PR #704, opened 2021-11-24
feat: removed placeholder animations to remove latency

rebelchris:
After some research we decided to remove the placeholder animations and show the plain placeholders.
We do this to improve performance of the application not forcing high GPU/CPU usage.

DD-200 #done


idoshamun (approved):
I love the fact that to introduce this change all you had to do is this.
Really shows the power of encapsulation.

PR #705, opened 2021-11-24
feat: added markdown support for the comments

rebelchris:
Added markdown powered comments.
The user is able to type markdown and we will parse it into html on the API side.

The user should also be able to edit and see the markdown again.

Out of scope:
Copy function for code blocks

Screenshots:

Edit comment:
Screenshot 2021-11-26 at 15 13 46

Rendered comment:
Screenshot 2021-11-26 at 15 13 39

Other elements:
Screenshot 2021-11-26 at 15 13 54

Other elements 2:
Screenshot 2021-11-26 at 15 14 01

DD-210 #done


packages/shared/src/components/comments/comments.module.css
```diff
@@ -1,6 +1,147 @@
...
+    .javascript .xml,
+    .tex .hljs-formula,
+    .xml .javascript,
+    .xml .vbscript,
+    .xml .css,
+    .xml .hljs-cdata)) {
+      opacity: 0.5;
+    }
+  }
+}
```
sshanzel, line 147: just needs another line for eof

packages/shared/src/components/comments/CommentMarkdown.tsx
```diff
@@ -0,0 +1,16 @@
+import React, { ReactElement } from 'react';
```
sshanzel, line 1: In time we might have another place where we will utilize html/markdown contents, so I think it is better to make the name more generalized. Also that this is not comment-specific.

One example is at profile details, soon enough we might allow markdown in their bio or somewhere else.

sshanzel (commented):
Just had some very minor comments, but what a great start on this awesome feature! 🚀 

Super excited about this to be in production! 🚢

packages/shared/src/components/comments/CommentMarkdown.tsx
rebelchris: Ah very good point Lee, thanks for this :D

rebelchris:
@sshanzel Would you mind taking another look?

packages/shared/src/components/Markdown.tsx
```diff
@@ -0,0 +1,16 @@
...
+
+export default function Markdown({
+  children,
+}: {
+  children: string;
+}): ReactElement {
+  return (
+    <div
+      className={styles.markdown}
+      dangerouslySetInnerHTML={{ __html: sanitize(children) }}
```
idoshamun, line 13: Love the fact that you remembered to sanitized the children

packages/shared/src/components/Markdown.tsx
```diff
@@ -0,0 +1,16 @@
+import React, { ReactElement } from 'react';
+import { sanitize } from 'dompurify';
+import styles from './markdown.module.css';
+
+export default function Markdown({
+  children,
+}: {
+  children: string;
```
idoshamun, line 8: I think that replacing the children's type or using it for the markdown is not a convention. It might be confusing.
Please use a different name.

packages/shared/src/components/markdown.module.css
```diff
@@ -0,0 +1,125 @@
+.markdown {
+  & :where(h1) {
```
idoshamun, line 2: Why do you use `& :where(x)` instead of `& x`?

packages/shared/src/components/markdown.module.css
```diff
@@ -0,0 +1,125 @@
...
+    @apply rounded-16;
+  }
+  & > :where(:first-child) {
+    @apply mt-0;
+  }
+  & > :where(:last-child) {
+    @apply mb-0;
+  }
+  & :where(pre) :global(:where(.hljs-comment, .hljs-title)) {
+    color: #989498;
```
idoshamun, line 63: Why do you use hard-coded colors here?

packages/shared/src/components/modals/DeleteCommentModal.spec.tsx
```diff
@@ -37,13 +37,6 @@ const renderComponent = (
   );
 };
 
-it('should close modal on cancel', async () => {
```
idoshamun, line 40: Why did you delete this test?

packages/shared/src/components/modals/DiscardCommentModal.spec.tsx
```diff
@@ -20,13 +20,6 @@ const renderComponent = (props: Partial<Props> = {}): RenderResult => {
   return render(<DiscardCommentModal {...defaultProps} {...props} />);
 };
 
-it('should close modal on cancel', async () => {
```
idoshamun, line 23: Why did you delete this test?

idoshamun (changes requested):
I don't see the "Markdown supported" link in the new comment modal as shown in the design.
Otherwise, this seems like an incredible work

packages/shared/src/components/markdown.module.css
rebelchris: More stricter selector in terms of resetting the `:first-child` and `:last-child` and child components.
Also enables us to target multiple at once for instance the `ul, ol` one.

I've taking this approach from Tailwind, after finding issues with resetting  child types.

packages/shared/src/components/markdown.module.css
rebelchris: Good question, just used a default color scheme that was available.
How would you suggest we go about it?

packages/shared/src/components/modals/DeleteCommentModal.spec.tsx
rebelchris: This has been deleted since the cancel button is now replaced with the "Markdown supported" link

packages/shared/src/components/markdown.module.css
idoshamun: You should advise @tsahimatsliah about the right colors to use.
You can send him the default palette and he'll send you back our palette.
It has to support light mode as well

packages/shared/src/components/markdown.module.css
idoshamun: Thanks for explaining :) Just make sure we have a proper browser support or that PostCSS compiles it to something every browser supports

idoshamun (changes requested):
Looks way better! Thanks :)

rebelchris:
Writing a subnote here:
Do not forget to do the API first and run the convert command 👏

idoshamun (approved):
Sweet!! Once we deploy the API I'll run the script and we can merge this

PR #708, opened 2021-11-30
feat: move feed customisation to a modal

rebelchris:
We decided to move the customisation for the feed to it's own modal.

This is as a separate pre-work task for the re layout.

Screenshot 2021-11-30 at 12 13 45


packages/shared/src/components/Settings.tsx
```diff
@@ -53,24 +48,15 @@ export default function Settings({
     { label: 'Light', value: 'light' },
     { label: 'Auto', value: 'auto' },
   ]);
+
   const Section = useMemo(
```
idoshamun, line 52: This can go outside now that it is not dependent on any variable

packages/shared/src/components/Settings.tsx
```diff
@@ -53,24 +48,15 @@ export default function Settings({
...
 
-  const SectionTitle = useMemo(
-    () =>
-      classed(
-        'h3',
-        'text-theme-label-tertiary mb-4 font-bold',
-        panelMode ? 'typo-callout' : 'typo-body',
-      ),
-    [panelMode],
+  const SectionTitle = classed(
```
idoshamun, line 57: This can go outside now that it is not dependent on any variable

packages/shared/src/components/modals/FeedSettingsModal.tsx
```diff
@@ -0,0 +1,28 @@
+import React, { ReactElement } from 'react';
+import dynamic from 'next/dynamic';
+import { ResponsiveModal } from './ResponsiveModal';
+import { ModalProps } from './StyledModal';
+import { ModalCloseButton } from './ModalCloseButton';
+
+const Settings = dynamic(
+  () => import(/* webpackChunkName: "settings" */ '../Settings'),
+);
```
idoshamun, line 9: We don't need to import it dynamically

idoshamun (changes requested):
Please check my comments and also fix the PR title to be more explicit :)

rebelchris:
@idoshamun Thanks! that even seems to have fixed the testing issue 😂
Also refactored to include Lee's amazing tooltip 🤘

idoshamun (approved):
Woohoo!

PR #710, opened 2021-11-30
refactor: hide publication date

sshanzel:
DD-340 #done

The separator now has 2 conditions for whether to be displayed or not.

The article publication date should also now be depending on our flagsmith value.

### DIFF
--- a/packages/shared/src/components/cards/PostCard.spec.tsx
+++ b/packages/shared/src/components/cards/PostCard.spec.tsx
@@ -1,10 +1,7 @@
 import React from 'react';
 import { render, RenderResult, screen, waitFor } from '@testing-library/react';
-import { act } from '@testing-library/react-hooks';
-import { IFlags } from 'flagsmith';
 import { Post } from '../../graphql/posts';
 import { PostCard, PostCardProps } from './PostCard';
-import { FeaturesContextProvider } from '../../contexts/FeaturesContext';
 
 const defaultPost: Post = {
   id: 'e3fd75b62cadd02073a31ee3444975cc',
@@ -61,15 +58,8 @@ beforeEach(() => {
   jest.clearAllMocks();
 });
 
-const renderComponent = (
-  props: Partial<PostCardProps> = {},
-  flags: IFlags = {},
-): RenderResult => {
-  return render(
-    <FeaturesContextProvider flags={flags}>
-      <PostCard {...defaultProps} {...props} />
-    </FeaturesContextProvider>,
-  );
+const renderComponent = (props: Partial<PostCardProps> = {}): RenderResult => {
+  return render(<PostCard {...defaultProps} {...props} />);
 };
 
 it('should call on link click on component left click', async () => {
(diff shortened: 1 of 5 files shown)


sshanzel:
Also, I noticed the max height for post card should actually be `366px`.

image

packages/shared/src/components/cards/PostCard.spec.tsx
```diff
@@ -107,10 +117,23 @@ it('should call on bookmark click on bookmark button click', async () => {
   );
 });
 
+it('should not display publication date based on features flags', async () => {
+  act(async () => {
```
idoshamun, line 121: Why do we need the `act` function here?

packages/shared/src/components/cards/PostCard.spec.tsx
```diff
@@ -107,10 +117,23 @@ it('should call on bookmark click on bookmark button click', async () => {
...
+    const el = await screen.findByText('Jun 13, 2018');
+    expect(el).not.toBeInTheDocument();
+  });
+});
+
 it('should format publication date', async () => {
-  renderComponent();
-  const el = await screen.findByText('Jun 13, 2018');
-  expect(el).toBeInTheDocument();
+  act(async () => {
```
idoshamun, line 132: Why do we need the `act` function here?

packages/shared/src/components/cards/PostMetadata.tsx
```diff
@@ -12,7 +25,9 @@ export default function PostMetadata({
   className?: string;
   children?: ReactNode;
 }): ReactElement {
+  const { flags } = useContext(FeaturesContext);
```
idoshamun, line 28: Exposing this component to context and especially to the `FeaturesContext` will create tight coupling between them. I think it's better to accept this flag as property from the outside.

idoshamun (changes requested):
A few fixes and we can ship :P

packages/shared/src/components/cards/PostCard.spec.tsx
sshanzel: I think some re-renders are triggered, and the tests were failing. After I applied the `act` to signify that rendering might take time, it worked.

sshanzel: Ah, my procedure was just wrong, which makes the act necessary.

idoshamun (approved):
Awesome!

rebelchris (approved):
Awesome work Lee!
And believe this is the first introduction of flagsmith on the monorepo right?

Looking forward to doing more of these little UI tests.

Subnote:
Had a giggle on their naming convention:
`IBulletTrainFeature`

idoshamun:
@rebelchris flagsmith was already there in several places but not that common

sshanzel:
Thank you for the reviews both!

sshanzel:
The feat has now been disabled. I will be merging this now. Thank you!

image

PR #713, opened 2021-12-01
fix: fix re focus state for tooltips

rebelchris:
There was an issue with the re-focus from closing modal which would always trigger the focus on tooltips.
To fix this I introduced a local variable that keeps track if the modal was the previous cause.

Do note we can not solve this with setState as this will not trigger since the component is technically still mounting.
Hate to use `let` here but it seems to be the only viable fix.

See video for resolution: (Tab still works, hover works and refocus doesn't show now)

https://user-images.githubusercontent.com/554874/144201177-42188cae-1dbf-47a5-8d02-54921528a041.mp4

### DIFF
--- a/packages/shared/src/components/tooltips/SimpleTooltip.tsx
+++ b/packages/shared/src/components/tooltips/SimpleTooltip.tsx
@@ -20,6 +20,7 @@ export function SimpleTooltip({
   content,
   ...props
 }: SimpleTooltipProps): ReactElement {
+  let shouldShow = true;
   const component = useMemo(
     () =>
       React.cloneElement(children, {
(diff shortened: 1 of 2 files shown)


packages/shared/src/components/tooltips/SimpleTooltip.tsx
```diff
@@ -29,11 +30,30 @@ export function SimpleTooltip({
...
+    }
+    shouldShow = true;
+  };
+
   return (
     <TippyTooltip
       shouldLoad={getShouldLoadTooltip()}
       {...props}
       content={content}
+      onClick
```
sshanzel, line 53: I think there is nothing declared in this scope as `onClick`.

sshanzel (changes requested):
Great work on this hotfix! Thank you so much! You did it!

packages/shared/src/components/tooltips/SimpleTooltip.tsx
rebelchris: Yep! 
Missed that in removing it, tried to bind to that first.

Will remove it

packages/shared/src/components/tooltips/SimpleTooltip.tsx
```diff
@@ -20,6 +20,7 @@ export function SimpleTooltip({
   content,
   ...props
 }: SimpleTooltipProps): ReactElement {
+  let shouldShow = true;
```
idoshamun, line 23: Can you add a comment here to explain that this resolves a focus issue?
Otherwise, we might get confused in the future.

idoshamun (changes requested):

packages/shared/src/components/tooltips/SimpleTooltip.tsx
rebelchris: Good point, will do 👏

rebelchris:
@idoshamun Would love to hear if this comment makes enough sense?

idoshamun (approved):
Perfect!

sshanzel (approved):
All perfect now! Just kindly check on the deployment before merging! 🚀

PR #715, opened 2021-12-01
refactor: forward ref of lazy image

sshanzel:
Tested locally using env variable for production api

### DIFF
--- a/packages/webapp/.env
+++ b/packages/webapp/.env
@@ -1,5 +1,5 @@
-NEXT_PUBLIC_API_URL=http://localhost:4000
-#NEXT_PUBLIC_API_URL=https://api.daily.dev
+#NEXT_PUBLIC_API_URL=http://localhost:4000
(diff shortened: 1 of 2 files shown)


packages/webapp/.env
```diff
@@ -1,5 +1,5 @@
-NEXT_PUBLIC_API_URL=http://localhost:4000
-#NEXT_PUBLIC_API_URL=https://api.daily.dev
+#NEXT_PUBLIC_API_URL=http://localhost:4000
+NEXT_PUBLIC_API_URL=https://api.daily.dev
```
idoshamun: Please revert this

idoshamun (approved):

packages/webapp/.env
sshanzel: This should be gone now. Thank you!

rebelchris (approved):
Nice catch! 👀

PR #716, opened 2021-12-01
fix: added optional padding to the responsive modal

rebelchris:
We re-use the responsive modal quite a lot, which is great.

However in some cases we want to remove the main modal's padding as this is per design for the new modal types.
There are quite a lot of these new ones coming, so decided to abstract by adding a padding parameter which is defaulted to true (as still most ones need the padding now)

Example of the first one that doesn't need it is actually the new FeedSettingsModal, but also the UpvotedPopupModal already had a hotfix for this issue.


idoshamun (approved):
Sweet!

PR #718, opened 2021-12-02
refactor: profile heatmap tooltip

sshanzel:
In this commit, I have included the customization capability of the tooltip which will also be utilized in the Profile Tooltip.

### DIFF
--- a/packages/shared/src/components/tooltips/BaseTooltip.tsx
+++ b/packages/shared/src/components/tooltips/BaseTooltip.tsx
@@ -58,6 +58,7 @@ export function BaseTooltip(
       onHide={onHide}
       delay={delay}
       allowHTML
+      animation={!(disableInAnimation && disableOutAnimation)}
       content={
         <BaseTooltipContainer
           {...container}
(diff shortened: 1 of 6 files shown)


idoshamun (approved):

packages/shared/src/components/tooltips/BaseTooltip.tsx
```diff
@@ -51,14 +58,20 @@ export function BaseTooltip(
       onHide={onHide}
       delay={delay}
       allowHTML
+      animation={!(disableInAnimation && disableOutAnimation)}
```
rebelchris, line 61: It can't ever hit this outAnimation because you return if it exists.

rebelchris (approved):
Nice work, really like the addition of the container padding and border-radius :)
Just one small note.

packages/shared/src/components/tooltips/BaseTooltip.tsx
sshanzel: Yes, it won't but this is for tippy's internal delay for animation which has to be applied here as well.

sshanzel:
Finished doing sanity checks. WIll be merging now. Thank you both!

packages/shared/src/components/tooltips/BaseTooltip.tsx
sshanzel: That return was to avoid our custom animation.

PR #721, opened 2021-12-02
feat: sidebar toggle mode

rebelchris:
This first PR introduces the new sidebar menu so we can start working on filling it up in the other tasks.
I did not introduce specific test, since they will change with the introduction of the other tasks.

[ ] Added a task in Jira to write extended tasks
[ ] Mobile version is a separate task
[ ] Bottom menu is separate task

Main point is we want to use this PR so we can start splitting up the other tasks.


packages/shared/src/components/MainLayout.tsx
```diff
@@ -46,31 +43,22 @@ export default function MainLayout({
...
-  );
-
   return (
     <>
       {!responsive && <PromotionalBanner />}
       <header
         className={`${
           styles.header
-        } relative flex items-center px-4 border-b border-theme-divider-tertiary tablet:px-8 laptop:px-4 ${
+        } relative flex items-center px-4 border-b border-theme-divider-tertiary tablet:px-8 laptop:px-4 py-2 ${
```
sshanzel, line 61: This is just an improvement in the existing code, I think we can turn this to using `classnames` instead of template strings.

packages/shared/src/components/MainLayout.tsx
```diff
@@ -111,47 +99,28 @@ export default function MainLayout({
...
-        {showRank && windowLoaded && (
-          <HeaderRankProgress className="absolute right-0 -bottom-px left-0 z-rank my-0 mx-auto transform translate-y-1/2" />
-        )}
       </header>
-      {children}
+      <main className="flex flex-row">
+        <aside
+          className="bg-theme-float transition-all transform duration-500 ease-in-out"
+          onClick={() => toggleOpenSidebar()}
+          style={{ width: openSidebar ? 240 : 44 }}
```
sshanzel, line 118: I think there are classes suitable for this width - namely `w-60` and `w-11`. Just to keep things coming from one place.

sshanzel (changes requested):
So far so good! Just a minor improvement and a very minor fix but nothing major and not a blocker.

Well done on this! 🚢

packages/shared/src/components/FeedLayout.tsx
```diff
@@ -81,18 +82,78 @@ const baseFeedSettings: FeedContextData[] = [
   },
 ];
 
+const sidebarOpenFeedSettings: FeedContextData[] = [
```
idoshamun, line 85: Is there any correlation between these settings and the original ones?
If so maybe we can code it. It looks like every element here is `i-1` in the previous array

packages/shared/src/components/Sidebar.tsx
```diff
@@ -1,105 +1,173 @@
...
+import Link from 'next/link';
+import SettingsContext from '../contexts/SettingsContext';
+import { Button } from './buttons/Button';
+import { FeedSettingsModal } from './modals/FeedSettingsModal';
+import ArrowIcon from '../../icons/arrow.svg';
+import HotIcon from '../../icons/hot.svg';
+import FilterIcon from '../../icons/filter.svg';
+import UpvoteIcon from '../../icons/upvote.svg';
+import SearchIcon from '../../icons/magnifying.svg';
+import DisqusIcon from '../../icons/comment.svg';
```
idoshamun, line 11: `Disqus` is this a typo?

packages/shared/src/components/Sidebar.tsx
```diff
@@ -1,105 +1,173 @@
...
-          {hightlightTrigger && (
-            <div
-              className="absolute left-0 -z-1 w-14 bg-theme-hover rounded-r-3xl"
-              style={{
-                top: '-1rem',
-                height: '5.5rem',
-                animation: 'rank-attention 2s infinite ease-in-out',
-              }}
+          <Button
+            onClick={() => toggleOpenSidebar()}
```
idoshamun, line 113:
```suggestion
            onClick={toggleOpenSidebar}
```

packages/shared/src/components/Sidebar.tsx
```diff
@@ -1,105 +1,173 @@
...
-        (feedSettings?.includeTags?.length ||
-          feedSettings?.blockedTags?.length ||
-          feedSettings?.excludeSources?.length ||
-          feedSettings?.advancedSettings?.length)
-      ) {
-        updateAlerts({ filter: false });
-      }
-    }
-  };
+  const ItemInner = ({ item }: { item: SidebarMenuItem }) => (
```
idoshamun, line 88: This can go outside the component

packages/shared/src/components/Sidebar.tsx
```diff
@@ -1,105 +1,173 @@
...
+                      type="button"
+                      className="flex flex-1 items-center px-3 h-7"
+                      onClick={item.action}
+                    >
+                      <ItemInner item={item} />
+                    </button>
+                  )}
+                </li>
+              ))}
+            </ul>
```
idoshamun, line 161: This is pretty hard to read. I suggest refactoring :)

idoshamun (changes requested):
Quite some work you did here.
I suggested some changes :)

packages/shared/src/components/FeedLayout.tsx
rebelchris: The page size is a bit random sometimes 2, sometimes 3

packages/shared/src/components/Sidebar.tsx
rebelchris: Then we should pass openSidebar to it, that's the only reason it's sitting inside now.
Not sure what's the better one here in terms of performance?

idoshamun: My personal preference is to take out everything that has little to no use in the closure and scope of the parent.
Make everything more readable.

idoshamun (approved):
This is a treat for my eyes now 🍬

sshanzel (approved):
Everything looks good to me now! 🚀

PR #722, opened 2021-12-02
feat: remove all responsive props on the MainLayout

rebelchris:
We had some unused occurrences of the responsive prop for the MainLayout.
This actually broke the layout for the reading history page, as it was not passed.

Since we don't use it anywhere, we decided to remove this prop.
This only caused jest not to acknowledge this component use, so decided to render an empty once in Jest only.

Sidenote:
This is also needed for the re-layout

### DIFF
--- a/packages/shared/src/components/PromotionalBanner.tsx
+++ b/packages/shared/src/components/PromotionalBanner.tsx
@@ -8,10 +8,11 @@ import { apiUrl } from '../lib/config';
 import { BANNER_QUERY, BannerData } from '../graphql/banner';
 import ProgressiveEnhancementContext from '../contexts/ProgressiveEnhancementContext';
 import usePersistentState from '../hooks/usePersistentState';
+import { isTesting } from '../lib/constants';
 
 export default function PromotionalBanner(): ReactElement {
   // Disable this component in Jest environment
-  if (typeof jest !== 'undefined') {
+  if (isTesting) {
     return <></>;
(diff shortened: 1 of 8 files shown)


packages/shared/src/components/PromotionalBanner.tsx
```diff
@@ -10,6 +10,10 @@ import ProgressiveEnhancementContext from '../contexts/ProgressiveEnhancementCon
 import usePersistentState from '../hooks/usePersistentState';
 
 export default function PromotionalBanner(): ReactElement {
+  // Disable this component in Jest environment
+  if (typeof jest !== 'undefined') {
```
sshanzel, line 14: I think we can use the constant variable we have here for this case:

https://github.com/dailydotdev/apps/blob/f687a80cbd1492fe316558638a8638edefc8824c/packages/shared/src/lib/constants.ts#L16-L17

sshanzel (changes requested):
Thank you for doing it across the app!

Though I have one question, why do we disable it in test environment?

packages/shared/src/components/PromotionalBanner.tsx
rebelchris: Well spotted! Should have used that one! 👀

sshanzel (approved):
LGTM! 🚀

idoshamun (approved):

PR #723, opened 2021-12-03
feat: profile tooltip

sshanzel:
DD-36 #done

I have included a tailwind [extended config](https://github.com/tailwindlabs/tailwindcss/discussions/2727#discussioncomment-686642) that I used to use which has now been merged. But since we haven't upgraded yet. I am applying it explicitly.

New component:
 - `ProfileTooltip` - this has all the necessary styling and configuration for a custom tooltip.

Details that must be present in the tooltip:
 - Profile picture (rank to be on the lower right corner) - must be a link
 - Name - must be a link
 - Handle - must be a link
 - Bio

I have also refactored the `ProfileLink` component which now doesn't have a `tooltip` (and `disableTooltip` props) out of the box. We made the `tooltip` component a wrapper which if you don't really need it, then just don't wrap the `children` in itself.

I also applied a lazy mechanism in our `BaseTooltip` as I noticed the content property is rendered whenever its target component is available. This impacts the performance greatly if a single page has many tooltips and custom ones.

### DIFF
--- a/packages/shared/src/components/TagLinks.tsx
+++ b/packages/shared/src/components/TagLinks.tsx
@@ -1,0 +1,46 @@
+import React, { ReactElement } from 'react';
+import classNames from 'classnames';
+import Link from 'next/link';
+import { Button } from './buttons/Button';
+
+interface TagLinkProps {
+  tag: string;
+  className?: string;
+}
+
+export function TagLink({ tag, className }: TagLinkProps): ReactElement {
+  return (
+    <Link href={`/tags/${tag}`} passHref key={tag}>
+      <Button
+        tag="a"
+        className={classNames('btn-tertiaryFloat xsmall', className)}
+      >
+        #{tag}
+      </Button>
+    </Link>
+  );
+}
+
+interface TagLinksProps {
+  tags: string[];
+  className?: string;
+  tagClassName?: string;
+}
+
+export function TagLinks({
+  tags,
+  className,
+  tagClassName,
+}: TagLinksProps): ReactElement {
+  if (tags.length === 0) {
+    return null;
+  }
+
+  return (
+    <div className={classNames('flex flex-wrap gap-2 mt-3 mb-4', className)}>
+      {tags.map((tag) => (
+        <TagLink key={tag} tag={tag} className={tagClassName} />
+      ))}
+    </div>
+  );
+}
--- a/packages/shared/src/components/comments/SubComment.tsx
+++ b/packages/shared/src/components/comments/SubComment.tsx
@@ -1,6 +1,5 @@
 import React, { ReactElement } from 'react';
 import classNames from 'classnames';
-import styles from './comments.module.css';
 import { Comment } from '../../graphql/comments';
 import { CommentBox, CommentPublishDate } from './common';
 import CommentActionButtons, {
@@ -55,7 +54,7 @@ export default function SubComment({
         <SubCommentBox>
           <CommentAuthor postAuthorId={postAuthorId} author={comment.author} />
           <CommentPublishDate comment={comment} />
-          <div className={classNames(styles.commentBox, 'mt-2')}>
+          <div className="mt-2">
             <Markdown content={comment.contentHtml} />
           </div>
         </SubCommentBox>
--- a/packages/shared/src/components/comments/SubComment.tsx
+++ b/packages/shared/src/components/comments/SubComment.tsx
@@ -1,5 +1,6 @@
 import React, { ReactElement } from 'react';
 import classNames from 'classnames';
+import styles from './comments.module.css';
 import { Comment } from '../../graphql/comments';
 import { CommentBox, CommentPublishDate } from './common';
 import CommentActionButtons, {
@@ -42,13 +43,17 @@ export default function SubComment({
             lastComment ? 'h-4' : 'bottom-0',
           )}
         />
-        <ProfileImageLink className="w-8 h-8" user={comment.author} />
+        <ProfileImageLink
+          user={comment.author}
+          picture={{ size: 'medium' }}
+          tooltip={{ content: comment.author.username }}
+        />
       </div>
       <div className="flex flex-col flex-1 items-stretch ml-2">
         <SubCommentBox>
           <CommentAuthor postAuthorId={postAuthorId} author={comment.author} />
           <CommentPublishDate comment={comment} />
-          <div className="mt-2">
+          <div className={classNames(styles.commentBox, 'mt-2')}>
             <Markdown content={comment.contentHtml} />
           </div>
         </SubCommentBox>
--- a/packages/shared/src/components/comments/SubComment.tsx
+++ b/packages/shared/src/components/comments/SubComment.tsx
@@ -10,6 +10,7 @@ import { ProfileImageLink } from '../profile/ProfileImageLink';
 import CommentAuthor from './CommentAuthor';
 import classed from '../../lib/classed';
 import Markdown from '../Markdown';
+import { ProfileTooltip } from '../profile/ProfileTooltip';
 
 export interface Props extends CommentActionProps {
   comment: Comment;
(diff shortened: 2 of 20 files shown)


sshanzel:
Preview:
image

packages/shared/src/components/profile/ProfileTolltipContent.tsx
```diff
@@ -0,0 +1,54 @@
...
+      <ProfileLink
+        className="mt-5 font-bold text-theme-label-primary"
+        user={user}
+      >
+        {user.name}
+      </ProfileLink>
+      <ProfileLink className="text-theme-label-secondary" user={user}>
+        @{user.username}
+      </ProfileLink>
+      <p className="mt-3 line-clamp-3 text-theme-label-tertiary">{user.bio}</p>
```
rebelchris, line 51: Maybe only render this conditionally?

packages/shared/src/components/comments/SubComment.tsx
```diff
@@ -42,13 +44,18 @@ export default function SubComment({
...
+            picture={{ size: 'medium' }}
+          />
+        </ProfileTooltip>
       </div>
       <div className="flex flex-col flex-1 items-stretch ml-2">
         <SubCommentBox>
           <CommentAuthor postAuthorId={postAuthorId} author={comment.author} />
           <CommentPublishDate comment={comment} />
-          <div className="mt-2">
+          <div className={classNames(styles.commentBox, 'mt-2')}>
```
rebelchris, line 58: What is added here that the commentBox is needed?

rebelchris (approved):
Wow really nice work @sshanzel ! 😃

packages/shared/src/components/profile/ProfileTolltipContent.tsx
sshanzel: Good idea! Let me work on that one 🚀

packages/shared/src/components/comments/SubComment.tsx
sshanzel: It was added before this line, but since the tooltip is inside of it, the selector inside that CSS module is getting applied to the tooltip content unnecessarily. It turns the links inside the tooltip as blue since the links don't have the className `commentAuthor` and I think it is safe to say that the actual comment box starts from this point.

sshanzel: It was added before this line, but since the tooltip is inside of it, the selector inside that CSS module is getting applied to the tooltip content unnecessarily. It turns the links inside the tooltip as blue since the links don't have the className `commentAuthor` and I think it is safe to say that the actual comment box starts from this point.

Long story short, this was a pre-existing class for links inside the comment box. Please advise if this should be removed now since we have the markdown.

idoshamun: @rebelchris your thoughts?

idoshamun:
@sshanzel you can go ahead and update our Tailwind version. It's a better solution if possible :)

packages/shared/src/components/comments/common.tsx
```diff
@@ -18,7 +17,8 @@ export function CommentPublishDate({
   );
 }
 
-export const commentBoxClassNames = `py-3 px-4 bg-theme-bg-secondary rounded-lg break-words typo-callout ${styles.commentBox}`;
```
idoshamun, line 21: Did you drop the `commentBox` class on purpose?

packages/shared/src/components/tooltips/LinkWithTooltip.tsx
```diff
@@ -18,7 +18,7 @@ export function LinkWithTooltip({
     () =>
       React.cloneElement(children, {
         ...children.props,
-        'aria-label': tooltip.content,
+        'aria-label': typeof tooltip.content === 'string' && tooltip.content,
```
idoshamun, line 21: Otherwise it might set `aria-label` to false

```suggestion
        'aria-label': typeof tooltip.content === 'string' ? tooltip.content : undefined,
```

idoshamun (changes requested):
Wow! Looks great.
Let's wait for the team response about the content of this tooltip. In the meantime, you can apply my requested changes

packages/shared/src/components/comments/SubComment.tsx
rebelchris: I don't think we should need this anymore as the markdown should have all the comments now.
We don't use the old slugify function anymore.

If it's being overruled we should fixed it in the markdown styling rather.

sshanzel: Sounds great. Especially since the CSS module only contains this specific styling. One less file for CSS once we remove this one. Thanks for the thoughts both of you!

packages/shared/src/components/comments/common.tsx
sshanzel: Yes, I had to apply it to a specific element to avoid unnecessary styling on some other parts but this should now be obsolete since the only content of this styling is the link-related we have discussed above.

sshanzel:
Thanks for the thoughts to both of you! Let me check if there will be any breaking changes that would require fixes for existing implementations on upgrading our tailwind version 🚀 .

sshanzel:
After further checking, the extended config we have here will only be available at tailwindcss version 3. It is currently at the `pre-release` stage. Once it gets released officially, I will create a task for migration.

sshanzel:
Once [this PR](https://github.com/dailydotdev/daily-api/pull/819) gets merged. The staging environment should work as intended for this feature. With that, this should still be reviewable.

idoshamun:
Before jumping into the review I highly suggest waiting for the data to load before showing the tooltip.
It creates a content overflow once you first show the tooltip and then add the tags when they get loaded.

packages/shared/src/components/TagLinks.tsx
```diff
@@ -0,0 +1,46 @@
...
+import { Button } from './buttons/Button';
+
+interface TagLinkProps {
+  tag: string;
+  className?: string;
+}
+
+export function TagLink({ tag, className }: TagLinkProps): ReactElement {
+  return (
+    <Link href={`/tags/${tag}`} passHref key={tag}>
```
idoshamun, line 13: I'm not sure if this should really be a link. You might want to ask it on the #product channel

packages/shared/src/graphql/users.ts
```diff
@@ -33,6 +38,18 @@ export const USER_READING_RANK_QUERY = gql`
...
 `;
 
+export const USER_TOOLTIP_CONTENT_QUERY = gql`
+  query UserTooltipContent($id: ID!) {
+    rank: userReadingRank(id: $id) {
+      currentRank
+    }
+    tags: userMostReadTags(id: $id) {
+      value
+      count
```
idoshamun, line 48: We don't use count property actually

idoshamun (changes requested):
Love the fact that you merged the GraphQL queries. You gotta love GraphQL 💪

packages/shared/src/components/TagLinks.tsx
sshanzel: Yes, [Tsahi has confirmed](https://dailydotdev.atlassian.net/wiki/spaces/HAN/pages/26280164/Profile+Tooltip?focusedCommentId=112394253#comment-112394253) that this should be a link to allow users to navigate to that specific tag.

packages/shared/src/graphql/users.ts
sshanzel: Ah, yes. I shall remove this part.

sshanzel:
GraphQL all the way!

Thank you for the review, I have now made the necessary changes for fetching the data first.

idoshamun (changes requested):
The code looks great but I did find an issue that I shared on Slack

idoshamun (changes requested):
It still flickers...

idoshamun (approved):
That's great! Eventually this problem helped you utilize better tippy :)

sshanzel:
I have to agree. This made things so much easier to manage.

PR #724, opened 2021-12-06
feat: temporary popup sidebar

sshanzel:
DD-347

As part of the relayout, the desired filtering should work differently from this PR but we will be shipping the custom feed separately and we need temporary access for filtering the feed itself.


sshanzel:
This should now be ready for review. Whew. Finally the tests have turned 🟢

packages/shared/src/components/Sidebar.tsx
```diff
@@ -193,6 +221,10 @@ export default function Sidebar(): ReactElement {
           onRequestClose={() => setShowSettings(false)}
         />
       )}
+      <FeedFilters
```
idoshamun, line 224: This can be loaded dynamically I think to reduce initial bundle size

packages/shared/src/components/filters/FeedFilters.tsx
```diff
@@ -0,0 +1,58 @@
...
+import FilterIcon from '../../../icons/arrow.svg';
+
+const asideWidth = sizeN(89);
+
+interface FeedFiltersProps {
+  isOpen?: boolean;
+  onBack?: () => void;
+}
+
+let timeout: NodeJS.Timeout;
```
idoshamun, line 15: This is a better typing as you currently rely on a node runtime which is actually not the right one.
Our runtime is the browser not node.

On another note, you should not define global variables for hooks. You need to use `useRef` for this.
There are many examples in this codebase and on the internet on how to use `setTimeout` with React Hooks. Give it a look, not that trivial.

```suggestion
let timeout: ReturnType<typeof setTimeout>;
```

packages/shared/src/components/filters/FeedFilters.tsx
```diff
@@ -0,0 +1,58 @@
...
+        transition: 'transform 0.2s linear 0.1s',
+      }}
+    >
+      <ClickableText
+        className="p-4 font-semibold no-underline text-theme-label-tertiary"
+        onClick={onBack}
+      >
+        <FilterIcon className="mr-2 text-2xl -rotate-90" />
+        Back
+      </ClickableText>
```
idoshamun, line 53: I'm not sure if it should be a `ClickableText` or a `Button` of sorts

packages/shared/src/components/filters/FilterRedDot.tsx
```diff
@@ -1,6 +1,6 @@
 import React, { ReactElement, useContext } from 'react';
 import SettingsIcon from '../../../icons/settings.svg';
-import RedDot from '../RedDot';
+import { RedDot } from '../notifs';
```
idoshamun, line 3: Please change `notifs` to something more explicit

idoshamun (changes requested):
Works smoothly but code needs some adjustments

packages/shared/src/components/Sidebar.spec.tsx
```diff
@@ -1,23 +1,46 @@
...
+import { FEED_SETTINGS_QUERY } from '../graphql/feedSettings';
+import { getFeedSettingsQueryKey } from '../hooks/useMutateFilters';
+import AlertContext, { AlertContextData } from '../contexts/AlertContext';
+import OnboardingContext from '../contexts/OnboardingContext';
+import { waitForNock } from '../../__tests__/helpers/utilities';
 
+let client: QueryClient;
+const updateAlerts = jest.fn();
 const toggleOpenSidebar = jest.fn();
+const incrementOnboardingStep = jest.fn();
```
rebelchris, line 22: Be aware we plan to remove the onboarding all together.
See slack discussion on Team Engineering

packages/shared/src/components/Sidebar.tsx
```diff
@@ -2,6 +2,7 @@ import React, { ReactElement, ReactNode, useContext, useState } from 'react';
 import Link from 'next/link';
 import SettingsContext from '../contexts/SettingsContext';
 import { Button } from './buttons/Button';
+import { BlueDot } from './notifs';
```
rebelchris, line 5: Thinking maybe we should rename this to AlertDot and have the option to define a color?
Just thinking as this will become more a generic component it might be worthwhile changing it now.

packages/shared/src/components/Sidebar.spec.tsx
```diff
@@ -36,25 +59,56 @@ const renderComponent = (
...
+              <Sidebar />
+            </OnboardingContext.Provider>
+          </SettingsContext.Provider>
+        </AuthContext.Provider>
+      </AlertContext.Provider>
+    </QueryClientProvider>,
   );
 };
 
+it('should remove red dot for filter alert when there is a pre-configured feedSettings', async () => {
```
rebelchris, line 97:
```suggestion
it('should remove alert dot for filter alert when there is a pre-configured feedSettings', async () => {
```

packages/shared/src/components/Sidebar.tsx
```diff
@@ -25,9 +29,10 @@ interface SidebarMenuItem {
   title: string;
   path?: string;
   action?: () => unknown;
+  notif?: ReactElement;
```
rebelchris, line 32: Would suggest to rename to `notification` just to keep everything written out in the codebase.

packages/shared/src/components/Sidebar.tsx
```diff
@@ -106,16 +114,36 @@ const NavItem = classed(
...
+  const showFilters = () => {
+    if (
+      alerts?.filter &&
+      (feedSettings?.includeTags?.length ||
+        feedSettings?.blockedTags?.length ||
+        feedSettings?.excludeSources?.length ||
+        feedSettings?.advancedSettings?.length)
+    ) {
+      updateAlerts({ filter: false });
+    }
```
rebelchris, line 131: We are now loading the complete feedSettings here, maybe we should move this to the opening of the feedFilter as that will already load them as well?
What do you think?

Note: We can keep the alerts context for the dot.

packages/shared/src/components/Sidebar.tsx
```diff
@@ -106,16 +114,36 @@ const NavItem = classed(
...
+  };
 
   const topMenuItems: SidebarMenuItems[] = [
     {
       key: 'Discover',
       items: [
         {
           icon: <ListIcon Icon={FilterIcon} />,
+          notif: alerts.filter && (
+            <BlueDot style={{ top: `-${0.125}rem`, right: `-${0.125}rem` }} />
```
rebelchris, line 143: Tailwind has classes for these values:

```jsx
-top-0.5
-right-0.5
```

packages/shared/src/components/filters/FeedFilters.tsx
```diff
@@ -0,0 +1,58 @@
...
+      className={classNames(
+        'fixed top-14 left-0 z-3 bottom-0 self-stretch bg-theme-bg-primary rounded-r-2xl border-t border-r border-theme-divider-primary overflow-y-auto',
+        !isOpen && 'pointer-events-none',
+        hidden && 'invisible',
+      )}
+      style={{
+        width: asideWidth,
+        transform: isOpen ? 'translateX(0)' : `translateX(-${asideWidth})`,
+        willChange: 'transform',
+        transition: 'transform 0.2s linear 0.1s',
```
rebelchris, line 44: I think we can use these directly from Tailwind.

packages/shared/src/components/filters/FeedFilters.tsx
rebelchris: To me it should be a button I think,
The same as we use in the multiLevelMenu

packages/shared/src/components/filters/FilterRedDot.tsx
rebelchris: And change RedDot to AlertDot and accept the color attribute

rebelchris (changes requested):
Well done Lee!
Really like this piece of work!

Just a side note, 
I had some really back UX because of the duplicate back buttons, maybe we should raise this with Tsahi.

I clicked the top back button thinking I would go back one page, but it closes the whole thing.

Screenshot 2021-12-07 at 10 03 45

packages/shared/src/components/Sidebar.tsx
sshanzel: I was thinking about that as well. Though I think it would be hard to have a clear definition of the alerts since let's say, for example, the blue one, there can be multiple shades of blue and for instance colors other than `fill` (current blue) will be used.

I think it would be best for @tsahimatsliah to define this and create a guideline if necessary.

packages/shared/src/components/Sidebar.tsx
sshanzel: I agree or we can probably use `alert` and reserve the `notification` for in-app notifs. Thanks for the thought!

packages/shared/src/components/Sidebar.tsx
sshanzel: I agree. This is the better way. I will work on this in the upcoming commits.

packages/shared/src/components/Sidebar.tsx
sshanzel: Hmm. Let me try that again, I used it before it wasn't making any effect and had to do it this way.

packages/shared/src/components/filters/FeedFilters.tsx
sshanzel: Oh, nice. Will look these up! Thank you!

packages/shared/src/components/Sidebar.tsx
sshanzel: That makes sense though I will have to add another mechanism to display the animation.

packages/shared/src/components/filters/FeedFilters.tsx
sshanzel: Alright, let me change this one to a `Button`

packages/shared/src/components/Sidebar.tsx
rebelchris: Oh might be because the alert has a default top offset already.

packages/shared/src/components/filters/FeedFilters.tsx
sshanzel: I'm gonna look it up! Thank you for this!

packages/shared/src/components/filters/FilterRedDot.tsx
sshanzel: Let me change this to `AlertDot` for now. The color attribute can be tricky at some point in the future, let's consult @tsahimatsliah on that part.

packages/shared/src/components/Sidebar.spec.tsx
sshanzel: Thank you! I will make sure to avoid these cases.

packages/shared/src/components/Sidebar.tsx
sshanzel: Or why not we use the ones for status since we already have those defined.

sshanzel:
This should now be ready for review.

Also, I have created a utility for testing to wait on rerendering/useEffects to run first before the assertions happen again. It can be noticed from [this commit](https://github.com/dailydotdev/apps/pull/724/commits/960d28575062536ef1fdb2c55a3fcc237a295ef7) the test has failed since the `useEffect` was not yet triggered. After the next commit, it passed.

This includes the change for the `close` button that was discussed with Tsahi 🚢 

Thank you both for your wonderful reviews! @idoshamun @rebelchris

packages/shared/__tests__/helpers/utilities.ts
```diff
@@ -1,6 +1,10 @@
+import { act } from '@testing-library/react-hooks';
 import { waitFor } from '@testing-library/react';
 import nock from 'nock';
 
 export const expectNockDone = (): void => expect(nock.isDone()).toBeTruthy();
 
 export const waitForNock = (): Promise<void> => waitFor(expectNockDone);
+
+export const waitForRerender = (ms = 100): Promise<void> =>
+  act(() => new Promise((resolve) => setTimeout(resolve, ms)));
```
rebelchris, line 10: Is this reliable? How do we know the 100ms is enough?
Just thinking ahead will this cause issues down the line if the codebase is even bigger?

rebelchris (approved):
Nice!

Really well done on making this super uniform!
Love the AlertDot.

Looks good to me though, just some small questions (non blocking)

packages/shared/__tests__/helpers/utilities.ts
sshanzel: I have tried 1ms earlier but it seems to have not worked. Now I tried it again. It worked. Oh, tests 😅 

Intuitively, `1ms` should work since our only goal is to put the things we need to test at the end of the event loop (after use effect).

sshanzel: Since the sleep-like utility felt like a hack than the actual test to use. I tried other ways and properly use the `act` API and it worked!

Thank you for your awesome comment! 🚀

packages/shared/src/components/filters/FeedFilters.tsx
```diff
@@ -12,46 +20,63 @@ interface FeedFiltersProps {
...
       return;
     }
 
-    timeout = setTimeout(() => setHidden(true), 300);
-  }, [isOpen]);
+    if (hidden) {
+      return;
+    }
+
+    timeoutRef.current = window.setTimeout(() => setHidden(true), 300);
```
idoshamun, line 57: We have to clear timeout when done
```suggestion
    timeoutRef.current = window.setTimeout(() => setHidden(true), 300);
    return () => timeoutRef.current && clearTimeout(timeoutRef.current);
```

packages/shared/src/components/filters/FeedFilters.tsx
```diff
@@ -12,46 +20,63 @@ interface FeedFiltersProps {
...
-      if (timeout) clearTimeout(timeout);
+      if (
+        alerts?.filter &&
+        (feedSettings?.includeTags?.length ||
+          feedSettings?.blockedTags?.length ||
+          feedSettings?.excludeSources?.length ||
+          feedSettings?.advancedSettings?.length)
+      ) {
+        updateAlerts({ filter: false });
+      }
```
idoshamun, line 42: Don't have this update statement in another component or a hook as well?

idoshamun (changes requested):

packages/shared/src/components/filters/FeedFilters.tsx
sshanzel: We don't but I think it would be a good idea to simplify this. Let me commit one thing.

idoshamun (approved):
Love this refactor :)

PR #725, opened 2021-12-06
fix: dropdown content overflow

sshanzel:
DD-351 #done

As reported [in this issue](https://github.com/dailydotdev/daily/issues/373), the content of our `Dropdown` component overflows or breaks up into two lines:
image

We applied truncating of the content and both cases should now be fixed. The `truncate` class has the following properties:
```
overflow: hidden;
text-overflow: ellipsis;
white-space: nowrap;
```

Preview of the fix:

Screen Shot 2021-12-06 at 2 39 30 PM

### DIFF
--- a/packages/shared/src/components/fields/Dropdown.tsx
+++ b/packages/shared/src/components/fields/Dropdown.tsx
@@ -99,10 +99,7 @@ export function Dropdown({
   };
 
   return (
-    <div
-      className={classNames('truncate', styles.dropdown, className)}
-      {...props}
-    >
+    <div className={classNames(styles.dropdown, className)} {...props}>
       <button
         type="button"
         ref={triggerRef}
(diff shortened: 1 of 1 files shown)


rebelchris (approved):
Nice work! And a one line fix, awesome 👏

idoshamun:
I think it's better to show also the three dots. I think the `ellipsis` should do it by default.
Maybe something is not 100% right with applied CSS or something is missing.
The `ellipsis` behavior can be a mystery.

sshanzel:
@idoshamun I agree `ellipsis` should have taken care of that. Though in our guideline for exceeding characters with input fields it does just get sliced off as [shown here](https://dailydotdev.slack.com/archives/C01L0QXQJNB/p1637589385001400). What d you think?

idoshamun:
So let's leave it as it is for now :)

idoshamun (approved):

PR #726, opened 2021-12-06
fix: fixed a testing case for brave browser

rebelchris:
In chrome we use the default new tab as: `chrome-search://local-ntp/local-ntp.html`.
Brave however doesn't provide a inner tab url, so we have to resolve to setting the chrome search default like so: `chrome://new-tab-page`.

I've added a constant check for brave browser.
Running it in Chrome:
Screenshot 2021-12-06 at 11 34 41

And in Brave:
Screenshot 2021-12-06 at 11 34 35

### DIFF
--- a/packages/extension/src/newtab/dnd.ts
+++ b/packages/extension/src/newtab/dnd.ts
@@ -20,9 +20,7 @@ const BRAVE_DEFAULT_URL = 'chrome://new-tab-page';
 
 const browserTest = () =>
   process.env.TARGET_BROWSER === 'chrome' ? CHROME_DEFAULT_URL : DEFAULT_URL;
-const braveTest = () => (isBrave() ? BRAVE_DEFAULT_URL : browserTest());
-
-export const getDefaultLink = (): string => braveTest();
+const getDefaultLink = () => (isBrave() ? BRAVE_DEFAULT_URL : browserTest());
 interface DndOption<T extends TimeFormat> {
   value: number;
   label: string;
--- a/packages/extension/src/newtab/dnd.ts
+++ b/packages/extension/src/newtab/dnd.ts
@@ -1,3 +1,4 @@
+import { isBrave } from '@dailydotdev/shared/src/lib/constants';
 import { addDays, addHours, addMinutes } from 'date-fns';
 
 export type TimeFormat =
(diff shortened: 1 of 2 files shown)


packages/extension/src/newtab/dnd.ts
```diff
@@ -15,9 +16,14 @@ export enum CustomTime {
...
 const DEFAULT_URL = 'https://www.google.com';
 const CHROME_DEFAULT_URL = 'chrome-search://local-ntp/local-ntp.html';
+const BRAVE_DEFAULT_URL = 'chrome://new-tab-page';
 
-export const getDefaultLink = (): string =>
-  process.env.TARGET_BROWSER === 'chrome' ? CHROME_DEFAULT_URL : DEFAULT_URL;
+export const getDefaultLink = (): string => {
+  const browserTest = () =>
+    process.env.TARGET_BROWSER === 'chrome' ? CHROME_DEFAULT_URL : DEFAULT_URL;
+  const braveTest = () => (isBrave() ? BRAVE_DEFAULT_URL : browserTest());
```
idoshamun, line 24: Looks like we can take this out of the function.
Functions generating more functions inside usually utilize the parent function's scope but here it's not necessary :)

packages/shared/src/lib/constants.ts
```diff
@@ -1,3 +1,6 @@
+// We have to declare the navigator because brave is not set in the types of the navigator, yet it does exist
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+declare const navigator: any;
```
idoshamun, line 3: How about this?

```suggestion
declare const navigator: Navigator & {brave?: {isBrave: unknown}};
```

packages/shared/src/lib/constants.ts
```diff
@@ -15,3 +18,16 @@ export const isDevelopment = process.env.NODE_ENV === 'development';
...
+  if (!window.Promise) {
+    return false;
+  }
+  if (
+    typeof navigator.brave === 'undefined' ||
+    typeof navigator.brave.isBrave !== 'function'
+  ) {
+    return false;
+  }
+  return true;
```
idoshamun, line 32: I think utilizing the optional chaining and returning the result of the statement is the way to go :)

```suggestion
  return typeof navigator.brave?.isBrave === 'function';
```

idoshamun (changes requested):
I have some ideas for refactoring. Other than that looks great. Quick response time :)

rebelchris:
@idoshamun Thanks! Refactored with your comments.
Missed a couple easy wins there 👀

packages/extension/src/newtab/dnd.ts
```diff
@@ -18,12 +18,11 @@ const DEFAULT_URL = 'https://www.google.com';
...
-  const browserTest = () =>
-    process.env.TARGET_BROWSER === 'chrome' ? CHROME_DEFAULT_URL : DEFAULT_URL;
-  const braveTest = () => (isBrave() ? BRAVE_DEFAULT_URL : browserTest());
-  return braveTest();
-};
+const browserTest = () =>
+  process.env.TARGET_BROWSER === 'chrome' ? CHROME_DEFAULT_URL : DEFAULT_URL;
+const braveTest = () => (isBrave() ? BRAVE_DEFAULT_URL : browserTest());
+
+export const getDefaultLink = (): string => braveTest();
```
idoshamun, line 25: Another easy win 👀 

```suggestion
const getDefaultLink = () => (isBrave() ? BRAVE_DEFAULT_URL : browserTest());
```

idoshamun (approved):
I approved it but just make sure to apply this one tiny change

PR #727, opened 2021-12-06
feat: added the sidebar bottom menu and refactored the sidebar

rebelchris:
Added the bottom menu options.
Also made the sidebar and header fixed.

The bottom bar introduces the invite people option, which for now will copy the link.

This PR has a major refactor on the sidebar, as I decided to extract certain components from the Sidebar.

DD-334 #done


packages/shared/src/components/sidebar/common.tsx
```diff
@@ -0,0 +1,111 @@
...
+  </>
+);
+export const btnClass = 'flex flex-1 items-center px-3 h-7';
+export const ButtonOrLink = ({
+  item,
+  children,
+}: {
+  item: SidebarMenuItem;
+  children?: ReactNode;
+}): ReactElement =>
```
sshanzel, line 52: I suggest we always create an interface/type when it comes to props for `ReactElement` functions with 2 or more properties. It is quite a bit confusing to read sometimes also when updating the properties themselves or adding new ones.

Could be just my personal POV, but what do you think about that?

packages/shared/src/components/sidebar/common.tsx
```diff
@@ -0,0 +1,111 @@
...
+      {children}
+    </button>
+  );
+export const MenuIcon = ({
+  openSidebar,
+  toggleOpenSidebar,
+}: {
+  openSidebar: boolean;
+  toggleOpenSidebar: () => Promise<void>;
+}): ReactElement => (
```
sshanzel, line 70: Tagging for the same concern I had above.

packages/shared/src/components/sidebar/common.tsx
```diff
@@ -0,0 +1,111 @@
...
+  'typo-footnote text-theme-label-quaternary h-7 flex items-center font-bold  transition-opacity',
+);
+export const RawNavItem = classed('li', 'flex items-center typo-callout');
+export const NavItem = ({
+  color,
+  children,
+}: {
+  color?: string;
+  children?: ReactNode;
+}): ReactElement => (
```
sshanzel, line 107: Tagging for the same concern I had above.

sshanzel (commented):
It is always so nice to see components being refactored and becoming slimmer. Great work!

Just had some minor non-blocking comments.

packages/shared/src/components/sidebar/common.tsx
rebelchris: That makes sense to me! let me work on that

rebelchris:
@sshanzel Thanks for that, yep makes total sense and refactored it 👀

packages/shared/src/components/sidebar/InvitePeople.tsx
```diff
@@ -0,0 +1,24 @@
+import React, { ReactElement } from 'react';
+import { useCopyLink } from '../../hooks/useCopyLink';
+import { ButtonOrLink, ItemInner, ListIcon, NavItem } from './common';
+import UserShareIcon from '../../../icons/user_share.svg';
+
+export default function InvitePeople({
```
idoshamun, line 6: When possible this should open native share

packages/shared/src/components/sidebar/InvitePeople.tsx
```diff
@@ -0,0 +1,24 @@
...
+  openSidebar,
+}: {
+  openSidebar: boolean;
+}): ReactElement {
+  const [copyingLink, copyLink] = useCopyLink(() => 'https://daily.dev/');
```
idoshamun, line 11: Please validate with Nimrod what should be the right link to share.
This can be either our landing page as you did now or `https://api.daily.dev/get` which is a direct link to the relevant store.

packages/shared/src/components/utilities.tsx
```diff
@@ -3,6 +3,8 @@ import classed from '../lib/classed';
 import styles from './utilities.module.css';
 import ArrowIcon from '../../icons/arrow.svg';
 
+export type TargetTypes = '_blank' | '_parent' | '_self' | '_top';
```
idoshamun, line 6: I think this type already exists out of the box, isn't it?

idoshamun (changes requested):
Some more work to do 💪

packages/shared/src/components/utilities.tsx
rebelchris: Oh yep! 👏

sshanzel (approved):
As neat as it can get! Thank you! 🚀

rebelchris:
@idoshamun And @sshanzel thanks for the feedback!
Modified the share to be the native one if we have it.

And shared the links to be referral links if we know the user, else show the default link.

idoshamun (approved):
LGTM! 🎉

PR #729, opened 2021-12-08
feat: added the hover and active styles for the new sidebar

rebelchris:
Added the hover styles and dynamic active style.
This is needed because there are cases where two items can be active (a page and popup action for instance).

DD-358 #done


packages/shared/src/components/sidebar/Sidebar.tsx
```diff
@@ -41,8 +42,7 @@ const bottomMenuItems: SidebarMenuItem[] = [
   {
     icon: <ListIcon Icon={TerminalIcon} />,
     title: 'Changelog',
-    path: 'https://changelog.daily.dev/',
-    target: '_blank',
+    path: '/sources/daily_updates',
```
idoshamun, line 45: A friendly reminder that this could be in the extension and not just the webapp.
Make sure to support this case. We have some samples throughout the codebase

packages/shared/src/components/sidebar/Sidebar.tsx
```diff
@@ -131,7 +133,10 @@ export default function Sidebar(): ReactElement {
                 {key}
               </NavHeader>
               {items.map((item) => (
-                <NavItem key={item.title}>
+                <NavItem
+                  key={item.title}
+                  active={item.active || item.path === router?.asPath}
```
idoshamun, line 138: Again we should support the extension as well here. You might want to think about getting this as a property from the outside. I mean the `router.asPath` part

idoshamun (changes requested):
Please see my comments :)

rebelchris:
@idoshamun Thanks for that, missed the complete navigation flow for the extension.
Rewrote quite a bit to make it work for both now :) (including the active states)

packages/extension/src/newtab/MainFeedPage.tsx
```diff
@@ -32,34 +28,40 @@ export type MainFeedPageProps = {
 export default function MainFeedPage({
   onPageChanged,
 }: MainFeedPageProps): ReactElement {
-  const { windowLoaded } = useContext(ProgressiveEnhancementContext);
   const { user } = useContext(AuthContext);
   const { onboardingStep } = useContext(OnboardingContext) || {};
   const [feedName, setFeedName] = useState<string>('default');
+  const [activePage, setActivePage] = useState<string>(
+    feedName === 'default' ? '/popular' : feedName,
```
idoshamun, line 35: Shouldn't this be: 

```suggestion
    feedName === 'default' ? '/' : feedName,
```

idoshamun: And anyway, I think it's best to compute this value rather than have it as a state.
We can compute it using the different states we already have. `useMemo` for the win

packages/shared/src/components/MainLayout.tsx
```diff
@@ -38,13 +42,19 @@ export default function MainLayout({
...
+  onNavTabClick,
+  enableSearch,
   onShowDndClick,
 }: MainLayoutProps): ReactElement {
+  const router = useRouter();
   const { windowLoaded } = useContext(ProgressiveEnhancementContext);
   const { user, showLogin, loadingUser } = useContext(AuthContext);
   const [showGreeting, setShowGreeting] = useState(false);
 
+  const useActivePage = activePage || router?.asPath;
```
idoshamun, line 56: I suggest receiving `router?.asPath` from the outside.
Let's keep this component platform-agnostic if possible.
You can utilize the existing `activePage` property to pass `router?.asPath` from the nextjs app

idoshamun (changes requested):

packages/shared/src/components/MainLayout.tsx
rebelchris: Sure thing! Fair enough ☺️

packages/extension/src/newtab/MainFeedPage.tsx
rebelchris: The path for the popular one is `/popular`, but the default also links to that.
So we need to make sure to set the activePage to `popular` if it's the default one.

I did change it to useMemo though :)

rebelchris:
@idoshamun Refactored from where we load the activePage now, and transformed the extension case to useMemo.
Would love to hear what you think :)

packages/shared/src/components/sidebar/Sidebar.tsx
```diff
@@ -80,25 +80,25 @@ export default function Sidebar({
           icon: <ListIcon Icon={HotIcon} />,
           title: 'Popular',
           path: '/popular',
-          action: () => onNavTabClick('popular'),
+          action: onNavTabClick ? () => onNavTabClick('popular') : null,
```
idoshamun, line 83: This might look better. What do you think?

```suggestion
          action: () => onNavTabClick.?('popular'),
```

idoshamun (changes requested):
One tiny improvement and we can merge!

idoshamun (approved):
Wooohoo!

PR #731, opened 2021-12-09
feat: move rank to sidebar

rebelchris:
This PR focusses on moving the rank from the header to the sidebar.
This also includes some new features:

- Small version of the rank (when the sidebar is closed)
- New text levels for a rank, this is dependant on your previous rank
- A new +1 animation when you not level up, but do get a +1 reading day

Small version:
Screenshot 2021-12-09 at 16 18 29

+1 animation:
https://user-images.githubusercontent.com/554874/145395165-678a9805-46ef-4495-9093-d1e6241464a3.mp4

Big version:
Screenshot 2021-12-09 at 16 18 36

Some notes:
- Animations still need to be checked with Tsahi, this is going to be another PR
- The new rank text: https://dailydotdev.atlassian.net/wiki/spaces/HAN/pages/113901569/Weekly+goal+texts


packages/shared/src/components/RankProgress.tsx
```diff
@@ -23,25 +26,37 @@ const notificationDuration = 300;
...
   className?: string;
   onRankAnimationFinish?: () => unknown;
   fillByDefault?: boolean;
+  smallVersion?: boolean;
+  showRadialProgress?: boolean;
+  showTextProgress?: boolean;
+  rankLastWeek?: number;
 };
 
+const getRank = (rank: number): string =>
```
rebelchris, line 41: Maybe rename this to getRankText

packages/shared/src/components/RankProgress.tsx
```diff
@@ -131,14 +151,23 @@ export function RankProgress({
...
+        } else {
+          progressAnimation?.cancel();
           firstBadgeAnimation.cancel();
           lastBadgeAnimation.cancel();
-          attentionAnimation.cancel();
+          attentionAnimation?.cancel();
           setForceColor(false);
           onRankAnimationFinish?.();
-        };
+        }
```
rebelchris, line 170: This could maybe be refactored into one liner?

packages/shared/src/components/RankProgress.tsx
```diff
@@ -163,65 +192,112 @@ export function RankProgress({
...
       (!rank || showRankAnimation || STEPS_PER_RANK[rank - 1] !== progress)
     ) {
-      setProgressDelta(progress - prevProgress);
+      if (!showRadialProgress) animateRank();
       setAnimatingProgress(true);
     }
     setPrevProgress(progress);
   }, [progress]);
 
+  const getNextRank = (useRank: number): string => {
```
rebelchris, line 201: Maybe rename this to getNextRankText

packages/shared/src/hooks/useReadingRank.ts
```diff
@@ -81,13 +87,14 @@ export default function useReadingRank(): ReturnType {
   const cacheRank = (
     rank: MyRankData = remoteRank,
     newNeverShowRankModal = neverShowRankModal,
-  ) =>
-    setCachedRank({
+  ) => {
+    return setCachedRank({
```
rebelchris, line 91: This should be the normal return :)

rebelchris (commented):
Some small notes

idoshamun (approved):
Brilliant!
Just make sure everything works as it should. This might be a bit tricky to test out but we have to make sure we don't introduce regressions

idoshamun:
Another comment that I see now is that there's a content overflow in the sidebar.
Because it takes time for the rank to load. I suggest reserving the place for it

rebelchris:
@idoshamun Great point!
I've changed the loading placeholder to be the empty heighten list placeholder so it works with semantics.

As for the testing.
I've gone through all possible states locally, by changing my database views to the possibilities 👀
I hope that covers it all

PR #735, opened 2021-12-12
fix: markdown link color was missing

rebelchris:
With the latest removal of the comment css we missed the actual markdown link color.

Reintroduced it as to color provided in spec.

Screenshot 2021-12-12 at 11 20 51


idoshamun (approved):
👀

PR #738, opened 2021-12-14
feat: mobile menu sidebar

rebelchris:
Hey guys,

Bear with me as this is a PR on the bigger size.

I have one question in regards to an issue I had:
- HeaderRankProgress and SidebarRankProgress as super similar, but they render completely different, is that fine with you?

Profile context menu:
Because we now have the sidebar we removed two items here:
- Reading History
- Feedback

Icons:
I had a discussion with Tsahi, and for now we will only slowly introduce the new icons.
Tsahi needs to re-check certain elements so we can't just swap them.
For now I added a new `filled` and `outline` folder in the icons setup for these new ones.

General:
We are moving around quite a lot of dynamic components, take the profile picture for instance.
This has a lot of functions and normally sits on the top right.
On the sidebar, but only mobile it shows inside the sidebar, and has different functionality, so had to "re-use" the component inner elements.


packages/shared/icons/filled/hamburger.svg
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
rebelchris, line 1: Note: We should cleanup this icon file (Ask Tsahi for the correct one)

packages/shared/src/components/FeedLayout.tsx
```diff
@@ -150,10 +151,15 @@ export default function FeedLayout({
...
+  const hiddenSidebar = useMedia(
+    [footerNavBarBreakpoint.replace('@media ', '')],
+    [true],
+    false,
+  );
 
   const currentSettings = useMedia(
     reversedBreakpoints,
-    openSidebar ? reversedSettingsSidebar : reversedSettings,
+    openSidebar && hiddenSidebar ? reversedSettingsSidebar : reversedSettings,
```
rebelchris, line 162: Is this one clear enough what the different between reversedSettingsSidebar and reversedSettings is?

packages/shared/src/components/MainLayout.tsx
```diff
@@ -19,6 +19,9 @@ import styles from './MainLayout.module.css';
 import ProfileButton from './profile/ProfileButton';
 import { LinkWithTooltip } from './tooltips/LinkWithTooltip';
 import Sidebar from './sidebar/Sidebar';
+import MenuIcon from '../../icons/filled/hamburger.svg';
```
rebelchris, line 22: Do note: We don't have a valid system for the new system, and currently we have to support the old and new ones for the time being.

We might want to refactor icon usage after relayout!

packages/shared/src/components/MainLayout.tsx
```diff
@@ -52,11 +55,18 @@ export default function MainLayout({
   const { windowLoaded } = useContext(ProgressiveEnhancementContext);
   const { user, showLogin, loadingUser } = useContext(AuthContext);
   const [showGreeting, setShowGreeting] = useState(false);
+  const showSidebar = useMedia(
```
rebelchris, line 58: Maybe make this sidebarVisible? because that's actually what's happening
Or sidebarRendered

packages/shared/src/components/RankProgress.tsx
```diff
@@ -169,10 +173,11 @@ export function RankProgress({
         } else {
```
rebelchris, line 173: Still think we could rearrange this function to be extracted?

packages/shared/src/components/RankProgress.tsx
```diff
@@ -293,10 +298,18 @@ export function RankProgress({
             unmountOnExit
           >
             <div className="flex flex-col items-start ml-3">
-              <span className="font-bold text-theme-rank typo-callout">
+              <span
+                className={`font-bold text-theme-rank typo-callout transition-opacity ${
```
rebelchris, line 302: This is no longer needed

packages/shared/src/components/profile/ProfileButton.tsx
```diff
@@ -21,10 +23,15 @@ export default function ProfileButton({
...
+          onClick={onClick || onMenuClick}
         >
-          <span className="mr-2 ml-3">{user.reputation ?? 0}</span>
+          <span className="hidden laptop:block mr-2 ml-3">
+            {user.reputation ?? 0}
+          </span>
           <ProfilePicture user={user} size="medium" />
+          <span className="laptop:hidden mr-2 ml-3">
+            {user.reputation ?? 0}
+          </span>
```
rebelchris, line 34: Think we can also remove this now :)

packages/shared/src/components/sidebar/Sidebar.tsx
```diff
@@ -56,14 +68,32 @@ const bottomMenuItems: SidebarMenuItem[] = [
...
 }: SidebarProps): ReactElement {
   const { alerts } = useContext(AlertContext);
+  const { user, showLogin, loadingUser } = useContext(AuthContext);
   const { isLoaded, isAnimated, setLoaded, setHidden } =
     useDynamicLoadedAnimation();
   const { openSidebar, toggleOpenSidebar } = useContext(SettingsContext);
   const [showSettings, setShowSettings] = useState(false);
+  const router = useRouter();
+
+  useEffect(() => {
```
rebelchris, line 86: We introduced this route change to close the sidebar, because we have actions and links triggering change.

packages/shared/src/components/sidebar/Sidebar.tsx
```diff
@@ -126,55 +159,110 @@ export default function Sidebar({
...
-                {key}
-              </NavHeader>
-              {items.map((item) => (
-                <NavItem
-                  key={item.title}
-                  active={item.active || item.path === activePage}
+      {openMobileSidebar && <SidebarBackdrop onClick={setOpenMobileSidebar} />}
+      <SidebarAside
+        className={classNames(
+          'w-70 laptop:-translate-x-0',
```
rebelchris, line 170: I think this could be inside the Sidebar aside

packages/shared/src/components/sidebar/Sidebar.tsx
```diff
@@ -126,55 +159,110 @@ export default function Sidebar({
...
+          openMobileSidebar ? '-translate-x-0' : '-translate-x-70',
+        )}
+      >
+        {showSidebar && (
+          <MenuIcon
+            openSidebar={openSidebar}
+            toggleOpenSidebar={toggleOpenSidebar}
+          />
+        )}
+        <div className="flex overflow-x-hidden overflow-y-auto flex-col h-full no-scrollbar">
```
rebelchris, line 181: This could get it's own wrapper component

packages/shared/src/components/sidebar/Sidebar.tsx
```diff
@@ -126,55 +159,110 @@ export default function Sidebar({
...
+          <Nav>
+            {!loadingUser && !showSidebar && (
```
rebelchris, line 183: Probably can extract this into it's own component

packages/shared/tailwind.config.js
```diff
@@ -138,7 +138,12 @@ module.exports = {
         'screen-80': '80vh',
         screen: '100vh',
       },
+      spacing: {
+        70: '17.5rem',
+      },
       width: {
+        13: '3.25rem',
```
rebelchris, line 145: Remove this one

rebelchris (commented):
Some small changes :)

sshanzel (changes requested):
Just had a few very minor comments. Day by day we are getting closer! Great work! 🚢

packages/shared/src/components/FeedLayout.tsx
idoshamun: I suggest extracting this one-liner to a function with a concise name.
You can define it outside the scope of the component and provide it all the details

packages/shared/src/components/MainLayout.tsx
idoshamun: When refactoring icon usage we might need to think again about SVGR.
I heard there are better ways now to handle icons.
SVGR can be memory intensive.

packages/shared/src/components/MainLayout.tsx
```diff
@@ -52,11 +55,18 @@ export default function MainLayout({
...
+    [footerNavBarBreakpoint.replace('@media ', '')],
+    [true],
+    false,
+  );
```
idoshamun, line 62: This seems to be repeated.
Please consider creating a hook for this.

Another suggestion would be to use some sort of context to reduce the number of event listeners. But not sure if we should really care about it.

packages/shared/src/components/MobileHeaderRankProgress.tsx
```diff
@@ -0,0 +1,25 @@
+import React, { CSSProperties, ReactElement } from 'react';
+import useReadingRank from '../hooks/useReadingRank';
+import RankProgressWrapper from './RankProgressWrapper';
+import { rankToColor } from '../lib/rank';
+
+export default function MobileHeaderRankProgress(): ReactElement {
+  const { isLoading, rank, nextRank } = useReadingRank();
+
+  if (isLoading) {
+    return <div className="">&nbsp;</div>;
```
idoshamun, line 10: We can also return `<></>`

packages/shared/src/components/MobileHeaderRankProgress.tsx
```diff
@@ -0,0 +1,25 @@
...
+  }
+
+  return (
+    <div
+      className=""
```
idoshamun, line 15: Why the empty className?

packages/shared/src/components/MobileHeaderRankProgress.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+export default function MobileHeaderRankProgress(): ReactElement {
+  const { isLoading, rank, nextRank } = useReadingRank();
+
+  if (isLoading) {
+    return <div className="">&nbsp;</div>;
+  }
+
+  return (
+    <div
```
idoshamun, line 14: Why do we need this parent `div`?

packages/shared/src/components/RankProgress.tsx
```diff
@@ -140,39 +140,43 @@ export function RankProgress({
...
+          lastBadgeAnimation?.cancel();
           attentionAnimation?.cancel();
           setForceColor(false);
           onRankAnimationFinish?.();
+        };
+
+        if (attentionAnimation) {
+          attentionAnimation.onfinish = () => {
+            cancelAnimations();
+          };
```
idoshamun, line 176:
```suggestion
          attentionAnimation.onfinish = cancelAnimations;
```

packages/shared/src/components/SidebarRankProgress.tsx
```diff
@@ -1,130 +1,32 @@
...
-        />
-      )}
-      {showAccountDetails && (
-        <AccountDetailsModal
-          isOpen={showAccountDetails}
-          onRequestClose={() => setShowAccountDetails(false)}
-        />
-      )}
-    </>
+    <li
```
idoshamun, line 18: To make sure we reserve the same size when loading. I'd suggest to use the same parent element and provide it with a different child depending on the state

packages/shared/src/components/profile/ProfileButton.tsx
```diff
@@ -21,9 +23,11 @@ export default function ProfileButton({
         <button
           type="button"
           className="flex items-center p-0 ml-0.5 font-bold no-underline rounded-lg border-none cursor-pointer text-theme-label-primary bg-theme-bg-secondary typo-callout focus-outline"
-          onClick={onMenuClick}
+          onClick={onClick || onMenuClick}
```
idoshamun, line 26: Why do we need both callbacks?

packages/shared/src/components/sidebar/Sidebar.tsx
idoshamun: This calls for a custom hook 😉

packages/shared/src/components/sidebar/common.tsx
```diff
@@ -86,8 +104,8 @@ export const ItemInner = ({
       {item.icon}
     </span>
     <span
-      className={`flex-1 text-left transition-opacity ${
-        openSidebar ? 'opacity-100' : 'opacity-0'
+      className={`flex-1 text-left  transition-opacity ${
```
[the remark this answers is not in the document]
idoshamun: Agree with the suggestion

idoshamun (changes requested):
If the difference between filled and outline icons is just a property maybe we can override it with some class. Not sure it that's the only difference. We might need to ask Tsahi.

Other than that, comments inside. This was quite a PR but overall seems solid!

packages/shared/src/components/profile/ProfileButton.tsx
rebelchris: We can actually removed this, Tsahi modified the design so this would no longer be needed.

packages/shared/src/components/sidebar/SidebarUserButton.tsx
```diff
@@ -0,0 +1,63 @@
...
+export default function SidebarUserButton({
+  showSidebar,
+  onShowDndClick,
+}: SidebarUserButtonProps): ReactElement {
+  const { user, showLogin, loadingUser } = useContext(AuthContext);
+
+  return (
+    <>
+      {!loadingUser && !showSidebar && (
+        <li className="flex flex-col p-6 pt-2">
```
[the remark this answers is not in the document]
rebelchris: Yeah the nav section is already adding `mt-4` so we can't offset it.
And only this nav element needs the different padding.

```
<nav class="my-4">
```

packages/shared/src/components/MobileHeaderRankProgress.tsx
rebelchris: We actually need it to use the top header space-between.
Else the button itself will just span across all available space.

rebelchris:
@sshanzel @idoshamun Thanks for your amazing review!
Think we made it even better now!

Managed to solve all but one issue from your review.
Would love to get your thought on it.

sshanzel (approved):
LGTM! 🚢 

Just had a question but awesome work!

packages/shared/src/components/FeedLayout.tsx
```diff
@@ -146,14 +152,23 @@ const reversedBreakpoints = feedBreakpoints
 const reversedSettings = baseFeedSettings.reverse();
 const reversedSettingsSidebar = sidebarOpenFeedSettings.reverse();
 
+const determineFeedSettings = ({
+  openSidebar,
+  showSidebar,
+}: DetermineFeedSettingsProps): FeedContextData[] => {
+  return openSidebar && showSidebar
```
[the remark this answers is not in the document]
rebelchris: Good question.

openSidebar is when the sidebar is in open mode (expanded)
and showSidebar is to determine is it's in the view at all (mobile does not have the sidebar in view)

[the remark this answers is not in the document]
idoshamun: Maybe we can be more explicit in the names. I agree with Lee's comment

packages/shared/src/components/MainLayout.tsx
```diff
@@ -57,42 +55,44 @@ export default function MainLayout({
...
+                  <LogoText
+                    className={classNames(
+                      styles.homeSvg,
+                      'hidden ml-1 laptop:block',
+                    )}
+                  />
+                </CSSTransition>
+              </a>
+            </LinkWithTooltip>
+          ))}
```
idoshamun, line 95: This can be extracted to a standalone component I guess.
I know it has been there for long time but it's time to make it a bit cleaner

idoshamun (changes requested):
Wow! So much cleaner.
Let's push it a bit and clean it some more.

rebelchris:
@sshanzel @idoshamun 

Refactored some more stuff...
-  Removed onboarding context in general as we don't use it anymore
- Menu button and gear icon changed to iconOnly buttons
- Refactored `showSidebar` to `sidebarRendered`
- Refactored  `openSidebar` to `sidebarExpanded`
- Extracted our logo to it's own component

packages/shared/src/components/MainLayout.tsx
```diff
@@ -40,6 +35,17 @@ const Greeting = dynamic(
...
 
+interface ShouldShowLogoProps {
+  mobileTitle?: string;
+  sidebarRendered?: boolean;
+}
+const shouldShowLogo = ({
+  mobileTitle,
+  sidebarRendered,
+}: ShouldShowLogoProps) => {
+  return !mobileTitle ? true : mobileTitle && sidebarRendered;
```
idoshamun, line 46: I guess it should be like this?

```suggestion
  return !mobileTitle ? true : (mobileTitle && sidebarRendered);
```

packages/shared/src/components/sidebar/common.tsx
```diff
@@ -144,24 +144,26 @@ export const ButtonOrLink = ({
...
+  toggleSidebarExpanded,
 }: MenuIconProps): ReactElement => (
   <SimpleTooltip
     placement="right"
-    content={`${openSidebar ? 'Close' : 'Open'} sidebar`}
+    content={`${sidebarExpanded ? 'Close' : 'Open'} sidebar`}
   >
     <Button
-      onClick={() => toggleOpenSidebar()}
+      onClick={() => toggleSidebarExpanded()}
```
idoshamun, line 155:
```suggestion
      onClick={toggleSidebarExpanded}
```

idoshamun (changes requested):
I love deleting code! 🤩

packages/shared/src/components/MainLayout.tsx
rebelchris: Prettier will remove those wrapping ()
So would say this is valid as it's inside a if else

idoshamun (approved):
Perfect!

PR #739, opened 2021-12-14
feat: refactor bookmark search page

rebelchris:
In the new design we no longer have a separate search page for the bookmarks.
This refactor makes sure the search bar is always active for the bookmark page.

Screenshot 2021-12-14 at 17 15 27
Screenshot 2021-12-14 at 17 15 35

For the mobile view we added a new option to the main layout including `mobileTitle` this title will get rendered in the header for certain pages.
This does need the mobile sidebar to finalise how it renders.


sshanzel (commented):
Just had a few very minor comments but very well done on working this relayout 💪

packages/shared/src/components/BookmarkFeedLayout.tsx
```diff
@@ -55,21 +50,11 @@ export default function BookmarkFeedLayout({
   return (
     <FeedPage>
       {children}
+      <FeedPageHeader className="mb-5">
+        <h3 className="font-bold typo-callout">Bookmarks</h3>
+      </FeedPageHeader>
```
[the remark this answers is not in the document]
rebelchris: So far I haven't seen it used on any other pages like this.
The main layout uses another version of this.

packages/shared/src/components/utilities.tsx
```diff
@@ -56,7 +56,12 @@ export const ActiveTabIndicator = classed(
 
 export const CustomFeedHeader = classed(
   'div',
-  'flex h-11 self-stretch items-center mb-6 text-theme-label-secondary typo-callout',
+  'flex h-11 self-stretch items-center mb-6 typo-callout',
+);
+
+export const FeedPageHeader = classed(
+  'nav',
```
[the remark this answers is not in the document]
rebelchris: Very good point!

packages/shared/src/components/BookmarkFeedLayout.tsx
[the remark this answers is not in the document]
sshanzel: Then I think this should be good to go 🚀

sshanzel (approved):
Woop! Woop! 🎉

packages/shared/src/components/utilities.tsx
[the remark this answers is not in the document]
idoshamun: Nice catch!

packages/shared/src/components/MainLayout.tsx
```diff
@@ -89,6 +91,12 @@ export default function MainLayout({
           />
         )}
         <div className="flex-1" />
+        {mobileTitle && (
+          <strong className="block laptop:hidden typo-callout">
```
idoshamun, line 95: Shouldn't this be `hx` element?

idoshamun (changes requested):
Looks great! Thanks for the info in the PR really helped me review it.
One UX question, what should happen when I hit the bookmarks button when I'm not logged in. Currently it shows a sign up popup. Is it the same now?

packages/shared/src/components/MainLayout.tsx
rebelchris: I'm not sure what that means?

idoshamun: h1, h2, h3, h something

sshanzel:
Was about to say the same for the reading history. Probably we should make it as a popup as well if we're gonna do that on the bookmarks to be consistent.

packages/shared/src/components/MainLayout.tsx
rebelchris: We generally don't set the callout to a heading from what I can see.

Happy to change to a H3 as this is what it is on desktop, WDYT?
Do keep in mind this section will also render the good morning message (which is a span)
And the rank, (also a span)

rebelchris:
@idoshamun / @sshanzel good call on that showLogin.
Actually missed that in the refactor.

Added it in this branch for reading history and bookmarks.
This takes priority over any other action that might have been set for the button.

packages/shared/src/components/MainLayout.tsx
idoshamun: H3 sounds good

idoshamun (approved):
Actually you added the `showLogin` in the mobile branch which is perfectly fine by me

PR #740, opened 2021-12-15
refactor: signup button copy experiment

idoshamun:
We want to try out different copies for the login button in the header.
For this purpose I added a new feature `signup_button_copy` in Flagsmith. This feature will hold the string for the button.
Please note, in mobile resolution we will keep the current copy due to space restrictions.

I also migrated the relevant signup events to our in-house analytics tracking so we can monitor the performance of the experiment.

DD-362 #done

### DIFF
--- a/packages/shared/src/components/LoginButton.tsx
+++ b/packages/shared/src/components/LoginButton.tsx
@@ -1,0 +1,40 @@
+import React, { ReactElement, useContext, useEffect } from 'react';
+import { Button } from './buttons/Button';
+import AuthContext from '../contexts/AuthContext';
+import FeaturesContext from '../contexts/FeaturesContext';
+import { getFeatureValue } from '../lib/featureManagement';
+import AnalyticsContext from '../contexts/AnalyticsContext';
+import { AnalyticsEvent } from '../hooks/analytics/useAnalyticsQueue';
+
+const getAnalyticsEvent = (
+  eventName: string,
+  copy: string,
+): Partial<AnalyticsEvent> => ({
+  event_name: eventName,
+  target_type: 'signup button',
+  target_id: 'header',
+  feed_item_title: copy,
+});
+
+export default function LoginButton(): ReactElement {
+  const { showLogin } = useContext(AuthContext);
+  const { flags } = useContext(FeaturesContext);
+  const { trackEvent } = useContext(AnalyticsContext);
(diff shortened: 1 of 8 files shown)


packages/shared/src/components/LoginButton.tsx
```diff
@@ -0,0 +1,40 @@
...
+  target_type: 'signup button',
+  target_id: 'header',
+  feed_item_title: copy,
+});
+
+export default function LoginButton(): ReactElement {
+  const { showLogin } = useContext(AuthContext);
+  const { flags } = useContext(FeaturesContext);
+  const { trackEvent } = useContext(AnalyticsContext);
+  const copy = getFeatureValue('signup_button_copy', flags) || 'Login';
```
rebelchris, line 23: Only suggestion is to change this to buttonCopy to not give the impression it copies anything.

rebelchris (changes requested):
Wow nice work!

1 Small naming remark.

packages/shared/src/components/LoginButton.tsx
idoshamun: Sure! I fixing now

packages/shared/src/components/LoginButton.tsx
```diff
@@ -0,0 +1,40 @@
...
+  target_type: 'signup button',
+  target_id: 'header',
+  feed_item_title: copy,
+});
+
+export default function LoginButton(): ReactElement {
+  const { showLogin } = useContext(AuthContext);
+  const { flags } = useContext(FeaturesContext);
+  const { trackEvent } = useContext(AnalyticsContext);
+  const buttonCopy = getFeatureValue('signup_button_copy', flags) || 'Login';
```
sshanzel, line 23: Wow, didn't we know we have this function. I will for sure use it next time.

I was thinking now to create an `enum` (or `type` which we can use as a parameter type for the function itself) or an `object` that will have these strings so it can be much easier to use on some other parts and ensure to have the same string that is being called. What do you think?

sshanzel (commented):
Just a single non-blocking comment but the new component is as neat as it can get! 🚢

rebelchris (approved):
Awesome! 👏

packages/shared/src/components/LoginButton.tsx
idoshamun: This is a great suggestion. I had auto merge turned on so I'll push another branch

PR #742, opened 2021-12-15
refactor: use features enum to store available flags

idoshamun:
Following Lee's suggestion, I introduce a new enum that will make it easy to use feature flags.


sshanzel (approved):
Wow, thank you so much for applying it across existing implementations! 🚢

PR #744, opened 2021-12-17
feat: article detail page

rebelchris:
This PR reworks the post detail page.
This is related to the introduction of the sidebar, where the post page would actually be too wide.

There are now three different kind of views:
- LaptopL -> Full view as before
- Laptop -> Middle view with everything stacked, but comments above sidebar
- Small -> Stacked, but comments are now a fixed element at the bottom.

See attached loom for more details.
https://www.loom.com/share/4cf1453544ee4f838692e87e9bfe2fec

DD-348 #done


idoshamun (changes requested):
`PostSidebar` is a somewhat confusing name given the real sidebar we have.
For me it sounds like a variation of the sidebar for the the article page.

Can you please introduce another refactor in this PR for article page? Take something from the main article page component and extract it into its own component or hook. I can't believe I wrote such a huge component myself 🤯

sshanzel:
I have finished refactoring the article page, there's quite a lot of extraction of components so I highly suggest reviewing commit by commit as I have separated the extraction in each one.

I have also renamed the `PostSidebar` to `PostWidgets` as if felt fitting since we have multiple widgets on that side.

sshanzel:
@rebelchris can't seem to ask review from you since you initially created the PR - feel free to do a review as well. Thank you!

packages/webapp/components/posts/PostWidgets.tsx
```diff
@@ -0,0 +1,62 @@
...
+    ssr: false,
+  },
+);
+
+const FurtherReading = dynamic(
+  () =>
+    import(
+      /* webpackChunkName: "furtherReading" */ '../widgets/FurtherReading'
+    ),
+);
```
idoshamun, line 22: This doesn't require dynamic import anymore

idoshamun (changes requested):
Wow! So clean now

rebelchris (commented):
Nice work Lee!
Thanks for extracting almost everything here 👏

packages/webapp/components/posts/PostWidgets.tsx
sshanzel: Good catch! Since this component won't be reached if there's no `post` (as we return a Fragment early on), I have removed the other dynamic imports here as well.

packages/webapp/components/posts/PostWidgets.tsx
```diff
@@ -49,13 +36,10 @@ export function PostWidgets({ postById }: PostWidgetsProps): ReactElement {
 
   return (
     <PageSidebar>
-      {postById && <ShareBar post={postById.post} />}
+      <ShareBar post={post} />
```
idoshamun, line 39: The check for `postById` was there on purpose because it's being fetched on load.
If this is not defined this might cause an error.
```suggestion
      {post && <ShareBar post={post} />}
```

packages/webapp/components/posts/PostWidgets.tsx
```diff
@@ -49,13 +36,10 @@ export function PostWidgets({ postById }: PostWidgetsProps): ReactElement {
...
       <ShareMobile share={sharePost} />
-      {postById?.post && tokenRefreshed && (
-        <FurtherReading
-          currentPost={postById.post}
-          className="laptopL:w-[19.5rem]"
-        />
+      {tokenRefreshed && (
```
idoshamun, line 41: The check for `postById` was there on purpose because it's being fetched on load.
If this is not defined this might cause an error.

```suggestion
      {post && tokenRefreshed && (
```

packages/webapp/pages/posts/[id].tsx
```diff
@@ -295,7 +295,7 @@ const PostPage = ({ id, postData }: Props): ReactElement => {
           )}
         </PageContainer>
         <NewComment user={user} onNewComment={openNewComment} />
-        <PostWidgets postById={postById} />
+        <PostWidgets post={postById.post} />
```
Comment (idoshamun) [inline packages/webapp/pages/posts/[id].tsx:298]:
```suggestion
        <PostWidgets post={postById?.post} />
```

idoshamun (changes requested):

packages/webapp/components/posts/PostWidgets.tsx
sshanzel: From the line of code below, we return whenever the `postById` or `postById.post` is `null` or `undefined`, so the code below this point (including the rendering part of this component) should be safe. Would we use this component for other parts as well?

https://github.com/dailydotdev/apps/blob/c7995b1e4b7071464e91aa6cd584466b184758ce/packages/webapp/pages/posts/%5Bid%5D.tsx#L201-L203

sshanzel: From the line of code below, we return whenever the `postById` or `postById.post` is `null` or `undefined`, so the code below this point (including the rendering part of this component) should be safe. Would we use this component for other parts as well?

https://github.com/dailydotdev/apps/blob/c7995b1e4b7071464e91aa6cd584466b184758ce/packages/webapp/pages/posts/%5Bid%5D.tsx#L201-L203

idoshamun: You're right! I missed that

idoshamun (approved):
Looks great!

sshanzel:
Thank you both for the awesome review! 🚢

PR #745, opened 2021-12-17
feat: added tracking events for the re-layout

rebelchris:
After discussing the events with Ido we decided on the following new events:

- Invite people clicked
- Toggle sidebar (mobile and desktop same tracking)
- Open customize modal
- Open Rank modal (generalised this one to trigger from the modal itself)

DD-356 #done


idoshamun (changes requested):
You used the wrong `trackEvent` function. I knew that's confusing!
You should use the other one with the custom hook.
This `trackEvent` uses Google Analytics and we want our custom solution.
We might want to change this name to prevent future confusion.

rebelchris:
@idoshamun Thanks for that mention.
I saw we replaced the Google `trackEvent` as `gaTrackEvent` sometimes.
So renamed it generally to that to make it more explicit.

Also added our in house tracking for all these new events.
Could you let me know if I did the `event_name`, `target_type` and `origin` in the right way?

idoshamun (changes requested):
A few notes:
* We use lowercase letters for the event name and all the other properties.
* Target type is not actually needed here, we can be more explicit in the `event_name` e.g `open sidebar`.
* The extra field is also not necessary as there's only one way to open the sidebar

For next time, I think it's better if I can document all these fields as part of the spec so you won't have to guess and we can reduce friction.

packages/shared/src/components/modals/RanksModal.tsx
```diff
@@ -197,6 +199,14 @@ export default function RanksModal({
   className,
   ...props
 }: RanksModalProps): ReactElement {
+  const { trackEvent } = useContext(AnalyticsContext);
+
+  useEffect(() => {
+    trackEvent({
+      event_name: 'show ranks modal',
+    });
+  }, []);
```
idoshamun, line 208: We need to track it when it opens. `useEffect` might be tricking us.
You can check in the login modal how I implemented it.

packages/shared/src/components/sidebar/Sidebar.tsx
```diff
@@ -137,7 +150,12 @@ export default function Sidebar({
         {
           icon: <ListIcon Icon={SettingsIcon} />,
           title: 'Customize',
-          action: () => setShowSettings(!showSettings),
+          action: () => {
+            trackEvent({
```
idoshamun, line 154: Maybe it's better to just unify the way we track modals. And add the relevant event in the settings component rather than here. What do you think?
We can also introduce a custom hook for tracking the state of a modal

idoshamun (changes requested):

packages/shared/src/components/sidebar/Sidebar.tsx
rebelchris: Agreed, might be better to do it inside the modals as this will be better for wherever they might get opened from.

rebelchris:
@idoshamun Thanks for the feedback and advice!

Introducing the new `useTrackModal` hook.
The hook will do all the tracking, and for conditionally rendered elements like:

```js
{showRanksModal && (
        <RanksModal />
)}
```

it will do a auto cleanup!
Meaning if the modal gets closed, or page destroyed, the cleanup function will fire a close event!

Because of this I had to remove the windowLoaded from the LoginModal.
This was causing the modal to always initiate and sending FAKE show events! (Already in the current setup, so you might see a decrease from now on)

packages/shared/src/components/modals/RanksModal.tsx
```diff
@@ -197,13 +198,19 @@ export default function RanksModal({
   className,
   ...props
 }: RanksModalProps): ReactElement {
+  useTrackModal({ isOpen: props.isOpen, title: 'ranks modal' });
+
+  const onClose = (event: React.MouseEvent | React.KeyboardEvent): void => {
+    onRequestClose(event);
+  };
```
idoshamun, line 205: Feels like you can use directly `onRequestClose`

packages/shared/src/hooks/useTrackModal.ts
```diff
@@ -0,0 +1,30 @@
...
+    };
+    trackEvent(trackObject);
+
+    return () => {
+      if (isOpen === true) {
+        // The modal was conditionally rendered we should fire a close event
+        trackObject.event_name = `close ${title}`;
+        trackEvent(trackObject);
+      }
+    };
```
idoshamun, line 28: I think it's more consistent in terms of javascript closure

```suggestion
      if (isOpen === true) {
    return () => {
        // The modal was conditionally rendered we should fire a close event
        trackObject.event_name = `close ${title}`;
        trackEvent(trackObject);
      };
    }
```

idoshamun (changes requested):
Looks great a few suggestions

rebelchris:
@idoshamun Because of the switch in the return value I have to omit the `useEffect` by using:
`// eslint-disable-next-line consistent-return`

Not sure if that's cleaner to be honest.

idoshamun (approved):
I didn't suggest the change to make it cleaner but to make it more predictable in terms of closure of anonymous functions. If you want to get rid of the eslint error you can return empty function otherwise. If you still this is just a mess you can revert back. I'm fine with it. Your choice :)

PR #746, opened 2021-12-17
feat: test issues

rebelchris:
We had some early test issues and all round things that still needed to be done.

- Main layout rendering issue solved (jumping logo and wrong classnames applied)
- Mobile view jumping header rank and logo
- Sidebar icon loading issue (We had to wait for window load)
- Bottom mobile menu to filter
- New layout for mobile where we can have the header as well as the footer nav bar

DD-365 #done


packages/shared/src/components/MainLayout.tsx
```diff
@@ -72,15 +78,17 @@ export default function MainLayout({
...
-        )}
-        {windowLoaded && greeting && (
-          <Greeting
-            user={user}
-            onEnter={() => setShowGreeting(true)}
-            onExit={() => setShowGreeting(false)}
-          />
+        {shouldShowLogo({ mobileTitle, sidebarRendered }) && windowLoaded && (
+          <div className="flex flex-row">
+            <Logo onLogoClick={onLogoClick} showGreeting={showGreeting} />
```
idoshamun, line 83: We can show the logo before the load, can't we?

packages/shared/src/components/MobileHeaderRankProgress.tsx
```diff
@@ -1,25 +1,45 @@
...
+const Wrapper = ({ children, ...props }: WrapperProps) => {
+  return (
+    <div className="w-8 h-8" {...props}>
+      {children}
+    </div>
+  );
+};
 
 export default function MobileHeaderRankProgress(): ReactElement {
+  const { windowLoaded } = useContext(ProgressiveEnhancementContext);
```
idoshamun, line 25: Why do we wait for `windowLoaded` here?
Once the rank is available we can show it

packages/shared/src/components/sidebar/Sidebar.tsx
```diff
@@ -168,39 +171,40 @@ export default function Sidebar({
...
-                    sidebarExpanded ? 'opacity-100 px-3' : 'opacity-0 px-0',
-                  )}
-                >
-                  {key}
-                </NavHeader>
-                {items.filter(mobileItemsFilter).map((item) => (
-                  <NavItem
-                    key={item.title}
-                    active={item.active || item.path === activePage}
+            {windowLoaded &&
```
idoshamun, line 174: Why do we have to wait for `windowLoaded`?

idoshamun (changes requested):
Please check my comments.
My main concern is that we would like to show as fast as possible the general layout of the page while avoiding content overflow. I think this is doable and feel free to advise if you need :)

packages/shared/src/components/MobileHeaderRankProgress.tsx
rebelchris: This causes a react leak same as for the sidebar.
Not sure if it can be solved otherwise but would love to hear alternatives.

idoshamun: What do you mean by "react leak"?

rebelchris:
@idoshamun Heavy refactor
Tried a lot of things, but this is the end.

Moving some things around and migrating to own components seems to be the clue.
Also upgraded all packages for extension and webapp.

packages/eslint-config/package-lock.json
```diff
@@ -1,3996 +1,29 @@
 {
 	"name": "@dailydotdev/eslint-config",
 	"version": "0.0.0",
-	"lockfileVersion": 2,
+	"lockfileVersion": 1,
```
idoshamun: Seems like something changed in the `eslint-config` package-lock

packages/shared/src/components/MainLayout.tsx
```diff
@@ -58,30 +57,30 @@ export default function MainLayout({
...
-      <header className="flex relative laptop:fixed laptop:top-0 laptop:left-0 z-3 flex-row-reverse laptop:flex-row justify-between items-center py-3 px-4 tablet:px-8 laptop:px-4 laptop:w-full h-14 border-b bg-theme-bg-primary border-theme-divider-tertiary">
-        {!sidebarRendered && <MobileHeaderRankProgress />}
-        {mobileTitle && (
-          <h3 className="block laptop:hidden typo-callout">{mobileTitle}</h3>
-        )}
-        {shouldShowLogo({ mobileTitle, sidebarRendered }) && (
-          <Logo onLogoClick={onLogoClick} showGreeting={showGreeting} />
-        )}
-        {windowLoaded && greeting && (
+  const LogoAndGreeting = () => {
```
idoshamun, line 65: Let's take it outside this component. (you can keep it in the same file)
If you need you can put its own state inside as well

idoshamun (changes requested):
Wow! I'm happy you bumped into this bug in a way.
This looks so much better

packages/shared/src/components/sidebar/Sidebar.tsx
```diff
@@ -78,75 +77,110 @@ export default function Sidebar({
...
+                item={item}
+                sidebarExpanded={sidebarExpanded || !sidebarRendered}
+              />
+            </ButtonOrLink>
+          </NavItem>
+        ))}
+      </NavSection>
+    );
+  };
+
```
sshanzel, line 183: I think we can safely say that we can create another file for this component.

packages/shared/src/components/sidebar/Sidebar.tsx
```diff
@@ -78,75 +77,110 @@ export default function Sidebar({
...
+      icon: <ListIcon Icon={EyeIcon} />,
+      title: 'Reading history',
+      path: `${process.env.NEXT_PUBLIC_WEBAPP_URL}history`,
+      requiresLogin: true,
+    },
+    {
+      icon: <ListIcon Icon={SettingsIcon} />,
+      title: 'Customize',
+      action: () => setShowSettings(!showSettings),
+      active: showSettings,
```
sshanzel, line 135: Now that we have separated our list for these items, I think we can take these two out and give them their own separate files then accept props to make it easier to manage.

I think in the future it will be easy to create a new set of sections and this one file would be filled so much by those lists. What do you think?

sshanzel (changes requested):
Wow. Thank you so much for this work. You even upgraded the packages which I can't imagine.

Just had two things mentioned but nothing so major. Well done on this! 🚀

packages/shared/src/components/sidebar/Sidebar.tsx
rebelchris: They will be receiving a lot of props though.
Not sure if we want to pass down so much information vs keeping it there?

The list will not be re-used anywhere

idoshamun (approved):

packages/shared/src/components/sidebar/Sidebar.tsx
idoshamun: I agree with @sshanzel that this might be way easier to read. But if it feels like too much of a hassle at this point, we can leave it for the next one to pick it up whenever we need to touch this again.
This PR is already pretty massive so I understand if we don't want to make it even heavier

sshanzel: Agreed. I think we can move forward with this.

sshanzel (approved):
Well done! 💪

PR #747, opened 2021-12-21
refactor: user settings in boot query

sshanzel:
DD-337 #done

As part of serving the user's settings in a much faster time, we decided to move it to the boot query.

Since I was already working on the `SettingsContext`, I took the opportunity to refactor and apply some types with the strings we work with.

It may be noticed that the `const isExtension = process.env.TARGET_BROWSER;` was taken inside the component, the test was failing due to the fact the value was generated when we run the test, and simply overriding the value make it fail. So the component has to check it by the runtime to see if it actually has the correct value.

The related `boot` changes PR is now just waiting for additional checks but this should be safe for review. Related PR:
https://github.com/dailydotdev/daily-gateway/pull/394

### DIFF
--- a/packages/shared/src/contexts/SettingsContext.tsx
+++ b/packages/shared/src/contexts/SettingsContext.tsx
@@ -163,11 +163,11 @@ export const SettingsContextProvider = ({
     const theme = storageWrapper.getItem(themeModeStorageKey) as ThemeMode;
     if (theme) {
       setCurrentTheme(theme);
-      return;
-    }
-    const lightMode = storageWrapper.getItem(deprecatedLightModeStorageKey);
-    if (lightMode === 'true') {
-      applyTheme(ThemeMode.Light);
+    } else {
+      const lightMode = storageWrapper.getItem(deprecatedLightModeStorageKey);
+      if (lightMode === 'true') {
+        applyTheme(ThemeMode.Light);
+      }
     }
   }, []);
 
--- a/packages/shared/src/contexts/SettingsContext.tsx
+++ b/packages/shared/src/contexts/SettingsContext.tsx
@@ -136,7 +136,7 @@ export const SettingsContextProvider = ({
   );
 
   useEffect(() => {
-    if (remoteSettings) {
+    if (remoteSettings && userId) {
       const { theme: remoteTheme, ...remoteData } = remoteSettings;
       const theme = themeModes[remoteTheme];
       setCurrentTheme(theme);
--- a/packages/shared/src/contexts/SettingsContext.tsx
+++ b/packages/shared/src/contexts/SettingsContext.tsx
@@ -171,18 +171,22 @@ export const SettingsContextProvider = ({
     }
   }, []);
 
-  const setSettings = async (
+  const updateRemoteSettingsFn = async (
     newSettings: Settings,
-    remoteTheme?: RemoteTheme,
+    theme: ThemeMode,
   ): Promise<void> => {
-    await setCachedSettings(newSettings);
-
     if (userId) {
+      const remoteTheme = getRemoteTheme(theme);
       await updateRemoteSettings({
         ...newSettings,
         theme: remoteTheme,
       });
     }
+  };
+
+  const setSettings = async (newSettings: Settings): Promise<void> => {
+    await setCachedSettings(newSettings);
+    await updateRemoteSettingsFn(newSettings, currentTheme);
   };
 
   const contextData = useMemo<SettingsContextData>(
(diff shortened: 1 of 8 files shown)


packages/webapp/pages/_app.tsx
```diff
@@ -182,17 +181,15 @@ export default function App(props: AppProps): ReactElement {
...
-                <AnalyticsContextProvider
-                  app="webapp"
-                  version={version}
-                  getPage={getPage}
-                >
-                  <InternalApp {...props} />
-                </AnalyticsContextProvider>
-              </OnboardingContextProvider>
-            </SettingsContextProvider>
+            <OnboardingContextProvider>
```
rebelchris, line 184: Onboarding should be removed.
This already happened in the main 260 branch, you might have to rebase

rebelchris (changes requested):
@sshanzel Nice work here!
Really cool you made things much more strict.

I have some small comments so far.

packages/shared/src/contexts/SettingsContext.tsx
```diff
@@ -59,104 +63,120 @@ const defaultSettings: Settings = {
...
+const getThemeMode = (remoteTheme: RemoteTheme): ThemeMode => {
+  switch (remoteTheme) {
+    case 'bright':
+      return ThemeMode.Light;
+    case 'darcula':
+      return ThemeMode.Dark;
+    default:
+      return ThemeMode.Auto;
+  }
+};
```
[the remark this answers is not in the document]
sshanzel: I think that's a good idea, in that case, we shall create another object for switching it in a reversed manner.

```
const themes = {
  [ThemeMode.Light]: 'bright',
  [ThemeMode.Dark]: 'darcula',
}
```

packages/webapp/pages/_app.tsx
sshanzel: I think we should rebase the 260 once this is merged to master as we don't want the current progress at 260 to get merged into production already.

rebelchris: Yes, rebase this branch from 260 
The onboarding should be completely gone then (only author onboarding is still a thing)

packages/shared/src/contexts/SettingsContext.tsx
```diff
@@ -59,104 +63,120 @@ const defaultSettings: Settings = {
...
+
+function applyTheme(themeMode: ThemeMode): void {
+  if (document.documentElement.classList.contains(themeMode)) {
+    return;
+  }
+
+  if (themeMode === ThemeMode.Auto) {
```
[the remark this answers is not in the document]
sshanzel: I took inspiration from your suggestion, what do you think about the snippet below (already pushed)?
```
function applyTheme(themeMode: ThemeMode): void {
  if (document.documentElement.classList.contains(themeMode)) {
    return;
  }

  document.documentElement.classList.remove(ThemeMode.Light);
  document.documentElement.classList.remove(ThemeMode.Auto);

  if (themeMode === ThemeMode.Dark) {
    return;
  }

  document.documentElement.classList.add(themeMode);
}
```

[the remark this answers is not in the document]
sshanzel: function applyTheme(themeMode: ThemeMode): void {
  if (document.documentElement.classList.contains(themeMode)) {
    return;
  }

  document.documentElement.classList.remove(ThemeMode.Light);
  document.documentElement.classList.remove(ThemeMode.Auto);

  if (themeMode === ThemeMode.Dark) {
    return;
  }

  document.documentElement.classList.add(themeMode);
}

packages/webapp/pages/_app.tsx
sshanzel: No, I mean, this is pointed at `master` branch since this is safe to be merged. If we rebase this from `260` the current progress of `260` will be included and if we merge this, those features will be included unnecessarily.

Or do you mean we should point this to 260 so we can rebase first?

rebelchris: Oh sorry didn't notice this was meant for the master already.
Does make sense yes.

- Merge this into master
- Rebase 260 from master
- ☑️

rebelchris (approved):
Nice looks super clean like that!
I like it

Good from my side so far 👏

idoshamun (approved):
The changes look great and much way cleaner.
The only concern I have and I'd like to ask you to validate it, is that now anyone who has light theme will have a flash of colors on load.
Can you check it locally please?

sshanzel:
Locally it seems to flash from black to light in this branch - but I tried the `master` branch and it seems to happen locally as well.

With that, the deployments from this PR seem to have not flashed and are almost the same with production now. Were you concerned because of the removal of `windowLoaded`? We can definitely have it back to ensure expected workability.

Note: I have also declared a global clean-up for our persistent values as you have wonderfully suggested.

The additional commits: found an edge case for logged-out, anonymous users now prioritize local cache than boot's settings default values.

https://user-images.githubusercontent.com/13744167/147043814-eb284368-e38b-4787-82dd-787c9fbbad97.mov

idoshamun:
@sshanzel let's merge it and I'll check what happens.
Initially I had my concerns because now it takes several `useEffect` until the theme is applied but we'll see

sshanzel:
Sure, let's do it 🚢

PR #748, opened 2021-12-21
feat: test and fixes

rebelchris:
In this PR we are fixing some edge-cases and re-evaluated the test cases we had.
The cases actually seem significant enough.

Decided to add a mobile render check for the sidebar to see if we are not breaking the mobile version.

- 0553e88add3d8e92b87bba3bdb65f4fac00491eb -> Added more space according to SC for the mobile version
- 8e89d4eae50933ddb1fcf149bf8eb07d37c95c2f -> Mobile backdrop was visible on desktop if you expand from mobile to desktop
- 275a38f96049cf01f1b939d8e789a9101919891b -> Reading history alignment was not set
- 22473237770ba8bee92b8aa86b3f881e0e5aaa2c -> Added a autoFocus attribute for the postSearch component
- 3c74a7d9fee0918135bc5b82245f13222a6e363f -> Added the mobile render test case for the sidebar

DD-350 #done

### DIFF
--- a/packages/shared/src/components/PostsSearch.tsx
+++ b/packages/shared/src/components/PostsSearch.tsx
@@ -107,11 +107,11 @@ export default function PostsSearch({
     }
   };
 
-  if (autoFocus) {
-    useEffect(() => {
+  useEffect(() => {
+    if (autoFocus) {
       searchBoxRef.current?.querySelector('input').focus();
-    }, [searchBoxRef]);
-  }
+    }
+  }, [searchBoxRef, autoFocus]);
 
   const isOpen = !!menuPosition && !!items.length;
   return (
--- a/packages/shared/src/components/PostsSearch.tsx
+++ b/packages/shared/src/components/PostsSearch.tsx
@@ -15,12 +15,14 @@ const AutoCompleteMenu = dynamic(() => import('./fields/AutoCompleteMenu'), {
 export type PostsSearchProps = {
   initialQuery?: string;
   placeholder?: string;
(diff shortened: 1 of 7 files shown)


packages/shared/src/components/PostsSearch.tsx
```diff
@@ -105,9 +107,11 @@ export default function PostsSearch({
...
   };
 
-  useEffect(() => {
-    searchBoxRef.current?.querySelector('input').focus();
-  }, [searchBoxRef]);
+  if (autoFocus) {
+    useEffect(() => {
+      searchBoxRef.current?.querySelector('input').focus();
+    }, [searchBoxRef]);
+  }
```
idoshamun, line 114: Not sure conditional hooks are allowed in React.
It's better to put the condition inside the `useEffect`. I know it can mess their internals

packages/shared/src/components/sidebar/Sidebar.spec.tsx
```diff
@@ -162,3 +167,11 @@ it('should set all navigation urls', async () => {
     );
   });
 });
+
+it('should render the mobile sidebar version on small screens', async () => {
+  await resizeWindow(1019, 768);
```
idoshamun, line 172: Nice trick!

idoshamun (changes requested):
One change and we're ready to go!

rebelchris:
@idoshamun Oh right!
Refactored to evaluate inside the `useEffect` hook

idoshamun (approved):
Woohoo!

PR #749, opened 2021-12-21
fix: mobile sidebar height and invisible share text

rebelchris:
We had some issues with the mobile sidebar height on the mobile view because of the 100vh usage.
Also the share link was invisible on mobile devices.

Introducing a fallback for Safari to fix the push of the sidebar height.


idoshamun (approved):
Perfect! Just make sure that there are no regressions due to this change. You can check the modals, different pages, different resolution, etc

PR #750, opened 2021-12-21
feat: introducing a swipeable way to toggle the sidebar

rebelchris:
Introducing the swipe-able hook to determine if we should open or close the sidebar on mobile.

Getting weird issues saying the hook usage is wrong, but not sure why.
This seems like a valid way of using this hook?


packages/shared/src/components/MainLayout.tsx
```diff
@@ -89,6 +90,19 @@ export default function MainLayout({
...
+        setOpenMobileSidebar(false);
+      }
+    },
+    onSwipedRight: () => {
+      if (!sidebarRendered && !openMobileSidebar) {
+        setOpenMobileSidebar(true);
+      }
+    },
+    preventDefaultTouchmoveEvent: true,
+  });
```
idoshamun, line 105: This calls for a custom hook 😉

packages/shared/package.json
```diff
@@ -112,6 +112,7 @@
     "lazysizes": "^5.3.2",
     "linkifyjs": "^2.1.9",
     "lodash.clonedeep": "^4.5.0",
-    "node-fetch": "^2.6.6"
+    "node-fetch": "^2.6.6",
+    "react-swipeable": "^6.2.0"
```
idoshamun, line 116: What's the package size? Have you checked if there are any lighter alternatives?

idoshamun (changes requested):
Looks

packages/shared/package.json
rebelchris: Unpacked Size
153 kB

Can have a look what else is out there :)

sshanzel (approved):
As neat as it can get! Just the hook that Ido have mentioned then I think this should be good to go 🚢

packages/shared/src/components/MainLayout.tsx
rebelchris: Even with a custom hook i'm getting the error that the hook usage is wrong.
Do you have any idea why this occurs?

packages/shared/package.json
idoshamun: Please fix the build issue so we can see how much payload it adds to our bundles

rebelchris: @idoshamun Having some issues with React not liking this hook implementation.
Do you have any guidance on where to look for this issue?

packages/shared/src/hooks/useSwipeableSidebar.ts
```diff
@@ -0,0 +1,33 @@
...
+      if (!sidebarRendered && !openMobileSidebar) {
+        setOpenMobileSidebar(true);
+      }
+    },
+    preventDefaultTouchmoveEvent: true,
+  });
+
+  return useMemo(() => {
+    return {
+      ...handlers,
```
idoshamun, line 30:
```suggestion
      handlers,
```

idoshamun (changes requested):
This should do it

idoshamun (approved):
Yay!

PR #754, opened 2021-12-22
fix: window is not defined yet on shareable link

sshanzel:
When `window` is not yet defined, the shareable link can't be fetched yet.

### DIFF
--- a/packages/shared/src/components/ShareBar.tsx
+++ b/packages/shared/src/components/ShareBar.tsx
@@ -18,6 +18,7 @@ import classed from '../lib/classed';
 import AnalyticsContext from '../contexts/AnalyticsContext';
 import { postAnalyticsEvent } from '../lib/feed';
 import { SimpleTooltip } from './tooltips/SimpleTooltip';
+import ProgressiveEnhancementContext from '../contexts/ProgressiveEnhancementContext';
 
 const ShareButton = classed(Button, 'my-1');
 const ColorfulShareButton = classed(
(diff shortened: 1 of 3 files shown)


idoshamun (changes requested):
Great catch! But actually `windowLoaded` is used for deferred loading of resources. The best approach here is to check if `window` is defined. Also I encourage you to add this check into the function itself

sshanzel:
That makes sense since it is a function we can ensure the value returned will be what is expected.

sshanzel:
That makes sense since it is a function we can ensure the value returned will be what is expected.

I have tried to check if `window` exists but it seems to fail still. So now I just enclosed it with a try-catch to ensure workability.

idoshamun:
The undefined test should work. I think we test it using typeof try looking for it in the codebase

sshanzel:
Unfortunately, it is not working. I think it is because the static props are running without the context of browser or window and just running as JS which could be why the error displayed is coming from the static props.

I also tried `window === undefined` but it still gives the error.

image

idoshamun:
Something doesn't make sense to me because it shouldn't fail if you make sure this is true:
`typeof window !== 'undefined'`

sshanzel:
I am actually confused as well. I have tried different variations and all other cases to see what will happen. This is actually the first time I encountered `window` is not even recognized to be checked.

image

idoshamun:
You're missing the quotes on the type check. It should be `'undefined'`

sshanzel:
Oh, wow. It was just a matter of quotes 😆  

Quick link: https://daily-webapp-mg6d9eswa-dailydotdev.vercel.app/posts/_xOdpqEBi

rebelchris (approved):
NICE! 👏

idoshamun (approved):
Woohoo!

PR #755, opened 2021-12-23
fix: missing sidebar tooltips

sshanzel:
Preview:

Screen Shot 2021-12-23 at 2 14 03 PM


packages/shared/src/components/sidebar/common.tsx
```diff
@@ -92,50 +96,86 @@ export const RawNavItem = classed(
...
+  title,
+  tooltip = {},
+}: SidebarMenuItem) => (
+  <SimpleTooltip {...tooltip} content={title} placement="right">
     <span
       className={classNames(
-        'flex-1 text-left transition-opacity',
-        sidebarExpanded ? 'opacity-100 delay-150' : 'opacity-0',
+        'relative mr-3',
+        tooltip.visible !== undefined && 'pointer-events-none',
```
rebelchris, line 121: Just noticed this does break the mobile view.
The one for the invitePeople seems to work.

Screenshot 2021-12-23 at 10 18 28

rebelchris (changes requested):
Nice one!

Just one remark on the mobile version render.

packages/shared/src/components/sidebar/common.tsx
sshanzel: Oh, let me check this one.

sshanzel: How did you check this one? It seems to work on the current deployment.

Screen Shot 2021-12-23 at 2 29 55 PM

rebelchris: Oh this might be a edge-case issue that was already existing.

Steps to reproduce:
1. Open desktop version
2. Set to small version (non-expanded icon only)
3. Resize to mobile version
4. Open sidebar

sshanzel: I think this is a different issue, this happens to me as well. Whenever I try to resize the browser manually, everything doesn't seem to follow all the time.

sshanzel: I think this is a pre-existing issue, I suggest we fix it in a separate PR so we can test out the tooltips in preview.

rebelchris (approved):
Approved this one, I'll look into this existing issue

idoshamun (approved):
Awesome!

PR #759, opened 2021-12-23
fix: responsive modal class selector prioritization

sshanzel:
The values for responsive modal are being overridden due to the class selector prioritization.

Now we have added another selector to have higher priority when using `StyledModal`

Before:
Screen Shot 2021-12-23 at 4 34 22 PM

Preview:
image

### DIFF
--- a/packages/shared/src/components/modals/ResponsiveModal.module.css
+++ b/packages/shared/src/components/modals/ResponsiveModal.module.css
@@ -10,7 +10,7 @@
     }
   }
 
-  & :global(.modal) {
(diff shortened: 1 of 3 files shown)


rebelchris (approved):
Oh wow you found it!
Actually weird we couldn't reproduce this before 🤔

sshanzel:
Totally crazy, locally it works as intended, but on any preview deployment before, it behaves differently. So

idoshamun (approved):
Impressive catch!

PR #760, opened 2021-12-23
fix: use links instead of buttons for sidebar actions

idoshamun:
`useNavButtonsNotLinks` was applied on all the actions even though some of them had to be links in the extension.
This PR splits the usage of this property between the different sections.

DD-376 #done


rebelchris (approved):
You did also push a lock file for the shared and a vcs idea file.
Not sure what would have changed in the lock.

PR #761, opened 2021-12-23
fix: Only show render sidebar on explicit false value

rebelchris:


idoshamun (approved):

sshanzel (approved):
Oooh. That's a tricky one to find! Great fix!

PR #767, opened 2021-12-24
fix: hotfixes relayout

rebelchris:
Fixed a placement issue of the mobile article page not being full width
Tooltip minimized hover was not covering the full space

### DIFF
--- a/packages/webapp/pages/posts/[id].tsx
+++ b/packages/webapp/pages/posts/[id].tsx
@@ -238,7 +238,7 @@ const PostPage = ({ id, postData }: Props): ReactElement => {
 
   return (
     <>
-      <div className="flex relative flex-col flex-wrap mx-auto w-full max-w-full">
+      <div className="flex relative flex-col flex-wrap mx-auto w-full laptopL:w-auto max-w-full">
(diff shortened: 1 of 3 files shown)


packages/webapp/pages/posts/[id].tsx
```diff
@@ -238,7 +238,7 @@ const PostPage = ({ id, postData }: Props): ReactElement => {
 
   return (
     <>
-      <div className="flex relative flex-col flex-wrap mx-auto max-w-full">
+      <div className="flex relative flex-col flex-wrap mx-auto w-full laptopL:w-auto max-w-full">
```
Comment (idoshamun) [inline packages/webapp/pages/posts/[id].tsx:241]: Why the `laptopL:auto`? We already set the width for this resolution in the PageContainer element

idoshamun (changes requested):
Looks great just one question

packages/webapp/pages/posts/[id].tsx
Comment (rebelchris) [reply packages/webapp/pages/posts/[id].tsx:241]: weird case where the full class would overwrite the one we set.
Auto would just take that one and unset the full width.

Comment (idoshamun) [reply packages/webapp/pages/posts/[id].tsx:241]: Ok!

idoshamun (approved):

PR #768, opened 2021-12-24
fix: sharebar copied text placement

sshanzel:
Preview before the fix:

image

Preview after the fix:
image


rebelchris (approved):
Well spotted Lee!

Thanks for picking this one up

idoshamun (approved):
Spot on!!

PR #772, opened 2021-12-27
fix: fix flagged version of login button

rebelchris:
Added the login button flagged version to be uniform.

Do note for the mobile sidebar version it will always render "Login" but at least this makes everything loaded from 1 component now.

DD-390 #done

### DIFF
--- a/packages/extension/package.json
+++ b/packages/extension/package.json
@@ -1,6 +1,6 @@
 {
   "name": "extension",
-  "version": "3.11.0",
+  "version": "3.11.1",
   "scripts": {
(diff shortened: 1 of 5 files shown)


sshanzel (approved):
🚢

idoshamun (changes requested):
The default value for the `buttonCopy` should be the experiment winner: `Access all features`.
I also need to validate with Tsahi and Nimrod what we want to do with the mobile version. It's a good point

rebelchris:
@idoshamun Updated the default copy, awaiting what we need to do with the mobile text.

idoshamun (approved):
Looks great!

PR #773, opened 2021-12-27
fix: register was not centered and timezone was not in default set

rebelchris:
Made sure the register page is centered and the timezone is part of the main fields as this is important information for the onboarding.

DD-391 #done


idoshamun (approved):
Perfect

PR #774, opened 2021-12-27
fix: devcard notif should not show when explicitly defined

sshanzel:
DD-393 #done

The previous mechanism for identifying the `rankLastSeen` [was removed](https://github.com/dailydotdev/apps/pull/763/files#diff-591811d5ed8112fe53d4882f8cfd46634f04cb39e40a12c0086ef7fe21510158L76-L84) but I have also realized to use the same process the other providers do when checking for the remote value's availability.

### DIFF
--- a/packages/shared/src/hooks/useReadingRank.ts
+++ b/packages/shared/src/hooks/useReadingRank.ts
@@ -39,13 +39,14 @@ const defaultRank: MyRankData = {
 const checkShouldShowRankModal = (
   rankLastSeen: Date,
   lastReadTime: Date,
+  loadedAlerts: boolean,
   neverShowRankModal?: boolean,
 ) => {
   if (neverShowRankModal !== null && neverShowRankModal !== undefined) {
     return !neverShowRankModal;
   }
 
-  if (rankLastSeen === undefined) {
+  if (loadedAlerts && !rankLastSeen) {
     return true;
   }
 
@@ -92,8 +93,9 @@ export default function useReadingRank(): ReturnType {
     });
 
   const shouldShowRankModal = checkShouldShowRankModal(
-    loadedAlerts && alerts?.rankLastSeen,
+    alerts?.rankLastSeen,
     cachedRank?.rank?.lastReadTime || remoteRank?.rank?.lastReadTime,
+    loadedAlerts,
     neverShowRankModal,
   );
 
--- a/packages/shared/src/hooks/useReadingRank.ts
+++ b/packages/shared/src/hooks/useReadingRank.ts
@@ -42,15 +42,11 @@ const checkShouldShowRankModal = (
   neverShowRankModal?: boolean,
 ) => {
   if (neverShowRankModal !== null && neverShowRankModal !== undefined) {
-    return neverShowRankModal;
+    return !neverShowRankModal;
   }
 
-  if (rankLastSeen === undefined) {
+  if (!rankLastSeen) {
     return true;
-  }
-
-  if (rankLastSeen === null) {
-    return false;
   }
 
   if (!lastReadTime) {
@@ -61,7 +57,7 @@ const checkShouldShowRankModal = (
 };
 
 export default function useReadingRank(): ReturnType {
-  const { alerts, updateAlerts } = useContext(AlertContext);
+  const { alerts, loadedAlerts, updateAlerts } = useContext(AlertContext);
   const { user, tokenRefreshed } = useContext(AuthContext);
   const [levelUp, setLevelUp] = useState(false);
   const queryClient = useQueryClient();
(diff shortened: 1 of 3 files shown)


packages/shared/src/hooks/useReadingRank.ts
```diff
@@ -42,26 +42,22 @@ const checkShouldShowRankModal = (
...
   if (rankLastSeen === undefined) {
     return true;
   }
 
-  if (rankLastSeen === null) {
-    return false;
-  }
-
-  if (!lastReadTime) {
+  if (!rankLastSeen) {
```
idoshamun, line 52: Why if it's `null` or `false` we need to not show the rank modal?
In what cases the `rankLastSeen` can be `null` or `false`?

idoshamun (changes requested):
Please make sure this works both for registered users and anonymous users before merging

packages/shared/src/hooks/useReadingRank.ts
sshanzel: The value `null` is our default value `onload` while we are still fetching the data of `rankLastSeen`, we don't want to accidentally show it - the reason for `false` is the way we pass the value for `rankLastSeen` like the snippet below. So when the value is not yet loaded (`loadedAlerts`) it will return false but if it is fetched already, it should give the actual value from `alerts.rankLastSeen` (be it `undefined` or the actual `Date`).

With that, I have noticed a different unexpected behavior related to this. Let me check that one out. 

https://github.com/dailydotdev/apps/blob/3e17c396312d2fe7b478866358cc92a716b7e730/packages/shared/src/hooks/useReadingRank.ts#L95

rebelchris (approved):
Besides Ido's comments nothing to add. 🎉

packages/shared/src/hooks/useReadingRank.ts
idoshamun: If `rankLastSeen` is null while loading then why do we need `loadedAlerts` for?

sshanzel: That is so we can indicate that the `null` value was not the default one but sent from the remote settings since our DB value is nullable. If it is `null` because of the default values, we should not show anything yet.

For this to be much clearer, let me take the `loadedAlerts` inside the function for easier readability and maintainability.

sshanzel:
@idoshamun has tested on both registered and unregistered and working as intended. I also included a fix for new users not displaying the modal due to the [additional condition](https://github.com/dailydotdev/apps/pull/774/commits/df2d8c6443226e708e132e4e7341a1c6bbeb8753).

The removed condition should not worry about leveling up and displaying the modal since we have a specific constant that thoroughly checks all the parameters.

sshanzel:
@idoshamun has tested on both registered and unregistered and working as intended. I also included a fix for new users of the first week not displaying the progress due to the additional condition.

Instead of being part of the first if statement where it just sets the value, it must also check if it still needs to display the animation.

packages/shared/src/hooks/useReadingRank.ts
```diff
@@ -115,19 +113,22 @@ export default function useReadingRank(): ReturnType {
...
         !cachedRank ||
-        remoteRank.rank.progressThisWeek < cachedRank.rank.progressThisWeek ||
-        !cachedRank.rank.rankLastWeek
+        remoteRank.rank.progressThisWeek < cachedRank.rank.progressThisWeek
       ) {
         cacheRank();
-        return;
+      } else if (cachedRank && !cachedRank.rank.rankLastWeek) {
+        cacheRank();
+        displayProgress();
```
idoshamun, line 128: Why do we need this case?
I'd suggest adding a comment above every case in this use effect. It's somewhat confusing

idoshamun (changes requested):
One question before we can proceed.

packages/shared/src/hooks/useReadingRank.ts
sshanzel95: After checking [this PR](https://github.com/dailydotdev/apps/pull/736), the condition `!cachedRank.rank.rankLastWeek` was added to fix a weird animation behavior. To support the case I kept its sole intention of updating the cache. But then the block statement where it was included, it returns immediately, it results in the progress not being shown for users on the first week that should be valid for animation - like they've read 4 consecutive days but since the `rank.rankLastWeek` was `null` they skipped the animation.

sshanzel: After checking this PR, the condition !cachedRank.rank.rankLastWeek was added to fix a weird animation behavior. To support the case I kept its sole intention of updating the cache. But then the block statement where it was included, it returns immediately, it results in the progress not being shown for users on the first week that should be valid for animation - like they've read 4 consecutive days but since the rank.rankLastWeek was null they skipped the animation.

sshanzel: After checking [this PR](https://github.com/dailydotdev/apps/pull/736), the condition `!cachedRank.rank.rankLastWeek` was added to fix a weird animation behavior. To support the case I kept its sole intention of updating the cache. But then the block statement where it was included, it returns immediately, it results in the progress not being shown for users on the first week that should be valid for animation - like they've read 4 consecutive days but since the rank.rankLastWeek was `null` they skipped the animation.

sshanzel: Let me add the comments now.

idoshamun (approved):

idoshamun:
Thanks a lot!

PR #775, opened 2021-12-27
fix: refactor usecases for the default feed state on reload

rebelchris:
Due to the recent changes we were no longer saving the default feed in the storage.
Since we changed how we load the feed, we had to reintroduce this both on the webapp and the extension side.

I've tested both and they work consistently now.

DD-394 #done


packages/extension/src/newtab/MainFeedPage.tsx
```diff
@@ -28,7 +29,11 @@ export default function MainFeedPage({
   onPageChanged,
 }: MainFeedPageProps): ReactElement {
   const { user } = useContext(AuthContext);
-  const [feedName, setFeedName] = useState<string>('default');
+  const [feedName, setFeedName] = usePersistentState(
```
idoshamun, line 32: Why is it used both here and in the `MainFeedLayout`?

packages/shared/src/components/MainFeedLayout.tsx
```diff
@@ -112,8 +112,17 @@ export default function MainFeedLayout({
...
+  const [defaultFeed, setDefaultFeed] = usePersistentState(
+    'defaultFeed',
+    null,
+    'popular',
+  );
   const feedName = feedNameProp === 'default' ? defaultFeed : feedNameProp;
+
+  if (defaultFeed !== null && feedName !== null && feedName !== defaultFeed) {
+    setDefaultFeed(feedName);
+  }
```
idoshamun, line 124: Not sure I understand the condition here and also make sure to use `useEffect` for this purpose

packages/shared/src/components/sidebar/Sidebar.tsx
```diff
@@ -126,7 +127,9 @@ export default function Sidebar({
   setOpenMobileSidebar,
   onShowDndClick,
 }: SidebarProps): ReactElement {
-  const activePage = activePageProp === '/' ? '/popular' : activePageProp;
+  const [defaultFeed] = usePersistentState('defaultFeed', null, 'popular');
```
idoshamun, line 130: Now I see that it is used here as well. Not a good practice all these values will not be in sync during the session and it could create some real mess.

idoshamun (changes requested):

rebelchris:
@idoshamun Refactored to a single source of modification.
Being the common component between extension and webapp: `MainFeedLayout`.

This just needed to send an update to the extension, which will happen once on the load change of the default state initialising.

idoshamun (changes requested):
I see that you still read the persistent state from two different place. Can you unify it please?
This may cause unstable behavior

rebelchris:
@idoshamun It's a bit difficult to refactor it from one place, as there is not top level shared component between the extension and the webapp.

Decided to refactor into a custom hook, the setting only happens in the shared file, but some other parts read from this custom hook.

Let me know what you think about this approach.

rebelchris:
@sshanzel @idoshamun I've added a new hook called `usePersistentContext`.
This is basically a mix between `usePersistentState` and `useQuery`.

This way we can create storage that is changeable throughout the application.
For now we decided on only using it for the `defaultFeed` variable.

But we can extend this in the future so `usePersistentState` will get replaced by this version as it's more reliable.
The usage of this hook is the same as the `usePersistentState` one, except it doesn't come with a `initialValue`.

idoshamun (approved):
Woohoo!!

PR #776, opened 2021-12-27
fix: on slow devices the boot would load after update overwritting

rebelchris:
The cached data would be overwritten while updating on slow connections.
We introduced a temporary update state that has priority.


idoshamun (approved):
💪

PR #777, opened 2021-12-27
fix: registration stuck-up on logout

sshanzel:
The reason for getting stuck on logout while at the registration page is we do `reload` only when logging out.

To better catch the case, we first check
- if there is any `redirect_uri` to utilize for redirection.
- if none - we redirect to our app's home page if it is logging out at the registration page (just to make sure to catch all cases).
- else we reload like how it is supposed to be before


rebelchris (approved):
Nice find! 🥳

idoshamun (approved):
Very nice! Please confirm that it works after the deployment

sshanzel:
Will immediately check once deployed 🚢

PR #778, opened 2021-12-27
fix: redirect user to registration if not yet confirmed

sshanzel:
Clicking the logout button on registration already works on the extension, but on the webapp, there needs an additional check whether we should `reload` to clear the local state and data (as it doesn't automatically clear up by the `window.location.replace` which should have been).

I have also found a bug (which I don't know if this should be directed at `nextjs` repo). In our `_app.tsx` file, it should detect when the user has arrived - like `window.location.reload` it goes through the `useEffect` as mentioned before it should also happen with `window.location.replace` since we are not using the `next/link` or `next/router` which is the native redirect to keep all local state and data.

Reproducing the bug mentioned above:
1. Register
2. Once the form is displayed, click our dailydev logo on the top-most left corner.
3. This won't redirect even though the user info has not yet been confirmed.

The expected behavior on step 3 should be redirected to the form again.

Fix: since `nextjs` acknowledges the changes in the route even using `window.location.replace` as nextjs-native redirect - we have included the `router.pathname` as a dependency.

### DIFF
--- a/packages/webapp/pages/_app.tsx
+++ b/packages/webapp/pages/_app.tsx
@@ -99,6 +99,10 @@ function InternalApp({ Component, pageProps, router }: AppProps): ReactElement {
       !user.infoConfirmed &&
       window.location.pathname.indexOf(REGISTRATION_PATH) !== 0
     ) {
+      /*  this will check if the user came from our registration
+          if the user has clicked the logout button in the registration - it should go through here because the user will be redirected
+          and since the window.location.replace did not clear all the local state and data like window.location.replace - we will have to force a reload
+      */
       const referrer = document.referrer && new URL(document.referrer);
       if (referrer.pathname?.indexOf(REGISTRATION_PATH) !== -1) {
         window.location.reload();
--- a/packages/webapp/pages/_app.tsx
+++ b/packages/webapp/pages/_app.tsx
@@ -13,7 +13,9 @@ import 'focus-visible';
 import Modal from 'react-modal';
 import { DefaultSeo } from 'next-seo';
 import { QueryClient, QueryClientProvider } from 'react-query';
-import AuthContext from '@dailydotdev/shared/src/contexts/AuthContext';
+import AuthContext, {
+  REGISTRATION_PATH,
+} from '@dailydotdev/shared/src/contexts/AuthContext';
 import { Router } from 'next/router';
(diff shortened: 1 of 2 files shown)


packages/webapp/pages/_app.tsx
```diff
@@ -95,14 +97,20 @@ function InternalApp({ Component, pageProps, router }: AppProps): ReactElement {
...
       user &&
       !user.infoConfirmed &&
-      window.location.pathname.indexOf('/register') !== 0
+      window.location.pathname.indexOf(REGISTRATION_PATH) !== 0
     ) {
+      const referrer = document.referrer && new URL(document.referrer);
+      if (referrer.pathname?.indexOf(REGISTRATION_PATH) !== -1) {
+        window.location.reload();
+        return;
+      }
```
idoshamun, line 106: Please add a comment what exactly this check does and why

idoshamun (changes requested):
Please double check it because if it has the potential to create a redirect loop or some undesired behavior.

packages/webapp/pages/_app.tsx
sshanzel: Yes, I have tested locally multiple times. If it came from the registration and the local state/data has not been cleared, we will have to force a `reload`. I have also tried all different combinations including on extensions.

This should not produce a loop since the `document.referrer` will change once the the reload has happened, so it won't be valid since it is not the registration path anymore. 
Coming from the registration page? The condition above this checks if it should not enter this block by validating if it is currently the registration page we are looking at.

The only loop-like behavior will be if the user doesn't fill-up the form (`user.infoConfirmed = false`) and keeps on clicking our dailydev logo (wants to go to the home page) which is our expected behavior since we want to finish up the form first before proceeding with the app if you are trying to register.

idoshamun (approved):
Awesome! Please check that in prod once it's deployed

sshanzel:
Just checked the webapp, it seems to work as intended on registration now 🚢

idoshamun:
Thanks! Can you confirm it works on the extension as well?

sshanzel:
Yes! The initial fix actually worked already for the extension, after the changes in this PR, it doesn't seem to have changed. Thank you! 🚀

idoshamun:
Super! It's also important to check for regressions. This is why I asked. :)

sshanzel:
I totally agree! Sometimes even subtle changes can affect a lot things 🚢

PR #779, opened 2021-12-28
refactor: article page classes layout

sshanzel:
DD-392 #done

This might require a bit of focus when reviewing (sorry about that) since I have refactored the classes and made the containers to use fewer and simple classes as much as possible.

Also, instead of applying the classes in one line (if there's a lot of differences), with `classed` I realized it accepts a rest argument which we can utilize to separate the different screen classes (as it is a bit difficult for complex layouts in one line) for easier understanding and readability. But it depends if it makes sense, probably just my personal pov.

```
classed(
  'tag',
  'base classes'
  'laptop classes',
  'tablet classes'
)
```

The jumpy widgets' container was caused by the `right-0` position which sets the element to the edge initially.

I have used flexbox so it will automatically set the placement (side-by-side) regardless of the box model and positioning. This also improves predictability. IMHO `absolute` values are pretty unstable at times, so I've decided to refactor the positioning.

I have also taken the `NewComment` inside the `PageContainer` so it won't have to think about its own box model styling which is basically the same with `PageContainer`. I have also removed its nested `div` as its classes were not necessary anymore.

**Note: I have used the light theme for the current layout - the dark theme is for the updated classes (this PR)**

I have used the browser's defined screen sizes in the console for toggling between previews.

**Screen resolution: 1440p**
current:
1440p - old
new:
1440p

**Screen resolution: 1080p**
current:
1080p - old
new:
1080p

**Screen resolution: iPad Pro**
current:
iPad Pro - old
new:
iPad Pro

**Screen resolution: iPad**
current:
iPad - old
new:
iPad

**Screen resolution: Mobile iPhone 6/7/8 Plus**

current:
Mobile iPhone 6:7:8 Plus - old
new:
Mobile iPhone 6:7:8 Plus

Screen recording for the behavior on load:
https://drive.google.com/file/d/1Hu5Zf9He9HutzuPZuXnFYzSh_oSsBsSk/view?usp=sharing

### DIFF
--- a/packages/shared/src/components/utilities.module.css
+++ b/packages/shared/src/components/utilities.module.css
@@ -16,11 +16,7 @@
   }
 }
 
-.newCommentContainer {
-  max-width: $pageMaxWidth;
-}
-
-.pageSidebar {
+.pageWidgets {
   max-width: $pageMaxWidth;
 }
 
--- a/packages/shared/src/components/utilities.tsx
+++ b/packages/shared/src/components/utilities.tsx
@@ -13,22 +13,30 @@ export const LegalNotice = classed(
   styles.legal,
 );
 
+export const pageBorders =
+  'laptop:border-r laptop:border-l laptop:border-theme-divider-tertiary';
+const pagePaddings = 'px-4 tablet:px-8';
+
 export const PageContainer = classed(
   'main',
   styles.pageContainer,
-  'relative flex flex-col w-full items-stretch px-4 z-1 tablet:px-8 tablet:self-center',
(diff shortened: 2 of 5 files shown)


rebelchris (approved):
Oh nice!
This is how I started, but it gave issues with the format of the newComment on the tablet view.

Seems like that has been fixed now, might be because of the new full width change.
No particular comments, looks to work on all types of content from what I tested on your version.

👏👏👏

idoshamun (approved):
Sweet! Great catch 👀

PR #780, opened 2021-12-28
test: update local alerts and mutation to be called

sshanzel:
DD-399

Explicit test cases to ensure whenever new changes are introduced things won't be broken.

We have realized the two items below require attention:
- should utilize default front-end settings for a first-time user.
- should not override settings for an anonymous user and utilize local settings.
- should call `updateAlerts` with mock
~~should keep the stored local version on double execution (still discussing with Chris about this case)~~

### DIFF
--- a/packages/shared/src/components/Settings.spec.tsx
+++ b/packages/shared/src/components/Settings.spec.tsx
@@ -20,6 +20,9 @@ import { LoggedUser } from '../lib/user';
 import defaultUser from '../../__tests__/fixture/loggedUser';
 import AuthContext from '../contexts/AuthContext';
 import { LoginModalMode } from '../types/LoginModalMode';
+import { BootDataProvider, BOOT_LOCAL_KEY } from '../contexts/BootProvider';
+import { apiUrl } from '../lib/config';
+import { BootCacheData } from '../lib/boot';
 
 beforeEach(() => {
   jest.clearAllMocks();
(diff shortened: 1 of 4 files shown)


sshanzel:
After discussing with Chris, the last item won't be possible to replicate since we even had to throttle the network speed to get the case being caught. Also, the case was related to getting out of focus on the browser and back to it to execute refetching the remote data.

rebelchris (approved):
Nice work on this!
These test looks significant enough for what we are testing 🤘

idoshamun (changes requested):
The tests look great! But I think this is not enough.
I'd like to see more test cases for the boot provider as it's a critical component in the system and can potentially break completely our app.
For example, we need to test the update alerts functions that was already broke in the last version. But we can test many more cases just to make sure boot provider works properly.
What do you think? It's also possible to do these in a different PR if you'd like.

sshanzel:
@idoshamun totally agree. We need to test as much as possible for the BootProvider as it is very critical.

In the updates I had for the alerts in this PR, it covers the issue we had last time not being able to execute the function.

Locally I have also tested if it would fail and it did when I reverted back the previous issue.

We can definitely create another PR to be solely focused on BootProvider.

idoshamun:
@sshanzel I'd love to see a dedicated PR with tests for the BootProvider

idoshamun (approved):

PR #782, opened 2021-12-29
fix: :lipstick: Fixing the article position

rebelchris:
We introduced a `tablet:self-center` on the PageContainer class, however for the article page this needs to stretch on laptop devices.

There are edge cases where the sidebar would be longer than the main part, forcing this to "center" while we want to stretch this to get the borders show fully.

DD-426 #done


idoshamun (approved):

idoshamun:
Nice catch! Feels like the article page is giving some a headache recently. Not sure what we can do about it

sshanzel (approved):
Nice catch!

PR #783, opened 2021-12-29
feat: my feed

sshanzel:
**Work In-Progress**

I just pushed a commit so we can create our draft PR for "My Feed".

Even though the initial commit was related to the dedicated page for custom feed, feel free to take that task if you want.


sshanzel:
I have now created a draft PR so we can merge pull from the master by just clicking the "Update branch" button.

cc @rebelchris @idoshamun

rebelchris (approved):
Woop woop let's ship this baby 🚀⛴

sshanzel (commented):
Looks good to my eyes 😍  Woop! Woop!

cc @idoshamun @rebelchris

idoshamun (approved):
This looks great!
The only I want you to confirm is that the analytics events for following and blocking tags are sent even when the user is not registered.
I want to make sure we create a proper funnel of events and investigate the user behavior.
Can you validate it before you merge?
You can check the network panel after clicking a tag in the manage tags window. It should dispatch a request to `/e`. It's ok it fails because you don't have the server running but check what's the payload of this request

sshanzel:
Below are the payloads for making changes as anonymous user 🤩 


Following payload:
Screen Shot 2022-01-13 at 7 33 13 PM

Blocking payload:
Screen Shot 2022-01-13 at 7 33 54 PM

PR #784, opened 2021-12-30
feat: added the MyFeedButton component

rebelchris:
This PR introduces the MyFeedButton component.
This component can render based on the alert flag.

1. User already filtered something:
2. User hasn't filtered yet

Option 1 will show the My feed button
Option 2 will show the flagged text component.

The different states:

Filtered expanded:
Screenshot 2021-12-30 at 15 19 45

Filtered small:
Screenshot 2021-12-30 at 15 19 55

Unfiltered expanded:
Screenshot 2021-12-30 at 15 19 16

Unfiltered small:
Screenshot 2021-12-30 at 15 19 22

For this component we also introduced 4 new flagsmith values:
1. MyFeedButtonCopy = 'my_feed_button_copy',
2. MyFeedButtonColor = 'my_feed_button_color',
3. MyFeedExplainerCopy = 'my_feed_explainer_copy',
4. MyFeedExplainerColor = 'my_feed_explainer_color',

### DIFF
--- a/packages/shared/src/components/sidebar/MyFeedButton.tsx
+++ b/packages/shared/src/components/sidebar/MyFeedButton.tsx
@@ -1,0 +1,114 @@
+import React, { ReactElement, useContext } from 'react';
+import classNames from 'classnames';
+import PlusIcon from '@dailydotdev/shared/icons/plus.svg';
+import FilterIcon from '@dailydotdev/shared/icons/outline/filter.svg';
+import { IFlags } from 'flagsmith';
+import { Button } from '../buttons/Button';
+import FeaturesContext from '../../contexts/FeaturesContext';
+import { Features, getFeatureValue } from '../../lib/featureManagement';
+import { ButtonOrLink, ItemInner, NavItem, SidebarMenuItem } from './common';
+
+const UnfilteredMyFeedButton = ({
+  sidebarExpanded,
+  flags,
+  action,
+}: {
+  sidebarExpanded: boolean;
+  flags: IFlags;
+  action: () => unknown;
+}) => {
+  const buttonCopy =
+    getFeatureValue(Features.MyFeedButtonCopy, flags) || 'Create my feed';
+  const explainerCopy =
+    getFeatureValue(Features.MyFeedExplainerCopy, flags) ||
+    'Devs with a personal feed get 11.5x more relevant articles';
+
+  return (
+    <div
+      className={classNames(
+        'flex flex-col items-center rounded-12 border-theme-status-success',
+        sidebarExpanded
+          ? 'shadow-2-avocado border p-3 m-4'
+          : 'mx-3 mb-2.5 mt-16',
+      )}
+    >
+      <p
+        className={classNames(
+          'typo-footnote  transition-all ',
+          sidebarExpanded
+            ? 'mb-3 h-auto transform opacity-100 ease-linear duration-200 translate-y-0 delay-200'
+            : 'm-0 transform duration-0 delay-0 opacity-0 translate-y-full h-0',
+        )}
+      >
+        {explainerCopy}
+      </p>
+
+      <Button
+        className="btn-primary-avocado"
+        buttonSize={sidebarExpanded ? 'medium' : 'xsmall'}
+        icon={<PlusIcon />}
+        iconOnly={!sidebarExpanded}
+        onClick={action}
+      >
+        {sidebarExpanded && buttonCopy}
+      </Button>
+    </div>
+  );
+};
+
+const FilteredMyFeedButton = ({
+  item,
+  sidebarExpanded,
+  action,
+}: {
+  item: SidebarMenuItem;
+  sidebarExpanded: boolean;
+  action: () => unknown;
+}) => {
+  return (
+    <NavItem className="mt-6">
+      <ButtonOrLink item={item}>
+        <ItemInner item={item} sidebarExpanded={sidebarExpanded} />
+      </ButtonOrLink>
+      <Button
+        iconOnly
+        className="mr-3"
+        buttonSize="xsmall"
+        icon={<FilterIcon />}
+        onClick={action}
+      />
+    </NavItem>
+  );
+};
+
+export default function MyFeedButton({
+  menuItem,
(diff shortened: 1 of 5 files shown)


packages/shared/src/components/sidebar/MyFeedButton.tsx
```diff
@@ -0,0 +1,159 @@
...
+  sidebarExpanded,
+  flags,
+  action,
+}: {
+  sidebarExpanded: boolean;
+  flags: IFlags;
+  action: () => unknown;
+}) => {
+  const buttonCopy =
+    getFeatureValue(Features.MyFeedButtonCopy, flags) || 'Create my feed';
```
idoshamun, line 54: I think it's better to add a default value support for `getFeatureValue`, it would look much nicer. What do you think?

packages/shared/src/components/sidebar/Sidebar.tsx
```diff
@@ -187,6 +179,13 @@ export default function Sidebar({
     },
   ];
 
+  const myFeedMenuItem: SidebarMenuItem = {
+    icon: <ListIcon Icon={HomeIcon} />,
+    title: 'My feed',
+    path: '/my-feed',
```
idoshamun, line 185: Should we go with `my-feed` or `myFeed`?

packages/shared/src/components/sidebar/Sidebar.tsx
```diff
@@ -248,6 +247,14 @@ export default function Sidebar({
               sidebarRendered={sidebarRendered}
               onShowDndClick={onShowDndClick}
             />
+            {sidebarRendered && (
+              <MyFeedButton
+                sidebarExpanded={sidebarExpanded}
+                filtered={!alerts?.filter}
+                menuItem={myFeedMenuItem}
+                action={setLoaded}
```
idoshamun, line 255: I know it's been here before but `setLoaded` is not explicit and it's not clear what it is doing only by reading the name. I advise renaming it.

idoshamun (changes requested):
Thanks for the great explanation in the PR's body. It was super helpful!

packages/shared/src/components/sidebar/MyFeedButton.tsx
rebelchris: Great point, i'll look into this 👏

packages/shared/src/components/sidebar/Sidebar.tsx
rebelchris: I was thinking from an URL point of view `my-feed` would look nicer, so we could use that?
Let me know what you and @sshanzel think.

packages/shared/src/components/sidebar/Sidebar.tsx
rebelchris: Agreed! Will rework this

packages/shared/src/components/sidebar/MyFeedButton.tsx
```diff
@@ -0,0 +1,159 @@
...
+
+const FilteredMyFeedButton = ({
+  item,
+  sidebarExpanded,
+  action,
+}: {
+  item: SidebarMenuItem;
+  sidebarExpanded: boolean;
+  action: () => unknown;
+}) => {
```
sshanzel, line 112: I think it would be nice there would be a type that they can share as I see some properties available on both components.

packages/shared/src/components/sidebar/MyFeedButton.tsx
```diff
@@ -0,0 +1,159 @@
...
+  menuItem,
+  sidebarExpanded,
+  filtered = false,
+  action,
+}: {
+  menuItem: SidebarMenuItem;
+  sidebarExpanded: boolean;
+  filtered: boolean;
+  action: () => unknown;
+}): ReactElement {
```
sshanzel, line 139: Just tagging this part as well.

packages/shared/src/components/sidebar/Sidebar.spec.tsx
```diff
@@ -104,7 +104,7 @@ const renderComponent = (
   );
 };
 
-it('should remove alert dot for filter alert when there is a pre-configured feedSettings', async () => {
+it('should remove feed filter button once user has filters', async () => {
```
sshanzel, line 107: I think this test was reused for something else. This is for making the alerts go away, the right test should be just checking if the button is gone on load if the user has filters.

sshanzel (changes requested):
The changes were neat! Just had some comments but great work here!

packages/shared/src/components/sidebar/Sidebar.tsx
rebelchris: As this hook is generic, I've extracted this name in the assignment.
Let me know if that works for this case.

packages/shared/src/components/sidebar/Sidebar.spec.tsx
rebelchris: This test is to specifically see if the mutation is still called.
I've added another test to check if the button changes it's this one:
`should show the my feed items if the user has filters`.

Let me know if that makes sense?

rebelchris:
@sshanzel Thanks for mentioning that.
Missed it in my own final refactor 😅.

Should be ready for your eyes again.

packages/shared/src/components/sidebar/Sidebar.tsx
sshanzel: Have the same opinion as Chris for the URL pov.

packages/shared/src/components/sidebar/Sidebar.spec.tsx
sshanzel: This mutation should be executed once the user has clicked the close button if I'm not mistaken. It would be the `alert.filter` to decide whether a message is displayed. The feed filter button is decided by `userFeedSettings`.

sshanzel (approved):
Great one! 💯

rebelchris:
@idoshamun should be ready for your review

idoshamun (approved):
Woop woop! 💪

PR #786, opened 2021-12-31
feat: introducing the new my feed alert option

rebelchris:
This PR introduces the `myFeed` filter.
This is a separate implementation from the actual button, so it will need some work once both are merged to finish this off and add the test to see if all the frontend actions work well together.

However we introduce this new alert once the user has the `my feed` available, depending on how it got available we show different text.

We added it for them:
Screenshot 2021-12-31 at 16 54 29

They filtered themselves:
Screenshot 2021-12-31 at 16 48 29

Related PR's
https://github.com/dailydotdev/daily-api/pull/843
https://github.com/dailydotdev/daily-gateway/pull/403

Known elements missing from this PR:
- Setting the alert to manual
- Linking this alert to the new feed button
- Setting the dynamic height of this section
- The link for the Learn more element
- Test cases for the flow, as we need the other UI elements for this

DD-429 #done

### DIFF
--- a/packages/shared/src/components/sidebar/MyFeedAlert.tsx
+++ b/packages/shared/src/components/sidebar/MyFeedAlert.tsx
@@ -6,7 +6,7 @@ import { AlertContainer, AlertCopy } from './common';
 import { Pointer, PointerColor } from '../Pointer';
 
 const alertCopy = {
-  migrated: `Psst, your feed has a new name! We've already applied your content filters to it. `,
+  migrated: `Psst, your feed has a new name! We've already applied your content filters to it.`,
   created: `🎉 Your feed is ready! Click here to manage your feed's settings`,
 };
 
--- a/packages/shared/src/components/sidebar/MyFeedAlert.tsx
+++ b/packages/shared/src/components/sidebar/MyFeedAlert.tsx
@@ -18,10 +18,6 @@ export default function MyFeedAlert({
   alerts,
   hideAlert,
 }: MyFeedAlertProps): ReactElement {
-  if (alerts?.filter || (!alerts?.filter && alerts?.myFeed === null)) {
-    return <></>;
-  }
-
   return (
     <AlertContainer>
       <Pointer color={PointerColor.Success} />
--- a/packages/shared/src/components/sidebar/MyFeedAlert.tsx
+++ b/packages/shared/src/components/sidebar/MyFeedAlert.tsx
@@ -26,7 +26,7 @@ export default function MyFeedAlert({
         {alerts.myFeed === 'migrated' && (
           <a
             className="hover:underline text-theme-label-link"
-            href=""
+            href="https://docs.daily.dev/docs/key-features/default-feeds#my-feed"
             target="_blank"
             rel="noopener"
           >
--- a/packages/shared/src/components/sidebar/MyFeedAlert.tsx
+++ b/packages/shared/src/components/sidebar/MyFeedAlert.tsx
@@ -27,7 +27,7 @@ export default function MyFeedAlert({
         {alertCopy[alerts.myFeed]}
         {alerts.myFeed === 'default' && (
           <a
-            className="text-theme-label-link hover:underline"
+            className="hover:underline text-theme-label-link"
             href=""
             target="_blank"
             rel="noopener"
--- a/packages/shared/src/components/sidebar/MyFeedAlert.tsx
+++ b/packages/shared/src/components/sidebar/MyFeedAlert.tsx
@@ -2,30 +2,32 @@ import React, { ReactElement } from 'react';
 import XIcon from '@dailydotdev/shared/icons/x.svg';
 import { Alerts } from '../../graphql/alerts';
 import { Button } from '../buttons/Button';
-import { AlertContainer, AlertCopy, Pointer } from './common';
+import { AlertContainer, AlertCopy } from './common';
+import { Pointer, PointerColor } from '../Pointer';
 
 const alertCopy = {
-  default: `Psst, your feed has a new name! We've already applied your content filters to it. `,
-  manual: `🎉 Your feed is ready! Click here to manage your feed's settings`,
+  migrated: `Psst, your feed has a new name! We've already applied your content filters to it. `,
+  created: `🎉 Your feed is ready! Click here to manage your feed's settings`,
 };
 
+interface MyFeedAlertProps {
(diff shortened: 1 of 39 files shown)


packages/shared/src/components/sidebar/common.tsx
```diff
@@ -90,10 +90,34 @@ export const NavHeader = classed(
...
+  'w-1.5 h-1.5 rounded-full bg-theme-status-success block',
+);
+const PointerLine = classed('i', 'w-px h-3.5 bg-theme-status-success block');
+
+export const Pointer = (): ReactElement => (
+  <PointerContainer>
+    <PointerPoint />
+    <PointerLine />
+  </PointerContainer>
+);
```
sshanzel, line 120: I think we can probably take this out and make a file for itself and allow props for `color` to allow flexibility and make changes easily.

sshanzel (changes requested):
Just had some minor comments but great work on this!

packages/shared/src/components/sidebar/MyFeedAlert.tsx
```diff
@@ -0,0 +1,47 @@
...
+};
+
+export default function MyFeedAlert({
+  alerts,
+  hideAlert,
+}: {
+  alerts: Alerts;
+  hideAlert: () => unknown;
+}): ReactElement {
+  if (alerts?.filter || (!alerts?.filter && alerts?.myFeed === null)) {
```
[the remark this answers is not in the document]
rebelchris: @sshanzel Where did you base that on?
We should not change anything in relation to the `alerts.filter`.

Please let me know if we need to have a meeting around this, as this will be very crucial.

packages/shared/src/components/sidebar/common.tsx
rebelchris: Great point!
Will work on this

packages/shared/src/components/sidebar/MyFeedAlert.tsx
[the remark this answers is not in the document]
sshanzel: Oh, cool. Seen the PR for the API now. The condition here looks fine.

rebelchris:
@sshanzel Refactored the Pointer to it's own component and refactored to reflect the API changes.

sshanzel (changes requested):
Looks great! I just think we are missing the tests.

rebelchris:
@sshanzel Thanks for that notice!
Now that we merged the feed button I could add the test as well.

So now introduced all test to evaluate the alert types based on the features set for a user.

rebelchris:
@idoshamun Perhaps you can already have a look at this one as you will be busy later

packages/shared/src/components/sidebar/MyFeedAlert.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+interface MyFeedAlertProps {
+  alerts: Alerts;
+  hideAlert: () => unknown;
+}
+export default function MyFeedAlert({
+  alerts,
+  hideAlert,
+}: MyFeedAlertProps): ReactElement {
+  if (alerts?.filter || (!alerts?.filter && alerts?.myFeed === null)) {
```
idoshamun, line 21: I'm wondering if this condition should be part of this component or if the parent component should not render it at all if that's the case

packages/shared/src/components/sidebar/MyFeedAlert.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+  return (
+    <AlertContainer>
+      <Pointer color={PointerColor.Success} />
+      <AlertCopy>
+        {alertCopy[alerts.myFeed]}
+        {alerts.myFeed === 'migrated' && (
+          <a
+            className="hover:underline text-theme-label-link"
+            href=""
```
idoshamun, line 33: Need to remember somehow to add this link 🤔

packages/shared/src/components/sidebar/Sidebar.spec.tsx
```diff
@@ -212,3 +212,37 @@ it('should show the my feed items if the user has filters', async () => {
   const section = await screen.findByText('My feed');
   expect(section).toBeInTheDocument();
 });
+
+it('should show the migrated message if the user has existing filters', async () => {
```
idoshamun, line 216: I'm missing a test to check that `updateAlerts` is called

idoshamun (changes requested):
A few comments...

packages/shared/src/components/sidebar/MyFeedAlert.tsx
sshanzel: I'll add a child ticket for these parts since right now the Sidebar has a constant variable to set `shouldShowMyFeed` always to true which also needs to be changed before running the QA.

packages/shared/src/components/sidebar/Sidebar.spec.tsx
rebelchris: That is the first test on line 118.
Felt a bit overkill to test it everytime.

What do you think?

rebelchris:
@sshanzel @idoshamun Refactored the render evaluation for the my feed alert.
Placed comments for the other points.

sshanzel (approved):
Overall it is super solid! We are now just missing the test that Ido has mentioned.

The test on line 118 runs when a user that has pre-existing filters opens the feed filters (executes mutation). We should also test if the mutation is called once the user has clicked the close button for the alert message since we have now two ways of hiding it for the mentioned user scenario.

Great work btw! 🎉

packages/shared/src/components/sidebar/Sidebar.tsx
```diff
@@ -240,6 +245,9 @@ export default function Sidebar({
     return <></>;
   }
 
+  const shouldHideMyFeedAlert =
+    alerts?.filter || (!alerts?.filter && alerts?.myFeed === null);
```
idoshamun, line 249: I suggest this. Otherwise, if alerts is undefined this condition will be true.
What we actually want to check is that it is not set to any actual value.

```suggestion
    alerts?.filter || (!alerts?.filter && !!alerts?.myFeed);
```

idoshamun (changes requested):
I found a bug I think, and also as Lee mentioned the test is missing

idoshamun (approved):
Sail away 🚢

PR #788, opened 2022-01-03
fix: fixed the default values for existing feature flags

rebelchris:
We recently introduced the default value for the feature flagged variables.
However we were not leveraging this in all places.

This PR solves that by switching the or cases to the default values


sshanzel (approved):
Super neat!

idoshamun (approved):

PR #789, opened 2022-01-03
fix: :bug: Removing the codeowners file

rebelchris:
After a discussion with @idoshamun we've decided to remove the codeowners file.
This will give myself and @sshanzel more control as being the active code reviewers for the apps repo.


idoshamun (approved):
OMG! I can't believe it 🥲

PR #790, opened 2022-01-04
feat: added the generic wrapper for the my feed ab test

rebelchris:
Introducing a new flagsmith value to determine if we should show the my feed in general or not.

DD-440 #done


idoshamun (approved):
Awesome!
I think this makes our new component compatible with master so we can think about merging it. Proceed as you seem fit.

sshanzel (approved):
Looks great!

PR #791, opened 2022-01-05
refactor: local storage utilization for feed filters

sshanzel:
DD-430 #done
[DD-442](https://dailydotdev.atlassian.net/jira/software/projects/DD/boards/1?selectedIssue=DD-442) #done - pre-existing issue (unblock a tag)
[DD-445](https://dailydotdev.atlassian.net/jira/software/projects/DD/boards/1?selectedIssue=DD-445) #done - pre-existing issue (reporting and hiding post)
[DD-446](https://dailydotdev.atlassian.net/jira/software/projects/DD/boards/1?selectedIssue=DD-446) #done - pre-existing issue (blocking a source and a tag)

This includes quite a few fixes since I was on it already and have thought of fixing it along the way.

The issues were:
- Unblock a tag on context menu after searching a tag
- Reporting a post as an anonymous user
- Hiding a post as an  anonymous user
- Blocking a tag as an  anonymous user
- Blocking a source as an anonymous user

I have also worked on some refactoring to apply proper types. One of the key ones were receiving a callback as a prop, I thought using the prefix `set` (i.e. `setValue`) is most commonly used on state-related declarations, so I changed the props to follow our common prefix in the codebase using `on` (i.e. `onValueChanged`).

At `FeedFilters.tsx` it can be noticed that I removed some mechanism for animation since it is being handled outside by our custom hook already, it shouldn't be necessary on the inside anymore.

Most of the core changes were part of the `useMutateFilters` where it manages the cache for `react-query` and instead of reworking the whole `FeedFilters` (and all the components underneath) to accept a dynamic prop where the outside component to decide whether we use data from local storage or the query.

We now just toggle whether we should utilize local storage or not so the core process flow won't be affected.

### DIFF
--- a/packages/shared/src/components/PostOptionsMenu.tsx
+++ b/packages/shared/src/components/PostOptionsMenu.tsx
@@ -3,7 +3,7 @@ import { Item } from '@dailydotdev/react-contexify';
 import dynamic from 'next/dynamic';
 import useFeedSettings from '../hooks/useFeedSettings';
 import useReportPost from '../hooks/useReportPost';
-import { Post } from '../graphql/posts';
+import { Post, ReportReason } from '../graphql/posts';
 import TrashIcon from '../../icons/trash.svg';
 import HammerIcon from '../../icons/hammer.svg';
 import EyeIcon from '../../icons/eye.svg';
@@ -37,6 +37,14 @@ const MenuIcon = ({ Icon }) => {
   return <Icon className="mr-2 text-2xl" />;
 };
 
+type ReportPostAsync = (
+  postIndex: number,
+  post: Post,
+  reason: ReportReason,
+  comment: string,
+  blockSource: boolean,
+) => Promise<unknown>;
+
 export default function PostOptionsMenu({
   postIndex,
   post,
@@ -64,18 +72,23 @@ export default function PostOptionsMenu({
     onRemovePost?.(_postIndex);
   };
 
-  const onReportPost = async (
+  const onReportPost: ReportPostAsync = async (
     reportPostIndex,
     reportedPost,
     reason,
     comment,
     blockSource,
   ): Promise<void> => {
-    reportPost({
+    const shouldProceed = await reportPost({
       id: reportedPost?.id,
       reason,
       comment,
     });
+
+    if (!shouldProceed) {
+      return;
+    }
+
     trackEvent(
       postAnalyticsEvent('report post', reportedPost, {
         extra: { origin: 'post context menu' },
@@ -91,26 +104,49 @@ export default function PostOptionsMenu({
 
   const onBlockSource = async (): Promise<void> => {
     setAvoidRefresh(true);
-    await onUnfollowSource({ source: post?.source });
+    const shouldProceed = await onUnfollowSource({
+      source: post?.source,
+      requireLogin: true,
+    });
+    if (!shouldProceed) {
+      setAvoidRefresh(false);
+      return;
+    }
     showMessageAndRemovePost(`🚫 ${post?.source?.name} blocked`, postIndex);
     setAvoidRefresh(false);
   };
 
   const onBlockTag = async (tag: string): Promise<void> => {
     setAvoidRefresh(true);
-    await onBlockTags({ tags: [tag] });
+    const shouldProceed = await onBlockTags({
+      tags: [tag],
+      requireLogin: true,
+    });
+
+    if (!shouldProceed) {
+      setAvoidRefresh(false);
+      return;
+    }
+
     await showMessageAndRemovePost(`⛔️ #${tag} blocked`, postIndex);
     setAvoidRefresh(false);
   };
 
   const onHidePost = async (): Promise<void> => {
-    const promise = hidePost(post.id);
+    const shouldProceed = await hidePost(post.id);
+
+    if (!shouldProceed) {
+      return;
+    }
+
     trackEvent(
       postAnalyticsEvent('hide post', post, {
         extra: { origin: 'post context menu' },
       }),
     );
-    await Promise.all([promise, onRemovePost?.(postIndex)]);
+
+    await onRemovePost?.(postIndex);
+
     if (!postIndex) {
       onMessage(
         '🙈 This article won’t show up on your feed anymore',
--- a/packages/shared/src/components/PostOptionsMenu.tsx
+++ b/packages/shared/src/components/PostOptionsMenu.tsx
@@ -79,13 +79,13 @@ export default function PostOptionsMenu({
     comment,
     blockSource,
   ): Promise<void> => {
-    const shouldProceed = await reportPost({
+    const { successful } = await reportPost({
       id: reportedPost?.id,
       reason,
       comment,
     });
 
-    if (!shouldProceed) {
+    if (!successful) {
       return;
     }
 
@@ -104,11 +104,11 @@ export default function PostOptionsMenu({
 
   const onBlockSource = async (): Promise<void> => {
     setAvoidRefresh(true);
-    const shouldProceed = await onUnfollowSource({
+    const { successful } = await onUnfollowSource({
       source: post?.source,
       requireLogin: true,
     });
-    if (!shouldProceed) {
+    if (!successful) {
       setAvoidRefresh(false);
       return;
     }
@@ -118,12 +118,12 @@ export default function PostOptionsMenu({
 
   const onBlockTag = async (tag: string): Promise<void> => {
     setAvoidRefresh(true);
-    const shouldProceed = await onBlockTags({
+    const { successful } = await onBlockTags({
       tags: [tag],
       requireLogin: true,
     });
 
-    if (!shouldProceed) {
+    if (!successful) {
       setAvoidRefresh(false);
       return;
     }
@@ -133,9 +133,9 @@ export default function PostOptionsMenu({
   };
 
   const onHidePost = async (): Promise<void> => {
-    const shouldProceed = await hidePost(post.id);
+    const { successful } = await hidePost(post.id);
 
-    if (!shouldProceed) {
+    if (!successful) {
       return;
     }
 
--- a/packages/shared/src/components/filters/FeedFilters.tsx
+++ b/packages/shared/src/components/filters/FeedFilters.tsx
@@ -12,7 +12,9 @@ import XIcon from '../../../icons/x.svg';
 import { menuItemClassNames } from '../multiLevelMenu/MultiLevelMenuMaster';
 import useFeedSettings from '../../hooks/useFeedSettings';
 import AlertContext from '../../contexts/AlertContext';
-import { useDynamicLoadedAnimation } from '../../hooks/useDynamicLoadAnimated';
+import { Source } from '../../graphql/sources';
+import UnblockModal from '../modals/UnblockModal';
+import AuthContext from '../../contexts/AuthContext';
 
 const asideWidth = sizeN(89);
 
@@ -21,20 +23,28 @@ interface FeedFiltersProps {
   onBack?: () => void;
 }
 
+interface UnblockItem {
+  tag?: string;
+  source?: Source;
+  action?: () => unknown;
+}
+
 export default function FeedFilters({
   isOpen,
   onBack,
 }: FeedFiltersProps): ReactElement {
+  const [unblockItem, setUnblockItem] = useState<UnblockItem>();
+  const { user } = useContext(AuthContext);
   const { alerts, updateAlerts } = useContext(AlertContext);
   const { hasAnyFilter } = useFeedSettings();
 
   useEffect(() => {
     if (isOpen) {
-      if (alerts?.filter && hasAnyFilter) {
+      if (alerts?.filter && hasAnyFilter && user) {
         updateAlerts({ filter: false });
       }
     }
-  }, [isOpen, alerts, hasAnyFilter]);
+  }, [isOpen, alerts, user, hasAnyFilter]);
 
   return (
     <aside
@@ -55,7 +65,15 @@ export default function FeedFilters({
           <XIcon className="text-2xl -rotate-90 text-theme-label-tertiary" />
         </button>
       </div>
-      <FilterMenu />
+      <FilterMenu onUnblockItem={setUnblockItem} />
+      {unblockItem && (
+        <UnblockModal
+          item={unblockItem}
+          isOpen={!!unblockItem}
+          onConfirm={unblockItem.action}
+          onRequestClose={() => setUnblockItem(null)}
+        />
+      )}
     </aside>
   );
 }
--- a/packages/shared/src/components/filters/FeedFilters.tsx
+++ b/packages/shared/src/components/filters/FeedFilters.tsx
@@ -16,6 +16,8 @@ import UnblockModal from '../modals/UnblockModal';
 import AuthContext from '../../contexts/AuthContext';
 import { Button } from '../buttons/Button';
 import { AllTagCategoriesData } from '../../graphql/feedSettings';
+import { Features, isFeaturedEnabled } from '../../lib/featureManagement';
+import FeaturesContext from '../../contexts/FeaturesContext';
 
 const asideWidth = sizeN(89);
 
@@ -39,6 +41,8 @@ export default function FeedFilters({
   const { user } = useContext(AuthContext);
   const { alerts, updateAlerts } = useContext(AlertContext);
   const { hasAnyFilter } = useFeedSettings();
+  const { flags } = useContext(FeaturesContext);
+  const shouldShowMyFeed = isFeaturedEnabled(Features.MyFeedOn, flags);
 
   useEffect(() => {
     if (isOpen) {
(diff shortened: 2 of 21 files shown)


packages/shared/src/components/filters/TagsFilter.spec.tsx
```diff
@@ -180,11 +197,14 @@ it('should follow a tag on click and remove filter alert if enabled', async () =
   const buttonDiv = await screen.findByTestId('tagCategoryTags');
   expect(buttonDiv).toBeVisible();
 
-  // eslint-disable-next-line testing-library/prefer-screen-queries
-  const button = await screen.findByText('#webdev');
-  button.click();
+  await act(async () => {
```
rebelchris, line 200: Are you sure the act is needed here?

packages/shared/src/hooks/useFeedSettings.ts
```diff
@@ -24,25 +33,64 @@ export type FeedSettingsReturnType = {
...
+  const { tagsCategories, feedSettings, advancedSettings } = feedQuery;
+
   useEffect(() => {
+    if (!user?.id) {
+      return;
+    }
+
     if (avoidRefresh) {
       return;
     }
```
rebelchris, line 96: I'm guessing we can combine these two right?

packages/shared/src/hooks/useFeedSettings.ts
```diff
@@ -57,6 +105,29 @@ export default function useFeedSettings(): FeedSettingsReturnType {
     }, 100);
   }, [tagsCategories, feedSettings, avoidRefresh]);
 
+  useEffect(() => {
+    const isEmpty = isObjectEmpty(feedSettings);
+    const isFeedQueryEmpty = Object.keys(feedQuery).length === 0;
```
rebelchris, line 110: Can you not use `isObjectEmpty` for the isFeedQueryEmpty as well?

rebelchris (changes requested):
Nice work Lee!
And thank you for fixing all the return types.

Indeed see we missed some cases for actions going through whilst they shouldn't.

Why did we choose the local storage, would it not have been easier to leverage the new `usePersistentContext`?

packages/shared/src/components/PostOptionsMenu.tsx
```diff
@@ -64,18 +72,23 @@ export default function PostOptionsMenu({
...
-    reportPost({
+    const shouldProceed = await reportPost({
       id: reportedPost?.id,
       reason,
       comment,
     });
+
+    if (!shouldProceed) {
+      return;
+    }
```
rebelchris, line 90: I'm not 100% sure if we should return and not fire the custom analytics?
What would cause the item to fail in this case?

rebelchris (changes requested):

packages/shared/src/components/filters/TagsFilter.spec.tsx
sshanzel: This was just from the existing test, so I think it is safe to say, yes.

packages/shared/src/hooks/useFeedSettings.ts
sshanzel: Ah, totally makes sense

packages/shared/src/components/PostOptionsMenu.tsx
sshanzel: Was thinking about that as well but thought we require login (which is part of the `shouldProceed`) to avoid abuse, people can start just reporting and we wouldn't know from whom.

packages/shared/src/hooks/useFeedSettings.ts
sshanzel: That makes sense, it will make the condition future proof even though we declare a default value of object literal in `feedQuery`. Let me refactor it.

packages/shared/src/components/filters/TagsFilter.spec.tsx
sshanzel: You are right! Seems like we don't need it. Have removed it, including on the existing test.

sshanzel:
Thank you for the awesome review Chris! The reason is, local storage is very straightforward, and we use the `useFeedSettings` on multiple areas, each of those instances (all async) will have to check the storage especially we utilize a variable to check if it has been checked already, it might produce unexpected behavior. So instead of doing a round trip to the store every time a component rendered has `useFeedSettings` it won't have to check if unnecessary.

sshanzel:
This should now be ready for a review, feel free once you only have the time. Thank you!

@rebelchris this time the `act` became a necessity as we will have to wait for a re-render.

cc @idoshamun

packages/shared/src/components/filters/FeedFilters.tsx
```diff
@@ -20,60 +26,78 @@ interface FeedFiltersProps {
...
-    timeoutRef.current = window.setTimeout(() => setHidden(true), 300);
-    // eslint-disable-next-line consistent-return
-    return () => timeoutRef.current && clearTimeout(timeoutRef.current);
-  }, [isOpen, alerts, hasAnyFilter]);
+  const onCreate = () => {
+    // apply analytics
+    const key = getFeedSettingsQueryKey(user);
+    const { feedSettings } = client.getQueryData(key) as AllTagCategoriesData;
+    updateLocalFeedSettings(feedSettings);
+    window.location.replace(`${process.env.NEXT_PUBLIC_WEBAPP_URL}register`);
```
idoshamun, line 60: We need to open the login modal. We can't redirect to registration directly.

packages/shared/src/components/filters/FeedFilters.tsx
```diff
@@ -20,60 +26,78 @@ interface FeedFiltersProps {
...
         className={classNames(
           menuItemClassNames,
-          'border-b border-theme-divider-tertiary',
+          'border-b border-theme-divider-tertiary flex-row justify-between',
         )}
       >
         <button onClick={onBack} type="button">
           <XIcon className="text-2xl -rotate-90 text-theme-label-tertiary" />
         </button>
+        {shouldShowMyFeed && (
```
idoshamun, line 81: This should be shown only for anonymous users

packages/shared/src/components/filters/FeedFilters.tsx
```diff
@@ -20,60 +26,78 @@ interface FeedFiltersProps {
...
-      <FilterMenu />
+      <FilterMenu onUnblockItem={setUnblockItem} />
+      {unblockItem && (
+        <UnblockModal
+          item={unblockItem}
+          isOpen={!!unblockItem}
+          onConfirm={unblockItem.action}
+          onRequestClose={() => setUnblockItem(null)}
+        />
+      )}
```
idoshamun, line 100: Why do you introduce this change?

packages/shared/src/hooks/useFeedSettings.ts
```diff
@@ -57,6 +102,32 @@ export default function useFeedSettings(): FeedSettingsReturnType {
...
+      feedSettings: { ...localFeedSettings },
+    }));
+  }, [feedQuery, feedSettings, loadedUserFromCache, user]);
+
+  useEffect(() => {
+    if (!hasClearedCache && loadedUserFromCache && user) {
+      hasClearedCache = true;
+      storage.removeItem(LOCAL_FEED_SETTINGS_KEY);
+    }
+  }, [loadedUserFromCache, user]);
```
idoshamun, line 129: What's that part?

packages/shared/src/components/filters/TagsFilter.spec.tsx
```diff
@@ -63,33 +67,44 @@ const createAllTagCategoriesMock = (
 
 let client: QueryClient;
 
+const defaultAlerts: Alerts = { filter: false };
+const defaultFlags: IFlags = {};
+const myFeedOnEnabledFlag = { my_feed_on: { enabled: true, value: 'true' } };
```
rebelchris, line 72: It's best to test this without the value, as we only check it from the enabled step.
This way we make sure there is no legacy string check anymore.

packages/shared/src/components/filters/TagsFilter.tsx
```diff
@@ -43,13 +60,16 @@ export default function TagsFilter({
...
   }, [feedSettings]);
 
-  const tagUnblockAction = ({ tags }) => {
-    setUnblockItem({
-      tag: tags,
-      action: () => onUnblockTags({ tags }),
+  const tagUnblockAction = ({ tags: [tag] }: TagActionArguments) => {
+    if (shouldShowMyFeed && !user) {
+      return onUnblockTags({ tags: [tag] });
+    }
```
rebelchris, line 66: Why is this a different action based on the shouldShowMyFeed?

packages/shared/src/hooks/useMutateFilters.ts
```diff
@@ -330,28 +369,22 @@ export default function useMutateFilters(user?: LoggedUser): ReturnType {
...
-            return newData;
-          },
-          user,
-        ),
+      onMutate: onUnfollowSource,
       onError: (err, _, rollback) => rollback(),
     },
   );
 
   return {
```
idoshamun, line 377: This should go into a `useMemo` otherwise it may cause tons of rerenders

packages/shared/src/hooks/useTagAndSource.ts
```diff
@@ -86,51 +106,58 @@ export default function useTagAndSource({
...
       event_name: 'unfollow',
       target_type: 'source',
       target_id: source?.id,
       extra: JSON.stringify({ origin, post_id: postId }),
     });
     await unfollowSource({ source });
+    return true;
   };
 
   return {
```
idoshamun, line 163: This should be in `useMemo`

rebelchris (changes requested):
Besides these comments, two general questions:
1. Why do we need to change the unblock action based on the shouldshowMyFeed?
2. As Ido will mention we need to show the showLogin instead of the register, and handle the tag setting on the register callback. (Remember people could login at any time and we still need to migrate their local feed settings)

packages/shared/src/hooks/useTagAndSource.ts
```diff
@@ -1,36 +1,48 @@
...
-  onUnfollowSource: ({ source }: SourceActionArguments) => Promise<unknown>;
-} {
+}: UseTagAndSourceProps): UseTagAndSource {
+  const { flags } = useContext(FeaturesContext);
+  const shouldShowMyFeed = isFeaturedEnabled(Features.MyFeedOn, flags);
   const { alerts, updateAlerts } = useContext(AlertContext);
   const { user, showLogin } = useContext(AuthContext);
   const { trackEvent } = useContext(AnalyticsContext);
+  const shouldShowLogin = (requireLogin?: boolean) =>
+    (!shouldShowMyFeed && !user) || (requireLogin && !user);
```
idoshamun, line 45: I think this is a wrong condition.
Because we require login for following tags but if you are part of the my feed experiment you shouldn't be asked to login. But the condition will still yield true.

idoshamun (changes requested):
I find this PR quite hard to review. I think the mistake was to blend together many fixes just because they involve the same components.
Reviewing the new flow would have been a hard task by itself, nonetheless with many more changes.
I pointed out in my review some critical issues that I found but I also request to take another look and try to refactor to make the code more readable. And also make sure this works locally in different scenarios.
This is the most complex part of this feature and we need to be very careful.
I'm in the Showcase room if you need me.

packages/shared/src/hooks/useTagAndSource.ts
sshanzel: When blocking sources and tags through the post menu options, I was thinking it should require login since they might think it should've been blocked but didn't happen.

So being part of the experiment can only happen in the actual FeedFilters in the sidebar.

packages/shared/src/components/filters/TagsFilter.tsx
sshanzel: Since blocking as an anonymous user doesn't take effect on the actual feed, I think it is not necessary to ask anymore, especially the question says the source will now be seen in the feed which is still seen even after blocking.

packages/shared/src/hooks/useFeedSettings.ts
sshanzel: This is to set the `feedSettings` initial values for anonymous users and part of the experiment since the property is undefined not `loggedIn`

packages/shared/src/components/filters/FeedFilters.tsx
sshanzel: Oh, I might have missed this on my initial plan on how to manage the data from the feed filters.

packages/shared/src/components/filters/FeedFilters.tsx
sshanzel: Oh, my mistake. Let me rectify this.

packages/shared/src/components/filters/TagsFilter.spec.tsx
sshanzel: I think we should test the actual function that is utilized by these features instead of each of its implementation, what do you think?

packages/shared/src/hooks/useTagAndSource.ts
sshanzel: Let me refactor this for better readability.

sshanzel:
First of all, thank you both for the awesome review! Apologies for not mentioning the affected files for the fixes, it should be the `useTagAndSources` where the functions it returns now give a boolean whether the action was successful or not.

@rebelchris for the item 1. - kindly check my reply. For item number 2. - we should handle syncing the actual filters if they clicked login at the register still since at the end of the day they will be redirected to the registration page after hitting a provider. With that, it should be covered by the PR dedicated to it.

packages/shared/src/components/filters/TagsFilter.spec.tsx
rebelchris: Sure we can also do that, just meaning to say if you simply remove the `value: 'true'` we make sure to not introduce and fake false tests

packages/shared/src/components/filters/TagsFilter.tsx
rebelchris: Ok, but what does it have to do with the My Feed?

sshanzel: When unblocking a tag they have blocked.

packages/shared/src/components/filters/TagsFilter.spec.tsx
sshanzel: Makes total sense. Should be removed by now.

sshanzel:
@idoshamun I have made the necessary refactor, though I am thinking, would it make more sense if the functions affected by the fixes return an object with `{ success: boolean }`, so it is much easier to understand from the consumer's end that the boolean value returned meant it was successful.

This should now be ready for another review 🚢 

@rebelchris

sshanzel:
Preview running locally:

https://user-images.githubusercontent.com/13744167/148389777-7d215dc6-6465-4e9d-a5cd-1d162b71a8e8.mov

packages/shared/src/hooks/useFeedSettings.ts
```diff
@@ -57,6 +102,38 @@ export default function useFeedSettings(): FeedSettingsReturnType {
...
```
idoshamun, line 135: Reading your comment I'm not sure we need this snippet.
The state we keep locally is used just as a way to sync the settings back once the user is registered.
If they never finish registration we should not show the preselected values in there every time they load the app.

rebelchris (approved):
Impressive work Lee!

This was not an easy one.
And now that I've seen the register change, it was exactly what I meant to say, so perfectly aligned 🎉

Go for it, from my side.

idoshamun (changes requested):
Way way better!

packages/shared/src/hooks/useFeedSettings.ts
sshanzel: Totally makes sense, let me fix this up quickly!

sshanzel:
Woop! Woop! Once again, thank you both for the reviews! 🚢

packages/shared/src/hooks/useFeedSettings.ts
```diff
@@ -76,11 +83,20 @@ export default function useFeedSettings(): FeedSettingsReturnType {
...
+        FEED_SETTINGS_QUERY,
+        { loggedIn: !!user?.id },
+      );
+
+      const feedSettings = getQueryFeedSettings(
+        req,
+        feedQuery,
+        !!user,
+        shouldShowMyFeed,
+      );
```
idoshamun, line 97: I think you should run the request only when it is needed.
Right now you dispatch the request and then check if you need it inside the `getQueryFeedSettings`

idoshamun (changes requested):

packages/shared/src/hooks/useFeedSettings.ts
sshanzel: ahh totally makes sense, less processing.

sshanzel: Though we would really still need this request as this contains the list of tags/keywords.

We will now only decide whether we will utilize remote feed settings.

idoshamun: Oh. Sorry I missed it. So it makes sense to keep it this way

idoshamun (approved):
Woohoo!

PR #792, opened 2022-01-05
fix: re-add the feed filters if the user has the ab test set to off

rebelchris:
We accidentally removed the feed filter option for people who have the AB test set to false.
This PR reintroduces it.

By doing so we had to re-arrange the flag loading a bit.

DD-447 #done

### DIFF
--- a/packages/shared/src/components/sidebar/Sidebar.tsx
+++ b/packages/shared/src/components/sidebar/Sidebar.tsx
@@ -146,7 +146,7 @@ export default function Sidebar({
     useContext(SettingsContext);
   const [showSettings, setShowSettings] = useState(false);
   const { flags } = useContext(FeaturesContext);
-  const shouldShowMyFeed = getFeatureValue(Features.MyFeedOn, flags, true);
+  const shouldShowMyFeed = getFeatureValue(Features.MyFeedOn, flags, false);
 
   useHideMobileSidebar({
     state: openMobileSidebar,
@@ -187,7 +187,7 @@ export default function Sidebar({
       hideOnMobile: true,
     },
   ];
-  if (shouldShowMyFeed !== 'true') {
+  if (shouldShowMyFeed) {
     discoverMenuItems.unshift({
       icon: <ListIcon Icon={FilterIcon} />,
       alert: alerts.filter && (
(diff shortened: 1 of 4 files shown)


sshanzel (approved):
Let's merge it! 💪

rebelchris:
@idoshamun Would you mind taking a second look as well 👏

packages/shared/src/components/sidebar/Sidebar.tsx
```diff
@@ -142,6 +145,8 @@ export default function Sidebar({
   const { sidebarExpanded, toggleSidebarExpanded, loadedSettings } =
     useContext(SettingsContext);
   const [showSettings, setShowSettings] = useState(false);
+  const { flags } = useContext(FeaturesContext);
+  const shouldShowMyFeed = getFeatureValue(Features.MyFeedOn, flags, 'false');
```
idoshamun, line 149: This is quite confusing because usually if you prefix a variable name with `should` you expect it to be boolean whereas in this case, it's a string.
Also from a Flagsmith point of view, we can utilize the `enabled` field instead of the actual value. So for the person who manages the deployment (aka Nimrod), it will be more straightforward and less error-prone.
What do you think?

idoshamun (changes requested):

packages/shared/src/components/sidebar/Sidebar.tsx
rebelchris: Agreed, let me rework the `getFeatureValue` to handle booleans :)

rebelchris:
@idoshamun Refactored the `getFeatureValue` function to handle booleans.
The idea is the default value will be false for these boolean checks.

Then once it's enabled in flagsmith we return true, else we return the default value (false).
Looking forward to hear what you think.

idoshamun (changes requested):
I'd imagined a new function that returns the `enabled` state of the flag rather than value.
This should make it easier to manage this kind of variants on the Flagsmith side.
Instead of using `getFeatureValue`, we can use `isFeaturedEnabled` for example in this case.
What do you think?

rebelchris:
@idoshamun Ah yes, makes total sense!
Refactored away and now we have this new `isFeaturedEnabled` method as well.

idoshamun (approved):
Perfect!

PR #793, opened 2022-01-05
refactor: ab signup popup

ayadotdev:
Added AB testing for login modal buttons and description copy. 
Removed LoginModalMode because it is no longer used. 

DD-412 #done

### DIFF
--- a/packages/shared/src/components/LoginButtons.tsx
+++ b/packages/shared/src/components/LoginButtons.tsx
@@ -1,25 +1,23 @@
-import React, { ReactElement, useContext } from 'react';
+import React, { HTMLAttributes, ReactElement, useContext } from 'react';
 import { useRouter } from 'next/router';
 import GitHubIcon from '../../icons/github.svg';
 import { privacyPolicy, termsOfService } from '../lib/constants';
 import { LegalNotice } from './utilities';
 import { Button } from './buttons/Button';
 import AuthContext from '../contexts/AuthContext';
-import FeaturesContext from '../contexts/FeaturesContext';
-import { Features, getFeatureValue } from '../lib/featureManagement';
 import { apiUrl } from '../lib/config';
 import AnalyticsContext from '../contexts/AnalyticsContext';
 
-export default function LoginButtons(): ReactElement {
+export interface LoginButtonsProps extends HTMLAttributes<HTMLDivElement> {
+  buttonCopyPrefix?: string;
+}
+
+export default function LoginButtons({
+  buttonCopyPrefix,
+}: LoginButtonsProps): ReactElement {
   const router = useRouter();
   const { getRedirectUri } = useContext(AuthContext);
   const { trackEvent } = useContext(AnalyticsContext);
-  const { flags } = useContext(FeaturesContext);
-  const buttonCopyPrefix = getFeatureValue(
-    Features.LoginModalButtonCopyPrefix,
-    flags,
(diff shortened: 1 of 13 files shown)


sshanzel (changes requested):
Looks solid overall! 🚢  I think the only thing missing would be the additional tests when the features flag is enabled to see if it actually renders what is expected. Great work on this! Let me know if you would need any help creating a test utilizing the `FeaturesContext` (where the flags are handled).

packages/shared/src/components/modals/LoginModal.tsx
```diff
@@ -32,9 +37,7 @@ export default function LoginModal({
       <ModalCloseButton onClick={onRequestClose} />
       <DailyDevLogo />
       <div className="mt-6 mb-8 text-center text-theme-label-secondary typo-callout">
-        {mode === LoginModalMode.ContentQuality
-          ? `Our community cares about content quality. We require social authentication to prevent abuse.`
-          : `Unlock useful features by signing in. A bunch of cool stuff like content filters and bookmarks are waiting just for you.`}
+        {loginModalDescriptionCopy}
       </div>
       <LoginButtons />
```
rebelchris, line 42: Seeing as we are already loading the flags and context here I think it would be nicer to determine the LoginButtons copy here and pass it as a child.

That way we don't have to load the context again in the child component.
What do you think?

rebelchris (changes requested):
Nice work on this setup!

Just the inline comment from my side

rebelchris:
@sshanzel We should be careful with introducing test cases for the AB filters as they will change overtime.
Perhaps we should consider adding a global test to determine if the `getFeatureValue` is working rather than for the specific functions?

What do you think?

idoshamun:
I agree with @rebelchris here. A/B experiments can be temporary, I wouldn't introduce a lot of hassle into implementing them unless they're quite complex. If you feel something here seriously needs to be tested, we can talk it over. 😄

packages/shared/src/components/modals/LoginModal.tsx
ayadotdev: @rebelchris That's a good point, I'll try it.

sshanzel:
That makes total sense. With that, all looks good to me now! 🚢

sshanzel (approved):
LGTM! 🚀

packages/shared/src/components/LoginButtons.tsx
```diff
@@ -8,7 +8,13 @@ import AuthContext from '../contexts/AuthContext';
 import { apiUrl } from '../lib/config';
 import AnalyticsContext from '../contexts/AnalyticsContext';
 
-export default function LoginButtons(): ReactElement {
+export interface LoginButtonsProps extends HTMLAttributes<HTMLDivElement> {
```
rebelchris, line 11: Some small notes, not breaking, but the `export` is not needed here as we are not using it outside this file.

Also the `extends HTMLAttributes<HTMLDivElement>` doesn't add anything here as we are not passing more props.

If you are remove those we should be ready to go 🎉

rebelchris (changes requested):

rebelchris (approved):
Perfect, just roping @idoshamun in for his final say 👀

idoshamun (approved):
Way to go! Looks fabulous. Let's ship it 🚢

PR #795, opened 2022-01-05
feat: my feed page

sshanzel:
DD-428 #done
DD-430 #done

A dedicated page for "My Feed" page/tab under `my-feed` route.

This PR also refactored the usage of popular page/tab to utilize anonymous feed query.

### DIFF
--- a/packages/extension/src/newtab/MainFeedPage.tsx
+++ b/packages/extension/src/newtab/MainFeedPage.tsx
@@ -11,7 +11,7 @@ import { HeaderButton } from '@dailydotdev/shared/src/components/buttons/common'
 import FeaturesContext from '@dailydotdev/shared/src/contexts/FeaturesContext';
 import {
   Features,
-  getFeatureValue,
+  isFeaturedEnabled,
 } from '@dailydotdev/shared/src/lib/featureManagement';
 import MostVisitedSites from './MostVisitedSites';
 
@@ -40,8 +40,7 @@ export default function MainFeedPage({
   const [showDnd, setShowDnd] = useState(false);
   const [defaultFeed] = usePersistentContext('defaultFeed', 'popular');
   const { flags } = useContext(FeaturesContext);
-  const myFeed = getFeatureValue(Features.MyFeedOn, flags, 'false');
-  const shouldShowMyFeed = myFeed === 'true';
+  const shouldShowMyFeed = isFeaturedEnabled(Features.MyFeedOn, flags);
   const enableSearch = () => {
     setIsSearchOn(true);
     setSearchQuery(null);
--- a/packages/extension/src/newtab/MainFeedPage.tsx
+++ b/packages/extension/src/newtab/MainFeedPage.tsx
@@ -1,18 +1,15 @@
 import React, { ReactElement, useContext, useMemo, useState } from 'react';
 import usePersistentContext from '@dailydotdev/shared/src/hooks/usePersistentContext';
 import MainLayout from '@dailydotdev/shared/src/components/MainLayout';
-import MainFeedLayout from '@dailydotdev/shared/src/components/MainFeedLayout';
+import MainFeedLayout, {
+  getShouldRedirect,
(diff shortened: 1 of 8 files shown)


packages/webapp/pages/my-feed.tsx
```diff
@@ -0,0 +1,28 @@
...
+import { NextSeoProps } from 'next-seo/lib/types';
+import { NextSeo } from 'next-seo';
+import {
+  getMainFeedLayout,
+  mainFeedLayoutProps,
+} from '../components/layouts/MainFeedPage';
+import { defaultOpenGraph, defaultSeo } from '../next-seo';
+
+const seo: NextSeoProps = {
+  title: 'My Feed posts on daily.dev',
```
rebelchris, line 11: Maybe double check this one with Nimrod to be safe

rebelchris (approved):
Nice work!
Nothing to add really.

Was thinking for a second do we really care if they find the `my-feed` url, but I guess it makes sense. 

(My reason for not doing it we could drop resource checks)

packages/extension/src/newtab/MainFeedPage.tsx
```diff
@@ -45,7 +52,10 @@ export default function MainFeedPage({
       setIsSearchOn(false);
     }
     setFeedName(tab);
-    onPageChanged(`/${tab}`);
+    const isMyFeed = tab === '/my-feed';
+    const shouldRedirect =
+      (isMyFeed && !shouldShowMyFeed) || (isMyFeed && !user);
```
idoshamun, line 57: If the user is not logged in we should redirect to the home page. If the user is not part of the experiment you can keep them on this page, it's probably because someone sent them a link

packages/shared/src/components/MainFeedLayout.tsx
```diff
@@ -44,10 +44,14 @@ type FeedQueryProps = {
...
 
 const propsByFeed: Record<string, FeedQueryProps> = {
-  popular: {
+  'my-feed': {
     query: ANONYMOUS_FEED_QUERY,
     queryIfLogged: FEED_QUERY,
   },
+  popular: {
+    query: ANONYMOUS_FEED_QUERY,
+    queryIfLogged: ANONYMOUS_FEED_QUERY,
```
idoshamun, line 53: `queryIfLogged` should be dependent on the A/B. If the A/B is off then it should be `FEED_QUERY`

packages/webapp/components/layouts/MainFeedPage.tsx
```diff
@@ -25,11 +37,17 @@ export default function MainFeedPage({
...
+  const { flags } = useContext(FeaturesContext);
+  const shouldShowMyFeed = isFeaturedEnabled(Features.MyFeedOn, flags);
   const [feedName, setFeedName] = useState(getFeedName(router?.pathname));
   const [isSearchOn, setIsSearchOn] = useState(router?.pathname === '/search');
 
   useEffect(() => {
-    if (router?.pathname === '/search') {
+    const isMyFeed = router?.pathname === '/my-feed';
+    if ((isMyFeed && !shouldShowMyFeed) || (isMyFeed && !user)) {
+      router.replace('popular');
```
idoshamun, line 49: Same comments as above. I would even turn this condition into a function that can be utilized in both places.
Also, make sure to redirect to the home page and not to popular.

idoshamun (changes requested):
Please check my comments 🙏

packages/webapp/components/layouts/MainFeedPage.tsx
sshanzel: I agree, let me turn this into a function to have an easier decision with the conditions.

sshanzel:
Should now be ready for another round of review. 

@rebelchris I have now removed all the parts relating to SEO, thank you for mentioning it!

cc @idoshamun

rebelchris (approved):
Nice work! 🎉🚀

idoshamun (approved):
Way to go!!

PR #796, opened 2022-01-05
feat: Added a features class to maintain default states in one place

rebelchris:
We decided to convert the existing Features Enum into a object containing the Flagsmith ID and the default value.
This way we remove any friction and make this file the single source of truth around flags.

Since a enum can't contain arrays of objects and a const object would introduce some changeable options I've decided to go for this class approach.
This gives us the abilities of an enum, while maintaining the dynamics of objects.

DD-448 #done

### DIFF
--- a/packages/shared/src/lib/featureManagement.ts
+++ b/packages/shared/src/lib/featureManagement.ts
@@ -1,21 +1,58 @@
 import { IFlags } from 'flagsmith';
 
-export enum Features {
-  SignupButtonCopy = 'signup_button_copy',
-  DevcardLimit = 'feat_limit_dev_card',
-  FeedVersion = 'feed_version',
-  HidePublicationDate = 'hide_publication_date',
-  LoginModalButtonCopyPrefix = 'login_modal_button_copy_prefix',
-  LoginModalDescriptionCopy = 'login_modal_description_copy',
+export class Features {
+  static readonly SignupButtonCopy = new Features(
+    'SignupButtonCopy',
+    'signup_button_copy',
+    'Access all features',
+  );
+
+  static readonly DevcardLimit = new Features(
+    'DevcardLimit',
+    'feat_limit_dev_card',
+  );
+
+  static readonly FeedVersion = new Features(
+    'FeedVersion',
+    'feed_version',
+    '1',
+  );
+
+  static readonly HidePublicationDate = new Features(
+    'HidePublicationDate',
+    'hide_publication_date',
+    '0',
+  );
+
+  static readonly LoginModalButtonCopyPrefix = new Features(
+    'LoginModalButtonCopyPrefix',
+    'login_modal_button_copy_prefix',
+    'Sign in with',
+  );
+
+  static readonly LoginModalDescriptionCopy = new Features(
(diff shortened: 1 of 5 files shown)


packages/shared/src/lib/featureManagement.ts
```diff
@@ -1,21 +1,58 @@
...
+  );
+
+  static readonly LoginModalDescriptionCopy = new Features(
+    'LoginModalDescriptionCopy',
+    'login_modal_description_copy',
+    'Unlock useful features by signing in. A bunch of cool stuff like content filters and bookmarks are waiting just for you.',
+  );
+
+  private constructor(
+    private readonly key: string,
```
idoshamun, line 40: Why do we need this property?

idoshamun (changes requested):
Looks great! Just one question

sshanzel:
I think this is a great idea. What about creating a custom Provider for the `FeaturesContext` which will return the features automatically and in that context we can decide what will be the actual value for each property (either through `getFeatureValue` or `isFeatureEnabled`). This way we won't have to always call the FeaturesContext and pass the flags extracted from it - just pass the features available from the context itself.

packages/shared/src/lib/featureManagement.ts
rebelchris: This actually makes sure we can access them as we did before with `Features.FeatureName`
Just thought it was nicer to use from the developer point of view.

idoshamun: Oh I thought that the fact they're defined as `static` already takes care of it, is it not?

rebelchris: No it's actually the class called itself with this constructor, so each static is a usage of this class:

```
static readonly LoginModalDescriptionCopy = new Features(
    'LoginModalDescriptionCopy',
    'login_modal_description_copy',
    'Unlock useful features by signing in. A bunch of cool stuff like content filters and bookmarks are waiting just for you.',
  );
```

And the magic is in the `toString` method on the class, making it default return the key.

idoshamun: Not sure that we're on the same page. If you define static variable `V` in class `C`, you should be able to access it by using `V.C`. This is why I'm asking why the magic is actually required (it's super cool though!!)

rebelchris: omg, haha I was half asleep I guess, removed it now indeed 😂

rebelchris:
@sshanzel I'm not quite sure I understand this one 100%
We would still be calling the FeaturesContext then right?

We would omit the usage of the `getFeatureValue` as that would now sit in the provider?

But maybe I'm reading the comment wrong.

sshanzel:
Like on auth or alerts context. They have custom provider. It we create a custom provider for context, they can have direct access to flags. This would resuce verbosity of having to call FeaturesContext just to have access to the flags data.

rebelchris:
@sshanzel Very good point actually! Shall we pick this up separate? I'll log it in Jira 👏
@idoshamun What's your opinion on this, good to add?

idoshamun (approved):
💪

PR #798, opened 2022-01-06
refactor: update page state for duration based events

idoshamun:
Duration-based events capture the page state at the time when the event starts. This can create cases where at first the page state was passive (open new tab) so all the impressions are marked as passive, and then the user decided to focus on the tab and continue their use on our platform. So in this case we want to mark all these impressions as active impressions or any other event for that matter.
Another important change is to mark events as active if the user scrolled down the page. Even if the page state is passive the user can still scroll it. I think that if the user scrolls it means that they're active and we should adjust the state accordingly.


sshanzel (approved):
Great thought about being active while scrolling btw 💪

PR #799, opened 2022-01-06
refactor: send request to create a feed if filters exist

sshanzel:
Only when this PR is merged I will need to update the fetching of local storage: https://github.com/dailydotdev/apps/pull/791/files

### DIFF
--- a/packages/shared/src/components/profile/ProfileForm.tsx
+++ b/packages/shared/src/components/profile/ProfileForm.tsx
@@ -7,6 +7,7 @@ import React, {
   useState,
 } from 'react';
 import classNames from 'classnames';
+import request from 'graphql-request';
 import { updateProfile, UserProfile } from '../../lib/user';
 import { TextField } from '../fields/TextField';
 import { Switch } from '../fields/Switch';
@@ -20,6 +21,8 @@ import {
   getTimeZoneOptions,
   getUserInitialTimezone,
 } from '../../lib/timezones';
+import useMutateFilters from '../../hooks/useMutateFilters';
+import { getLocalFeedSettings } from '../../hooks/useFeedSettings';
 
 const REQUIRED_FIELDS_COUNT = 4;
 const timeZoneOptions = getTimeZoneOptions();
@@ -59,6 +62,8 @@ export default function ProfileForm({
       update: mode === 'update',
     }),
   );
+  const { followTags, blockTag, unfollowSource, updateAdvancedSettings } =
+    useMutateFilters();
   const [usernameHint, setUsernameHint] = useState<string>();
   const [twitterHint, setTwitterHint] = useState<string>();
   const [githubHint, setGithubHint] = useState<string>();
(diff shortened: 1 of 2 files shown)


packages/shared/src/components/profile/ProfileForm.tsx
```diff
@@ -123,6 +131,28 @@ export default function ProfileForm({
       const filledFields = Object.keys(data).filter(
         (key) => data[key] !== undefined && data[key] !== null,
       );
+      if (shouldShowMyFeed) {
```
rebelchris, line 134: Just a note here, I think we can remove this check.
There could be an edge case where the users AB test might have been removed/changed.

I think if they have local feed settings, we should just always apply them anyway.
What do you think?

sshanzel: Very sharp thinking. Let me remove this one.

rebelchris (approved):
Nice clean, simple and it does the trick 💖

idoshamun (changes requested):
Let's utilize GraphQL power.
Also I'm missing relevant tests

packages/shared/src/components/profile/ProfileForm.tsx
```diff
@@ -123,6 +127,26 @@ export default function ProfileForm({
...
+        : Promise.resolve();
+      const onUpdateAdvancedSettings = advancedSettings?.length
+        ? updateAdvancedSettings({ advancedSettings })
+        : Promise.resolve();
+      await Promise.all([
+        onFollowTags,
+        onBlockTags,
+        onUnfollowSource,
+        onUpdateAdvancedSettings,
+      ]);
```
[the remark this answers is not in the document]
sshanzel: Oh, I totally agree.

rebelchris (approved):
Wow amazing how much we could refactor to one graphQL query.

packages/shared/src/components/profile/ProfileForm.tsx
```diff
@@ -123,6 +130,14 @@ export default function ProfileForm({
       const filledFields = Object.keys(data).filter(
         (key) => data[key] !== undefined && data[key] !== null,
       );
+
+      const feedSettings = getLocalFeedSettings(true);
+
+      if (feedSettings) {
+        await updateFeedFilters(feedSettings);
```
rebelchris, line 137: As mentioned I think we should set the update filter here if the user had any filters.

rebelchris (changes requested):
A new remark in combination with the discussion in the other branch.
The user already has filters, so it's fair to say we should update the alert for them.

As this would have happened when the filtered while being logged in.

sshanzel:
So cool to see we have batched 3 requests into 1.

@idoshamun should now be ready for your sharp eyes 🦅 

@rebelchris I have now included the mutation for filters as we have confirmed there would be no issue in this part.

packages/shared/src/graphql/feedSettings.ts
```diff
@@ -98,6 +98,23 @@ export const FEED_SETTINGS_QUERY = gql`
...
+  ) {
+    addFiltersToFeed(filters: $filters) {
+      id
+    }
+    updateFeedAdvancedSettings(settings: $settings) {
+      id
+    }
+    updateUserAlerts(data: { filter: false }) {
+      filter
+    }
```
rebelchris, line 114: Wow nice!!

Is it still correct to do this manually in the sidebar? 
Or should we maybe consider later on refactoring to something similar?

packages/shared/src/components/profile/ProfileForm.spec.tsx
```diff
@@ -37,24 +45,54 @@ const defaultUser = {
...
+  blockedTags: ['javascript'],
+  includeTags: ['react'],
+  excludeSources: [{ id: 'test', name: 'Source', image: 'a.b.c' }],
+};
+
+const createUpdateFeedFiltersMock = (
+  mock: ReturnType<typeof jest.fn>,
+  { advancedSettings, ...rest }: FeedSettings = defaultFeedSettings,
+) => {
+  const variables = { filters: rest, settings: advancedSettings };
```
rebelchris, line 59: Perhaps we should include the alerts setting in this test as well?
What do you think?

rebelchris (changes requested):
Wow Lee!

Impressive work and so happy to see with could just enable this in GraphQL 🌟 
One little note on testing for the alert, not sure what you think on that point?

packages/shared/src/graphql/feedSettings.ts
sshanzel: I am planning to refactor to apply it on users having the experiment disabled as we still need support the case outside experiment.

packages/shared/src/components/profile/ProfileForm.spec.tsx
sshanzel: I tried but it fails, I think it is due to the fact we are mocking the parameters and the value was strongly typed.

rebelchris (approved):
Nice, well done on this Lee!

sshanzel:
Thanks for the awesome reviews @rebelchris! 🚢

idoshamun (approved):
Sweet! Let's merge :)

PR #802, opened 2022-01-06
refactor: make signup modal copies dynamic

ayadotdev:
Added AB testing for signup modal title and buttons copies. 
Added AB testing to hide a profile image.

DD-416 #done

### DIFF
--- a/packages/shared/src/lib/featureManagement.ts
+++ b/packages/shared/src/lib/featureManagement.ts
@@ -28,22 +28,22 @@ export class Features {
     'Unlock useful features by signing in. A bunch of cool stuff like content filters and bookmarks are waiting just for you.',
   );
 
-  static readonly SignupModalTitleCopy = new Features(
-    'signup_modal_title_copy',
+  static readonly SignupTitleCopy = new Features(
+    'signup_title_copy',
     'Set up your profile',
   );
 
-  static readonly HideSignupModalProfileImage = new Features(
-    'hide_signup_modal_profile_image',
+  static readonly HideSignupProfileImage = new Features(
+    'hide_signup_profile_image',
   );
 
-  static readonly SignupModalSubmitButtonCopy = new Features(
-    'signup_modal_submit_button_copy',
+  static readonly SignupSubmitButtonCopy = new Features(
+    'signup_submit_button_copy',
     'Finish',
   );
 
-  static readonly SignupModalLogoutButtonCopy = new Features(
-    'signup_modal_logout_button_copy',
+  static readonly SignupLogoutButtonCopy = new Features(
+    'signup_logout_button_copy',
     'Logout',
   );
 
--- a/packages/shared/src/lib/featureManagement.ts
+++ b/packages/shared/src/lib/featureManagement.ts
@@ -28,6 +28,26 @@ export class Features {
     'Unlock useful features by signing in. A bunch of cool stuff like content filters and bookmarks are waiting just for you.',
   );
 
+  static readonly SignupModalTitleCopy = new Features(
+    'signup_modal_title_copy',
+    'Set up your profile',
+  );
+
+  static readonly HideSignupModalProfileImage = new Features(
+    'hide_signup_modal_profile_image',
+    '0',
+  );
+
+  static readonly SignupModalSubmitButtonCopy = new Features(
+    'signup_modal_submit_button_copy',
+    'Finish',
+  );
+
+  static readonly SignupModalLogoutButtonCopy = new Features(
+    'signup_modal_logout_button_copy',
+    'Logout',
+  );
+
   private constructor(
     public readonly id: string,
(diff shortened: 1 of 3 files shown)


packages/webapp/pages/register.tsx
```diff
@@ -40,11 +50,16 @@ export default function Register(): ReactElement {
...
+              {getSignupModalFeatureValue(Features.SignupModalTitleCopy)}
+            </ProfileHeading>
             <h2 className="self-start my-2 text-theme-label-tertiary typo-callout">
               Please fill in your details below
             </h2>
-            <EditImageWithJoinedDate user={user} />
+            {!parseInt(
+              getSignupModalFeatureValue(Features.HideSignupModalProfileImage),
+              10,
+            ) && <EditImageWithJoinedDate user={user} />}
```
rebelchris, line 62: For this one it would be better to use `isFeaturedEnabled` 

It worked like this:

```js
{!isFeaturedEnabled(Features.HideSignupModalProfileImage) && <EditImageWithJoinedDate user={user} />}
```

rebelchris (changes requested):
One small thing, other than that, it looks amazing 👏
Well done 🎉

ayadotdev:
Note:
Removing a profile image there is a space between headings and the Profile form which has some margin.

Screenshot 2022-01-06 at 18 59 13

sshanzel (changes requested):
It is super awesome that you noted the empty space when the profile image is hidden, the link below to the SC should be able to help with how much the margin must be applied - the spacing should exactly be `24px`.

Let me know if you need help navigating through SC 🚢 

https://www.sketch.com/s/28f2d32c-2923-45b9-91d5-ad7c27be17c9/a/7yr21AP#Inspector

packages/webapp/pages/register.tsx
```diff
@@ -12,13 +12,27 @@ import {
...
   const { trackEvent } = useContext(AnalyticsContext);
+  const { flags } = useContext(FeaturesContext);
+
+  const getSignupModalFeatureValue = (featureFlag: Features) => {
+    return getFeatureValue(featureFlag, flags);
+  };
+
+  const isFeatureEnabled = (featureFlag: Features) => {
+    return parseInt(getSignupModalFeatureValue(featureFlag), 10);
+  };
```
rebelchris, line 35: Oh sorry my bad, I see we don't have this setup in the `featureManagement` here.
This should be sitting in there as soon as we merge our current branch.

For now you could add the following code to the `featureManagement` file so we have it aligned.

```js

export const isFeaturedEnabled = (key: Features, flags: IFlags): boolean =>
  flags[key?.id]?.enabled;
```

See the following PR for the new file:
https://github.com/dailydotdev/apps/blob/DD-354-my-feed/packages/shared/src/lib/featureManagement.ts

packages/shared/src/lib/featureManagement.ts
```diff
@@ -28,6 +28,26 @@ export class Features {
...
   );
 
+  static readonly SignupModalTitleCopy = new Features(
+    'signup_modal_title_copy',
+    'Set up your profile',
+  );
+
+  static readonly HideSignupModalProfileImage = new Features(
+    'hide_signup_modal_profile_image',
+    '0',
```
rebelchris, line 38: When using the boolean checked values, we can remove the default value here.

rebelchris (changes requested):
Almost there!

My bad for forgetting we don't have this in master yet.

rebelchris (approved):
Nice this is perfect, if you can just check the margin note from Lee it will be done 🎉

sshanzel (approved):
LGTM! 🚀

packages/webapp/pages/register.tsx
```diff
@@ -12,13 +12,24 @@ import {
...
 export default function Register(): ReactElement {
   const { user, logout } = useContext(AuthContext);
   const router = useRouter();
   const [disableSubmit, setDisableSubmit] = useState<boolean>(true);
   const { trackEvent } = useContext(AnalyticsContext);
+  const { flags } = useContext(FeaturesContext);
+
+  const getSignupModalFeatureValue = (featureFlag: Features) => {
+    return getFeatureValue(featureFlag, flags);
+  };
```
idoshamun, line 32: Not sure if we need this function actually...

idoshamun (commented):
Looks great I had just one optimization comment but nothing major.
Anyway, for next time it's better to wait for my review before merging :)

packages/webapp/pages/register.tsx
ayadotdev: @idoshamun I added that function for readability purpose. But you are right, it may be redundant. Happy to remove it if you feel it is better without :thumbsup:

idoshamun: Let it be for now 😁

PR #803, opened 2022-01-07
fix: navigation not shown even when user has filters

sshanzel:
DD-457 #done
DD-455 #done

~~Currently, the decision to identify whether the user has any filter is based on the `alerts.filter` which is used to display alerts, for us to declare a user has any filter is to check if there's an actual filter applied on their feed settings. The case results in the "My Feed" not being displayed.~~

image

Also, I noticed a false positive in one of the test cases namely `should remove feed filter button once user has filters`
I think the primary purpose of this test is to see that the button is not rendered anymore since the user has an existing filter. With that, it clicks the "Create my feed" button which should have not been there in the first place. Then it tests the `updateAlerts` to be called which should only happen when the user has clicked the close button (already covered) or the user with pre-existing filters open the `FeedFilters` but it proceeds since that specific button also opens `FeedFilters`.

In terms of testing the experiment, I have made it to be explicitly defined so we can be sure our pre-existing tests were not affected by it.

### DIFF
--- a/packages/shared/src/hooks/useFeedSettings.ts
+++ b/packages/shared/src/hooks/useFeedSettings.ts
@@ -33,6 +33,12 @@ export type FeedSettingsReturnType = {
 
 let avoidRefresh = false;
 
+export const getHasAnyFilter = (feedSettings: FeedSettings): boolean =>
+  feedSettings?.includeTags?.length > 0 ||
+  feedSettings?.blockedTags?.length > 0 ||
+  feedSettings?.excludeSources?.length > 0 ||
+  feedSettings?.advancedSettings?.length > 0;
+
 export const LOCAL_FEED_SETTINGS_KEY = 'feedSettings:local';
 
 export const updateLocalFeedSettings = (feedSettings: FeedSettings): void => {
@@ -110,19 +116,13 @@ export default function useFeedSettings(): FeedSettingsReturnType {
     }, 100);
   }, [tagsCategories, feedSettings, avoidRefresh]);
 
-  const hasAnyFilter =
-    feedSettings?.includeTags?.length > 0 ||
-    feedSettings?.blockedTags?.length > 0 ||
-    feedSettings?.excludeSources?.length > 0 ||
-    feedSettings?.advancedSettings?.length > 0;
-
   return useMemo(() => {
     return {
       tagsCategories,
       feedSettings,
       isLoading,
       advancedSettings,
-      hasAnyFilter,
+      hasAnyFilter: getHasAnyFilter(feedSettings),
       setAvoidRefresh,
     };
   }, [
@@ -130,7 +130,6 @@ export default function useFeedSettings(): FeedSettingsReturnType {
     feedSettings,
     isLoading,
     advancedSettings,
-    hasAnyFilter,
     setAvoidRefresh,
   ]);
 }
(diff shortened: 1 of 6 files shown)


rebelchris:
@sshanzel This was actually a managed decision logged in the beginning of taking on this task.
My understanding was a person now will be able to reset their actual filters and it would remove their My Feed?

See discussion here:
https://dailydotdev.slack.com/archives/C02AA18RD1D/p1640844453027100

rebelchris:
@sshanzel My idea was to set the `filters = false` on your other branch.
That captures the registered user. 

Since that is actually what happens.
I would really suggest going that way, as with this we introduce a condition that the user can revert, which will cause confusion.

sshanzel:
That's a good idea. Then we should apply it to the registration PR. With that, I think we will have to keep the tests here since as a new user they shouldn't be seeing the button again if we will turn the `alerts.filter` to `false` soon as possible.

sshanzel:
That's a good idea. Then we should apply it to the registration PR, the only issue would if in that split second after successfully registering in the `gateway` is the `api` would be able to immediately receive and save the user since the `userId` is an FK at `alerts` table. It is literally a split second after the `updateProfile` execution and now saves the filter.

With that, I think we will have to keep the tests here since the `alerts.filter` shouldn't be updated when they click the "Create my feed" button.

rebelchris:
@sshanzel The register page doesn't have the sidebar, so shouldn't be a problem I think.
Agreed on the test they make sense to me 👏

sshanzel:
When they get redirected to the home page, they will have `alerts.filter` = true before, now that should be covered if we're gonna update the `false` value.

sshanzel:
@idoshamun what do you think about updating the FK constraints from the `alerts` in `api` DB  to ensure it won't produce an error when the user has registered with filters so we can utilize the `alerts.filter` = `false` for catching the edge case of users with pre-existing filters but removed them.

idoshamun:
I'm not a fan of removing the foreign key from the table. Can you please elaborate on the issue we're trying to solve with this solution?

packages/shared/src/hooks/useFeedSettings.ts
```diff
@@ -33,6 +33,12 @@ export type FeedSettingsReturnType = {
 
 let avoidRefresh = false;
 
+export const getHasAnyFilter = (feedSettings: FeedSettings): boolean =>
```
rebelchris, line 36: In the current implementation this is only used in the test, so pretty sure we can remove it right?

rebelchris (changes requested):
Think we can remove the hasAnyFilters one.
What do you think?

packages/shared/src/hooks/useFeedSettings.ts
sshanzel: It is also being used in the `useFeedSettings`, just to be sure they are using the same process to fetch the correct value.

https://github.com/dailydotdev/apps/blob/3fdfe38302ba5fcde4b71fc3a6d44c390070c643/packages/shared/src/hooks/useFeedSettings.ts#L125

sshanzel:
After further discussion, the scenario where the user information might not be available yet right after submitting registration was deemed a concern we won't have to worry about.

This PR now just contains a fix for updating the `alerts.filter` on users not part of the experiment and the tests cases aligned to the actual scenario.

idoshamun (approved):
Alright!!

PR #805, opened 2022-01-10
Dd 342 extension build

rebelchris:
Added a new job to test the build of the extension.
It happened in the past we introduced a breaking change only on the extension part, and it's critical to catch these issues early on.

This PR introduces a new job that runs parallel to the tests *(and uses the same cache in available)

Screenshot 2022-01-10 at 13 48 56

DD-342 #done

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -81,3 +81,4 @@ workflows:
   build:
     jobs:
       - build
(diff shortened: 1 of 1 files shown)


idoshamun (approved):
Awesome!!

PR #807, opened 2022-01-10
feat: :sparkles: Moved the DND button to the sidebar

rebelchris:
In this PR I've moved the DnD button to the sidebar vs having it in the profile dropdown.

We still need to load it from the MainFeedPage, as the alert is shown on top of the page.

We also changed the name to be "Focus mode" and the icon to a Moon.
The icon is the filled one for now, as we don't use the outline anywhere else.

Screenshot 2022-01-10 at 14 52 56

DD-439 #done


idoshamun (approved):
Woop woop!

PR #808, opened 2022-01-10
feat: :sparkles: Moved to Tailwind v3 with all the new elements needed

rebelchris:
This PR introduces the merge from Tailwind V2 to V3.
Enabling JIT by default so no need to state it anymore.

Also variants are enabled by default so we could remove it.

And it comes with it's own tailwindcss/nesting module, which cleans up the process a lot.

Ran tests on webapp and extension, all seems to work as we expect.

DD-359 #done 

Note: (Might be good to look at the preview with a couple of people?)


sshanzel (changes requested):
Super cool that you have taken this task! 🎉 

In our config, we have a custom value for `width` namely `w-fit` which is now supported out-of-the-box. I think we will have to remove that and see it applies the `width: fit-content` style.

Just a general question, since we have migrated the `flex-shrink` to `shrink` - should we migrate the `flex-grow` to `grow` as well? I understand there are much more implementations of `grow` and things would still be working, just wanted to know your thoughts about it.

rebelchris:
@sshanzel Thanks for the `w-fit` remark removed that.
As for flex-grow we actually don't use it. we use `flex-1` which implements a combo, so there is no need to refactor those as I understand.

The other thing would be the `transform` and `filter` classes, but thought we might want to do that later for safety purposes.

@idoshamun Made sure everything works on build now, and double checked manually from the build.

idoshamun (approved):
💪

sshanzel (approved):
LGTM! 🚀

PR #809, opened 2022-01-11
fix: my feed alert close button

sshanzel:
- [x] The color of the close button should be “Tertiary” (currently at “Primary”).

- [x] The green dot is missing on the sidebar when minimized.

- [x] The close icon should be a button rather than an icon to have the proper styles and effects.


rebelchris (approved):
Nice work!

ayadotdev (approved):

idoshamun (approved):
Woop!

PR #812, opened 2022-01-11
refactor: explicitly set 'created' for my feed

sshanzel:


sshanzel:
I think it is safe to say that we can close this PR now with the changes in the PR for API:

https://github.com/dailydotdev/daily-api/pull/849

sshanzel:
Reopened to ensure the correct value for `myFeed` is registered. Currently, the `alerts.filter` has shown the accurate and expected value.

cc @idoshamun @rebelchris

rebelchris (approved):
Makes sense.
Thanks for including this.

Was this an issue for the people already having alerts?
As we set the default value to `created` just want to make sure that still works as expected.

sshanzel:
Yes, somehow the default value is being utilized. Now that we have the default value of `myFeed` as `null`, we will need to explicitly set the value as 'created'.

idoshamun:
According to your recent change in the API, shouldn't it happen automatically?

sshanzel:
Yes, I think a race could be happening since the CDC is running on parallel and the request for making the alerts to `false` as well.

PR #813, opened 2022-01-11
fix: boot data invalidation

sshanzel:
When the registration finishes and we proceed with the redirection the boot query is not being requested which should actually happen as the data were updated (even on window.location.replace the state doesn't refresh).

For the accurate alerts to be displayed right after registration, the user needs to refetch the boot data.

### DIFF
--- a/packages/shared/src/components/profile/ProfileForm.tsx
+++ b/packages/shared/src/components/profile/ProfileForm.tsx
@@ -7,6 +7,7 @@ import React, {
   useState,
 } from 'react';
 import classNames from 'classnames';
+import { useQueryClient } from 'react-query';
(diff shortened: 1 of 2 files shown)


packages/shared/src/components/profile/ProfileForm.tsx
```diff
@@ -139,6 +142,7 @@ export default function ProfileForm({
       if (feedSettings) {
         await updateFeedFilters(feedSettings);
         storage.removeItem(LOCAL_FEED_SETTINGS_KEY);
+        await client.invalidateQueries(BOOT_QUERY_KEY);
```
idoshamun, line 145: Love it!

idoshamun (approved):
Great catch!

PR #814, opened 2022-01-11
fix: my feed navigation item styles

sshanzel:
- Fixes the my-feed being active (displaying active CSS)
- Fixes extension navigation


idoshamun (approved):
Looks great! This reminded me that we must check the feature on the extension as well before releasing.

sshanzel:
Totally, I will also run the end-to-end test on extension locally.

PR #815, opened 2022-01-11
refactor: change font size for text fields and dropdown label to 17px

ayadotdev:
Changed font size for text fields and dropdown label to 17px as indicated in the style guide.

DD-315 #done


ayadotdev:
Screenshot 2022-01-11 at 21 03 22

Note. The text area font size is 15px. I couldn't find this component in the style guide. Should it be also changed to 17px?

sshanzel (changes requested):
For this ticket, I think there are lots of changes necessary as the guideline affected the whole form.

I commented on the actual ticket and asked Tsahi if the old guideline is available so we can easily see the comparison.

But I think there's gonna be a lot of work for this.

Feel free to ping us if you have any questions understanding the guideline provided in the ticket.

rebelchris (changes requested):
This one is fixed well.
Did you manage to check the other points as well to reflect the guidelines?

I think I noted the border-radius for instance.

packages/shared/src/components/fields/TextField.tsx
```diff
@@ -137,6 +137,7 @@ export function TextField({
           },
           styles.field,
         )}
+        style={{ ...style, borderRadius: '0.625rem' }}
```
rebelchris, line 140: The border radius should be `rounded-10`
We should already have that class.

rebelchris (commented):
Great setup already.
I think I went through all fields on the profile form and the spec seems perfect there.

The only thing you might want to check with Tsahi is the "Right Count"
He did not mention this in the color changes, but ask him if it is the same as the "Right Icon" color schemes.

Example on hover it's primary:
Screenshot 2022-01-13 at 14 42 53

packages/shared/src/components/fields/TextField.tsx
```diff
@@ -137,6 +137,7 @@ export function TextField({
           },
           styles.field,
         )}
+        style={{ ...style, borderRadius: 'rounded-10' }}
```
rebelchris, line 140: Oh sorry, actually meant we can apply if to the parent element. no need to set a style for it.

rebelchris (changes requested):
Looks great one small thing

packages/shared/src/components/fields/TextField.tsx
ayadotdev: This component already has rounded-10 (line 133: `compact ? 'h-9 rounded-10`'). Should I remove this line `style={{ ...style, borderRadius: 'rounded-10' }}`?

rebelchris: Ah yes, then it can actually be removed as it would add them double.
But maybe double-check this after you remove it.

rebelchris (approved):
Well done!
In the end it wasn't that much changes, but glad to see everything is inline with the guideline now.

@idoshamun @sshanzel If you guys could have a second set of eyes 🥳

packages/shared/src/components/fields/TextField.module.css
```diff
@@ -10,7 +10,6 @@
 
   &:global(.invalid) {
     box-shadow: inset 0.125rem 0 0 0 var(--theme-status-error);
-    border-color: var(--theme-status-error);
```
idoshamun, line 13: Why are we dropping this?

idoshamun (changes requested):
Only one question, other than that it looks great

sshanzel (approved):
LGTM! 🚢

packages/shared/src/components/fields/TextField.module.css
ayadotdev: According to the guideline this is a design for a text field with Error status: 

Screenshot 2022-01-19 at 13 26 56

With `border-color: var(--theme-status-error);` the full border applies:

Screenshot 2022-01-19 at 13 41 40

Please let me know if misinterpreted something.

idoshamun: Ok, you're right! Let's proceed :)

idoshamun (approved):

PR #817, opened 2022-01-12
fix: registration with filters through extension and redirect accurately

sshanzel:
Introduced a hook that has the steps to do when registering the filters from local storage. Also centralized the flag variable for my feed.

We are now passing a query parameter for when the user has the flag enabled and registered but started from the extension.

### DIFF
--- a/packages/shared/src/components/profile/ProfileForm.tsx
+++ b/packages/shared/src/components/profile/ProfileForm.tsx
@@ -22,7 +22,6 @@ import {
 } from '../../lib/timezones';
 import { Features, isFeaturedEnabled } from '../../lib/featureManagement';
 import FeaturesContext from '../../contexts/FeaturesContext';
-import { useMyFeed } from '../../hooks/useMyFeed';
 
 const REQUIRED_FIELDS_COUNT = 4;
 const timeZoneOptions = getTimeZoneOptions();
@@ -57,7 +56,6 @@ export default function ProfileForm({
   ...props
 }: ProfileFormProps): ReactElement {
   const { user, updateUser } = useContext(AuthContext);
-  const { registerLocalFilters } = useMyFeed();
   const formRef = useRef<HTMLFormElement>(null);
 
   const [userTimeZone, setUserTimeZone] = useState<string>(
@@ -132,12 +130,7 @@ export default function ProfileForm({
         (key) => data[key] !== undefined && data[key] !== null,
       );
 
-      const { hasFilters } = await registerLocalFilters();
-
-      onSuccessfulSubmit?.(
-        filledFields.length > REQUIRED_FIELDS_COUNT,
-        hasFilters,
-      );
+      onSuccessfulSubmit?.(filledFields.length > REQUIRED_FIELDS_COUNT);
     }
   };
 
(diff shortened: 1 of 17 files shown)


rebelchris (approved):
Nice, super nice catch!
Wasn't a 100% sure if the webapp profile form doesn't interfear, but thanks for explaining on slack.

rebelchris (approved):
Nice work, love the fact that you are setting the active feed to my-feed. Should we also consider doing that on the webapp?

sshanzel:
Thank you for the review!

Great point! Though I think it will automatically be set since we will redirect the user to `/my-feed`. What do you think?

packages/webapp/pages/register.tsx
```diff
@@ -33,18 +33,31 @@ export default function Register(): ReactElement {
...
 
+  const getRedirectUri = (hasFilters?: boolean) => {
+    const uri = router.query.redirect_uri as string;
+
+    if (!uri) {
+      return hasFilters ? '/my-feed' : '/';
+    }
+
+    return hasFilters ? `${uri}?create_filter=true` : uri;
+  };
```
idoshamun, line 44: I originally thought of adding this variable here: https://github.com/dailydotdev/apps/blob/master/packages/shared/src/components/LoginButtons.tsx#L24
In the login button, rather than messing with the url in registration.
It feels like a more solid implementation, I think, and makes the registration page unaware of everything.
What do you think?

idoshamun (changes requested):
Please check my comment

packages/webapp/pages/register.tsx
sshanzel: Ohh. Very interesting. Let me work on this one first thing!

sshanzel:
@idoshamun I have now refactored it to apply the parameter in the `redirect_uri`. Thank you for the review! 🚢

packages/shared/src/components/profile/ProfileForm.tsx
```diff
@@ -40,7 +31,10 @@ export type RegistrationMode = 'default' | 'author' | 'update';
 
 export interface ProfileFormProps extends HTMLAttributes<HTMLFormElement> {
   setDisableSubmit?: (disable: boolean) => void;
-  onSuccessfulSubmit?: (optionalFields: boolean) => void | Promise<void>;
+  onSuccessfulSubmit?: (
+    optionalFields: boolean,
+    hasFilters?: boolean,
```
idoshamun, line 36: No needed any more I guess...

packages/shared/src/hooks/useMyFeed.ts
```diff
@@ -0,0 +1,52 @@
...
+    await client.invalidateQueries(BOOT_QUERY_KEY);
+
+    return { hasFilters: true };
+  };
+
+  return useMemo(
+    () => ({
+      registerLocalFilters,
+      shouldShowMyFeed,
+      hasAnyLocalFilter: !!getLocalFeedSettings(true),
```
idoshamun, line 48: I think we can optimize this so we will not have to read from local storage all the time

idoshamun (changes requested):
Almost there

idoshamun (approved):
Woohoo!!!

PR #818, opened 2022-01-12
refactor: multi-level menu directly opened tab

sshanzel:
When the user clicks the "Create my feed" button, it should display the "Tags menu" tab automatically.

### DIFF
--- a/packages/shared/src/components/multiLevelMenu/MultiLevelMenu.tsx
+++ b/packages/shared/src/components/multiLevelMenu/MultiLevelMenu.tsx
@@ -1,19 +1,37 @@
-import React, { useState, ReactElement } from 'react';
+import React, { useState, ReactElement, useEffect } from 'react';
 import { MenuItem } from '../filters/common';
 import MultiLevelMenuDetail from './MultiLevelMenuDetail';
 import MultiLevelMenuMaster from './MultiLevelMenuMaster';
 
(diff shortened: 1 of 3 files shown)


rebelchris (approved):
Nice work, nothing to add from my side.

packages/shared/src/components/multiLevelMenu/MultiLevelMenu.tsx
```diff
@@ -1,20 +1,38 @@
...
+    }
+
+    const menuItem = menuItems.find((item) => item.title === directlyOpenedTab);
+
+    if (!menuItem) {
+      return;
+    }
+
+    setMultiLevelMenuDetailItem(menuItem);
+  }, [directlyOpenedTab]);
```
idoshamun, line 34: Instead of `useEffect`, we can inject the initial value for the `multiLevelMenuDetailItem` in its `useState`.
I think it's the preferred implementation

idoshamun (commented):
There's some room for improvement there :)

packages/shared/src/components/multiLevelMenu/MultiLevelMenu.tsx
sshanzel: Oh, sorry about merging early. I agree that would make things much cleaner. Let me push this one quickly.

PR #819, opened 2022-01-13
fix: postcss fix for the extension

rebelchris:
A hotfix for the extension to use the double require import, where next will handle it for us.


idoshamun (approved):

sshanzel (approved):
💪

PR #820, opened 2022-01-14
fix: rendering my feed nav item on mobile

sshanzel:
Just as on our current implementation, we are not allowing feed filters to be changed on mobile hence, not seen in my feed nav item.

Preview:

image


idoshamun (approved):
Looks great just make that it doesn't show the CTA on mobile and that everything on desktop works

sshanzel:
Confirmed, unnecessary components are not displayed on mobile. Merging! 🚢

PR #821, opened 2022-01-14
fix: my feed as the default feed

rebelchris:
Introducing a fix for setting my-feed as the default if the user has not selected one yet.

The problem with this is that we actually don't wait for data to return, and react-query (in the persistent context) does not fire on changing the default value.

This resulted in it always being the popular feed.

In this PR I've introduced a workaround in the persistentContext.
By using useEffect and checking if there is no stored value, we can actually optimistically update the default value if it changed.

Discussed three approaches:
1. Wait rendering feed until boot loaded (Not really an option as we don't want to block loads)
2. This approach built out
3. Doing this same effect but in the components, caused longer re-render time and had to be done in three places

Super keen to hear what you think of this

DD-467 #done


idoshamun:
We are already waiting for the boot to finish before rendering because we need to refresh the token so I wonder why it doesn't work still?

rebelchris:
In this case the issue might be more around the `defaultFeed` variable from usePersistentContext never updating when it changes.

It starts as undefined, which will become the evaluation criteria and forcing it to set `popular`.

idoshamun:
So does this PR addresses it correctly? Can I review it?

rebelchris:
@idoshamun yes this fix works a the most stable solution from the tested ones.
You can review it.

However we might want to consider adding a ticket for the comment you described above

packages/shared/src/hooks/usePersistentContext.ts
```diff
@@ -24,6 +25,14 @@ export default function usePersistentContext<T>(
...
   );
 
+  useEffect(() => {
+    const hasDefaultValue = getAsyncCache(key, null);
+    hasDefaultValue.then((defaultValue) => {
+      if (defaultValue !== null) return;
+      queryClient.setQueryData(key, valueWhenCacheEmpty);
+    });
+  }, [valueWhenCacheEmpty]);
+
```
idoshamun, line 35: A better solution for this would be to use a different query key for `useQuery`. Instead of `key`, we can use `[key, valueWhenCacheEmpty]`. This will make sure to trigger again the function when one of them changes which is exactly what we want to do.

idoshamun (changes requested):

rebelchris:
@idoshamun Refactored, way nicer like this 🤘

idoshamun (approved):
Woohoo!

PR #822, opened 2022-01-14
refactor: opt-in settings for utilizing ranking algorithm by sorting date 

sshanzel:
I have deployed this on preview so we can have better testing of this feature:

https://preview.app.daily.dev/popular


packages/shared/src/components/MainFeedLayout.tsx
```diff
@@ -123,6 +123,12 @@ const getQueryBasedOnLogin = (
   return null;
 };
 
+const algorithms = [
```
rebelchris, line 126: Would a typescript ENUM not be nicer here?

rebelchris (commented):
The code looks like a solid implementation
Do have some issues while using it.

First load: only get one article?
Screenshot 2022-01-14 at 13 31 19

Switching to my-feed and back shows all articles?

Then toggling on the setting, while on the page gives me this:
Screenshot 2022-01-14 at 13 32 00

Again switching feed will fix it.

Now it seems the popular always only returns one
(might not be related to this issue)

sshanzel:
I think you could be having network issues? I tried the scenarios and it seems to display the necessary elements.


https://user-images.githubusercontent.com/13744167/149510184-ca2c7df9-6f00-42bf-9ec1-96edda05e908.mov

packages/shared/src/components/MainFeedLayout.tsx
sshanzel: Good idea. I think I can create one for the value property since the array is a must for the options. Let me push that real quick.

rebelchris (approved):
Nice work!

The previously mentioned things are a general issue, not related to this PR.

packages/shared/src/components/MainFeedLayout.tsx
```diff
@@ -230,9 +240,21 @@ export default function MainFeedLayout({
...
-    const variables = isUpvoted
-      ? { ...query.variables, period: periods[selectedPeriod].value }
-      : query.variables;
+
+    const getVariables = () => {
+      if (isUpvoted) {
+        return { ...query.variables, period: periods[selectedPeriod].value };
+      }
+
+      if (sortingEnabled && isSortableFeed) {
```
idoshamun, line 249: I'd suggest this change. Otherwise, once `sortingEnabled` will be turned on it will refresh the feed.
In any way, the default is by popularity so it should make any different

```suggestion
      if (isSortableFeed) {
```

idoshamun: Not sure I follow. We don't save the selected sorting method locally, the user has to explicitly choose it every time so when `selectedAlgo` is changed we need to trigger our `useMemo` again. So just add it to the deps list of the `useMemo`

idoshamun (approved):
Nice catch! I see what you mean now :)

PR #823, opened 2022-01-14
fix: Sidebar transform animation was missing

rebelchris:
This ticket was intended for the sidebar showing on mobile initial load.
However this seems to be no longer existing, testing all browser options, and iPhone/android phone.

Not sure if we already fixed it.
The default setup is that it's hidden though, so makes sense it never shows.

Did notice the animation was missing from the mobile menu, so reintroduced this.

DD-436 #done


sshanzel (approved):
Awesome work on this! Can't imagine myself finding this specific class to add.

The animation is now back, though from time to time I suffer as well from the sidebar showing the animation onload.

rebelchris:
Oh you are getting it sometimes?
On what device?

Really weird as it defaults to be hidden 👀

idoshamun (approved):
For me the problem with the sidebar rarely happens but when it does it is mostly in article pages

PR #825, opened 2022-01-14
fix: removing the recent feed

rebelchris:
As we are no longer providing this option and are re-introducing how this will work we no longer need the actual recent endpoints to work.

DD-443 #done


sshanzel (approved):
Just in time with the rework of date ranking algo! Thank you for covering this up! 🙌

idoshamun (approved):

PR #827, opened 2022-01-16
fix: access safely to the flags object

idoshamun:
When I set up my new Mac and bootstrapped from scratch our development environment, I noticed that I receive critical errors on the first load. This happened when trying to access the flags object when it is still null. I use optional chaining to make sure we handle these situations correctly.


rebelchris (approved):
Nice find!

This might actually cause some edge cases that we saw on the iPad for instance.

PR #828, opened 2022-01-16
fix: track active page state correctly

idoshamun:
Our feed engine uses the client's indication of whether the session is active or not for certain decisions.
After some data analysis and debugging, I found out that there's some issue in how we track it.
The edge case opens new tab -> type URL -> navigate to the new URL will be marked as an active session.
This fix makes sure the user clicked somewhere in the window object or at least scrolled down a bit before the session was marked as active.
I also fixed an edge case where events were not sent on page unload.


rebelchris (approved):
Nice work!
Nothing really to add here, are we sure this covers all events possible defining active states?

idoshamun:
@rebelchris this is my only concern actually. But I tried to think what can indicate an active state other than a click or scroll? Also on the initial load, I check if the document is focused which covers another edge case...

sshanzel (approved):
Great catch here! Nothing to add from my side, looks neat!

PR #829, opened 2022-01-17
fix: spaces only comment

sshanzel:
DD-476 #done

On submitting a comment - it should return as soon as the input is noticed to be empty or contains spaces only.

### DIFF
--- a/packages/shared/src/components/cards/CommentPopup.tsx
+++ b/packages/shared/src/components/cards/CommentPopup.tsx
@@ -101,7 +101,7 @@ export default function CommentPopup({
           <Button
             icon={<CommentIcon />}
             onClick={() => onSubmit?.(comment)}
-            disabled={!comment?.length}
+            disabled={!comment?.trim().length}
             loading={loading}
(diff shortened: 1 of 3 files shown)


rebelchris (approved):
Solid, nice work Lee.

We could potentially have a helper for the method.
but since it's only trim now it should be fine.

Thinking ahead if we want to check multiple things.

Also weird case, but technically people could have empty elements as well: `&nbsp` or `<p></p>`.

Not sure if it's worthwhile, but maybe good to note down for a future addition?

sshanzel:
Totally agree with you. Let me create a ticket to create one method and run through our inputs in the app which we will have to ensure empty spaces are not entered.

idoshamun (approved):
Great fix! I'd love to see its counterpart on the API side that will not allow empty comments.
Please also remember to close the GH issue

sshanzel:
Merging this now. The API counterpart can be found from the link below:

https://github.com/dailydotdev/daily-api/pull/859

PR #830, opened 2022-01-17
fix: people with recent feed in storage get a error

rebelchris:
There is an edge case where people still had the `recent` feed in their local storage items.
This would cause the feed query to break.

To fix this we now delete the key, and reset the feed locally to their popular one.
I did not want to update their storage as this might be weird behaviour from us as an application, now we leave the ball in their hands.

### DIFF
--- a/packages/extension/src/newtab/MainFeedPage.tsx
+++ b/packages/extension/src/newtab/MainFeedPage.tsx
@@ -43,10 +43,14 @@ export default function MainFeedPage({
   const [searchQuery, setSearchQuery] = useState<string>();
   const [showDnd, setShowDnd] = useState(false);
   const { registerLocalFilters, shouldShowMyFeed } = useMyFeed();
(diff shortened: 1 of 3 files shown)


idoshamun (changes requested):
I suggest a different approach for this fix.
We can introduce a new custom hook that will return the default feed. It seems we have quite duplicate logic around it.
Regardless, we can also add another parameter for `usePersistentContext` which is an array of the possible values. If the returned value is not in the possible values we can fallback to the default value. What do you think?
I believe this mechanism will be helpful for other use cases

sshanzel (approved):
Wow, can't imagine people still have something related to recent feed. Lol

rebelchris:
@idoshamun Thanks! Think this looks way neater 🤘

idoshamun (approved):
Looks rad!

PR #832, opened 2022-01-17
fix: :bug: Fixed a bug in the most visited sites stack

rebelchris:
We where using a `gap-4` which caused all individual items in the most visited sites to space out.
I've decided to add separate margins to the dropdowns.

And add a items align so they are centered with the button.

Before:
image

After:
Screenshot 2022-01-17 at 15 34 43

DD-478 #done

### DIFF
--- a/packages/shared/src/components/MainFeedLayout.tsx
+++ b/packages/shared/src/components/MainFeedLayout.tsx
@@ -199,11 +199,11 @@ export default function MainFeedLayout({
   const header = (
     <LayoutHeader>
       {!isSearchOn && <h3 className="typo-headline">{feedTitles[feedName]}</h3>}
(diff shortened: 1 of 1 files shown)


sshanzel (commented):
This looks like a great fix! Though I was thinking to fix it in the container of the `MostVisitedSites` component to make it predictable and avoid future glitches since it will have its own set of rules for its container.

What do you think of it?

rebelchris:
It doesn't have a container at the moment.
But I think we can pick that up in the custom links task maybe?

sshanzel:
Yes, I mean we will have to define an actual container for it instead of a fragment.

sshanzel (approved):
That makes sense, I'll look into it since it would at least take time to finish the custom links.

idoshamun (approved):
Great fix!

PR #833, opened 2022-01-18
feat: added ranking analytics

rebelchris:
Added the ranking property to our feed extra's.
This only gets added if it's available and passed through to all the individual tracking options.

DD-481 #done

### DIFF
--- a/packages/shared/src/hooks/feed/useCommentPopup.ts
+++ b/packages/shared/src/hooks/feed/useCommentPopup.ts
@@ -7,7 +7,7 @@ import {
 } from '../../graphql/comments';
 import { apiUrl } from '../../lib/config';
 import AnalyticsContext from '../../contexts/AnalyticsContext';
-import { postAnalyticsEvent } from '../../lib/feed';
+import { feedAnalyticsExtra, postAnalyticsEvent } from '../../lib/feed';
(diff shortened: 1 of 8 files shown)


packages/shared/src/hooks/feed/useCommentPopup.ts
```diff
@@ -48,7 +51,11 @@ export default function useCommentPopup(feedName: string): {
             columns,
             column,
             row,
-            extra: { origin: 'feed', feed: feedName },
+            extra: {
+              origin: 'feed',
+              feed: feedName,
+              ...(ranking && { ranking }),
```
idoshamun, line 57: Maybe we should create a global function that generates this `extra` object for the feed-based events?
It will be easier to keep track where it is used and that we keep the right format

idoshamun (changes requested):
Looks great. I have a suggestion to improve this implementation, see below :)

packages/shared/src/hooks/feed/useCommentPopup.ts
rebelchris: Great idea, working on that

rebelchris:
@idoshamun What do you think of this approach?

idoshamun (approved):
Awesome!

PR #834, opened 2022-01-18
refactor: feed ranking algo persistence

sshanzel:
DD-482 #done 

We now persist the value for the value applied on which ranking algorithm to use.

### DIFF
--- a/packages/shared/src/components/MainFeedLayout.tsx
+++ b/packages/shared/src/components/MainFeedLayout.tsx
@@ -191,6 +191,8 @@ export default function MainFeedLayout({
   const [selectedAlgo, setSelectedAlgo, loadedAlgo] = usePersistentContext(
     DEFAULT_ALGORITHM_KEY,
     0,
+    [0, 1],
+    0,
   );
   const [selectedPeriod, setSelectedPeriod] = useState(0);
   const search = (
--- a/packages/shared/src/components/MainFeedLayout.tsx
+++ b/packages/shared/src/components/MainFeedLayout.tsx
@@ -28,7 +28,7 @@ import classed from '../lib/classed';
 import { useMyFeed } from '../hooks/useMyFeed';
 import useDefaultFeed from '../hooks/useDefaultFeed';
 import SettingsContext from '../contexts/SettingsContext';
-import usePersistentState from '../hooks/usePersistentState';
+import usePersistentContext from '../hooks/usePersistentContext';
 
 const SearchEmptyScreen = dynamic(
   () => import(/* webpackChunkName: "emptySearch" */ './SearchEmptyScreen'),
@@ -188,7 +188,7 @@ export default function MainFeedLayout({
     query = { query: null };
   }
 
-  const [selectedAlgo, setSelectedAlgo, loadedAlgo] = usePersistentState(
+  const [selectedAlgo, setSelectedAlgo, loadedAlgo] = usePersistentContext(
     DEFAULT_ALGORITHM_KEY,
     0,
   );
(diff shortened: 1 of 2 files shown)


packages/shared/src/components/MainFeedLayout.tsx
```diff
@@ -186,9 +188,11 @@ export default function MainFeedLayout({
     query = { query: null };
   }
 
-  const [selectedAlgo, setSelectedAlgo] = useState(0);
+  const [selectedAlgo, setSelectedAlgo, loadedAlgo] = usePersistentState(
```
rebelchris, line 191: This actually sounds like a perfect match for usePersistentContext 🦸‍♀️

rebelchris (changes requested):
Nice one!

I would suggest swapping to use `usePersistentContext` it's a bit more stable.
And technically we should move all implementations to it.

What do you think?

packages/shared/src/components/MainFeedLayout.tsx
sshanzel: I actually did use it before. I think there's a bug in it. It doesn't change the value even if I call the second element which should replace the value.

sshanzel:
@rebelchris I have fixed an existing issue for `usePersistentContext` - it had a mismatched key when updating. The persistent value is now utilizing the said fixed context.

rebelchris (approved):
Omg, totally missed that one!
Thanks for spotting it, we accidentally introduced it in the last release where we added the default value to the key.

This looks great now! 💖

idoshamun (approved):
💪

sshanzel:
@rebelchris the tests have failed since on load the `valueWhenCacheEmpty` is not yet utilized while processing the cache value. Kindly advise if the commit below should be safe to apply

https://github.com/dailydotdev/apps/pull/834/commits/62f99e81bc66715548a2ddf853242d69977677b6

packages/shared/src/hooks/usePersistentContext.ts
```diff
@@ -21,24 +21,27 @@ export default function usePersistentContext<T>(
...
+      },
+      onError: (_, __, rollback) => {
+        queryClient.setQueryData(queryKey, rollback);
       },
-      onError: () => {},
     },
   );
 
-  return [data, updateValue];
+  return [data ?? valueWhenCacheEmpty, updateValue, isFetched];
```
rebelchris, line 46: I actually think we should not block this.
If we simply return data it should be quick enough?

rebelchris (commented):
Don't think you have to catch it here.

packages/shared/src/hooks/usePersistentContext.ts
rebelchris: But then again, your fix would not be breaking, just causes it to re-render twice I guess.
Not quite sure why the test fail then, we use this in other places as well.

@idoshamun What do you think about the above?

sshanzel: I think it already re-renders twice since the initial value is undefined while fetching the data - then returns the actual or placeholder value.

With that, I can handle the case outside to set a value if undefined. I was also thinking if we should add an initial value parameter like how it used to be at `usePersistentState`

idoshamun: `valueWhenCacheEmpty` should be used after we validate that the cache is not presented. In this case, we actually use it as a default value which is quite different. In `usePersistentState`, we have an `initialValue` parameter that is used specifically for the purpose of a default value. So if we want to stick to the new behavior I suggest renaming accordingly. Please make sure it doesn't break any of the existing usages of this hook because it potentially can.
https://github.com/dailydotdev/apps/blob/master/packages/shared/src/hooks/usePersistentState.ts#L4

sshanzel: Cool. Let me just then add a parameter for a fallback value for when the data is yet undefined so we can be sure things will still be compatible with existing implementations.

rebelchris (approved):
Nice this looks great! 🤩

idoshamun (approved):

PR #835, opened 2022-01-18
feat: tldr to article pages

ayadotdev:
Added TLDR to the article page. When the TLDR content is more than 300 characters "Show more" link is displayed. When the content is expanded "Show less" link is displayed. When the content is less than 300 characters, "Show more" link is not displayed.

Screenshot 2022-01-18 at 15 54 43

Screenshot 2022-01-18 at 15 55 22

Screenshot 2022-01-18 at 15 59 50

DD-279 #done


packages/shared/src/components/buttons/ClickableText.tsx
```diff
@@ -21,6 +23,8 @@ function ClickableTextComponent<Tag extends AvailableTags>(
     children,
     tag: Tag = 'button',
     defaultTypo = true,
+    flexRowApplied = true,
+    underlined = true,
```
rebelchris, line 27: I think we are actually better of keeping this outside the component.
This specific spec guideline uses it, but the overal one doesn't.

You can set classNames for this specific use-case like so:

```
<ClickableText className="flex flex-row">
```

I think you might have noticed it's not inline because the default tag is a button.
You can change it to a link like so:

```
<ClickableText tag="a" />
```

packages/shared/src/components/cards/PostTLDRSummary.tsx
```diff
@@ -0,0 +1,26 @@
+import React, { ReactElement } from 'react';
+import { Post } from '../../graphql/posts';
+import { TLDRSummary } from '../utilities';
+import ShowMoreContent from './ShowMoreContent';
+
+interface PostTLDRSummaryProps {
+  post: Post;
+}
+
+export default function PostTLDRSummary({
```
rebelchris, line 10: I think PostSummary might be a better name here.

packages/shared/src/components/cards/PostTLDRSummary.tsx
```diff
@@ -0,0 +1,26 @@
...
+  post,
+}: PostTLDRSummaryProps): ReactElement {
+  return (
+    <TLDRSummary>
```
rebelchris, line 14: This might also be more inline if it's SummaryContainer or something like that?
What do you think?

packages/shared/src/components/cards/ShowMoreContent.tsx
```diff
@@ -0,0 +1,51 @@
...
+        {isTextExpanded ? content : getSlicedContent()}
+        &nbsp;
+        <ClickableText
+          className="text-theme-label-link"
+          flexRowApplied={false}
+          underlined={false}
+          onClick={toggleTextExpanded}
+        >
+          {displayShowMoreLink()}
+        </ClickableText>
```
rebelchris, line 47: It would be better to completely not render this if it's not applicable.

packages/shared/src/components/cards/ShowMoreContent.tsx
```diff
@@ -0,0 +1,51 @@
...
+    }
+    return '';
+  };
+
+  const getSlicedContent = () => {
+    return content?.slice(0, charactersLimit);
+  };
+  return (
+    <div className={className}>
+      <p className="typo-callout">
```
rebelchris, line 36: I think this should be `typo-headline` looking at the spec.

Screenshot 2022-01-19 at 09 52 23

packages/shared/src/components/fields/Dropdown.tsx
```diff
@@ -30,7 +30,7 @@ export interface DropdownProps {
 
 const getButtonSizeClass = (buttonSize: string): string => {
   if (buttonSize === 'select') {
-    return 'h-9 rounded-10 text-theme-label-primary typo-callout';
+    return 'h-9 rounded-10 text-theme-label-primary typo-body';
```
rebelchris, line 33: Why did you change this one?

packages/shared/src/components/fields/TextField.module.css
```diff
@@ -10,7 +10,6 @@
 
   &:global(.invalid) {
     box-shadow: inset 0.125rem 0 0 0 var(--theme-status-error);
-    border-color: var(--theme-status-error);
```
rebelchris, line 13: Also not sure why this one is removed here?

packages/shared/src/components/fields/TextField.tsx
```diff
@@ -120,7 +120,7 @@ export function TextField({
     >
       {compact && (
         <label
-          className="px-2 mb-1 font-bold text-theme-label-primary"
```
rebelchris, line 123: I think you might have mixed two Pr's here

packages/shared/src/components/utilities.tsx
```diff
@@ -89,3 +89,11 @@ export const customFeedIcon = 'text-2xl text-theme-label-tertiary mr-2';
...
 
 export const SummaryArrow = classed(ArrowIcon, 'icon arrow ml-auto text-xl');
+
+export const TLDRSummaryBorder = 'border-l border-theme-status-cabbage pl-4';
+
+export const TLDRSummary = classed(
+  'div',
+  'text-theme-label-secondary multi-truncate',
+  TLDRSummaryBorder,
+);
```
rebelchris, line 99: I would combine these two, don't think we need them seperate so:

```js
export const TLDRSummary = classed(
  'div',
  'text-theme-label-secondary multi-truncate border-l border-theme-status-cabbage pl-4',
);
```

rebelchris: Also this should have the top and bottom margin as Tsahi mentioned.

rebelchris (changes requested):
Really nice work Aya!
Have some remarks, but love this setup. 🥳

packages/shared/src/components/fields/common.tsx
```diff
@@ -3,7 +3,7 @@ import styles from './fields.module.css';
 
 export const FieldInput = classed(
   'input',
-  'min-w-0 text-theme-label-primary bg-transparent typo-callout caret-theme-label-link focus:outline-none',
+  'min-w-0 text-theme-label-primary bg-transparent typo-body caret-theme-label-link focus:outline-none',
```
sshanzel, line 6: Tagging changes in this file as well.

packages/shared/src/components/profile/ProfileForm.tsx
```diff
@@ -280,7 +280,7 @@ export default function ProfileForm({
         }
       />
       <div className="flex flex-col items-stretch self-stretch mb-3">
-        <p className="px-2 mb-1 font-bold text-theme-label-primary">
+        <p className="px-2 mb-1 font-bold text-theme-label-primary typo-caption1">
```
sshanzel, line 283: Tagging changes in this file as well.

packages/webapp/pages/posts/[id].tsx
```diff
@@ -243,23 +244,30 @@ const PostPage = ({ id, postData }: Props): ReactElement => {
...
           <PostHeader post={postById.post} />
-
           <a {...postLinkProps} className="cursor-pointer">
-            <h1 className="my-2 font-bold typo-title2">
+            <h1 className="my-2 font-bold typo-large-title">
               {postById?.post.title}
             </h1>
           </a>
-
+          <PostTLDRSummary post={postById.post} />
```
Comment (sshanzel) [inline packages/webapp/pages/posts/[id].tsx:270]: I think we should dynamically render this only if the `post.summary` is actually available - this would display the TLDR even summary should not be available.

packages/shared/src/components/buttons/ClickableText.tsx
```diff
@@ -8,6 +8,8 @@ export interface BaseClickableTextProps {
   pressed?: boolean;
   tag?: React.ElementType & AvailableTags;
   defaultTypo?: boolean;
+  flexRowApplied?: boolean;
+  underlined?: boolean;
```
sshanzel, line 12: These properties will violate the guidelines we have for `ClickableText`. I think this shouldn't be necessary.

sshanzel (changes requested):
Wow, super amazing you took this task. Very exciting to see this get merged.

I had quite a few comments but you are on the right track! It was mostly the changes that shouldn't be made since those were not related to the task at hand.

👏

packages/shared/src/components/buttons/ClickableText.tsx
rebelchris: Oh and as extra: The underline already is applied to the default ClickableText element on hover.

packages/shared/src/components/cards/PostTLDRSummary.tsx
ayadotdev: It makes sense 👍

packages/shared/src/components/cards/ShowMoreContent.tsx
ayadotdev: I applied `typo-headline` before but with that all text becomes bold:

Screenshot 2022-01-19 at 13 49 23

packages/shared/src/components/cards/PostTLDRSummary.tsx
```diff
@@ -0,0 +1,26 @@
+import React, { ReactElement } from 'react';
+import { Post } from '../../graphql/posts';
+import { TLDRSummary } from '../utilities';
+import ShowMoreContent from './ShowMoreContent';
+
+interface PostTLDRSummaryProps {
+  post: Post;
```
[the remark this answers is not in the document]
ayadotdev: Nice one, Lee!

packages/shared/src/components/cards/ShowMoreContent.tsx
rebelchris: Hmm, you might have to double check with Tsahi.
I think it should be 17px, but maybe not bold.

ayadotdev: There is a bold font-weight in typo-headline:
```
 '.typo-headline': {
      'font-size': '1.0625rem',
      'line-height': '1.5rem',
      'font-weight': 'bold',
    }
```

ayadotdev: Clarified with @tsahimatsliah, it should be typo-body.

packages/shared/src/components/fields/TextField.tsx
```diff
@@ -137,6 +137,7 @@ export function TextField({
           },
           styles.field,
         )}
+        style={{ ...style, borderRadius: 'rounded-10' }}
```
rebelchris, line 140: This think this is not right?
Not sure this is even needed here

packages/shared/src/components/utilities.tsx
```diff
@@ -89,3 +89,9 @@ export const customFeedIcon = 'text-2xl text-theme-label-tertiary mr-2';
 export const Summary = classed('summary', 'cursor-pointer focus-outline');
 
 export const SummaryArrow = classed(ArrowIcon, 'icon arrow ml-auto text-xl');
+
+export const SummaryContainer = classed(
+  'div',
+  'text-theme-label-secondary multi-truncate my-6',
+  'border-l border-theme-status-cabbage pl-4',
```
rebelchris, line 96: I would combine these in one line

packages/shared/src/components/cards/ShowMoreContent.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+  const toggleTextExpanded = () => {
+    setIsTextExpanded(!isTextExpanded);
+  };
+
+  const displayShowMoreLink = () => {
+    if (
+      content &&
+      content?.length > charactersLimit &&
+      content?.length - charactersLimit > threshold
```
rebelchris, line 30: Think we can do this in one go:

```js
content?.length > charactersLimit + threshold
```

rebelchris (changes requested):
Really nice work!
Some small things, almost there 🥳

packages/shared/src/components/fields/TextField.tsx
ayadotdev: Weird, I removed it yesterday.

rebelchris (approved):
Super nice work you did here! 🎉
Love the tests

packages/shared/src/components/cards/ShowMoreContent.tsx
```diff
@@ -0,0 +1,53 @@
...
+  const toggleTextExpanded = () => {
+    setIsTextExpanded(!isTextExpanded);
+  };
+
+  const displayShowMoreLink = () => {
+    if (content && content?.length > charactersLimit + threshold) {
+      return true;
+    }
+    return false;
+  };
```
idoshamun, line 31: It looks cleaner in my opinion

```suggestion
  const displayShowMoreLink = () => content && content?.length > charactersLimit + threshold;
```

packages/shared/src/components/cards/ShowMoreContent.tsx
```diff
@@ -0,0 +1,53 @@
...
+  threshold = 50,
+  contentPrefix,
+  className,
+}: ShowMoreContentProps): ReactElement {
+  const [isTextExpanded, setIsTextExpanded] = useState(false);
+  const linkName = isTextExpanded ? 'Show less' : 'Show more';
+
+  const toggleTextExpanded = () => {
+    setIsTextExpanded(!isTextExpanded);
+  };
```
idoshamun, line 24:
```suggestion
  const toggleTextExpanded = () => setIsTextExpanded(!isTextExpanded);
```

packages/shared/src/components/cards/ShowMoreContent.tsx
```diff
@@ -0,0 +1,53 @@
...
+  const displayShowMoreLink = () => {
+    if (content && content?.length > charactersLimit + threshold) {
+      return true;
+    }
+    return false;
+  };
+
+  const getContent = () => {
+    return isTextExpanded ? content : content?.slice(0, charactersLimit);
+  };
```
idoshamun, line 35:
```suggestion
  const getContent = () => (isTextExpanded ? content : content?.slice(0, charactersLimit));
```

packages/shared/src/components/cards/ShowMoreContent.tsx
```diff
@@ -0,0 +1,53 @@
...
+  return (
+    <div className={className}>
+      <p className="typo-body">
+        {contentPrefix}
+        {getContent()}{' '}
+        {displayShowMoreLink() && (
+          <ClickableText
+            className="inline-flex text-theme-label-link"
+            onClick={toggleTextExpanded}
+            tag="a"
```
idoshamun, line 45: Why is that a `a` element? I think the `tag` property can be dropped

packages/shared/src/graphql/posts.ts
```diff
@@ -103,6 +104,7 @@ export const POST_BY_ID_QUERY = gql`
         username
       }
       description
+      summary
```
idoshamun, line 107: We need to add this to `POST_BY_ID_STATIC_FIELDS_QUERY` so it will be cached for the next visits.
In the article page itself you can find this function `getStaticProps` that fetches the necessary data for the page before it arrives to the client and then we revalidate on the client side as well. I highly suggest reading about nextjs's incremental static regeneration for better understanding what's going on under the hood.

@rebelchris @sshanzel  this can be another valid point for future reviews in pages that use ISR or SSR.

idoshamun (changes requested):
Some cosmetics suggestion and an important change.

Overall this looks solid and this is also quite advanced, it's so great to see your progress from day to day. 💪

packages/shared/src/components/cards/ShowMoreContent.tsx
rebelchris: I'm pretty sure this should render as a <a> not a <button> right?
Since it's an inline anchor element.

The button would also push it to it's own line.

idoshamun: Hmm interesting. The second argument could be easily fixed but the first one is interesting.
I think anchors should only be links and this one is actually a button. So in terms of accessibility and semantics, I do think it's a button but you made me somewhat question it. Could you check it, please?

rebelchris: Screenshot 2022-01-20 at 13 42 28

To me this is really a <a href=""> not a <button>

But let me know if you think otherwise

rebelchris: But according to A11Y rules you are 100% right and it should be a button (my bad)
Screenshot 2022-01-20 at 13 44 34

rebelchris: Reading a bit more, there seems to be quite the split thinking about these and some field experts even say it should not even exist in this way.

https://blog.prototypr.io/youre-not-still-using-read-more-are-you-16466ae326e7

idoshamun: Let's go with a button then :)

sshanzel (approved):
LGTM! 🚀

packages/shared/src/graphql/posts.ts
rebelchris: On this topic, I get the difference, but why can't they use the same GraphQL endpoint?

idoshamun: Because some of the properties depend on the logged-in user, this will not be available and might even throw an exception. We can try that of course.

idoshamun (approved):
Let's ship!

PR #836, opened 2022-01-18
fix: cleanup tailwind classes

rebelchris:
Since the Tailwind `break-words` is broken by design, we decided to take hands in own matters and override those to make them do what they promise.

Giving us the `break-words` class to break words and `break-words-overflow` to define the overflow for breaking words.

I also took the liberty in this PR to clean-out no longer needed Tailwind classes for `transform`.

And found a couple weird edge cases I fixed:

1. When posts had the hot filter they did not match the full height:
Screenshot 2022-01-18 at 16 42 23

(Fixed by adding the h-full to the card)
Related the z-index was off, so decided to force to `-1` to always be behind the cards
 
2. Profile bio had no overflow:
Screenshot 2022-01-18 at 17 03 26

Fixed by adding `break-words`
Screenshot 2022-01-18 at 17 04 13

### DIFF
--- a/packages/shared/src/components/InteractionCounter.tsx
+++ b/packages/shared/src/components/InteractionCounter.tsx
@@ -34,15 +34,13 @@ export default function InteractionCounter({
     <span className={`relative overflow-hidden ${styles.interactionCounter}`}>
       <span
         className={
-          animate
-            ? 'transform opacity-0 -translate-y-full'
-            : 'transform opacity-100 translate-y-0'
+          animate ? 'opacity-0 -translate-y-full' : 'opacity-100 translate-y-0'
         }
       >
         {shownValue}
       </span>
       <span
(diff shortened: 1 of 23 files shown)


sshanzel (approved):
Wow. Thank you so much for cleaning up all the unnecessary classes. This makes it much easier to understand and make changes with the styles.

idoshamun (approved):
I think there is a reason the flag's z-index was set to 0. Please double check locally that everything works including hovering the flag and the tooltip

rebelchris:
@idoshamun As a double test:

Chrome:
Screenshot 2022-01-19 at 09 07 54

Safari:
Screenshot 2022-01-19 at 09 04 58

Firefox:
Screenshot 2022-01-19 at 09 06 59

Tooltip hover:
Screenshot 2022-01-19 at 09 07 34

All seems to be working well.

PR #837, opened 2022-01-19
build: webpack5 upgrade

rebelchris:
Upgraded to webpack5 🥳

DD-214 #done


idoshamun (approved):
🤖

rebelchris:
Small hold on extension dev breaking (build is ok, but dev fails)

rebelchris:
Some background on the latest commit.

The build was going well as it doesn't use this plugin, so for dev we use a plugin called `webpack-extension-reloader`.

The sole purpose is to reload the extension.

The original registry: https://www.npmjs.com/package/webpack-extension-reloader
Problem: Last update 2 years ago, making it incompatible with Webpack 5 (due to `webpack-sources` using a wrong chunking method)

Some people are trying to publish fixed to it, but the owner seem non responding.
https://github.com/rubenspgcavalcante/webpack-extension-reloader/pull/128

Then I found another company who forked it, and took over maintenance:
https://github.com/SimplifyJobs/webpack-ext-reloader/

It seems to at least have gotten the update to webpack5 which solves our dev problem.

As this is only used in dev, I don't think it's breaking change, but wanted you guys to be informed around the decision here.

CC: @idoshamun @sshanzel

packages/shared/package.json
```diff
@@ -20,9 +20,9 @@
     "graphql": "^15.5.0",
     "graphql-request": "^3.4.0",
     "idb-keyval": "^5.0.6",
-    "next": "^10.2.3",
-    "react": "^17.0.5",
-    "react-dom": "^17.0.5",
+    "next": "^11.1.3",
+    "react": "^16.0.0",
+    "react-dom": "^16.0.0",
```
sshanzel, line 25: Why did we have to downgrade our react packages?

sshanzel (approved):
This is just as cool as it can be! 🚢 

Just had one non-blocking question.

packages/shared/package.json
rebelchris: I think this might be a upgrade that went wrong as 17.0.5 doesn't even exist.

Screenshot 2022-01-20 at 09 25 17

rebelchris: But upgraded it to the latest version now

PR #840, opened 2022-01-20
feat: virtual grid removal

rebelchris:
We decided to move away from the virtual grid for now.
It proved to give a lot of weird edge-cases of some where not solvable.

This PR is removing the virtual grid and making use of CSS-grid for the rendering.
I've tried to keep the analytics in tact by mimicking the col/row and gap collection.

DD-367 #done 

Note>: We might want to run the preview deployment for the team to see if anyone has weird behaviour on this one.


rebelchris:
Hmm seems Tailwind can't inject the classes since they are dynamic.
Can we define them as always render classes, or what would be the best solution here?

I made sure they were fully qualified and not partial dynamic classes.

idoshamun:
@rebelchris that's a weird behavior, maybe this file is not in the `purge` settings of tailwind? Can you confirm, please?

rebelchris:
@idoshamun Yep they are the same as before.

```js
module.exports = {
  ...config,
  content: [
    './pages/**/*.{js,ts,jsx,tsx}',
    './components/**/*.{js,ts,jsx,tsx}',
    './node_modules/@dailydotdev/shared/src/**/*.{js,ts,jsx,tsx}',
  ],
};
```

Note in Tailwind 3 it's renamed to `content` what was `purge` before.
The file is: `src/components/Feed.tsx` so it would hit.

idoshamun:
@rebelchris could be a bug perhaps? Let's add it to the allow list for now and add a comment that explains it :)

rebelchris:
@idoshamun the purge issue is solved in this version.
With the other layout shift fix, this version should be ready to go 👀

idoshamun (approved):
Let's do it!

PR #843, opened 2022-01-20
feat: custom links

sshanzel:
DD-85 #done

Kindly note that I have made some changes in the `TextField` utilizing the `fieldType` property - after checking the guideline there seem to be quite a few changes necessary with the field itself. I will open a ticket to further continue on other types (primary and secondary) to be in line with the actual guideline since a lot of existing implementations must be checked carefully.

Note: this is only available for extension so this has to be pulled locally to make it work.


sshanzel:
Whew, this was one of the most comprehensive tests I have written, I think.

Should now be complete from end to end and ready for review.

rebelchris (changes requested):
Hey Lee,
First of all, super well done on this and amazing work on those test cases! 🤩

Generic feedback:

I'm really not a fan of our determination to show the custom links, see comment I made in the DR as well.

Not sure if this auto fill is what we want?
It might confuse people, but maybe it's as intended
Screenshot 2022-01-21 at 13 33 42

Perhaps check with Nimrod, but something like "Manage top sites" would be nicer than options?
Screenshot 2022-01-21 at 13 35 05

It seems to reset my choice on every load, might be local but please verify.
This in return will reset my own set sites everytime I toggle, making it super hard to manage

packages/extension/src/newtab/ShortcutLinksModal.tsx
```diff
@@ -0,0 +1,107 @@
...
+      padding={false}
+      contentClassName="max-h-[40rem]"
+      style={{ ...style, content: { maxHeight: '40rem' } }}
+    >
+      <form ref={formRef} onSubmit={onSubmit}>
+        <ModalHeader>
+          <h3 className="font-bold typo-title3">Shortcuts</h3>
+          <div className="flex-1" />
+          <Button className="mr-3 btn-primary" buttonSize="small" type="submit">
+            save changes
```
ayadotdev, line 57: This is my first very useful comment 😅  I noticed a lower case letter in "Save changes".

sshanzel: Great spot! But that was actually part of the design spec 😅

So happy to see you take part in the review!

packages/extension/src/newtab/CustomLinks.tsx
```diff
@@ -0,0 +1,47 @@
...
+        <a
+          href={url}
+          rel="noopener noreferrer"
+          className={classNames(
+            'focus-outline w-8 h-8 rounded-lg overflow-hidden bg-white',
+            i >= 4 && 'hidden laptopL:block',
+          )}
+          key={url}
+        >
+          <img
```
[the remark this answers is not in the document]
sshanzel: Whenever the site's image is not available it falls back to the first letter of the site.

Screen Shot 2022-01-21 at 10 55 03 PM

packages/extension/src/newtab/CustomLinks.tsx
```diff
@@ -0,0 +1,47 @@
...
+}: CustomLinksProps): ReactElement {
+  return (
+    <div className="flex flex-row gap-2 p-2 rounded-14 border border-theme-divider-secondary">
+      {links.map((url, i) => (
+        <a
+          href={url}
+          rel="noopener noreferrer"
+          className={classNames(
+            'focus-outline w-8 h-8 rounded-lg overflow-hidden bg-white',
+            i >= 4 && 'hidden laptopL:block',
```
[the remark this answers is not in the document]
sshanzel: It was part of the old code, probably to hide it on smaller screen sizes as I see. Also that there was no mobile design.

packages/extension/src/newtab/CardSelection.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+export function CardSelection({
+  title,
+  description,
+  icon,
+  isActive,
+  onClick,
+}: CardSelectionProps): ReactElement {
+  return (
+    <button
```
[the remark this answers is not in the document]
sshanzel: `<Button>` has a pre-defined set of values for its size and there's currently no available square box as large as this.

Good point though, do you have anything in mind we should switch to? I previously used `div` but `button` felt more fitting.

packages/extension/src/newtab/ShortcutLinksModal.tsx
ayadotdev: Then I am sorry! I'll double check the next time 😄

packages/extension/src/newtab/CardSelection.tsx
```diff
@@ -0,0 +1,47 @@
...
+        'relative flex flex-col gap-1 py-3 px-4 items-center border border-theme-divider-tertiary rounded-16 hover:cursor-pointer',
+        isActive
+          ? 'border-theme-divider-primary'
+          : 'border-theme-divider-tertiary',
+      )}
+      onClick={onClick}
+      type="button"
+    >
+      {isActive && (
+        <Button
```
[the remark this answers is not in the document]
sshanzel: I think the `iconOnly` shouldn't be necessary anymore since the deciding factor is the computed value inside (also I think the prop is not being used anywhere).

https://github.com/dailydotdev/apps/blob/8479fcbc05149818870eca7ef5b4f6be32b87d7f/packages/shared/src/components/buttons/Button.tsx#L58

packages/extension/src/newtab/ShortcutLinksModal.tsx
sshanzel: Actually, I just checked and I was wrong! It should have been on caps! You were right! My bad! Thank you for the additional pair of eyes.

packages/extension/src/newtab/CardSelection.tsx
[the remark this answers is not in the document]
sshanzel: `<Button>` has a pre-defined set of values for its size and there's currently no available square box as large as this.

I realized the icon should only be an icon. I've removed the button on its implementation and just displayed the icon itself.

sshanzel:
Kindly note that I have made some changes in the `TextField` utilizing the `fieldType` property - after checking the guideline there seem to be quite a few changes necessary with the field itself. I will open a ticket to further continue on other types (primary and secondary) to be in line with the actual guideline since a lot of existing implementations must be checked carefully.

This should now be ready for review - cc @rebelchris @ayadotdev @idoshamun

packages/shared/icons/sites.svg
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
rebelchris, line 1: Non breaking, but Tashi has a way to clean this up so it's smaller.
We might have to ask him at one point to do all icons in the minimized way.

Could you add a ticket so we don't forget?

rebelchris (approved):
Very happy with this feature.
Some non breaking comments 🥳

packages/shared/src/components/fields/TextField.module.css
```diff
@@ -3,6 +3,11 @@
     box-shadow: inset 0.125rem 0 0 0 var(--theme-label-primary);
   }
 
+  &:global(.focused.readOnly) {
+    box-shadow: inset 0.125rem 0 0 0 var(--theme-focus);
+    border-color: var(--theme-focus);
+  }
```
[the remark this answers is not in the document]
sshanzel: I think this should be across the as part of the guideline but let me confirm with Tsahi first. Thanks for pointing that out!

packages/shared/src/components/fields/TextField.tsx
```diff
@@ -132,14 +141,15 @@ export function TextField({
...
+            readOnly,
             focused,
             invalid,
           },
           styles.field,
         )}
       >
         <div className="flex flex-col flex-1 items-start max-w-full">
-          {!compact && (
+          {!compact && fieldType !== 'tertiary' && (
```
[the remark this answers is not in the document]
sshanzel: Totally agree with it, I created one ticket dedicated for the migration of this component. Can be found here: https://dailydotdev.atlassian.net/jira/software/projects/DD/boards/1?selectedIssue=DD-495

packages/shared/icons/sites.svg
sshanzel: Would love to! Let me stir one up! 🚀

packages/shared/src/components/fields/TextField.module.css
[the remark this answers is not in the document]
sshanzel: Tsahi has confirmed the style should be applied across the types. This should now be safe to be merged 🚢

packages/extension/src/newtab/LinksForm.tsx
```diff
@@ -0,0 +1,56 @@
...
+      return;
+    }
+
+    setStaleInput((state) => ({ ...state, [i]: false }));
+  };
+
+  return (
+    <div className="flex flex-col gap-4">
+      {list.map((_, i) => (
+        <TextField
```
idoshamun, line 39: Why don't we apply for any local check as well?
Are we relying only on server-side validation?

EDIT: Giving it another look, I see we have local validations but we can use the built-in input type for doing this validation. I think it will be better than a js based solution

idoshamun (changes requested):
Open question about client-side validation.
Also it feels that there's a lot going on and it wasn't trivial for me to read through the code. Unfortunately, there's nothing I can highlight specifically. So what I suggest is you giving the PR a second read and if you find anything that can improved go for it please.

packages/extension/src/newtab/LinksForm.tsx
sshanzel: Was thinking of not fully managing every field (callback and checks) with a separate state of their own since all of their validations will run through the same check and hints so I decided to create an object instead.

I tried checking the `ProfileForm` and it handles each input's validity separately, should we do the same for this scenario?

sshanzel (commented):
@idoshamun thank you for the review and pointing out to use the built-in validation - I have now made the necessary changes.

Also ran through and cleared up some unnecessary functions and files 🚢

packages/extension/src/newtab/ShortcutLinksModal.tsx
```diff
@@ -49,11 +45,16 @@ export default function CustomLinksModal({
...
+      <form ref={formRef}>
         <ModalHeader>
           <h3 className="font-bold typo-title3">Shortcuts</h3>
           <div className="flex-1" />
-          <Button className="mr-3 btn-primary" buttonSize="small" type="submit">
+          <Button
+            className="mr-3 btn-primary"
+            buttonSize="small"
+            type="button"
+            onClick={(e) => onSubmit(e)}
```
idoshamun, line 56: I think the previous implementation was preferred. That you listened to the `onSubmit` of the form. More accessible. Why did you decide to change?

idoshamun (changes requested):

packages/extension/src/newtab/ShortcutLinksModal.tsx
sshanzel: I agree the previous one was better - though I encountered a few issues with it, I have found a workaround to keep `onSubmit`.

sshanzel:
@idoshamun I have now pushed, what I think was better in terms of handling the validation than before. Also now utilizes the `onSubmit` of the actual form. Thank you!

idoshamun (approved):
Please confirm your changes locally before releasing a version

sshanzel:
Have done local QA from end to end with a build production. Will merge now and create a release to be sent in slack 🚢

PR #844, opened 2022-01-20
fix: cut summary when not reaching threshold

ayadotdev:
Fixing a bug when the summary is cut if not reaching threshold (300-350 characters).


idoshamun (approved):
🙌🍻

rebelchris (approved):
Super nice that you included tests for this 🙌🙌

PR #845, opened 2022-01-21
fix: :bug: The ranks modal was not scrollable on desktop

rebelchris:
The ranks modal was not scrollable on desktop screen with small heights.

This made it impossible for people to reach the "generate" button.

I've added a wrapper on the content part, as we don't want the rank symbol to scroll.
The fix works only on desktop size as the mobile version already had a scrollbar.

Desktop:
Screenshot 2022-01-21 at 10 46 08

Mobile:
Screenshot 2022-01-21 at 10 46 21

DD-487 #done


sshanzel (approved):
LGTM! 🚀

packages/shared/src/components/modals/RanksModal.tsx
```diff
@@ -214,29 +214,35 @@ export default function RanksModal({
...
         </div>
       )}
-      <h1 className="mt-4 mobileL:mt-11 font-bold uppercase typo-title2">
-        Your weekly goal
-      </h1>
-      <h2 className="mt-1 font-normal text-theme-label-secondary typo-callout">
-        Read content you love to stay updated
-      </h2>
-      <TimezoneText onShowAccount={onShowAccount} />
+      <div className="flex overflow-y-auto flex-col items-center responsiveModalBreakpoint:max-h-[calc(100vh-5rem)]">
```
idoshamun, line 217: `responsiveModalBreakpoint:max-h-[calc(100vh-5rem)]` feels a bit too much for me. Don't you think?
We might want to think about a more profound solution for modals to improve their DX.
So the modal will be scrollable out of the box without having to think about it.

idoshamun (changes requested):

packages/shared/src/components/modals/RanksModal.tsx
rebelchris: The other modals actually are, but since this one has a different layout we can't use it.
It would break the ranks badge outside of it.

Can have another look to see what we can come up with.

idoshamun: In this case, that's perfectly fine. I'd just love to hear what you think about the long JIT class that I mentioned above.

rebelchris: It is a bit long, but not sure how we could improve on it, as we need the dynamic height.
It also doesn't help `responsiveModalBreakpoint:` is quite long compared to `laptop:` for instance.

Do you have any suggestion on improving it?
Perhaps a definition of the max h as the ranks modal so it becomes:

```js
responsiveModalBreakpoint:max-h-rank-modal
```

idoshamun: Yes, this would be much better for example. Otherwise, we can use CSS modules as well.

idoshamun (approved):
Sweet!

PR #846, opened 2022-01-21
refactor: summary content sliced by words

ayadotdev:
Added a function to split TLDR summary content by words instead of characters.


sshanzel (approved):
Just as I would have imagined! 🚢

rebelchris (approved):
Nice work!

idoshamun (changes requested):
Why do we use JavaScript for the truncation? Maybe it's better to use a CSS alternative that can handle it correctly like the (text-overflow)[https://www.w3schools.com/cssref/css3_pr_text-overflow.asp] property.

ayadotdev:
> Why do we use JavaScript for the truncation? Maybe it's better to use a CSS alternative that can handle it correctly like the [text-overflow](https://www.w3schools.com/cssref/css3_pr_text-overflow.asp) property.

It looks like `text-overflow` property hides the content based on the container width which makes it difficult to control (display the last entire word). 

Another simplified and more readable/maintainable approach, in my opinion, is to treat the content string as an array and simply control the amount of words to display (also both limit and threshold refer to words and not characters).

```
const contentWords = content.split(' ');

const getSlicedContent = () => contentWords.slice(0, wordsLimit).join(' ');

const displayShowMoreLink = () =>
    contentWords.length > wordsLimit + wordsThreshold;

const getContent = () =>
    isTextExpanded || !displayShowMoreLink() ? content : getSlicedContent();
```
WDYT?

@idoshamun @rebelchris @sshanzel

idoshamun:
Ok, I understand your concern regarding the `text-overflow` property. As for the actual splitting, I think by default we should split by word as you mentioned above but also need to check for the result's length. Because sometimes there could be very long words that will cause the short version to be much longer than expected. For example, if the TLDR includes code snippets without spaces or a URL, etc...

ayadotdev:
I discussed this with Chris @rebelchris and I think maybe it is better to leave the function like it is for now as, otherwise, I am afraid we can add unnecessary complexity. Let me know if it makes sense for you as well @idoshamun

idoshamun:
@ayadotdev ok then!

idoshamun (approved):

PR #848, opened 2022-01-21
fix: feed layout shift

rebelchris:
Hey Ido,

A lot of the layout shift was due to the ads loading, in certain cases it would occur the isFetching was false, but on the next re-render it would re-evaluate and start re-adding the ad.

This version has no flicker at all, but the negative is that there is a possibility for the empty placeholders to show if there is no ad for the spot.

You can try it out here:
https://daily-webapp-9sh1bbnp5-dailydotdev.vercel.app/

Not sure if this is the most reliable, but wanted to try it out at-least to see what happens.

DD-485 #done


rebelchris:
@idoshamun @sshanzel Would love for you guys to try this out and let me know what you think about the responsiveness of the feed?

So the layout shift off the cards, should be fully static now.

idoshamun (approved):
I'm definitely OK with the edge case you present. Fix looks solid!

PR #850, opened 2022-01-24
feat: refactor theme colors

rebelchris:
Introducing all the theme colors as options for the AB testing. 
I've also modified the way this is handled to have a fallback/default value when a color is not found.

Color that can be used:

-   avocado
-   burger
-   blueCheese
-  lettuce
- cheese
- bun
- ketchup
- bacon
- cabbage
- onion
- water
- primary

DD-491 #done


packages/shared/src/components/sidebar/MyFeedButton.tsx
```diff
@@ -80,9 +48,15 @@ const UnfilteredMyFeedButton = ({
   const { trackEvent } = useContext(AnalyticsContext);
 
   const buttonCopy = getFeatureValue(Features.MyFeedButtonCopy, flags);
-  const buttonColor = getFeatureValue(Features.MyFeedButtonColor, flags);
+  const buttonColor = getThemeColor(
+    getFeatureValue(Features.MyFeedButtonColor, flags),
+    'primary',
```
idoshamun, line 53: Don't we want to use the default value of the feature instead of falling back to `primary`? I think it makes more sense

idoshamun (changes requested):
Looks great I do think we need to change the name of the actual flag name in Flagsmith of both features. Otherwise, we might crash older versions that don't support what you introduce in this PR. You can delete the old flags that you created

rebelchris:
Thanks @idoshamun Refactored the flagsmith keys and re-added them in flagsmith, with all options defined already.
Also reworked how we load them.

idoshamun (approved):
Looks great!

PR #851, opened 2022-01-25
hotfix: ad load

rebelchris:
We had an issue where the ad infinite query would not load on time.
this introduces a useEffect way of reloading the ads if they take longer than the next feed page.


sshanzel (approved):
Great fix! 🚢

PR #853, opened 2022-01-26
chore: the docker postgres images was outdated

rebelchris:
People where not able to spool up the apps docker image.
As mentioned before, might be worthwhile moving away from it at one stage, or rework how we want people to use it.

We don't actively maintain this image and every change on the API/Gateway can break this behaviour.

For now this fix will make sure it's up to date again.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -13,16 +13,6 @@ services:
     ports:
       - "6379:6379"
 
-  daily-postgres:
-    image: postgres:11.6-alpine
-    ports:
-      - "5432:5432"
-    volumes:
-      - db:/var/lib/postgresql/data
-    environment:
-      - POSTGRES_DB=api
-      - POSTGRES_PASSWORD=12345
-
   daily-mysql:
     image: mariadb
     environment:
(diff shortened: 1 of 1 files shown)


idoshamun (commented):
It seems that we use the same image. I'm not sure what is the change except the volumes. If the problem is the image so nothing changed there. The change you present here makes it create two databases `api` and `api_test` instead of one

rebelchris:
Not sure what the issue was, but I generally just make these inline with the docker configs for the individual projects as we use them.

The postgres was failing, so thought we might have introduced the second database and it failed on that?

idoshamun:
I haven't checked if it fails but if you confirmed locally that it works than we're all set and I can approve it

rebelchris:
@idoshamun Yep, ran the previous version, fails on Postgres, made these changes and it spools up correctly.

idoshamun (approved):

tevez07b9:
@rebelchris , I took the latest pull form master but both api and postgres containers keep crashing
image

Please help

jonatan5524:
> @rebelchris , I took the latest pull form master but both api and postgres containers keep crashing image
> 
> Please help

experiencing the same behavior  Permission denied, daily-postgres crashes:
image

System: Ubutnu 20.04.3 LTS
Docker: Docker version 20.10.13
Docker-compose: docker-compose version 1.29.2

rebelchris:
@tevez07b9 @jonatan5524 I've found and solved a small bug in our docker setup.
Would you mind trying it again?

Try to pull the latest master version first.

Since it failed you'll have to full down everything with the following command:

```bash
docker-compose down --volumes
```

Then re-up them:

```bash
docker-compose pull && docker-compose up
```

It might occur the gateway fails on the first boot, you can manually start that one again once everything else is up.

tevez07b9:
Everything works, thank you so much @rebelchris

jonatan5524:
> @tevez07b9 @jonatan5524 I've found and solved a small bug in our docker setup. Would you mind trying it again?
> 
> Try to pull the latest master version first.
> 
> Since it failed you'll have to full down everything with the following command:
> 
> ```shell
> docker-compose down --volumes
> ```
> 
> Then re-up them:
> 
> ```shell
> docker-compose pull && docker-compose up
> ```
> 
> It might occur the gateway fails on the first boot, you can manually start that one again once everything else is up.

Worked! thank you.

PR #855, opened 2022-01-26
fix: page container placement

sshanzel:
DD-435 #done
DD-385 #done

After further checking the previous implementations, I realized the previously defined margin on left was used to keep the content at the absolute center. With that, we now have a sidebar which means the static margin won't fit anymore so we have to calculate how much padding there is and the actual width of the sidebar itself - whether it is in expanded mode or not. The content push was a result of a race on rendering the components when the Shareable links are not yet available, the main content consumes the right side as well - resulting in to push to the left when the other component becomes available.

This fixes the profile page and the other pages affected by the failing tests since those pages utilize the `PageContainer` which now is at the absolute center all the time.

Also, this PR improves the UX for when the user minimizes or expands the sidebar - it doesn't re-adjust anymore.

Quicklinks:
https://preview.app.daily.dev/posts/V4219b1HY
https://preview.app.daily.dev/palant

Before:
https://user-images.githubusercontent.com/13744167/151149642-21c3bce7-ed90-4ea2-bcfc-9d7060d46ccb.mov

This PR:
https://user-images.githubusercontent.com/13744167/151149459-b851f66d-7c78-4762-b5d3-7e3cff782341.mov


rebelchris (changes requested):
Not sure this is really the right fix.
I assumed it should always be centered depending on the sidebar state? We might have to check with Tsahi.

Also now the profile page is never centered:
Screenshot 2022-01-26 at 16 01 26

And it can even only be half visible:
Screenshot 2022-01-26 at 16 01 15

Same for article (also noted another issue now, see the hot line)
Screenshot 2022-01-26 at 16 04 43

sshanzel:
@rebelchris how do these cuts happen? Is it when you resize the browser or when you load having that size already?

rebelchris:
@sshanzel By resizing indeed, it seems just around tablet size.

sshanzel:
@rebelchris thank you for reporting the concern, there was indeed a glitch - I have fixed it on the profile page, let me work on the article page first thing tomorrow.

Also, Tsahi has agreed we should now display the content in the screen's absolute center.

rebelchris:
@sshanzel I still have a edge case where about 10px are overlapping (And see the right side, it's cut off)
Screenshot 2022-01-27 at 09 28 00

sshanzel:
@rebelchris thanks for the feedback! Looking at it, it made me go back to the design spec and realized on the laptop - the widgets should already be placed below, I have applied the necessary styles based on screen size.

Should now be as what is intended.

laptopL: https://www.sketch.com/s/2867ec63-0966-4647-b1d7-b6fe696ba251/a/qeva7Me#Inspector
laptop: https://www.sketch.com/s/2867ec63-0966-4647-b1d7-b6fe696ba251/a/L0x2Me4#Inspector

rebelchris:
@sshanzel The breaking is fine now
Pretty sure there is a 1 or 2px overlap, but really not worth checking out.

I did however see some new edge cases of elements being pushed wrong for the article page.
Screenshot 2022-01-27 at 13 17 38

Top is pushed together and bottom has no margin

sshanzel:
@rebelchris the top padding should've been fixed by this: https://github.com/dailydotdev/apps/pull/855/commits/b83dc6fa6fc02f65c28ef68f32fb45ab330df46c probably got late in deployment by the time you checked. Let me check the below padding on my local.

sshanzel:
@rebelchris both sides should now have their paddings! Thank you for the sharp eyes! 🔭

rebelchris:
@sshanzel Well done, these two look great!

Did realise there are more pages we should fix that use this PageContainer 👀

- https://preview.app.daily.dev/history
- https://preview.app.daily.dev/dsdsds
- https://preview.app.daily.dev/register

sshanzel:
Can't believe I ended up with a much simpler solution now. I think this is much more solid than before.

@rebelchris thank you for the feedback! Every page now should automatically be centered.

rebelchris (approved):
Nice work!

Lovely and simple solution!
And we got to remove a obsolete once off container 🥳

idoshamun (approved):
Please validate that there is no regression before your merge.
This kind of changes has a potential to break completely the layout.
Make sure you test out all affected pages in all resolutions, otherwise we can't be sure

sshanzel:
Have checked each and every page. I have now fixed the missed out bookmarks page to not be centered 🎉

PR #857, opened 2022-01-27
fix: :bug: Fix for the hiding of the read history

rebelchris:
## Changes

- We changed the timezone for reading history to be reflecting the users timezone
- We delete based on a UTC timezone, so we need to pass along that value instead

## Testing

N/A

## Related

https://github.com/dailydotdev/daily-api/pull/865

DD-498 #done


sshanzel (approved):
Nice catch!

idoshamun (approved):

PR #858, opened 2022-01-27
refactor: text field guideline update

sshanzel:
## Changes

The primary change to look for is the variation of the font colors based on which `fieldType` the input is. The `tertiary` is a new addition that is used already in the custom links so there are no pre-existing implementations of it.

The `compact` property is now gone and was replaced by being the `secondary` field type.

Primary is the common and default implementation of the text field.

This now supports icons inside the text field itself and the padding varies between having one or not.

## Testing

- Did you introduce testing for this PR
- ☑️ Did you test the webapp? **Yes**
- ☑️ Did you test the extension? **Yes**

DD-495 #done

**Note:** just waiting for Tsahi's answer regarding the right icon but this should be alright to be reviewed.

### DIFF
--- a/packages/shared/src/components/fields/TextField.spec.tsx
+++ b/packages/shared/src/components/fields/TextField.spec.tsx
@@ -57,7 +57,7 @@ it('should set hint role as alert when invalid', async () => {
   await waitFor(() => expect(el).toHaveAttribute('role', 'alert'));
 });
 
-it('should show both label and placeholder in compact mode', async () => {
+it('should show both label and placeholder in secondary mode', async () => {
   renderComponent({ fieldType: 'secondary', placeholder: 'Placeholder' });
   const input = getInput();
   await waitFor(() => expect(input.placeholder).toEqual('Placeholder'));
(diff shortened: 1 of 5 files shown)


packages/shared/src/components/fields/TextField.spec.tsx
```diff
@@ -58,7 +58,7 @@ it('should set hint role as alert when invalid', async () => {
 });
 
 it('should show both label and placeholder in compact mode', async () => {
```
rebelchris, line 60: Can we rename this to `secondary` mode? (The actual test)

rebelchris, line 60: Also can we add a specific test for this new one?

packages/shared/src/components/fields/TextField.tsx
```diff
@@ -52,6 +56,9 @@ export function TextField({
     onInput: baseOnInput,
     focusInput,
   } = useInputField(value, valueChanged);
+  const isPrimary = fieldType === 'primary';
+  const isSecondary = fieldType === 'secondary';
+  const isTertiary = fieldType === 'tertiary';
```
rebelchris, line 61: Could we maybe make this a bit more explicit?

rebelchris (changes requested):
Very nice setup!
Would only state we could add some more tests around the specific types?

Perhaps we can add to check for certain classes?
This will quickly show us if we ever move one by accident

What do you think?

sshanzel:
@rebelchris thank you for the review! Totally agree with it! This is a reused component across the app, tests must be very strict in this scenario.

rebelchris (approved):
Nice work on this

idoshamun (commented):
Awesome!

PR #859, opened 2022-01-31
refactor: pull request template description

sshanzel:
Just fixed a typo and made the checkbox actually require the requestor to tick. My bad I think I was not clear about the checkbox and what the requestor needs to keep in mind when creating one.


rebelchris (approved):
Nice!

PR #860, opened 2022-01-31
feat: add weekly goals toggle in customize settings

ayadotdev:
## Changes

This PR adds the new toggle for Weekly goals widget in Customize settings. It is enabled by default and is toggable only for logged in users.

Untoggled:
Screenshot 2022-01-31 at 11 58 32

Toggled:
Screenshot 2022-01-31 at 11 58 50

## Testing

Added the unit tests as well.

Comment: for the other toggles only the capital letter is in upper case. "Show Weekly goal widget" title seems inconsistent. WDYT?

DD-506 #done

### DIFF
--- a/packages/shared/src/components/Settings.tsx
+++ b/packages/shared/src/components/Settings.tsx
@@ -47,6 +47,8 @@ export default function Settings({
     toggleShowTopSites,
     sortingEnabled,
     toggleSortingEnabled,
+    showWeeklyGoals,
+    toggleShowWeeklyGoals,
   } = useContext(SettingsContext);
   const [themes, setThemes] = useState([
     { label: 'Dark', value: 'dark' },
@@ -147,6 +149,16 @@ export default function Settings({
             compact={false}
           >
             Show feed sorting menu
+          </Switch>
+          <Switch
+            inputId="weekly-goal-widget-switch"
+            name="weekly-goal-widget"
+            className="my-3 big"
+            checked={showWeeklyGoals}
+            onToggle={toggleShowWeeklyGoals}
+            compact={false}
+          >
+            Show Weekly Goal widget
           </Switch>
         </div>
       </Section>
--- a/packages/shared/src/components/Settings.tsx
+++ b/packages/shared/src/components/Settings.tsx
@@ -56,7 +56,7 @@ export default function Settings({
     { label: 'Auto', value: 'auto' },
   ]);
 
-  const onToggleForLoggedUnUsers = (
+  const onToggleForLoggedInUsers = (
     onToggleFunc: () => Promise<void> | void,
   ): Promise<void> | void => {
     if (!user) {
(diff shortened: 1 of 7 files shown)


packages/shared/src/components/Settings.tsx
```diff
@@ -111,7 +115,7 @@ export default function Settings({
             name="hide-read"
             className="my-3"
             checked={showOnlyUnreadPosts}
-            onToggle={onShowOnlyUnreadPosts}
+            onToggle={() => onToggleForLoggedUnUsers(toggleShowOnlyUnreadPosts)}
```
rebelchris, line 118: We could use it directly as: `{onToggleForLoggedUnUsers(toggleShowOnlyUnreadPosts)}`

packages/shared/src/components/Settings.tsx
```diff
@@ -148,6 +152,16 @@ export default function Settings({
           >
             Show feed sorting menu
           </Switch>
+          <Switch
+            inputId="weekly-goal-widget-switch"
+            name="weekly-goal-widget"
+            className="my-3 big"
+            checked={showWeeklyGoals}
+            onToggle={() => onToggleForLoggedUnUsers(toggleShowWeeklyGoals)}
```
rebelchris, line 160: Same here:
`onToggle={onToggleForLoggedUnUsers(toggleShowWeeklyGoals)}`

rebelchris (changes requested):
Thanks for adding the proper tests everywhere!
Looks great 🙏

packages/shared/src/components/Settings.tsx
```diff
@@ -47,20 +47,24 @@ export default function Settings({
...
+    toggleShowWeeklyGoals,
   } = useContext(SettingsContext);
   const [themes, setThemes] = useState([
     { label: 'Dark', value: 'dark' },
     { label: 'Light', value: 'light' },
     { label: 'Auto', value: 'auto' },
   ]);
 
-  const onShowOnlyUnreadPosts = (): Promise<void> | void => {
+  const onToggleForLoggedUnUsers = (
```
idoshamun, line 59: I'm not sure about this function name

packages/shared/src/components/Settings.tsx
idoshamun: I'm not sure you're right. Because `onToggleForLoggedUnUsers` doesn't generate a function.
So if we call it directly, it will execute the function immediately on render and this is not what we want to achieve here. I hope I didn't miss anything

idoshamun (changes requested):
I love the idea of creating a reusable function the enforces the login check

packages/shared/src/components/Settings.tsx
rebelchris: Ah woeps totally missed that one 🤯

idoshamun (approved):
Way to go!

rebelchris (approved):

PR #861, opened 2022-01-31
feat: ranks modal

rebelchris:
## Changes

- Introducing the new rank modal
- Not finished see below for missing parts

## Missing

- ~~Actual data changes from the API~~
- The new icons
- ~~Colors for each rank~~
- Tags from API
- ~~Earn text based on actual rank status~~

## Testing

- Added tests for the new ranks

DD-508 #done
DD-507 #done


ayadotdev:
Great job Chris! 🥇

packages/shared/src/lib/rank.ts
```diff
@@ -2,6 +2,74 @@ export const NO_RANK = 'No rank';
 export const FINAL_RANK = 'You made it 🎉';
 export const STEPS_PER_RANK = [3, 4, 5, 6, 7];
 export const RANK_NAMES = ['Bronze', 'Silver', 'Gold', 'Platinum', 'Diamond'];
```
sshanzel, line 4: We probably need to update this as well or the areas that implement this to utilize the new one. What do you think?

sshanzel (commented):
This is superb! Just had one question but this PR looks great already!

packages/shared/src/lib/rank.ts
rebelchris: Yep we'll use the new RANKS object for this as well, wanted to take this on in DD-507 (Aya's ticket)

rebelchris: Actually @sshanzel decided to add this in this PR :)
Feel free to check again

packages/shared/src/components/modals/RanksModal.tsx
```diff
@@ -79,88 +25,183 @@ export interface RanksModalProps extends ModalProps {
...
+        {getNextRankText({
+          nextRank: rank + 1,
+          rank,
+          finalRank: isFinalRank,
+          progress,
+          showNextLevel: false,
+        })}
+      </p>
+    </section>
+  );
```
idoshamun, line 136: Looks somewhat lengthy. I think we can refactor a bit and split to helper components

packages/shared/src/components/modals/RanksModal.tsx
```diff
@@ -1,73 +1,19 @@
-import React, {
-  CSSProperties,
-  ReactElement,
-  ReactNode,
-  useContext,
-} from 'react';
+import React, { CSSProperties, ReactElement, useContext } from 'react';
```
idoshamun, line 1: Let's split this file into multiple files. It has so many things going on here. We can create a directory for the `RanksModal` and have everything there

idoshamun (changes requested):
Massive! I have some comments inside.
Make sure to test also the new rank modal and see that everything works there with these changes.

packages/shared/src/components/modals/RanksModal.tsx
rebelchris: On it 🥳

rebelchris:
@idoshamun I've refactored everything into small components and added some classed components to make things more clear.

Let me know what you think 🥳
@sshanzel Would also love to hear your thoughts on naming espacially.

packages/shared/src/components/modals/common.tsx
```diff
@@ -4,3 +4,7 @@ export const ModalHeader = classed(
   'header',
   'flex flex-row items-center py-2 px-3 h-14 border-b border-theme-divider-tertiary',
```
sshanzel, line 5: Not sure if this is part of the new guideline - but the padding for the header is now `24px` = `px-6`. If so, then should we make small updates from here or do it in a single separate PR?
```suggestion
  'flex flex-row items-center py-3 px-3 h-14 border-b border-theme-divider-tertiary',
```

packages/shared/src/components/modals/RanksModal/TimezoneText.tsx
```diff
@@ -0,0 +1,40 @@
...
+          onClick={() => onShowAccount()}
+        >
+          account details.
+        </ClickableText>
+      </SimpleTooltip>
+    </ModalText>
+  );
+
+  const loggedTimezoneDescription = !user?.timezone ? accountDetails : null;
+  const timezoneDescription = !user ? signIn : loggedTimezoneDescription;
```
sshanzel, line 36: Just my personal POV - I think we can return early if the user does not exist so we can reduce the ternary operators. Just felt a bit of difficulty reading these two lines. But feel free to ignore this if it won't make too much sense.
Ex:
```
if (!user) {
  return <ModalText className={classes}>
    To fit the weekly goal to your time zone, please sign up and add it in the
    account details.
  </ModalText>
}
```

packages/shared/src/components/modals/RanksModal/index.tsx
```diff
@@ -0,0 +1,55 @@
...
+  onRequestClose,
+  onShowAccount,
+  className,
+  ...props
+}: RanksModalProps): ReactElement {
+  const { user } = useContext(AuthContext);
+  useTrackModal({ isOpen: props.isOpen, title: 'ranks modal' });
+  const currentRank = rank;
+
+  // Todo: Refactor these to load from API
```
sshanzel, line 29: Just tagging for reference. API should now be available. Woohoo!

sshanzel (commented):
Looks neat! ✨ 

Just had some non-blocking comments/questions.

packages/shared/src/components/modals/common.tsx
rebelchris: Let's refactor right away, it should reflect all headers anyway.
I will add another ticket to revisit modals as they don't all use this sytem.

packages/shared/src/components/modals/RanksModal/TimezoneText.tsx
rebelchris: Sure sounds great, i'll change it :)

idoshamun:
@rebelchris something fails here. Just wanted to let you know while I'm reviewing

idoshamun (approved):
This looks so much better!
Clean, neat, well organized. I love it.

sshanzel (approved):
Spotless! ✨

idoshamun (approved):
This looks good, my only worry is that we not tested it enough.
This feature has so many edge cases, please do your best to cover them manually.

PR #862, opened 2022-02-01
feat: dailyfest - gamification (#860)

rebelchris:
## Changes

- Introducing the new gamification layer
- New ranks (7 days)

## Manual Testing

- On those affected packages:

DD-502 #done


packages/shared/src/components/DevCardPlaceholder.tsx
```diff
@@ -1,21 +1,33 @@
 import React, { ReactElement } from 'react';
 
+const getColorPrefix = (rank: number, isLocked?: boolean) => {
+  if (isLocked) {
+    return '--theme-rank-1-color';
+  }
+
+  return rank > 0 ? `--theme-rank-${rank}-color` : '--theme-rank-5-color';
```
rebelchris, line 8: @sshanzel is the default 5 rank ok here?

packages/shared/src/components/modals/NewRankModal.tsx
```diff
@@ -41,8 +41,8 @@ export default function NewRankModal({
   style,
   ...props
 }: NewRankModalProps): ReactElement {
-  const [shownRank, setShownRank] = useState(rank - 1);
-  const [shownProgress, setShownProgress] = useState(progress);
+  const [shownRank, setShownRank] = useState(rank > 0 ? rank - 1 : 0);
```
rebelchris, line 44: I'm actually guessing this could use the function Lee made as well.

packages/shared/src/components/modals/RanksModal/RanksTags.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+const RanksTags = ({
+  tags,
+  isColorPrimary,
+  limit = 6,
+}: RanksTagsProps): ReactElement => {
+  return (
+    <RanksTagsSection>
+      <ModalSubTitle>Reading status per tag</ModalSubTitle>
+      <RanksTagsList style={{ gridTemplateRows: '1fr 1fr 1fr' }}>
```
rebelchris, line 14: Still not really sure why we can't just use tailwind classes for this?
I'm pretty sure that rows does this exact same thing?

packages/shared/src/components/modals/RanksModal/index.tsx
```diff
@@ -0,0 +1,75 @@
...
+  return (
+    <ResponsiveModal
+      {...props}
+      onRequestClose={onRequestClose}
+      padding={false}
+      style={{ content: { maxWidth: '26.25rem' } }}
+    >
+      <ModalHeader>
+        <h3 className="font-bold typo-title3">Weekly reading goal</h3>
+        <ModalCloseButton onClick={onRequestClose} style={{ top: '0.5rem' }} />
```
rebelchris, line 41: Why don't we use `top-2` class?

idoshamun (approved):

rebelchris (commented):
Looks great, small little things

packages/shared/src/components/DevCardPlaceholder.tsx
sshanzel: Not sure which one to use, though for the areas that used to be using the default is now utilizing what's the user's actual rank.

packages/shared/src/components/modals/RanksModal/index.tsx
sshanzel: It is not being applied since there's a top class inside - unable to explicitly override.

packages/shared/src/components/modals/RanksModal/RanksTags.tsx
sshanzel: It actually has a different value - https://tailwindcss.com/docs/grid-template-rows

Tried arbitrary value but no luck: `grid-rows-[1fr 1fr 1fr]`.

packages/shared/src/components/DevCardPlaceholder.tsx
rebelchris: Let's use this for now 🙌

packages/shared/src/components/modals/RanksModal/RanksTags.tsx
rebelchris: For future reference `repeat(3, minmax(0, 1fr));` is the same as writing `1fr 1fr 1fr`.
The main issue here is this is a new class and the next cache doesn't take these into account.

I'll raise a separate issue for this.
Fixed it now.

packages/shared/src/components/modals/RanksModal/index.tsx
rebelchris: Removed as we don't use it anymore.

PR #863, opened 2022-02-01
refactor: profile section related to reading history

sshanzel:
## Changes

This displays the tags read by the users based on the date range filter selected.

Currently missing the changes for reading history ranks - will wait until the PR that has the icons get merged before applying the updates.

~~Currently, the date range is already utilized but will further update the placement once the design is released officially. API is yet to be merged to test on the staging env but already works locally.~~

With that, feel free to review and leave a comment.

## Manual Testing

- On those affected packages:

Preview:
Screen Shot 2022-02-01 at 10 34 14 PM

Screen Shot 2022-02-01 at 10 34 55 PM

DD-{number} #done


packages/webapp/pages/[userId]/index.tsx
```diff
@@ -252,6 +251,22 @@ const ProfilePage = ({ profile }: ProfileLayoutProps): ReactElement => {
               />
             )}
           </ActivityContainer>
+          <ActivityContainer>
+            <ActivitySectionTitle>
+              Top tags by reading days
+            </ActivitySectionTitle>
+            <div className="grid grid-cols-1 tablet:grid-cols-2 gap-3 tablet:gap-10">
+              {readingHistory.userReadingTopTags?.map((topTag) => (
```
Comment (rebelchris) [inline packages/webapp/pages/[userId]/index.tsx:259]: Maybe just call this "Tag"

packages/webapp/pages/[userId]/index.tsx
```diff
@@ -252,6 +251,22 @@ const ProfilePage = ({ profile }: ProfileLayoutProps): ReactElement => {
...
+                <ReadingTagProgress
+                  key={topTag.tag}
+                  topTag={topTag}
+                  isFilterSameYear={
+                    dropdownOptions[selectedHistoryYear] === CURRENT_YEAR
+                  }
```
Comment (rebelchris) [inline packages/webapp/pages/[userId]/index.tsx:265]: Why does this matter?

rebelchris (changes requested):
Looks solid, just not sure why we need the same year passing?

packages/webapp/pages/[userId]/index.tsx
Comment (sshanzel) [reply packages/webapp/pages/[userId]/index.tsx:265]: For example, you selected 2022 and we only have 32 days yet - the label would be "7/32 reading days"

packages/webapp/pages/[userId]/index.tsx
Comment (sshanzel) [reply packages/webapp/pages/[userId]/index.tsx:259]: I think it would kinda look odd if we're gonna do `tag.tag` (like from the key) since it has a property tag. What do you think?

Comment (rebelchris) [reply packages/webapp/pages/[userId]/index.tsx:259]: Ah yes makes sense 🙏

packages/webapp/pages/[userId]/index.tsx
Comment (rebelchris) [reply packages/webapp/pages/[userId]/index.tsx:265]: Cant we just always get the max days for the selected year then?

rebelchris (approved):

packages/shared/src/components/profile/ReadingTagProgress.tsx
```diff
@@ -0,0 +1,38 @@
...
+}
+
+const now = new Date();
+
+export function ReadingTagProgress({
+  topTag: { tag, readingDays, percentage },
+  isFilterSameYear = true,
+}: ReadingTagProgressProps): ReactElement {
+  const value = `${(percentage * 100).toFixed(0)}%`;
+  const denominator = isFilterSameYear ? getDayOfYear(now) : getDaysInYear(now);
```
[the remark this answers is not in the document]
sshanzel: The denominator would be different if it is on the same year. In 2022, we're only 32 days of the year. So instead of `3/365 reading days` it would progress based on what day of the year we are currently at `3/32 reading days`. What do you think?

packages/webapp/pages/[userId]/index.tsx
Comment (sshanzel) [reply packages/webapp/pages/[userId]/index.tsx:265]: Totally cool. Please have some rest! You've been grinding the fest! 💪

sshanzel:
This utilizes the v2 APIs now. @rebelchris I also took your suggestion to use `tag` than the previous `topTags` now that we utilize the `MostReadTags` that utilizes `value`.

sshanzel:
This should now be complete. Woohoo! cc @rebelchris @idoshamun 

Previews

1080p:
Screen Shot 2022-02-03 at 2 38 47 PM

720p:
Screen Shot 2022-02-03 at 2 39 18 PM

iPad Mini:
Screen Shot 2022-02-03 at 2 39 43 PM

iPhone 12 Pro:
Screen Shot 2022-02-03 at 2 40 10 PM

packages/shared/src/graphql/users.ts
```diff
@@ -24,7 +24,7 @@ export const USER_STATS_QUERY = gql`
 
 export type UserReadingRank = { currentRank: number };
 export type UserReadingRankData = { userReadingRank: UserReadingRank };
-export type MostReadTag = { value: string; count: number };
+export type MostReadTag = { value: string; count: number; percentage?: number };
 export type UserTooltipContentData = {
   rank: UserReadingRank;
   tags: MostReadTag[];
```
rebelchris, line 30: Should we not uniform this with the TopTag interface?

packages/webapp/__tests__/ProfilePage.tsx
```diff
@@ -360,3 +379,14 @@ it('should show the reading rank history of the user', async () => {
...
 });
+
+it('should show the top reading tags of the user', async () => {
+  renderComponent();
+  await waitForNock();
+  await Promise.all(
+    defaultTopTags.map(async ({ value: tag }) => {
+      const el = await screen.findByText(`#${tag}`);
+      expect(el).toBeInTheDocument();
+    }),
```
rebelchris, line 390: Not breaking, but we could include a check to see if the bar has a width property of {x}px?

packages/webapp/pages/[userId]/index.tsx
```diff
@@ -85,24 +78,30 @@ const readHistoryToTooltip = (
   );
 };
 
+interface RankHistoryProps {
+  rank: number;
+  rankName: string;
+  count: number;
+}
```
Comment (rebelchris) [inline packages/webapp/pages/[userId]/index.tsx:85]: Maybe we should also extract this from the `lib/ranks` props?
then we make sure they are the same type if we ever change something.

packages/webapp/pages/[userId]/index.tsx
```diff
@@ -217,8 +208,23 @@ const ProfilePage = ({ profile }: ProfileLayoutProps): ReactElement => {
...
+                clickableTitle="weekly goals"
+                link="https://docs.daily.dev/docs/your-profile/weekly-goal"
+              />
+              <Dropdown
+                className="hidden laptop:block ml-auto"
+                selectedIndex={selectedHistoryYear}
+                options={dropdownOptions}
+                onChange={(val, index) => setSelectedHistoryYear(index)}
+                buttonSize="small"
+                style={{ width: '8rem', minWidth: 'fit-content' }}
```
Comment (rebelchris) [inline packages/webapp/pages/[userId]/index.tsx:224]: Can't we use classnames for this?

rebelchris (changes requested):
wow nice work Lee!
Just have some concerns about interfaces that I think should be shared across the app.

packages/shared/src/components/profile/ReadingTagProgress.tsx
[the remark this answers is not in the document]
idoshamun: Reading days is something you should receive from the server. It's not the number of days the user could have read articles. It's the number of days they actually read an article. We need to adjust the API to return this value as well. We can add another field `total` to the existing query.

packages/shared/src/components/profile/ActivitySection.tsx
```diff
@@ -2,18 +2,51 @@ import React, { Fragment, ReactElement, ReactNode } from 'react';
...
 );
 
+interface TitleWithLinkProps {
+  title: string;
+  subtitle: string;
+  clickableTitle: string;
+  link: string;
+}
+
+export const TitleWithLink = ({
```
idoshamun, line 31: The name is a bit misleading because when I see the word Title I expect to see also `h1`, `h2`, or any `hx` element but here it's a span. So not sure how we want to address it.

idoshamun (changes requested):
Overall it looks good but there are some required changes

packages/shared/src/graphql/users.ts
sshanzel: They have different APIs that have different shapes. So I think we won't be able to unify them.

packages/webapp/__tests__/ProfilePage.tsx
sshanzel: Why do we want to check the width?

packages/webapp/pages/[userId]/index.tsx
Comment (sshanzel) [reply packages/webapp/pages/[userId]/index.tsx:85]: Good point!

packages/webapp/pages/[userId]/index.tsx
Comment (sshanzel) [reply packages/webapp/pages/[userId]/index.tsx:224]: I think we can. This was just copy and pasted from the previous implementation - forgot to update. Nice catch!

packages/shared/src/components/profile/ActivitySection.tsx
sshanzel: Makes total sense.

packages/webapp/__tests__/ProfilePage.tsx
sshanzel: Oh, okay. The percentage. Good point. Let me add that in the specific component's test suite.

idoshamun (approved):
Yessss!

sshanzel:
@rebelchris would love to hear your feedback! 🚢

rebelchris (approved):
Nice work Lee

PR #864, opened 2022-02-01
refactor: opt-out weekly goals

ayadotdev:
## Changes

Control the visibility of the sidebar rank badge based on the user settings.

## Manual Testing

✅ 

DD-506 #done

### DIFF
--- a/packages/shared/src/components/Settings.spec.tsx
+++ b/packages/shared/src/components/Settings.spec.tsx
@@ -39,7 +39,7 @@ const defaultSettings: RemoteSettings = {
   showTopSites: true,
   sidebarExpanded: true,
   sortingEnabled: false,
-  showWeeklyGoals: true,
+  optOutWeeklyGoal: true,
 };
 
 const updateSettings = jest.fn();
@@ -267,7 +267,7 @@ it('should mutate feed sorting enabled setting', () =>
   }));
 
 it('should mutate show weekly goals widget setting', () =>
-  testSettingsMutation({ showWeeklyGoals: false }, async () => {
+  testSettingsMutation({ optOutWeeklyGoal: false }, async () => {
     const checkboxes = await screen.findAllByRole('checkbox');
     const checkbox = checkboxes.find((el) =>
       // eslint-disable-next-line testing-library/no-node-access, testing-library/prefer-screen-queries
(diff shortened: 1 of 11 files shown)


packages/shared/src/components/SidebarRankProgress.spec.tsx
```diff
@@ -105,3 +133,26 @@ it('should show rank for anonymous users', async () => {
     expect(screen.queryAllByTestId('remainingPath').length).toEqual(2);
   });
 });
+
+it('should show rank if opt out weekly goals widget is checked', async () => {
```
idoshamun, line 137:
```suggestion
it('should show rank if show weekly goals toggle is checked', async () => {
```

packages/shared/src/components/Settings.spec.tsx
```diff
@@ -171,11 +171,11 @@ it('should utilize local cache settings for anonymous users', async () => {
   );
 });
 
-const testSettingsMutation = async (
+export const testSettingsMutation = async (
```
rebelchris, line 174: This doesn't need to be exported

rebelchris (changes requested):
Nice work, once small note from my side

packages/shared/src/contexts/SettingsContext.tsx
```diff
@@ -94,7 +94,7 @@ const defaultSettings: RemoteSettings = {
   showTopSites: true,
   sidebarExpanded: true,
   sortingEnabled: false,
-  showWeeklyGoals: true,
+  optOutWeeklyGoal: true,
```
sshanzel, line 97: The default value here now should be `false` since we want to initially include everyone.
```suggestion
  optOutWeeklyGoal: false,
```

sshanzel (changes requested):
Just had one comment but great work here!

packages/shared/src/contexts/SettingsContext.tsx
ayadotdev: Nice spot, thanks Lee.

idoshamun (approved):
Woohoo!

sshanzel (approved):
LGTM!

rebelchris (approved):
Ship it 🥳

PR #865, opened 2022-02-03
feat: added the new rank symbols and final colors

rebelchris:
## Changes

- Added the new symbol SVG paths
- Added the correct colors to the ranks variable

## Manual Testing

- On those affected packages:

DD-524 #done


idoshamun (approved):
I see that in many cases you use this color `#F5F8FC` directly. Please validate with Tsahi that this color is not theme based and that we can use it directly. Once that's set you can merge :)

rebelchris:
Well spotted Ido, I've adjusted it to be dynamic based on the theme.
The white is white in both version so that one should be ok.

idoshamun (approved):

PR #869, opened 2022-02-03
fix: invalidate ranks query on read

idoshamun:
## Changes

On increment reading rank, we need to invalidate the ranks query to reflect the updates in the same session.

## Manual Testing

- On those affected packages:


sshanzel (approved):
💪

rebelchris (approved):
Nice one

PR #870, opened 2022-02-03
fix: gamification feedback

sshanzel:
## Changes

- Describe what this PR does
- top tags space vertically is too big
- top tags is missing 1
- userReadingRankHIstory is offset incorrectly
- max width modal 420px
- links profile should open a new tab

## Manual Testing

- On those affected packages:

DD-{number} #done

### DIFF
--- a/packages/shared/src/components/profile/ActivitySection.tsx
+++ b/packages/shared/src/components/profile/ActivitySection.tsx
@@ -42,7 +42,7 @@ export const ActivitySectionHeader = ({
         {title}
         <ActivitySectionSubTitle>
           {subtitle}
-          <ClickableText tag="a" className="ml-1" href={link}>
+          <ClickableText tag="a" target="_blank" className="ml-1" href={link}>
             {clickableTitle}
           </ClickableText>
         </ActivitySectionSubTitle>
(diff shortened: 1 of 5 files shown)


packages/shared/src/components/profile/ActivitySection.tsx
```diff
@@ -42,7 +42,7 @@ export const ActivitySectionHeader = ({
         {title}
         <ActivitySectionSubTitle>
           {subtitle}
-          <ClickableText tag="a" className="ml-1" href={link}>
+          <ClickableText tag="a" target="_blank" className="ml-1" href={link}>
```
rebelchris, line 45: I think there should be two buttons here right?

rebelchris (approved):
There should be two links on the profile page, otherwise good work!

packages/shared/src/components/profile/ActivitySection.tsx
sshanzel: This is a reused component so both of them do this now.

sshanzel:
Preview:

Screen Shot 2022-02-03 at 10 06 23 PM

idoshamun (approved):
Go go go go!

PR #872, opened 2022-02-03
fix: don't show the bookmarks sharing modal while loading

idoshamun:
## Changes

Defer showing the modal while loading the data from the server.
This should prevent the glitch of the toggle switch from turning on immediately.

## Manual Testing

- On those affected packages:

DD-{number} #done

### DIFF
--- a/packages/shared/src/components/BookmarkFeedLayout.tsx
+++ b/packages/shared/src/components/BookmarkFeedLayout.tsx
@@ -5,6 +5,7 @@ import React, {
   useMemo,
   useState,
 } from 'react';
+import dynamic from 'next/dynamic';
 import { BOOKMARKS_FEED_QUERY, SEARCH_BOOKMARKS_QUERY } from '../graphql/feed';
 import AuthContext from '../contexts/AuthContext';
 import { CustomFeedHeader, FeedPage, FeedPageHeader } from './utilities';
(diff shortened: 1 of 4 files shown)


packages/shared/src/components/BookmarkFeedLayout.tsx
```diff
@@ -5,17 +5,16 @@ import React, {
   useMemo,
   useState,
 } from 'react';
+import dynamic from 'next/dynamic';
```
rebelchris, line 8: What are we loading dynamic?

packages/shared/src/components/modals/SharedBookmarksModal.tsx
```diff
@@ -49,10 +49,14 @@ export default function SharedBookmarksModal({
     },
   );
 
-  const [isRssUrlCopied, copyRssUrl] = useCopyLink(
+  const [, copyRssUrl] = useCopyLink(
```
rebelchris, line 52: This seems off?

rebelchris (changes requested):
Two small questions otherwise good to go

packages/shared/src/components/BookmarkFeedLayout.tsx
idoshamun: This is a lint fix. The current branch is failing on lint so I ran a lint fix. Basically this import was a bit down before and the lint moved it up here

packages/shared/src/components/modals/SharedBookmarksModal.tsx
idoshamun: Lint fix because it throws an error that there's an unused variable and this was the suggest fix

rebelchris (approved):

PR #874, opened 2022-02-03
fix: update cache key to prevent client error

idoshamun:
## Changes

The current implementation breaks if the previous cache version is present.
I changed the cache key to fix this issue and ignore the existing cache.

## Manual Testing

- On those affected packages:


rebelchris (approved):
Perfect

PR #877, opened 2022-02-04
fix: rank modal ui concerns

sshanzel:
## Changes

- Describe what this PR does
- placeholder tag
- radial completed step color

**Note:** I had to wrap the generate devcard button in a div due to a web limitation on disabled elements having their pointer-events not work even when overridden, hence, tooltip won't work.

Preview:

image

## Manual Testing

- On those affected packages:

DD-{number} #done


packages/shared/src/components/modals/RanksModal/RanksTags.tsx
```diff
@@ -3,14 +3,17 @@ import { ModalSubTitle } from '../common';
...
-const RanksTags = ({ tags }: RanksTagsProps): ReactElement => {
+const RanksTags = ({ tags, limit = 6 }: RanksTagsProps): ReactElement => {
   return (
     <RanksTagsSection>
       <ModalSubTitle>Reading status per tag</ModalSubTitle>
-      <RanksTagsList>
-        {tags.map((tag) => (
-          <RankTagItem tag={tag} key={tag.tag} />
-        ))}
+      <RanksTagsList style={{ gridTemplateRows: '1fr 1fr 1fr' }}>
```
rebelchris, line 10: Why don't we use the classname for this?

packages/shared/src/components/GoToDevCardButton.tsx
```diff
@@ -1,30 +1,37 @@
...
-    <Button
-      className={classNames('btn-primary', className)}
-      tag="a"
-      buttonSize="small"
-      href={`${process.env.NEXT_PUBLIC_WEBAPP_URL}devcard`}
-      onClick={onGenerateDevCardClick}
-    >
-      {children}
-    </Button>
+    <SimpleTooltip content="Sign up to Unlock">
```
rebelchris, line 22: Now it will always show this text right?
Also for signed in users?

rebelchris (changes requested):
Super nice work, got some small things we should look into.

packages/shared/src/components/modals/RanksModal/RanksTags.tsx
sshanzel: I tried `grid-rows-['1fr 1fr 1fr']` but  didn't work.

packages/shared/src/components/modals/RanksModal/RankTagItem.tsx
```diff
@@ -5,14 +5,21 @@ import styles from '../../RankProgress.module.css';
 import { RankTag, RankTagPill, RankTagProps } from './common';
 
 const RankTagItem = ({ tag }: RankTagProps): ReactElement => {
+  const isPlaceholder = tag === undefined || tag === null;
```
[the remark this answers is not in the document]
sshanzel: I think that's on the API side with redis. I tried searching but nothing came up.

packages/shared/src/components/GoToDevCardButton.tsx
sshanzel: Oh, wait. Forgot to revert as I was debugging earlier. Nice spot.

packages/shared/src/components/modals/RanksModal/RanksTags.tsx
rebelchris: It should be `grid-rows-3`

packages/shared/src/components/modals/RanksModal/RankTagItem.tsx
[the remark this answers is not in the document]
rebelchris: Oh no problem, just wanted to double check :)

packages/shared/src/components/GoToDevCardButton.tsx
```diff
@@ -1,30 +1,37 @@
...
-    <Button
-      className={classNames('btn-primary', className)}
-      tag="a"
-      buttonSize="small"
-      href={`${process.env.NEXT_PUBLIC_WEBAPP_URL}devcard`}
-      onClick={onGenerateDevCardClick}
-    >
-      {children}
-    </Button>
+    <SimpleTooltip content={isLocked && 'Sign up to Unlock'}>
```
rebelchris, line 22: Does it now not render when you hover it in non locked mode?

rebelchris (changes requested):

rebelchris (commented):
Screenshot 2022-02-04 at 15 40 21

Also just noticed these are now always white, which is not correct.

packages/shared/src/components/GoToDevCardButton.tsx
sshanzel: Yes it does now 🚀 

Screen Shot 2022-02-04 at 9 40 58 PM

packages/shared/src/components/modals/RanksModal/RanksTags.tsx
sshanzel: `grid-rows-3` provides a different value specifically `grid-template-rows: repeat(3, minmax(0, 1fr));` - which works differently.

sshanzel:
The radial progress color should now be fixed 🚢

rebelchris:
@sshanzel I've added an extra step to validate the amount of reads.
Would you mind checking that commit :)

And just noticed we also still missing this section:
Screenshot 2022-02-04 at 16 28 55

packages/shared/src/components/DevCardPlaceholder.tsx
```diff
@@ -56,15 +56,15 @@ export default function DevCardPlaceholder({
           id="devcard_placeholder_svg__a"
           width={2300}
           height={2030}
-          xlinkHref={!isLocked && profileImage}
```
sshanzel: @rebelchris  Why did you have to remove the condition? I think in the design it should not be displayed.

packages/shared/src/components/modals/RanksModal/DevCardFooter.tsx
```diff
@@ -20,17 +31,29 @@ export default function DevCardFooter({
...
+              <div className="ml-2 flex flex-1 relative h-2 rounded-full bg-theme-active">
+                <div
+                  className="absolute bottom-0 left-0 h-full rounded-full bg-theme-label-primary"
+                  style={{ width: (100 * reads) / devCardLimit }}
+                  data-testId="tagProgress"
+                />
+              </div>
+            </div>
+          ) : (
+            <GoToDevCardButton
```
sshanzel, line 53: @rebelchris In every scenario, in the design, the devcard button is always rendered. Why do we have to conditionally render it?

packages/shared/src/components/DevCardPlaceholder.tsx
rebelchris: See here :)

packages/shared/src/components/modals/RanksModal/DevCardFooter.tsx
rebelchris: See above comment:
https://www.sketch.com/s/2867ec63-0966-4647-b1d7-b6fe696ba251/a/jgA5lWz

packages/shared/src/components/DevCardPlaceholder.tsx
sshanzel: Awesome. Then that should cover it 🚢

packages/shared/src/components/modals/RanksModal/DevCardFooter.tsx
sshanzel: Nice. Then all the changes were valid. Thanks for the help!

packages/shared/src/components/modals/RanksModal/index.tsx
```diff
@@ -39,11 +41,31 @@ export default function RanksModal({
...
       <IntroSection onShowAccount={onShowAccount} user={user} />
-      <RanksBadges rank={currentRank} progress={progress} />
-      <RanksTags tags={tags} />
-      {!hideProgress && reads > devCardLimit && (
-        <DevCardFooter rank={rank} user={user} />
+      <RanksBadges rank={rank} progress={progress} />
+      {!user && (
+        <div className="flex flex-col items-center mt-2">
+          <span className="typo-footnote">
+            Sign up to add your weekly achievements to your profile.
```
rebelchris, line 48: I think this should also be based on the rank color right?

rebelchris (approved):
Just one small comment 🙌

packages/shared/src/components/modals/RanksModal/index.tsx
sshanzel: I think this should be good, in the design, it was a gold rank but had the text in primary.

rebelchris: You're right!
My mind played tricks on me

sshanzel: Lol, many things having their own variations for color 😆

packages/shared/src/components/DevCardPlaceholder.tsx
```diff
@@ -3,19 +3,31 @@ import React, { ReactElement } from 'react';
...
 }): ReactElement {
-  const colorPrefix =
-    rank > 0 ? `--theme-rank-${rank}-color` : '--theme-rank-5-color';
+  const getColorPrefix = () => {
+    if (isLocked) {
+      return '--theme-rank-1-color';
+    }
+
+    return rank > 0 ? `--theme-rank-${rank}-color` : '--theme-rank-5-color';
+  };
```
idoshamun, line 20: This can go outside the component

packages/shared/src/components/RadialProgress.module.css
```diff
@@ -15,5 +15,9 @@
     transform-origin: center;
     transition: stroke-dashoffset 0.5s ease-out
       var(--radial-progress-transition-delay);
+
+    &:global(.modal) {
+      stroke: var(--theme-label-primary);
+    }
```
idoshamun, line 21: A more explicit class name should be used I think. `modal` is pretty vague (at least for me)

packages/shared/src/components/RadialProgress.tsx
```diff
@@ -66,6 +66,7 @@ export type RadialProgressProps = {
   onTransitionEnd?: () => unknown;
   className?: string;
   style?: CSSProperties;
+  isAtModal?: boolean;
```
idoshamun, line 69: The naming comment goes in here as well

packages/shared/src/components/RankProgressWrapper.tsx
```diff
@@ -34,11 +32,7 @@ export default function RankProgressWrapper({
   const { user } = useContext(AuthContext);
   const [showAccountDetails, setShowAccountDetails] = useState(false);
   const [showRanksModal, setShowRanksModal] = useState(false);
-  const { flags } = useContext(FeaturesContext);
-  const devCardLimit = parseInt(
-    getFeatureValue(Features.DevcardLimit, flags),
-    10,
-  );
```
idoshamun, line 41: Nimrod mentioned that he would like to keep the support if this is already implemented.
One thing, I'd change is to make it to 50 by default instead of null.
I mean we will always limit the DevCard generation but the exact threshold is received from Flagsmith with a default of 50

idoshamun (changes requested):
Some changes are required :)

idoshamun (approved):
Perfect!

PR #878, opened 2022-02-04
feat: sharable bookmarks

ayadotdev:
## Changes

This PR adds Share bookmarks functionality which allows to switch between public and private modes.

## Manual Testing

There are some Tailwind test warnings unsolved.

DD-517 #done


packages/shared/src/components/BookmarkFeedLayout.tsx
```diff
@@ -55,9 +66,32 @@ export default function BookmarkFeedLayout({
       <FeedPageHeader className="mb-5">
         <h3 className="font-bold typo-callout">Bookmarks</h3>
       </FeedPageHeader>
-      <CustomFeedHeader className="relative mb-6">
+      <CustomFeedHeader className="flex relative mb-6">
```
rebelchris, line 69: I think the relative class became obsolete because we no longer use absolute positioning inside right?

packages/shared/src/components/BookmarkFeedLayout.tsx
```diff
@@ -55,9 +66,32 @@ export default function BookmarkFeedLayout({
...
+        >
+          Share bookmarks
+        </Button>
+        <Button
+          className="ml-4 btn-secondary block laptop:hidden"
+          buttonSize="medium"
+          icon={<SourceIcon />}
+          onClick={() => setShowSharedBookmarks(true)}
+          iconOnly
+        />
```
rebelchris, line 86: Would love to hear from @sshanzel and @idoshamun what they think
I suggested this in favor of abusing the useMediaHook which becomes super heavy on the JS side of things.

packages/shared/src/components/BookmarkFeedLayout.tsx
```diff
@@ -55,9 +66,32 @@ export default function BookmarkFeedLayout({
       <FeedPageHeader className="mb-5">
         <h3 className="font-bold typo-callout">Bookmarks</h3>
       </FeedPageHeader>
-      <CustomFeedHeader className="relative mb-6">
+      <CustomFeedHeader className="flex relative mb-6">
         {searchChildren}
+
+        <Button
+          className="ml-4 btn-secondary hidden laptop:flex"
+          buttonSize="medium"
```
rebelchris, line 74: I think when it's medium we don't need to define it actually.

packages/shared/src/components/BookmarkFeedLayout.tsx
```diff
@@ -55,9 +66,32 @@ export default function BookmarkFeedLayout({
...
+          icon={<SourceIcon />}
+          onClick={() => setShowSharedBookmarks(true)}
+        >
+          Share bookmarks
+        </Button>
+        <Button
+          className="ml-4 btn-secondary block laptop:hidden"
```
rebelchris, line 81: I would change the block to flex just to be more inlined.

packages/shared/src/components/modals/SharedBookmarksModal.tsx
```diff
@@ -0,0 +1,139 @@
...
+
+export default function SharedBookmarksModal({
+  className,
+  ...props
+}: ModalProps): ReactElement {
+  const { data: bookmarksSharingData, isFetched } =
+    useQuery<BookmarksSharingData>('bookmarksSharing', () =>
+      request(`${apiUrl}/graphql`, BOOKMARK_SHARING_QUERY),
+    );
+  const queryClient = useQueryClient();
```
rebelchris, line 33: Super nitpicky, but can we have this above the query that is defined first?

packages/shared/src/components/modals/SharedBookmarksModal.tsx
```diff
@@ -0,0 +1,139 @@
...
+        {...props}
+      >
+        <header className="flex items-center justify-between py-4 px-6 w-full border-b border-theme-divider-tertiary">
+          <h3 className="pl-2 font-bold typo-title3">Bookmarks sharing</h3>
+          <ModalCloseButton onClick={props.onRequestClose} />
+        </header>
+        <section className="flex flex-col py-6 px-6 mobileL:px-10">
+          <Switch
+            inputId="share-bookmarks-switch"
+            name="hide-read"
```
rebelchris, line 72: Not sure if this name reflects what it does, perhaps we can name it `share-bookmarks`
What do you think?

packages/shared/src/components/modals/SharedBookmarksModal.tsx
```diff
@@ -0,0 +1,139 @@
...
+          <p className="typo-callout text-theme-label-tertiary">
+            Need inspiration? we prepared some tutorials explaining some best
+            practices of integrating your bookmarks with other platforms.
+          </p>
+          <div className="flex justify-between mt-4">
+            <Button
+              rel="noopener noreferrer"
+              className="btn-secondary"
+              buttonSize="small"
+              href=""
```
rebelchris, line 122: We shouldn't forget to add this URL before we go live

packages/shared/src/components/utilities.tsx
```diff
@@ -125,6 +125,11 @@ export const FormErrorMessage = classed(
 
 export const ProfileHeading = classed('h1', 'm-0 self-start typo-title2');
 
+export const BookmarksModelHeading = classed(
+  'h1',
+  'm-0 self-start typo-title3',
+);
```
rebelchris, line 131: This is defined, but not used

rebelchris (changes requested):
Really well done!

I have some minor comments we can fix, but it looks amazing already 🥳

packages/shared/src/components/modals/SharedBookmarksModal.tsx
rebelchris: https://docs.daily.dev/docs/key-features/bookmarks

packages/shared/src/components/BookmarkFeedLayout.tsx
ayadotdev: You are right, I just tested without and it behaves correctly.

packages/shared/src/components/modals/SharedBookmarksModal.tsx
ayadotdev: Good catch.

packages/shared/src/components/BookmarkFeedLayout.tsx
sshanzel: I am always in favor of utilizing CSS for any screen-related condition unless impossible.

sshanzel: Almost every prop in this Component is the same, I think we can store it in a variable and just spread it to make it easier to change in the future.

packages/shared/src/components/BookmarkFeedLayout.tsx
```diff
@@ -55,9 +66,32 @@ export default function BookmarkFeedLayout({
...
+          buttonSize="medium"
+          icon={<SourceIcon />}
+          onClick={() => setShowSharedBookmarks(true)}
+          iconOnly
```
sshanzel, line 83: This prop shouldn't be necessary anymore - it is not utilized in the `Button` component itself as there's a specific variable that handles it inside.
```suggestion
```

packages/shared/src/components/modals/ModalCloseButton.tsx
```diff
@@ -11,11 +11,11 @@ function ModalCloseButtonComponent(
     <Button
       {...props}
       ref={ref}
-      className={classNames('btn-tertiary right-4 top-4', className)}
+      className={classNames('btn-tertiary', className)}
       buttonSize="small"
       title="Close"
       icon={<XIcon />}
-      style={{ position: 'absolute', ...style }}
+      style={{ ...style }}
```
sshanzel, line 18: This shouldn't be necessary once we remove the destructuring above.

packages/shared/src/components/BookmarkFeedLayout.tsx
```diff
@@ -55,9 +66,30 @@ export default function BookmarkFeedLayout({
...
+          onClick={() => setShowSharedBookmarks(true)}
+        >
+          Share bookmarks
+        </Button>
+        <Button
+          className="ml-4 btn-secondary flex laptop:hidden"
+          icon={<SourceIcon />}
+          onClick={() => setShowSharedBookmarks(true)}
+          iconOnly
+        />
```
sshanzel, line 84: I think it would be better to store the props in this button in a variable and just spread it for maintainability purposes - we would only have to change in one place when needed.

sshanzel (changes requested):
Just had a few comments but I think we are on the right track!

packages/shared/src/components/BookmarkFeedLayout.tsx
ayadotdev: What about something like this ? 

```
const shareBookmarksButton = (style: string, text?: string) => (
    <Button
      className={style}
      icon={<SourceIcon />}
      onClick={() => setShowSharedBookmarks(true)}
    >
      {text}
    </Button>
  );

  return (
    ...
        {shareBookmarksButton(
          'hidden laptop:flex ml-4 btn-secondary',
          'Share bookmarks',
        )}
        {shareBookmarksButton('flex laptop:hidden ml-4 btn-secondary')}
...

```

sshanzel: Great idea!

sshanzel (approved):
Let's go shipping! 🚢 🚢 🚢

rebelchris (changes requested):
Great!

Just noticed we're not testing any of the new things.
That would be my only concern.

Perhaps we can test the mutation and the switching to showcase the input field?

packages/shared/src/components/modals/SharedBookmarksModal.spec.tsx
```diff
@@ -0,0 +1,90 @@
...
+const renderComponent = (mocks: MockedGraphQLResponse[] = []): RenderResult => {
+  const client = new QueryClient();
+
+  mocks.forEach(mockGraphQL);
+  return render(
+    <QueryClientProvider client={client}>
+      <SharedBookmarksModal
+        isOpen
+        onRequestClose={onRequestClose}
+        ariaHideApp={false}
```
rebelchris, line 38: What do we use this for?

packages/shared/src/components/modals/SharedBookmarksModal.spec.tsx
```diff
@@ -0,0 +1,90 @@
...
+  ]);
+  // eslint-disable-next-line testing-library/no-node-access, testing-library/prefer-screen-queries
+  const checkboxes = await screen.findAllByRole('checkbox');
+  const checkbox = checkboxes.find((el) =>
+    // eslint-disable-next-line testing-library/no-node-access, testing-library/prefer-screen-queries
+    queryByText(el.parentElement, 'Public mode'),
+  ) as HTMLInputElement;
+  fireEvent.click(checkbox);
+  await waitFor(() => mutationCalled);
+  await waitFor(() => onRequestClose.mock.calls.length === 1);
```
rebelchris, line 89: We don't call the modal close once you toggle right?
So this test should not be valid I think.

rebelchris (changes requested):
Nice work!

Could you maybe include a check if the input field is now in the screen?
That way we make sure the frontend also works.

packages/shared/src/components/BookmarkFeedLayout.tsx
```diff
@@ -19,13 +22,21 @@ export type BookmarkFeedLayoutProps = {
   onSearchButtonClick?: () => unknown;
 };
 
+const SharedBookmarksModal = dynamic(
+  () =>
+    import(
+      /* webpackChunkName: "accountDetailsModal" */ './modals/SharedBookmarksModal'
```
idoshamun, line 28: The chunk name should be changed

```suggestion
      /* webpackChunkName: "sharedBookmarksModal" */ './modals/SharedBookmarksModal'
```

packages/shared/src/components/modals/SharedBookmarksModal.tsx
```diff
@@ -0,0 +1,140 @@
...
+  const [isRssUrlCopied, copyRssUrl] = useCopyLink(
+    () => bookmarksSharingData?.bookmarksSharing?.rssUrl,
+  );
+
+  if (!isFetched || !bookmarksSharingData?.bookmarksSharing) {
+    return <></>;
+  }
+
+  return (
+    <>
```
idoshamun, line 61: This empty fragment is not needed.
You can use `ResponsiveModal` as the direct parent

idoshamun (changes requested):
Looks great!
I love how you ended up handling the tests.
Two minor comments to address before merging.

packages/shared/src/components/modals/SharedBookmarksModal.tsx
idoshamun: @ayadotdev I see you resolved this but didn't apply my suggestion

ayadotdev: @idoshamun pushed now 👍

idoshamun (approved):

rebelchris (approved):
Wohoo Let's go 🚀

PR #879, opened 2022-02-04
feat: new rank modal

rebelchris:
## Changes

- Fixed the rank animations
- Small badge should only show the +1 animation if the user opt out of the popup
- Popup should never show the +1 
- The progress was showing the wrong circle 
- Small issues with it being called wrong, and styling being off.

## Manual Testing

- On those affected packages:

DD-523 #done


packages/shared/src/components/RankProgress.tsx
```diff
@@ -109,7 +169,9 @@ export function RankProgress({
         )
       : null;
 
-    const firstBadgeAnimation = badgeRef.current.animate(
+    const firstBadgeAnimation = (
+      badgeRef.current ? badgeRef.current : plusRef.current
```
sshanzel, line 173: Would probably be better off declaring this in a variable to only have a single point of change for maintainability. What do you think?

Like:
`const animatedRef = badgeRef.current || plusRef.current`

packages/shared/src/components/RankProgress.tsx
```diff
@@ -23,6 +24,63 @@ import styles from './RankProgress.module.css';
...
+  if (!smallVersion)
+    return (
+      <RankElement shownRank={shownRank} rank={rank} badgeRef={badgeRef} />
+    );
+
+  return !showRankAnimation || !animatingProgress ? (
+    <RankElement shownRank={shownRank} rank={rank} badgeRef={badgeRef} />
+  ) : (
+    <strong
+      ref={plusRef}
```
sshanzel, line 76: Why did we have to have two refs to utilize? We are rendering conditionally so there should only be one reference in any scenario.

sshanzel (changes requested):
Just had some comments but we are getting closer every second! 🙌

packages/shared/src/components/RankProgress.tsx
rebelchris: Great idea, I'll refactor it

packages/shared/src/components/RankProgress.tsx
rebelchris: Only one of them will render.
It just means the top one will never animate between the +1 and the rank, the bottom one can.

rebelchris: Oh you mean why the badge and plus are different, the one is a SVG reference the other a HTML element reference.

sshanzel: Sorry for the short unclear question. I mean why do we have to use two refs, namely `badgeRef` and `plusRef`, I think we can make them use just one ref since only one element will be available in every scenario.

rebelchris: Not sure if it's possible.
The one is a SVG the other a HTML element.

Once the current ref is set to SVG you can't swap over to HTML.

sshanzel: I think it should be possible through callback ref but, I can say it might not be worth the change. Well done! 💪

sshanzel (approved):
LGTM! 🚢

packages/shared/src/components/RankProgress.tsx
```diff
@@ -23,6 +24,63 @@ import styles from './RankProgress.module.css';
...
+  smallVersion,
+  showRankAnimation,
+  animatingProgress,
+}: GetRankElementProps): ReactElement => {
+  if (!smallVersion)
+    return (
+      <RankElement shownRank={shownRank} rank={rank} badgeRef={badgeRef} />
+    );
+
+  return !showRankAnimation || !animatingProgress ? (
```
idoshamun, line 72: It looks like line 69 and line 73 are the same. Maybe we can remove the tertiary operator and merge its condition into the if statement in line 67

idoshamun (approved):
Non blocking comment for your decision.
Other than that, we're ready to roll 🛼

packages/shared/src/components/RankProgress.tsx
rebelchris: @idoshamun Would you mind checking if this is what you meant?

idoshamun (approved):
Exactly!

PR #881, opened 2022-02-07
feat: Added the my feed position variations and stylings for each version

rebelchris:
## Changes

- Added the AB test for the my feed position
- Refactored how we pass the feed filters on the feed pages
- Added specific styling for each type

image

## Manual Testing

- On those affected packages:

DD-529 #done

### DIFF
--- a/packages/shared/src/components/CreateMyFeedButton.tsx
+++ b/packages/shared/src/components/CreateMyFeedButton.tsx
@@ -1,0 +1,128 @@
+import React, { ReactElement, useContext, useEffect } from 'react';
+import classNames from 'classnames';
+import { IFlags } from 'flagsmith';
+import PlusIcon from '../../icons/plus.svg';
+import UserIcon from '../../icons/user.svg';
+import { Button } from './buttons/Button';
+import { Features, getFeatureValue } from '../lib/featureManagement';
+import AnalyticsContext from '../contexts/AnalyticsContext';
+import { AnalyticsEvent } from '../hooks/analytics/useAnalyticsQueue';
+import { getThemeColor, ThemeColor } from './utilities';
+
+interface CreateMyFeedButtonProps {
+  type: string;
+  sidebarExpanded?: boolean;
+  flags: IFlags;
+  action: () => unknown;
+}
+
+const getAnalyticsEvent = (
+  eventName: string,
+  copy: string,
+): Partial<AnalyticsEvent> => ({
+  event_name: eventName,
+  target_type: 'my feed button',
+  target_id: 'sidebar',
+  feed_item_title: copy,
+});
+
+interface ClassProps {
+  explainerColor?: ThemeColor;
+  sidebarExpanded?: boolean;
+}
+const wrapperClasses = (sidebarExpanded: ClassProps) => {
+  return {
+    sidebar: sidebarExpanded ? 'h-[8.125rem] justify-center' : 'h-[8.125rem] ',
+    feed_top: 'w-full items-center mb-8',
+  };
+};
+
+const innerWrapClasses = ({ explainerColor, sidebarExpanded }: ClassProps) => {
+  return {
+    sidebar: sidebarExpanded
+      ? `flex-col ${explainerColor.shadow} border p-3 m-4`
+      : 'flex-col mx-3',
+    feed_title: `flex-row-reverse`,
+    feed_top: `${explainerColor.shadow} p-2 border`,
+    feed_ad: `${explainerColor.shadow} p-2 border flex-col flex-1 justify-center`,
+  };
+};
+const textClass = (sidebarExpanded: ClassProps) => {
+  return {
+    sidebar: sidebarExpanded
+      ? 'typo-footnote w-[11.25rem] mb-3 transform opacity-100 ease-linear duration-200  delay-200'
+      : 'typo-footnote w-[11.25rem] mb-3 transform duration-0 delay-0 opacity-0',
+    feed_top: 'typo-footnote ml-2',
+    feed_title: 'typo-footnote',
+    feed_ad: 'typo-body font-bold mx-6 text-center mb-6',
+  };
+};
+const buttonClass = {
+  sidebar: 'w-full',
+  feed_title: 'w-auto mr-4',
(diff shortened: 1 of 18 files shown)


packages/shared/src/hooks/useMyFeed.ts
```diff
@@ -25,6 +30,7 @@ export function useMyFeed(): UseMyFeed {
   const { user } = useContext(AuthContext);
   const { flags } = useContext(FeaturesContext);
   const shouldShowMyFeed = isFeaturedEnabled(Features.MyFeedOn, flags);
+  const myFeedPosition = getFeatureValue(Features.MyFeedPosition, flags);
```
idoshamun, line 33: We want to use our new function that chooses a value out of a set of possibilities otherwise fall back to the default `sidebar`. We might introduce new variations in the future and don't want the version to break.

idoshamun (changes requested):
Sweet! A few changes before we merge

packages/shared/src/hooks/useMyFeed.ts
rebelchris: This is the case, the defaults are managed in the `featureManagement` class.

See here: 

https://github.com/dailydotdev/apps/blob/1dee521b54563cb0e9224dc2a9fb7a2cad1561d2/packages/shared/src/lib/featureManagement.ts#L33

```js
  static readonly MyFeedPosition = new Features('my_feed_position', 'sidebar');
```

rebelchris:
@idoshamun Thanks for that amazing review!
The possibility fallback should already be in place, the other points I've addressed 🙌

packages/shared/src/components/CreateMyFeedButton.tsx
```diff
@@ -0,0 +1,129 @@
...
+  return {
+    sidebar: sidebarExpanded
+      ? `flex-col ${explainerColor.shadow} border p-3 m-4`
+      : 'flex-col mx-3',
+    feed_title: `flex-row-reverse`,
+    feed_top: `${explainerColor.shadow} p-2 border`,
+    feed_ad: `${explainerColor.shadow} p-2 border flex-col flex-1 justify-center`,
+  };
+};
+const textClass = (sidebarExpanded: ClassProps) => {
```
sshanzel, line 51: I think this needs destructuring.
```suggestion
const textClass = ({ sidebarExpanded }: ClassProps) => {
```

packages/shared/src/components/CreateMyFeedButton.tsx
```diff
@@ -0,0 +1,129 @@
...
+  target_type: 'my feed button',
+  target_id: type,
+  feed_item_title: copy,
+});
+
+interface ClassProps {
+  explainerColor?: ThemeColor;
+  sidebarExpanded?: boolean;
+}
+const wrapperClasses = (sidebarExpanded: ClassProps) => {
```
sshanzel, line 34: I think this needs destructuring.
```suggestion
const wrapperClasses = ({ sidebarExpanded }: ClassProps) => {
```

packages/shared/src/components/CreateMyFeedButton.tsx
```diff
@@ -0,0 +1,129 @@
...
+  return {
+    sidebar: sidebarExpanded ? 'h-[8.125rem] justify-center' : 'h-[8.125rem] ',
```
sshanzel, line 36: Both of them have the same height, I think we can do the condition on the centering part only.
```suggestion
    sidebar: classNames('h-[8.125rem]', sidebarExpanded && 'justify-center'),
```

packages/shared/src/components/CreateMyFeedButton.tsx
```diff
@@ -0,0 +1,128 @@
...
+interface CreateMyFeedButtonProps {
+  type: string;
+  sidebarExpanded?: boolean;
+  flags: IFlags;
+  action: () => unknown;
+}
+
+const getAnalyticsEvent = (
+  eventName: string,
+  copy: string,
```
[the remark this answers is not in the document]
idoshamun: Yes :)

packages/shared/src/hooks/useMyFeed.ts
idoshamun: It's not enough actually. Let's say in this version we support two values for this feature `sidebar` and `feed_ad` (I know that in reality there's more but just an example). Then the next version comes and we want to add a new type `header`. We set it in Flagsmith and now the old version breaks. Whereas if we check in every version of the received value is one of the supported values and if not fall back to default, it will be backward compatible.

packages/shared/src/components/CreateMyFeedButton.tsx
```diff
@@ -0,0 +1,129 @@
...
+    feed_title: `flex-row-reverse`,
+    feed_top: `${explainerColor.shadow} p-2 border`,
+    feed_ad: `${explainerColor.shadow} p-2 border flex-col flex-1 justify-center`,
+  };
+};
+const textClass = (sidebarExpanded: ClassProps) => {
+  return {
+    sidebar: sidebarExpanded
+      ? 'typo-footnote w-[11.25rem] mb-3 transform opacity-100 ease-linear duration-200  delay-200'
+      : 'typo-footnote w-[11.25rem] mb-3 transform duration-0 delay-0 opacity-0',
```
sshanzel, line 55: I think we can unify some of the classes.
```suggestion
    sidebar: classNames('typo-footnote w-[11.25rem] mb-3 transform', sidebarExpanded
      ? 'opacity-100 ease-linear duration-200  delay-200'
      : 'duration-0 delay-0 opacity-0'),
```

sshanzel (changes requested):
So awesome to see that we have a reusable component for different scenarios! Just had a few comments but nothing major 🚢

rebelchris:
@sshanzel and @idoshamun thanks so much!
Added a validTypes object that can be set optional.

This should also help with defining what types we allow for each of the features.
And added some basic tests to test if nothing breaks over time in this feature.

sshanzel (approved):
LGTM! 🚢

idoshamun (approved):
Awesome!

PR #883, opened 2022-02-08
fix: missed rank index

sshanzel:
## Changes

Describe what this PR does
- Fixes missed indexes in the `RANKS` object.

## Manual Testing

- On those affected packages:

DD-{number} #done

### DIFF
--- a/packages/shared/src/components/RankProgress.tsx
+++ b/packages/shared/src/components/RankProgress.tsx
@@ -144,7 +144,7 @@ export function RankProgress({
     }
 
     if (!finalRank) {
-      return RANKS[rankIndex].steps;
+      return RANKS[rank].steps;
     }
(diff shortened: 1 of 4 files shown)


rebelchris (approved):
Are there no other places we could also need this?
Otherwise really great work 🙌

idoshamun (approved):
I agree with Chris. Make sure there's no other place we use the ranks without `getRank`

sshanzel:
Thank you both for the awesome review!

@rebelchris can you please have a look at the last commit that it is valid?

rebelchris:
@sshanzel Does seem valid yes!
It should never fail as this is a safetype check based on negative values so I think you got them all now 🙌

Thanks for going through them 🚀

PR #884, opened 2022-02-08
fix: radial progress color

sshanzel:
## Changes

Describe what this PR does
The radial progress color must be consistent with the color in the sidebar. There's a discrepancy when the user is trying to re-earn the previous rank.

Fixed preview:
image

## Manual Testing

- On those affected packages:

DD-{number} #done

### DIFF
--- a/packages/shared/src/components/RankProgressWrapper.tsx
+++ b/packages/shared/src/components/RankProgressWrapper.tsx
@@ -90,6 +90,7 @@ export default function RankProgressWrapper({
           onRequestClose={closeRanksModal}
           onShowAccount={() => setShowAccountDetails(true)}
           reads={reads}
+          previousRank={rankLastWeek}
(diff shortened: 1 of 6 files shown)


rebelchris (approved):
Screenshot 2022-02-08 at 10 41 28

This was my current one, was it only on the legendary rank where this issue occurred?
Nevertheless, it makes sense so not sure why mine showed up correctly already.

sshanzel:
@rebelchris it was also for every re-earn situation.

I think there's also a bug in the screenshot you provided. It says "earn 1/6" for Legendary which I think the offset should only be at Diamond.

rebelchris:
Its on the preview i think you already fixed this in another scenario not?

sshanzel:
Ah, I think you are right, we have this already covered. Things should be what is intended now.

idoshamun (approved):
Great fix

PR #885, opened 2022-02-08
fix: :test_tube: Added rank tests

rebelchris:
## Changes

- Added tests for all the ranks when there is no current rank
- Added tests for all the ranks when there is a previous week rank

## Manual Testing

- On those affected packages:


packages/shared/src/components/SidebarRankProgress.spec.tsx
```diff
@@ -158,3 +159,62 @@ it('should not show rank if show weekly goals toggle is not checked', async () =
     expect(screen.queryByTestId('completedPath')).not.toBeInTheDocument();
   });
 });
+
+RANKS.forEach((rank) => {
```
idoshamun, line 163: Nice!!

idoshamun (approved):
I love how you loop through the ranks

sshanzel (approved):
Looking great! I will add some tests for the modal in a separate PR dedicated for fixing Tsahi's concern.

PR #886, opened 2022-02-08
fix: in some cases the close button was below the rank text

rebelchris:
## Changes

- Realised we had this issue open in regards to the new rank modal, so perfect time to fix it.
- Is was possible for the rank text to be overlapping the close button
- By setting a fixed z-index we ensure it's always higher.

## Manual Testing

- On those affected packages:

DD-499 #done


idoshamun (approved):
Great spot

PR #894, opened 2022-02-10
fix: :bug: Fix the re-earn state for the popup modal

rebelchris:
## Changes

- We did not take the previous rank into account when it comes to the re-earn for the popup.
- Also fixed the state of the badges in the popup

## Manual Testing

- On those affected packages:


idoshamun (approved):
Woohoo!

PR #895, opened 2022-02-10
fix: forgot to check the my feed button should actually be rendered

rebelchris:
## Changes

- Forgot to add the actual filter check on the feed options
- This let to the button always to show, even when you filtered before.

## Manual Testing

- On those affected packages:


idoshamun (approved):
Please confirm that it works and release a new patch version

PR #896, opened 2022-02-10
fix: devcard overlap on sidebar

sshanzel:
## Changes

The sidebar overlaps on the content - we have now changed the breakpoints.

Max size before breaking:
Screen Shot 2022-02-10 at 7 11 45 PM

## Manual Testing

- On those affected packages:

DD-{number} #done

### DIFF
--- a/packages/webapp/pages/devcard.tsx
+++ b/packages/webapp/pages/devcard.tsx
@@ -179,7 +179,7 @@ const Step2 = ({
   return (
     <div className="flex flex-col self-stretch laptop:self-center mx-2 mt-5">
       <h1 className="mx-3 mb-8 font-bold typo-title1">Share your #DevCard</h1>
-      <main className="grid grid-cols-1 laptop:grid-cols-2 gap-10 laptop:gap-20">
+      <main className="grid grid-cols-1 laptopL:grid-cols-2 gap-10 laptop:gap-20">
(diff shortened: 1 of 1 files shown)


rebelchris (approved):
Cool!

Thanks for checking this

idoshamun (changes requested):
I think the gap should change breakpoint accordingly, isn't it?

sshanzel:
@idoshamun it used to be not. Let me try and check this approach.

idoshamun:
@sshanzel I mean instead of `laptop:gap-20"` use `laptopL:gap-20`

sshanzel:
@idoshamun looked perfect with your suggestion! 🚀 

image

sshanzel:
@idoshamun yes I did, having the initial change would break immediately on laptop screen. I think this was forgotten to be considered when the sidebar change was designed. Should we reach out to Tsahi on how he would want it in consideration of the sidebar?

idoshamun:
Yes. This would be the best

sshanzel:
After discussing with Tsahi on what should be the final layout, he mentioned that we should set specific size on the devcard itself and the textarea (will vary based on screen size). Then on laptop it should still be on the same row.

Have also checked different screen sizes and still displaying as intended.

image

PR #897, opened 2022-02-10
fix: :bug: Fix breaking on non-existing rank call

rebelchris:
## Changes

- We didn't implement `getRank` everywhere 
- In this case it was failing in the `NewRankModal`
- `You earned the {RANKS[rank - 1].name?.toLowerCase()} rank`
- Should have been:
- `You earned the {RANKS[getRank(rank)].name?.toLowerCase()} rank`

The case is that for rank 0 it will try and get -1 which doesn't exists.
Because of this danger I decided to implement the `getRank` function everywhere this pattern was used.
It can't really break since it will check if it's 0 or not.

Tested for these use-cases on webapp and extension now.

## Manual Testing

- On those affected packages:

### DIFF
--- a/packages/shared/src/components/Rank.tsx
+++ b/packages/shared/src/components/Rank.tsx
@@ -221,10 +221,10 @@ const rankPaths: ((fill: string) => ReactNode)[] = [
 ];
 
 const getColorPath = (rank, id) => {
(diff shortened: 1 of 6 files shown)


idoshamun (approved):
Thanks a lot for doing this 😊

PR #898, opened 2022-02-10
fix: :bug: Making sure the mobile rank only shows on mobile

rebelchris:
## Changes

- We always rendered the mobile header rank wrapper
- This wrapper contains the `NewRankModal` so it would show up twice on desktop.
- Added a `sidebarRendered` prop to it so it will render the empty wrapper on desktop.

## Manual Testing

- On those affected packages:

### DIFF
--- a/packages/shared/src/components/MobileHeaderRankProgress.tsx
+++ b/packages/shared/src/components/MobileHeaderRankProgress.tsx
@@ -27,7 +27,11 @@ export default function MobileHeaderRankProgress({
 }: MobileHeaderRankProgressProps): ReactElement {
   const { isLoading, rank, nextRank } = useReadingRank();
 
-  if (isLoading || sidebarRendered) {
+  if (sidebarRendered) {
+    return <></>;
+  }
+
+  if (isLoading) {
     return <Wrapper>&nbsp;</Wrapper>;
   }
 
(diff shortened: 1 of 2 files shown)


packages/shared/src/components/MobileHeaderRankProgress.tsx
```diff
@@ -19,10 +19,15 @@ const Wrapper = ({ children, ...props }: WrapperProps) => {
...
+interface MobileHeaderRankProgressProps {
+  sidebarRendered: boolean;
+}
+export default function MobileHeaderRankProgress({
+  sidebarRendered,
+}: MobileHeaderRankProgressProps): ReactElement {
   const { isLoading, rank, nextRank } = useReadingRank();
 
-  if (isLoading) {
+  if (isLoading || sidebarRendered) {
```
idoshamun, line 30: If sidebar rendered is true shouldn't we just return an empty element? `<></>`

idoshamun (changes requested):

packages/shared/src/components/MobileHeaderRankProgress.tsx
rebelchris: Changed it to this now.
Initially I was afraid we would get the flicker back on mobile, but it seems to be fine.

sshanzel (approved):
LGTM! 🚢

idoshamun (approved):

PR #900, opened 2022-02-14
fix: use the getRank function everywhere we access the ranks array

idoshamun:
I found a missing spot where we don't use the `getRank` before accessing the `RANKS` array.

DD-548 #done

### DIFF
--- a/packages/shared/src/lib/rank.ts
+++ b/packages/shared/src/lib/rank.ts
@@ -115,13 +115,14 @@ export const getNextRankText = ({
   rankLastWeek,
   showNextLevel = true,
 }: GetNextRankTextProps): string => {
-  const { steps } = RANKS[getRank(rank)];
+  const clampedRank = getRank(rank);
+  const { steps } = RANKS[clampedRank];
   if (finalRank && progress >= steps) return FINAL_RANK;
   if (finalRank || (nextRank === rankLastWeek && progress < steps))
     return `Re-earn: ${progress}/${steps} days`;
   if (nextRank === 0) return `Earn: ${progress ?? 0}/1 days`;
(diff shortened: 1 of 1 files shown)


packages/shared/src/lib/rank.ts
```diff
@@ -115,13 +115,14 @@ export const getNextRankText = ({
...
-  const { steps } = RANKS[getRank(rank)];
+  const clampedRank = getRank(rank);
+  const { steps } = RANKS[clampedRank];
   if (finalRank && progress >= steps) return FINAL_RANK;
   if (finalRank || (nextRank === rankLastWeek && progress < steps))
     return `Re-earn: ${progress}/${steps} days`;
   if (nextRank === 0) return `Earn: ${progress ?? 0}/1 days`;
-  if (showNextLevel) return `Next level: ${RANKS[rank].name}`;
-  return `Earn: ${progress ?? 0}/${RANKS[rank].steps} days`;
+  if (showNextLevel) return `Next level: ${RANKS[clampedRank].name}`;
```
sshanzel, line 124: I think this should be using the raw `rank` as it will have a plus one which is the next level this condition is supposed to display.

sshanzel (changes requested):
Just had one comment but I think the change in the last return should be correct.

rebelchris (approved):
Not sure if the Next level is needed, but it can't hurt.
Let's try this out, it might have been the earn ;)

packages/shared/src/lib/rank.ts
rebelchris: We might have to introduce a function that will handle the zero, but doesn't do the -1
What do you think?

rebelchris (changes requested):

PR #901, opened 2022-02-14
feat: add stats, actions and menu to reading history posts

ayadotdev:
## Changes

- Adds stats and actions to posts in the reading history
- Adds a menu button on the reading history post
- Adds "Share article" and "Save to bookmarks" from reading history functionality

https://www.loom.com/share/891e2cb839d347818490fc9d9dcbb11c

## Manual Testing

- On those affected packages:

## Things missing (to be confirmed with the product team)

- Link copied message after clicking on Share article

DD-473 #done


packages/shared/icons/outline/x.svg
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
rebelchris, line 1: If we are using the filled/outlined version could we please move the existing one?
Technically there is another task for moving the SVG to the modern one.

Appreciate you already moved these, but would love to have them in the correct folders.

packages/shared/src/components/PostOptionsReadingHistoryMenu.tsx
```diff
@@ -0,0 +1,156 @@
...
+import { useCopyLink } from '../hooks/useCopyLink';
+
+const PortalMenu = dynamic(() => import('./fields/PortalMenu'), {
+  ssr: false,
+});
+
+export type PostOptionsReadingHistoryMenuProps = {
+  post: ReadHistoryPost;
+  onHidden?: () => unknown;
+  displayCopiedMEssageFunc?: () => void;
```
rebelchris, line 22: Typo here
```suggestion
  displayCopiedMessageFunc?: () => void;
```

packages/shared/src/components/PostOptionsReadingHistoryMenu.tsx
```diff
@@ -0,0 +1,156 @@
...
+      pages: newItems,
+    });
+    return () => {
+      queryClient.setQueryData<ReadHistoryInfiniteData>(
+        historyQueryKey,
+        oldReadingHistory,
+      );
+    };
+  };
+/* eslint-disable no-param-reassign */
```
rebelchris, line 66: Why is this needed here?

packages/shared/src/components/PostOptionsReadingHistoryMenu.tsx
```diff
@@ -0,0 +1,156 @@
...
+  onHidden?: () => unknown;
+  displayCopiedMEssageFunc?: () => void;
+  onHide?: (postId: string) => unknown;
+};
+
+const MenuIcon = ({ Icon }) => {
+  return <Icon className="mr-2 text-2xl" />;
+};
+
+/* eslint-disable no-param-reassign */
```
rebelchris, line 30: Which issue is this giving?

packages/shared/src/components/PostOptionsReadingHistoryMenu.tsx
```diff
@@ -0,0 +1,156 @@
...
+export default function PostOptionsReadingHistoryMenu({
+  post,
+  onHidden,
+  onHide,
+}: PostOptionsReadingHistoryMenuProps): ReactElement {
+  const { user } = useContext(AuthContext);
+  const queryClient = useQueryClient();
+  const historyQueryKey = ['readHistory', user?.id];
+
+  /* eslint-disable @typescript-eslint/no-unused-vars */
```
rebelchris, line 77: Also here, which variable is unused?

packages/shared/src/components/PostOptionsReadingHistoryMenu.tsx
```diff
@@ -0,0 +1,156 @@
...
+  return (
+    <>
+      <PortalMenu
+        disableBoundariesCheck
+        id="reading-history-options-context"
+        className="menu-primary"
+        animation="fade"
+        onHidden={onHidden}
+      >
+        <Item key="share" className="typo-callout" onClick={copyPostLink}>
```
rebelchris, line 129: Key is not really needed here I don't think.

packages/shared/src/components/PostOptionsReadingHistoryMenu.tsx
```diff
@@ -0,0 +1,156 @@
...
+            {getBookmarkMenuText()}
+          </a>
+        </Item>
+        <Item
+          key="close"
+          className="laptop:hidden typo-callout"
+          onClick={() => onHide(post?.id)}
+        >
+          <a className="flex w-full typo-callout">
+            <MenuIcon Icon={XIcon} /> Remove article
```
rebelchris, line 150: When using the new icons the idea is actually that they change on certain states, if we don't use that yet I would actually suggest sticking to the old ones.

packages/shared/src/components/PostOptionsReadingHistoryMenu.tsx
```diff
@@ -0,0 +1,156 @@
...
+          onClick={onBookmarkReadingHistoryPost}
+        >
+          <a className="flex w-full typo-callout">
+            {getBookmarkIcon()}
+            {getBookmarkMenuText()}
+          </a>
+        </Item>
+        <Item
+          key="close"
+          className="laptop:hidden typo-callout"
```
rebelchris, line 146: Did you test this by switching between laptop/mobile?
It might happen that it will keep it hidden once it set that.

packages/shared/src/components/PostOptionsReadingHistoryMenu.tsx
```diff
@@ -0,0 +1,156 @@
...
+          className="typo-callout"
+          onClick={onBookmarkReadingHistoryPost}
+        >
+          <a className="flex w-full typo-callout">
+            {getBookmarkIcon()}
+            {getBookmarkMenuText()}
+          </a>
+        </Item>
+        <Item
+          key="close"
```
rebelchris, line 145: Why do we define the key here?

packages/shared/src/components/cards/PostMetadataReadingHistory.tsx
```diff
@@ -0,0 +1,45 @@
+import React, { ReactElement, ReactNode } from 'react';
+import classNames from 'classnames';
+import { ReadHistoryPost } from '../../graphql/posts';
+import classed from '../../lib/classed';
+
+const Separator = classed(
+  'div',
+  'mx-1 w-0.5 h-0.5 rounded-full bg-theme-label-tertiary',
+);
```
rebelchris, line 9: Perhaps we can move this to a more general file with common components?

packages/shared/src/components/cards/PostMetadataReadingHistory.tsx
```diff
@@ -0,0 +1,45 @@
+import React, { ReactElement, ReactNode } from 'react';
```
rebelchris, line 1: I think you copied this file from the extension right?
This is not really specific to reading history.

I would suggest completely moving this to the shared components and removing the one in the extension.
(Side benefit: I also need this for the companion app)

packages/shared/src/components/history/ReadingHistoryList.tsx
```diff
@@ -65,8 +77,17 @@ function ReadHistoryList({
...
       <InfiniteScrollScreenOffset ref={infiniteScrollRef} />
+      <PostOptionsReadingHistoryMenu
+        post={readingHistoryContextItem?.post}
+        onHidden={() => setReadingHistoryContextItem(null)}
+        onHide={(postId) =>
+          onHide({
+            postId,
+            timestamp: readingHistoryContextItem.timestampDb,
+            ...queryIndexes,
+          })
```
rebelchris, line 88: This seems a bit messy, can we not access al these things from the readingHistoryContextItem?

packages/shared/src/graphql/posts.ts
```diff
@@ -33,7 +33,7 @@ export interface Post {
   featuredComments?: Comment[];
   trending?: number;
   description?: string;
-  summary: string;
+  summary?: string;
```
rebelchris, line 36: Why did we change this?

packages/shared/src/graphql/sources.ts
```diff
@@ -3,7 +3,7 @@ import { gql } from 'graphql-request';
 export interface Source {
   __typename?: string;
   id?: string;
-  name: string;
+  name?: string;
```
rebelchris, line 6: Why is this changed?

packages/shared/src/hooks/useReadingHistoryContextMenu.ts
```diff
@@ -0,0 +1,40 @@
...
+  setReadingHistoryContextItem;
+  onReadingHistoryContextOptions;
+  queryIndexes;
+} {
+  const [readingHistoryContextItem, setReadingHistoryContextItem] =
+    useState<ReadHistory>();
+  const [queryIndexes, setQueryIndexes] = useState<QueryIndexes>({
+    page: -1,
+    edge: -1,
+  });
```
rebelchris, line 17: Why do we set them to -1, what's the thought behind that vs null or 0?

packages/shared/src/components/PostOptionsReadingHistoryMenu.tsx
```diff
@@ -0,0 +1,156 @@
+import React, { ReactElement, useContext } from 'react';
+import { Item } from '@dailydotdev/react-contexify';
+import dynamic from 'next/dynamic';
+import { QueryClient, QueryKey, useQueryClient } from 'react-query';
+import { ReadHistoryPost } from '../graphql/posts';
+import ShareIcon from '../../icons/share.svg';
+import BookmarkIcon from '../../icons/outline/bookmark_outlined.svg';
+import BookmarkIconFilled from '../../icons/filled/bookmark_filled.svg';
```
rebelchris, line 8: I'm not sure we need these changed versions actually, we should keep the icon uniform throughout the app for now.

rebelchris (changes requested):
Wow really nice work already!
Looks exciting.

I have some small issues and questions, but very happy to see you tackle this issue 🙌

packages/shared/src/components/PostOptionsReadingHistoryMenu.tsx
ayadotdev: `isPostLinkCopied` is unused.  Is there a way to destructure only `copyPostLink` from `useCopyLink()`?

packages/shared/src/components/PostOptionsReadingHistoryMenu.tsx
ayadotdev: Should I use the old x icon even if there is the new one in the design guide?

packages/shared/src/components/PostOptionsReadingHistoryMenu.tsx
ayadotdev: Yes, I tested switching between mobile and laptop multiple times, all seems to work.

packages/shared/src/components/cards/PostMetadataReadingHistory.tsx
ayadotdev: I copied it from /shared/src/components/cards/PostMetadata.tsx, just created another one for Reading history item. It is already in the shared folder 🤔

packages/shared/src/graphql/posts.ts
ayadotdev: This is super weird. Good catch.

packages/shared/src/graphql/sources.ts
ayadotdev: Same. Weird.

packages/shared/src/components/PostOptionsReadingHistoryMenu.tsx
ayadotdev: Should the Bookmark icon still be color filled once the article is bookmarked?

packages/shared/src/components/history/ReadingHistoryList.tsx
ayadotdev: Agree, it is a bit messy, but ReadingHistoryContextItem doesn't contain queryIndexes at the moment. Do you suggest to refactor the useReadingHistoryContextMenu hook to include queryIndexes inside the Item?

packages/shared/src/components/PostOptionsReadingHistoryMenu.tsx
rebelchris: Yes I think it can be the same approach as we use now for the bookmark icon

rebelchris (changes requested):
Some small questions, but well done on the changes so far 🙌

packages/shared/src/components/history/ReadingHistoryList.tsx
sshanzel: Unfortunately, we can't get this through `readingHistoryContextItem` as the values needed are part of the data pages array. I think another way is to loop through every item to search for this specific history though I think it would be too much to loop through again. We can definitely refactor this once we can find another viable solution for this scenario but for now, imho this should be alright.

ayadotdev:
@rebelchris 
Screenshot 2022-02-17 at 12 45 21

I tried, I got this:

Screenshot 2022-02-17 at 12 43 52

packages/shared/src/components/PostOptionsReadingHistoryMenu.tsx
rebelchris: @ayadotdev Sorry remove the `_` so it's just [, copyPostLink]

packages/shared/src/components/PostOptionsReadingHistoryMenu.tsx
```diff
@@ -0,0 +1,148 @@
...
+const MenuIcon = ({ Icon }) => {
+  return <Icon className="mr-2 text-2xl" />;
+};
+
+/* eslint-disable no-param-reassign */
+const updateReadingHistoryPost =
+  (
+    queryClient: QueryClient,
+    historyQueryKey: QueryKey,
+    update: (oldPost: ReadHistoryPost) => Partial<ReadHistoryPost>,
```
sshanzel, line 35: Instead of a function, why not use a plain object here? I see you passed a node but was not necessarily used in the implementations.

packages/shared/src/components/PostOptionsReadingHistoryMenu.tsx
```diff
@@ -0,0 +1,148 @@
...
+            ...update(node.post),
+          };
+        }
+        return { node };
+      });
+      readHistory.edges = modified;
+      return {
+        readHistory,
+      };
+    });
```
sshanzel, line 54: We have access to the indexes outside, we can definitely pass them down as props to the actual component that uses this function. Then we can just directly find the element to update than looping through all the elements. What do you think? You can check sample usage in the `useReadingHistory`.

packages/shared/src/components/cards/PostMetadataReadingHistory.tsx
sshanzel: I agree with Chris. If this is displaying the same thing, let's use one component. I think the only change you need will be the type of `post` to accommodate the two scenarios.

packages/shared/src/hooks/useReadingHistoryContextMenu.ts
```diff
@@ -0,0 +1,40 @@
+import { useContextMenu } from '@dailydotdev/react-contexify';
```
sshanzel, line 1: I can see this file as something we can create a generic hook that accepts an id and a generic type to manage the actual data it receives and stored as a state.

We can refactor it now but feel free to just create a ticket to do it in the future as this still provides what we actually need.

sshanzel (changes requested):
Just had a few comments but you've got great progress here!

packages/shared/src/components/cards/PostMetadata.tsx
```diff
@@ -1,30 +1,29 @@
...
   className?: string;
   children?: ReactNode;
   typoClassName?: string;
 }
 
 export default function PostMetadata({
-  post,
+  createdAt,
+  readTime,
+  numUpvotes,
```
rebelchris, line 19: The numUpvotes is now never passed?

packages/shared/src/components/cards/PostMetadata.tsx
```diff
@@ -1,30 +1,29 @@
 import React, { ReactElement, ReactNode, useMemo } from 'react';
 import classNames from 'classnames';
-import { Post } from '../../graphql/posts';
+import { Post, ReadHistoryPost } from '../../graphql/posts';
```
rebelchris, line 3: I don't think you need these anymore.

packages/shared/src/components/cards/PostMetadata.tsx
```diff
@@ -1,30 +1,29 @@
...
-
-const Separator = classed(
-  'div',
-  'mx-1 w-0.5 h-0.5 rounded-full bg-theme-label-tertiary',
-);
+import { Separator } from './common';
 
 interface PostMetadataProps {
-  post: Post;
+  createdAt?: string;
```
rebelchris, line 8: I think the main benefit of using the `Post` interface is that we can get those properties from one place in case they might change.

Up to you, but optionally we could use a `Partial<Post, 'createdAt' | 'readTime' | 'numUpvotes>` type.

rebelchris (changes requested):
Some minor notes on the meta data, but well done on making it uniform 🙌

packages/shared/src/components/cards/PostMetadata.tsx
ayadotdev: In this case I was thinking that PostMetadata component could be only responsible to **how** to show the data and the components who are calling PostMetadata are responsible to **what** to show from the Post. WDYT?

packages/shared/src/components/cards/PostMetadata.tsx
ayadotdev: Yes, ReadingHistoryItem is passing numUpvotes.

sshanzel:
@sshanzel95 @sshanzel95

packages/shared/src/components/PostOptionsReadingHistoryMenu.tsx
ayadotdev: Maybe we can have a look at it together?

packages/shared/src/hooks/useReadingHistoryContextMenu.ts
ayadotdev: Thanks Lee for the suggestion. I created a separate task for this. https://dailydotdev.atlassian.net/browse/DD-557

packages/shared/src/components/PostOptionsReadingHistoryMenu.tsx
sshanzel: Whenever you are free! 🚀

packages/shared/src/components/cards/PostMetadata.tsx
```diff
@@ -35,10 +33,16 @@ export default function PostMetadata({
         className,
       )}
     >
-      {!!post.createdAt && <time dateTime={post.createdAt}>{date}</time>}
-      {!!post.createdAt && !!post.readTime && <Separator />}
-      {!!post.readTime && (
-        <span data-testid="readTime">{post.readTime}m read time</span>
+      {!!createdAt && <span>{date}</span>}
+      {!!createdAt && !!readTime && <Separator />}
+      {!!readTime && readTime >= 0 && (
```
sshanzel, line 38:
```suggestion
      {!!readTime && (
```

packages/shared/src/components/cards/PostMetadata.tsx
```diff
@@ -35,10 +33,16 @@ export default function PostMetadata({
...
+        <span data-testid="readTime">{readTime}m read time</span>
+      )}
+      {!!readTime && readTime >= 0 && numUpvotes >= 0 && <Separator />}
```
sshanzel, line 41:
```suggestion
      {(!!createdAt || !!readTime) && !!numUpvotes && <Separator />}
```

packages/shared/src/components/cards/PostMetadata.tsx
```diff
@@ -35,10 +33,16 @@ export default function PostMetadata({
...
+      {!!numUpvotes && numUpvotes >= 0 && (
```
sshanzel, line 42:
```suggestion
      {!!numUpvotes && (
```

sshanzel (approved):
Just had a few suggestions, but all looks good to me now! Well done on this PR!!! 🚢

packages/shared/src/components/cards/PostMetadata.tsx
rebelchris: Hmm I personally think it could give us issues by not leveraging the type already existing.
For instance readTime, let's say it's a number now, but we decide to make it a string because of some reason.

But either way will do really.

rebelchris (changes requested):
Almost there! 
I just have one small concern about naming, and a little tip for cleaner components (non-breaking)

I think we are about there where @idoshamun can have a look 🙌

packages/shared/src/components/PostOptionsReadingHistoryMenu.tsx
```diff
@@ -0,0 +1,141 @@
...
+
+const PortalMenu = dynamic(() => import('./fields/PortalMenu'), {
+  ssr: false,
+});
+
+export type PostOptionsReadingHistoryMenuProps = {
+  post: ReadHistoryPost;
+  onHidden?: () => unknown;
+  displayCopiedMessageFunc?: () => void;
+  onHide?: (postId: string) => unknown;
```
idoshamun, line 25: `onHide` and `onHidden` are quite confusing

packages/shared/src/components/PostOptionsReadingHistoryMenu.tsx
```diff
@@ -0,0 +1,141 @@
...
+  onHide,
+  indexes,
+}: PostOptionsReadingHistoryMenuProps): ReactElement {
+  const { user } = useContext(AuthContext);
+  const queryClient = useQueryClient();
+  const historyQueryKey = ['readHistory', user?.id];
+
+  const [, copyPostLink] = useCopyLink(() => post.commentsPermalink);
+
+  const { bookmark, removeBookmark } = useBookmarkPost({
```
idoshamun, line 70: I don't remember if `useBookmarkPost` publishes an analytics event but we do need to publish events here on bookmark or remove bookmark.

packages/shared/src/components/PostOptionsReadingHistoryMenu.tsx
```diff
@@ -0,0 +1,141 @@
...
+  post,
+  onHidden,
+  onHide,
+  indexes,
+}: PostOptionsReadingHistoryMenuProps): ReactElement {
+  const { user } = useContext(AuthContext);
+  const queryClient = useQueryClient();
+  const historyQueryKey = ['readHistory', user?.id];
+
+  const [, copyPostLink] = useCopyLink(() => post.commentsPermalink);
```
idoshamun, line 68: I'm not sure if we have a hook for sharing an article specifically. But I think we should because we need to publish an analytics event here. Also when possible should be open the native share process. This is the default user experience we want to promote, you can find it also in the article page

packages/shared/src/components/PostOptionsReadingHistoryMenu.tsx
```diff
@@ -0,0 +1,141 @@
...
+    }
+    return <MenuIcon Icon={BookmarkIcon} />;
+  };
+
+  const getBookmarkMenuText = () => {
+    if (post?.bookmarked) {
+      return 'Remove from bookmarks';
+    }
+    return 'Save to bookmarks';
+  };
```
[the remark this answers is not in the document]
idoshamun: I also suggest unifying them into one function that already returns both as a single element instead of two functions

packages/shared/src/components/PostOptionsReadingHistoryMenu.tsx
```diff
@@ -0,0 +1,141 @@
...
+    <>
+      <PortalMenu
+        disableBoundariesCheck
+        id="reading-history-options-context"
+        className="menu-primary"
+        animation="fade"
+        onHidden={onHidden}
+      >
+        <Item className="typo-callout" onClick={onBookmarkReadingHistoryPost}>
+          <a className="flex w-full typo-callout">
```
idoshamun, line 120: Why do you use an anchor element here?

packages/shared/src/components/PostOptionsReadingHistoryMenu.tsx
```diff
@@ -0,0 +1,141 @@
...
+            {getBookmarkIcon()}
+            {getBookmarkMenuText()}
+          </a>
+        </Item>
+        <Item className="typo-callout" onClick={copyPostLink}>
+          <a className="flex w-full typo-callout">
```
idoshamun, line 126: Why do you use an anchor element here?

packages/shared/src/components/PostOptionsReadingHistoryMenu.tsx
```diff
@@ -0,0 +1,141 @@
...
+            <MenuIcon Icon={ShareIcon} /> Share article
+          </a>
+        </Item>
+        <Item
+          className="laptop:hidden typo-callout"
+          onClick={() => onHide(post?.id)}
+        >
+          <a className="flex w-full typo-callout">
```
idoshamun, line 134: Why do you use an anchor element here?

packages/shared/src/components/cards/PostMetadata.tsx
idoshamun: Chris is right. You should use `Pick<Post, 'createdAt' | 'readTime' | 'numUpvotes'>` to make sure types are aligned throughout the app. You still get what you want @ayadotdev where the caller will be in charge of extracting the properties but it will be aligned with the right types

packages/shared/src/components/cards/PostMetadata.tsx
```diff
@@ -35,10 +33,16 @@ export default function PostMetadata({
         className,
       )}
     >
-      {!!post.createdAt && <time dateTime={post.createdAt}>{date}</time>}
-      {!!post.createdAt && !!post.readTime && <Separator />}
-      {!!post.readTime && (
-        <span data-testid="readTime">{post.readTime}m read time</span>
+      {!!createdAt && <span>{date}</span>}
```
idoshamun, line 36: Why don't we use the `time` element here anymore? Semantically it's the correct thing to do

idoshamun (changes requested):
Great PR! But we still have some work to do here before we can merge

packages/shared/src/components/PostOptionsReadingHistoryMenu.tsx
```diff
@@ -0,0 +1,141 @@
...
+        historyQueryKey,
+        oldReadingHistory,
+      );
+    };
+  };
+
+export default function PostOptionsReadingHistoryMenu({
+  post,
+  onHidden,
+  onHide,
```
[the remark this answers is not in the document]
ayadotdev: Great idea!

packages/shared/src/components/cards/PostMetadata.tsx
ayadotdev: @rebelchris @idoshamun I got your suggestions, it makes sense to align the types through the app. The issue is that some components display only some metadata, e.x. the card component doesn't display numUpvotes but displays createdAt and readTime, while the ReadingHistoryItem displays readTime, numUpvotes but not the createdAt. It means that to manage that logic the caller should pass some flags together with post metadata which seems a bit convoluted. WDYT?

packages/shared/src/components/PostOptionsReadingHistoryMenu.tsx
ayadotdev: @idoshamun we actually have a `useCopyPostLink` hook with `gaTrackEvent()` inside. Would that be enough?

packages/shared/src/components/PostOptionsReadingHistoryMenu.tsx
ayadotdev: @idoshamun I reused it from our PostOptionsMenu. Should we use something else?

packages/shared/src/components/PostOptionsReadingHistoryMenu.tsx
idoshamun: No, we don't need `gaTrackEvent` anymore actually. Maybe @rebelchris can help out with onboarding information for our analytics system

packages/shared/src/components/PostOptionsReadingHistoryMenu.tsx
idoshamun: Shouldn't be there as well I think. Anchor elements are for links not for general-purpose buttons. I think the `Item` component is enough

ayadotdev: You are right, the anchor is not needed. I still have to use span to enforce the styles.

idoshamun: Why can't you apply them on the `Item` directly?

ayadotdev: It seems that the text style in the Item is ignored if I do it.

idoshamun: It's worth checking the `Item` component itself in this case

packages/shared/src/components/PostOptionsReadingHistoryMenu.tsx
rebelchris: Sure we can do it tomorrow if you want?

packages/shared/src/components/cards/PostMetadata.tsx
rebelchris: I think that's why we should use optional characters.
Which Partial can do, depending on which one is always required?

I can help you set this type up if you want.
The main goal is to have shared types throughout the app, which we currently don't always do.

We could even mix with a Union type actually. So it can either be part of a Post or part of the ReadingHistoryItem?

sshanzel: I think what Aya is concerned about is that `post` in the feed will always have `numUpvotes` and it should not be displayed in the card's metadata (which will automatically be displayed if the whole object was passed).

ayadotdev: Yes, Lee, exactly.

rebelchris: Oh right, but that has nothing to do with the Type of the Props?
You can still pass the individual items for each one, so the feed is the only one that passes `numUpvotes`.

We are concerned about the types of `PostMetadataProps`.
Let me know if you want to have a chat about this.

packages/shared/src/components/PostOptionsReadingHistoryMenu.tsx
rebelchris: @idoshamun We have to wrap something inside the `Item` if another typo is needed.
We set the inner div with `typo-footnote` by default (CSS)

So the only way to use other fonts is to override it.
We might have to look into this if it's still needed actually. or it should be `typo-callout` by default.
But suggestion we open a separate ticket for that.

packages/shared/src/components/PostOptionsReadingHistoryMenu.tsx
rebelchris: Don't forget to change the actual trackEvent once we have the information.
I've added a TODO in the code for this.

packages/shared/src/components/PostOptionsReadingHistoryMenu.tsx
rebelchris: Don't forget to change the TrackEvent once we have the information.
I've added a TODO in the code for this.

packages/shared/src/components/PostOptionsReadingHistoryMenu.tsx
idoshamun: So let's use a `span` then as suggested by @ayadotdev

packages/shared/src/components/PostOptionsReadingHistoryMenu.tsx
```diff
@@ -0,0 +1,168 @@
...
+      {
+        bookmarked: false,
+      },
+      indexes,
+    ),
+  });
+
+  const onBookmarkReadingHistoryPost = async (): Promise<void> => {
+    const bookmarked = post?.bookmarked;
+    const trackEventFunction = trackEventFunc(
```
rebelchris, line 128: Nice one of reusing this!
I'm personally not a fan of `func` but it's not breaking so leaving it up to @idoshamun  if he wants that changed

rebelchris (approved):
Wow, just Wow Aya!
Super well done on pulling this one through 👏

Final word for @idoshamun 👀

packages/shared/src/components/PostOptionsReadingHistoryMenu.tsx
ayadotdev: I am not a fan either! 😅

packages/shared/src/components/PostOptionsReadingHistoryMenu.tsx
```diff
@@ -0,0 +1,168 @@
...
+  const queryClient = useQueryClient();
+  const historyQueryKey = ['readHistory', user?.id];
+
+  const [, copyLink] = useCopyLink(() => post.commentsPermalink);
+
+  const trackEventFunc = (eventName: string) => {
+    return () =>
+      trackEvent(
+        postAnalyticsEvent(eventName, post, {
+          extra: { origin: 'reading history post menu' },
```
idoshamun, line 95: Aligned with the existing naming

```suggestion
          extra: { origin: 'reading history context menu' },
```

packages/shared/src/components/PostOptionsReadingHistoryMenu.tsx
idoshamun: Instead of a function that returns a function, we can do a regular function that calls `trackEvent`, and then for `useShareOrCopyLink` we can use the `bind` function. Something like `trackEventFunc.bind(null, 'share post')` will generate a "new" function that receives no arguments. What do you think?

See here more info about it:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_objects/Function/bind

packages/shared/src/hooks/useShareOrCopyLink.ts
```diff
@@ -0,0 +1,30 @@
...
+      try {
+        await navigator.share({
+          text,
+          url: link,
+        });
+      } catch (err) {
+        // Do nothing
+      }
+    } else {
+      copyLink?.();
```
idoshamun, line 25: I think it's better to have the copy functionality here and then call a callback after the actual copy to show a notification of sorts

packages/shared/src/hooks/useShareOrCopyLink.ts
```diff
@@ -0,0 +1,30 @@
...
+  copyLink?: () => Promise<void>;
+}
+export function useShareOrCopyLink({
+  link,
+  text,
+  trackEvent,
+  copyLink,
+}: UseShareOrCopyLinkProps): () => Promise<void> {
+  const onShareOrCopy = async () => {
+    trackEvent?.();
```
idoshamun, line 14: I think it's better to track the event here with an actual function call to the real `trackEvent` and receive a function that generates the event instead of doing the call as well

idoshamun (changes requested):
Almost there!
I think I missed the additional fields in the reading history graphql query. Can you confirm you added them please?

packages/shared/src/hooks/useShareOrCopyLink.ts
rebelchris: I think the the difficult part is that some use the post copy hook, which will auto hide the message and some don't
That's why I thought it would be better to leave it up to the implementation to decide what copy function to use.

idoshamun: Make sense then. Let's keep the current implementation

packages/shared/src/components/PostOptionsReadingHistoryMenu.tsx
rebelchris: @idoshamun Out of interest what's the main argument to do that compared to how we had it?

idoshamun: @rebelchris I tried to provide a better solution following your comment that you are not a fan of the `func` thingy. So I provided an alternative that kind of suits this case better so we won't have to generate a function and then call it in this case.

rebelchris: Oh right i only meant the shortcut name "func" i hate seeing those for some reason 😂🤯

idoshamun: Oh ok then. Anyway, whatever works for you two. I don't mind just finding the `bind` solution a bit cleaner.

packages/shared/src/hooks/useShareOrCopyLink.ts
ayadotdev: I am not 100% sure what it means.

ayadotdev: @idoshamun I moved both track events inside the hooks. Let me know if this works for you.

packages/shared/src/hooks/useBookmarkPost.ts
```diff
@@ -1,15 +1,19 @@
...
   REMOVE_BOOKMARK_MUTATION,
 } from '../graphql/posts';
+import AnalyticsContext from '../contexts/AnalyticsContext';
 
 type MutateFunc<T> = (variables: T) => Promise<(() => void) | undefined>;
 type UseBookmarkPostParams<T> = {
   onBookmarkMutate: MutateFunc<T>;
   onRemoveBookmarkMutate: MutateFunc<T>;
+  onBookmarkTrackObject?: Record<string, unknown>;
+  onRemoveBookmarkTrackObject?: Record<string, unknown>;
```
idoshamun, line 16: One thing, it's better to have it as a function rather than an actual object.
Something like below. Because otherwise, we may provide some out-of-date values. Providing a function makes sure everything is up to date and generated only when necessary.
```suggestion
  onBookmarkTrackObject?: () => Record<string, unknown>;
  onRemoveBookmarkTrackObject?: () => Record<string, unknown>;
```

packages/shared/src/hooks/useShareOrCopyLink.ts
```diff
@@ -1,17 +1,22 @@
+import { useContext } from 'react';
+import AnalyticsContext from '../contexts/AnalyticsContext';
+
 interface UseShareOrCopyLinkProps {
   link: string;
   text: string;
-  trackEvent?: () => void;
   copyLink?: () => Promise<void>;
+  trackObject?: Record<string, unknown>;
```
idoshamun, line 8: Please see my comment above :)

packages/shared/src/lib/feed.ts
```diff
@@ -85,18 +85,18 @@ export function postAnalyticsEvent(
...
+    feed_item_title: post?.title,
+    post_author_id: post?.author?.id,
+    post_created_at: post?.createdAt,
+    post_comments_count: post?.numComments,
+    post_read_time: post?.readTime,
+    post_tags: post?.tags,
+    post_source_id: post?.source?.id,
+    post_trending_value: post?.trending,
+    post_upvotes_count: post?.numUpvotes,
+    target_id: post?.id,
```
idoshamun, line 99: Why is this needed? If `post` is not defined we better don't do anything in this case

idoshamun (changes requested):
Almost there! 💪

ayadotdev:
@idoshamun PR is ready, I addressed all the unresolved conversations. Please have a look and feel free to mark them resolved in case it is ok.

packages/shared/src/hooks/useBookmarkPost.ts
```diff
@@ -43,7 +43,7 @@ export default function useBookmarkPost<
     {
       onMutate: onBookmarkMutate,
       onError: (err, _, rollback) => rollback?.(),
-      onSuccess: () => trackEvent(onBookmarkTrackObject()),
+      onSuccess: () => trackEvent(onBookmarkTrackObject?.()),
```
idoshamun, line 46: If `onBookmarkTrackObject` is not defined we should not call `trackEvent`

```suggestion
      onSuccess: () => onBookmarkTrackObject && trackEvent(onBookmarkTrackObject()),
```

packages/shared/src/hooks/useBookmarkPost.ts
```diff
@@ -60,7 +60,7 @@ export default function useBookmarkPost<
     {
       onMutate: onRemoveBookmarkMutate,
       onError: (err, _, rollback) => rollback?.(),
-      onSuccess: () => trackEvent(onRemoveBookmarkTrackObject()),
+      onSuccess: () => trackEvent(onRemoveBookmarkTrackObject?.()),
```
idoshamun, line 63: Same here

```suggestion
      onSuccess: () => onRemoveBookmarkTrackObject && trackEvent(onRemoveBookmarkTrackObject()),
```

idoshamun (changes requested):
Last one :)

packages/shared/src/hooks/useBookmarkPost.ts
ayadotdev: @idoshamun done 👍

packages/shared/src/hooks/useBookmarkPost.ts
ayadotdev: @idoshamun done 👍

idoshamun (approved):
Yessss!!!

PR #906, opened 2022-02-18
feat: mention user in comments

sshanzel:
## Changes

Describe what this PR does
- created a common component for user's short info (image, name, username)
- once the user starts typing with at (@) character we will send a query and display the returned users
- when the user mentions someone, it must display differently by applying an additional style.

## Manual Testing

- On those affected packages:

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [x] Laptop (1020px)

DD-469 #done


sshanzel:
Should now be ready for review, just gotta work on those suddenly failing tests.

packages/shared/src/components/RecommendedMention.tsx
```diff
@@ -0,0 +1,48 @@
...
+  users,
+  selected,
+  onClick,
+}: RecommendedMentionProps): ReactElement {
+  if (!users?.length) {
+    return null;
+  }
+
+  return (
+    <ul
```
idoshamun, line 27: I'm not sure what should be the accessibility aspects of this component but you can have a look at Twitter and see if they added any attributes or keyboard behavior, etc. Please let me know your findings before you implement

packages/shared/src/components/modals/NewCommentModal.tsx
```diff
@@ -236,6 +258,8 @@ export default function NewCommentModal({
     ) {
       await sendComment(event);
     }
+
+    onMentionKeypress(event);
```
idoshamun, line 262: I think this is better

```suggestion
else {
    onMentionKeypress(event);
}  
```

packages/shared/src/components/modals/NewCommentModal.tsx
```diff
@@ -316,13 +345,42 @@ export default function NewCommentModal({
...
+            className: 'shadow',
+            paddingClassName: 'p-0',
+            roundedClassName: 'rounded-16',
+            bgClassName: 'bg-theme-bg-tertiary',
+          }}
+          reference={commentRef}
+          arrow={false}
+          placement="top-start"
+          visible={mentionQuery !== undefined}
+        />
```
idoshamun, line 369: This can go into its own component

packages/shared/src/components/modals/NewCommentModal.tsx
```diff
@@ -316,13 +345,42 @@ export default function NewCommentModal({
...
         style={{ minHeight: '1rem' }}
       >
         {errorMessage && <span role="alert">{errorMessage}</span>}
       </div>
-      <footer className="flex justify-between items-center py-2 border-t border-theme-divider-tertiary">
+      <footer className="flex items-center py-3 px-3 border-theme-divider-tertiary">
+        <Button
+          className="mx-1 border border-theme-label-primary btn-tertiary"
+          buttonSize="small"
+          icon={<AtIcon />}
```
idoshamun, line 381: Why is this a button? What happens on click?

packages/shared/src/hooks/useProfileTooltip.ts
```diff
@@ -0,0 +1,39 @@
...
+  USER_TOOLTIP_CONTENT_QUERY,
+} from '../graphql/users';
+import { apiUrl } from '../lib/config';
+
+interface UseProfileTooltip {
+  fetchInfo: () => unknown;
+  data?: UserTooltipContentData;
+}
+
+export const useProfileTooltip = (userId: string): UseProfileTooltip => {
```
idoshamun, line 15: I'd like to suggest a different design instead of `fetchInfo`. You should provide another argument `enabled: boolean` to the hook and then you can provide it to the `enabled` prop of react query.
The current design seems somewhat like an anti pattern. What do you think?

packages/shared/src/hooks/useUserMention.ts
```diff
@@ -0,0 +1,205 @@
...
+          query: mentionQuery,
+        }),
+      {
+        enabled: !!user,
+        refetchOnWindowFocus: false,
+        refetchOnMount: false,
+      },
+    );
+  const { recommendedMentions: mentionUsers } = data;
+  const fetchUsers = useDebounce(refetch, 300);
```
idoshamun, line 124: Is this the same way we handle throttling on the search page? I think we should use the same mechanism

packages/shared/src/hooks/useUserMention.ts
```diff
@@ -0,0 +1,205 @@
...
+  const filteredUsers = mentionUsers?.filter((recommendation) =>
+    mentions.every((mention) => `@${recommendation.username}` !== mention),
+  );
```
idoshamun, line 127: What do you check here?

packages/shared/src/hooks/useUserMention.ts
```diff
@@ -0,0 +1,205 @@
...
+  const { data = { recommendedMentions: [] }, refetch } =
+    useQuery<RecommendedMentionsData>(
+      key,
+      () =>
+        request(`${apiUrl}/graphql`, RECOMMEND_MENTIONS_QUERY, {
+          postId,
+          query: mentionQuery,
+        }),
+      {
+        enabled: !!user,
```
idoshamun, line 118: You need to check here if `mentionQuery` is not `undefined` as well

packages/shared/src/hooks/useUserMention.ts
```diff
@@ -0,0 +1,205 @@
+import { useQuery, useQueryClient } from 'react-query';
```
idoshamun, line 1: This file can use some more refactoring and better naming.
Some of the functions here are hard to follow and understand because there's quite a lot of complex logic here

packages/shared/src/components/Markdown.tsx
```diff
@@ -1,12 +1,83 @@
...
+export default function Markdown({ content }: MarkdownProps): ReactElement {
+  const [username, setUsername] = useState('');
+  const [instance, setInstance] = useState<TippyInstance>();
+  const queryKey = ['profile', username];
+  const { data: fetchedProfile } = useQuery<PublicProfile>(
+    queryKey,
+    () => getProfile(username),
+    { enabled: !!username },
+  );
+  const { fetchInfo, data } = useProfileTooltip(fetchedProfile?.id);
```
idoshamun, line 32: Here we do double fetch. We run `getProfile` and inside `useProfileTooltip` we ask for the profile tooltip info.
Why can't we do only one? Maybe instead of filling the username in `data-mention-username` we should fill the user id. It will save a call. What do you think?

idoshamun (changes requested):
Wow man! This is brilliant engineering right there. Of course, I have some comments but still that's a massive work. Way to go 💪

packages/shared/src/components/RecommendedMention.tsx
sshanzel: The noticeable thing on their structure looks to be utilizing a role attribute as `listbox` (select element) for the container and `option` role for each item. Then an aria attribute `aria-selected` for which one is currently selected (which now I think a must, to implement).

packages/shared/src/hooks/useUserMention.ts
sshanzel: It is currently using a timeout. Though their process flow is the same, should we update the search field implementations to utilize this hook?

idoshamun: On a different ticket perhaps

packages/shared/src/hooks/useUserMention.ts
sshanzel: Just so it won't have to, again, display the user that was already mentioned. Though after checking Twitter and GH, they allow users to be mentioned again in FE multiple times. In the backend though, it is already handled to be allowed. Let me abolish this.

packages/shared/src/components/RecommendedMention.tsx
```diff
@@ -0,0 +1,48 @@
...
+      className="flex overflow-hidden flex-col rounded-16 border border-theme-divider-secondary text-theme-label-primary"
+      style={{ width: '15rem' }}
```
rebelchris, line 29: Perhaps we can change this to be a JIT class?

packages/shared/src/components/modals/NewCommentModal.tsx
```diff
@@ -316,13 +346,42 @@ export default function NewCommentModal({
...
+      {getShouldLoadTooltip() && (
+        <BaseTooltip
+          content={
+            <RecommendedMention
+              users={mentionQuery !== undefined && mentions}
+              selected={selected}
+              onClick={onMentionClick}
+            />
+          }
+          offset={[getOffsetPerChar(offsetX), getOffsetPerLine(offsetY) + -90]}
```
rebelchris, line 358: Does this not contradict when people are zoomed/using weird fonts? 
(not sure, might be worth checking)

packages/shared/src/hooks/useUserMention.ts
```diff
@@ -0,0 +1,205 @@
...
+
+  const onMention = (username: string) => {
+    const query = `@${mentionQuery}`;
+    const replacement = `@${username}`;
+    const element = commentRef?.current;
+    const content = element.innerHTML.replace(query, replacement);
+    // eslint-disable-next-line no-param-reassign
+    element.innerHTML = content;
+    setMentionQuery(undefined);
+    client.setQueryData(key, []);
```
rebelchris, line 137: Not a 100% sure, but you might also want to invalidate the query at this point.
Perhaps we can test it to see if the query keeps refreshing for instance?

rebelchris (changes requested):
Wow Lee, really impressive work!
Looking forward to testing this out.

I added some small remarks from my side, let me know what you think

packages/shared/src/components/RecommendedMention.tsx
rebelchris: Ah yes probably, what mostly works is stopping the server, removing the .next file in your webapp and re-running.

packages/shared/src/hooks/useUserMention.ts
sshanzel: I agree a better way would be invalidating the query - You are right, it won't request again due to invalidation since the `mentionQuery` is already `undefined`. Though we won't be able to actually create a test for it as it resides inside a tooltip which we actually can't test since then. This is why I had to test the specific component on its own.

packages/shared/src/hooks/useUserMention.ts
sshanzel: Alright, let me create one for it.

rebelchris (commented):
Also don't forget to add some test for anonymous users as the call needs auth.

packages/shared/src/components/RecommendedMention.tsx
sshanzel: Worked! After re-run and deleting in-between. Thanks!

packages/shared/src/components/modals/NewCommentModal.tsx
sshanzel: Looks like we're good to go! 

Screen Shot 2022-03-07 at 10 37 46 PM

packages/shared/src/hooks/useUserMention.ts
sshanzel: Unfortunately, even though the `mentionQuery` is already undefined. I have to keep setting it as an empty array to avoid the additional request and flick on the UI when invalidating.

packages/shared/src/hooks/useProfileTooltip.ts
sshanzel: Was actually thinking about that, what made me do it this way is, that we won't have to manage an additional state outside for every consumer utilizing this hook.

idoshamun: Ok then. Let's keep it this way :)

packages/shared/src/components/RecommendedMention.tsx
sshanzel: For now, I will just apply the `aria-selected` attribute as it seems to be a must for this scenario. What do you think?

idoshamun: I think the roles are also important so apply them as well, please.
They turn this component into a sort of input component, dropdown to be more accurate. So it's important to set these.

packages/shared/src/components/RecommendedMention.tsx
sshanzel: @rebelchris looks like on our build it doesn't generate the arbitrary value. Have you had any situation where you were able to make it work in production?

Preview deployment:
Screen Shot 2022-03-08 at 8 02 16 PM

sshanzel:
Should now be ready for another round of review! 🎉 🚢 

@idoshamun I have refactored the `useUserMention` and extracted some generic functions. With the filename though, I am not sure what to use for this one. Do you have one in mind?

@rebelchris we have a login barrier in our text comment modal popup, so anonymous users won't be able to type anything. I think the test for auth in this scenario shouldn't be necessary anymore. What do you think?

packages/shared/src/components/RecommendedMention.tsx
rebelchris: I haven't seen it be an issue in production.
But don't let this block you, rather move back to style solution and add a separate ticket for this 😮‍💨

sshanzel:
Thank you both for the awesome review and kind words! 🙇  this should now be ready for review! 🎉 🚢 

@idoshamun I have now refactored the `useUserMention` and extracted the generic functions in a separate module. As for the filename, I'm not sure what to replace, do you have one in mind?

@rebelchris we have a login barrier for our comment modal popup, I think it should not be necessary anymore to test the anonymous user. What do you think?

packages/shared/src/components/RecommendedMention.tsx
sshanzel: Hmm. Let me clear my cache once I come back, could be just a cache issue on my end if it will be working on yours.

sshanzel:
@rebelchris for testing different scenarios based on the sequence of letters. The result would be solely dependent on our API (integration testing). I think that scenario was covered in the API PR. Mocking the results I think won't provide much accuracy to whether the output was actually correct. What do you think?

idoshamun:
@sshanzel I think it's better to wait with the review until we will conclude the manual testing.
One point I did notice, we have some users with empty usernames because they didn't finish registration. We should exclude them. It's on the API side obviously.

rebelchris:
@sshanzel yep, don't worry about that test, we'll figure those out in the preview tests manually.

sshanzel:
Whew! This should now fix the edge cases for mentioning someone. Thanks a lot to Chris for listing out a lot of it as well. [Issues from this thread](https://dailydotdev.slack.com/archives/C01L0QXQJNB/p1646743316148079).

Should be ready for review 🚢

sshanzel:
This should now be ready for review 🙌 

cc @idoshamun @rebelchris @ayadotdev

packages/shared/src/components/Markdown.tsx
```diff
@@ -1,12 +1,81 @@
...
+
+interface MarkdownProps {
   content: string;
-}): ReactElement {
+}
+
+type TippyInstance = ReturnType<typeof tippy>[0];
+
+const getUserPermalink = (username: string) =>
+  `${process.env.NEXT_PUBLIC_WEBAPP_URL}${username}`;
```
rebelchris, line 19: Just thinking ahead, perhaps we can move this to a more generic file and export it?
This seems like something we'll need more often.

packages/shared/src/components/modals/NewCommentModal.tsx
```diff
@@ -265,7 +285,12 @@ export default function NewCommentModal({
 
   return (
     <ResponsiveModal
-      {...{ contentRef: modalRef, onRequestClose: confirmClose, ...props }}
+      {...{
+        contentRef: modalRef,
+        onRequestClose: confirmClose,
+        style: { content: { overflow: 'initial' } },
```
rebelchris, line 291: Do we need the overflow initial here? 
Could we not use `overflow-auto` here?

packages/shared/src/components/profile/ProfileTooltipContent.tsx
```diff
@@ -42,11 +42,11 @@ export function ProfileTooltipContent({
           {user.bio}
         </p>
       )}
-      {tags?.length && (
+      {tags?.length ? (
```
rebelchris, line 45: Why do we need the hardcoded null return?

rebelchris (changes requested):
Getting so close, i'll do some UI testing now

packages/shared/src/components/Markdown.tsx
sshanzel: I agree this should be in a different place.

packages/shared/src/components/modals/NewCommentModal.tsx
sshanzel: In the past, we can't. I have now introduced a change that makes this unnecessary anymore 🚢

packages/shared/src/components/profile/ProfileTooltipContent.tsx
sshanzel: It displays the string `0` when the length is 0 instead of conditionally rendering. If we turn it into a boolean, it might return the word `false`.

packages/shared/src/components/Markdown.tsx
```diff
@@ -1,12 +1,79 @@
...
+interface MarkdownProps {
   content: string;
-}): ReactElement {
+}
+
+type TippyInstance = ReturnType<typeof tippy>[0];
+
+export default function Markdown({ content }: MarkdownProps): ReactElement {
+  const [userId, setUserId] = useState('');
+  const [instance, setInstance] = useState<TippyInstance>();
```
idoshamun, line 21: Why do we keep it in the state?
I see only where we assign a value to `instance` but not where we use it

packages/shared/src/components/profile/UserShortInfo.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+  return (
+    <Element
+      {...props}
+      className={classNames(
+        'flex flex-row py-3 px-6 hover:bg-theme-hover',
+        className,
+      )}
+    >
+      <ProfilePicture user={{ username, image }} size={imageSize} />
```
idoshamun, line 47: The following `{{ username, image }}` always creates a new object.
It's better to store it in `useMemo`

packages/shared/src/components/tooltips/RecommendedMentionTooltip.tsx
```diff
@@ -0,0 +1,64 @@
...
+      container={{
+        className: 'shadow',
+        paddingClassName: 'p-0',
+        roundedClassName: 'rounded-16',
+        bgClassName: 'bg-theme-bg-tertiary',
+      }}
+      reference={elementRef}
+      arrow={false}
+      placement="top-start"
+      visible={query !== undefined}
```
idoshamun, line 61: I'm not sure this is the right way to check for `undefined`. Please confirm it

packages/shared/src/hooks/useUserMention.ts
```diff
@@ -0,0 +1,231 @@
...
+          return;
+        }
+
+        event.preventDefault();
+        onMention(mentions[selected].username);
+        return;
+      }
+
+      if (event.key === 'Backspace') {
+        await nextTick();
```
idoshamun, line 135: Why do we need this `nextTick` all over?

packages/shared/src/hooks/useUserMention.ts
```diff
@@ -0,0 +1,231 @@
...
+
+interface UseUserMentionProps {
+  postId: string;
+  onInput: (content: string) => unknown;
+  commentRef?: MutableRefObject<HTMLDivElement>;
+}
+
+const ARROW_KEYS = ['ArrowUp', 'ArrowDown'];
+
+export function useUserMention({
```
idoshamun, line 49: The business logic here of how you handle the different chars like `Backspace` and other long code sections need more documentation I think. It will be hard to understand it in a few days.

idoshamun (changes requested):

packages/shared/src/components/Markdown.tsx
sshanzel: We will be using it in the very last part of generating the tooltip content. Below is line:

https://github.com/dailydotdev/apps/blob/fc87a2724404f9a9c0ef33e217fdca1c22f967f7/packages/shared/src/components/Markdown.tsx#L74

packages/shared/src/components/tooltips/RecommendedMentionTooltip.tsx
sshanzel: I think it would be much safer to check the type. Let me refactor it.

packages/shared/src/hooks/useUserMention.ts
sshanzel: I used to have this to utilize the `commentRef` as it is not updated the same time as when the user hits a key. The values are not matched. With that, I have reduced the usage and ensured things are still working as intended.

packages/shared/src/hooks/useUserMention.ts
sshanzel: I have now added more information relating to this on the specs page in the product section at point number 5.

packages/shared/src/hooks/useUserMention.ts
idoshamun: Hopefully there's no race condition here

sshanzel: Yes, I think there should be none, this is used now only at the initial phase.

packages/shared/src/hooks/useUserMention.ts
```diff
@@ -91,6 +92,10 @@ export function useUserMention({
   };
 
   const onMention = (username: string) => {
+    if (mentions.length === 0) {
```
idoshamun, line 95: Safer, I think :)

```suggestion
    if (!mentions?.length) {
```

packages/shared/src/components/tooltips/RecommendedMentionTooltip.tsx
```diff
@@ -39,7 +39,7 @@ export function RecommendedMentionTooltip({
       appendTo={document?.body || 'parent'}
       content={
         <RecommendedMention
-          users={query !== undefined && mentions}
+          users={typeof query !== 'undefined' && mentions}
```
idoshamun, line 42: I think I saw more places where you checked for `undefined` like this `query !== undefined`.
Can you please validate?

idoshamun (changes requested):
Last review!

idoshamun (approved):
Yes!!!! Nailed it 💪

rebelchris (approved):
Nice work Lee 🛳

idoshamun (approved):
Much more simple now! I love it :)

rebelchris (approved):
Wow Lee, this is stunning so clean with the textareas!
From my side good to go 💖

PR #907, opened 2022-02-23
fix: do not show new rank on opt-out gamification

sshanzel:
## Changes

Describe what this PR does
- Will now check if the user opted-out of the gamification to never show the new rank popup modal
- Centralized the variable whether to display the rank popup instead of the component deciding it.
- Also refactored the `checkShouldShowRankModal` to bring outside the absolute conditions to reduce the number of parameters
- Rank new popup is disabled when the settings modal is opened.

## Manual Testing

- On those affected packages:

- [x] MobileL (420px)
- [x] Tablet (656px)
- [x] Laptop (1020px)

DD-{number} #done


rebelchris (changes requested):
For the sidebar this was already captured in the sidebar file.
```js
{sidebarRendered && !optOutWeeklyGoal && (
              <SidebarRankProgress sidebarExpanded={sidebarExpanded} />
            )}
```

Perhaps we should rather use that way for the header one as well? *(Else we'll still show the rank badge on mobile actually)

The actually intention for this ticket was the edge case when toggling from optOut to optIn, and it looks as that would still happen in this scenario?

However, super happy you actually found another bug on mobile 👀

sshanzel:
@rebelchris thank you for the notes. I have now included a condition to not level up when the weekly goal is disabled.

rebelchris:
@sshanzel I've tested it and the main issue for this ticket is still at hand.
Let me know if you want me to show you?

Screenshot 2022-02-23 at 09 19 22

sshanzel:
Oh, how were you able to reproduce it? I tried it locally, and it seemed to be working as intended.  I understand this is for toggling the opt-out of the weekly goal and suddenly this popup would show.

rebelchris:
@sshanzel Yes exactly!

- Make sure you are about to level up and you enabled the new rank modal!
- Opt out
- Add a view in the DB (It won't trigger with click, unless you are running the debezium?)
- Refresh page
- Opt in
- Popup shows

sshanzel:
Can you try to re-run the dev in the terminal after pulling? I tried again just now, here's a sample recording:

https://user-images.githubusercontent.com/13744167/155277450-8c8acace-225f-41b8-a8fb-43d5d3385a48.mov

sshanzel:
After digging through the `useReadingRank` - I am starting to think that we should create a context provider for rank-related concerns so we can centralize all the requests and the data used across the app and be in sync with each other. Currently, each instance makes its own requests and updates of the cache, which is quite redundant and unpredictable.

With that, this PR should now be ready to fix the primary issue at hand.

cc @rebelchris @idoshamun

rebelchris:
@sshanzel It indeed does no longer show the popup, however the rank badge freezes on this:
Screenshot 2022-02-23 at 17 41 40

Until i refresh and then i'll get the modal (which is good)
But I think this might have introduced a new edge case

sshanzel:
@rebelchris the latest should have covered it now. Looking forward to your review! 🚢

rebelchris (approved):
@sshanzel Sorry took me a while to test is, feel free next time to ping me to have a look 😅
It all seems to work fine, if you can fix the test we should be ready to go

rebelchris:
@sshanzel I see this is the flaky test, I'll log a ticket for that separate tomorrow morning.
It happens every other time, so probably runs out of sync or something.

You can go ahead and merge this 💖

sshanzel:
@rebelchris wow, you've got some magic hands. I've re-run this multitple times and no luck 🤣

PR #911, opened 2022-03-01
feat: add dynamic signup button color

ayadotdev:
## Changes

This PR adds dynamic signup button color class.

Screenshot 2022-03-01 at 09 56 38

## Manual Testing

- On those affected packages:

DD-407 #done


packages/shared/src/components/LoginButton.tsx
```diff
@@ -43,7 +48,7 @@ export default function LoginButton({
     <Button
       onClick={onClick}
       icon={icon}
-      className={classNames('btn-primary', className)}
+      className={classNames('btn-primary', className, buttonColor.button)}
```
rebelchris, line 51: This buttonColor should actually overwrite the btn-primary, else we have contradicting classes.

packages/shared/src/lib/featureManagement.ts
```diff
@@ -6,6 +6,11 @@ export class Features {
     'Access all features',
   );
 
+  static readonly SignupButtonColor = new Features(
+    'signup_button_theme_color',
+    'primary',
+  );
```
rebelchris, line 12: Think about this, something I forgot to mention is that we now also have `validOptions` as the third parameter here.
This should be an array of the options we allow to be valid.

```suggestion
  static readonly SignupButtonColor = new Features(
    'signup_button_theme_color',
    'primary',
    [
    'primary',
    'color1',
    'color2',
  ]
  );
```

rebelchris (changes requested):
Awesome, looks great just two minor changes and we should be ready to go

rebelchris (approved):
Nice work!
@sshanzel Would you mind having a look before we merge

sshanzel (approved):
Looks perfect to my eyes! 🤩 🚢

PR #912, opened 2022-03-02
fix: dnd banner missing when active

sshanzel:
## Changes

Describe what this PR does
- refactored the DndContext to have a custom provider to have easier maintainability.
- moved from `usePersistentState` to `usePersistentContext`
- New property on main feed layout to accept custom banner.

Screen Shot 2022-03-02 at 2 35 34 PM

Screen Shot 2022-03-02 at 2 35 43 PM

Screen Shot 2022-03-02 at 2 35 50 PM

## Manual Testing

- On those affected packages:

- [x] MobileL (420px)
- [x] Tablet (656px)
- [x] Laptop (1020px)

DD-563 #done


rebelchris (approved):
Looks great @sshanzel 

Just would like to ask you to test if the normal banners still work as intended and to double check the webapp with those.

packages/extension/src/newtab/DndContext.tsx
```diff
@@ -0,0 +1,38 @@
+import usePersistentContext from '@dailydotdev/shared/src/hooks/usePersistentContext';
+import React, { ReactElement, ReactNode } from 'react';
+
+export type DndSettings = { expiration: Date; link: string };
+
+export interface DndContextData {
+  dndSettings: DndSettings;
+  isActive: boolean;
+  onDndSettings: (settings: DndSettings) => Promise<void>;
```
idoshamun, line 9: Why did you decide to rename it to `on` instead of `set`?

idoshamun (approved):
Non blocking question. Other than that, we're good to go :)

packages/extension/src/newtab/DndContext.tsx
sshanzel: Just wanted to make things are uniform and easier to understand. We usually use the prefix `set` for state-related inside the component itself, to easily know that it is coming from either props or a destructured object.

PR #913, opened 2022-03-02
refactor: add dynamic post heading font

ayadotdev:
## Changes

This PR adds dynamic post heading font class for Post Card mode and Post List mode.

image-20220207-105127

## Manual Testing

- On those affected packages:

DD-525 #done


packages/shared/src/components/utilities.tsx
```diff
@@ -70,10 +75,30 @@ const themeColors = {
...
   },
 };
+
 export const getThemeColor = (color: string, fallback: string): ThemeColor => {
   return themeColors[color] ?? themeColors[fallback];
 };
 
+const themeFonts = {
+  bodyBold: {
+    heading: 'font-bold typo-body',
```
rebelchris, line 85: I actually think we can omit the heading from this, as there will not be any other versions.

packages/shared/src/components/cards/PostCard.tsx
```diff
@@ -69,12 +73,17 @@ export const PostCard = forwardRef(function PostCard(
...
     showImage = true,
     style,
+    flags,
     ...props
   }: PostCardProps,
   ref: Ref<HTMLElement>,
 ): ReactElement {
   const [selectedComment, setSelectedComment] = useState<Comment>();
   const { trending } = post;
+  const postHeadingFont = getThemeFont(
```
rebelchris, line 83: I think it might be better to load it in the Feed.tsx file and then pass down to FeedItemComponent as a option?
That way we only have to load it once.

rebelchris (changes requested):
Nice work, just some loading optimilizations

packages/shared/src/components/utilities.tsx
ayadotdev: @rebelchris please have a look and resolve if it is ok.

rebelchris (approved):
Lovely, well done!
Let's just see if @sshanzel spots anything, but this looks golden to me 🙌

packages/shared/src/components/utilities.tsx
```diff
@@ -70,10 +71,22 @@ const themeColors = {
...
 };
+
 export const getThemeColor = (color: string, fallback: string): ThemeColor => {
   return themeColors[color] ?? themeColors[fallback];
 };
 
+const themeFonts = {
+  bodyBold: 'font-bold typo-body',
+  body: 'typo-body',
+  title: 'typo-title3',
```
sshanzel, line 82: I suggest we either add the number itself or the number word so the property will be accurate because we have an actual class for `typo-title`. Kindly coordinate with the product team so they can be wary of the values they actually use.
```suggestion
  title3: 'typo-title3',
```

packages/shared/src/components/utilities.tsx
```diff
@@ -70,10 +71,22 @@ const themeColors = {
...
+  titleBold: 'font-bold typo-title3',
```
sshanzel, line 83: I suggest the same here
```suggestion
  title: 'typo-title3',
  title3Bold: 'font-bold typo-title3',
```

sshanzel (changes requested):
Just had a few comments, but everything is looking tight!

sshanzel (approved):
LGTM! 🚢

packages/shared/src/components/utilities.tsx
ayadotdev: I changed the names 👍  @rebelchris Is it ok to use a number in a variable name or should we as an actual word for it?

rebelchris: All good for me 🙌

PR #918, opened 2022-03-09
fix: remove all third party tracking

rebelchris:
## Changes

- Removing all the third party tracking we have
- Google Analytics
- Amplitude
- Twitter tracking pixel

## Manual Testing

- On those affected packages:

- Did you test the modified components media queries? (N/A)
- [x] MobileL (420px)
- [x] Tablet (656px)
- [x] Laptop (1020px)

DD-363 #done
DD-560 #done


sshanzel (approved):
It's so satisfying to see all the deleted files and lines of codes ✨ 

Nothing to add from my side really, great job on this! 💪

packages/shared/src/components/Feed.spec.tsx
```diff
@@ -534,39 +534,6 @@ it('should increase reading rank progress for anonymous users', async () => {
   });
 });
 
-it('should not increase reading rank progress for anonymous users after the first rank', async () => {
```
idoshamun, line 537: This might have blended into it this PR? It isn't related to this scope.

packages/shared/src/hooks/useAdImpressions.ts
```diff
@@ -1,7 +1,6 @@
 import { isSameDay } from 'date-fns';
 import usePersistentState from './usePersistentState';
 import { Ad } from '../graphql/posts';
-import { logRevenue, gaTrackEvent } from '../lib/analytics';
```
idoshamun, line 4: Please check if we need this file at all. I'm not sure we need this hook anymore

packages/shared/src/hooks/useCopyPostLink.ts
```diff
@@ -8,11 +7,6 @@ export function useCopyPostLink(): [boolean, () => Promise<void>] {
   return [
     copying,
     () => {
```
idoshamun, line 9: We can simply return `copy`. No need for anonymous function

idoshamun (changes requested):
You must tell me how it felt to delete so many lines of code 😆

packages/shared/src/hooks/useAdImpressions.ts
rebelchris: Ah yes looks like we don't need this at all as it only kept the count, but we don't do anything with that now.

rebelchris:
@idoshamun Fully removed the ad hook, is there any other way the ads know seen count now?
It did feel good to remove all this stuff, but also scary 🤯

idoshamun (approved):
Once merged please check that ads impressions and events in general are being tracked.
Just a sanity check

PR #919, opened 2022-03-09
refactor: on boot provider load apply theme

sshanzel:
## Changes

Describe what this PR does
There is a flick of black to white when on light mode at extension, though after further investigation, it seems to be unavoidable. I checked the previous implementation before introducing the major changes but it seems to flick still for a microsecond. So I took back the previous implementation (as I felt the flick was much quicker) and adjusted a bit on when we triggered the application of the theme (when we first load data from boot/cache).
At webapp though, it seems to work as intended.

I even created a production build that renders nothing but applies the light theme on `App.js` just to see if it would actually stop flicking, but it didn't.

## Manual Testing

- On those affected packages:

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [x] Laptop (1020px)

DD-{number} #done


rebelchris:
@sshanzel just thinking out loud, could we not introduce something in the content script? This fires quicker and it might be able to inject if we read directly from local storage?

I'm using this example for the companion and haven't seen it flicker there in the different modes.

sshanzel:
@rebelchris that sounds a great idea! Do you have some sample of your work I can check as reference?

rebelchris:
@sshanzel Not sure if it helps, but this is how I used it in the companion:
https://github.com/dailydotdev/apps/blob/DD-458-companion-extension/packages/extension/src/companion/index.tsx

sshanzel:
I still see some flicker but it is much faster now. I think I will proceed with your recommended process. Thank you for the reference!

rebelchris (approved):
Nice, love how we changed it to load from outside the app.
We just have to validate we didn't include unwanted unsets.

Meaning it could be if you have multiple tabs open they override changed data now.
Could you make sure to test that use-case?

Otherwise the code looks great like this.

sshanzel:
Looks like we still got things covered 🙌 

https://user-images.githubusercontent.com/13744167/159444826-c269c17e-a2ce-4efe-8a00-4b54bbcebd47.mov

packages/shared/src/contexts/BootProvider.tsx
```diff
@@ -54,13 +58,20 @@ function filteredProps<T extends Record<string, unknown>>(
 export type BootDataProviderProps = {
   children?: ReactNode;
   app: string;
+  localBootData?: BootCacheData;
   getRedirectUri: () => string;
 };
 
-const getLocalBootData = () => {
+export const getLocalBootData = (isInitialLoad = false): BootCacheData => {
```
idoshamun, line 65: This function should not apply the theme but only return what is stored in the cache

packages/shared/src/contexts/BootProvider.tsx
```diff
@@ -86,11 +97,13 @@ const updateLocalBootData = (
...
   children,
   app,
+  localBootData,
   getRedirectUri,
 }: BootDataProviderProps): ReactElement => {
   const queryClient = useQueryClient();
-  const [cachedBootData, setCachedBootData] =
-    useState<Partial<BootCacheData>>();
+  const [cachedBootData, setCachedBootData] = useState<Partial<BootCacheData>>(
+    () => localBootData || getLocalBootData(true),
```
idoshamun, line 105: If we assume `localBootData` is always provided we don't need the second part. Ideally we will read the cache only once

idoshamun (changes requested):
Please check my comments also I didn't find the equivalent fix for the webapp

packages/shared/src/contexts/BootProvider.tsx
sshanzel: Makes total sense. Let me work it out.

packages/shared/src/contexts/BootProvider.tsx
sshanzel: In the instance of webapp which the flicker doesn't happen anymore, it doesn't make the fetch in a much earlier time. But we can definitely do it as well when the webapp initializes.

idoshamun (approved):
Looks great! Please validate that everything works once merged :)

sshanzel:
Confirmed workability!

PR #920, opened 2022-03-15
feat: comment markdown preview

sshanzel:
## Changes

Describe what this PR does
- Introduced a new component for creating tabs 
- Apply tabs component in the comment modal
- Make a request on generating the Markdown.

Preview:
Screen Shot 2022-03-15 at 8 45 06 PM

## Manual Testing

- On those affected packages:

- [x] MobileL (420px)
- [x] Tablet (656px)
- [x] Laptop (1020px)

DD-468 #done


packages/shared/src/components/modals/NewCommentModal.tsx
```diff
@@ -263,90 +266,117 @@ export default function NewCommentModal({
...
+
   return (
     <ResponsiveModal
       {...{ contentRef: modalRef, onRequestClose: confirmClose, ...props }}
+      padding={false}
+      contentClassName="p-2"
     >
-      <article
-        className={`flex flex-col items-stretch ${commentBoxClassNames}`}
+      <TabContainer
```
rebelchris, line 285: Maybe we should move the tabs to their own component, this is getting quite big now.
Let me know what you think

packages/shared/src/components/modals/NewCommentModal.tsx
```diff
@@ -263,90 +266,117 @@ export default function NewCommentModal({
...
-      </div>
-      <div className="flex px-2">
-        <ProfilePicture user={user} size="small" />
-        <div
-          className={classNames(
-            'ml-3 flex-1 text-theme-label-primary bg-none border-none caret-theme-label-link whitespace-pre-line break-words break-words-overflow typo-subhead',
-            styles.textarea,
+          <div
+            className="my-2 mx-3 text-theme-status-error typo-caption1"
+            style={{ minHeight: '1rem' }}
```
rebelchris, line 344: Can't we add a class for this somehow?

packages/shared/src/components/modals/NewCommentModal.tsx
```diff
@@ -263,90 +266,117 @@ export default function NewCommentModal({
...
+            </Button>
+          </footer>
+          <DiscardCommentModal
+            isOpen={showDiscardModal}
+            onRequestClose={() => setShowDiscardModal(false)}
+            onDeleteComment={onRequestClose}
+            shouldCloseOnOverlayClick={false}
+          />
+        </Tab>
+        <Tab label="Preview" style={{ minHeight: '20rem' }}>
```
rebelchris, line 374: Same here, the minHeight is a bit of a pain to look at.

rebelchris (changes requested):
Looks pretty slick Lee!
Did you check this in all possible comment boxes? I think we have 2 different implementations.

packages/shared/src/components/modals/NewCommentModal.tsx
sshanzel: I think we have been moving away from CSS modules right? Probably the reason why the `1rem` was left there since then. Unfortunately, there could be some bug in the arbitrary value of tailwindcss and it results to be not working. What do you think an option we can take other than this?

packages/shared/src/components/modals/NewCommentModal.tsx
sshanzel: Great idea! Let me extract this one out.

sshanzel: Great idea! Let me work on that 🚢

packages/shared/src/components/modals/NewCommentModal.tsx
rebelchris: Hmm we really need to get this tailwind to work again, don't worry if that's blocking this.
We'll have to refactor quite a few things for that.

sshanzel: Think I should create a ticket for this so once we get tailwind working again, we will migrate everything at once. What do you think?

rebelchris (approved):
Love it!
Makes the file way more readable.

I do feel we could extract the render of the `CommentBox` a bit more.
Perhaps with in file components? (non-blocking, but love to hear what you think)

sshanzel:
Totally agree with your thoughts. It could use more destructuring. Thank you so much for the amazing review! 🚢

PR #921, opened 2022-03-16
refactor: search input guideline

sshanzel:
## Changes

Describe what this PR does
- Search field has now two types: Primary and Secondary
- Right icon now varies based on the field type
- Updated existing implementations to utilize the built-in support of the new guideline on different cases.
- New SVG assets for the magnifying icon.

Primary:
Screen Shot 2022-03-16 at 8 33 18 PM

Secondary:
Screen Shot 2022-03-16 at 8 33 30 PM
Screen Shot 2022-03-16 at 8 33 51 PM
Screen Shot 2022-03-16 at 8 34 09 PM

As part of our new guideline, the following issue got fixed as well. We now conditionally render the clear button instead of keeping it in the DOM: https://github.com/dailydotdev/daily/issues/342

## Manual Testing

- On those affected packages:

- [x] MobileL (420px)
- [x] Tablet (656px)
- [x] Laptop (1020px)

DD-254 #done


packages/shared/src/components/fields/SearchField.tsx
```diff
@@ -73,6 +81,15 @@ export const SearchField = forwardRef(function SearchField(
...
   };
 
+  const isPrimary = fieldType === 'primary';
+  const isSecondary = fieldType === 'secondary';
+  const SearchIcon = focused ? MagnifyingFilledIcon : MagnifyingOutlineIcon;
+  const buttonIcon = isPrimary ? (
+    <XIcon className="text-lg icon group-hover:text-theme-label-primary" />
+  ) : (
+    <ArrowIcon style={{ transform: 'rotate(90deg)' }} />
+  );
```
rebelchris, line 91: Perhaps move this one out to be it's own small inline component?
Also the rotate we should have as a tailwind class out of the box right?

packages/shared/src/components/fields/SearchField.tsx
```diff
@@ -111,21 +140,26 @@ export const SearchField = forwardRef(function SearchField(
...
       />
-      {rightChildren || (
-        <button
-          type="button"
+      {((hasInput && isPrimary) || isSecondary) && (
+        <Button
+          {...rightButtonProps}
+          className={isSecondary ? 'btn-primary' : 'btn-tertiary'}
+          buttonSize="xsmall"
           title="Clear query"
```
rebelchris, line 154: Should we not make this optional for the future?

rebelchris (changes requested):
Nice work Lee,
Love the fact that we have consise 2 clear types now.

And nice that you launched the first double-state icon 🙌
Just some minor questions

rebelchris (changes requested):
Some test based remarks:

- Field border radius should be 14px, it's now 12px
- The search icon states it should be 11 (assuming 12px) and can see only 8
Screenshot 2022-03-17 at 09 12 59
- Primary focus is missing the border
- Please double check these for the secondary as well, as I can't test those.

sshanzel:
For the border-radius, it actually changes based on the size of the actual input. Medium is for `12px` and for Large is `14px. 

The icon actually only has `8px` as you will have to focus from the icon to the relative element for checking.
Screen Shot 2022-03-17 at 3 38 34 PM

Let me check why the border is not reflecting as intended. Thank you for the review!

packages/shared/src/components/fields/SearchField.tsx
sshanzel: Ah, yes. True, we should.

sshanzel:
Woop! Woop!

Screen Shot 2022-03-17 at 5 13 02 PM

packages/shared/src/components/PostsSearch.tsx
```diff
@@ -122,7 +122,7 @@ export default function PostsSearch({
       <SearchField
         className="flex-1 compact"
         inputId="posts-search"
-        compact
+        fieldSize="medium"
```
rebelchris, line 125: I see now that it's correct for the "feed", but could you check with Tsahi that he intended the feeds to have the "medium" search

packages/shared/src/components/fields/SearchField.tsx
```diff
@@ -73,17 +87,36 @@ export const SearchField = forwardRef(function SearchField(
     setInput(null);
   };
 
+  const isPrimary = fieldType === 'primary';
+  const isSecondary = fieldType === 'secondary';
+  const SearchIcon = focused ? MagnifyingFilledIcon : MagnifyingOutlineIcon;
```
rebelchris, line 92: You don't have to do this now, but could you note that we add a ticket to make a hook for this?

rebelchris (approved):
Awesome work @sshanzel 
All good to go, just have some remarks we should just validate/write down.

Perhaps @idoshamun might want to take a look at this one, (Only if he wants, else feel free to merge)

packages/shared/src/components/PostsSearch.tsx
sshanzel: I think the size is really necessary as it came from the guideline.

Screen Shot 2022-03-22 at 4 51 58 PM

packages/shared/src/components/fields/SearchField.tsx
sshanzel: That's a great idea.

packages/shared/src/components/PostsSearch.tsx
rebelchris: Yes it is needed

My question is to ask Tsahi if "medium" is the right size for the feeds 🙌

sshanzel: Oh, sorry for the misunderstanding. Let me ping him up. Thank you!

idoshamun:
@rebelchris I'm not highly familiar with the specs of this change, so can't really speak my mind here

rebelchris:
@idoshamun No problem, @sshanzel Go for it.
The changes look great from my side 🙌

PR #923, opened 2022-03-17
refactor: let variable for avoiding refresh

sshanzel:
## Changes

Describe what this PR does
There used to be an issue when the feed filters are opened and the user has blocked a tag. The feed unintentionally gets invalidated. The solution was to create a let variable to handle the "state".

The reason why a local `useState` wasn't working when applied is we instantiate multiple `useFeedSettings` which results in instantiating their own states that should have been a centralized variable.

The fix I applied was to utilize our `usePersistentContext` as it is saved in the cache where it will be constant across different instances.

Additional refactor: We instantiate the `useFeedSettings` in multiple areas, all those instances make their own unnecessary GraphQL request for the actual data, which in fact when we load the `FeedFilters` component, we will already have the information needed by its child components that also instantiated the said hook. Hence, I created an additional parameter that will determine whether to make a request or not but still fetches the cache data.

## Manual Testing

- On those affected packages:

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [x] Laptop (1020px)

DD-{number} #done


packages/shared/src/hooks/useFeedSettings.ts
```diff
@@ -98,6 +105,7 @@ export default function useFeedSettings(): FeedSettingsReturnType {
 
       return { ...req, feedSettings };
     },
+    { enabled: isTesting || !observerOnly },
```
rebelchris, line 108: Do we have cases where we set the value but not watch in the same hook?
Just not really sure what the use-case for not having it is?

rebelchris (approved):
Love this refactor, just not sure on that observer?
Would you mind explaining that a little bit more?

packages/shared/src/hooks/useFeedSettings.ts
sshanzel: This is when we open the feed filters popup it should make a graphql request - now when we open the `Tags` section, it also uses the same hook, which since it is a different instance, it would again make the same graphql request, which it should not because the data was already available.

What do you think about this? I am also having some doubts if the reduction of calls would be worth it since we are not yet experiencing any issues with the number of requests.

rebelchris: My main concern was if it would cause any "negative" effects.
I wouldn't see any, and like you said we don't have any issues with the performance yet.

But maybe @idoshamun has an idea on this?

sshanzel: Probably this should be a subject on a different PR if in any case. I can revert it for now so we can proceed with the primary concern.

PR #924, opened 2022-03-21
refactor: preview comment min height

sshanzel:
## Changes

- Describe what this PR does
@idoshamun after checking through the scenarios, it seems that there would really be certain cases where the height won't be equal, so calculating it on the go would still produce a flicker due to the height differences (ex. when displaying an image, content would automatically be larger.).

With that, I have decided to make the min-height to be the same as the actual min-height of the comment, it won't be noticeable with our minimum number of rows. What do you think?

## Manual Testing

- On those affected packages:

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

DD-{number} #done


idoshamun:
This works for most cases I think :)

idoshamun (approved):

PR #925, opened 2022-03-21
feat: add device id to analytics events

idoshamun:
## Changes

The device ID is used to identify the specific device that is sending the events to our server.
It is used to filter out bots, cookie blockers, and other users who might generate noise in our tracking system.
It is generated by the client and saved locally in IndexedDB so it doesn't rely on cookies.

## Manual Testing

- On those affected packages:

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)


packages/shared/src/hooks/analytics/useAnalyticsSharedProps.ts
```diff
@@ -6,11 +6,25 @@ import {
...
+import { v4 as uuidv4 } from 'uuid';
 import { AnalyticsEvent } from './useAnalyticsQueue';
 import FeaturesContext from '../../contexts/FeaturesContext';
 import SettingsContext from '../../contexts/SettingsContext';
 import AuthContext from '../../contexts/AuthContext';
 
+const DEVICE_ID_KEY = 'device_id';
+
+const getOrGenerateDeviceId = async (): Promise<string> => {
+  const deviceId = await getCache<string | undefined>(DEVICE_ID_KEY);
```
rebelchris, line 19: Could you not leverage the usePersistentContext hook?
that technically uses the idb cache

rebelchris (changes requested):
Love it, just wondering if we can leverage our existing hook for this.

packages/shared/src/hooks/analytics/useAnalyticsSharedProps.ts
idoshamun: The reason I didn't use our hook is that it doesn't support lazy evaluating the default value and I didn't want to introduce it just for this case on one hand, and on the other hand, I didn't want to generate a new UUID when it's not necessary because the device id is already presented. What do you think?

PR #928, opened 2022-03-22
feat: ab testing feed filtering

ayadotdev:
## Changes

This PR adds AB testing modal options for creating feed filtering. 
There is a default Sidebar option and 4 new modal options:

v1-default
v2
v3
v4
v5

## Todo list

- [x] add new modal types features
- [x] add new modal
- [x] adjust a close button position on v3 (right screen angle)
- [x] add logic for UI elements order (such as search bar, heading, paragraph, Create my feed button)
- [x] add new unit tests

## Manual Testing

- On those affected packages:

- [x] MobileL (420px)
- [x] Tablet (656px)
- [x] Laptop (1020px)

DD-490 #done


packages/shared/src/components/modals/CreateMyFeedModal.tsx
```diff
@@ -0,0 +1,120 @@
...
+      {type === 'v5' && (
+        <>
+          <Button className="mr-3 w-40 btn-secondary" onClick={onRequestClose}>
+            Cancel
+          </Button>
+          <CreateFeedFilterButton
+            className={classNames(buttonClass[type], 'btn-primary-cabbage')}
+          />
+        </>
+      )}
```
rebelchris, line 56: Can we not combine the <CreateFeedFilterButton> they seem twice the same, so we can just render the close button dynamically.

packages/shared/src/components/modals/CreateMyFeedModal.tsx
```diff
@@ -0,0 +1,120 @@
...
+  onRequestClose,
+}: CloseButtonProps) => {
+  return (
+    <Button
+      className={classNames('btn-tertiary', closeButtonClassName)}
+      buttonSize={size || 'small'}
+      title="Close"
+      icon={<XIcon />}
+      onClick={onRequestClose}
+      position={position || ''}
```
rebelchris, line 74: Why do we need the || '' here?

packages/shared/src/components/modals/CreateMyFeedModal.tsx
```diff
@@ -0,0 +1,120 @@
...
+import XIcon from '../../../icons/x.svg';
+import { Button, ButtonSize } from '../buttons/Button';
+
+export type TypeProps = {
+  type: string;
+};
+
+export type CreateMyFeedModalProps = TypeProps & ModalProps;
+
+export type LayoutModalProps = TypeProps & Partial<ModalProps>;
```
rebelchris, line 18: I see you export all these, are they used anywhere else?

rebelchris (changes requested):
Early comments as a first draft.
Like the setup, still would love to see more happen with CSS personally though.

Let me know if you want to look at that together?

packages/shared/src/components/modals/CreateMyFeedModal.tsx
ayadotdev: I am not sure now, I removed it.

packages/shared/src/components/modals/CreateMyFeedModal.tsx
ayadotdev: They are not 👍

packages/shared/src/components/modals/CreateMyFeedModal.tsx
ayadotdev: ✅

ayadotdev: @rebelchris yes, I would like to look at that together 🙏

packages/shared/src/components/modals/CreateMyFeedModal.tsx
```diff
@@ -0,0 +1,121 @@
...
+  return (
+    <footer className="flex fixed responsiveModalBreakpoint:sticky bottom-0 justify-center items-center py-3 border-t border-theme-divider-tertiary bg-theme-bg-tertiary">
+      {type === 'v4' && getCreateButton('v4')}
+      {type === 'v5' && (
+        <>
+          <Button className="mr-3 w-40 btn-secondary" onClick={onRequestClose}>
+            Cancel
+          </Button>
+          {getCreateButton('v5')}
+        </>
```
rebelchris, line 57: I see we have the `getCreateButton` twice here, I think we can have this only once like we did in the other space.

packages/shared/src/components/filters/TagsFilter.tsx
```diff
@@ -68,19 +87,42 @@ export default function TagsFilter({
...
+      <div
+        className={classNames(
+          containerClass[feedFilterModalType] ?? 'flex-col',
+          'flex px-6 pb-6',
+        )}
+      >
+        <SearchField
+          inputId="search-filters"
+          placeholder="Search"
+          className={classNames(searchFieldClass[feedFilterModalType], 'mb-6')}
```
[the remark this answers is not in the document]
rebelchris: Hmm, let's go with this if it works solid 🙌

packages/shared/src/components/modals/CreateMyFeedModal.tsx
```diff
@@ -0,0 +1,119 @@
...
+};
+
+export default function CreateMyFeedModal({
+  className,
+  onRequestClose,
+  type,
+  ...modalProps
+}: CreateMyFeedModalProps): ReactElement {
+  return (
+    <>
```
sshanzel, line 83: I think we can safely remove this extra `Fragment`.

sshanzel (changes requested):
Well done! You are on the right track! 🚢 

Just had a few comments, great work!

packages/shared/src/components/filters/TagsFilter.tsx
```diff
@@ -68,19 +87,42 @@ export default function TagsFilter({
...
-        ref={searchRef}
-        valueChanged={setQuery}
-      />
-      {query?.length > 0 ? (
+      <div
+        className={classNames(
+          containerClass[feedFilterModalType] ?? 'flex-col',
+          'flex px-6 pb-6',
+        )}
+      >
```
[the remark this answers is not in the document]
ayadotdev: I think it is as inside of the parent element we have 2 blocks and we apply a col-reverse only for one of them (with title, search bar and paragraph), so applying classes to the only parent element will break styles for some modal versions.

packages/shared/src/components/modals/CreateMyFeedModal.tsx
```diff
@@ -0,0 +1,119 @@
...
+import XIcon from '../../../icons/x.svg';
+import { Button, ButtonSize } from '../buttons/Button';
+
+type TypeProps = {
+  type: string;
+};
+
+type CreateMyFeedModalProps = TypeProps & ModalProps;
+
+type LayoutModalProps = TypeProps & Partial<ModalProps>;
```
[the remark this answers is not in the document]
ayadotdev: Nice, I didn't know the difference.

packages/shared/src/components/modals/CreateMyFeedModal.tsx
```diff
@@ -0,0 +1,119 @@
...
+  v3: 'Choose tags to follow',
+  v4: 'Feed filters',
+  v5: 'Choose tags to follow',
+};
+
+const headerClass = {
+  v4: 'flex-row-reverse',
+};
+
+const getFooter = ({ type, onRequestClose }: LayoutModalProps) => {
```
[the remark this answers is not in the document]
ayadotdev: The difference should be only he name of the function?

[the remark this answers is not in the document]
sshanzel: Would also be how we render it. Reading the render would allow us to easily understand we expect a component - a function can return anything, so we won't have to look at it if not really necessary.

packages/shared/src/components/modals/CreateMyFeedModal.tsx
```diff
@@ -0,0 +1,105 @@
...
+          <CreateFeedFilterButton
+            className={classNames(buttonClass[type], 'btn-primary-cabbage')}
+            icon={type === 'v2' && <PlusIcon />}
+          />
+        )}
+      </header>
+      <section className="mt-6">
+        <TagsFilter />
+      </section>
+      {['v4', 'v5'].includes(type) && ModalFooter({ type, onRequestClose })}
```
sshanzel, line 102:
```suggestion
      {['v4', 'v5'].includes(type) && <ModalFooter type={type} onRequestClose={onRequestClose} />}
```

packages/shared/src/components/modals/CreateMyFeedModal.tsx
```diff
@@ -0,0 +1,105 @@
...
+      )}
+      <header
+        className={classNames(
+          headerClass[type],
+          'flex fixed responsiveModalBreakpoint:sticky top-0 left-0 z-3 justify-between items-center py-4 px-6 w-full border-b border-theme-divider-tertiary bg-theme-bg-tertiary',
+        )}
+      >
+        {['v2', 'v4'].includes(type) && (
+          <Button
+            className={classNames('btn-tertiary')}
```
sshanzel, line 84:
```suggestion
            className="btn-tertiary"
```

packages/shared/src/components/modals/CreateMyFeedModal.tsx
```diff
@@ -0,0 +1,105 @@
...
+    <ResponsiveModal
+      className={classNames(className, styles.accountDetailsModal)}
+      {...modalProps}
+      onRequestClose={onRequestClose}
+    >
+      {type === 'v3' && (
+        <Button
+          className={classNames(
+            'btn-tertiary-float bg-theme-bg-tertiary fixed top-8 right-8',
+          )}
```
sshanzel, line 68:
```suggestion
          className="btn-tertiary-float bg-theme-bg-tertiary fixed top-8 right-8"
```

packages/shared/src/components/filters/TagsFilter.tsx
```diff
@@ -23,6 +24,23 @@ import { FilterMenuProps } from './common';
...
+import FeaturesContext from '../../contexts/FeaturesContext';
+
+const containerClass = {
+  v3: 'flex-col-reverse',
+  v5: 'flex-col-reverse',
+};
+
+const searchFieldClass = {
+  v3: 'mt-6 mb-2',
+  v5: 'mb-2',
```
sshanzel, line 37: As you mentioned, you wanted to override the bottom margin of `v5`. Why not put it here instead?

Reference: https://github.com/dailydotdev/apps/pull/928/files#r833365205
```suggestion
  v5: 'mb-6',
```

packages/shared/src/components/filters/TagsFilter.tsx
sshanzel, line 106: The `v3` value would be overriden, you would have to make it an optional value, else you will have to set something for every case.
```suggestion
          className={classNames(searchFieldClass[feedFilterModalType] || 'mb-6')}
```

sshanzel (changes requested):
Just a few more changes, but you are almost there!

packages/shared/src/components/filters/TagsFilter.tsx
ayadotdev: It seems to me that margin 2 is the correct one because there is also a paragraph with its padding of 6. For version 5 margin should be 8.

ayadotdev:
@rebelchris @sshanzel please check and in case approve 💟

packages/shared/src/components/filters/TagsFilter.tsx
sshanzel: Alright, thought you really wanted the 6 since you mentioned it in your reply to Chris. This should be okay now as you have verified.

sshanzel (changes requested):
Great refactoring skills right there, Aya! We're almost there! 🙏

sshanzel (approved):
LGTM!!! Well done! 🚢 🚢 🚢

packages/shared/src/components/CreateFeedFilterButton.tsx
```diff
@@ -0,0 +1,35 @@
...
+  const { user, showLogin } = useContext(AuthContext);
+  const { trackEvent } = useContext(AnalyticsContext);
+  const client = useQueryClient();
+
+  const onCreate = () => {
+    trackEvent({
+      event_name: 'click',
+      target_type: 'create feed filters',
+      target_id: 'feed-filters',
+    });
```
rebelchris, line 24: We might want to check with @idoshamun  if we want to get the specific version for which this button was clicked.

rebelchris (approved):
Wow, well done as so much improved and abstracted 🙌

packages/shared/src/components/CreateFeedFilterButton.tsx
idoshamun: Yes, you are right. We want to set a different `target_id` to every option so we can select the best performing option based on conversion rate

ayadotdev: I made the `target_id` dynamic based on the version, so we have a different one for each type. Could you please have a look and approve if it makes sense for you? @idoshamun

idoshamun (approved):
Tracking looks great 💪

PR #930, opened 2022-03-22
feat: companion extension

rebelchris:
## Changes

- Added the companion application
- Added basic companion render (chrome) test cases as the individual actions are all existing implementations.

## Manual Testing

- On those affected packages:

- [x] MobileL (420px) (Not super applicable, but it works)
- [x] Tablet (656px)
- [x] Laptop (1020px)

DD-458 #done

### DIFF
--- a/packages/extension/src/background/index.ts
+++ b/packages/extension/src/background/index.ts
@@ -10,7 +10,7 @@ import {
   getLocalBootData,
 } from '@dailydotdev/shared/src/contexts/BootProvider';
 import { getOrGenerateDeviceId } from '@dailydotdev/shared/src/hooks/analytics/useDeviceId';
-import { registerBrowserContentScripts } from '../companion/useExtensionPermission';
+import { getContentScriptPermissionAndRegister } from '../companion/useExtensionPermission';
 import {
   getExtensionAlerts,
   updateExtensionAlerts,
@@ -64,12 +64,7 @@ const sendBootData = async (req, sender) => {
 };
 
 async function handleMessages(message, sender) {
-  const permissions = await browser.permissions.contains({
-    origins: ['*://*/*'],
-  });
-  if (permissions) {
-    await registerBrowserContentScripts();
-  }
+  await getContentScriptPermissionAndRegister();
 
   if (message.type === 'CONTENT_LOADED') {
     sendBootData(message, sender);
@@ -140,4 +135,8 @@ browser.runtime.onInstalled.addListener(async (details) => {
   await Promise.all([
     browser.runtime.setUninstallURL('https://daily.dev/uninstall'),
   ]);
+});
+
+browser.runtime.onStartup.addListener(async () => {
+  await getContentScriptPermissionAndRegister();
 });
--- a/packages/extension/src/companion/CompanionPermission.tsx
+++ b/packages/extension/src/companion/CompanionPermission.tsx
@@ -17,7 +17,7 @@ const CompanionPermissionComponent = (
   _,
   ref: Ref<HTMLDivElement>,
 ): ReactElement => {
-  const { registerContentScripts } = useExtensionPermission({
+  const { requestContentScripts } = useExtensionPermission({
     origin: 'companion permission popup',
   });
   const { flags } = useContext(FeaturesContext);
@@ -43,7 +43,7 @@ const CompanionPermissionComponent = (
         </p>
         <Button
           className="mt-1 w-[12.5rem] btn btn-primary"
-          onClick={registerContentScripts}
+          onClick={requestContentScripts}
           buttonSize="small"
         >
           {button}
--- a/packages/extension/src/companion/useCompanionSettings.ts
+++ b/packages/extension/src/companion/useCompanionSettings.ts
@@ -20,7 +20,7 @@ export const useCompanionSettings = (origin: string): UseCompanionSettings => {
   );
   const { optOutCompanion, toggleOptOutCompanion, loadedSettings } =
     useContext(SettingsContext);
-  const { contentScriptGranted, registerContentScripts } =
+  const { contentScriptGranted, requestContentScripts } =
     useExtensionPermission({ origin });
 
   useEffect(() => {
@@ -38,7 +38,7 @@ export const useCompanionSettings = (origin: string): UseCompanionSettings => {
       return;
     }
 
-    registerContentScripts().then((granted) => {
+    requestContentScripts().then((granted) => {
       if (!granted) {
         toggleOptOutCompanion();
       }
--- a/packages/extension/src/companion/useExtensionPermission.ts
+++ b/packages/extension/src/companion/useExtensionPermission.ts
@@ -8,7 +8,7 @@ import AnalyticsContext from '@dailydotdev/shared/src/contexts/AnalyticsContext'
 interface UseExtensionPermission {
   isFetched?: boolean;
   contentScriptGranted: boolean;
-  registerContentScripts: () => Promise<boolean>;
+  requestContentScripts: () => Promise<boolean>;
 }
 
 export const registerBrowserContentScripts =
@@ -22,13 +22,19 @@ export const registerBrowserContentScripts =
       ],
     });
 
-const getContentScriptPermission = async () => {
-  const permissions = await browser.permissions.contains({
+export const getContentScriptPermission = (): Promise<boolean> =>
+  browser.permissions.contains({
     origins: ['*://*/*'],
   });
 
-  return !!permissions;
-};
+export const getContentScriptPermissionAndRegister =
+  async (): Promise<void> => {
+    const permission = await getContentScriptPermission();
+
+    if (permission) {
+      await registerBrowserContentScripts();
+    }
+  };
 
 const contentScriptKey = 'permission_key';
 
@@ -52,7 +58,7 @@ export const useExtensionPermission = ({
     },
   );
 
-  const registerContentScripts = async () => {
+  const requestContentScripts = async () => {
     trackEvent({
       event_name: 'request content scripts',
       extra: JSON.stringify({ origin }),
@@ -75,7 +81,7 @@ export const useExtensionPermission = ({
   };
 
   return useMemo(
-    () => ({ contentScriptGranted, registerContentScripts, isFetched }),
+    () => ({ contentScriptGranted, requestContentScripts, isFetched }),
     [contentScriptGranted, isFetched],
   );
 };
--- a/packages/extension/src/newtab/App.tsx
+++ b/packages/extension/src/newtab/App.tsx
@@ -25,6 +25,10 @@ import { version } from '../../package.json';
 import MainFeedPage from './MainFeedPage';
 import { DndContextProvider } from './DndContext';
 import { BootDataProvider } from '../../../shared/src/contexts/BootProvider';
+import {
+  registerBrowserContentScripts,
+  useExtensionPermission,
+} from '../companion/useExtensionPermission';
 
 const AnalyticsConsentModal = dynamic(() => import('./AnalyticsConsentModal'));
 
@@ -57,6 +61,9 @@ function InternalApp({
     shouldShowLogin,
     loginState,
   } = useContext(AuthContext);
+  const { contentScriptGranted } = useExtensionPermission({
+    origin: 'on extension load',
+  });
   const [analyticsConsent, setAnalyticsConsent] = usePersistentState(
     'consent',
     false,
@@ -88,6 +95,12 @@ function InternalApp({
     routeChangedCallbackRef.current();
     dismissToast();
   };
+
+  useEffect(() => {
+    if (contentScriptGranted) {
+      registerBrowserContentScripts();
+    }
+  }, [contentScriptGranted]);
 
   return (
     <DndContextProvider>


packages/extension/package.json
```diff
@@ -36,6 +36,7 @@
     "react-intersection-observer": "^8.32.2",
     "react-modal": "^3.14.3",
     "react-query": "^3.31.0",
+    "react-shadow-root": "^6.1.1",
```
rebelchris, line 39: Not sure if we actually need this anymore

packages/extension/package.json
```diff
@@ -110,9 +111,11 @@
     "postcss-loader": "^4.0.4",
     "postcss-mixins": "^8.1.0",
     "postcss-nesting": "^7.0.1",
+    "postcss-rem-to-responsive-pixel": "^5.1.1",
     "postcss-simple-vars": "^6.0.3",
     "prettier": "^2.4.1",
     "resolve-url-loader": "^3.1.4",
+    "@open-wc/testing": "^2.1.2",
```
rebelchris, line 118: Double check this one as well

packages/extension/src/background/index.ts
```diff
@@ -15,6 +27,95 @@ const cacheAmplitudeDeviceId = async ({
   }
 };
 
+const excludedCompanionOrigins = [
+  'https://twitter.com',
+  'https://www.google.com',
+  'chrome-extension://',
+];
```
rebelchris, line 34: This is kind of a first draft of hard-coded domains we would never want to fire on.
We can extend this or make this work differently in the future.

The idea was to have sites that devs use a lot, but never are featured blacklisted from Querying the API

packages/extension/src/companion/App.tsx
```diff
@@ -0,0 +1,251 @@
...
+    id: 'companion-options-context',
+  });
+  const onContextOptions = (event: React.MouseEvent): void => {
+    showCompanionOptionsMenu(event, {
+      position: { x: 48, y: 265 },
+    });
+  };
+
+  return (
+    <div
```
rebelchris, line 100: Perhaps we can refactor this into multiple smaller components

packages/shared/src/styles/base.css
```diff
@@ -1,8 +1,15 @@
-html {
+html, #daily-companion-wrapper {
```
rebelchris, line 1: Check if we still use this one please

packages/shared/src/styles/base.css
```diff
@@ -21,6 +28,10 @@ body {
   &.ReactModal__Body--open .ReactModal__Content > * {
     overscroll-behavior-y: none;
   }
+   &.companion {
+  min-width: auto;
+  overflow: hidden;
+    }
```
rebelchris, line 34: Thinks this can be removed as well

rebelchris (commented):
Some small notes

packages/extension/src/background/index.ts
idoshamun: We can add stackoverflow, gmail, google meet, google calendar. Don't know if it makes sense

packages/extension/public/css/daily-companion-app.css
```diff
@@ -0,0 +1,8 @@
+daily-companion-app {
+  line-height: 1.5; /* 1 */
```
sshanzel, line 2: Just a question, what does the number comments signify?

packages/extension/src/background/index.ts
```diff
@@ -15,6 +27,95 @@ const cacheAmplitudeDeviceId = async ({
...
+
+const isExcluded = (origin: string) =>
+  excludedCompanionOrigins.some((e) => origin.includes(e));
+
+const sendBootData = async (req, sender) => {
+  if (isExcluded(sender?.origin)) {
+    return;
+  }
+  const cacheData = JSON.parse(localStorage.getItem('boot:local'));
```
sshanzel, line 43: I think we have a generic function that fetches the local boot data.

packages/extension/src/companion/CompanionContextMenu.tsx
```diff
@@ -0,0 +1,119 @@
...
+  onReport,
+  onBlockSource,
+  onDisableCompanion,
+}: {
+  postData: BootData;
+  onMessage?: (message: string, timeout?: number) => Promise<unknown>;
+  onReport;
+  onBlockSource;
+  onDisableCompanion;
+}): ReactElement {
```
sshanzel, line 24: I think we can create an interface for this so we can also explicitly define the parameters.

packages/shared/src/lib/boot.ts
```diff
@@ -13,17 +13,24 @@ export type Boot = {
...
   flags: IFlags;
   settings: RemoteSettings;
+  data?;
 };
 
 export type BootCacheData = Pick<
   Boot,
-  'user' | 'alerts' | 'settings' | 'flags'
->;
+  'user' | 'alerts' | 'settings' | 'flags' | 'data'
```
sshanzel, line 21: What is the content of `data` property? I think we should change it to something much more explicit.

packages/shared/src/components/modals/ReportPostModal.tsx
```diff
@@ -14,7 +14,7 @@ import {
 
 export interface Props extends ModalProps {
   postIndex: number;
-  post: Post;
+  post?: Partial<Post>;
```
sshanzel, line 17: Why do we have to make this as a `Partial`type? I think the post should be required to report something.

packages/extension/src/companion/common.tsx
```diff
@@ -0,0 +1,20 @@
+import { Post } from '@dailydotdev/shared/src/graphql/posts';
+
+export type BootData = Pick<
```
sshanzel, line 3:
```suggestion
export type PostBootData = Pick<
```

packages/extension/src/companion/useCompanionActions.ts
```diff
@@ -0,0 +1,143 @@
...
+  removeUpvote: (variables: T) => Promise<void>;
+  disableCompanion: (variables: T) => Promise<void>;
+};
+
+export default function useCompanionActions<
+  T extends { id?: string; reason?: string; comment?: string } = {
+    id?: string;
+    reason?: string;
+    comment?: string;
+  },
```
sshanzel, line 26: I think we should create interfaces for this, to make it easier to understand.

sshanzel (changes requested):
Wow, I can't imagine the amount of work you did put into this. Must have been crazy moments 💪 💪 💪 

Just had a few comments and questions but nothing major. Super amazing work you got here, Chris! 🎸 🎸 🎸

packages/shared/src/components/modals/ReportPostModal.tsx
rebelchris: It should indeed, I must actually remake it to be type of CompanionPost/Post
As the CompanionPost has less data compared to a full post.

packages/extension/public/css/daily-companion-app.css
rebelchris: To be completely honest, no idea.
It was something we had setup in the extension, and was thinking it might have been a fallback for certain old browsers?

packages/extension/src/background/index.ts
idoshamun: Following up on our call yesterday, I see that `excludedCompanionOrigins` is your own implementation and not something we define on the manifest level. So it's great to have this list and when needed we can just remove sites from this list if we suddenly want to support them

packages/extension/src/background/index.ts
```diff
@@ -1,5 +1,17 @@
...
+  REMOVE_BOOKMARK_MUTATION,
+  REPORT_POST_MUTATION,
+  UPVOTE_MUTATION,
+} from '@dailydotdev/shared/src/graphql/posts';
+import request from 'graphql-request';
+import { ADD_FILTERS_TO_FEED_MUTATION } from '@dailydotdev/shared/src/graphql/feedSettings';
+import { UPDATE_USER_SETTINGS_MUTATION } from '@dailydotdev/shared/src/graphql/settings';
+import { BOOT_LOCAL_KEY } from '@dailydotdev/shared/src/contexts/BootProvider';
 
 const cacheAmplitudeDeviceId = async ({
```
idoshamun, line 16: This is no longer needed I think.
It should have been removed during the third party tracking task but we probably forgot

packages/extension/src/background/index.ts
idoshamun: Not sure we need the cached data, to be honest. Because the companion boot is a requirement before we can render anything. So when we need to render something we already have the most up-to-date boot data available.

packages/extension/src/background/index.ts
```diff
@@ -15,6 +27,95 @@ const cacheAmplitudeDeviceId = async ({
...
+  });
+};
+
+function handleMessages(message, sender) {
+  if (message.type === 'CONTENT_LOADED') {
+    sendBootData(message, sender);
+    return null;
+  }
+
+  if (message.type === 'UPVOTE_POST') {
```
idoshamun, line 61: A reminder that we want to change the API with the background worker to be something like a proxy to the `request` function rather than implementing a new type for every graphql request.

packages/extension/src/companion/index.tsx
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+import ReactDOM from 'react-dom';
+import { browser } from 'webextension-polyfill-ts';
+import { themeModes } from '@dailydotdev/shared/src/contexts/SettingsContext';
+import { CompanionBootData } from './common';
+import App from './App';
+
+const init = async (data: CompanionBootData, settings) => {
```
idoshamun, line 8: I suggest a more explicit name like `render` or `renderApp`

packages/extension/src/background/index.ts
```diff
@@ -15,6 +27,95 @@ const cacheAmplitudeDeviceId = async ({
...
+  }
+  await browser.tabs.sendMessage(sender?.tab?.id, {
+    data,
+    settings: settingsOutput,
+  });
+};
+
+function handleMessages(message, sender) {
+  if (message.type === 'CONTENT_LOADED') {
+    sendBootData(message, sender);
```
idoshamun, line 57: If the user has opted out we should not send a request to our server at all, I think. What do you think?

packages/extension/src/companion/App.tsx
```diff
@@ -0,0 +1,251 @@
...
+  if (isOptOutCompanion) {
+    return <></>;
+  }
+
+  return (
+    <div>
+      <style>
+        @import &quot;chrome-extension://{browser.runtime.id}
+        /css/companion.css&quot;;
+      </style>
```
idoshamun, line 242: Is there no other way to achieve it? How do we do it in the extension?

packages/extension/src/companion/App.tsx
idoshamun: I agree this is too big and hard to read

packages/extension/src/companion/CompanionContextMenu.tsx
```diff
@@ -0,0 +1,119 @@
...
+          url: shareLink,
+        });
+      } catch (err) {
+        // Do nothing
+      }
+    } else {
+      await navigator.clipboard.writeText(shareLink);
+      onMessage('✅ Copied link to clipboard');
+    }
+  };
```
idoshamun, line 43: Don't we have a function like this already?

packages/extension/src/companion/CompanionContextMenu.tsx
```diff
@@ -0,0 +1,119 @@
...
+          <EyeIcon className="mr-2 text-xl" /> Disable widget
+        </Item>
+      </Menu>
+      {reportModal && (
+        <RepostPostModal
+          post={postData}
+          parentSelector={() =>
+            document
+              .querySelector('daily-companion-app')
+              .shadowRoot.querySelector('#daily-companion-wrapper')
```
idoshamun, line 97: I suggest turning it into a function

packages/shared/src/components/modals/LoginModal.tsx
```diff
@@ -34,7 +33,6 @@ export default function LoginModal({
       className={classNames(styles.loginModal, className)}
     >
       <ModalCloseButton onClick={onRequestClose} />
-      <DailyDevLogo />
```
idoshamun, line 37: Why did you remove the logo?

packages/shared/src/contexts/BootProvider.tsx
```diff
@@ -70,12 +70,13 @@ const updateLocalBootData = (
   current: Partial<BootCacheData>,
   boot: Partial<BootCacheData>,
 ) => {
-  const localData = { ...current, ...boot };
+  const localData = { ...current, ...boot, last_modifier: 'extension' };
```
idoshamun, line 73: Why not camelCase?

packages/shared/src/contexts/BootProvider.tsx
```diff
@@ -112,20 +113,20 @@ export const BootDataProvider = ({
...
     } else {
-      if (lastAppliedChange) {
+      if (cachedData?.last_modifier !== 'companion' && lastAppliedChange) {
         updatedData = { ...updatedData, ...lastAppliedChange };
       }
       setLastAppliedChange(null);
     }
-
-    const updated = updateLocalBootData(cachedBootData, updatedData);
+    const updated = updateLocalBootData(cachedData, updatedData);
```
idoshamun, line 129: Not sure I understand the reason for this change

packages/shared/src/styles/base.css
```diff
@@ -1,8 +1,15 @@
...
   color: var(--theme-label-primary);
   height: -webkit-fill-available;
 }
+daily-companion-app {
+  line-height: 1.5; /* 1 */
+  -webkit-text-size-adjust: 100%; /* 2 */
+  -moz-tab-size: 4; /* 3 */
+  tab-size: 4; /* 3 */
+  font-family: theme('fontFamily.sans', ui-sans-serif, system-ui, -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, "Helvetica Neue", Arial, "Noto Sans", sans-serif, "Apple Color Emoji", "Segoe UI Emoji", "Segoe UI Symbol", "Noto Color Emoji"); /* 4 */
+}
```
idoshamun, line 12: If you copied it from somewhere else maybe we can utilize the same selector?

packages/extension/public/css/daily-companion-app.css
sshanzel: Oh, cool. Thought this was something new. Thank you!

idoshamun (changes requested):
I did a high level review looked for some big chunks and put my comments.
You overcame so many challenges in this single PR that it is definitely amazing 🤩 
Can't wait to see the impact of it

packages/extension/src/background/index.ts
rebelchris: This part is only for anonymous user syncing if they modify the theme in the extension and move over to a companion enabled website.

packages/extension/src/background/index.ts
rebelchris: The main issue is that this info only comes from the boot query.
We could do it based on the local cache, but it might not always be the latest value.

packages/extension/src/companion/App.tsx
rebelchris: The main issue is actually shadow dom not being able to access regular imports in React.
I can have another look at it though.

packages/shared/src/contexts/BootProvider.tsx
rebelchris: Thanks for raising this.
We introduced the lastAppliedChange check which could favorite local data over remote data.
however with the introduction of the companion there is a "third" race condition.

When the companion runs a update we always want to favorite that, (it's the only way this last_modifier get's set to companion)

Else we got into edge cases where you turn of the companion (from the companion)
And the extension would re-enable it again from cache or remote data.

Let me know if that makes sense, happy to showcase it if that helps

packages/extension/src/background/index.ts
idoshamun: I see ok then...

packages/extension/src/background/index.ts
idoshamun: I'm ok with this tradeoff. If the user opted out we should reduce the overhead of this extension to minimum

packages/extension/src/companion/App.tsx
idoshamun: Actually, I'm thinking of a webpack trick perhaps. The best way would be to inline the CSS to there during the build. Not sure if it's easy to achieve though.

packages/shared/src/contexts/BootProvider.tsx
idoshamun: We need to document it somehow. It will be very hard to maintain this piece without understanding the side effects

packages/extension/src/background/index.ts
rebelchris: Refactoring in separate PR

packages/shared/src/contexts/BootProvider.tsx
rebelchris: Let me create a DR for this and describe the flow we choose

packages/extension/src/companion/App.tsx
rebelchris: I tried some of the known webpack tips:
- import directly -> doesn't build in the shadow dom
- import as variable and inject output -> outputs empty

What would be the benefit of doing that vs the import like this?
As that companion.css gets dynamically build for every build. (It's a packed version of the global.css)

idoshamun: To streamline the way we use global css files but don't put any more time into it. It's not that important

packages/extension/src/companion/companionRequest.tsx
```diff
@@ -0,0 +1,16 @@
...
+      type: 'GRAPHQL_REQUEST',
+      url: args?.[0],
+      document: args?.[1],
+      variables: args?.[2],
+    });
+    return null;
+  },
+};
+
+export const companionRequest = new Proxy(request, proxyRequest);
```
idoshamun, line 16: What's the advantage of using `Proxy`?
Because it seems like you must import the `graphql-request` module here although you don't really use it

idoshamun (changes requested):
Looks great only one question

packages/extension/src/companion/companionRequest.tsx
rebelchris: I think the main advantage is that is copies the signature so we don't have to explicitly state those, and we can simply "modify" what it actually does.

Would you rather have a function that just has the same props?

idoshamun: I would favor a solution that doesn't involve importing unnecessary modules in this case

sshanzel (approved):
Sharp work! Nothing to add from my end. Can't believe we are nearing the merge of this feature. Great work, no doubt! 🚢

packages/extension/src/companion/App.tsx
```diff
@@ -0,0 +1,77 @@
...
+              visit={visit}
+              tokenRefreshed
+              getRedirectUri={() => '/'}
+              updateUser={() => null}
+            >
+              <SettingsContextProvider settings={settings}>
+                <AlertContextProvider alerts={alerts}>
+                  <AnalyticsContextProvider
+                    app="companion"
+                    version="1"
```
idoshamun, line 59: Can you provide the real extension version here, please?

packages/extension/src/companion/App.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+  return (
+    <div>
+      <style>
+        @import &quot;chrome-extension://{browser.runtime.id}
+        /css/companion.css&quot;;
+      </style>
+      <RouterContext.Provider value={router}>
+        <QueryClientProvider client={queryClient}>
+          <FeaturesContext.Provider value={{ flags }}>
```
idoshamun, line 47: Usually, it's better to have `{{ flags }}` memoized. Otherwise it creates a new object on every render

packages/extension/src/companion/index.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+const renderApp = ({ ...props }: CompanionData) => {
+  const { settings } = props;
+  getCompanionWrapper().classList.add(themeModes[settings.theme]);
+
+  // Set target of the React app to shadow dom
+  ReactDOM.render(<App {...props} />, getCompanionWrapper());
+};
+
+browser.runtime.onMessage.addListener(async (request) => {
```
idoshamun, line 16: Why do we need the `async` for?
Can't we do the destructuring in the function declaration instead of the line below?

packages/shared/src/hooks/analytics/useAnalyticsQueue.ts
```diff
@@ -16,7 +16,10 @@ export type PushToQueueFunc = (events: AnalyticsEvent[]) => void;
 
 const ANALYTICS_ENDPOINT = `${apiUrl}/e`;
 
-export default function useAnalyticsQueue(): {
+type UseAnalyticsQueueProps = {
+  method: typeof fetch;
```
idoshamun, line 20: `method` is not explicit enough. Maybe `fetchLike` or `customFetch`, or anything else that hint this is a fetch method

packages/shared/src/hooks/analytics/useAnalyticsSharedProps.ts
```diff
@@ -28,6 +28,7 @@ const getOrGenerateDeviceId = async (): Promise<string> => {
 export default function useAnalyticsSharedProps(
   app: string,
   version: string,
+  deviceId?: string,
```
idoshamun, line 31: It is better to not mix two ways of using this hook:
* Getting the device id as an argument
* The hook loads the device id from storage by itself
Due to the companion constraints, we have to use the first method so I suggest applying it throughout the application

idoshamun (changes requested):
The tracking system looks very good but I have a few comments.
I think we also have to track bookmark / remove bookmark, open / close companion sidebar.
If there are any user interactions that we are not tracking in the companion and I missed please let me know. I'm not very aware to the different versions of the companion.

packages/shared/src/hooks/analytics/useAnalyticsQueue.ts
rebelchris: How does `fetchMethod` sound like?
I don't want to make it sound like it's only the proxy as it can also be the "regular" fetch

packages/shared/src/hooks/analytics/useAnalyticsSharedProps.ts
rebelchris: I'm a bit stuck what's wisdom here.
If we use option1 how will we pass it?

As the current `getOrGenerateDeviceId` is a promise that returns a string.
I don't want to pre-block from the parent components, nor have the option the deviceId is not set yet.

Perhaps passing a promise as the `deviceId` option could be an option?
That way we can either pass the initial one, or a plain string resolve?

rebelchris: I actually created a new look for the device ID, let me know how that looks.

packages/shared/src/hooks/analytics/useAnalyticsQueue.ts
idoshamun: Makes sense ☺️

idoshamun:
@rebelchris What's left to be done in terms of development? Maybe we can roll an internal version to try out.
Also, have you checked how to tackle the manifest change thing?

packages/extension/src/companion/App.tsx
```diff
@@ -32,10 +33,15 @@ export default function App({
...
     settings?.optOutCompanion,
   );
+
   if (isOptOutCompanion) {
     return <></>;
   }
 
+  const memoizedFlags = useMemo(() => {
+    return flags;
+  }, [flags]);
```
idoshamun, line 43: Just my favorite, not required.

```suggestion
  const memoizedFlags = useMemo(() => flags, [flags]);
```

packages/extension/src/companion/App.tsx
```diff
@@ -44,7 +50,7 @@ export default function App({
       </style>
       <RouterContext.Provider value={router}>
         <QueryClientProvider client={queryClient}>
-          <FeaturesContext.Provider value={{ flags }}>
+          <FeaturesContext.Provider value={{ flags: memoizedFlags }}>
```
idoshamun, line 53: The idea of moving to memoized flags was to not create an object on every render.
My suggestion removes the need to create the object every time.

```suggestion
          <FeaturesContext.Provider value={memoizedFlags}>
```

packages/extension/src/companion/CompanionMenu.tsx
```diff
@@ -17,11 +17,22 @@ import { PostBootData } from '@dailydotdev/shared/src/lib/boot';
...
   Modal.setAppElement('daily-companion-app');
 }
 
+const postEventName = (
+  update: Pick<PostBootData, 'upvoted' | 'bookmarked'>,
+): string => {
+  if ('upvoted' in update) {
+    return `${!update.upvoted ? 'remove ' : ''}upvote post`;
+  }
+  return `${!update.bookmarked ? 'remove ' : ''}post bookmark`;
```
idoshamun, line 33: We already have these event names somewhere. Maybe we should consider consolidating it somehow?
I'm afraid that when we change something they will not be aligned. This can be a future work as well

idoshamun (changes requested):
Wow! We're almost there 🤯

idoshamun:
There's an issue with the line between the tooltip and the sidebar. It seems to be not aligned

image

sshanzel:
> There's an issue with the line between the tooltip and the sidebar. It seems to be not aligned

I noticed this as well. Thought this was just an isolated case on my local env. Let me log this.

sshanzel:
@rebelchris feel free to review this commit https://github.com/dailydotdev/apps/pull/930/commits/2de70b7e7a44ea8d53a0681ac2dee8a753f99df7 - it is just related to fetching user info for profile tooltip. Thank you!

rebelchris:
@idoshamun @sshanzel I want to see if we can do a last sanity check on this before we merge into master.
This will make further "bugfixing" easier for the companion in my opinion.

packages/extension/src/companion/useCompanionActions.ts
```diff
@@ -0,0 +1,205 @@
...
+      }),
+    {
+      onMutate: () => undefined,
+      onError: (_, __, rollback) => {
+        rollback?.();
+      },
+    },
+  );
+
+  return {
```
idoshamun, line 194: use memo alert!

idoshamun (approved):
One minor request. Other than that, we're good to go. May the force be with us

sshanzel:
Have verified locally that fixing the PR conflict was a success. The toast notification works as intended in the companion as well.

cc @rebelchris @idoshamun 

Chrome:
image

Firefox:
image

idoshamun:
Thank you for verifying it @sshanzel !

PR #932, opened 2022-03-24
fix: tooltip mention issues

sshanzel:
- Fixes the styling edge case.
- Fixed tooltip animation
- Fixed missing bio of users
- Fixed the multiple instances of tooltip

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

DD-{number} #done

### DIFF
--- a/packages/shared/src/styles/base.css
+++ b/packages/shared/src/styles/base.css
@@ -54,12 +54,14 @@ body {
   --theme-status-help: theme('colors.cheese.40');
   --theme-status-success: theme('colors.avocado.40');
   --theme-status-warning: theme('colors.bun.40');
-  --theme-status-cabbage: theme('colors.cabbage.40'); /* temporary */
+  --theme-status-cabbage: theme('colors.cabbage.40');
   --theme-status-fill: theme('colors.water.40');
 
   --theme-shadow2: theme('boxShadow.2-black');
   --theme-shadow3: theme('boxShadow.3-black');
 
+  --theme-mention: rgba(206, 61, 243, 0.24);
+  --theme-mention-hover: rgba(206, 61, 243, 0.32);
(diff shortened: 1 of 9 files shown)


packages/shared/src/styles/base.css
```diff
@@ -54,12 +54,14 @@ body {
...
   --theme-status-warning: theme('colors.bun.40');
-  --theme-status-cabbage: theme('colors.cabbage.40'); /* temporary */
+  --theme-status-cabbage: theme('colors.cabbage.40');
   --theme-status-fill: theme('colors.water.40');
 
   --theme-shadow2: theme('boxShadow.2-black');
   --theme-shadow3: theme('boxShadow.3-black');
 
+  --theme-mention: rgba(206, 61, 243, 0.24);
+  --theme-mention-hover: rgba(206, 61, 243, 0.32);
```
sshanzel, line 64: Unfortunately, the value like `theme('colors.pepper.70')66` (with 66 to provide opacity percentage on hex values) doesn't provide the accurate color of the requirements. We will have to use the RGB equivalent but it is not being used in our configuration so I had to specifically get the cabbage color specs. Pseudo-elements have been proven to result in edge cases.

rebelchris (approved):
Is the hover calculation not super heavy if a lot of people are tagged?
What was the issue with it calculating it out of the box?

For the rest, it looks good to me, as long as you tested this out fully 🙌

sshanzel:
The issue initially was, I instantiated a tooltip for each mention if it gets hovered, so now, each comment will be sharing a single instance of tooltip and just update the value of it rather than having their own that can be preset at the same time.

Performance-wise regarding hover, I think it should not be too heavy to operate seamlessly, it will be just a regular event handler to set a state.

Thanks for the quick review!

PR #934, opened 2022-03-25
feat: article modal

sshanzel:
- Open the modal by clicking the comment button in the feed
- Restructuring the layout based on the new design (most notable is at tablet size)
- Navigation actions on top of the modal, should not be available on the actual article page
- The navigation should be fixed the user scrolled
- The URL should change whenever the user opens the article modal and navigates through (webapp only)

### WIP
- ~~additional test cases~~
- ~~url change on navigation~~

- [x] MobileL (420px)
- [x] Tablet (656px)
- [x] Laptop (1020px)

DD-211


idoshamun:
@sshanzel I haven't reviewed yet, but I did some QA. I think the design of the header when scrolling is not according to the design so you might want to ask design team. I would love to know what events we track in the article modal?
I think we need to track every time the modal is opened just like we do with page views. We can call it `article modal view` and the relevant post id as `target_id`. We also want to track clicks on the arrows and the existing engagement like bookmark, upvote, impression, click etc (I think this should already exist on the article page).

idoshamun:
Another issue I noticed is that when opening the article modal and then hitting refresh it first shows a 404 page and then immediately update

idoshamun:
Mobile design is also not in place. Please check it with a real device and you will see that the recommended and hot discussions are overlaying the comments. Also before scrolling you can't close the modal. The close buttons shows only after you scroll

sshanzel:
@idoshamun fixed the issue with the mobile design and the header concerns. Currently, the events being tracked is when the modal opens up, when the user navigates left and right, when the user hits "Read article". Then for the pre-existing ones, we now changed the `origin` if the user is using the modal or on the actual page.

cc @rebelchris

idoshamun:
I think when hovering over the author we should show the profile tooltip
image

idoshamun:
When closing the modal it should revert the address bar back to what it was

idoshamun:
When clicking "next" in the article modal it doesn't fetch the next feed page if necessary so it throws an error

packages/shared/src/components/post/PostContent.tsx
```diff
@@ -89,7 +89,7 @@ const PageBodyContainer = classed(
 
 const PostContainer = classed(
   'main',
-  'flex flex-col flex-1 px-8 pb-20 laptop:max-w-[40rem] tablet:border-r tablet:border-theme-divider-tertiary',
+  'flex flex-col flex-1 px-8 tablet:pb-20 laptop:max-w-[40rem] tablet:border-r tablet:border-theme-divider-tertiary',
```
idoshamun, line 92: Please make sure it doesn't break the article page

packages/shared/src/components/post/PostContent.tsx
```diff
@@ -226,10 +230,11 @@ export function PostContent({
     return <Custom404 />;
   }
 
+  const analyticsOrigin = navigation ? 'article page' : 'article modal';
```
idoshamun, line 233: This condition is a bit tricky and not future proof in my opinion

packages/shared/src/components/post/PostModalActions.tsx
```diff
@@ -56,6 +62,14 @@ export function PostModalActions({
     });
   };
 
+  const onReadArticle = () => {
+    trackEvent(
+      postAnalyticsEvent('click', post, {
```
idoshamun, line 67: Let's change the event name so we will not be confused with the click on the card.
I know we used to have `click` before on the article page but now it's a great time to change it.
Also is it used by the article page as well?

```suggestion
      postAnalyticsEvent('go to link', post, {
```

packages/shared/src/hooks/analytics/useAnalyticsContextData.ts
```diff
@@ -14,6 +14,8 @@ const generateEventId = (now = new Date()): string => {
   return `${timePart}${randomStr}`;
 };
 
+export type PostOrigin = 'article page' | 'article modal';
```
idoshamun, line 17: I think it's a duplicate

idoshamun (changes requested):
Looks great almost ready. I dropped a few comments outside the review following my QA

packages/shared/src/components/post/PostContent.tsx
sshanzel: Yes. These changes were actually intentional since there was a change in the actual layout.

packages/shared/src/components/post/PostModalActions.tsx
sshanzel: Yes. This button will be available on the article page as well.

packages/shared/src/hooks/analytics/useAnalyticsContextData.ts
sshanzel: I've kept this one and removed the other one now. Thank you for the notice!

packages/shared/src/components/Feed.tsx
```diff
@@ -131,6 +131,7 @@ export default function Feed<T>({
     insaneMode,
     loadedSettings,
   } = useContext(SettingsContext);
+  const [isFetchingNextPage, setIsFetchingNextPage] = useState(false);
```
idoshamun, line 134: You should use the state that is returned from `useFeed` for this purpose. This hook controls the requests for the feed. I think we already have an existing mechanism that prevents from fetching another page if there's a pending request. You can check inside the hook itself

packages/shared/src/components/Feed.tsx
```diff
@@ -283,25 +284,42 @@ export default function Feed<T>({
...
 
+    if (items[index + 1].type === 'placeholder') {
+      return null;
+    }
+
+    setIsFetchingNextPage(false);
     const offset = items[index + 1].type !== 'post' ? 2 : 1;
     const item = items[index + offset];
 
     if (item?.type !== 'post') {
```
idoshamun, line 304: This will cause a weird behavior that if I hit next and there's an ad the modal will close itself right?

packages/shared/src/components/Feed.tsx
```diff
@@ -283,25 +284,42 @@ export default function Feed<T>({
...
-    setOpenedPost(item.post.id);
+
+    return setOpenedPost(item.post.id);
   };
 
+  useEffect(() => {
+    if (openedPost && isFetchingNextPage) {
+      onNext();
+    }
+  }, [items, openedPost, isFetchingNextPage]);
```
idoshamun, line 321: Why do we need this?

packages/shared/src/components/modals/PostModal.tsx
```diff
@@ -35,7 +36,7 @@ export function PostModal({
   }, [id]);
 
   const onClose: typeof onRequestClose = (e) => {
-    if (isExtension) {
+    if (!isExtension) {
       window.history.replaceState({}, `Feed`, '/');
```
idoshamun, line 40: It's not necessarily the home page. It can be any feed page

idoshamun (changes requested):
Please also let me know once you fix the issue I raised before in the PR as part of my QA

packages/shared/src/components/Feed.tsx
sshanzel: Because we will have to wait for the items to get updated when you reach the end of the list. I tried doing async in a single function and re-calling the `onNext` in itself but it was still referencing the same object (old list). So I had to ensure the object was updated already before moving forward.

packages/shared/src/components/Feed.tsx
sshanzel: No. This was just required because of typescript. But handling the ad will be at line 301 where it decides how much should be the offset.

packages/shared/src/components/Feed.tsx
sshanzel: Yes, but we had to create a separate one since it might not know when the user actually clicked the next button. If the user just scrolled the background while the modal is open, it would trigger the `onNext`.

packages/shared/src/components/Feed.tsx
```diff
@@ -144,6 +142,14 @@ export default function Feed<T>({
       variables,
     );
   const { ranking } = (variables as RankVariables) || {};
+  const {
+    onOpenModal,
+    onCloseModal,
+    onPrevious,
+    onNext,
+    article,
```
idoshamun, line 150: Looking at it again, we should consider renaming this variable.
Suggestion: `selectedArticle` / `selectedPost`

packages/shared/src/components/Feed.tsx
```diff
@@ -330,11 +263,11 @@ export default function Feed<T>({
...
-      {openedPost && (
+      {article && (
         <PostModal
-          id={openedPost}
-          isOpen={!!openedPost}
-          onRequestClose={() => setOpenedPost('')}
+          isOpen
+          id={article.id}
+          onRequestClose={onCloseModal}
           navigation={{ onPreviousPost: onPrevious, onNextPost: onNext }}
```
idoshamun, line 271: It'd better to split this prop to `onPreviousPost` and `onNextPost` rather than a single prop

packages/shared/src/components/modals/PostModal.tsx
```diff
@@ -22,6 +22,7 @@ export function PostModal({
   isFetchingNextPage,
   ...props
 }: PostModalProps): ReactElement {
+  const [currentPage, setCurrentPage] = useState('');
```
idoshamun, line 25: It's more accurate this way, rather than using empty string

```suggestion
  const [currentPage, setCurrentPage] = useState<string>();
```

packages/shared/src/components/modals/PostModal.tsx
```diff
@@ -32,12 +33,17 @@ export function PostModal({
...
+    }
+
     window.history.replaceState({}, `Post: ${id}`, `/posts/${id}`);
   }, [id]);
 
   const onClose: typeof onRequestClose = (e) => {
     if (!isExtension) {
-      window.history.replaceState({}, `Feed`, '/');
+      window.history.replaceState({}, `Feed`, currentPage);
+      setCurrentPage('');
```
idoshamun, line 46:
```suggestion
      setCurrentPage();
```

packages/shared/src/components/modals/PostModal.tsx
```diff
@@ -32,12 +33,17 @@ export function PostModal({
...
 
+    if (!currentPage) {
+      setCurrentPage(window.location.pathname);
+    }
+
     window.history.replaceState({}, `Post: ${id}`, `/posts/${id}`);
   }, [id]);
 
   const onClose: typeof onRequestClose = (e) => {
     if (!isExtension) {
```
idoshamun, line 44:
```suggestion
    if (!isExtension && currentPage) {
```

idoshamun (changes requested):
Some minor comments and we're ready to roll as far as I care 😉

packages/shared/src/components/modals/PostModal.tsx
```diff
@@ -0,0 +1,75 @@
...
+    }
+    onRequestClose(e);
+  };
+
+  useEffect(() => {
+    trackEvent({
+      event_name: 'article modal view',
+      target_id: id,
+    });
+  }, []);
```
rebelchris, line 59: This should actually fire based on the `id` prop as we need this for the tracking.

packages/shared/src/components/post/PostContent.tsx
```diff
@@ -0,0 +1,386 @@
...
+  useSubscription(
+    () => ({
+      query: POSTS_ENGAGED_SUBSCRIPTION,
+      variables: {
+        ids: [id],
+      },
+    }),
+    {
+      next: (data: PostsEngaged) => {
+        queryClient.setQueryData<PostData>(postQueryKey, (oldPost) => ({
```
rebelchris: Keep in mind to not overwrite the fix we did today.

packages/shared/src/components/Feed.tsx
sshanzel: This component actually extends prop from `ModalProps` but just wanted to isolate the props for navigation. Though we can definitely separate it. Let me refactor it quickly! 🚢

packages/shared/src/components/post/PostModalActions.tsx
```diff
@@ -0,0 +1,135 @@
...
+export interface PostModalActionsProps {
+  post: Post;
+  onClose?: MouseEventHandler | KeyboardEventHandler;
+  className?: string;
+  style?: CSSProperties;
+  inlineActions?: boolean;
+  origin?: PostOrigin;
+}
+
+const Container = classed('div', 'flex flex-row items-center');
```
rebelchris, line 37: Not breaking, but perhaps we should consider making these classes generic at one stage

packages/shared/src/components/post/PostNavigation.module.css
```diff
@@ -0,0 +1,14 @@
+.fixedPostsTitle {
+  overflow: hidden;
+  white-space: nowrap;
+  text-overflow: ellipsis;
+  max-width: calc(100vw - 18.75rem);
```
rebelchris, line 5: Can we not move these to tailwind classes?

sshanzel:
@idoshamun thank you for the awesome review! I have now applied the refactoring! 🚢 🚢 🚢

packages/shared/src/hooks/usePostModalNavigation.ts
```diff
@@ -0,0 +1,89 @@
...
+        for (; index > 0 && items[index].type !== 'post'; index -= 1);
+        const item = items[index];
+        if (item.type !== 'post') {
+          return;
+        }
+
+        trackEvent({
+          origin: 'article modal',
+          event_name: 'navigate previous',
+          target_id: item.post.id,
```
rebelchris, line 39: This should be the old post (so the one we are coming from)
And we should actually use `postAnalyticsEvent` for all these, as it tracks post data.

packages/shared/src/hooks/usePostModalNavigation.ts
```diff
@@ -0,0 +1,89 @@
...
+
+        if (item.type !== 'post') {
+          return;
+        }
+
+        setIsFetchingNextPage(false);
+        trackEvent({
+          origin: 'article modal',
+          event_name: 'navigate next',
+          target_id: item.post.id,
```
rebelchris, line 70: Same comment here as the one above

rebelchris (changes requested):
Wow, really blown away by what you put down here! 🤘

rebelchris:
Long titles in You might also like are not breaking (we fixed this in the main cards)

Screenshot 2022-04-20 at 15 47 32

rebelchris:
Locally I get two tooltip windows open over each other, not sure where the bottom one comes from, and it's only locally.
So no big rush, but maybe it rings a bell for you.

Screenshot 2022-04-20 at 15 48 48

rebelchris:
The space is missing here in your version.

Screenshot 2022-04-20 at 15 51 28

packages/shared/src/components/Feed.spec.tsx
```diff
@@ -840,3 +843,81 @@ it('should block a tag', async () => {
...
+  const [first] = await screen.findAllByLabelText('Comments');
+  fireEvent.click(first);
+
+  await screen.findByRole('dialog');
+  const title = await screen.findByTestId('post-modal-title');
+  expect(title).toHaveTextContent(firstPost.node.title);
+
+  const nav1 = await screen.findByRole('navigation');
+  // eslint-disable-next-line testing-library/no-node-access
+  const [, next] = Array.from(nav1.getElementsByTagName('button'));
```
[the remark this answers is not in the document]
sshanzel: I just copy-pasted this from the previous file, not really sure about this tho.

packages/shared/src/components/modals/PostModal.tsx
sshanzel: I agree, though we track the changes in the navigation part where we also include the id.

packages/shared/src/components/post/PostContent.tsx
sshanzel: This should be fixed by the recent merge now. Thank you for pointing it out!

packages/shared/src/components/post/PostModalActions.tsx
sshanzel: Totally agree. We probably need to come up with a better naming on that one.

packages/shared/src/components/widgets/PostUsersHighlights.tsx
```diff
@@ -0,0 +1,136 @@
...
+        userType="source"
+      />
+      {author && <UserHighlight {...author} userType="author" />}
+      {featuredComments?.map((comment) => (
+        <UserHighlight
+          key={comment.id}
+          {...comment.author}
+          userType="featured"
+        />
+      ))}
```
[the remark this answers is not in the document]
sshanzel: Right. Let me remove this. Had a wrong perception on what to do with this part.

packages/shared/src/components/Custom404.tsx
```diff
@@ -0,0 +1,24 @@
...
+    <div
+      className="flex flex-col justify-center items-center mx-auto min-h-screen"
+      data-testid="notFound"
+    >
+      {children}
+      <HelloWorldSvg
+        style={{ width: '55%', maxWidth: '32.75rem' }}
+        className="self-center mb-10"
+      />
+      <h1 className="mx-9 font-bold text-center break-words-overflow typo-title1">
```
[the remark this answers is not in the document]
sshanzel: I just copy-pasted this from the previous file, not really sure about this tho.

packages/shared/src/components/Feed.spec.tsx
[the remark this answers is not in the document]
sshanzel: Perfect! Thank you for the tip!

sshanzel:
@rebelchris thank you for the amazing review! Concerns should now be fixed! 🚢 🚢 🚢

packages/shared/src/hooks/usePostModalNavigation.ts
```diff
@@ -64,14 +65,14 @@ export const usePostModalNavigation = (
         }
 
         setIsFetchingNextPage(false);
-        trackEvent({
-          origin: 'article modal',
-          event_name: 'navigate next',
-          target_id: item.post.id,
-        });
+        trackEvent(
+          postAnalyticsEvent('navigate next', item.post, {
```
idoshamun, line 69: We should use here the current post, not the new one.
This makes more sense from the perspective of data analysis.

packages/shared/src/components/modals/PostModal.tsx
sshanzel: Since there will be some changes in the tracking of navigation, I shall apply your comment here.

sshanzel:
Since I am a huge fan of keyboard navigation, I have also included the navigation for hitting left and right going previous and next respectively.

Reference: https://dailydotdev.slack.com/archives/C02AA18RD1D/p1648317972666129?thread_ts=1648317600.314799&cid=C02AA18RD1D

packages/shared/src/components/modals/PostModal.module.css
```diff
@@ -0,0 +1,40 @@
...
+    border: none;
+    border-radius: 0;
+
+    @screen responsiveModalBreakpoint {
+      position: relative;
+      min-height: unset;
+    }
+
+    @screen laptop {
+      max-height: calc(100% - 7rem);
```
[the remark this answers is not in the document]
rebelchris: Taking that into account perhaps you want to check with Tsahi/Tomer if they don't want the scrolled header to have the border radius?

packages/shared/src/components/modals/PostModal.tsx
rebelchris: I meant in the useEffect we should put the id as a dependency, that way we are sure this fires correct.

packages/shared/src/components/modals/PostModal.tsx
```diff
@@ -58,6 +58,33 @@ export function PostModal({
     });
   }, [id]);
 
+  useEffect(() => {
```
rebelchris, line 61: Love this, but would personally consider moving this to a custom hook, it will make this file a bit less bloated and reusable for the future.

the hook could be something like useKeyboardNavigation and it can accept certain valid keys, en possible return each valid key as a function like onValid{Key}Press

Let me know what you think.

rebelchris (changes requested):
Wow amazing and the keyboard is really ground breaking, I was telling Ido I missed that during our review yesterday so well done 👏

One small remark on the keyboard navigation, but up to you if you like to refactor now or add a note for a new ticket?

And then one small thing left from my side to validate with design team.
Otherwise good to go 🚀

rebelchris:
Also still have the header behaving super weird on iOS (Chrome)
It works way better than yesterday, but the header kind of moves between full screen and what you see on the image below.

there also seems some horizontal drift (like 2px's) but during scroll this gives you a drunk scroll behaviour (from left to right)
Let me know if it will help to show you perhaps?

IMG_5381

rebelchris:
The message on the modal is fixed to the buttons and the post never actually hides.
Perhaps we should ask the product team what should happen in this instance?

Screenshot 2022-04-21 at 07 17 32

packages/shared/src/components/modals/PostModal.tsx
sshanzel: Yes. That wasn't necessary initially since in the navigation we provide the new ids. Now that we changed the parameters for the navigation and uses the old id we now will apply the `id` as an actual dependency here.

packages/shared/src/components/modals/PostModal.tsx
idoshamun: I agree

packages/shared/src/components/modals/PostModal.tsx
```diff
@@ -56,7 +56,34 @@ export function PostModal({
...
+  }, [id]);
+
+  useEffect(() => {
+    const arrowkeyNavigation = (e: KeyboardEvent & { path: HTMLElement[] }) => {
+      if (e.key !== 'ArrowLeft' && e.key !== 'ArrowRight') {
+        return;
+      }
+
+      const [base] = e.path;
+      const inputs = ['select', 'input'];
```
idoshamun, line 68:
```suggestion
      const inputs = ['select', 'input', 'textarea'];
```

packages/shared/src/components/modals/PostModal.tsx
```diff
@@ -0,0 +1,106 @@
...
+  const onClose: typeof onRequestClose = (e) => {
+    if (!isExtension && currentPage) {
+      window.history.replaceState({}, `Feed`, currentPage);
+      setCurrentPage(undefined);
+    }
+    onRequestClose(e);
+  };
+
+  useEffect(() => {
+    trackEvent({
```
idoshamun, line 55: This should be a post event as well. Please check if there are any other events that we might have missed during the review

idoshamun (changes requested):
Can't wait already

packages/shared/src/components/modals/PostModal.module.css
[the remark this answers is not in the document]
sshanzel: Let me create a scheduled message for them since it's holiday in their place.

sshanzel:
@rebelchris I have refactored the structure of the fixed header a bit, I think it is much more stable now. Thank you!

sshanzel:
@rebelchris thank you! Yes, I noticed that as well but it seems to be a markdown issue with links not breaking to two lines. Let me create a ticket for this.

packages/shared/src/components/post/PostContent.tsx
```diff
@@ -280,22 +289,24 @@ export function PostContent({
...
+              className={classNames(
+                'flex',
+                padding,
+                position,
+                isFixed &&
+                  'z-3 w-full bg-theme-bg-secondary border-b border-theme-divider-tertiary px-6',
+                isFixed && styles.fixedPostsNavigation,
+              )}
+              shouldDisplayTitle={isFixed}
+              postActionsProps={{ post: postById.post, onClose }}
```
idoshamun, line 306: It's not advised to create a new object on every render, so consider splitting it into two different props. Nothing blocking though.

idoshamun (approved):
Yay!! Check out my comments but you have my approval 💯

rebelchris (approved):
Besides the notification margin we are good to go.
Perhaps just add the margin-left same as the button has? 🤘

PR #936, opened 2022-03-25
fix: interactiveness of mention popup

sshanzel:
- In my previous PR we introduced manual handling of state on where we should display the tooltip or not. Unfortunately, it introduced an edge case where the popup lost its interactiveness.
- Also refactored the use debounce hook to be cancellable
- Fixed the primary concern in this PR

Verified workability in this link: https://daily-webapp-i28cjgu9g-dailydotdev.vercel.app/posts/RcIbqXVDf

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

DD-{number} #done

### DIFF
--- a/packages/shared/src/components/Markdown.tsx
+++ b/packages/shared/src/components/Markdown.tsx
@@ -4,6 +4,7 @@ import styles from './markdown.module.css';
 import { ProfileTooltip } from './profile/ProfileTooltip';
 import { useProfileTooltip } from '../hooks/useProfileTooltip';
 import { CaretOffset } from '../lib/element';
+import useDebounce from '../hooks/useDebounce';
 
 interface MarkdownProps {
   content: string;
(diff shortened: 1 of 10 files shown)


rebelchris (changes requested):
I think the code is fine, but I did spot two issues (probably not related to this issue)

Issue one:
When scrolled up top the mention bounces in and out by loosing focus:
test

And issue two is a bit of the opposite
When scrolling the mention down it shows the popup half way over the mention.
The orange line is actually where the mention sits.
Screenshot 2022-03-25 at 14 25 42

Would you mind looking into those?
Especially issue 1?

sshanzel:
@rebelchris thank you for the review! There is indeed a bug where it pops in and out. Regarding the second issue. Unfortunately, that's unavoidable, it is due to the fact that our tooltip has large height and when the screen doesn't fit anymore on both ends, it would need to adjust a bit which results to tooltip overlapping the reference. What do you think?

sshanzel:
@rebelchris let me know if you want me to make a bit more adjustments here. Thank you! 🙌

rebelchris (approved):
@sshanzel Nice work, looks to work way better now!
I did note sometimes it takes a very long time to show up (especially for Ido)
Just remember to log a ticket for that.

sshanzel:
@rebelchris thank you for the review! Ah, yes, let me log it now! 🚢

PR #943, opened 2022-03-31
fix: explicitly defining the start date of heatmap

sshanzel:
- The core issue was when generating the heatmap - we have a function that creates an array, based on the week columns we will have to create.
- The concern came from the way we generate the start date - we utilize the function `new Date(0)` and set its year. Unfortunately, it looks like from `UTC-1` and lower, the value it returns is `31-12-1969` which is supposed to be `01-01-1970`.
- To avoid the incorrect start date, I have now explicitly defined that the calculation should start at `01-01-YYYY` based on the selected year.

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

## Relevant Issues
- https://github.com/dailydotdev/daily/issues/501

DD-573 #done


rebelchris (approved):
Wohoo, so happy you found and solved this bug 🙌

PR #950, opened 2022-04-18
build: update next-pwa to the latest version

idoshamun:
I think the overload cache issue was caused only for users who used the webapp for a long time with an old service worker definition. The old SW didn't set a limit on the image cache, thus causing it to explode (my local cache was 4GB 🤯).

I update next-pwa to the latest version to make sure we are aligned with the newest settings and also confirmed locally that cache is limited to ~300mb.
I'm still not 100% what takes up all this storage because I tried to delete the caches one by one and still remained with ~170mb. But I think 300mb is a decent cap for now.

Closes dailydotdev/daily#372
DD-437 #done

### DIFF
--- a/packages/extension/package-lock.json
+++ b/packages/extension/package-lock.json
@@ -1,6 +1,6 @@
 {
 	"name": "extension",
-	"version": "3.15.4",
+	"version": "3.15.5",
 	"lockfileVersion": 1,
 	"requires": true,
 	"dependencies": {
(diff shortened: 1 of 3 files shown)


rebelchris (approved):
Very interesting, happy to see it's solved in the latest version.
Thanks for debugging this.

Let's plan a new rollout for this week.

idoshamun:
It happens only on the webapp. The extension doesn't have a service worker on.
image

PR #951, opened 2022-04-20
fix: do not update article page on posts engaged event

idoshamun:

Earlier we removed the filter functionality of the `postsEngaged` subscription. This caused the article page to receive every event that is occurring in the system.
Unfortunately, we didn't handle the case where we receive an event of another post in the callback.
This is now fixed :)

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

DD-597 #done


rebelchris (approved):
Nice work 🙌

PR #953, opened 2022-04-27
fix: window being undefined

sshanzel:
- The reason window is null, is when we navigate to the source page, the getStaticProps doesn't have the context of window yet, so it crashes.

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

DD-{number} #done

### DIFF
--- a/packages/shared/src/hooks/usePostModalNavigation.ts
+++ b/packages/shared/src/hooks/usePostModalNavigation.ts
@@ -83,8 +83,10 @@ export const usePostModalNavigation = (
     [items, openedPostIndex, isFetchingNextPage],
   );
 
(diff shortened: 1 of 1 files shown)


idoshamun (commented):
How come window is null? And what happens in the keyboard hook when it receives null?

sshanzel:
It will return and do nothing, but calling `window` itself is resulting to crash is it is not available in the actual scope.

sshanzel:
The reason window is null, is when we navigate to the `source` page, the `getStaticProps` doesn't have the context of `window` yet, so it crashes.

idoshamun (approved):

PR #957, opened 2022-04-27
fix: upvote modal user list profile tooltip

sshanzel:
- Modified user short profile info component to include profile tooltip (can be disabled, for the case of mention recommendations)
- Forwards reference Profile Picture component.
- includes id in the graphql request for upvoters

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [x] Laptop (1020px)

DD-598 #done


rebelchris (approved):
Nice work!

The only thing I have while using it and scrolling through a long list of upvotes is that the modal pops up twice then.

Perhaps check with product/design what the desired flow is on scroll?

packages/shared/src/components/profile/ProfileTooltip.tsx
```diff
@@ -38,9 +41,29 @@ export function ProfileTooltip({
...
+  const handler = useRef<() => void>();
+
+  const onShow = () => {
+    if (!scrollingContainer) {
+      return;
+    }
+
+    scrollingContainer.removeEventListener('scroll', handler.current);
+    handler.current = () => query.setQueryData(['readingRank', user.id], {});
+    scrollingContainer.addEventListener('scroll', handler.current);
```
rebelchris, line 54: I'm not a 100% sure why we need to manually set the queryData here on the scroll listener?
Is a regular reference to the element not enough?

rebelchris (changes requested):
Nice fix, it does solve the issue I had.
However one small question around the scrollingContainer variable.

packages/shared/src/components/profile/ProfileTooltip.tsx
sshanzel: This is to clear the current tooltip and close it after scrolling - we force to empty the data so it would be unmounted.

> Is a regular reference to the element not enough?
That's actually a great point, was trying to make it abide with React's approach. Should we move to native and just pass a selector? What do you think?

rebelchris: I think it might be good enough with regular reference as this might be a bit overkill,  but leave it up to you if this is needed for some other reason then it can make sense.

sshanzel: I think for the moment, we can move this now. Once we find other areas requiring this implementation, I think that can be worth moving this existing implementation to query selector.

PR #958, opened 2022-04-27
feat: feed filters for new users

ayadotdev:
#### 1 step.
New users are welcomed to create My feed before they are being let into daily.dev. There are 2 versions of the intro modal:
 
1. With no skip button:
Screenshot 2022-04-27 at 21 38 19

2. With Skip button:
Screenshot 2022-04-27 at 21 38 49

#### 2 step.
After clicking Continue/Create my feed button the filtering process should remain the same. E.x. Any Feed filter modal:
Screenshot 2022-04-27 at 21 41 47

https://www.loom.com/share/f69a8ac7d84340a7a7041d4977a9bb95

Done: 
- [x] used usePersistantContext to detect if the user is the first
- [x] created a wrapper component with Intro Modal and Feed filter modal
- [x] added a wrapper component to be rendered in Sidebar component

- [x] MobileL (420px)
- [x] Tablet (656px)
- [x] Laptop (1020px)

DD-546 #done
https://www.loom.com/share/ed88da4ac9184320b1cb7db1a3d8b224


packages/shared/src/components/modals/FeedFiltersIntroModal.tsx
```diff
@@ -0,0 +1,110 @@
...
+  onOpenFeedFilterModal: (event: React.MouseEvent<Element, MouseEvent>) => void;
+} & Pick<ModalProps, 'onRequestClose'>;
+
+type FeedFiltersIntroModalProps = FeedFiltersProps &
+  ModalFooterProps &
+  ModalProps;
+
+const footerClass = {
+  test1: 'justify-center',
+  test2: 'justify-between',
```
rebelchris, line 26: I think it might be better to be more explicit as `introTest1` for instance

packages/shared/src/components/modals/FeedFiltersIntroModal.tsx
```diff
@@ -0,0 +1,110 @@
...
+}: FeedFiltersIntroModalProps): ReactElement {
+  const tagsRows = [
+    ['', 'docker', '', 'kubernetes', ''],
+    ['', '', 'architecture', '', ''],
+    ['', '', '', 'devops', ''],
+    ['', 'cloud', '', '', ''],
+  ];
+
+  return (
+    <ResponsiveModal className={classNames(className)} {...modalProps}>
```
rebelchris, line 79: I think the className is obsolete here as we don't refer to it on the parent.

```suggestion
    <ResponsiveModal {...modalProps}>
```

packages/shared/src/components/modals/FeedFiltersIntroModal.tsx
```diff
@@ -0,0 +1,110 @@
...
+      <section className="flex overflow-hidden flex-col items-center py-6 px-6 mobileL:px-10 mt-24">
+        <UserIcon className="my-3 mx-2 w-12 h-12" />
```
rebelchris, line 81: I think you need to adjust the size of the icon by using `text-{}` (not sure what the correct one is)

packages/shared/src/components/modals/FeedFiltersIntroModal.tsx
```diff
@@ -0,0 +1,110 @@
...
+        <h3 className="mt-1 font-bold typo-large-title">Create my feed</h3>
+        <p className="mt-3 mb-16 text-center typo-title3 text-theme-label-tertiary">
+          Devs with a personal feed get 11.5x more relevant articles
```
rebelchris, line 84: We should use the existing AB test here: MyFeedExplainerCopy

packages/shared/src/components/modals/FeedFiltersIntroModal.tsx
```diff
@@ -0,0 +1,110 @@
...
+    ['', '', 'architecture', '', ''],
+    ['', '', '', 'devops', ''],
+    ['', 'cloud', '', '', ''],
+  ];
+
+  return (
+    <ResponsiveModal className={classNames(className)} {...modalProps}>
+      <section className="flex overflow-hidden flex-col items-center py-6 px-6 mobileL:px-10 mt-24">
+        <UserIcon className="my-3 mx-2 w-12 h-12" />
+        <h3 className="mt-1 font-bold typo-large-title">Create my feed</h3>
```
rebelchris, line 82: Also here, we should use the existing AB test: MyFeedButtonCopy

packages/shared/src/components/modals/FeedFiltersIntroModal.tsx
```diff
@@ -0,0 +1,110 @@
...
+        <p className="mt-3 mb-16 text-center typo-title3 text-theme-label-tertiary">
+          Devs with a personal feed get 11.5x more relevant articles
+        </p>
+        <FeedFiltersIntroModalTagsContainer className="mb-24">
```
rebelchris, line 86: Can we not put the className directly on this component

packages/shared/src/components/modals/FeedFiltersIntroModalTags.tsx
```diff
@@ -0,0 +1,45 @@
...
+};
+
+const getPillWidth = (isFirstTagPill: boolean, isLastTagPill: boolean) => {
+  if (isFirstTagPill || isLastTagPill) {
+    return 'flex-1';
+  }
+  return getRandomTagWidth();
+};
+
+const FeedTagItem = ({
```
rebelchris, line 25: This name doesn't represent the file

packages/shared/src/components/modals/FeedFiltersIntroModalTags.tsx
```diff
@@ -0,0 +1,45 @@
...
+  tag,
+  isLastTagPill,
+  isFirstTagPill,
+}: FeedTagItemProps): ReactElement => {
+  const pillWidth = useMemo(
+    () => getPillWidth(isFirstTagPill, isLastTagPill),
+    [isFirstTagPill, isLastTagPill],
+  );
```
rebelchris, line 33: I think we can use the global hook

```suggestion
  const pillWidth = useMemo(
    () => getPillWidth(isFirstTagPill, isLastTagPill),
    [],
  );
```

packages/shared/src/components/modals/FeedFiltersIntroModalTags.tsx
```diff
@@ -0,0 +1,45 @@
+import React, { ReactElement, useMemo } from 'react';
+import { FeedFiltersIntroModalTagPill } from '../utilities';
+
+type FeedTagItemProps = {
+  tag: string;
+  isLastTagPill: boolean;
+  isFirstTagPill: boolean;
```
rebelchris, line 7: Naming could be more like `isLast` `isFirst`

packages/shared/src/components/modals/FeedFiltersIntroModalTags.tsx
```diff
@@ -0,0 +1,45 @@
...
+};
+
+const tagPillWidths = ['w-[4.5rem]', 'w-24', 'w-[7.5rem]'];
```
rebelchris, line 10: We can safely use `widths`

packages/shared/src/components/modals/FeedFiltersIntroModalTags.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+const getRandomTagWidth = () => {
```
rebelchris, line 12: getRandomWidth

packages/shared/src/components/modals/FeedFiltersIntroModalTags.tsx
```diff
@@ -0,0 +1,45 @@
...
+  const randomIndex = Math.floor(Math.random() * tagPillWidths.length);
+  const randomWidthValue = tagPillWidths[randomIndex];
+  return randomWidthValue;
+};
+
+const getPillWidth = (isFirstTagPill: boolean, isLastTagPill: boolean) => {
```
rebelchris, line 18: getWidth

packages/shared/src/components/CreateFeedFilterButton.tsx
```diff
@@ -11,12 +11,14 @@ import { Button, ButtonProps } from './buttons/Button';
 
 type TypeProps = {
   feedFilterModalType: string;
+  buttonText: string;
```
rebelchris, line 14: Why did we make this a property

packages/shared/src/components/modals/FeedFiltersWrapperModal.tsx
```diff
@@ -54,30 +62,35 @@ const ModalFooter = ({
...
         )}
         feedFilterModalType={feedFilterModalType}
+        buttonText="Create"
       />
     </footer>
   );
 };
 
-export default function CreateMyFeedModal({
+const FeedFiltersModal = ({
```
rebelchris, line 71: Perhaps we should move this to it's own file

packages/shared/src/components/modals/RanksModal/common.tsx
```diff
@@ -75,6 +75,6 @@ export const RanksTagsList = classed(
 );
 export const RankTagPill = classed(
   'strong',
-  'flex items-center px-3 ml-2 h-6 truncate rounded-8 typo-callout text-theme-label-tertiary',
+  'flex items-center h-6 truncate rounded-8 typo-callout text-theme-label-tertiary',
```
rebelchris, line 78: Did you make sure this doesn't break existing implementations

packages/shared/src/components/sidebar/Sidebar.tsx
```diff
@@ -156,10 +163,35 @@ export default function Sidebar({
...
+
   const shouldShowDnD = !!process.env.TARGET_BROWSER;
   const { flags } = useContext(FeaturesContext);
   const popularFeedCopy = getFeatureValue(Features.PopularFeedCopy, flags);
   const feedFilterModal = getFeatureValue(Features.FeedFilterModal, flags);
+  const feedFilterModalOnboarding = getFeatureValue(
+    Features.FeedFilterModalOnboarding,
+    flags,
+  );
+  const FIRST_TIME_SESSION = 'firstTimeSession';
```
rebelchris, line 176: This can be outside of the function (top of component)

packages/shared/src/components/sidebar/Sidebar.tsx
```diff
@@ -156,10 +163,35 @@ export default function Sidebar({
...
+      if (feedFilterModalOnboarding === 'control') {
+        setTimeout(() => {
+          openFeedFilters();
+        }, 300);
+      } else {
+        openFeedFilters();
+        setShowIntroModal(true);
+      }
+    }
+  }, [isFirstSession]);
```
rebelchris, line 194: We don't want to listen to this change, it can simply fire once, so you can use a empty array `[]`

packages/shared/src/components/sidebar/Sidebar.tsx
```diff
@@ -156,10 +163,35 @@ export default function Sidebar({
...
+    FIRST_TIME_SESSION,
+    true,
+  );
+  useEffect(() => {
+    if (isFirstSession) {
+      setIsFirstSession(false);
+      if (feedFilterModalOnboarding === 'control') {
+        setTimeout(() => {
+          openFeedFilters();
+        }, 300);
```
rebelchris, line 188: Not ideal, perhaps we can use @sshanzel implementation of waiting for the assets to load (used in the companion)
Or any other suggestions perhaps?

The main issue here is that we use the dynamic hook, but it's not loaded yet.

packages/shared/src/lib/featureManagement.ts
```diff
@@ -98,6 +98,12 @@ export class Features {
     'v5',
   ]);
 
+  static readonly FeedFilterModalOnboarding = new Features(
+    'feed_filter_modal_onboarding',
+    'control',
+    ['control', 'test1', 'test2'],
```
rebelchris, line 104: Perhaps we should be more explicit here by adding intro prefix to it.

rebelchris (changes requested):
On the right track for sure!
Great work already, some small refactor comments, hit me up when it's ready for a next review.

ayadotdev:
@rebelchris Chris, thanks you for the review, take a look, when you can if I managed to fix everything.

packages/shared/src/components/modals/FeedFiltersModal/FeedFiltersIntroModal.tsx
```diff
@@ -0,0 +1,117 @@
...
+  feedFilterModalType,
+  onOpenFeedFilterModal,
+  ...modalProps
+}: FeedFiltersIntroModalProps): ReactElement {
+  const tagsRows = [
+    ['', 'docker', '', 'kubernetes', ''],
+    ['', '', 'architecture', '', ''],
+    ['', '', '', 'devops', ''],
+    ['', 'cloud', '', '', ''],
+  ];
```
rebelchris, line 78: I think we can safely move this outside of the component (so top of file)

packages/shared/src/components/modals/FeedFiltersModal/FeedFiltersIntroModal.tsx
```diff
@@ -0,0 +1,117 @@
...
+  const introExplainerCopy = getFeatureValue(
+    Features.MyFeedExplainerCopy,
+    flags,
+  );
+
+  return (
+    <ResponsiveModal className={className} {...modalProps}>
+      <section className="flex overflow-hidden flex-col items-center py-6 px-6 mobileL:px-10 mt-24">
+        <UserIcon className="typo-giga2" />
+        <h3 className="mt-4 font-bold typo-large-title">Create my feed</h3>
```
rebelchris, line 89: I think you forgot the flagged AB test version here.

packages/shared/src/components/modals/FeedFiltersModal/FeedFiltersIntroModal.tsx
```diff
@@ -0,0 +1,117 @@
...
+          Skip
+        </Button>
+      )}
+
+      <Button
+        className="mr-4 w-40 btn-primary-cabbage"
+        onClick={onOpenFeedFilterModal}
+      >
+        {feedFilterOnboardingModalType === 'introTest1'
+          ? 'Create my feed'
```
rebelchris, line 57: This probably also needs the MyFeedButtonCopy ab test

rebelchris (changes requested):
Nice work, really looks way cleaner!
Had some minor remarks and looking forward to seeing you and Lee review it as well 🙌

packages/shared/src/components/modals/FeedFiltersModal/FeedFiltersIntroModal.tsx
ayadotdev: Even if it is the button that only opens the actual Feed filter?

packages/shared/src/components/modals/FeedFiltersModal/FeedFiltersIntroModal.tsx
ayadotdev: Should it be MyFeedButtonCopy?

sshanzel (changes requested):
Great job! Super amazing to see you worked on this large A/B testing that we have. Just have few very minor comments, but you are definitely right on track! 🚢

packages/shared/src/components/modals/FeedFiltersModal/FeedFiltersIntroModal.tsx
```diff
@@ -0,0 +1,117 @@
...
+  return (
+    <footer
+      className={classNames(
+        footerClass[feedFilterOnboardingModalType],
+        'flex fixed responsiveModalBreakpoint:sticky bottom-0 py-3 border-t border-theme-divider-tertiary bg-theme-bg-tertiary',
+      )}
+    >
+      {feedFilterOnboardingModalType === 'introTest2' && (
+        <Button
+          className="ml-4 w-20 text-theme-label-tertiary btn-default"
```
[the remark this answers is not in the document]
ayadotdev: I removed `btn-default `but I had to keep `text-theme-label-tertiary` as removing it causes the change of button text color.

packages/shared/src/components/modals/FeedFiltersModal/FeedFiltersIntroModal.tsx
```diff
@@ -0,0 +1,117 @@
...
+} & Pick<ModalProps, 'onRequestClose'>;
+
+type FeedFiltersIntroModalProps = FeedFiltersProps &
+  ModalFooterProps &
+  ModalProps;
+
+const footerClass = {
+  introTest1: 'justify-center',
+  introTest2: 'justify-between',
+};
```
rebelchris, line 36: Missed it, but this can also move outside the component :)

packages/shared/src/components/modals/FeedFiltersModal/FeedFiltersIntroModalTags.tsx
```diff
@@ -0,0 +1,40 @@
...
+  const randomIndex = Math.floor(Math.random() * widths.length);
+  return widths[randomIndex];
+};
+
+const getWidth = (shouldBeCut: boolean) => {
+  if (shouldBeCut) {
+    return 'flex-1';
+  }
+  return getRandomWidth();
+};
```
rebelchris, line 21: As this isn't reused and relatively small should we not perhaps combine them?

```suggestion
const getWidth = (shouldBeCut: boolean) => {
  if (shouldBeCut) {
    return 'flex-1';
  }
  
  const randomIndex = Math.floor(Math.random() * widths.length);
  return widths[randomIndex];
};
```

packages/shared/src/components/sidebar/Sidebar.tsx
```diff
@@ -156,10 +165,30 @@ export default function Sidebar({
...
+
+  useEffect(() => {
+    if (isFirstSession && isSessionLoaded) {
+      setIsFirstSession(false);
+      if (feedFilterModalOnboarding === 'control') {
+        openFeedFilters();
+      } else {
+        openFeedFilters();
+        setShowIntroModal(true);
+      }
```
rebelchris, line 189:
```suggestion
      openFeedFilters();
      if (feedFilterModalOnboarding !== 'control') {
        setShowIntroModal(true);
      }
```

rebelchris (changes requested):
Almost there! 🏃‍♂️

packages/shared/src/components/sidebar/Sidebar.tsx
```diff
@@ -156,10 +165,30 @@ export default function Sidebar({
...
   const shouldShowDnD = !!process.env.TARGET_BROWSER;
   const { flags } = useContext(FeaturesContext);
   const popularFeedCopy = getFeatureValue(Features.PopularFeedCopy, flags);
   const feedFilterModal = getFeatureValue(Features.FeedFilterModal, flags);
+  const feedFilterModalOnboarding = getFeatureValue(
+    Features.FeedFilterModalOnboarding,
+    flags,
+  );
+  const [isFirstSession, setIsFirstSession, isSessionLoaded] =
+    usePersistentContext(FIRST_TIME_SESSION, true);
```
idoshamun, line 179: Two comments:
* I think this should be in some context rather than in the sidebar. Because other components might need it as well.
* This is not so bulletproof because existing users will also see this dialog and it's not the expected behavior. We want only new users to see it. We have to think about another way to achieve it. It's a global application challenge I think. How to identify a new user? I think you need some Gateway work here as well.

idoshamun (changes requested):
Please see my comments.
Also because it's a big one we need to test carefully manually.

packages/shared/src/components/modals/FeedFiltersModal/FeedFiltersIntroModal.tsx
```diff
@@ -0,0 +1,117 @@
...
+}: ModalFooterProps) => {
+  return (
+    <footer
+      className={classNames(
+        footerClass[feedFilterOnboardingModalType],
+        'flex fixed responsiveModalBreakpoint:sticky bottom-0 py-3 border-t border-theme-divider-tertiary bg-theme-bg-tertiary',
+      )}
+    >
+      {feedFilterOnboardingModalType === 'introTest2' && (
+        <Button
```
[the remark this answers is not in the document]
rebelchris: @idoshamun Do we also want to track impressions for these popup variants in general?

packages/shared/src/components/sidebar/Sidebar.tsx
rebelchris: I'll raise a sub ticket for this to identify first time users.
Think it's a bit out of scope for Aya to create this.

rebelchris: However raising another question, should it not simply show for all users who DON'T have feed filters yet?
Rather than first time visitors?

packages/shared/src/components/modals/FeedFiltersModal/FeedFiltersIntroModal.tsx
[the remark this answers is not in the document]
sshanzel: If that's the case then it is most likely inaccurate. The text theme should be part of the button type you are using. I think you should use `btn-tertiary`

If you have any doubt, kindly check with the design team which type should actually be used.

image

packages/shared/src/components/modals/FeedFiltersModal/FeedFiltersModal.tsx
```diff
@@ -1,19 +1,23 @@
...
+import XIcon from '../../../../icons/x.svg';
+import { Button } from '../../buttons/Button';
 
 type TypeProps = {
   feedFilterModalType: string;
+  showIntroModal: boolean;
+  feedFilterOnboardingModalType: string;
+  actionToOpenFeedFilters: () => unknown;
+  onIntroClose: () => unknown;
+  onCloseFeedFilterModal: () => unknown;
```
sshanzel, line 17: This property shouldn't actually be necessary anymore since the type you combine this with already has the property that handles the closing of the modal. I suggest using that one than creating an additional property.
The property I mean to be utilized is the `onRequestClose` rather than this new one.
If you want the property name to be more intuitive, you can put an alias to the said property on destructuring.
```suggestion
```

packages/shared/src/components/modals/FeedFiltersModal/FeedFiltersIntroModal.tsx
```diff
@@ -0,0 +1,140 @@
...
+          className="ml-4 w-20 text-theme-label-tertiary"
```
sshanzel, line 75: I think it is inaccurate if we have to use a specific label color in our button. The text theme should be part of the button type you are using. I think you should use `btn-tertiary`

If you have any doubt, kindly check with the design team which type should actually be used.

image

```suggestion
          className="ml-4 w-20 btn-tertiary"
```

sshanzel (changes requested):
Wow, you've made amazing progress here. Well done! 🚢 

Just had two minor comments other than what others have pointed out, but things are looking good to me.

Also, a general comment. I think it is great that your variable names are very descriptive, though I think they can be shortened in some cases. Just like as they always say: "As short as possible as long as necessary".

packages/shared/src/components/modals/FeedFiltersModal/FeedFiltersIntroModal.tsx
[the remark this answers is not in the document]
idoshamun: If we have it already for the other variants then yes. We need to be consistent

packages/shared/src/components/sidebar/Sidebar.tsx
idoshamun: Some of the variants here don't make sense for existing users as they will make the app not usable for someone who didn't filter. It might be too strict in my opinion but I'd love to hear what you think?

packages/shared/src/hooks/useDynamicLoadAnimated.ts
```diff
@@ -16,18 +16,21 @@ export const useDynamicLoadedAnimation = ({
...
 
-    setIsAnimated(true);
+    setTimeout(() => {
+      setIsAnimated(true);
+    });
   }, [isLoaded]);
 
   useEffect(() => {
-    if (isAnimated) {
+    if (isAnimated || !isInitialised) {
```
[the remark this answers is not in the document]
rebelchris: @sshanzel We actually don't want it to act on any updates on `isInitialsed` it was a design choice to not auto close the sidebar :)

packages/shared/src/components/modals/FeedFiltersModal/FeedFiltersIntroModal.tsx
```diff
@@ -30,16 +37,58 @@ type FeedFiltersIntroModalProps = FeedFiltersProps &
...
   onRequestClose,
   onOpenFeedFilterModal,
 }: ModalFooterProps) => {
+  const { trackEvent } = useContext(AnalyticsContext);
+
+  useEffect(() => {
+    trackEvent(
+      getAnalyticsEvent(
+        'impression',
+        'skip button',
```
idoshamun, line 63:
```suggestion
        'filters skip button',
```

packages/shared/src/components/sidebar/Sidebar.tsx
```diff
@@ -175,16 +176,15 @@ export default function Sidebar({
     Features.FeedFilterModalOnboarding,
     flags,
   );
+
   const [isFirstSession, setIsFirstSession, isSessionLoaded] =
-    usePersistentContext(FIRST_TIME_SESSION, true);
+    usePersistentContext(FIRST_TIME_SESSION, isFirstVisit);
```
idoshamun, line 181: One edge case that comes into my mind, is the user enters daily.dev for the first time, sees the onboarding but does not finish it. Comes back again and suddenly they will not see it any more right?

packages/shared/src/components/modals/FeedFiltersModal/FeedFiltersIntroModal.tsx
idoshamun, line 63: I think the right value here is `open feed filter button` by looking a the other parameters. Can you confirm?
I also see that in other places (outside the scope of this change), we used the value `my feed button` for indicating a click on the CTA that opens the feed filters. Can you please confirm as well and if so please change wherever necessary

packages/shared/src/components/modals/FeedFiltersModal/FeedFiltersIntroModal.tsx
```diff
@@ -30,16 +37,58 @@ type FeedFiltersIntroModalProps = FeedFiltersProps &
...
+          ? 'Create my feed'
+          : 'Continue',
+      ),
+    );
+  }, []);
+
+  const closeModal = (event: React.MouseEvent) => {
+    getAnalyticsEvent(
+      'click',
+      'skip button',
```
idoshamun, line 75:
```suggestion
      'filters skip button',
```

idoshamun (changes requested):
Please check my comments

packages/shared/src/components/sidebar/Sidebar.tsx
rebelchris: Sorry, my assumption was this was how it was supposed to work.
So we want to keep showing this until they act?

It might become annoying for users if they have to close this modal time over?

rebelchris:
@idoshamun I've refactored the tracking as discussed.
Also included a new impression for the actual feed filters.
This includes a new addition to the `useFeedSettings` hook to keep track of the last known origin (default is `manual`)

Would love to hear what you and @sshanzel think of this approach.

packages/shared/src/components/modals/ResponsiveModal.module.css
```diff
@@ -17,7 +17,8 @@
   & :global(.modal) {
     position: absolute;
     max-width: 30rem;
-    min-height: 100%;
+      max-height: 40rem;
+      min-height: 100%;
```
sshanzel, line 21: Just gotta remove the extra indents, I think.
```suggestion
    max-height: 40rem;
    min-height: 100%;
```

sshanzel (approved):
Just the indent part but nothing blocking. Amazing work on this PR! 🚢

packages/shared/src/components/sidebar/Sidebar.tsx
idoshamun: If they dismissed it, it should be gone forever I agree. I originally meant in cases where they didn't interact at all or didn't finish the process we want to keep showing it until they dismiss manually. Does that make sense?

packages/shared/src/components/MainFeedLayout.tsx
```diff
@@ -222,7 +229,7 @@ export default function MainFeedLayout({
       return (
         <CreateMyFeedButton
           type={myFeedPosition}
-          action={openFeedFilters}
+          action={() => ''}
```
rebelchris, line 232: This doesn't look correct

packages/shared/src/components/filters/FilterMenu.tsx
```diff
@@ -34,7 +34,9 @@ export default function FilterMenu({
     {
       icon: <HashtagIcon className="mr-3 text-xl" />,
```
rebelchris, line 35: This could get upgraded to the sizing spec

rebelchris (changes requested):
@idoshamun Refactored the way we use this first session.
It's only valid if a user has `isFirstSession` set.

Then we set the local version to true, and only once they close or perform an action on the modal will it be set to false.

This way it keeps showing until they interact with it.

idoshamun (changes requested):
The feed filters mechanism feels a bit messy now. It was hard for me to track what's exactly going on and why. I believe we need better encapsulation of some business logic there. If you want we can do a session together to solve everything

rebelchris:
@idoshamun Agreed this is quite a lot of messy code, I even thought of doing it from scratch, but seeing the PR was already nearly accepted decided to leave it.

Would love to perhaps do a quick session and get some high level points.
Happy to take it from there.

PR #959, opened 2022-04-27
refactor: comment box from textarea to div contenteditable

sshanzel:
- Based on our design guideline, we have set a specific max height for the comment modal, with that, scrolling from inside must be handled.
- Have set max-height on the original comment to reply to since we let the textarea part just consume what's available.

- [x] MobileL (420px)
- [x] Tablet (656px)
- [x] Laptop (1020px)

DD-{number} #done


rebelchris:
@sshanzel Nice work buddy, however think we should really put this one through the design team.
I've raised a concern here: https://dailydotdev.slack.com/archives/C02AA18RD1D/p1651144522037189

sshanzel:
@rebelchris thank you for the awesome review! Regarding the fix, I am not actually a fan of JS-solution for its height. I would really much prefer the `div` with `contenteditable` though initially, it was pretty complex and hard to maintain when I worked on the user mention (specifically getting the caret position) so I had to switch to textarea.

If we want to publish to the extension now, in line with the article modal, I think we can settle with this for now, and create a ticket to go back to `contenteditable` but refactor the user mention as well to work with it.

sshanzel:
I mean, I never mentioned the companion. I meant about publishing the extension for article modal as I think we haven't shipped one for it yet.

If this is nothing to rush on, then let me see how we can manage it through contenteditable.

sshanzel:
@rebelchris feel free to take a peek once you have some spare time. Have thoroughly tested the workability of the mentions as it was the primary concern for utilizing `div` with `contenteditable`.

rebelchris (commented):
Looks like you found a great way around this.
Before launching would you mind putting this on Preview and have it tested internally.

This is something we want to be careful with while rolling out as the contenteditable element might have some edge behaviours.

Pleae let me know if available and I'll go through it once more.

sshanzel:
@rebelchris sorry for lacking the info, this should be available in the preview deployment below. Thank you so much for the effort! 🙇 

https://preview.app.daily.dev/

packages/shared/src/components/modals/CommentBox.tsx
```diff
@@ -36,12 +36,14 @@ export interface CommentBoxProps {
...
   onKeyDown: (
-    e: KeyboardEvent<HTMLTextAreaElement>,
-    defaultCallback?: KeyboardEventHandler<HTMLTextAreaElement>,
+    e: KeyboardEvent<HTMLDivElement>,
+    defaultCallback?: KeyboardEventHandler<HTMLDivElement>,
   ) => unknown;
   post: Post;
 }
 
+const cleanupText = (text: string) => text.replaceAll?.('\xa0', ' ') || text;
```
rebelchris, line 45: It might be unclear that `xa0` is empty space. perhaps we should rename cleanupText to state cleanupEmptySpaces?

But this is very optional, happy either way.

rebelchris: Second note, perhaps we can put this in general utils? as we most likely will need it in other places later on.

packages/shared/src/components/modals/CommentBox.tsx
```diff
@@ -91,77 +94,81 @@ function CommentBox({
...
-        selected={selected}
-        query={mentionQuery}
-        onMentionClick={onMentionClick}
-      />
-      <div
-        className="my-2 mx-3 text-theme-status-error typo-caption1"
-        style={{ minHeight: '1rem' }}
-      >
-        {errorMessage && <span role="alert">{errorMessage}</span>}
+        <div
```
rebelchris, line 166: Do we always want to show this empty div?
I see we use it as the margin offset, so it's fine, but might not be best use-case?

rebelchris (approved):
Wow Lee, it's amazing how you simplified this in the contenteditable div.

I see basic support is pretty good, but perhaps you could ask Graham if he sees any accessibility concerns (they should not block, but might raise add-on tasks for later)

Furthermore, one thing I would like to raise UX wise.
When you open a long comment the overflow scroll is still there.
Can we perhaps try and remove the overflow scroll so only the comment box scrolls?

Otherwise see my small notes, but we should be good to go.

packages/shared/src/components/modals/CommentBox.tsx
sshanzel: Totally agree with your suggested name. Will work on it real quick! ⛴️

packages/shared/src/components/modals/CommentBox.tsx
sshanzel: Was thinking about that as well. Probably it was placed before so the UI won't jumpy when it shows? Sounds like not a big issue for me UX-wise. Should we conditionally render this now? What do you think?

sshanzel:
@rebelchris thank you very much for the amazing review!!!

> Can we perhaps try and remove the overflow scroll so only the comment box scrolls?

I think that might produce an unwanted behavior if the user is replying in a very long comment. If we limit the scrolling part on the current user's comment only, there would only be very little space for the users to use (might also become unusable if replying to a comment longer than the image below). What do you think?

image

rebelchris:
@sshanzel Sorry didn't mean overflow, meant the overscroll behaviour.
So when you scroll the modal the feed or page itself also scrolls.

sshanzel:
It seems after trying, it doesn't scroll the outside elements when it is opened and the comment box is the only part where it scrolls. Probably we can sync in Gather or in our upcoming meeting (Friday).

With that, I will be merging a few moments from now, if in case we figured out the issue with scrolling, I think we can create a ticket for it with the inclusion of Graham's feedback if any. 

https://user-images.githubusercontent.com/13744167/169220297-72461346-c4bb-4f02-abcd-50cc150847a5.mov

rebelchris:
@sshanzel for this one, perhaps wait on @idoshamun feedback as well, since it's quite a big one.

sshanzel:
After discussing with Chris, we have decided to ship the new comment box fix now. Thank you! 🚢

idoshamun:
No worries :)
I know it takes me some time to review PRs now.

PR #962, opened 2022-04-29
chore: PR template to add reminder for checking actual mobile devices

sshanzel:
@rebelchris what do you think of this change? Do you think it can add value to our future PRs? If really necessary in some cases, we can probably suggest utilizing an emulator to further check.

With that, it might also be a good idea to create a doc page of what software we suggest for emulating different devices.

image


idoshamun (approved):
Great addition. But please wait for @rebelchris approval

rebelchris (approved):
Nice, great addition indeed 💖

PR #967, opened 2022-05-04
fix: cleanup reading rank state

rebelchris:
## Changes
Add cleanups to the useReadingRank

- The useReadingRank hook had zero cleanups and quite a lot of timeout and event listeners.
- This resulted in memory leaks, thus we clean them up now
- Hopefully this also addressed the flaky tests (yet to be proven)

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-10

### DIFF
--- a/packages/shared/src/hooks/useReadingRank.ts
+++ b/packages/shared/src/hooks/useReadingRank.ts
@@ -1,4 +1,4 @@
-import { useContext, useEffect, useState } from 'react';
+import { useContext, useEffect, useState, useCallback, useRef } from 'react';
 import { useQuery, useQueryClient } from 'react-query';
 import request from 'graphql-request';
 import { isThisISOWeek, isToday } from 'date-fns';
@@ -117,19 +117,29 @@ export default function useReadingRank(
       neverShowRankModal,
     );
 
-  const updateShownProgress = async () => {
+  const unmountRef = useRef(false);
+  const timeoutRef = useRef<number>();
+  const visibilityRef = useRef(null);
+
+  const updateShownProgress = useCallback(() => {
+    if (unmountRef) {
+      return;
+    }
+
+    visibilityRef.current = () => {
+      timeoutRef.current = window.setTimeout(updateShownProgress, 1000);
+    };
+
     if (document.visibilityState === 'hidden') {
-      document.addEventListener(
-        'visibilitychange',
-        () => setTimeout(updateShownProgress, 1000),
-        { once: true },
-      );
+      document.addEventListener('visibilitychange', visibilityRef.current, {
+        once: true,
+      });
     } else if (cachedRank?.rank.currentRank === remoteRank?.rank.currentRank) {
-      await cacheRank();
+      cacheRank();
     } else {
       setLevelUp(true);
     }
-  };
+  }, [cachedRank]);
 
   useEffect(() => {
     if (!disableNewRankPopup) {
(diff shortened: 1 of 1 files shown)


packages/shared/src/hooks/useReadingRank.ts
```diff
@@ -117,30 +117,57 @@ export default function useReadingRank(
...
+
     if (document.visibilityState === 'hidden') {
-      document.addEventListener(
-        'visibilitychange',
-        () => setTimeout(updateShownProgress, 1000),
-        { once: true },
-      );
+      document.addEventListener('visibilitychange', visibilityRef.current, {
+        once: true,
+      });
```
sshanzel, line 136: I just realized, this condition can be met multiple times, so if there is an existing event handler, I think it should get cleaned up as well. If there is none, this does not result in the app crashing, so this should be safe. What do you think?
```suggestion
      document.removeEventListener('visibilitychange', visibilityRef.current);
      document.addEventListener('visibilitychange', visibilityRef.current, {
        once: true,
      });
```

sshanzel (approved):
Nice! Very well done on how you managed the visibility reference. Just had one comment I also forgot earlier in the session.

packages/shared/src/hooks/useReadingRank.ts
rebelchris: Yep 100% agree, we missed that one!

sshanzel:
I see no flaking! Ho! Ho! Ho! 💯

rebelchris:
@idoshamun This one I just wanted to ask if you have any concerns with the changes?

packages/shared/src/hooks/useReadingRank.ts
```diff
@@ -117,30 +117,58 @@ export default function useReadingRank(
       neverShowRankModal,
     );
 
-  const updateShownProgress = async () => {
+  const unmountRef = useRef(false);
+  const timeoutRef = useRef<number>();
+  const visibilityRef = useRef(null);
+
+  const updateShownProgress = useCallback(() => {
+    if (unmountRef) {
```
idoshamun, line 125: I think you originally meant this
```suggestion
    if (unmountRef?.current) {
```

idoshamun (changes requested):
It's a bit hard for me to understand what's happening here.
I think it needs some refactoring or a better explanation.

packages/shared/src/hooks/useReadingRank.ts
```diff
@@ -117,30 +117,58 @@ export default function useReadingRank(
...
-        () => setTimeout(updateShownProgress, 1000),
-        { once: true },
-      );
+      document.removeEventListener('visibilitychange', visibilityRef.current);
+      document.addEventListener('visibilitychange', visibilityRef.current, {
+        once: true,
+      });
     } else if (cachedRank?.rank.currentRank === remoteRank?.rank.currentRank) {
-      await cacheRank();
+      cacheRank();
```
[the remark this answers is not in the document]
rebelchris: We moved to a useCallback, but I think we missed this as it shouldn't have been removed.

rebelchris:
@idoshamun I'm not really sure how we can make this beast of a function more readable.
It's tricky to follow in the first place.

A bit more context.
We open a lot of event listeners and timeouts in this hook.
Before they where never removed, this is not fixed by setting a cleanup useEffect function.

Second to that, there was a possibility of the event listeners firing a function after unmount.
For this we introduce the `unmountRef` variable.
It sets to true on unmount, and for each function we evaluate if that's the case it should not execute the function to prevent setting state that no longer exists.

Let me know if this makes sense, or perhaps a DR explaining the hook would make more sense?
I tried to evaluate how we can make it more clear, but so far it seems hard to refactor as a whole.

idoshamun (approved):

PR #968, opened 2022-05-04
feat: toast component

sshanzel:
- A newly introduced toast component for notifications
- Auto dismisses by default after 5 seconds (default)
- Can immediately close the notification or set shorter time
- Can include undo button to revert recently made action
- Migrated existing implementations to utilize the new notification component
- For accessibility concerns, we apply `aria-live` on dynamic regions that got affected after the action 
~~**Note:** until this PR (https://github.com/dailydotdev/daily-api/pull/934/files) is merged, undoing the report and hiding of posts won't be working. For the other actions, undo should work.~~

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

DD-{number} #done


packages/shared/src/components/notifications/Toast.tsx
```diff
@@ -0,0 +1,153 @@
...
+  const [timer, setTimer] = useState(0);
+  const { data: toast } = useQuery<ToastNotification>(TOAST_NOTIF_KEY);
+  const setToast = (data: ToastNotification) =>
+    client.setQueryData(TOAST_NOTIF_KEY, data);
+
+  useEffect(() => {
+    if (toast && timer === 0 && intervalId) {
+      window.clearInterval(intervalId);
+      setIntervalId(null);
+      setTimeout(() => setToast(null), IN_OUT_ANIMATION);
```
rebelchris, line 52: We probably also want to clear this timeout on cleanup.

rebelchris (changes requested):
Nice setup Lee!

Would you mind putting this one also on a preview domain, would love to run through some things.
Overall, first impressions are really good, but like to validate some things before approving.

Also surprised you didn't need a custom context for this :)

packages/shared/src/components/notifications/Toast.tsx
sshanzel: Indeed, this has to be cleaned up! 🔭

sshanzel:
@rebelchris thank you for the review! I have now applied a cleanup for the timeout. This PR should be available in the preview deployment below:

packages/shared/src/components/Settings.tsx
```diff
@@ -109,11 +115,11 @@ export default function Settings({
       </Section>
       <Section>
         <SectionTitle>Preferences</SectionTitle>
-        <div className="flex flex-col items-start pl-1.5 -my-0.5">
+        <SectionContent>
           <Switch
             inputId="hide-read-switch"
             name="hide-read"
-            className="my-3"
+            className="my-3 big"
```
rebelchris, line 122: What does this big class do, I don't see any reference of that in our styles?

packages/shared/src/components/notifications/Toast.tsx
```diff
@@ -0,0 +1,171 @@
...
+const Content = classed('div', 'relative flex flex-row items-center ml-2');
+const Message = classed(
+  'div',
+  'flex-1 mr-2 typo-subhead text-theme-label-invert',
+);
+const Progress = classed(
+  'span',
+  styles.toastProgress,
+  'absolute -bottom-2 h-1 ease-in-out bg-theme-status-cabbage rounded-full',
+);
```
rebelchris, line 38: Up to you, but perhaps we should move these to a common or util file int he notifications directory? (Just cleaner I think)

non-blocking

packages/shared/src/components/notifications/Toast.tsx
```diff
@@ -0,0 +1,171 @@
...
+const INTERVAL_COUNT = 10;
+const IN_OUT_ANIMATION = 140;
+const TEMPORARY_ID = 1;
```
rebelchris, line 41: What's this used for?

packages/shared/src/components/notifications/Toast.tsx
```diff
@@ -0,0 +1,171 @@
...
+      );
+    }
+  }, [timer, toast, intervalId]);
+
+  useEffect(
+    () => () => {
+      window.clearTimeout(timeout.current);
+    },
+    [],
+  );
```
rebelchris, line 72: Is this not redundant if we already have it in the useEffect below?

rebelchris (changes requested):
Wow Lee, really well done on this task.
Also love the fact that you took accessibility into account for this one.

Perhaps also ask Graham to double check the behaviour.

Would you mind fixing (or adding a new task) to also add the notifications to the reading history as we are currently missing them there.

As mentioned having issues with the undo, but raised seperate.
I can't seem to test the notifications with the settings turned off, not sure if this is not available on the preview?

some notes you might want to check with product:
- Should we change the "invite people" in the sidebar to this toast?
- Copy link in the article detail doesn't show toast (see screenshot)
- Same goes for in article modal
Screenshot 2022-05-18 at 10 32 08
 - Please check if we have any other implementations of "Copy link"

packages/extension/src/newtab/App.tsx
```diff
@@ -78,10 +79,13 @@ function InternalApp({
     }
   }, [routeChangedCallbackRef]);
 
+  const { dismissToast } = useToastNotification();
+
   const onPageChanged = (page: string): void => {
     // eslint-disable-next-line no-param-reassign
     pageRef.current = page;
     routeChangedCallbackRef.current();
+    dismissToast();
```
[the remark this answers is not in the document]
sshanzel: It is part of the behavior to automatically dismiss notifications when navigating to a different page.

image

packages/shared/src/components/Settings.tsx
sshanzel: I think that is something related to the switch - not quite sure though but all of the switches inside the Settings have that class.

packages/shared/src/components/notifications/Toast.tsx
sshanzel: This is the only cleanup for the `timeout`. The `useEffect` below is used for the `interval`.

packages/shared/src/components/notifications/Toast.tsx
sshanzel: Great idea. Let me move these ones! Thank you!

sshanzel:
@rebelchris thank you for the awesome review! I have now migrated all the copy links implementation to utilize the toast notification.
Not sure what could be causing you to not be able to utilize the settings for notifications. Was there anything in the logs? I'll be more than happy to check it with you tomorrow if you have some spare time.

@idoshamun feel free to review once you have some spare time! 🔍

packages/shared/src/components/Settings.tsx
rebelchris: I don't see any use for it, could you see if we can remove all of these?
Don't want to have random classes that don't do anything if that makes sense

sshanzel: Totally agree. I have instead created a common component since all of the switches are the same on some properties then removed the unnecessary class. Thank you! 🚢

rebelchris (approved):
Love this! Way to go Lee 💖
Only concern from testing is the block -> undo doesn't unhide the article now.
We can raise that as a separate concern from my perspective.

packages/shared/src/components/notifications/Toast.tsx
```diff
@@ -0,0 +1,164 @@
...
+    const handler = () => dismissToast();
+    window.addEventListener('scroll', handler);
+    router.events.on('routeChangeStart', handler);
+
+    // eslint-disable-next-line consistent-return
+    return () => {
+      window.removeEventListener('scroll', handler);
+      router.events.off('routeChangeStart', handler);
+    };
+  }, [dismissToast]);
```
idoshamun, line 127: This logic seems a bit too to maintain in one component.
We should consider refactoring to functions or hooks. It's hard to understand what these useEffects are all about without carefully following the code.

packages/shared/src/components/ShareNewCommentPopup.tsx
```diff
@@ -97,7 +97,7 @@ export default function ShareNewCommentPopup({
         <Button
           className="btn-primary"
           buttonSize="small"
-          onClick={copyLink}
+          onClick={() => copyLink()}
```
idoshamun, line 100: Why can't we keep the old implementation?

packages/shared/src/components/ShareMobile.tsx
```diff
@@ -16,7 +16,7 @@ export function ShareMobile({ share }: Props): ReactElement {
     <WidgetContainer className="flex laptop:hidden flex-col gap-2 items-start py-3 px-1">
       <Button
         buttonSize="small"
-        onClick={copyLink}
+        onClick={() => copyLink()}
```
idoshamun, line 19: Why can't we keep the old implementation?

packages/shared/src/components/ShareBar.tsx
```diff
@@ -44,19 +42,9 @@ export default function ShareBar({ post }: { post: Post }): ReactElement {
...
-              styles.copied,
-            )}
-          >
-            Copied!
-          </div>
-        )}
         <SimpleTooltip content="Copy link">
           <ShareButton
-            onClick={copyLink}
+            onClick={() => copyLink()}
```
idoshamun, line 47: Why can't we keep the old implementation?

packages/shared/src/components/PostOptionsReadingHistoryMenu.tsx
```diff
@@ -144,7 +139,7 @@ export default function PostOptionsReadingHistoryMenu({
             {getBookmarkIconAndMenuText(post?.bookmarked)}
           </span>
         </Item>
-        <Item className="typo-callout" onClick={onShareOrCopyLink}>
+        <Item className="typo-callout" onClick={() => onShareOrCopyLink()}>
```
idoshamun, line 142: Why can't we keep the old implementation?

packages/shared/src/hooks/useToastNotification.ts
```diff
@@ -0,0 +1,47 @@
...
+  const { data: toast } = useQuery<ToastNotification>(TOAST_NOTIF_KEY);
+  const setToastNotification = (data: ToastNotification) =>
+    client.setQueryData(TOAST_NOTIF_KEY, data);
+
+  const displayToast = (
+    message: string,
+    { timer = 5000, ...props }: NotifyOptionalProps = {},
+  ) => setToastNotification({ message, timer, ...props });
+
+  return {
```
idoshamun, line 42: You might want to consider `useMemo` here because you generate a new object on every render

idoshamun (changes requested):
Wow! I love the behavior of this toast. It also significantly reduced the spaghetti code of showing a notification in the app.
Beside the comments, we also need to tweak the user experience according to the discussion below
image

But that's a fine job right there! Way to go 💪

packages/shared/src/components/ShareNewCommentPopup.tsx
sshanzel: The `copyLink` function now accepts parameters, and the `onClick` natively passes an `event` object, so it is giving an incompatible error.

packages/shared/src/components/notifications/Toast.tsx
sshanzel: The initialization of the toast notification is actually dependent to be a `useEffect` since we will have multiple instances of `useToastNotification`.

There are two options for us to work with the Notifications.
One is the current implementation.
The second is by rendering a toast component on each and every component where it will need the notification instead of the `useToastNotification` hook - from there only we can turn these ones into a function.

Perhaps I should add a comment on what the hooks actually do. What do you think?

idoshamun: I love the high level implementation that you followed which is to have a single component and use a hook to trigger it.
My concern was very tactical, the fact that we have a lot of business logic inside the components makes it hard to read through. I suggest some refactoring around it. Just to make sure a new developer comes in and is able to understand what's going on

packages/shared/src/hooks/useTimedAnimation.ts
```diff
@@ -0,0 +1,87 @@
...
+  animationDuration: number;
+  onAnimationEnd?: () => void;
+  onAnimationStop?: () => void;
+}
+
+const TEMPORARY_ID = 1;
+const INTERVAL_COUNT = 10;
+const IN_OUT_ANIMATION = 140;
+
+export const useTimedAnimation = ({
```
idoshamun, line 21: I see that we use a timer to indicate when the animation is done, and an interval to update the progress.
May I suggest using only an interval and when the animation is done trigger it from there?
This will make it less complicated and more synced I think

packages/shared/src/hooks/useTimedAnimation.ts
```diff
@@ -0,0 +1,87 @@
...
+
+interface UseTimedAnimationProps {
+  autoEndAnimation: boolean;
+  animationDuration: number;
+  onAnimationEnd?: () => void;
+  onAnimationStop?: () => void;
+}
+
+const TEMPORARY_ID = 1;
+const INTERVAL_COUNT = 10;
```
idoshamun, line 18: `INTERVAL_COUNT` name is not very explicit in this case.
I would suggest `PROGRESS_INTERVAL`. As it indicates how often we should trigger the progress update right?

packages/shared/src/hooks/useTimedAnimation.ts
```diff
@@ -0,0 +1,87 @@
...
+}
+
+interface UseTimedAnimationProps {
+  autoEndAnimation: boolean;
+  animationDuration: number;
+  onAnimationEnd?: () => void;
+  onAnimationStop?: () => void;
+}
+
+const TEMPORARY_ID = 1;
```
idoshamun, line 17: Why do we need this instead of using null?

packages/shared/src/hooks/useTimedAnimation.ts
```diff
@@ -0,0 +1,87 @@
...
+  const timeout = useRef<number>();
+  const [timer, setTimer] = useState(0);
+  const [intervalId, setIntervalId] = useState<number>(null);
+
+  const endAnimation = useCallback(() => {
+    if (!isNullOrUndefined(animationDuration)) {
+      return;
+    }
+
+    window.clearInterval(intervalId);
```
idoshamun, line 36: We don't set the `intervalId` to null here. I suggest a function that both clears the interval and sets the variable to null

idoshamun (changes requested):
Looks much better. Still a few comments on the new hook

packages/shared/src/hooks/useTimedAnimation.ts
sshanzel: Yes, that was intentional so we can allow the out animation to process first before dismounting. It is then set to null in the line below:
https://github.com/dailydotdev/apps/blob/7efbf1e531f4484524f980d3730d8a42c733a900/packages/shared/src/hooks/useTimedAnimation.ts#L45

packages/shared/src/hooks/useTimedAnimation.ts
idoshamun: Ok I see now. In this case I'd just rename this constant as it's not clear that this is its purpose

packages/shared/src/hooks/useTimedAnimation.ts
sshanzel: If we will be removing the `timer` we won't have any reference of when to stop the interval itself as intervals run indefinitely until explicitly stopped.

sshanzel: If we will be removing the timer we won't have any reference of when to stop the interval itself as intervals run indefinitely until explicitly stopped.

idoshamun: But you use `setTimer` to keep track of the progress, don't you?

sshanzel: Yes. Since we need the re-render so the component would know the change of value to display.

sshanzel: Ah. And replace the interval to do it. Let me see how we can achieve it since we set value of 0 for unending animations. I'll check how it would turn out if we do -1.

sshanzel:
@idoshamun instead of relying on the use effects, I have converted things to what you have initially suggested, to explicitly call the animations for easier maintenance.

packages/shared/src/hooks/useTimedAnimation.ts
```diff
@@ -1,80 +1,81 @@
...
+  const startAnimation = (duration: number) => {
+    if (isNullOrUndefined(duration) || duration <= 0) {
       return;
     }
 
-    if (animationDuration <= 0) {
+    setTimer(duration);
+
+    if (!autoEndAnimation) {
+      clearInterval();
```
idoshamun, line 55: This can go outside the `if` because in both cases you apply this function

packages/shared/src/hooks/useTimedAnimation.ts
```diff
@@ -1,80 +1,81 @@
...
+        ),
+      PROGRESS_INTERVAL,
     );
-  }, [animationDuration]);
+  };
+
+  useEffect(() => {
+    // when the timer ends we need to do cleanups
+    // we delay the callback execution so we can let the slide out animation finish
+    if (timer === 0 && interval?.current) {
```
idoshamun, line 73: To make sure we don't introduce infinite interval

```suggestion
    if (timer <= 0 && interval?.current) {
```

packages/shared/src/hooks/useTimedAnimation.ts
```diff
@@ -1,80 +1,81 @@
...
+      clearInterval();
+      window.clearTimeout(timeout.current);
+      timeout.current = window.setTimeout(onAnimationEnd, outAnimationDuration);
```
idoshamun, line 76: This no longer needed, am I right?

idoshamun (changes requested):
A few more things. Looks much better 😬

packages/shared/src/hooks/useTimedAnimation.ts
sshanzel: This one should still be needed, for out animation, so we won't clear things out until it is finished.

sshanzel:
@idoshamun after testing the preview deployment, [undoing something in the feed doesn't feel like it undid something](https://dailydotdev.atlassian.net/browse/DD-629) since we just remove data from the cache. I tried to recreate the data to be programmatically put back into the list but it doesn't work in preview deployment. So to make things simple, I now invalidate the query instead of manually reverting back the post.

The total changes can be found in this single commit (https://github.com/dailydotdev/apps/pull/968/commits/b8f352fb1cb0c684170868e92cc3483d2319bad6) - let me know if have concerns about it. Thank you!

idoshamun:
@sshanzel but this means the feed will completely reload, isn't it?

sshanzel:
@idoshamun yes. So at least the changes can be seen when undo is clicked than no activity. I'm thinking we might get flooded by bug reports.

PR #970, opened 2022-05-06
feat: add community links env limit

rebelchris:
- Add env variable for community links per day

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-39 #done


sshanzel:
I think we should add this on the API side and the validation will be part of the mutation. What do you think?

rebelchris:
Thought about that as well, but we need the value in the frontend ui, so eitherway we'll need to share it.

idoshamun:
This should be provided by the API in my opinion. Otherwise we would have to release a version to change it.

rebelchris:
@idoshamun Do you have any suggestions on how we can push this back to the apps, as we need the value to display in the UI.

idoshamun:
I think we're aligned on this looking at the DR? just making sure @rebelchris

PR #971, opened 2022-05-09
feat: recommend article modal

rebelchris:
## Changes
Introducing the recommend an article modal.

## Blocking
This task has some blocking elements
- API submission -> Throwing error ✅
- API missing for open of the modal (limit and user daily rate)
- Link request form missing
- Test cases missing because of API ✅

- People with no access can request access (external)
- People with access can submit a link
- We evaluate the link based on multiple criteria

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-40


rebelchris:
@sshanzel Here is my initial frontend, i've added a blocking session so we can see what is needed before we can close this.
Would love a initial review/test from your side as well to see if it works with the API changes.

packages/shared/src/components/modals/RecommendAnArticle.tsx
```diff
@@ -0,0 +1,182 @@
...
+            disabled={!isEnabled}
+            hint={urlHint}
+            valid={!urlHint}
+            valueChanged={onUrlChanged}
+          />
+          {isSubmitted && (
+            <>
+              <div className="flex items-center mt-4 mb-6 h-8 font-bold rounded-10 w-fit bg-theme-status-success text-theme-label-invert padding px-[0.9375rem]">
+                Request sent
+              </div>
```
sshanzel, line 136: I noticed in the Sketch that this is actually a `Button` component though and I really like that you made it a static element instead since there is nothing actionable about it.

I think we should raise it to the Design team for better UX as the users might think it is a button to click but nothing will happen.

packages/shared/src/components/modals/RecommendAnArticle.tsx
```diff
@@ -0,0 +1,182 @@
...
+            </>
+          )}
+
+          {isEnabled ? (
+            <Button
+              className="mt-4 btn-primary "
+              buttonSize="small"
+              type="submit"
+              aria-label="Submit article"
+              disabled={!isEnabled || !enableSubmission}
```
sshanzel, line 149: Since it is conditionally being rendered, I think the first condition can be omitted.
```suggestion
              disabled={!enableSubmission}
```

packages/shared/src/components/modals/RecommendAnArticle.tsx
```diff
@@ -0,0 +1,182 @@
...
+            >
+              <span className={isValidating && 'invisible'}>
+                Submit article
+              </span>
+              {isValidating && (
+                <Loader className="absolute inset-x-0 top-1 mx-auto" />
+              )}
+            </Button>
```
sshanzel, line 157: I think we should make this built-in the `Button` component, it looks like this will be used on many occasions. What do you think? Also not sure if this actually part of the design guideline already and we just haven't had the chance to update the implementation.

packages/shared/src/components/modals/RecommendAnArticle.tsx
```diff
@@ -0,0 +1,182 @@
...
+              Request access
+            </Button>
+          )}
+        </form>
+      </section>
+      {existingArticle && (
+        <div>
+          <div className="px-10 pb-3.5 border-b border-theme-divider-tertiary">
+            <p className="font-bold typo-callout">Article exists</p>
+          </div>
```
sshanzel, line 176: Not sure if my suggestion of removing the container is actually a good idea, though I think `h4` would probably be more fitting semantically. What do you think? Also, just wanted to check if the additional empty `div` can be removed or not?
```suggestion
        <div>
          <h4 className="px-10 pb-3.5 border-b border-theme-divider-tertiary font-bold typo-callout">Article exists</h4>
```

packages/shared/src/components/sidebar/Sidebar.tsx
```diff
@@ -156,10 +162,15 @@ export default function Sidebar({
     optOutWeeklyGoal,
   } = useContext(SettingsContext);
   const [showSettings, setShowSettings] = useState(false);
+  const [showRecommendAnArticle, setShowRecommendAnArticle] = useState(false);
```
sshanzel, line 165: This is just a general comment on the term that we're gonna be using. I think it would be better to use `suggest` than `recommend` as the latter might get used in the future like in a "Recommendation" feature where we recommend articles to read. I feel like the current term is bringing a different meaning (actually got a bit confused at first). Also that we use `suggest` in our backend terms.

```suggestion
  const [showSuggestArticle, setShowSuggestArticle] = useState(false);
```

sshanzel (commented):
Just had very minor, non-blocking comments but I think you have built the foundation very well. Amazing work as always!!! 🚢

packages/shared/src/components/modals/RecommendAnArticle.tsx
rebelchris: Very good point, raising it now :)

packages/shared/src/components/modals/RecommendAnArticle.tsx
rebelchris: Very good point, let me refactor

packages/shared/src/components/modals/RecommendAnArticle.tsx
rebelchris: Good point, I'll look if I can remove the extra one, and agree on the h4 semanticly seen.

packages/shared/src/components/sidebar/Sidebar.tsx
rebelchris: Not sure, the whole system is called recommend an article everywhere.
Perhaps we should raise it with product/graham.

sshanzel: I agree. We should coordinate this with the product team.

packages/shared/src/components/modals/SubmitArticle.spec.tsx
```diff
@@ -0,0 +1,164 @@
...
+  mockGraphQL({
+    request: {
+      query: SUBMIT_ARTICLE_MUTATION,
+      variables: { url: 'http://blog.daily.dev/blog/article-1' },
+    },
+    result: {
+      data: {
+        submitArticle: null,
+      },
+      errors: [
```
idoshamun, line 121: I suggested changing this behavior in the API PR. GraphQL errors are not type-safe and thus I'm not a fan of using them heavily.

packages/shared/src/components/modals/SubmitArticle.tsx
```diff
@@ -0,0 +1,211 @@
...
+      setEnableSubmission(submitFormRef.current.checkValidity());
+    }
+  };
+
+  const onSubmit = async (
+    event: React.FormEvent<HTMLFormElement>,
+  ): Promise<void> => {
+    event.preventDefault();
+
+    trackEvent({
```
idoshamun, line 75: Let's add the actual URL as extra please

packages/shared/src/components/modals/SubmitArticle.tsx
```diff
@@ -0,0 +1,211 @@
...
+  const { mutateAsync: submitArticle } = useMutation<unknown, unknown, string>(
+    (articleUrl: string) =>
+      request(`${apiUrl}/graphql`, SUBMIT_ARTICLE_MUTATION, {
+        url: articleUrl,
+      }),
+  );
+
+  const submitArticleFailEvent = (reason: string): void => {
+    trackEvent({
+      event_name: 'submit article failed',
```
idoshamun, line 51:
```suggestion
      event_name: 'submit article fail',
```

packages/shared/src/components/modals/SubmitArticle.tsx
```diff
@@ -0,0 +1,211 @@
...
+    try {
+      await submitArticle(data.articleUrl);
+      setIsSubmitted(true);
+      trackEvent({ event_name: 'submit article succeed' });
+    } catch (err) {
+      const error = JSON.parse(JSON.stringify(err));
+      try {
+        const errorMessage = JSON.parse(error?.response?.errors[0]?.message);
+        setExistingArticle(errorMessage);
+        submitArticleFailEvent('Article exist already');
```
idoshamun, line 98:
```suggestion
        submitArticleFailEvent('Article exists already');
```

packages/shared/src/graphql/submitArticle.ts
```diff
@@ -0,0 +1,9 @@
+import { gql } from 'graphql-request';
+
+export const SUBMIT_ARTICLE_MUTATION = gql`
+  mutation SubmitArticle($url: String!) {
```
idoshamun, line 4: I asked to change the return type of this mutation in the API PR

packages/shared/src/components/sidebar/Sidebar.tsx
```diff
@@ -177,6 +193,14 @@ export default function Sidebar({
     toggleSidebarExpanded();
   };
 
+  const trackAndShowSubmitArticle = () => {
+    trackEvent({
```
idoshamun, line 197: Is it possible to add here a flag that indicates whether the user has access to the feature or not?
If not we can find a more appropriate place for it.
I'd like to separate between conversion rates of people who have access to those who don't

idoshamun (changes requested):
Mainly tracking concerns, other than that it looks fabulous 🧙‍♂️

packages/shared/src/components/modals/SubmitArticle.tsx
rebelchris: Just checking, you wanted to track every click right, or only ones where the URL is definitely set?

packages/shared/src/components/sidebar/Sidebar.tsx
rebelchris: If added an extra: { has_access: boolean } is that ok?

packages/shared/src/components/modals/SubmitArticle.spec.tsx
rebelchris: This is changed to the API

packages/shared/src/components/modals/SubmitArticle.tsx
rebelchris: You can see the change proposed now :)

packages/shared/src/components/sidebar/Sidebar.tsx
idoshamun: Yes. In the extra field

packages/shared/src/components/modals/SubmitArticle.tsx
```diff
@@ -0,0 +1,224 @@
...
+
+    setIsValidating(true);
+
+    if (!data.articleUrl) {
+      setUrlHint('Please submit a valid URL');
+      setIsValidating(false);
+      return;
+    }
+
+    const defaultErrorMessage = 'Something went wrong, try again';
```
idoshamun, line 96: This can go outside as it's a constant

packages/shared/src/components/modals/SubmitArticle.tsx
```diff
@@ -0,0 +1,224 @@
...
+        tag="a"
+        href={requestAccessSubmitArticle}
+        target="_blank"
+        rel="noopener"
+        aria-label="Request access"
+        className="mt-4 w-fit btn-secondary"
+      >
+        Request access
+      </Button>
+    );
```
idoshamun, line 164: Not sure I want this part anymore from a product perspective. I'll talk to Tsahi about it

idoshamun (changes requested):
Looks very good! Almost ready

Can we add a flag to decide if we should submit article button even?
This way we can rollout the version without actually releasing the feature.

idoshamun:
We can also utilize the existing flags to achieve what I mentioned above instead of creating a new one

packages/shared/src/components/modals/SubmitArticle.tsx
rebelchris: This is gone :)

rebelchris:
@idoshamun Decided to add a new flag.
The existing ones are controlled by multiple variables so it's hard to overwrite otherwise.

With this new flag we can safely rollout and give permissions to team members first :).

kirillkurko (approved):

packages/shared/src/components/sidebar/Sidebar.tsx
```diff
@@ -309,6 +297,17 @@ export default function Sidebar({
       active: showSettings,
     },
   ];
+  if (submitArticleOnOn) {
```
idoshamun, line 300: Feels like a typo

```suggestion
  if (submitArticleOn) {
```

idoshamun (approved):
Looks great please don't merge yet though :)

PR #972, opened 2022-05-10
Wt 45 scout badge in comment reply section

ayadotdev:
Adds a scout badge to comment and reply sections (hardcoded).

Screenshot 2022-05-10 at 09 04 40

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-45 #done


packages/shared/src/components/comments/CommentBadge.tsx
```diff
@@ -0,0 +1,11 @@
+import React, { ReactElement } from 'react';
+import ScoutIcon from '../../../icons/filled/scout.svg';
+
+export default function CommentBadge(): ReactElement {
```
rebelchris, line 4: I would maybe call it ScoutBadge as it's more explicit as to what it is.

rebelchris (changes requested):
One small remark in naming, otherwise fine.
Please do note somewhere we need to make this dynamic at one stage. (Perhaps in the WT-4 PR)

packages/shared/src/components/comments/ScoutBadge.tsx
```diff
@@ -0,0 +1,11 @@
+import React, { ReactElement } from 'react';
+import ScoutIcon from '../../../icons/filled/scout.svg';
+
+export default function ScoutBadge(): ReactElement {
+  return (
+    <span className="flex items-center ml-2 text-theme-color-bun typo-footnote">
+      <ScoutIcon className="mr-0.5 text-base icon" />
```
sshanzel, line 7: I think it should be `mr-1`.
image

I don't think you should be using font size for the image size. Also, the `icon` class shouldn't be necessary. I don't think we have a specific selector in our base CSS for this.
```suggestion
      <ScoutIcon className="mr-1 w-3 h-3" />
```

sshanzel (changes requested):
Just had a minor comment regarding the classes of the icon. But overall, well done on the new component!

sshanzel (approved):
LGTM! Well done! 🚢

rebelchris (approved):
Well done 👏

PR #974, opened 2022-05-12
feat: add scout and author badges to user profile

rebelchris:
- Introduce the author and scout badges to the user profile
- Fallback on source image

- [] Have you done sanity checks in the extension? (n/a)

- [x] MobileL (420px)
- [x] Tablet (656px)
- [x] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-42 #done


packages/shared/src/components/profile/PostsSection.tsx
```diff
@@ -53,6 +56,43 @@ const PostStat = classed(
...
+const iconImage = (post: Post) => {
+  if (post.isAuthor || post.isScout) {
+    return (
+      <div
+        className={classNames(
+          iconImageClass,
+          'absolute flex items-center justify-center',
+        )}
+      >
+        {post.isAuthor ? (
```
sshanzel, line 71: I think we can just use the `authorId` property to compare right? Since if the user did scout an article, its `authorId` will either be if it was anyone's article or it will be empty?

rebelchris: This is related to which users profile you are viewing, that's why I had to introduce the two selects as well.
Happy to chat on gather so we can close these?

sshanzel (approved):
LGTM! 🚢

PR #975, opened 2022-05-15
fix: tooltip's unnecessary aria-label

sshanzel:
- Related concern: https://dailydotdev.slack.com/archives/C01L0QXQJNB/p1652520139195169
- For tooltips that are not of type string - the `aria-label` has a value of `false` due to the condition - we now set a condition to only add the property when necessary.

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

DD-{number} #done

### DIFF
--- a/packages/shared/src/components/tooltips/SimpleTooltip.tsx
+++ b/packages/shared/src/components/tooltips/SimpleTooltip.tsx
@@ -23,14 +23,17 @@ export function SimpleTooltip({
    * Read more on the PR: https://github.com/dailydotdev/apps/pull/713
    */
   let shouldShow = true;
-  const component = useMemo(
-    () =>
-      React.cloneElement(children, {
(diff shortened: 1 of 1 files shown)


rebelchris (approved):
Love this fix!

Small note though, I'd prefer it if we can log these in the future and not jump on them on the go.
It's more a policy we are trying to adapt with the cycles.

sshanzel:
> Love this fix!
> 
> Small note though, I'd prefer it if we can log these in the future and not jump on them on the go. It's more a policy we are trying to adapt with the cycles.

Oh, cool. I will log it in the future. Though I think it would be great to discuss these cases in our web meeting since this was an out-of-work hours thing.

PR #976, opened 2022-05-15
feat: discussion in companion

sshanzel:

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

DD-{number} #done

### DIFF
--- a/packages/extension/src/companion/CompanionContent.tsx
+++ b/packages/extension/src/companion/CompanionContent.tsx
@@ -1,8 +1,6 @@
-import React, { ReactElement, useState } from 'react';
+import React, { ReactElement, useContext, useState } from 'react';
 import LogoIcon from '@dailydotdev/shared/src/svg/LogoIcon';
 import CopyIcon from '@dailydotdev/shared/icons/copy.svg';
-import { apiUrl } from '@dailydotdev/shared/src/lib/config';
-import { useInfiniteQuery, useQueryClient } from 'react-query';
 import {
   HotLabel,
   TLDRText,
@@ -10,22 +8,19 @@ import {
 import '@dailydotdev/shared/src/styles/globals.css';
 import SimpleTooltip from '@dailydotdev/shared/src/components/tooltips/SimpleTooltip';
 import { PostBootData } from '@dailydotdev/shared/src/lib/boot';
-import UpvotedPopupModal from '@dailydotdev/shared/src/components/modals/UpvotedPopupModal';
-import { POST_UPVOTES_BY_ID_QUERY } from '@dailydotdev/shared/src/graphql/posts';
-import {
-  DEFAULT_UPVOTES_PER_PAGE,
-  UpvotesData,
-} from '@dailydotdev/shared/src/graphql/common';
-import { ClickableText } from '@dailydotdev/shared/src/components/buttons/ClickableText';
 import { Button } from '@dailydotdev/shared/src/components/buttons/Button';
 import { useCopyLink } from '@dailydotdev/shared/src/hooks/useCopyLink';
+import { usePostComment } from '@dailydotdev/shared/src/hooks/usePostComment';
 import classNames from 'classnames';
 import useNotification from '@dailydotdev/shared/src/hooks/useNotification';
 import { CardNotification } from '@dailydotdev/shared/src/components/cards/Card';
+import { PostComments } from '@dailydotdev/shared/src/components/post/PostComments';
+import { NewComment } from '@dailydotdev/shared/src/components/post/NewComment';
+import AuthContext from '@dailydotdev/shared/src/contexts/AuthContext';
+import NewCommentModal from '@dailydotdev/shared/src/components/modals/NewCommentModal';
+import { CompanionDiscussion } from './CompanionDiscussion';
+import { useBackgroundRequest } from './useBackgroundRequest';
 import { getCompanionWrapper } from './common';
-import { companionRequest } from './companionRequest';
-import { useRawBackgroundRequest } from './useRawBackgroundRequest';
-import { useBackgroundPaginatedRequest } from './useBackgroundPaginatedRequest';
 
 type CompanionContentProps = {
   post: PostBootData;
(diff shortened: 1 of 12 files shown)


packages/shared/src/components/comments/CommentActionButtons.tsx
```diff
@@ -49,9 +50,11 @@ export default function CommentActionButtons({
     setNumUpvotes(comment.numUpvotes);
   }, [comment]);
 
+  const { companionRequest } = useCompanionProtocol();
+  const requestMethod = companionRequest || request;
```
rebelchris, line 54: Should we not opt to simply pass this as we do with other companion affected request methods?

rebelchris: I added this comment before seeing how you use the companion protocol, but see my remarks in the main comment. :)

sshanzel: Was also thinking about that too, though I realized I had to drill quite a bit of props to get where the component requires it. So I thought of using this. Thank you!

packages/extension/src/companion/CompanionContent.tsx
```diff
@@ -110,37 +126,45 @@ export default function CompanionContent({
...
-            target="_parent"
-            className="hover:underline"
-          >
-            {post?.numComments.toLocaleString()}
-            {` Comment${post?.numComments === 1 ? '' : 's'}`}
-          </a>
-        )}
-      </div>
-      {isUpvotesOpen && (
+      <CompanionDiscussion
```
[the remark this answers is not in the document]
sshanzel: I agree there's a lot of clutter now. Let's see how I can refactor this as the discussion's absolute position requires to be relevant to the `CompanionContent`

packages/shared/src/components/comments/CommentActionButtons.tsx
sshanzel: > Like the approach you took on the new hook, although we might have to try this out, it might be super heavy on refocus and add a lot of state that we don't actively use. (Meaning it will never change once you are inside the companion, so why do we use react query for that part?)

I was actually thinking of using Context instead of `useQuery`, thinking it would be quite light still but since you mentioned it, I think we can also go for the Context. 
The only concern I had was we will have to create another ContextProvider since I don't see any of our existing providers where this will fit. And that new context will only have this.
We can also probably declare a static variable just to check if it is the extension then we will use the companion request.
What do you think?

rebelchris: Hmm yeah my initial thoughts would be context, but like you mention it would be very small.
Very perhaps we could use it in the bootContext as this now also has a explicit call for bootquery for companion?
Perhaps we can even hook into that and set some setting?

But as mentioned a static variable for the companion would also work for me, for this case.

sshanzel: Agreed. Let's go for the static variable then and see how it works out.

sshanzel: Now that I worked on it again, I remembered why I was forced to not use static variable - we will have to install `webextension-polyfill-ts` shared package and it is throwing an error if we are trying to load it in an app outside of being an extension, so the webapp always crashes and not load. I think we will have to use Context then.

packages/shared/src/components/comments/CommentActionButtons.tsx
```diff
@@ -49,9 +49,10 @@ export default function CommentActionButtons({
     setNumUpvotes(comment.numUpvotes);
   }, [comment]);
 
+  const { requestMethod } = useCompanionProtocol();
```
rebelchris, line 52: Reading this now it's actually a bit weird that we use "companion" in the shared components.

Perhaps we should simply name it useRequestProtocol?
Not sure if I'm being dramatic here, but just find this should have nothing to do with the companion.

rebelchris (changes requested):
Well done Lee, I think we are almost there!
Just have one concern around naming.

packages/shared/src/components/comments/CommentActionButtons.tsx
sshanzel: Was thinking about this as well but had doubts due to the fetch method in it. Though I agree, request would fit better and doesn't necessarily mean it is only for graphql request.

packages/extension/src/companion/companionRequest.tsx
```diff
@@ -8,7 +8,7 @@ const proxyRequest = {
       url: args?.[0],
       document: args?.[1],
       variables: args?.[2],
-      queryKey: args?.[3],
+      headers: args?.[3],
```
idoshamun, line 11: Why did we decide to use headers instead of queryKey?

packages/shared/src/components/comments/CommentActionButtons.tsx
idoshamun: I agree. Shared components should be agnostic to application specific terminology such as `companion`

packages/shared/src/hooks/useCompanionProtocol.ts
```diff
@@ -0,0 +1,17 @@
...
+  fetchMethod?: typeof fetch;
+} => {
+  const client = useQueryClient();
+  const { companionRequest, companionFetch } =
+    client.getQueryData<CompanionProtocol>(COMPANION_PROTOCOL_KEY) || {};
+
+  const requestMethod = companionRequest || request;
+  const fetchMethod = companionFetch || fetch;
+
+  return { requestMethod, fetchMethod };
```
idoshamun, line 16: Consider useMemo here

packages/shared/src/hooks/usePostComment.ts
```diff
@@ -0,0 +1,223 @@
...
+    }
+  }, [enableShowShareNewComment]);
+
+  useEffect(() => {
+    if (initializeNewComment) {
+      openNewComment();
+    }
+  }, [initializeNewComment]);
+
+  return {
```
idoshamun, line 212: Consider useMemo

packages/shared/src/hooks/useUpvoteQuery.ts
```diff
@@ -0,0 +1,69 @@
...
+      upvotes,
+      requestQuery: {
+        queryKey: ['commentUpvotes', commentId],
+        query: COMMENT_UPVOTES_BY_ID_QUERY,
+        params: { id: commentId, first: DEFAULT_UPVOTES_PER_PAGE },
+      },
+    });
+  };
+
+  return {
```
idoshamun, line 63: Consider useMemo

packages/shared/src/lib/func.ts
```diff
@@ -1,2 +1,10 @@
 export const nextTick = (): Promise<unknown> =>
   new Promise((resolve) => setTimeout(resolve));
+
+export const safeJsonParse = <T = unknown>(data: string): T | string => {
```
idoshamun, line 4: This name is not the most explicit name I think.
It's not about safe but about returning a default or a fallback value

idoshamun (changes requested):
That was a big one! So much work 🤩

packages/extension/src/companion/companionRequest.tsx
sshanzel: I felt like it was much more fitting to use the headers, and that the third parameter was the header part - not sure why my initial implementation didn't throw an error but the other one did.

packages/shared/src/lib/func.ts
sshanzel: Makes total sense. I refactored it to `parseOrDefault`. Thank you! 🚢

packages/shared/src/hooks/useRequestProtocol.ts
```diff
@@ -0,0 +1,21 @@
...
+import { useQueryClient } from 'react-query';
+import { CompanionProtocol, COMPANION_PROTOCOL_KEY } from '../graphql/common';
+
+export const useRequestProtocol = (): {
+  requestMethod?: typeof request;
+  fetchMethod?: typeof fetch;
+} => {
+  const client = useQueryClient();
+  const { companionRequest, companionFetch } =
+    client.getQueryData<CompanionProtocol>(COMPANION_PROTOCOL_KEY) || {};
```
rebelchris, line 12: For this hook.

I think it's better if the hook can simply can accept a request and fetch method, but we set it from the companion itself?

Now we kind of determine that here, but the companion should be the source of truth for which requestmethod it uses.

Not sure if this makes sense?

packages/webapp/__tests__/PostPage.tsx
```diff
@@ -574,67 +519,3 @@ it('should not cut summary when there is a summary without reaching threshold',
   const fullSummary = await screen.findByText(summaryText);
   expect(fullSummary).toBeInTheDocument();
 });
-
```
rebelchris, line 577: I see we removed quite a lot of test cases are these all handled elsewhere?

rebelchris (changes requested):
Really amazing, still not convinced on the hook use.
I would really like it to be independant of the companion.

So can we have it simply accept a type of request but default to request itself?

packages/webapp/__tests__/PostPage.tsx
sshanzel: Yes, this was just reverted to handle the test cases in its own separate suite.

packages/shared/src/hooks/useRequestProtocol.ts
sshanzel: Totally agree. Let me refactor this one.

packages/webapp/__tests__/PostPage.tsx
rebelchris: Perfect, then forget about this comment, wanted to verify :D

packages/shared/src/hooks/useRequestProtocol.ts
rebelchris: Amazing refactor, loving this 💖

rebelchris (approved):
Lee, super amazing work 🥷

sshanzel:
@idoshamun thanks for the review! The newly introduced hooks should now utilize `useMemo`. Let me know if you have further concerns, will quickly work on it! ⚡ 🚢

idoshamun (approved):
Yes!!!

PR #982, opened 2022-05-23
feat: companion less intrusive asking of permission

sshanzel:
- Introduced a new hook to manage extension permissions
- Companion Permission component to contain the title and description (subject for A/B test).

~~WIP: Test cases~~

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [x] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-53 #done


packages/extension/src/background/index.ts
```diff
@@ -53,6 +54,20 @@ const sendBootData = async (req, sender) => {
 };
 
 async function handleMessages(message, sender) {
+  const permissions = await browser.permissions.contains({
+    origins: ['*://*/*'],
+  });
+  if (permissions) {
+    await browser.contentScripts.register({
```
rebelchris, line 61: Can't we leverage the created hook here?

rebelchris (changes requested):
Some quick notes, also don't forget to wrap the companion show ab test around the settings.
So people with off don't see anything.

Will do a bigger review back on my pc

packages/extension/src/background/index.ts
sshanzel: Not actually sure if this is used anywhere. I think this has a similar code to the one in the hook. Let me check to remove this quickly and see how it behaves.

packages/extension/src/companion/CompanionPermission.tsx
```diff
@@ -0,0 +1,85 @@
...
+} from '@dailydotdev/shared/src/lib/featureManagement';
+import { useExtensionPermission } from './useExtensionPermission';
+
+interface Content {
+  title: string;
+  description: string;
+}
+
+const contentVariation: Record<string, Content> = {
+  v1: {
```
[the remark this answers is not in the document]
sshanzel: I thought it would be fixed values from this image? Though I agree it would be much more flexible if we won't restrict it by version.
image

.circleci/config.yml
```diff
@@ -24,14 +24,14 @@ jobs:
           filename: combined-package-lock.txt
       - restore_cache:
           keys:
-            - deps-v3-{{ .Branch }}-{{ checksum "combined-package-lock.txt" }}
-            - deps-v3-{{ .Branch }}
-            - deps-v3-
+            - deps-v4-{{ .Branch }}-{{ checksum "combined-package-lock.txt" }}
+            - deps-v4-{{ .Branch }}
+            - deps-v4-
```
sshanzel, line 29: @rebelchris I have updated the deps reference for the keys. Am I missing something?

packages/extension/src/background/index.ts
rebelchris: Oh please don't remove this!
The background worker needs this to spin up the first time the user has given permissions as they technically two different instances.

packages/extension/src/companion/CompanionPermission.tsx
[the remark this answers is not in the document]
rebelchris: I mean generally when we roll out AB test we create a string variable for the title with V1 as the default.
But perhaps @idoshamun can shed some light on how he likes to proceed.

.circleci/config.yml
rebelchris: Nope it should be just the keys, I had this before for a update, never got that to work.
Perhaps we can leave this part and see what happens when we merge into companion branch?

packages/extension/src/companion/CompanionPermission.tsx
```diff
@@ -0,0 +1,68 @@
...
+        <Button
+          className="mt-1 w-[12.5rem] btn btn-primary"
+          onClick={registerContentScripts}
+          buttonSize="small"
+        >
+          Add the companion now!
+        </Button>
+      </CompanionSection>
+      <CompanionSection>
+        <a className="flex relative justify-center items-center" href="#">
```
rebelchris, line 47: I've asked Graham for the link for this

packages/extension/src/companion/CompanionPermission.tsx
```diff
@@ -0,0 +1,68 @@
...
+          <img
+            src="./companion_preview.png"
```
rebelchris, line 49: This is fine for me, but perhaps Ido might suggest to host it on the cloudinary? (Not sure what we default favor)

packages/extension/src/companion/CompanionPermission.tsx
```diff
@@ -0,0 +1,68 @@
...
+            alt="Companion video preview"
+            className="rounded-10 w-[11.25rem] h-[6.875]"
+          />
+          <PlayIcon className="absolute w-10 h-10" />
+        </a>
+        <a
+          href="#"
```
rebelchris, line 56: Same here, just a note so we won't forget to update

packages/extension/src/companion/CompanionPermission.tsx
```diff
@@ -0,0 +1,68 @@
...
+          className="mt-2 text-center underline typo-footnote text-theme-status-cabbage"
+        >
+          Watch the Companion
+          <br />
+          Overview
```
rebelchris, line 61: I think this is also a AB test variant.

packages/extension/src/newtab/MainFeedPage.tsx
```diff
@@ -108,13 +135,42 @@ export default function MainFeedPage({
...
-        user && (
-          <SimpleTooltip content="Do Not Disturb" placement="bottom">
-            <HeaderButton
-              icon={<TimerIcon />}
-              className="btn-tertiary"
-              onClick={() => setShowDnd(true)}
-              pressed={showDnd}
+        placement === 'header' &&
+        !contentScriptGranted && (
+          <SimpleTooltip
```
rebelchris, line 140: I feel like this can be it's own component as it will be re-used in the sidebar later on.

rebelchris (changes requested):
Amazing changes, you are too quick 🚀
Some small notes, but really well done 💖

packages/extension/src/companion/CompanionPermission.tsx
sshanzel: Thank you! I've also created a test that they must have the same value so always update both of them.

packages/extension/src/newtab/MainFeedPage.tsx
sshanzel: Totally makes sense. Let me refactor this one! 🚢

packages/extension/src/companion/CompanionPermission.tsx
sshanzel: I think it is only the title and description that we planned to A/B test for the upcoming release? Let me check with the team.

sshanzel: This is for the text of the link to the video. Do you mean the "Add the companion" button

packages/extension/src/companion/CompanionPermission.tsx
sshanzel: This should be the only concern now. Once both PRs are merged, we can be ready for internal testing 🚢

packages/extension/src/companion/CompanionPermission.tsx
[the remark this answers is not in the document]
sshanzel: I have now specified different variables for title and description to have more flexibility on how we would want it to be AB tested. Thank you for the thoughts!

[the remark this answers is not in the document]
idoshamun: I agree with @rebelchris. The version is for the layout not the copy

packages/extension/src/background/index.ts
```diff
@@ -53,6 +55,13 @@ const sendBootData = async (req, sender) => {
 };
 
 async function handleMessages(message, sender) {
+  const permissions = await browser.permissions.contains({
+    origins: ['*://*/*'],
+  });
+  if (permissions) {
+    await registerBrowserContentScripts();
+  }
```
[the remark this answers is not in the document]
sshanzel: Yes. Below was Chris' earlier response regarding this part.

> The background worker needs this to spin up the first time the user has given permissions as they are technically two different instances.

packages/extension/src/companion/CompanionPermission.tsx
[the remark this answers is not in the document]
sshanzel: Cool. The current implementation should support two different variables for title and description copy.

packages/shared/src/components/Settings.tsx
```diff
@@ -166,21 +173,23 @@ export default function Settings({
...
+          <div className="flex flex-col items-start pl-1.5 -my-0.5">
+            <Switch
+              inputId="hide-companion-switch"
+              name="hide-companion"
+              className="my-3"
+              checked={!optOutCompanion}
+              onToggle={toggleOptOutCompanion}
+              compact={false}
+            >
+              Enable widget on articles
```
idoshamun, line 188: Is this the final copy and design?
I'm not sure we need a dedicated section for it and the copy doesn't match IMO

rebelchris: Small note we need this to turn off the companion else users have no way to disable it. But we can raise the copy comment

packages/extension/src/companion/CompanionPopupButton.tsx
```diff
@@ -0,0 +1,77 @@
...
+    ? CompanionFilledIcon
+    : CompanionOutlineIcon;
+
+  useEffect(() => {
+    if (!loadedUserFromCache) {
+      return;
+    }
+
+    setShowCompanionPermission(!!user);
+  }, [user, loadedUserFromCache]);
```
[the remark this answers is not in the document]
sshanzel: If the user is logged in, the default value of showing the popup should be on load.

packages/extension/src/companion/CompanionPopupButton.tsx
```diff
@@ -0,0 +1,77 @@
...
+interface CompanionPopupButtonProps {
+  placement: string;
+}
+
+export const CompanionPopupButton = ({
+  placement,
+}: CompanionPopupButtonProps): ReactElement => {
+  if (placement === 'off') {
+    return null;
+  }
```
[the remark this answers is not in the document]
sshanzel: Yes, we do that as well from the outside, though I have thought of passing the parameter for when this is used in the sidebar and the UI might be different than just a button.

packages/extension/src/newtab/MainFeedPage.tsx
```diff
@@ -108,16 +118,7 @@ export default function MainFeedPage({
...
-          <SimpleTooltip content="Do Not Disturb" placement="bottom">
-            <HeaderButton
-              icon={<TimerIcon />}
-              className="btn-tertiary"
-              onClick={() => setShowDnd(true)}
-              pressed={showDnd}
-            />
-          </SimpleTooltip>
-        )
+        placement === 'header' && <CompanionPopupButton placement={placement} />
```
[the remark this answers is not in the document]
sshanzel: We don't use it here, anymore. It is now part of the sidebar, and the passed parameter `additionalButtons` wasn't being used as well.

packages/extension/src/companion/CompanionPopupButton.tsx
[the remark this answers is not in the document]
idoshamun: It's not if the user is logged in, it's if the user already exists when we rolled it out. I suggest using the alerts system we built

packages/extension/src/companion/CompanionPopupButton.tsx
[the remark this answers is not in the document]
idoshamun: I think it's redundant. Let's remove it if it's safe to do

sshanzel:
@idoshamun thanks for the review. I have now applied to track when the user has initiated the request for permission and only when it is approved (since we can the percentage of denied will be the other half of approved).

packages/extension/src/background/index.ts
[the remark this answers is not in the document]
rebelchris: It's inside the handleMessages as this get's triggered from the new tab.
Mostly about the first page you open after enabling the companion where this is needed

packages/extension/src/companion/CompanionPopupButton.tsx
```diff
@@ -0,0 +1,75 @@
...
+          showCompanionPermission
+            ? 'btn-primary-cabbage'
+            : 'btn-secondary-cabbage',
+        )}
+        icon={
+          <CompanionIcon
+            className={
+              showCompanionPermission
+                ? 'w-7 h-7 text-theme-label-primary'
+                : 'w-6 h-6 text-theme-status-cabbage'
```
rebelchris, line 68: Why do we change the size based on permissions here?

packages/shared/src/components/Settings.tsx
```diff
@@ -164,21 +171,18 @@ export default function Settings({
...
-          >
-            Enable widget on articles
-          </Switch>
+          {companionPlacement !== 'off' && (
+            <Switch
+              inputId="hide-companion-switch"
+              name="hide-companion"
+              className="my-3"
+              checked={!optOutCompanion}
+              onToggle={toggleOptOutCompanion}
```
rebelchris, line 180: As raised on [slack](https://dailydotdev.slack.com/archives/C03G1FZPTEX/p1653458794644749) we should also request the permission if they are not granted yet here.

packages/shared/src/lib/featureManagement.ts
```diff
@@ -98,6 +98,32 @@ export class Features {
...
+  );
+
+  static readonly CompanionPermissionDescription = new Features(
+    'companion_permission_description',
+    'Heads up! We need to ask for some extra permissions so you can enjoy the power of the companion.',
+  );
+
+  static readonly CompanionPermissionLink = new Features(
+    'companion_permission_link',
+    'Watch the Companion\nOverview',
```
rebelchris, line 119: Not sure if we even need the new line?
Is it not just a design issue where they showcased the new line for that example?

rebelchris (changes requested):
Amazing work Lee, really impressed 💖

packages/extension/src/companion/CompanionPopupButton.tsx
sshanzel: The exported size isn't equal. Not sure if it's because of being filled or not.

packages/shared/src/lib/featureManagement.ts
sshanzel: Yes, it is from the design. Not sure if the new line was intentional but it looked much better so I applied it.

rebelchris: Ah ok perfect

packages/extension/src/companion/CompanionPopupButton.tsx
rebelchris: Agreed, let me note it in that existing ticket ;)

packages/shared/src/components/Settings.tsx
sshanzel: Oh, okay. Though that would make the behavior of the toggle inconsistent, wouldn't it? Now, it made me question, shouldn't we make this available only if the user has already accepted the permission?

rebelchris: Would you mind raising this with Product?

sshanzel: Totally cool. Was just finishing the work in API then was about to send a message to the channel. Thank you! 🚢

packages/extension/src/companion/useExtensionPermission.ts
```diff
@@ -0,0 +1,71 @@
...
+      if (permissions) {
+        client.setQueryData(contentScriptKey, true);
+      }
+    };
+    permissionCall();
+  }, []);
+
+  const registerContentScripts = async () => {
+    trackEvent({
+      origin,
```
rebelchris, line 53: Origin is not available on trackEvent it should be set as an `extra` object.
We have some examples already of this.

rebelchris (changes requested):

packages/extension/src/companion/useExtensionPermission.ts
rebelchris: @sshanzel Pushed a fix for this.

sshanzel: Yes, I tried to have it as an object but it is being rejected by typescript. Which leads the build to fail.

idoshamun:
> Also, we tried changing the keys for the pipeline's cache but it seems to have not been cleared off, still. Have you tried something similar before?

That's weird. Maybe it's another reason? I will give it another look later

packages/shared/src/lib/standardizeUrl.ts
```diff
@@ -0,0 +1,22 @@
...
+const subtractDomain = (url: string): string => {
+  const matches = url.match(
+    /^(?:https?:\/\/)?(?:[^@/\n]+@)?(?:www\.)?([^:/?\n]+)/i,
+  );
+  return matches && matches[1];
+};
+
+const removeTrailingSlash = (url: string): string => url.replace(/\/$/, '');
+
+export const standardizeURL = (url: string): string => {
```
idoshamun, line 15: In the future, we will have to do the standardization on the server-side for sure because this list will be dynamic.
Do we want to fix it before we launch perhaps? But waiting holding back the internal testing

rebelchris: Let's fix before launch, but after internal if that makes sense?
I'll simply send the full URL to the API and add this function there.

Let me know if that works for you?

rebelchris:
@idoshamun The cache issue was due to some other issues, I think I solved those now

sshanzel:
@idoshamun the alert for managing when the popup for the companion should be displayed is now available locally. To simulate it, kindly re-install. It should not do anything, but if you update the code, let's say for `featureManagement` file to set the placement of companion from `off` to `header` - it triggers update, so once it refreshes it should have the desired value.

packages/shared/src/lib/standardizeUrl.ts
rebelchris: @idoshamun Removed from this project and introduced the same on the API level.
https://github.com/dailydotdev/daily-api/pull/946

packages/extension/src/background/index.ts
```diff
@@ -107,7 +118,15 @@ browser.browserAction.onClicked.addListener(() => {
   browser.tabs.create({ url, active: true });
 });
 
-browser.runtime.onInstalled.addListener(async () => {
+browser.runtime.onInstalled.addListener(async (details) => {
+  if (details.reason === 'update') {
+    updateCompanionAlerts({ displayCompanionPopup: true });
+  }
```
rebelchris, line 124: What happens when a new users installs -> then updates to a next version?
Perhaps we should lock in to only this version or something specific?

Let me know what you think

rebelchris (changes requested):
Nice addition, just one question from my side.

packages/extension/src/background/index.ts
sshanzel: Good point. I think we need to add another check if the value is explicitly false, meaning the user has seen the popup (as it was undefined initially), we won't turn this to true. WDYT?

sshanzel:
@rebelchris let me know what you think of the change to check whether to update the local data. Thank you!

packages/extension/src/companion/CompanionPermission.spec.tsx
```diff
@@ -0,0 +1,89 @@
...
+    });
+  });
+
+  it('should render the image preview for the video', async () => {
+    renderComponent();
+    const img = await screen.findByAltText('Companion video preview');
+    expect(img).toBeInTheDocument();
+  });
+
+  it('should ask for permission when button is clicked', async () => {
```
rebelchris, line 80: Could we perhaps add a test case of the localStorage mimic here?
I want to make sure future code doesn't overwrite that functionality

packages/extension/src/companion/CompanionPopupButton.tsx
```diff
@@ -0,0 +1,104 @@
...
+    return {};
+  }
+};
+
+export const updateCompanionAlerts = (
+  current: CompanionAlerts,
+  updated: CompanionAlerts,
+): CompanionAlerts => {
+  const data = { ...current, ...updated };
+  window.localStorage.setItem(COMPANION_ALERTS_LOCAL_KEY, JSON.stringify(data));
```
rebelchris, line 36: Should we not opt to use `storageWrapper` for this?

idoshamun (changes requested):

rebelchris (changes requested):
Found some small things, but we are almost there

packages/extension/src/companion/CompanionPopupButton.tsx
```diff
@@ -10,29 +8,52 @@ import AnalyticsContext from '@dailydotdev/shared/src/contexts/AnalyticsContext'
...
+  return data;
+};
+
 export const CompanionPopupButton = (): ReactElement => {
+  const alerts = useRef(getCompanionAlerts());
```
idoshamun, line 41: Why `useRef`? I think it's better to just utilize this value as the initial value for the use state

packages/extension/src/companion/CompanionPopupButton.tsx
```diff
@@ -0,0 +1,104 @@
...
+
+  if (!data) {
+    return {};
+  }
+
+  try {
+    return JSON.parse(data);
+  } catch (ex) {
+    return {};
+  }
```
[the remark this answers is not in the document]
sshanzel: It would return what was given, but for this one, it will return an empty object literal

packages/extension/src/companion/CompanionPopupButton.tsx
sshanzel: Good point, yes, we should use it here.

packages/extension/src/companion/CompanionPopupButton.tsx
```diff
@@ -10,29 +8,52 @@ import AnalyticsContext from '@dailydotdev/shared/src/contexts/AnalyticsContext'
 import { CompanionPermission } from './CompanionPermission';
 import { useExtensionPermission } from './useExtensionPermission';
 
+export const COMPANION_ALERTS_LOCAL_KEY = 'companion:alerts';
+
+interface CompanionAlerts {
+  displayCompanionPopup?: boolean;
+}
+
+export const getCompanionAlerts = (): CompanionAlerts => {
```
idoshamun, line 17: I think it's better to have this function in a separate file. Otherwise, when you import it in the background worker it might bloat its bundle size eventually

packages/extension/src/companion/CompanionPopupButton.tsx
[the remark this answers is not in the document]
rebelchris: Cool leave it for now perhaps in the future we can extend it to include a actual default :)

packages/extension/src/background/index.ts
```diff
@@ -118,7 +121,19 @@ browser.browserAction.onClicked.addListener(() => {
   browser.tabs.create({ url, active: true });
 });
 
-browser.runtime.onInstalled.addListener(async () => {
+browser.runtime.onInstalled.addListener(async (details) => {
+  const data = getCompanionAlerts();
```
idoshamun, line 125: Should we generalize it perhaps? Not just for companion alerts but every alert we possibly need and then expose it in a context. We can do it later as well but just change the name of the storage key.
cc @rebelchris

idoshamun (changes requested):

packages/extension/src/background/index.ts
rebelchris: Very good point, if we can generalize this to be non companion related we and in it's own scope we are there :)

packages/extension/src/companion/CompanionPopupButton.tsx
sshanzel: We intend the value to be always accurate but we don't want the rerender triggered when updated as the use effect will produce a loop.

idoshamun: Not sure I understand why

idoshamun (approved):
Incredible!!

rebelchris (approved):
Lets go internal release party 🥳

PR #986, opened 2022-05-26
fix: height calculation accuracy

sshanzel:
- Since we require pixel-perfect calculation, we are now utilizing px instead of rem.

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

DD-{number} #done

### DIFF
--- a/packages/extension/src/companion/CompanionContent.tsx
+++ b/packages/extension/src/companion/CompanionContent.tsx
@@ -27,6 +27,10 @@ type CompanionContentProps = {
   onViewComments?: (state: boolean) => void;
 };
 
+const COMPANION_TOP_OFFSET_REM = 7.5;
+const PX_PER_REM = 16;
+const remToPx = (rem: number) => rem * PX_PER_REM;
+
 export default function CompanionContent({
   post,
   viewComments,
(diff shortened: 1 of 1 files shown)


packages/extension/src/companion/CompanionContent.tsx
```diff
@@ -59,10 +63,9 @@ export default function CompanionContent({
     }
 
     const { height } = el.getBoundingClientRect();
-    const topOffset = 7.5;
-    const remValue = height / 16;
-    const rem = `${remValue + topOffset}rem`;
-    setHeightRem(rem);
+    const topOffset = remToPx(COMPANION_TOP_OFFSET_REM);
```
idoshamun, line 66: I suggest already storing the offset in pixels instead of calculating it at runtime

sshanzel: Totally agree. I have now pushed the changes. 🚢

packages/extension/src/companion/CompanionContent.tsx
```diff
@@ -27,14 +27,18 @@ type CompanionContentProps = {
   onViewComments?: (state: boolean) => void;
 };
 
+const PX_PER_REM = 16;
+const COMPANION_TOP_OFFSET_REM = 7.5;
+const COMPANION_TOP_OFFSET_PX = PX_PER_REM * COMPANION_TOP_OFFSET_REM;
```
idoshamun, line 32: We already know the value so let's do it this way. It's simpler.

```suggestion
const COMPANION_TOP_OFFSET_PX = 120;
```

PR #987, opened 2022-05-27
fix: toggling discussion and user mention

sshanzel:
- Should open the comments list even when the companion is minimized.
- Fixes the user mention on shadow dom with content editable (due to Selection API)

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

DD-{number} #done

### DIFF
--- a/packages/extension/src/companion/Companion.tsx
+++ b/packages/extension/src/companion/Companion.tsx
@@ -113,6 +113,14 @@ export default function Companion({
     routeChangedCallbackRef.current();
   }, [containerRef]);
 
+  useEffect(() => {
+    if (!isCommentsOpen) {
+      return;
+    }
+
+    setCompanionState(true);
+  }, [isCommentsOpen]);
+
   return (
     <Container
       containerRef={containerRef}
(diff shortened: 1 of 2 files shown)


packages/extension/src/companion/Companion.tsx
```diff
@@ -113,6 +113,14 @@ export default function Companion({
     routeChangedCallbackRef.current();
   }, [containerRef]);
 
+  useEffect(() => {
```
idoshamun, line 116: I think instead of doing `useEffect`, we should add `setCompanionState(true)` to  `onOpenComments`. Otherwise, it may feel like magic

packages/shared/src/lib/element.ts
```diff
@@ -7,8 +7,13 @@ type Row = number;
 export type CaretOffset = [number, number];
 export type CaretPosition = [Column, Row];
 
+const getShadowDom = (): Document =>
+  document.querySelector('daily-companion-app')
+    ?.shadowRoot as unknown as Document;
+
 export function getCaretPostition(el: Element): CaretPosition {
-  const sel = window.getSelection();
+  const dom = getShadowDom() || window;
```
idoshamun, line 15: Nice!

idoshamun (changes requested):
One minor comment, other than that looks great

idoshamun (approved):

PR #996, opened 2022-06-04
refactor(docker-compose): add health-check and depends-on

### LINKED ISSUE #635: Why to use wait-for-it, when docker-compose has depends_on and healthchecks?
Hi! That's just a question, not a big deal, but why to use `wait-for-it.sh`, instead of [`healthcheck`](https://docs.docker.com/compose/compose-file/compose-file-v2/#healthcheck) and [`depends_on`](https://docs.docker.com/compose/compose-file/compose-file-v2/#depends_on). 

For example, there is how it could be used:

```yaml
daily-postgres:
  ...
  test: [ "CMD-SHELL", "pg_isready -U ${POSTGRES_USER:-postgres}" ]
  ...

daily-api:
  ...
  depends_on:
      daily-postgres:
        condition: service_healthy
  ...
```

I personally use it in all my own projects, and it works great, and in theory, even better then `wait-for-it.sh`, since it can be configured in any way you want

jonatan5524:
Replace the wait-for.sh in docker-compose with depends-on and healthcheck

Closes #635 

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

DD-{number} #done

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -8,6 +8,11 @@ volumes:
 services:
   daily-redis:
     image: redis:5.0.10
+    healthcheck:
+      test: redis-cli ping
+      interval: 1s
+      timeout: 3s
+      retries: 60
     command: redis-server --appendonly yes
     volumes:
       - redis:/data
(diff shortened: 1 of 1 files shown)


idoshamun (approved):
I wasn't even aware of the `healthcheck` feature. This is so useful!
Thanks a lot :)

rebelchris:
Thanks a lot @jonatan5524 for putting this in place 💖

PR #997, opened 2022-06-06
fix: refresh token utilization

sshanzel:
## Changes

I think this is one comment from our beta user has really helped us finding the root cause.

> As an FYI, I couldn’t commit the above message because of the “Access denied” error, but after a refresh of the page, I could. Same thing with the “markdown preview”, after the refresh it works. So perhaps a timing issue since I took a while to write the comment? Not sure, but I’ll keep an eye to see if I can reproduce it.

- We received quite a few feedback relating to making a comment. After further investigation, it was realized the token for making requests was not refreshed, meaning, every request gets rejected (upvotes/comments/upvotes modal/etc).
- We are now utilizing the custom hook used in boot for refreshing the token.
- Instead of re-requesting the whole boot data - we now instead request for the refreshed token.

Blocked until this PR is merged: https://github.com/dailydotdev/daily-gateway/pull/471

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

Manually waited for quite some while to ensure the token is refreshed (vice-versa), to ensure workability.
Preview:
image

image

As it can be noticed, the requests have succeeded after the refresh token call.

Preview before the fix:
image

DD-WT-113 #done

### DIFF
--- a/packages/extension/src/background/index.ts
+++ b/packages/extension/src/background/index.ts
@@ -1,5 +1,5 @@
 import 'content-scripts-register-polyfill';
-import { browser } from 'webextension-polyfill-ts';
+import { browser, Runtime } from 'webextension-polyfill-ts';
 import { getBootData } from '@dailydotdev/shared/src/lib/boot';
 import { apiUrl } from '@dailydotdev/shared/src/lib/config';
 import { parseOrDefault } from '@dailydotdev/shared/src/lib/func';
@@ -69,7 +69,26 @@ const sendBootData = async (req, sender) => {
   });
 };
 
-async function handleMessages(message, sender) {
+const sendRequestResponse = async (
+  requestKey: string,
+  req: Promise<unknown>,
+  sender: Runtime.MessageSender,
+  variables,
+) => {
+  const key = parseOrDefault(requestKey);
+  const url = sender?.tab?.url?.split('?')[0];
+  const [deviceId, res] = await Promise.all([getOrGenerateDeviceId(), req]);
+
+  return browser.tabs.sendMessage(sender?.tab?.id, {
+    deviceId,
+    url,
+    res,
+    req: { variables },
+    key,
+  });
+};
+
+async function handleMessages(message, sender: Runtime.MessageSender) {
   await getContentScriptPermissionAndRegister();
 
   if (message.type === 'CONTENT_LOADED') {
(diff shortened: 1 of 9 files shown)


sshanzel:
@rebelchris @idoshamun in time this gets approved. In my humble opinion, I think we should consider publishing to stores once merged as the actions themselves (upvotes/making a comment/etc) are what we expect from the users to accomplish.

idoshamun:
The `boot` endpoint should already renew the access token when it's applicable. I suggest not introducing a new endpoint for refreshing a token at the moment. We can call boot if necessary again. I wonder why the token wasn't actually refreshed as we have everything in place. We need to find the root cause before introducing something new.

sshanzel:
> I suggest not introducing a new endpoint for refreshing a token at the moment.

The reason why I created another one for refreshing the token, is the `boot` itself fetches a lot of data which are not needed in the situation.

> I wonder why the token wasn't actually refreshed as we have everything in place. We need to find the root cause before introducing something new.

We instantiate the `useRefreshToken` hook in the `BootProvider` and we did not use that custom provider in the companion.

With that, we can definitely request from `boot` again, if we don't mind the amount of data processed in that endpoint just to refresh the token.

sshanzel:
There's a whole other permanent fix for this but I hesitated for the moment since we have the upcoming re-work of authentication.

The fix I was thinking of is, to create a middleware in the API server whenever we check the jwt validity and to refresh the token if found to be expired. So we won't have to create a timeout and force the refresh of token from the frontend.

idoshamun:
> to create a middleware in the API server whenever we check the jwt validity and to refresh the token if found to be expired. So we won't have to create a timeout and force the refresh of token from the frontend.

This is can introduce some security issues, so I wouldn't go down this path.

> With that, we can definitely request from boot again if we don't mind the amount of data processed in that endpoint just to refresh the token.

I don't mind really. It will not happen that often. Thanks for considering it though :)
We might optimize this flow with the new authentication but for now let's keep it simple.

sshanzel:
Still works as intended. Attached Preview if the performed manual testing:

image

idoshamun (approved):
Looks great now! Please wait for Chris's review as well

rebelchris (approved):
Nice this looks great!
Great find on abstracting the refresh token hook.

PR #1000, opened 2022-06-06
fix: firefox companion multiple instances

sshanzel:
Every step of the way, we check content scripts if permission is already given. With Chrome, no matter how many you register your scripts, when re-injecting it, the current one just gets replaced. For Firefox, it registers another instance.

There are two ways to go about this issue.

The first is (https://github.com/dailydotdev/apps/commit/099a46205e2b80153224ca51e23bd0a8a77e35be), to target where we inject the scripts and exclude Firefox if `permissions.contains` return `true`. The API only returns true if it was registered already.

The second one is (https://github.com/dailydotdev/apps/commit/c0648292321a8512330de3b60161eef977ed8fb1), before creating another instance in the DOM, we will check first if our element already exists. 

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

DD-{number} #done

### DIFF
--- a/packages/extension/src/companion/useExtensionPermission.ts
+++ b/packages/extension/src/companion/useExtensionPermission.ts
@@ -29,9 +29,10 @@ export const getContentScriptPermission = (): Promise<boolean> =>
 
 export const getContentScriptPermissionAndRegister =
   async (): Promise<void> => {
+    const isFirefox = process.env.TARGET_BROWSER === 'firefox';
     const permission = await getContentScriptPermission();
 
-    if (permission) {
+    if (permission && !isFirefox) {
       await registerBrowserContentScripts();
     }
   };
--- a/packages/extension/src/content/index.tsx
+++ b/packages/extension/src/content/index.tsx
@@ -1,16 +1,20 @@
 import { browser } from 'webextension-polyfill-ts';
 
-// Inject app div
-const appContainer = document.createElement('daily-companion-app');
-document.body.appendChild(appContainer);
+const hasRendered = !!document.querySelector('daily-companion-app');
 
-// Create shadow dom
-const shadow = document
-  .querySelector('daily-companion-app')
-  .attachShadow({ mode: 'open' });
+if (!hasRendered) {
+  // Inject app div
+  const appContainer = document.createElement('daily-companion-app');
+  document.body.appendChild(appContainer);
 
-const wrapper = document.createElement('div');
-wrapper.id = 'daily-companion-wrapper';
-shadow.appendChild(wrapper);
+  // Create shadow dom
+  const shadow = document
+    .querySelector('daily-companion-app')
+    .attachShadow({ mode: 'open' });
 
-browser.runtime.sendMessage({ type: 'CONTENT_LOADED' });
+  const wrapper = document.createElement('div');
+  wrapper.id = 'daily-companion-wrapper';
+  shadow.appendChild(wrapper);
+
+  browser.runtime.sendMessage({ type: 'CONTENT_LOADED' });
+}


idoshamun:
The problem with the second implementation is that you will still have so many registered content scripts but you actually get rid of the symptom. You don't solve the problem. I suggest going with the #1 but I'd love to hear @rebelchris as well.
Also, I suggest a careful testing procedure for this.

idoshamun:
Also @sshanzel why do we call this register function so many times?

sshanzel:
> The problem with the second implementation is that you will still have so many registered content scripts but you actually get rid of the symptom.

Yes, I have the same sentiment for the second one.

> Also @sshanzel why do we call this register function so many times?

There was an issue we had when we close the whole browser using Chrome, the companion won't load anymore. So we inject in every part where the browser might start running.

For re-injecting every call to `handleMessages` - I'm not sure what was the reason behind the call itself. Probably @rebelchris has some idea about it.

idoshamun:
People also reported the companion suddenly disappears. I wonder if it has to do with this

sshanzel:
> People also reported the companion suddenly disappears. I wonder if it has to do with this

Hmm. I think that could be a separate issue since we don't have yet any cleanup that removes the element from the DOM.

rebelchris:
@sshanzel The register on handleMessages might actually not be needed anymore now that we introduced the lifecycle callback.
Perhaps we should give it a try by only using that?

We want to minimize the number of calls of register wherever possible.
This was initially the only point we could register.

sshanzel:
> Perhaps we should give it a try by only using that?

Sure, let me try it after I make progress on the navigation issue.

> could this not simply be if(!permission)?

For Chrome, it is actually necessary to inject again, even though the permission (and injected on the first instance) was already given. Really annoying sometimes they work differently on very important cases like this 😓

sshanzel:
@rebelchris tested it out on Firefox and Chrome. Including the edge-cases we had. Things still work as intended 🙌

sshanzel:
@idoshamun looking forward to your review. Thanks for the thoughts! 🚢 🚢 🚢

sshanzel:
@rebelchris @idoshamun after rigorously testing out the app. It seems we still need the check of permission in `handleMessages`. The issue from Chrome when you close the browser has switched to FF even though we had the `onStartup` event handler to register the scripts.

With that, I have changed the way we recognize whether the script was injected already or not.

packages/extension/src/companion/useExtensionPermission.ts
```diff
@@ -27,12 +27,14 @@ export const getContentScriptPermission = (): Promise<boolean> =>
     origins: ['*://*/*'],
   });
 
+let hasInjectedScripts = false;
```
rebelchris, line 30: Can't we use useState here?

packages/extension/src/content/index.tsx
```diff
@@ -1,25 +1,29 @@
 import { browser } from 'webextension-polyfill-ts';
 
-// Inject app div
-const appContainer = document.createElement('daily-companion-app');
-document.body.appendChild(appContainer);
+const hasRendered = !!document.querySelector('daily-companion-app');
```
rebelchris, line 3: for some reason I felt like `isRendered` made more sense looking at the rest of our code-base (but it's an optional change)

rebelchris (approved):
Some minor things, non blocking, nice work on this.

packages/extension/src/content/index.tsx
sshanzel: Totally agree. Let me update it quickly! 🚢

packages/extension/src/companion/useExtensionPermission.ts
sshanzel: Good point, though the function is not a hook as it is also needed to run in the background worker, so we can't turn it to state.

packages/extension/src/companion/index.tsx
```diff
@@ -28,7 +28,7 @@ browser.runtime.onMessage.addListener(
     visit,
     accessToken,
   }) => {
-    if (settings.optOutCompanion) {
+    if (!settings || settings?.optOutCompanion) {
```
idoshamun, line 31: Why are we checking for `!settings`? is it still usable by anonymous users?

idoshamun (approved):
One question. Other than that, it looks great!

packages/extension/src/companion/index.tsx
sshanzel: Yes. Our anonymous users store cache data for their settings which are still being utilized. Also, boot data returns the default value for the anonymous one.

Reference snippet:
https://github.com/dailydotdev/apps/blob/af42b78916c8ed4d2dd782450e11bc1ac341141f/packages/extension/src/background/index.ts#L55-L58

PR #1002, opened 2022-06-07
feat: new outline icons

kirillkurko:

This PR introduces new outline icons used across app

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

DD-{number} #done


packages/shared/src/components/Icon.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+const Icon: React.VFC<Props> = ({
+  filled = false,
+  size = 'small',
+  className = '',
+  IconOutlined,
+  IconFilled,
+}) => {
+  const iconSize = IconSize[size];
+  const classes = `w-${iconSize} h-${iconSize} pointer-events-none ${className}`;
```
sshanzel, line 27: For concatenating classes, we use `classNames` to make it much easier so we won't have to explicitly set an empty string value for `className`, plus other stuffs.

packages/shared/src/components/Icon.tsx
```diff
@@ -0,0 +1,40 @@
+import React from 'react';
+
+export type Size = 'small' | 'medium' | 'large';
+
+const IconSize = {
+  small: 5,
+  medium: 6,
+  large: 8,
+};
```
sshanzel, line 9: When using tailwindcss with dynamic classes, you must not concatenate the value manually. You must have the whole value of the expected class as the former will get ignored by tree-shaking.
```suggestion
const IconSize = {
  small: 'w-5 h-5',
  medium: 'w-6 h-6',
  large: 'w-8 h-8',
};
```

packages/shared/src/components/Icon.tsx
```diff
@@ -0,0 +1,40 @@
...
+  return (
+    <>
+      {filled ? (
+        <IconFilled className={classes} />
+      ) : (
+        <IconOutlined className={classes} />
+      )}
+    </>
+  );
+};
```
sshanzel, line 38: I think we can omit the extra fragment and simplify passing of the props.
```suggestion
  const IconComponent = filled ? IconFilled : IconOutlined;
  return <IconComponent className={classes} />;
```

packages/shared/src/components/ProfileMenu.tsx
```diff
@@ -44,12 +44,13 @@ export default function ProfileMenu(): ReactElement {
         <Item>
           <Link href={user.permalink} passHref prefetch={false}>
             <a className="flex w-full">
-              <UserIcon className="mr-2 text-xl" /> Profile
+              <UserIcon filled={false} className="mr-2 text-xl" /> Profile
```
sshanzel, line 47: Really sorry, I know you weren't the one who added the value for `text-xl` but can you check the equivalent value for it in terms of our guideline sizing? Thank you!

packages/shared/src/components/ProfileMenu.tsx
```diff
@@ -44,12 +44,13 @@ export default function ProfileMenu(): ReactElement {
...
             </a>
           </Link>
         </Item>
         <Item onClick={() => setShowAccountDetails(true)}>
-          <SettingsIcon className="mr-2 text-xl" /> Account details
+          <SettingsIcon filled={false} className="mr-2 text-xl" /> Account
```
sshanzel, line 52: Same goes here for `text-xl`. Thank you! 🙇

packages/shared/src/components/icons/Bookmark/index.tsx
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+import Icon, { Size } from '../../Icon';
+import OutlinedIcon from './outlined.svg';
+import FilledIcon from './filled.svg';
+
+type Props = {
```
sshanzel, line 6: I think we should create an interface that will be shared across the Icon components. It is most likely the shape of their props are the same.

packages/shared/src/components/icons/Copy/index.tsx
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+import Icon, { Size } from '../../Icon';
+import OutlinedIcon from './outlined.svg';
+import FilledIcon from './filled.svg';
+
+type Props = {
+  filled?: boolean;
+  size?: Size;
+};
```
sshanzel, line 9: Tagging for visibility regarding the concern for shared interface.

packages/shared/src/components/icons/DevCard/index.tsx
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+import Icon from '../../Icon';
+import OutlinedIcon from './outlined.svg';
+import FilledIcon from './filled.svg';
+
+type Props = {
+  filled?: boolean;
+  className?: string;
+};
```
sshanzel, line 9: Tagging for visibility regarding the concern for shared interface.

packages/shared/src/components/icons/Discuss/index.tsx
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+import Icon, { Size } from '../../Icon';
+import OutlinedIcon from './outlined.svg';
+import FilledIcon from './filled.svg';
+
+type Props = {
+  filled?: boolean;
+  size?: Size;
+};
```
sshanzel, line 9: Tagging for visibility regarding the concern for shared interface.

packages/shared/src/components/icons/Eye/index.tsx
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+import Icon from '../../Icon';
+import OutlinedIcon from './outlined.svg';
+import FilledIcon from './filled.svg';
+
+type Props = {
+  filled?: boolean;
+};
```
sshanzel, line 8: Tagging for visibility regarding the concern for shared interface.

packages/shared/src/components/icons/Filter/index.tsx
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+import Icon, { Size } from '../../Icon';
+import OutlinedIcon from './outlined.svg';
+import FilledIcon from './filled.svg';
+
+type Props = {
+  filled?: boolean;
+  size: Size;
+};
```
sshanzel: Tagging for visibility regarding the concern for shared interface.

packages/shared/src/components/icons/Home/index.tsx
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+import Icon, { Size } from '../../Icon';
+import OutlinedIcon from './outlined.svg';
+import FilledIcon from './filled.svg';
+
+type Props = {
+  filled?: boolean;
+  size?: Size;
+};
```
sshanzel, line 9: Tagging for visibility regarding the concern for shared interface.

packages/shared/src/components/icons/Hot/index.tsx
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+import Icon from '../../Icon';
+import OutlinedIcon from './outlined.svg';
+import FilledIcon from './filled.svg';
+
+type Props = {
+  filled?: boolean;
+};
```
sshanzel, line 8: Tagging for visibility regarding the concern for shared interface.

packages/shared/src/components/icons/Moon/index.tsx
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+import Icon from '../../Icon';
+import OutlinedIcon from './outlined.svg';
+import FilledIcon from './filled.svg';
+
+type Props = {
+  filled?: boolean;
+};
```
sshanzel, line 8: Tagging for visibility regarding the concern for shared interface.

packages/shared/src/components/icons/Search/index.tsx
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+import Icon, { Size } from '../../Icon';
+import OutlinedIcon from './outlined.svg';
+import FilledIcon from './filled.svg';
+
+type Props = {
+  filled?: boolean;
+  size?: Size;
+};
```
sshanzel, line 9: Tagging for visibility regarding the concern for shared interface.

packages/shared/src/components/icons/Upvote/index.tsx
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+import Icon, { Size } from '../../Icon';
+import OutlinedIcon from './outlined.svg';
+import FilledIcon from './filled.svg';
+
+type Props = {
+  filled?: boolean;
+  size?: Size;
+};
```
sshanzel, line 9: Tagging for visibility regarding the concern for shared interface.

packages/shared/src/components/icons/User/index.tsx
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+import Icon, { Size } from '../../Icon';
+import OutlinedIcon from './outlined.svg';
+import FilledIcon from './filled.svg';
+
+type Props = {
+  filled?: boolean;
+  size?: Size;
+  className: string;
+};
```
sshanzel, line 10: Tagging for visibility regarding the concern for shared interface.

sshanzel (changes requested):
Amazing work right there! Just had minor comments regarding conventions and practices with tailwindcss 🚢 🚢 🚢 

Apologies for the number of comments, but I just tagged the areas where it requires the same refactoring.

I just have one concern regarding the dedicated components for each kind of Icon (Bookmark, Upvote, etc.). I think it is great though it is probably over abstracting. I understand it has benefit of easier use, but with more files require more maintenance. I think we can just instead use the base component you created for those cases since they're only different with the source files. WDYT @rebelchris?

packages/shared/src/components/Icon.tsx
```diff
@@ -0,0 +1,40 @@
...
+  IconFilled: React.ComponentType<{ className }>;
+};
+
+const Icon: React.VFC<Props> = ({
+  filled = false,
+  size = 'small',
+  className = '',
+  IconOutlined,
+  IconFilled,
+}) => {
```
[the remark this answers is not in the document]
kirillkurko: Don't you think my way is more clear? There is no explicit return type and using React.VFC or React.FC you can mark is component accepts children or not. Plus you see this info earlier - no need to visually scroll through props if you just want to find props type

packages/shared/src/components/Icon.tsx
kirillkurko: Oh, of course, thanks

packages/shared/src/components/Icon.tsx
kirillkurko: Got it

packages/shared/src/components/Icon.tsx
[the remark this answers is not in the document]
sshanzel: In my personal projects, I actually use the same thing, so I totally agree with you (using React.FC). It has just become a convention throughout the codebase. Feel free to suggest it to the whole team in Slack I would suggest.

packages/shared/src/components/ProfileMenu.tsx
kirillkurko: Can you help me where I can check this?

kirillkurko:
> Amazing work right there! Just had minor comments regarding conventions and practices with tailwindcss 🚢 🚢 🚢
> 
> Apologies for the number of comments, but I just tagged the areas where it requires the same refactoring.
> 
> I just have one concern regarding the dedicated components for each kind of Icon (Bookmark, Upvote, etc.). I think it is great though it is probably over abstracting. I understand it has benefit of easier use, but with more files require more maintenance. I think we can just instead use the base component you created for those cases since they're only different with the source files. WDYT @rebelchris?

I also though about this, we have a bunch of files with my approach (what I don't like), but I've been initially thinking about imports, with my way you just need to import an icon and that's it, and with only one general icon we need three imports per icon (base component and two .svg files)

packages/shared/src/components/icons/Search/index.tsx
kirillkurko: Done

packages/shared/src/components/Icon.tsx
kirillkurko: Done

packages/shared/src/components/Icon.tsx
kirillkurko: Done

packages/shared/src/components/Icon.tsx
```diff
@@ -0,0 +1,42 @@
+import React from 'react';
+import classNames from 'classnames';
+
+export type Size = 'small' | 'medium' | 'large';
```
sshanzel, line 4: Since this is now an exported type, I think it would be better to be more explicit about it from `Size` to `IconSize`.
```suggestion
export type IconSize = 'small' | 'medium' | 'large';
```

rebelchris (changes requested):
Wow really nice work @kirillkurko 💖

I'm in two minds about the implementation around the singular extra files.
It would make things way easier for developer to find the icons and how they, but adding new icons might become a bit of a pain.

However quite ok with this for now if everyone agrees.

Some generic feedback:
Please go through all the existing icons, I think we should remove all the old ones and mark the ones we don't have replacement for.
Also not sure why the bookmark one is in a different folder?

Some files that still have old icons:
- https://github.com/dailydotdev/apps/blob/220f77a371b0e3c242549b15bd2f35e7e2348e5f/packages/shared/src/components/BookmarkEmptyScreen.tsx#L3
- https://github.com/dailydotdev/apps/blob/220f77a371b0e3c242549b15bd2f35e7e2348e5f/packages/shared/src/components/BookmarkFeedLayout.tsx#L16
- https://github.com/dailydotdev/apps/blob/220f77a371b0e3c242549b15bd2f35e7e2348e5f/packages/shared/src/components/CreateMyFeedButton.tsx#L4
- https://github.com/dailydotdev/apps/blob/220f77a371b0e3c242549b15bd2f35e7e2348e5f/packages/shared/src/components/CreateMyFeedButton.tsx#L5
- https://github.com/dailydotdev/apps/blob/220f77a371b0e3c242549b15bd2f35e7e2348e5f/packages/shared/src/components/FeedEmptyScreen.tsx#L9

These are some examples.
Perhaps the easiest way to do this is to go over each icon in the current `icons` folder, search for `iconName.svg`.
Replace the implementation and then remove the physical icon?

packages/shared/src/components/icons/Home/index.tsx
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+import Icon, { IconProps } from '../../Icon';
+import OutlinedIcon from './outlined.svg';
+import FilledIcon from './filled.svg';
+
+const HomeIcon: React.VFC<IconProps> = ({
```
sshanzel, line 6: Kindly check my concern regarding the `props` for easier maintenance. Tagging for visibility. Thank you!

packages/shared/src/components/icons/Hot/index.tsx
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
+import Icon, { IconProps } from '../../Icon';
+import OutlinedIcon from './outlined.svg';
+import FilledIcon from './filled.svg';
+
+const HotIcon: React.VFC<IconProps> = ({ filled = false }) => (
```
sshanzel, line 6: Kindly check my concern regarding the `props` for easier maintenance. Tagging for visibility. Thank you!

packages/shared/src/components/icons/Upvote/index.tsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import Icon, { IconProps } from '../../Icon';
+import OutlinedIcon from './outlined.svg';
+import FilledIcon from './filled.svg';
+
+const UpvoteIcon: React.VFC<IconProps> = ({ filled = false, size }) => (
```
sshanzel, line 6: Kindly check my concern regarding the `props` for easier maintenance. Tagging for visibility. Thank you!

sshanzel (changes requested):
Wow, you work very fast! And the refactoring is as clean as it can get! Well done! 🚢 

Just had some comments regarding maintenance work.

packages/shared/src/components/Icon.tsx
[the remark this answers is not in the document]
rebelchris: I don't know the exact differences between FC/VFC/our way, but I think FC/VFC have some issues with Preact and from what I read last on VFC is that it's planned to be deprecated and migrated to FC.
Which in return brings a lot more issues again.

I would personally leave everything to our conventions for now.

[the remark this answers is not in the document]
idoshamun: If it's going to be deprecated then we can put it aside. I suggest for now let's stick to our convention and log it for further research

sshanzel (changes requested):
One last thing about the convention we follow. Other than that, I think things are looking great!!! 🚢

docker-compose.yml
```diff
@@ -119,4 +119,4 @@ services:
       - COOKIES_KEY=topsecret
       - JWT_SECRET='|r+.2!!!.Qf_-|63*%.D'
       - JWT_AUDIENCE='Daily Staging'
-      - JWT_ISSUER='Daily API Staging'
```
rebelchris, line 122: Why was there a change here and in the `prettier.xml`?

packages/shared/src/components/Icon.tsx
```diff
@@ -0,0 +1,47 @@
+import React from 'react';
+import classNames from 'classnames';
+
+type Size = 'small' | 'medium' | 'large' | 'extralarge';
```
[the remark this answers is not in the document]
rebelchris: Agree with Lee here, that would make it uniform to how we address sizing at the moment.

rebelchris:
I also still see these icons where they not changed?
Perhaps ask Tsahi if they need to change or not.

import GitHubIcon from '@dailydotdev/shared/icons/github.svg';
import TwitterIcon from '@dailydotdev/shared/icons/twitter.svg';
import HashnodeIcon from '@dailydotdev/shared/icons/hashnode.svg';

docker-compose.yml
kirillkurko: This is one updated automatically (I've been running prettier for all files) and it just inserted a new line. Regarding prettier.xml - you can see this one inside` .idea` folder and it's used by JetBrains IDEs to keep configurations. Usually it's added to .gitgnore, but not in our case and by default prettier is not working on save action what I'm usually change - because of this prettier.xml file was updated and added to git changelist. We should probably add .idea folder to .gitignore, what do you think @rebelchris?

rebelchris: Ah perfect lets exclude the idea folder for now yes :)

rebelchris (approved):
Nice make sure to double check the companion if you pull the master in this one :)

sshanzel (approved):
I have to second what Chris have said. The new icons looks so refreshing! Wow. Just amazing. Thank you so much for all the effort. You did amazing! 🚢 🚢 🚢

PR #1003, opened 2022-06-07
fix: prevent comments from reloading to placeholders

rebelchris:
- Our companion request method was retuning `null` as a default, this causes react query to refresh to `null` then only to the actual data
- The fix introduced a optional object that can be returned, this optional object is the temporary previous object.

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-124 #done

### DIFF
--- a/packages/extension/src/companion/companionRequest.tsx
+++ b/packages/extension/src/companion/companionRequest.tsx
@@ -1,6 +1,6 @@
 import request from 'graphql-request';
 import { browser } from 'webextension-polyfill-ts';
-import { initialDataKey } from './common';
+import { initialDataKey } from '@dailydotdev/shared/src/lib/constants';
 
 const proxyRequest = {
   apply(_, __, args) {
--- a/packages/extension/src/companion/companionRequest.tsx
+++ b/packages/extension/src/companion/companionRequest.tsx
@@ -1,10 +1,10 @@
 import request from 'graphql-request';
 import { browser } from 'webextension-polyfill-ts';
+import { initialDataKey } from './common';
 
 const proxyRequest = {
   apply(_, __, args) {
-    const initialCol = 'initial';
-    const { [initialCol]: initial, ...variables } = args?.[2];
+    const { [initialDataKey]: initial, ...variables } = args?.[2];
 
     browser.runtime.sendMessage({
       type: 'GRAPHQL_REQUEST',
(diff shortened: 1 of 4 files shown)


packages/extension/src/companion/companionRequest.tsx
```diff
@@ -3,14 +3,18 @@ import { browser } from 'webextension-polyfill-ts';
 
 const proxyRequest = {
   apply(_, __, args) {
+    const initialCol = 'initial';
```
sshanzel, line 6: I think we should take this outside the function and export it, so the reference will always be the same, even if we decide to change keys in the future, we will be safe. What do you think?

sshanzel (approved):
Just had one comment, but it is so nice to see how our functions grow to support customizations. Love the fix you have here! 🚢

packages/extension/src/companion/companionRequest.tsx
sshanzel: Oh, yes. I totally forgot about that. Probably in one of our common files? Not exactly sure where the specific one would fit though.

sshanzel: Oh. Really sorry. I meant to be placed in the shared package as well, so the key in the actual request would be able to utilize it as well. Sorry for the inconvenience 🙇

rebelchris: Oh wait let me refactor I totally got your idea wrong 🤯

packages/shared/src/components/post/PostComments.tsx
```diff
@@ -92,7 +92,7 @@ export function PostComments({
         requestMethod(
           `${apiUrl}/graphql`,
           POST_COMMENTS_QUERY,
-          { postId: id },
+          { postId: id, initial: comments },
```
sshanzel, line 95: @rebelchris I meant for this case as well. So we can always be sure that the keys are matched.
```suggestion
          { postId: id, [initialDataKey]: comments },
```

sshanzel, line 95: @rebelchris I meant for this case as well. So we can always be sure that the keys are matched. Probably just a personal preference, but we can let this go as it is not quite critical. Just for typos.
```suggestion
          { postId: id, [initialDataKey]: comments },
```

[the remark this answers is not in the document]
rebelchris: Yep, misunderstood you in the first comment, fixed it now :)

packages/extension/src/companion/companionRequest.tsx
sshanzel: Thank you! Sorry for lacking clarity! 🚢

PR #1005, opened 2022-06-08
fix: companion missing at some point

sshanzel:
## Changes
One of the probable causes why the companion goes missing "at some point", as mentioned by users, is they have probably clicked a header or navigated the ToC that made them target a div (adds `#container` in the URL). Hence, it goes missing as we had an issue with URLs on those criteria.

- The `href` property of a URL object still returns the link with `#` character. So we concatenated the origin and pathname.
image

Quick links:
Sample:
Working since then: https://www.joshwcomeau.com/css/surprising-truth-about-pixels-and-accessibility
Not working before the fix: https://www.joshwcomeau.com/css/surprising-truth-about-pixels-and-accessibility/#accessibility-considerations
Kindly try to navigate in the ToC as well if you want.

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-120 #done

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -8,6 +8,11 @@ volumes:
 services:
   daily-redis:
     image: redis:5.0.10
+    healthcheck:
+      test: redis-cli ping
+      interval: 1s
+      timeout: 3s
+      retries: 60
     command: redis-server --appendonly yes
     volumes:
       - redis:/data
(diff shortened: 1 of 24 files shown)


rebelchris (approved):
Looks good, could you just verify this also works in FF.
They have different ways of using the browser polyfil.

sshanzel:
Yes, actually that was the reason why I switched to using `Tabs.Tab`. In FF, they don't recognize the `Runtime.MessageSender.origin`. After further testing, things seem to still work as intended! 🚢

idoshamun (approved):
Fabulous!

PR #1016, opened 2022-06-12
refactor: feed post card layout

sshanzel:
# Post Card V2

- Created a custom context provider for the FeaturesContext as it is quite verbose to pull a single value for AB testing. We're already calling the context for pulling the flags, so I thought of why not embed to easily manage and centralize the value (helpful in development).
- The custom context provider is backward compatible as I kept the value of the flags.
- Removed the featured comments feature.
- Removed the comment popup from the card when a user has upvoted.
- AB Test the new version of Post Card.
- AB Test the behavior of clicking the card itself to open the Post Modal by default.
- AB Test the engagement elements in the Post Card.

Tip: most of the changes revolved around the `ActionsButton` as the whole layout kind of has still large resemblance from one another.

The flagsmith parameters are not yet created, but this PR is currently deployed at: https://preview2.app.daily.dev/

_Have verified locally all the variation including on all different screen sizes_

- [x] MobileL (420px)
- [x] Tablet (656px)
- [x] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-51 #done


sshanzel:
@idoshamun tagging you as a reviewer, just to see if you have anything that must be changed with tracking but feel free to ignore if there is none.

packages/shared/src/components/cards/PostCard.tsx
```diff
@@ -67,89 +65,108 @@ export const PostCard = forwardRef(function PostCard(
...
+          />
+        )}
+        {!showImage && (
+          <PostAuthor post={post} className="hidden tablet:flex mx-4 mt-2" />
+        )}
+        {showImage && (
+          <CardImage
+            imgAlt="Post Cover image"
+            imgSrc={post.image}
+            fallbackSrc="https://res.cloudinary.com/daily-now/image/upload/f_auto/v1/placeholders/1"
```
rebelchris, line 138: Random question, do we want to keep this with the lazy loading or move to the native one?

packages/shared/src/components/cards/SourceButton.tsx
```diff
@@ -1,33 +1,35 @@
...
       <a className={classNames('flex cursor-pointer', className)} style={style}>
         <img
-          src={post.source.image}
-          alt={post.source.name}
-          className="w-6 h-6 rounded-full bg-theme-bg-tertiary"
+          src={source.image}
+          alt={source.name}
+          className="w-8 h-8 rounded-full bg-theme-bg-tertiary"
         />
       </a>
```
rebelchris, line 32: I think we can actually use the ProfilePicture component for this, we use it in other places for the a rounded source as well.

rebelchris (changes requested):
Nice, love the context you used for the flags, do you think we should move all flags to that use-case or only the `isFeaturedEnabled` one?

Furthermore I did a manual test on all variants and we're almost there!

- Clicking article in modal mode adds for rank (check with product)
- We can safely remove the hide publication date flag
- We can safely remove the complete post font flag
- List options missing source? (control list)
Screenshot 2022-06-17 at 14 46 09
- List option menu alignment is wrong:
Screenshot 2022-06-17 at 14 46 09
- Static icons should be smaller?  (I think, the design is blocked for me)
- The static buttons have some weird hover:
Screenshot 2022-06-17 at 14 41 29
Screenshot 2022-06-17 at 14 44 02

packages/shared/src/components/Feed.tsx
```diff
@@ -114,15 +113,8 @@ export default function Feed<T>({
...
-  const { flags } = useContext(FeaturesContext);
-  const postHeadingFont = getThemeFont(
-    getFeatureValue(Features.PostCardHeadingFont, flags),
-    Features.PostCardHeadingFont.defaultValue,
-  );
-  const displayPublicationDate = !parseInt(
-    getFeatureValue(Features.HidePublicationDate, flags),
-    10,
-  );
+  const { postModalByDefault, postHeadingFont, displayPublicationDate } =
```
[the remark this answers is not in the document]
rebelchris: I already removed them in another PR https://github.com/dailydotdev/apps/pull/1036

sshanzel:
> Nice, love the context you used for the flags, do you think we should move all flags to that use-case or only the isFeaturedEnabled one?

Thank you! I think we should move all, I just didn't decide to do it here as the PR might get bloated. Was thinking of having it in a separate PR. WDYT?

For the missing source, I think it should be the screenshot you have included, right?
image

packages/shared/src/components/cards/PostCard.tsx
sshanzel: Ah, yes. It is our mission to move everything to the native one, let's refactor this as well. Thank you!

packages/shared/src/components/cards/ActionButtons.tsx
```diff
@@ -11,86 +11,160 @@ import CommentIcon from '../icons/Discuss';
...
+  onMenuClick,
   onShare,
   className,
   children,
+  insaneMode,
 }: ActionButtonsProps): ReactElement {
+  const { postCardVersion, postEngagementNonClickable, postModalByDefault } =
+    useContext(FeaturesContext);
+  const isV2 = postCardVersion === 'v2';
+  const buttonStyles = postEngagementNonClickable ? {} : { width: rem(78) };
```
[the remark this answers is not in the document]
sshanzel: This was the value to the styling of the component prior to the PR, should we move it to explicitly stating the actual computed rem value? WDYT?

packages/shared/src/components/cards/SourceButton.tsx
sshanzel: Totally agree. Let me work this out 🚢

packages/shared/src/components/cards/ActionButtons.tsx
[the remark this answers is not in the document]
rebelchris: Hmm, yeah not sure where this comes from, seems out of line with what we have.
Would love it if you can make it uniform.

rebelchris:
@sshanzel Yeah it works for most possibilities. but this one you can see we don't have the sources.
Screenshot 2022-06-17 at 14 46 09

sshanzel:
> @sshanzel Yeah it works for most possibilities. but this one you can see we don't have the sources.

Oh, wow. Do you have any steps to reproduce this? Or this just happens when you reload?

packages/shared/src/components/cards/ActionButtons.tsx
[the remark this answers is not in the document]
sshanzel: It was mostly defined as an inline styling before:

https://github.com/dailydotdev/apps/pull/1016/files#diff-05bb8102e25658f2248bef0feee962e1774f95603cb144f2a4652497e0bdd94aL53

Sure, let's make it utilize arbitrary values if it works. Thank you!

sshanzel:
I haven't been able to reproduce the missing source part of the list mode. Though I think it was due to the switching of features context value as I noticed from the screenshot, the font weight of the title is not bold (resets when hitting save, I noticed to happen when in development mode).

With that, let me know if the issue persists on a fresh load. I have also updated the list of properties in the features context based on the last cleanup you've made. Thank you!

packages/shared/src/components/cards/ActionButtons.tsx
```diff
@@ -1,96 +1,170 @@
...
+    className: classNames(
+      'btn-tertiary-avocado',
+      !postEngagementNonClickable && 'w-[4.875rem]',
+    ),
+  };
+
+  const bookmarkButton = (
+    <SimpleTooltip content={post.bookmarked ? 'Remove bookmark' : 'Bookmark'}>
+      <Button
+        icon={<BookmarkIcon filled={post.bookmarked} size="medium" />}
```
rebelchris, line 77: You no longer need to provide sizes for icons inside buttons 🎉

rebelchris (changes requested):
Wow amazing, love how you uniformed the Image component.
That is a great addition to our "native" components.

I see the source correct now, however the options button sits to far to the left in list mode.
It should be sitting at the far right.
Screenshot 2022-06-21 at 16 41 17

Otherwise we are good to go I think 🚀

sshanzel:
The placement of the actions should correct in control mode, I think.

image

rebelchris (approved):
Nice, looks solid!
Super hyped to see this in action Lee 🤯

packages/shared/src/components/cards/PostCard.tsx
```diff
@@ -66,86 +65,109 @@ export const PostCard = forwardRef(function PostCard(
...
+  const customStyle = !showImage ? { minHeight: '15.125rem' } : {};
   const card = (
     <Card
       {...props}
-      className={getPostClassNames(post, selectedComment, className)}
+      className={getPostClassNames(post, className)}
       style={{ ...style, ...customStyle }}
       ref={ref}
     >
       <PostLink post={post} openNewTab={openNewTab} onLinkClick={onLinkClick} />
```
idoshamun, line 95: We should use `PostLink` only when we don't open the article modal by default. Otherwise, there's no need to have background anchor element. The only button that opens that article and it's an actual link is the "read article" button

packages/shared/src/components/image/Image.tsx
```diff
@@ -0,0 +1,17 @@
+import React, { ImgHTMLAttributes, ReactElement, SyntheticEvent } from 'react';
+
+interface ImageProps extends ImgHTMLAttributes<HTMLImageElement> {
+  fallbackSrc?: string;
+}
+
+export const Image = ({ fallbackSrc, ...props }: ImageProps): ReactElement => {
```
idoshamun, line 7: Why do we use a new `Image` element?

packages/shared/src/components/cards/PostCard.tsx
sshanzel: Yes it is for opening a link, though it can be reused to instead make the whole card clickable. But we can make a separate element for it if necessary. Or make it a button element instead if the user has post by modal.

idoshamun: Yes having a link in the background just to have something clickable is not the right thing from a semantic HTML perspective. You can use a button as you mentioned

packages/shared/src/components/FeedItemComponent.tsx
```diff
@@ -138,10 +116,13 @@ export default function FeedItemComponent({
           onUpvoteClick={(post, upvoted) =>
             onUpvote(post, index, row, column, upvoted)
           }
-          onLinkClick={(post) => onPostClick(post, index, row, column)}
+          onLinkClick={(post, e) => onPostClick(post, index, row, column, e)}
           onBookmarkClick={(post, bookmarked) =>
             onBookmark(post, index, row, column, bookmarked)
           }
+          onReadArticleClick={(e) =>
+            onPostClick(item.post, index, row, column, e, true)
```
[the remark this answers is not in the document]
idoshamun: I'm not sure we want to track it as a view at the moment. But let me get back to you on that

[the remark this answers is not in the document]
idoshamun: We want to flag an article as read-only when the user actually visits the article

packages/shared/src/components/image/Image.tsx
sshanzel: It was discussed that we will be moving forward in utilizing the native lazy loading.

packages/shared/src/components/cards/PostCard.tsx
sshanzel: I agree. Let's turn this one into a button only when necessary.

packages/shared/src/components/FeedItemComponent.tsx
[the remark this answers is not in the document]
sshanzel: Oh, cool. Then this should count as read since they will be redirected to the actual article, right?

With that, I will label the event based on our taxonomy as 'go to link' but register the `view` that the user has made. WDYT?

packages/shared/src/components/Feed.tsx
```diff
@@ -197,6 +193,23 @@ export default function Feed<T>({
...
+    column,
+    event,
+    fromReadArticle,
+  ) => {
+    if (!postModalByDefault || fromReadArticle) {
+      await onPostClick(post, index, row, column);
+      return;
+    }
+
+    event.preventDefault();
```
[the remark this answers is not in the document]
sshanzel: It was due to the issue in utilizing the anchor element as the clickable one but since we are replacing it now with a button, we will be removing this part.

packages/shared/src/components/FeedItemComponent.tsx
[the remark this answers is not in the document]
idoshamun: By "register as view", what do you mean exactly? Because views are automatically counted once you visit the redirector link. Do you refer to the local weekly badge count?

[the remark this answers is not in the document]
sshanzel: Yes, I mean we add an entry in our database at `view` table when the post link is opened (clicked the card), and with the `ReadArticle` button, it opens the link (but not through the whole card click anymore), so I am thinking it should perform the same things and marks the post as already read by the user.

I think the read article button inside post modal button doesn't create a new entry in our database whenever it is clicked though it contains all the parameters for tracking the post. That's why I am wondering if we should create an entry on the new card read article button, and the post modal's read article button should behave the same way?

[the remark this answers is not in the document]
sshanzel: Or wait. I think I am actually mistaken. It seems we register view in the redirector, and we just locally update the cache about clicking a post card link.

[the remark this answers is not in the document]
idoshamun: You're right. The redirector counts automatically the views and it's just the local cache that you update on article view

idoshamun:
@sshanzel let me know when it's ready for another round

sshanzel:
@idoshamun should now be ready for review! 🎉 

I have returned but commented out the comment popup feature as recommended. As for tracking, I applied the appropriate event for clicking the modal and for clicking the read article button. Kindly check the parameter I added to understand whether the post by modal is enabled or not.

For the API part. I think we should merge this first and ensure the extension won't request the query for featured comments anymore before the removal in the backend so we won't encounter any error in the frontend. Or do we want to keep the query properties but remove the cron job responsible for it?

idoshamun:
@sshanzel we should proceed with the latter. We don't want to introduce breaking changes. Even when we merge this, there will still be clients running old version

packages/shared/src/components/cards/PostButton.tsx
```diff
@@ -0,0 +1,20 @@
...
+
+export default function PostButton({
+  title,
+  onClick,
+}: PostButtonProps): ReactElement {
+  return (
+    <CardButton
+      title={title}
+      onClick={onClick}
+      onMouseUp={(event) => event.button === 1 && onClick?.(event)}
```
idoshamun, line 17: If it is used only for real buttons, we should not support middle click. The middle click here supports links

packages/shared/src/components/cards/PostCard.tsx
```diff
@@ -83,6 +85,23 @@ export const PostCard = forwardRef(function PostCard(
       ),
     [postCardVersion],
   );
+
+  const getClickablePost = () => {
+    const onPostCardClick = () => onPostClick(post);
```
idoshamun, line 90: This can go outside this inner function

packages/shared/src/components/cards/PostCard.tsx
```diff
@@ -83,6 +85,23 @@ export const PostCard = forwardRef(function PostCard(
       ),
     [postCardVersion],
   );
+
+  const getClickablePost = () => {
```
idoshamun, line 89: It doesn't have to be a function. This can be a variable that stores the right elements

packages/shared/src/components/cards/PostLink.tsx
```diff
@@ -1,27 +1,28 @@
...
-      href={post.permalink}
+      href={href}
+      title={title}
       {...(openNewTab
         ? { target: '_blank', rel: 'noopener' }
         : { target: '_self' })}
-      title={post.title}
-      onClick={(e) => onLinkClick?.(post, e)}
-      onMouseUp={(event) => event.button === 1 && onLinkClick?.(post)}
+      onClick={(e) => onLinkClick?.(e)}
```
idoshamun, line 24: This should be fine

```suggestion
      onClick={onLinkClick}
```

packages/shared/src/components/cards/PostList.tsx
```diff
@@ -42,25 +36,43 @@ export const PostList = forwardRef(function PostList(
...
 ): ReactElement {
   const { trending } = post;
-  const { postCardVersion } = useContext(FeaturesContext);
   const isV1 = postCardVersion === 'v1';
   const isV2 = postCardVersion === 'v2';
   const ActionsContainer = isV2
     ? useMemo(() => classed('div', 'flex flex-row items-center w-full'), [isV2])
     : React.Fragment;
 
+  const getClickablePost = () => {
```
idoshamun, line 53: Same comments as for the card

packages/shared/src/hooks/feed/useFeedOnPostClick.ts
```diff
@@ -20,19 +24,26 @@ export default function useFeedOnPostClick(
...
-  return async (post, index, row, column): Promise<void> => {
+  return async (post, index, row, column, optional): Promise<void> => {
     trackEvent(
-      postAnalyticsEvent('click', post, {
+      postAnalyticsEvent(event, post, {
         columns,
         column,
         row,
-        ...feedAnalyticsExtra(feedName, ranking),
+        ...feedAnalyticsExtra(feedName, ranking, { modal: postModalByDefault }),
```
idoshamun, line 39: This additional param isn't needed at the moment. But thanks for thinking about it :)

idoshamun (approved):
Please double check everything and also make sure the right events are being tracked.
It's a very sensitive PR

PR #1017, opened 2022-06-13
fix: outline icons spacing and size

kirillkurko:

Few icons sizes and spacings were updated (found inconsistencies during testing)

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

DD-{number} #done

### DIFF
--- a/packages/shared/src/components/BookmarkFeedLayout.tsx
+++ b/packages/shared/src/components/BookmarkFeedLayout.tsx
@@ -63,7 +63,7 @@ export default function BookmarkFeedLayout({
   const shareBookmarksButton = (style: string, text?: string) => (
     <Button
       className={style}
-      icon={<SourceIcon />}
+      icon={<SourceIcon className="icon" />}
       onClick={() => setShowSharedBookmarks(true)}
     >
       {text}
--- a/packages/shared/src/components/widgets/PostUsersHighlights.tsx
+++ b/packages/shared/src/components/widgets/PostUsersHighlights.tsx
@@ -87,10 +87,10 @@ const UserHighlight = (props: SourceAuthorProps) => {
       {Icon && (
         <Icon
           className={classNames(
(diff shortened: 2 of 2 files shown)


packages/shared/src/components/BookmarkFeedLayout.tsx
```diff
@@ -63,7 +63,7 @@ export default function BookmarkFeedLayout({
   const shareBookmarksButton = (style: string, text?: string) => (
     <Button
       className={style}
-      icon={<SourceIcon />}
+      icon={<SourceIcon className="icon" />}
```
sshanzel, line 66: Can you check what are the styles applied for using the icon class? I think this doesn't do anything anymore but I could be wrong.

sshanzel (approved):
Just some things that needs checking but looks good to me overall! ⛴️

packages/shared/src/components/widgets/PostUsersHighlights.tsx
```diff
@@ -87,10 +87,10 @@ const UserHighlight = (props: SourceAuthorProps) => {
       {Icon && (
         <Icon
           className={classNames(
-            'absolute w-5 h-5 top-11 left-11',
+            'absolute w-5 h-5 top-10 left-10',
```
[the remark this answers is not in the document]
kirillkurko: Yes, I checked both icons before making this change

packages/shared/src/components/BookmarkFeedLayout.tsx
kirillkurko: I do not remember exactly, but it applies some font-size (doesn't really matter it this case) and some margin, so icon is not so close to the text

rebelchris (approved):
Looks good to me, we can evaluate the icon class later, I think only buttons use it now.

PR #1021, opened 2022-06-14
fix: font rendered different on some sites

rebelchris:
- We adapted font-rendering from each website, but we wanted to render smoothly
- This PR introduces the same font rendering for all websites so we uniform our companion fonts.

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-108 #done

### DIFF
--- a/packages/extension/public/css/daily-companion-app.css
+++ b/packages/extension/public/css/daily-companion-app.css
@@ -8,4 +8,7 @@ daily-companion-app {
   position: fixed;
   z-index: 2147483647;
   font-feature-settings: normal!important;
+  text-rendering: optimizeLegibility;
+  -webkit-font-smoothing: antialiased;
+  -moz-osx-font-smoothing: grayscale;
 }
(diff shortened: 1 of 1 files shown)


packages/extension/public/css/daily-companion-app.css
```diff
@@ -8,4 +8,7 @@ daily-companion-app {
   position: fixed;
   z-index: 2147483647;
   font-feature-settings: normal!important;
```
kirillkurko, line 10: I'm going to work as a prettier here - let's insert a whitespace before !important  😀

kirillkurko (approved):
Just a minor comment

packages/extension/public/css/daily-companion-app.css
rebelchris: I think both variants are valid, but added a space 👍

sshanzel (approved):
Whoa! So nice you found the fix for this! 🚢 🚢 🚢

PR #1022, opened 2022-06-14
Wt 140 companion icons

rebelchris:
- The old icon class was 24px, this one renders 28px to the new icon spec.
- Currently only used by the companion buttons anyway (actively, as it's a button > .icon selector)

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-140 #done

### DIFF
--- a/packages/extension/public/css/daily-companion-app.css
+++ b/packages/extension/public/css/daily-companion-app.css
@@ -8,7 +8,4 @@ daily-companion-app {
   position: fixed;
   z-index: 2147483647;
   font-feature-settings: normal!important;
-  text-rendering: optimizeLegibility;
-  -webkit-font-smoothing: antialiased;
-  -moz-osx-font-smoothing: grayscale;
 }
--- a/packages/extension/public/css/daily-companion-app.css
+++ b/packages/extension/public/css/daily-companion-app.css
@@ -8,4 +8,7 @@ daily-companion-app {
   position: fixed;
   z-index: 2147483647;
   font-feature-settings: normal!important;
+  text-rendering: optimizeLegibility;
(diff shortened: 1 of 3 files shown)


packages/shared/src/styles/components/buttons.css
```diff
@@ -10,7 +10,7 @@
   & .icon {
     width: 1em;
     height: 1em;
-    font-size: 1.5rem;
+    font-size: 1.75rem;
```
kirillkurko, line 13: So this change is going to affect other places where this class is used and they will be bigger, right? Maybe we can remove className from the icons above and use on the available sizes (size prop)

packages/extension/public/css/daily-companion-app.css
```diff
@@ -8,4 +8,7 @@ daily-companion-app {
   position: fixed;
   z-index: 2147483647;
   font-feature-settings: normal!important;
+  text-rendering: optimizeLegibility;
```
kirillkurko, line 11: You already have these changes in other PR, but I will repeat my comment - let's insert space before !important

packages/shared/src/styles/components/buttons.css
rebelchris: The icon class is only valid for icons inside buttons.
The other ones it doesn't affect too. But we might want to re-evaluate the button icons globally (see tsahi comment: https://dailydotdev.slack.com/archives/C01L0QXQJNB/p1655216371347519?thread_ts=1655214705.501559&cid=C01L0QXQJNB)

PR #1028, opened 2022-06-15
feat: introducing scout tracking and updated post author tracking

rebelchris:
- We didn't include post author in every graphQL query
- Added new `post_scout_id` parameter
- Fixed some scout elements on the comments

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-144 #done

### DIFF
--- a/packages/extension/package-lock.json
+++ b/packages/extension/package-lock.json
@@ -1,6 +1,6 @@
 {
 	"name": "extension",
-	"version": "3.16.3",
+	"version": "3.16.4",
 	"lockfileVersion": 1,
 	"requires": true,
 	"dependencies": {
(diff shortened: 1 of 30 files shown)


rebelchris:
@idoshamun Tagged you from a analytics perspective only :)

kirillkurko (approved):
Looks good, but what I see outside of this task is that that MainComment and SubComment components are really similar, so maybe we can create a task to implement kinda general comment component that can be used for replacing both mentioned component

rebelchris:
@sshanzel We where missing the id from the author query, so you see that field being added.
Also one new refactor in the companion boot data to simply include `author` and `scout` completly.

@kirillkurko Great idea, I'll add it to the backlog :)

sshanzel:
@rebelchris ah, saw it now! Thank you! Think I'm going blind 😆

PR #1029, opened 2022-06-15
refactor: button icon sizes inline with the new icon guideline

kirillkurko:

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

DD-{number} #done


packages/shared/src/components/post/AuthorOnboarding.tsx
```diff
@@ -27,7 +27,7 @@ export function AuthorOnboarding({
         )}
         style={{ gridTemplateColumns: 'repeat(2, max-content)' }}
       >
-        <FeatherIcon />
+        <FeatherIcon size="xlarge" className="text-theme-status-help icon" />
```
rebelchris, line 30: I think this one needs to be filled right?

kirillkurko: This PR is still in draft, but I will reach to Tsahi about guideline for filled icons and double check all of them

packages/shared/src/components/ShareBar.tsx
```diff
@@ -56,7 +56,7 @@ export default function ShareBar({ post }: { post: Post }): ReactElement {
             target="_blank"
             rel="noopener"
             onClick={() => onClick('whatsapp')}
-            icon={<WhatsappIcon />}
+            icon={<WhatsappIcon style={{ fontSize: '1.75rem' }} />}
```
rebelchris, line 59: Can't we set a size for this as well?

rebelchris (changes requested):
I think we might miss some implementations.
For example on the companion you set them to `medium` which is 24px right?
It should be 28px for button icons.

That's why I suggested perhaps we can have a special icon inside a button class that could be determined by the button wrapper?

Let me know what you think.

packages/shared/src/components/ShareBar.tsx
kirillkurko: This icon is not a component, but a simple .svg and I think we can keep it as it is

kirillkurko:
> I think we might miss some implementations. For example on the companion you set them to `medium` which is 24px right? It should be 28px for button icons.
> 
> That's why I suggested perhaps we can have a special icon inside a button class that could be determined by the button wrapper?
> 
> Let me know what you think.

"medium" is for "More options" menu because button size here is 32px, so 24px for icon. And upvote, comment, bookmark and menu icon sizes are set to "large" - 40px (CompanionMenu.tsx)

packages/shared/src/components/Icon.tsx
```diff
@@ -37,6 +38,7 @@ const Icon = ({
   return (
     <IconComponent
       className={classNames(IconSize[size], 'pointer-events-none', className)}
+      style={size === 'large' ? { fontSize: '1.75rem' } : null}
```
rebelchris, line 41: This piece i'm referring to.

What happen to medium sized buttons with a icon now, they would be not the correct size.

Also this way of adding it seems a bit hacky to me.

In my opinion we should add something in the `<Button />` component (our own one) and handle icon sizes based on the button size.

They will be different then the regular icon size in some case:
XSmall button => Small icon
Small button => Medium icon
Medium button => new w-7 h-7 map (or the large if we don't use it otherwise)
Large button => Xlarge icon

My idea is to remove the struggle for the developer to even have to think about the size, the button should handle it.
Let me know if that makes sense?

kirillkurko: Why medium sized buttons will have wrong icon size?

Agree that way of adding is tricky, because there is no w-7 h-7 (so we need to add it).

We can make button component handle icon size, but we still have a lot of button-looking links where we need to care about icons size

packages/shared/src/components/Icon.tsx
```diff
@@ -1,13 +1,14 @@
...
 
-type Size = 'small' | 'medium' | 'large' | 'xlarge';
+type Size = 'small' | 'medium' | 'large' | 'xlarge' | 'xxlarge';
 
 const IconSize: Record<Size, string> = {
   small: 'w-5 h-5',
   medium: 'w-6 h-6',
-  large: 'w-8 h-8',
-  xlarge: 'w-10 h-10',
+  large: 'w-auto w-auto',
```
sshanzel, line 9: I'm not sure being large has auto values. It may happen where it can even be bigger than the `xlarge`. I think you should introduce one more size being named `auto`.

packages/shared/src/components/Icon.tsx
sshanzel, line 41: I think it is confusing to have auto values but then set a specific styling value. We need to make it consistent across different sizes.

sshanzel (changes requested):
Just had a few comments around the consistency of our `size` property but the PR looks amazing! Well done on keeping the icons accurate as how it was design!

Also, what is the default value for the `size` property?

packages/shared/src/components/Icon.tsx
kirillkurko: Yeah, it was a hack way of adding a new size just because h-7 w-7 wasn't working for me (next.js cache 😅), now it's good

packages/shared/src/components/Icon.tsx
kirillkurko: It was a part of hacky way of adding large size (complete style was w-auto h-auto and fontSize: 1.75rem), but now we have w-7 h-7

packages/webapp/components/CookieBanner.tsx
```diff
@@ -40,7 +40,7 @@ export default function CookieBanner({
       <ModalCloseButton onClick={close} />
       <CookieIcon
         className="hidden laptop:block mb-4 text-theme-label-primary"
-        style={{ fontSize: '3.5rem' }}
+        style={{ fontSize: '3.5rem', width: 'auto', height: 'auto' }}
```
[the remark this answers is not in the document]
kirillkurko: Now we don't have it, and it's a little bit different thing - if we want to go with a custom size through fontSize we need to reset width and height, because they are always set inside icon with some tailwind classes

packages/shared/src/components/Icon.tsx
sshanzel: Ah, yes. We have this ongoing issue with nextjs - we're hoping to solve it soon enough. If in case the thing we're working on is applicable in extension as well, personally I would suggest testing things in there so you won't have to clear cache and re-run every time.

packages/webapp/components/CookieBanner.tsx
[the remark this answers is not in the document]
sshanzel: I would suggest adding a size property that adds these auto values so we can be consistent with how we set the width and height.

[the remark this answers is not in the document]
kirillkurko: What about xxxlarge size prop for this one? It will be w-14 h-14? So no fontSize at all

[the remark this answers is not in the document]
sshanzel: That can work, if we it is deemed to be the actual next size. If not I would suggest just override the height and width. But long as we are still using the height and width, I think we are consistent enough.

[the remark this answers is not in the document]
kirillkurko: Really this size will be too specific and used only in one place, so let's go with auto for now

packages/shared/src/components/ShareBar.tsx
sshanzel, line 59: Why are we not using our pre-defined sizes?
```suggestion
            icon={<WhatsappIcon size="large" />}
```

packages/shared/src/components/ShareBar.tsx
```diff
@@ -67,7 +67,7 @@ export default function ShareBar({ post }: { post: Post }): ReactElement {
             target="_blank"
             rel="noopener"
             onClick={() => onClick('twitter')}
-            icon={<TwitterIcon />}
+            icon={<TwitterIcon style={{ fontSize: '1.75rem' }} />}
```
sshanzel, line 70: Why are we not using our pre-defined sizes?

packages/shared/src/components/ShareBar.tsx
```diff
@@ -78,7 +78,7 @@ export default function ShareBar({ post }: { post: Post }): ReactElement {
             target="_blank"
             rel="noopener"
             onClick={() => onClick('facebook')}
-            icon={<FacebookIcon />}
+            icon={<FacebookIcon style={{ fontSize: '1.75rem' }} />}
```
sshanzel, line 81: Why are we not using our pre-defined sizes?

packages/shared/src/components/buttons/Button.tsx
```diff
@@ -45,7 +62,7 @@ function ButtonComponent<TagName extends AllowedTags>(
     pressed,
     icon,
     rightIcon,
-    buttonSize,
+    buttonSize = 'medium',
```
sshanzel, line 65: What is the size if this is left empty? Be careful with changing the default value as there are a lot of areas where this component was used.

packages/webapp/components/CookieBanner.tsx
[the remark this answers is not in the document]
rebelchris: I would prefer fixed sizes as well, if we don't have them we should add them.
I think the xxxlarge can work, I added that recently for a similar approach.

packages/webapp/components/CookieBanner.tsx
```diff
@@ -39,6 +39,7 @@ export default function CookieBanner({
     >
       <ModalCloseButton onClick={close} />
       <CookieIcon
+        size="auto"
         className="hidden laptop:block mb-4 text-theme-label-primary"
         style={{ fontSize: '3.5rem' }}
```
sshanzel, line 44: One of the reasons we did the refactoring of icons was to be consistent with how we implemented the icons. If we're gonna keep using font-size for changing its actual size, incoming developers would have doubts about which one to use.

We have moved all other implementations to use width and height. So kindly change the style property here to override the width and height style.

sshanzel (changes requested):
Just one question with the button and the others were just regarding consistency. Other than that, things look neat to me! 🚢

packages/webapp/components/CookieBanner.tsx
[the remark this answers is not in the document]
kirillkurko: Done

packages/shared/src/components/ShareBar.tsx
kirillkurko: Already discussed with Chris - these icons are some old .svg and not components I created, so there is no size prop. We can create components in the same way I created them for other icons, but I don't think we need it, because it will be just a wrapper on a single .svg (no filled/outlined)

packages/webapp/components/CookieBanner.tsx
```diff
@@ -39,6 +39,7 @@ export default function CookieBanner({
...
       />
```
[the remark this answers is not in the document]
kirillkurko: Done

packages/shared/src/components/ShareBar.tsx
sshanzel: Alrighty! Thanks for the notes!

packages/shared/src/components/buttons/Button.tsx
kirillkurko: It always was medium, but wasn't explicitly specified in props specified, but it can be figured out from button styles

packages/shared/src/components/CreateMyFeedButton.tsx
```diff
@@ -65,7 +65,7 @@ export default function CreateMyFeedButton({
             buttonColor.button,
           )}
           buttonSize="small"
-          icon={<PlusIcon />}
+          icon={<PlusIcon className="icon" />}
```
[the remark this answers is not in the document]
rebelchris: Would you mind doing one last check if we use the icon class anywhere where it's not applicable?
I want to take this opportunity to clean up some legacy stuff.

packages/shared/src/components/buttons/Button.tsx
sshanzel: Nice! This should be good to go then 🚢

rebelchris (approved):
Besides my main `icon` comment we should be good from my side.

kirillkurko:
> Besides my main `icon` comment we should be good from my side.



> Besides my main `icon` comment we should be good from my side.



> Besides my main `icon` comment we should be good from my side.

@rebelchris what I see is that icon class is required in all places being used, but I was able to clean it up for buttons and now button component is responsible for adding it

PR #1036, opened 2022-06-17
fix: cleanup unused and finished experiments

rebelchris:
- Cleanup unused experiments
- Text related ones remain active, but changed to winners
- Logic related ones get completely removed

Note: We should only turn them off once we deploy the extension to a new version and give it a week or so.

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

DD-{number} #done

### DIFF
--- a/packages/shared/src/components/FeedItemComponent.tsx
+++ b/packages/shared/src/components/FeedItemComponent.tsx
@@ -23,11 +23,9 @@ export type FeedItemComponentProps = {
   useList: boolean;
   openNewTab: boolean;
   insaneMode: boolean;
-  displayPublicationDate: boolean;
   nativeShareSupport: boolean;
   postMenuIndex: number | undefined;
   showCommentPopupId: string | undefined;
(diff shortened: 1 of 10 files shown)


packages/shared/src/components/FeedItemComponent.tsx
```diff
@@ -137,7 +133,7 @@ export default function FeedItemComponent({
           ref={inViewRef}
           post={{
             ...item.post,
-            createdAt: displayPublicationDate && item.post.createdAt,
+            createdAt: item.post.createdAt,
```
sshanzel, line 136: I think we can just pass `item.post` here and no need for destructuring since there is no modification with the data anymore.

sshanzel (approved):
So many deleted codes ❤️ 

Just had one non-blocking comment, but all looks amazing! 🚢

PR #1039, opened 2022-06-21
fix: back button on post modal

sshanzel:
- Instead of replacing history on navigation - we should push the history based on the user's action.
- This should allow the user to utilize the native browser's function of the back and forward button
- This means the back button for mobile should function as intended as well. This will greatly improve users' experience, I believe.
- Since we have a custom hook that manages the navigation - I migrated everything related to navigation in that hook itself.

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

DD-634 #done

### DIFF
--- a/packages/shared/src/hooks/usePostModalNavigation.ts
+++ b/packages/shared/src/hooks/usePostModalNavigation.ts
@@ -8,8 +8,8 @@ import { useKeyboardNavigation } from './useKeyboardNavigation';
 interface UsePostModalNavigation {
   onPrevious: () => void;
   onNext: () => Promise<void>;
-  onOpenModal: (index: number) => void;
-  onCloseModal: () => void;
+  onOpenModal: (index: number, fromPopState?: boolean) => void;
+  onCloseModal: (fromPopState?: boolean) => void;
   isFetchingNextPage?: boolean;
   selectedPost: Post | null;
 }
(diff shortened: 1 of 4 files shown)


packages/shared/src/hooks/usePostModalNavigation.ts
```diff
@@ -18,15 +18,83 @@ export const usePostModalNavigation = (
...
   fetchPage: () => Promise<unknown>,
 ): UsePostModalNavigation => {
+  const [currentPage, setCurrentPage] = useState<string>();
+  const isExtension = !!process.env.TARGET_BROWSER;
   const [openedPostIndex, setOpenedPostIndex] = useState<number>(null);
   const [isFetchingNextPage, setIsFetchingNextPage] = useState(false);
   const { trackEvent } = useContext(AnalyticsContext);
 
+  const changeHistory = (data: unknown, title: string, url: string) => {
+    if (!isExtension) {
```
rebelchris, line 28: Is this method not supported in the extension? (Don't we do something similar with the feed types?)

rebelchris (approved):
Wow this is pretty slick.
Are we sure we can't use this for the extension?

And please make sure to test it out really well in different browsers as well.

PR #1041, opened 2022-06-21
style: fix visual bugs on the dropdown on the main feed

AnishDe12020:
## Changes
Fixes https://github.com/dailydotdev/daily/issues/586

- Fixes the text being cut off and box shadow on focus being cut off
image

- [x] MobileL (420px)
- [x] Tablet (656px)
- [x] Laptop (1020px)

- [ ] iOS (Chrome and Safari) 
- [x] Android


packages/shared/src/components/fields/Dropdown.tsx
```diff
@@ -114,9 +114,7 @@ export function Dropdown({
         aria-expanded={isVisible}
       >
         {icon}
-        <span className="flex flex-1 mr-1 truncate">
-          {options[selectedIndex]}
-        </span>
+        <span className="flex flex-1 truncate">{options[selectedIndex]}</span>
```
rebelchris, line 117: I think we need the margin right for the arrow icon.
Perhaps it's better to remove the parent container set width, not sure why we needed that.

packages/shared/src/components/MainFeedLayout.tsx
```diff
@@ -235,7 +235,7 @@ export default function MainFeedLayout({
   const header = (
     <LayoutHeader className="flex-row">
       {!isSearchOn && getFeedTitle()}
-      <div className="flex flex-row flex-wrap gap-4 items-center mr-px">
+      <div className="flex flex-row flex-wrap gap-4 items-center py-2 mr-px">
```
rebelchris, line 238: Looks like an overflow issue then?
Perhaps we can rather leverage border instead of the box-shadow for the border then?

rebelchris (changes requested):
Nice, thanks so much for this contribution 💪
I think we can make it a little bit more robust with the changes described.

AnishDe12020:
Hey @rebelchris, I removed the explicit setting of the width but using border with it cause small layout shifts when clicking on the dropdown. After some googling, I figured out that I can use `inset` for our back shadow (so no overflow issue and no layout shifts). 

Do let me know if it is fine

packages/shared/src/components/MainFeedLayout.tsx
```diff
@@ -277,7 +277,6 @@ export default function MainFeedLayout({
         {navChildren}
         {isUpvoted && (
           <Dropdown
-            className="w-44"
```
rebelchris, line 280: Sorry, my bad we use this to prevent layout shifts.

packages/shared/src/components/MainFeedLayout.tsx
```diff
@@ -287,7 +286,6 @@ export default function MainFeedLayout({
         )}
         {sortingEnabled && isSortableFeed && (
           <Dropdown
```
rebelchris, line 288: This should work ;)
```suggestion
          <Dropdown className="w-44"
          
```

rebelchris (changes requested):
Almost there, I think if we settle for the described width's we are there.
The inset works fine it seems 💪

AnishDe12020:
I have added back the width (`w-44` tailwind CSS class) to both the dropdowns. Also, the slightly higher width for the sorting dropdowns (10.5rem vs `w-44` that is 11rem) seems to make the recommended text appear completely :D

rebelchris (approved):
Perfect, thanks so much for this work @AnishDe12020 🥳

PR #1044, opened 2022-06-22
refactor: fetching shadow dom on extension only

sshanzel:
- Have changed the decision process where the user mention tooltip will append the result so we won't have to bother from the list and ensure it will always be the actual DOM for webapp.

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

DD-{number} #done


rebelchris (approved):
Nice find 💪
Would we still need to redeploy extension for this to activate?

PR #1045, opened 2022-06-22
fix: setting return state on modal close

sshanzel:
- There was a race condition when setting the current page where it falls back.

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

DD-{number} #done


kirillkurko (approved):
Nice catch! 💪

sshanzel:
@rebelchris this should now handle the issue when hitting back and the post is not found in the feed. Before I really didn't know what to execute so I just close the modal, now it is more clear that we should actually trigger reload so the actual article page will be displayed, instead.

In the article page, when you hit the back button from the process we were emulating, it won't do anything since the history came from `pushState` not from pushing in nextjs router. So we have to execute a reload.

rebelchris (approved):
This looks pretty solid from my side.

PR #1052, opened 2022-06-23
fix: extension comment modal size

kirillkurko:

Now we use tailwind class instead of inline style, so it's our plugin can convert it from rem to pixels and it's solves issue with modal height in companion and we don't rely on site font size

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

DD-{number} #done

### DIFF
--- a/packages/shared/src/components/modals/CommentBox.tsx
+++ b/packages/shared/src/components/modals/CommentBox.tsx
@@ -103,7 +103,7 @@ function CommentBox({
 
   return (
     <>
-      <div className="overflow-auto" style={{ maxHeight: '31rem' }}>
+      <div className="overflow-auto max-h-[31rem]">
         <article
           className={classNames(
             'flex flex-col items-stretch',


rebelchris (approved):
Nice one! Did you validate the output is now in pixels?

sshanzel:
@kirillkurko kindly validate in the preview deployment in the webapp as well since the reason we were not using tailwindcss arbitrary value is, that it's unstable. Sometimes it generates the desired value sometimes it does not with our current setup with the webapp.

kirillkurko:
> @kirillkurko kindly validate in the preview deployment in the webapp as well since the reason we were not using tailwindcss arbitrary value is, that it's unstable. Sometimes it generates the desired value sometimes it does not with our current setup with the webapp.
> 
> Note: the issue I mentioned is on production build.
> 
> You can set this up in one of our vercel's preview deployment, so the comment box can actually be shown.

Got it, okay, but now issue with login blocks the testing, because I cannot sigh up to deployed preview

sshanzel:
> Got it, okay, but now issue with login blocks the testing, because I cannot sigh up to deployed preview

Yes, we have preview deployments in Vercel where it would allow login. If you haven't tried it in our setup, just let me know, and I'll walk you through how to set it up.

kirillkurko:
> Yes, we have preview deployments in Vercel where it would allow login. If you haven't tried it in our setup, just let me know, and I'll walk you through how to set it up.

It will great if you will show me it

sshanzel (approved):
Woop! Woop! Tested it out in preview deployment, and it worked! 🚢

PR #1053, opened 2022-06-24
feat: change validation reference for profile form

rebelchris:
- We currently blocked the submit button, but this caused issues with auto-fill plugins
- We already leverage browser form validation 
- Second to the browser validation did we move the validation check to the submit function

- [x] Have you done sanity checks in the webapp

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-153 #done

### DIFF
--- a/packages/shared/src/components/profile/ProfileForm.tsx
+++ b/packages/shared/src/components/profile/ProfileForm.tsx
@@ -30,7 +30,6 @@ const timeZoneValues = timeZoneOptions.map((timeZone) => timeZone.label);
 export type RegistrationMode = 'default' | 'author' | 'update';
 
 export interface ProfileFormProps extends HTMLAttributes<HTMLFormElement> {
-  setDisableSubmit?: (disable: boolean) => void;
   onSuccessfulSubmit?: (optionalFields: boolean) => void | Promise<void>;
   mode?: RegistrationMode;
 }
@@ -46,7 +45,6 @@ const FormSwitch = classed(Switch, 'my-3', styles.formSwitch);
 const defaultEmailHint = 'Not publicly shared';
 
 export default function ProfileForm({
-  setDisableSubmit,
   onSuccessfulSubmit,
   mode,
   className,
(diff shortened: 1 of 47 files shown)


kirillkurko (approved):
Nice, let's merge it

rebelchris:
@idoshamun Just a quick check, do you see any danger in these changes?
I tested it out in 3 browsers and seems to be rock solid 🥳

packages/shared/src/components/profile/ProfileForm.tsx
```diff
@@ -97,37 +87,38 @@ export default function ProfileForm({
...
-          setEmailHint('This email is already used');
-        } else if (res.field === 'username') {
-          setUsernameHint('This username is already taken');
-        } else if (res.field === 'twitter') {
-          setTwitterHint('This Twitter handle is already used');
-        } else if (res.field === 'github') {
-          setGithubHint('This GitHub handle is already used');
-        } else if (res.field === 'hashnode') {
-          setGithubHint('This Hashnode handle is already used');
+    if (formRef.current.checkValidity()) {
```
idoshamun, line 91: What happens in case the form is not valid? Can you attach a print screen with an example please?

packages/shared/src/components/profile/ProfileForm.tsx
```diff
@@ -97,37 +87,38 @@ export default function ProfileForm({
 
   const onSubmit = async (event: FormEvent): Promise<void> => {
     event.preventDefault();
-    setDisableSubmit?.(true);
```
idoshamun, line 100: I think we need to disable the button while submitting so the user will not double click on it

idoshamun (changes requested):
Some concerns before we merge

packages/shared/src/components/profile/ProfileForm.tsx
rebelchris: Ah valid point indeed!

packages/shared/src/components/profile/ProfileForm.tsx
rebelchris: You basically get these native HTML errors besides the ones we have set.

Screenshot 2022-06-24 at 17 47 14

Screenshot 2022-06-24 at 17 46 58

idoshamun: Not the best UI for this but given it's about to be changed soon that's ok

idoshamun (approved):

sshanzel (approved):
LGTM! 🚀

PR #1054, opened 2022-06-25
fix: tag button icon changes when selected 

kirillkurko:

Now classNames are passed to button icon when doing close, so they are applied. It resolves all the issues with icon buttons (mostly rotation). Also some classes were removed from tag button icon, because we don't need them

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

DD-{number} #done

### DIFF
--- a/packages/shared/src/components/filters/TagButton.tsx
+++ b/packages/shared/src/components/filters/TagButton.tsx
@@ -59,7 +59,7 @@ const UnfollowTagButton = ({
   <GenericTagButton
     {...props}
     className={classNames('btn-primary', className)}
-    icon={<PlusIcon className="ml-2 text-xl transition-transform rotate-45" />}
+    icon={<PlusIcon className="ml-2 transition-transform rotate-45" />}
     action={action}
     tag={tag}
   />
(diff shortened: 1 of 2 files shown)


packages/shared/src/components/filters/TagButton.tsx
```diff
@@ -78,7 +78,7 @@ const FollowTagButton = ({
   <GenericTagButton
     {...props}
     className={classNames('btn-tag', className)}
-    icon={<PlusIcon className="ml-2 text-xl transition-transform rotate-0" />}
```
rebelchris, line 81: I think we used the transform here to do the undo transform could you verify that still animates? (so both ways)

rebelchris (changes requested):

packages/shared/src/components/filters/TagButton.tsx
kirillkurko: Yeah, you are right, it is so tiny that I haven't initially noticed it

rebelchris (approved):
Thanks for fixing these 💪

PR #1062, opened 2022-06-29
fix: header to display both create feed and dropdowns

sshanzel:
- From the link, it looks like both elements need to be present even when the user has no filters yet.

https://miro.com/app/board/uXjVOPPtumw=/?invite_link_id=273386981736

Preview:

image

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-172 #done

### DIFF
--- a/packages/shared/src/components/CreateMyFeedButton.tsx
+++ b/packages/shared/src/components/CreateMyFeedButton.tsx
@@ -48,7 +48,7 @@ export default function CreateMyFeedButton({
   }, [buttonCopy]);
 
   return (
-    <div className="flex flex-col items-center mb-8 w-full">
+    <div className="flex flex-col items-center mb-4 w-full">
       <div
         className={classNames(
           'p-2 border flex-col tablet:flex-row flex items-center rounded-12',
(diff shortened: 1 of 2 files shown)


rebelchris:
Wow can't believe we missed that! 🤯

rebelchris (approved):
Wow can't believe we missed that 🤯

rebelchris (changes requested):
Wow, can't believe we missed this.

However I think we still have the issue with the dropdown not showing.

Also noticed another issue, there is a UX jump if you switch between the feeds.
I think this should be solved here as well.

sshanzel:
> However I think we still have the issue with the dropdown not showing.

You will have to go to the Popular feed to then open the Customize Modal and toggle the enable sorting of feed. It is only shown in Popular and My Feed.

> Also noticed another issue, there is a UX jump if you switch between the feeds.
I think this should be solved here as well.

With the fixed height, the jumpy shift should now be gone.

rebelchris (approved):
Nice, love the fixed height.
Also totally forgot about that settings checkbox, so good to go 💖

PR #1063, opened 2022-06-30
fix: icon sizes and spacings fixed 

kirillkurko:
- profile social icons size updated
- empty bookmark screen icon size updated
- icon component is temporally changed and now outline icon is optional 

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

DD-{number} #done


rebelchris (changes requested):
This will collide a bit with my open PR.
I think for the social icons we should keep the naming uniform.

- primary = white
- secondary = color

Alternatively the files can be named white.svg and color.svg and we can set the respective ones as primary/secondary (for now of course filled/outlined)

Let me know if that makes sense.

kirillkurko:
> 

Yeah, I'm agree, I made it this way just for now, because your changes is not yet there. If you want I can go with seconds suggestion for now and use filled/outlined (but for some of the icons there is not color version, so I need to pass the same white icon twice as filled and outlined)

packages/shared/src/components/modals/SharedBookmarksModal.tsx
```diff
@@ -119,7 +119,7 @@ export default function SharedBookmarksModal({
             <DiscordIcon />
             <TwitterIcon />
             <SlackIcon />
-            <GithubIcon />
+            <GithubIcon size="medium" />
```
rebelchris, line 122: Why is this one specifically medium?

packages/webapp/pages/devcard.tsx
```diff
@@ -282,7 +282,7 @@ const Step2 = ({
               defaultTypo={false}
               target="_blank"
             >
-              <GitHubIcon className="mr-2" />
+              <GitHubIcon className="mr-2 w-auto h-auto" />
```
rebelchris, line 285: Did we not add a auto class?

rebelchris (changes requested):
Some small remarks, but looks great already.

packages/webapp/pages/devcard.tsx
kirillkurko: No, we don't and really I don't want to introduce this one - I think it's better to go with concrete sizes set through size prop, but now we have two places where GitHub icon is 16px and 17px (I already mentioned it in a Slack thread), so it's two new sizes for our icon component. So I prefer to go with className here and update it later when we will have a guideline for icon sizes.

packages/shared/src/components/modals/SharedBookmarksModal.tsx
kirillkurko: Other icons are .svg and size is set by the wrapping di, but this one is a component, so I need to explicitly set size for it

rebelchris: Can we not convert them all to components then?

packages/webapp/pages/devcard.tsx
kirillkurko: Current size here is 17px and to keep it 17px we need to set w-auto and h-auto, because height is set by wrapping element. And in account details GitHub icon is 16px. These are two new sizes and I'm not even sure that 17px is something that should exist, but I think it will be better to keep it as it is while we don't have a guideline

packages/shared/src/components/modals/SharedBookmarksModal.tsx
kirillkurko: Yes, I just made it

rebelchris (approved):
Nice! 👍

PR #1064, opened 2022-06-30
refactor: usage of root node for getting window or shadow root

sshanzel:
- After reading through the proposals of what will be the official API for getting the `Selection` on shadow DOM, there seems to be no clear date of when to expect it. 
- Though after going through the links below, I was able to find a better solution for fetching the actual root.
- We now require an element to be passed to identify which root it belongs to when we use `getRootNode`.

Relevant links:
- https://github.com/WICG/webcomponents/issues/79
- https://github.com/mfreed7/shadow-dom-selection#proposal (Proposal)

One example is if you visit our preview deployment, which is not whitelisted in our background worker, and if you inspect the HTML, you'll notice the shadow DOM to be present but the user mention still works since the `getRootNode` provides a more accurate for which the element was embedded.

Preview deployment: https://preview.app.daily.dev/

### **Please make sure existing components are not breaking/affected by this PR**

I have also manually tested the preview deployment on all major browsers we support (including Safari and Edge).

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-104 #done


rebelchris (approved):
Wow interesting find!
Looks great, thanks for testing and validating this better fix :)

sshanzel:
@rebelchris I refactored it a bit to be safer from edge cases on any other browsers outside of Chromium, just to be sure. Let me know if have any concerns about it. Thank you for the awesome review!

rebelchris (approved):

PR #1065, opened 2022-07-01
fix: remove auto permission popup

rebelchris:
- Since we are done with the launch we no longer need the auto-popup alert

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-175 #done


sshanzel (approved):
Nice cleanup!

Though I have a feeling we might require another kind of alert in the future for our extension, I'm not sure about removing the custom hook completely. But feel free to move on if you think the excess files won't be necessary 🚢

rebelchris:
@sshanzel Think you might be right we need it in the future so returned those two :)

PR #1067, opened 2022-07-01
fix: cleanup all affecting timeouts

rebelchris:
- We had a outstanding task to introduce cleanups for all affected setTimeouts
- There are exceptions for instance onCopyLink as this is a user action and only fires on event it can be omitted for now.
- Tested all affected changes locally

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-194  #done


packages/shared/src/components/modals/NewRankModal.tsx
```diff
@@ -73,24 +75,40 @@ export default function NewRankModal({
 
   useEffect(() => {
```
kirillkurko, line 76: I feel like we have this kinda low-level logic of handling listener and timeouts. My first idea was to have useAnimateRank hook with this logic hidden, but I'm not sure it can be done in a clear way (one of the reasons is that we have state updates inside), so it's just more like a side note

kirillkurko (commented):
Nice one! 
What I think that we can improve is to move timeout delays to constants with meaningful names and use them instead of numbers.

And overall thing I feel about timeouts and event listeners is that they are usually on a lover abstraction level than what you want to see inside components, so it's better to keep them hidden inside custom hooks

kirillkurko (approved):

packages/shared/src/components/modals/NewRankModal.tsx
kirillkurko: Yes, but I think there is nothing wrong with hooks used only in one place if they help to keep components more readable and hide implementation details from the component itself, so hook works like a contract for a component and can be implemented in different ways.

Disclaimer - example is abstract and I'm not talking about what we have in this PR and I just want to explain how I feel about hooks. 

Example:

```
// @hooks/useAnimateRank


// different possible ways of implementation
// 1. Using refs and timeouts 
// 2. Using some animation library
export const useAnimateRank = () => {
// Exact implementation here
}

```

```
// RankComponent.tsx

import useAnimateRank from '@hooks/useAnimateRank';

...
...

const RankComponent = () => {
   
   // Implementation can be different 
   // But hook is used only in this way
   // So no changes in component required when hook implementation changed
   useAnimateRank();

   return <>{/*component JSX here*/}</>
};

...
...

```

Hope it makes sense, but for now we are good with what you just implemented

rebelchris: Totally and 100% agree this would be amazing to rework, let me add a new backlog item for it.

sshanzel:
Now that we have specific ways of using timeouts and intervals, I think it would be a great idea to create a custom hook that will encapsulate the reference and will return two methods. One is to trigger the function, second is to trigger the cleanup manually.

Then we can probably have an additional use effect inside that will ensure when it unmounts, the reference will be cleaned up automatically.

This would also avoid future mistakes of forgetting to clean up these events. We will just be using `useTimeout` or `useInterval`

kirillkurko:
> Now that we have specific ways of using timeouts and intervals, I think it would be a great idea to create a custom hook that will encapsulate the reference and will return two methods. One is to trigger the function, second is to trigger the cleanup manually.
> 
> Then we can probably have an additional use effect inside that will ensure when it unmounts, the reference will be cleaned up automatically.
> 
> This would also avoid future mistakes of forgetting to clean up these events. We will just be using `useTimeout` or `useInterval`
> 
> Though this can be worked in a separate PR, just let me know what you think so I can move forward in reviewing if so.

I think that's a great suggestion

packages/shared/src/components/Greeting.tsx
```diff
@@ -69,21 +79,15 @@ export default function Greeting({
       }
 
       const now = new Date();
-      const lastGreeting = await getCache<Date>('greeting');
+      const lastGreeting = await getCache<Date>('greeting+');
```
kirillkurko, line 82: Is this plus expected?

kirillkurko (commented):
So much cleaner - good job! I just left a small comment. Also I see empty useTimeout.ts file, let's delete it if we don't need

packages/shared/src/components/Greeting.tsx
rebelchris: Nope, my bad :)

rebelchris:
@kirillkurko Has been removed.
Additionally I want to introduce a `useEventListener` hook, but rather as a separate PR

kirillkurko (approved):
Cool, I think now we are good to merge it

sshanzel:
> Additionally I want to introduce a useEventListener hook, but rather as a separate PR

That's a great idea! 🚢

sshanzel (approved):
Not only it is so much cleaner, but it is also now much easier to read the code. Amazing work indeed! 🚢

PR #1069, opened 2022-07-01
fix: make test cases use uniform data objects

rebelchris:
- We had multiple test fixtures, uniformed to be all in the shared repo now
- We had multiple test having their own variables, using the fixtures now
- Cleaned some test as they where using hardcoded strings to test vs variables from the fixtures
- (Not perfect, but addresses a lot of issues we had with tests)

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-196 #done


packages/shared/__tests__/fixture/comment.ts
```diff
@@ -0,0 +1,14 @@
...
+const comment: Comment = {
+  id: 'c1',
+  content: 'my comment',
+  contentHtml: '<p>my comment</p>',
+  author,
+  createdAt: new Date(2017, 1, 10, 0, 0).toISOString(),
+  upvoted: false,
+  permalink: 'https://daily.dev',
+  numUpvotes: 0,
+};
```
kirillkurko, line 13: My perfectionist just cannot ignore it - let's insert an empty line here 🤣

kirillkurko (approved):
This one is great, so happy to see how much we removed 😀

sshanzel (approved):
Wow, this is just amazing! Can't imagine the time and effort you have put into this to ensure test cases will result in the same. Thank you! 🙇

PR #1070, opened 2022-07-05
fix: companion popup placement

sshanzel:
- After further checking, it seems the root cause of miscalculation in the offset (and had to use) was the max width that was set default from tippyjs. Our component exceeds it, but since the value was explicitly set, it has been thought to be the actual width.
- Also checked from the actual library if the removal of the max width is safe, and there seems to be no issue about it: https://github.com/atomiks/tippyjs/issues/451#issuecomment-473252780

There was also an issue before for its misplacement when the user is not logged in. This has resolved that issue as well. Have thoroughly checked and all seems to be in place now and much more stable.

Issues to close:
- https://github.com/dailydotdev/daily/issues/620
- https://github.com/dailydotdev/daily/issues/574

Will close the relevant issues once we have verified the fix.

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-198 #done

### DIFF
--- a/packages/extension/src/companion/CompanionPopupButton.tsx
+++ b/packages/extension/src/companion/CompanionPopupButton.tsx
@@ -53,7 +53,6 @@ export const CompanionPopupButton = ({
     <SimpleTooltip
       content={<CompanionPermission />}
       placement="bottom-start"
-      offset={[-140, 12]}
       container={{
         paddingClassName: 'px-6 py-4',
         bgClassName: 'bg-theme-bg-primary',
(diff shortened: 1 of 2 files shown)


rebelchris (approved):
Just make sure this doesn't introduce any other edgecase with existing tooltips.

Perhaps just check with super long source name tooltip, user profile and some generic ones.
And second perhaps some other that use offset (if any)

Other then that, ready to go.

sshanzel:
We should be safe with the mentions as it has only a max of 280px. Let's merge! 🚢

PR #1074, opened 2022-07-07
Wt 218 fallback block

rebelchris:
- Introducing a temporary solution to enable facebook sharing.
- Crawler would not find data as it was still eager loading, this way we block and wait for the data

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-218 #done

### DIFF
--- a/packages/shared/src/components/fields/TextField.tsx
+++ b/packages/shared/src/components/fields/TextField.tsx
@@ -35,6 +35,7 @@ interface InputFontColorProps {
   disabled?: boolean;
   focused?: boolean;
   hasInput?: boolean;
+  actionIcon?: React.ReactElement<IconProps>;
 }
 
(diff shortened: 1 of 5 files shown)


idoshamun:
Are we sure there's no other easy way?

rebelchris:
@idoshamun Not really sure, we can't use the new only block bots feature yet.
I don't really know what other easy options we could have at this point.

Unless we disable facebook for now as I don't envision even many people use it.
(Compared to what latency we introduce here)

Do you want to put this on hold?
I just didn't want to block a new extension

idoshamun:
Don't wait with the new extension for it. Let's discuss it next week to see if there are more alternatives

PR #1078, opened 2022-07-07
feat: new authentication

sshanzel:
## Changes
There are a lot of things going on as this is almost like a new feature. The core idea here is to revamp the authentication flow. We will have our modal as the primary way of logging in, which will have different variations, then we will also have a dedicated login page as v3 and v4.

There are a lot of newly introduced components but solely used in the authentication flow. I think it would be better to go through the code instead of explaining here since those are mostly container elements or forms.

## TODO
- [ ] Dynamic import

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

DD-{number} #done


packages/shared/src/components/DailyCircle.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+function DailyCircle({
+  size = 'medium',
+  className,
+}: DailyCircleProps): ReactElement {
+  return (
+    <div
+      className={classNames('rounded-full', sizeMap[size], className)}
+      style={{
+        background: 'linear-gradient(135deg, #E436FE, #1B3C67)',
```
sshanzel, line 26: The color values are not found in our design system.

packages/shared/src/components/auth/common.tsx
```diff
@@ -0,0 +1,100 @@
...
+    icon: <GitHubIcon />,
+    provider: 'GitHub',
+    style: { backgroundColor: '#2D313A' },
+  },
+  apple: {
+    icon: <AppleIcon />,
+    provider: 'Apple',
+    style: { backgroundColor: '#0E1217' },
+  },
+};
```
sshanzel, line 46: Utilize existing color classes introduced in a recently merged PR for social providers.

packages/shared/src/components/modals/DiscardActionModal.tsx
```diff
@@ -0,0 +1,42 @@
+import React, { ReactElement, MouseEvent } from 'react';
```
sshanzel, line 1: It would probably be a better idea to create a generic modal for confirmation-related ones.

packages/webapp/pages/login.tsx
```diff
@@ -0,0 +1,52 @@
...
+            <AuthModalHeading className="z-1 ml-32 typo-mega1">
+              Where developer
+              <br />
+              grow together.
+            </AuthModalHeading>
+          )}
+        </div>
+        {!isV4 && (
+          <img
+            src="/login.png"
```
sshanzel, line 42: Host the image in our cloudinary before going live.

sshanzel:
This should be ready for initial review. Only the base structure for the elements to display but none of the actual functions as it requires integration.

cc @rebelchris @kirillkurko

packages/shared/src/components/auth/RegistrationForm.tsx
```diff
@@ -0,0 +1,164 @@
...
+
+export const RegistrationForm = ({
+  email,
+  socialAccount,
+  formRef,
+  onClose,
+  onBack,
+  isV2,
+}: RegistrationFormProps): ReactElement => {
+  const [isNameValid, setIsNameValid] = useState<boolean>();
```
kirillkurko, line 51: A general question - haven't we ever thought about having a form-handling solution? Probably this is something we can discuss on our web-team meeting @rebelchris

packages/shared/src/lib/featureManagement.ts
```diff
@@ -130,6 +130,13 @@ export class Features {
     ['control', 'v1', 'v2', 'v3'],
   );
 
+  static readonly AuthVersion = new Features('auth_version', 'v1', [
```
kirillkurko, line 133: Check the way of implementation of `AdditionalInteractionButton` below introduced by Chris (in `featureValues.ts` you will have an enum `AuthVersions` that will be used here instead of strings)

packages/webapp/pages/login.tsx
```diff
@@ -0,0 +1,52 @@
...
+import Logo from '@dailydotdev/shared/src/components/Logo';
+import useAuthForms from '@dailydotdev/shared/src/hooks/useAuthForms';
+import React, { ReactElement, useContext } from 'react';
+import classNames from 'classnames';
+import FeaturesContext from '@dailydotdev/shared/src/contexts/FeaturesContext';
+
+function Login(): ReactElement {
+  const { onSocialProviderChange, formRef, socialAccount } = useAuthForms();
+  const { authVersion } = useContext(FeaturesContext);
+  const isV4 = authVersion === 'v4';
```
kirillkurko, line 12: Just to point out that enum can be used

packages/shared/src/components/auth/IncompleteRegistration.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+  const timezoneUpdated = (timezone: string) => {
+    const findTimeZoneRow = timeZoneOptions.find((_timeZone) => {
+      return _timeZone.label === timezone;
+    });
+    setUserTimeZone(findTimeZoneRow.value);
+  };
+
+  const hour = getHourTimezone(userTimeZone);
+  const Background = hour >= 6 && hour < 18 ? DaytimeIcon : NighttimeIcon;
```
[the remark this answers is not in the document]
sshanzel: I think I created one in another PR.

packages/shared/src/components/auth/RegistrationForm.tsx
sshanzel: Good point. Was thinking about that as well. My first option is `react-forms`.

packages/shared/src/components/auth/IncompleteRegistration.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+const timeZoneOptions = getTimeZoneOptions();
+const timeZoneValues = timeZoneOptions.map((timeZone) => timeZone.label);
+
+function IncompleteRegistration(props: ModalProps): ReactElement {
+  const [userTimeZone, setUserTimeZone] = useState<string>(
+    getUserInitialTimezone({ userTimezone: 'utc' }),
+  );
+
+  const timezoneUpdated = (timezone: string) => {
```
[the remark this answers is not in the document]
sshanzel: This was a copy and pasted code from another file. Probably better to create a hook to manage timezone-related stuff.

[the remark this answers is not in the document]
kirillkurko: Oh yeah, then hook will be great

packages/shared/src/components/auth/AuthDefault.tsx
```diff
@@ -0,0 +1,98 @@
...
+interface AuthDefaultProps {
+  children?: ReactNode;
+  onClose?: CloseModalFunc;
+  onSignup?: (email: string) => unknown;
+  onProviderClick?: (provider: string) => unknown;
+  onForgotPassword?: () => unknown;
+  isV2?: boolean;
+}
+
+export const AuthDefault = ({
```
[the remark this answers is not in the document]
sshanzel: There's nothing special with this one.

packages/shared/src/components/auth/AuthModal.tsx
```diff
@@ -0,0 +1,97 @@
...
+    onContainerChange,
+    onSocialProviderChange,
+    formRef,
+    socialAccount,
+    container,
+    isDiscardOpen,
+  } = useAuthForms({
+    onDiscard: onRequestClose,
+  });
+  const isV1 = authVersion === 'v1';
```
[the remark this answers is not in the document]
sshanzel: Good point. I forgot to include what was recently introduced process of managing these feature values.

packages/shared/src/components/auth/IncompleteRegistration.tsx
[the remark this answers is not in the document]
sshanzel: Will apply your thoughts on the said hook. Thank you!

packages/shared/src/components/auth/RegistrationForm.tsx
kirillkurko: It's something we will probably research if we decided to go with some solution - have never heard `react-forms`, but `react-hook-form` is a great choice from my experience (zero dependencies, small and fast)

packages/shared/src/components/auth/AuthDefault.tsx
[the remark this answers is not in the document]
kirillkurko: There is only one export and usually we use `function` keyword for our components, so maybe we can convert to `function` and export as default?

packages/shared/src/components/auth/RegistrationForm.tsx
sshanzel: I might have forgotten the name, but I think, yes, that's the same I'm referring to.

packages/shared/src/components/auth/AuthDefault.tsx
```diff
@@ -0,0 +1,100 @@
...
+interface AuthDefaultProps {
+  children?: ReactNode;
+  onClose?: CloseModalFunc;
+  onSignup?: (email: string) => unknown;
+  onProviderClick?: (provider: string) => unknown;
+  onForgotPassword?: () => unknown;
+  isV2?: boolean;
+}
+
+const AuthDefault = ({
```
rebelchris, line 21: What exactly is AuthDefault?
It almost looks like a modal, but not really sure it is.

rebelchris (changes requested):
Amazing work and really happy to see the massive effort in this PR.
Have some generic comments for now.

packages/shared/src/hooks/useAuthForms.ts
```diff
@@ -0,0 +1,72 @@
...
+      onDiscardAttempt,
+      onDiscardCancelled,
+      onContainerChange: setContainer,
+      onSocialProviderChange: setSocialAccount,
+      container,
+      isDiscardOpen,
+      socialAccount,
+      formRef,
+    }),
+    [formRef?.current, container, isDiscardOpen, socialAccount],
```
kirillkurko, line 68: Do we need a `formRef?.current` as a dependency here? Mutable values like ref.current do not triggers re-renders, so result of `useMemo` will be the same if only `formRef?.current` is updated

kirillkurko (commented):
Looks good, left just one comment and some comments from Chris are still pending

packages/shared/src/components/auth/AuthDefault.tsx
sshanzel: Not the modal entirely but the screenshot below. The reason is, that it is also used on the login page where it is not part of any modal, but the exact same box.

image

packages/shared/src/components/auth/AuthModal.module.css
```diff
@@ -0,0 +1,44 @@
...
+    overflow: hidden;
+    background: var(--theme-background-primary);
+
+    @screen laptop {
+      max-width: 35rem;
+    }
+  }
+}
+
+.circles {
```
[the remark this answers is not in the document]
sshanzel: I think we can. Sharp eyes! 🔭

packages/shared/src/components/auth/AuthModalFooter.tsx
```diff
@@ -0,0 +1,34 @@
+import classNames from 'classnames';
+import React, { ReactElement } from 'react';
+import { ClickableText } from '../buttons/ClickableText';
+
+interface AuthModalFooterProps {
+  className?: string;
+  shouldLogin: boolean;
```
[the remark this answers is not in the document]
sshanzel: Ah, yes, I agree. I took this out from a previous component and used the same parameter which had context back then. I think `isLogin` or something similar should be better.

packages/shared/src/components/modals/LoginModal.tsx
```diff
@@ -1,45 +1,11 @@
...
-      <ModalCloseButton onClick={onRequestClose} />
-      <DailyDevLogo />
-      <div className="mt-6 mb-8 text-center text-theme-label-secondary typo-callout">
-        {loginModalDescriptionCopy}
-      </div>
-      <LoginButtons />
-      {children}
-    </StyledModal>
-  );
+export default function LoginModal(props: LoginModalProps): ReactElement {
```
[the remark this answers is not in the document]
sshanzel: This is just temporary, my bad, forgot to add a comment. This was to enable it for demo, but this will be changed before production.

packages/shared/src/hooks/useAuthForms.ts
sshanzel: Yes, we don't expect a re-render here, but it would still look for when the dependency has changed its value, which a re-render from the component where this hook was used, happens.

packages/shared/src/components/auth/RegistrationForm.tsx
sshanzel: We have a built-in one to help us with forms, I haven't checked which parts can be salvaged, I suggest we put it in another ticket as researching for options and which one to use would take time. Also migrating the existing implementations would require work. Though if we have time, then I think we can take it on in this same shipment.

packages/shared/src/hooks/useAuthForms.ts
kirillkurko: Not sure we are on the same place, please check this [codesandox](https://codesandbox.io/s/still-moon-vmbvu1?file=/src/App.tsx) - refs can be safely excluded from dependencies, because React do not track their values

kirillkurko: Yeah, I understand, but ref values are out of React flow, so they are updated not at the time of rendering (and time of execution of hooks), so you can just remove `ref.current` from dependencies

Some replies from Dan Abramov:
https://github.com/facebook/react/issues/15164
https://twitter.com/dan_abramov/status/1105918794239803392?s=21&t=xdo_Ith7_fYnXDHWYQtnjw

@rebelchris looks like react-hooks ESLint rules do not work for us because react-hooks probably not included in plugins in our config (or maybe it was intentional)

sshanzel: I'm not sure you get the point. That's why he said, "In general cases". As I've been saying the whole time, we are not looking for a re-render here explicitly when the refs change. It is a safety net for when the outside state triggered a re-render, it would at least let the hook realize the reference in the dependency could have changed.

sshanzel: The point is clear in "In general" cases you don't require dependencies that don't trigger a re-render. That's why it was mentioned it is not part of React flow as it does not trigger re-render when its reference/value changes. And we are not waiting explicitly when that changed. In the end, the dependencies (including the refs) would still be taken a look at if it has changed, which will then re-compute. That's why you'll notice in most of our cases, the refs are not really included, but in this case, it was.

packages/shared/src/components/auth/AuthDefault.tsx
rebelchris: Oh ok, then it makes sense :)

packages/shared/src/components/MainLayout.tsx
```diff
@@ -103,18 +124,62 @@ export default function MainLayout({
   enableSearch,
   onShowDndClick,
 }: MainLayoutProps): ReactElement {
-  const { user, loadingUser } = useContext(AuthContext);
+  const router = useRouter();
```
idoshamun, line 127: I think we should consider taking out everything auth related from here.
It already has quite a lot of code here

packages/shared/src/components/MainLayout.tsx
```diff
@@ -123,8 +188,30 @@ export default function MainLayout({
     setOpenMobileSidebar(state);
   };
 
+  useEffect(() => {
```
idoshamun, line 191: Are we sure incomplete registration should be a state rather than a derived value from the user object?

packages/shared/src/hooks/useAuthForms.ts
```diff
@@ -0,0 +1,64 @@
...
+    const values = Object.values(form);
+
+    if (values.some((value) => !!value)) {
+      return setIsDiscardOpen(true);
+    }
+
+    return onDiscard(e);
+  };
+
+  const onDiscardCancelled: CloseAuthModalFunc = (e) => {
```
idoshamun, line 46: `Canceled` I'm up for the american version of the word :P

idoshamun (approved):
As much as I was able to cover, it looked good.
I'd go over the big components and see where we can refactor some logic out.
The authentication logic is quite complex so we have to make sure it's living separately from the rest

packages/shared/src/components/MainLayout.tsx
sshanzel: Right. We can just derive this from the user object without the needing of any state.

packages/shared/src/components/MainLayout.tsx
sshanzel: I have now completely extracted auth-related code blocks from the MainLayout.

packages/shared/src/components/auth/AccountCompletionModals.tsx
```diff
@@ -0,0 +1,105 @@
...
+  });
+  const onUpdateProfile = ({ name, username }: SocialRegistrationFormValues) =>
+    updateUserProfile({ name, username });
+
+  const displayErrorMessage = (text: string) => {
+    displayToast(text);
+    window.history.replaceState(null, '', '/');
+  };
+
+  useQuery(
```
rebelchris, line 45: Is this not an existing kratos flow call we already have a hook for?

packages/shared/src/components/auth/ProviderButton.tsx
```diff
@@ -0,0 +1,36 @@
...
+  ...props
+}: ProviderButtonProps): ReactElement {
+  return (
+    <Button
+      key={provider}
+      {...props}
+      className={classNames('btn-secondary', className)}
+      style={{
+        justifyContent: label ? 'flex-start' : 'center',
+        color: '#FFFFFF',
```
rebelchris, line 25: Why do we hard set the color here?

packages/shared/src/components/auth/RegistrationForm.tsx
```diff
@@ -0,0 +1,145 @@
...
+      if (!values['traits.name']?.length) {
+        setHints['traits.name'] = 'Please provide name.';
+      }
+      if (!values['traits.username']?.length) {
+        setHints['traits.username'] = 'Please provide username.';
+      }
+
+      onUpdateHints(setHints);
+      return;
+    }
```
rebelchris, line 64: Perhaps we can refactor this into some kind of reusable hook

packages/shared/src/components/alert/AlertContainer.tsx
```diff
@@ -0,0 +1,33 @@
...
+  container?: string;
+  overlay?: string;
+}
+
+interface AlertContainerProps {
+  children: ReactNode;
+  className?: ClassName;
+}
+
+export const AlertBackground = classed(
```
rebelchris, line 14: We can probably extract anything classed to the common file?

packages/shared/src/components/auth/EmailVerificationSent.tsx
```diff
@@ -0,0 +1,84 @@
...
+  initializeKratosFlow,
+  submitKratosFlow,
+} from '../../lib/kratos';
+import {
+  AccountRecoveryParameters,
+  getNodeByKey,
+  VerificationParams,
+} from '../../lib/auth';
+
+const BodyText = classed(
```
rebelchris, line 18: We can probably extract to common file?

packages/shared/src/components/auth/common.tsx
```diff
@@ -0,0 +1,121 @@
...
+
+export const getProviderMapClone = (
+  map: ProviderMap = providerMap,
+): ProviderMap => cloneDeep(map);
+
+export const providers: Provider[] = Object.values(providerMap);
+
+const Container = classed(
```
rebelchris, line 66: Extract this?

rebelchris (changes requested):

packages/shared/src/components/auth/AccountCompletionModals.tsx
sshanzel: I created a new one but it is still on the works. I think I can tweak it to make this one be part of that.

rebelchris (approved):

packages/shared/src/components/alert/AlertContainer.tsx
sshanzel: Good point! Let me work on your concerns about doing some cleaning 🚢

PR #1079, opened 2022-07-08
refactor: comment box from div contenteditable to textarea

sshanzel:
- It was almost impossible to make a comment in FF previously. The root cause was the div contenteditable.
- The reason behind its usage was to natively support the dynamically changing height of the comment box.
- But since it is more important for the users to be able to make a comment, we decided to go back to using textarea.
- Moved from keydown to keyup as it is much more accurate of when we should trigger the mentions.

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-134 #done

### DIFF
--- a/packages/shared/src/components/modals/CommentBox.tsx
+++ b/packages/shared/src/components/modals/CommentBox.tsx
@@ -98,8 +98,7 @@ function CommentBox({
   };
 
   const handleKeydown = (e: KeyboardEvent<HTMLTextAreaElement>) => {
-    const defaultCallback = () => onMentionKeypress(e);
-    onKeyDown(e, defaultCallback);
+    onKeyDown(e);
     e.stopPropagation();
   };
 
@@ -162,6 +161,7 @@ function CommentBox({
             placeholder="Write your comment..."
             onInput={onTextareaInput}
             onKeyDown={handleKeydown}
+            onKeyUp={onMentionKeypress}
             onClick={onInputClick}
             onPaste={onPaste}
             tabIndex={0}
--- a/packages/shared/src/components/modals/CommentBox.tsx
+++ b/packages/shared/src/components/modals/CommentBox.tsx
@@ -92,7 +92,7 @@ function CommentBox({
   };
 
   const handleKeydown = (e: KeyboardEvent<HTMLTextAreaElement>) => {
-    if ((e.ctrlKey || e.metaKey) && e.keyCode === 13 && input?.length) {
+    if ((e.ctrlKey || e.metaKey) && e.key === 'Enter' && input?.length) {
       return sendComment(e);
     }
 
--- a/packages/shared/src/components/modals/CommentBox.tsx
+++ b/packages/shared/src/components/modals/CommentBox.tsx
@@ -3,7 +3,6 @@ import React, {
   useContext,
   useEffect,
   ClipboardEvent,
-  KeyboardEventHandler,
   MouseEvent,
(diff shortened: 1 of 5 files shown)


packages/shared/src/components/modals/CommentBox.tsx
```diff
@@ -54,34 +48,36 @@ function CommentBox({
...
     commentRef.current?.focus();
-    if (commentRef.current && editContent) {
-      commentRef.current.innerText = editContent;
+    if (commentRef.current) {
+      if (editContent) {
+        commentRef.current.value = editContent;
+      }
+      commentRef.current.setAttribute(
+        'data-min-height',
+        commentRef.current.offsetHeight.toString(),
```
rebelchris, line 79: Why do we set it as a string when we need to receive it as int anyway?

packages/shared/src/hooks/useUserMention.ts
```diff
@@ -203,54 +208,58 @@ export function useUserMention({
...
 
       setQuery(undefined);
-    },
-    [query],
-  );
+    };
 
-  useEffect(() => {
-    document.addEventListener('mousedown', userClicked);
+    const dom = commentRef.current?.getRootNode();
```
rebelchris, line 239: Just wondering if this doesn't cause issues in the companion?

rebelchris (approved):
Some small questions, non blocking.
Just make sure to test it our for all three applications and in FF and Chrome. (Don't forget to test the mention popup/preview/long comments etc)

packages/shared/src/components/modals/CommentBox.tsx
```diff
@@ -95,10 +91,27 @@ function CommentBox({
     }
   };
 
-  const handleKeydown = (e: KeyboardEvent<HTMLDivElement>) => {
-    const defaultCallback = () => onMentionKeypress(e);
-    onKeyDown(e, defaultCallback);
-    e.stopPropagation();
+  const handleKeydown = (e: KeyboardEvent<HTMLTextAreaElement>) => {
+    if ((e.ctrlKey || e.metaKey) && e.keyCode === 13 && input?.length) {
```
kirillkurko, line 95: What if we will we move 13 to a constant with a clear name e.g. ENTER_KEY_CODE to avoid "magic numbers"

kirillkurko (approved):
Looks good, left just a small comment

packages/shared/src/hooks/useUserMention.ts
sshanzel: That was actually the reason for this change, as it proves to have some issues when using just "document", so to make it work for the Companion, we need to add the listener to the companion's root node.

packages/shared/src/components/modals/CommentBox.tsx
sshanzel: I think this is due to typescript asking it to be a string than a number since attributes are strings. Though let me double check.

packages/shared/src/components/modals/CommentBox.tsx
sshanzel: Sharp eyes! This was a copy-pasted code from another file, didn't realize it had some of it. Let me create a constant.

sshanzel:
Thank you both for the awesome reviews! After carefully testing out all three apps, I have verified things are working as intended! 🚢

PR #1084, opened 2022-07-09
chore: upgrade to next 12

rebelchris:
- Upgrading to Next 12, but still keeping our babel in favor of SWC (as we don't have all the plugins in SWC yet)
- Does introduce a 5% bundle increate so it's a consideration to make. (Known issue if you don't go full SWC in Next 12)

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-226 #done


idoshamun:
Can we perhaps migrate to SWC? What's the missing part?

rebelchris:
@idoshamun I think at this point we use the following the webapp babel:

- macros -> Not available in SWC yet
- babel-plugin-module-resolver (For our monorepo dependencies so not sure we can use the native next module paths for this)
- react-remove-properties -> Available in SWC
- dynamic-import-node (Can possibly be removed

idoshamun (approved):
The only thing to check is if we bundle (p)react twice to the final bundle.
When I used babel and webpack I had to carefully adjust it so the shared package will not bundle one version of react and the webapp another.

rebelchris:
@idoshamun great point, do you have any guidance on how you checked this before?
Not sure how to tell the difference for the final build.

idoshamun:
You have to use the bundle inspector. We can do it together. Not sure I remember it well enough to guide you through it without seeing it firsthand

packages/shared/src/components/post/AuthorOnboarding.tsx
```diff
@@ -39,10 +38,9 @@ export function AuthorOnboarding({
         <li>Gain reputation points by earning upvotes on articles you wrote</li>
       </ol>
       <div
-        className="grid grid-flow-col gap-x-4 mt-6"
+        className="grid grid-flow-col gap-x-4 mt-6 max-w-[4.625rem]"
```
sshanzel, line 41: I think the `sizeN` has a multiplier of `0.25`.
```suggestion
        className="grid grid-flow-col gap-x-4 mt-6 max-w-[18.5rem]"
```

packages/shared/src/components/utilities.tsx
```diff
@@ -118,7 +118,7 @@ export const NewCommentContainer = classed(
 
 export const ResponsivePageContainer = classed(
   PageContainer,
-  'py-6 laptop:border-theme-divider-tertiary laptop:border-l laptop:border-r laptop:min-h-screen',
+  'py-6 laptop:border-theme-divider-tertiary laptop:border-l laptop:border-r laptop:min-h-screen pb-[1.4375rem]',
```
sshanzel, line 121: The original value was 23 which is 5.75rem, with that, I think we can round it off to just 24 to lessen arbitrary value. What do you think?
```suggestion
  'py-6 laptop:border-theme-divider-tertiary laptop:border-l laptop:border-r laptop:min-h-screen pb-24',
```

sshanzel (approved):
The performance gains here is the next level Dev Experience! Thank you for the initiative! ❤️ 

Just had some values for correction, but all looks so amazing to me! 🚢

kirillkurko (approved):

packages/shared/src/components/utilities.tsx
rebelchris: Ah good point, took it very literal, but yeah think we can do this rounding.

rebelchris:
@idoshamun I've run the bundle analyzer, not really sure what to look for, but it seems the shared doesn't import it's own react right?

Screenshot 2022-07-12 at 10 42 31
Screenshot 2022-07-12 at 10 43 34

idoshamun:
Yes if you can't find preact twice. Or both react and preact in there then it should be fine

PR #1089, opened 2022-07-12
fix: we removed search params on the companion side

rebelchris:
- In this commit we removed the search params
- This was not intended, we only wanted to remove `#` and other weird symbols not the search params
- This fix introduces the whole search string

Edge case:
If the page has multiple search params for these allowed domains it won't be able to find them.
For instance:

```js
https://news.ycombinator.com/item?id=32069418 = valid
https://news.ycombinator.com/item?id=32069418&utm=blabla = not valid
```

Personally think it's good enough for now.
(It only affects domains where we allow querystrings)

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-240 #done

### DIFF
--- a/packages/extension/src/background/index.ts
+++ b/packages/extension/src/background/index.ts
@@ -32,7 +32,7 @@ const isExcluded = (origin: string) => {
 };
 
 const sendBootData = async (_, tab: Tabs.Tab) => {
-  const { origin, pathname } = new URL(tab.url);
+  const { origin, pathname, search } = new URL(tab.url);
   if (isExcluded(origin)) {
     return;
   }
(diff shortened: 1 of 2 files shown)


sshanzel (approved):
Nice find! I think we should also update a function called `removeLinkTargetElement` and make [line 45](https://github.com/dailydotdev/apps/pull/1089/files#diff-67919b1001a13ae9267131777d30ccbd353d2968244149ebd184d2d7a1e864a3R36) utilize the said function.

idoshamun (approved):
Nice catch! 👀

kirillkurko (approved):

PR #1092, opened 2022-07-13
fix: className so value is converted to pixels

kirillkurko:
Share post modal can be of the wrong width on the companion because of style with width set in rem. Now classNames is used, so value is converted to pixels, but we need to use !important to override the rule of the responsive modal used under the hood of share post modal 

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done

### DIFF
--- a/packages/shared/src/components/modals/SharePostModal.tsx
+++ b/packages/shared/src/components/modals/SharePostModal.tsx
@@ -66,7 +66,7 @@ export default function SharePostModal({
     <ResponsiveModal
       padding={false}
       {...props}
-      style={{ content: { maxWidth: '26.25rem' } }}
+      contentClassName="!max-w-[26.25rem]"
     >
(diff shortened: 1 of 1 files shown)


packages/shared/src/components/modals/SharePostModal.tsx
```diff
@@ -66,7 +66,7 @@ export default function SharePostModal({
     <ResponsiveModal
       padding={false}
       {...props}
-      style={{ content: { maxWidth: '26.25rem' } }}
+      contentClassName="!max-w-[26.25rem]"
```
rebelchris, line 69: Do we need the ! important?

rebelchris (approved):
Thanks! :)

packages/shared/src/components/modals/SharePostModal.tsx
kirillkurko: Yes, we need, because responsive modal has max-width: 480px rule (and I guess this is why it was previously done will inline styles), but to convert to pixels and override the rule we need to use important

PR #1096, opened 2022-07-14
fix: refresh token causing boot to refetch

sshanzel:
- The value we use to check how much it should wait for our refresh token to be requested was `undefined` at initialization.
- It produces `NaN` due to arithmetic operation on `undefined`, so the setTimeout was thinking to fire at will.
- Also took the chance to remove the undefined search query parameter in the boot request.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-232 #done

### DIFF
--- a/packages/shared/src/hooks/useRefreshToken.ts
+++ b/packages/shared/src/hooks/useRefreshToken.ts
@@ -2,7 +2,6 @@ import { useEffect } from 'react';
 import { differenceInMilliseconds } from 'date-fns';
 import { AccessToken } from '../lib/boot';
 import useDebounce from './useDebounce';
-import { nextTick } from '../lib/func';
 
 export function useRefreshToken(
   accessToken: AccessToken,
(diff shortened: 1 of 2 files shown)


rebelchris (approved):
Why did we even use this here?
I'm just wanting to make sure this wasn't there for a specific reason 😅

Perhaps Ido might know why it was used before?

sshanzel:
> Perhaps Ido might know why it was used before?

Yes, was having deep thoughts about it as well, but after checking through, I think we can have it here since the computation won't have a major performance issue.

The other option is to create a state so that it would trigger a re-render and the `useDebounce` be alerted for the change of `delay` parameter. This I think is much more complex so I opted for the current change. What do you think?

rebelchris:
Alright, thanks for the additional information.
Let's go for this.

We can see if it doesn't produce other edge-cases

PR #1114, opened 2022-07-25
feat: login and registration including error management and tests

sshanzel:
## Changes

TODO: refetch boot on login and registration

- Check whether email exists before proceeding to go for registration
- Login error management
- Registration error management

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-187 #done


sshanzel:
@rebelchris would love to know what you think about this PR including its test cases if you think it does cover what we need to have.

packages/shared/src/components/auth/AuthDefault.tsx
```diff
@@ -26,18 +30,28 @@ const AuthDefault = ({
   isV2,
 }: AuthDefaultProps): ReactElement => {
   const [shouldLogin, setShouldLogin] = useState(isV2);
+  const { mutateAsync: checkEmail } = useMutation((emailParam: string) =>
+    request(`${apiUrl}/graphql`, QUERY_USER_BY_EMAIL, { email: emailParam }),
+  );
```
rebelchris, line 35: I'm not really sure we should do this.
Think Kratos should have a pre-hook for this and do it their side.

rebelchris (changes requested):
I'm not really sure some of these events should even be handled by us.
Would love to raise this in auth channel first.

Would you mind taking that on?

packages/shared/src/components/auth/AuthDefault.tsx
sshanzel: Totally agree. What do you think of commenting this block out for the meantime, so we can push forward with what we have here?

kirillkurko (approved):

packages/shared/src/components/auth/ForgotPasswordForm.tsx
```diff
@@ -1,26 +1,68 @@
...
+  initialEmail,
   onBack,
   onClose,
 }: ForgotPasswordFormProps): ReactElement {
+  const [hint, setHint] = useState('');
   const [emailSent, setEmailSent] = useState(false);
-  const onSendEmail = () => setEmailSent(true);
+  const { data: recovery } = useQuery('recovery', initializeRecovery, {
+    ...disabledRefetch,
+  });
```
rebelchris, line 38: We'll probably have to introduce a manual refetch based on the resend?
Just noting it here for you to test with. (could be it's multi try)

rebelchris (changes requested):
Nice, we probably need to do a run-over as some have been deprecated/changed.
Let me know if you need any help on this.

I think we can focus on this one first as it's quite impactful.

packages/shared/src/components/auth/ForgotPasswordForm.tsx
sshanzel: we can probably change keys, an incrementing number state of how much the user has sent already so that it would auto-refetch. What do you think?

sshanzel: I'll fix the deprecated change on merge now.

rebelchris: Sure also fine, just make sure it's even needed.
If we don't need to refetch it's obviously better. *(Kratos might keep the CSRF valid)

sshanzel: Yes, sometimes, the flow gets invalid after a period of time (which is tied to the token). I think we will need it.

packages/shared/src/components/auth/ForgotPasswordForm.tsx
```diff
@@ -1,32 +1,85 @@
...
+  const { data: recovery } = useQuery(
+    ['recovery', sentCount],
+    () => initializeKratosFlow(AuthFlow.Recovery),
+    {
+      ...disabledRefetch,
+    },
+  );
+
+  const { mutateAsync: sendEmail, isLoading } = useMutation(
+    (params: ValidateReceoveryParams) => submitKratosFlow(params),
```
rebelchris, line 52:
```suggestion
    (params: ValidateRecoveryParams) => submitKratosFlow(params),
```

Small typo

rebelchris (changes requested):

packages/shared/src/components/auth/ForgotPasswordForm.tsx
```diff
@@ -1,32 +1,85 @@
...
+  );
+
+  const { mutateAsync: sendEmail, isLoading } = useMutation(
+    (params: ValidateRecoveryParams) => submitKratosFlow(params),
+    {
+      onSuccess: ({ error }) => {
+        if (error) {
+          const requestError = getErrorMessage(error.ui.messages);
+          const emailError = getNodeByKey('email', error.ui.nodes);
+          const formError = getErrorMessage(emailError?.messages);
```
rebelchris, line 58: Should we perhaps consider abstracting this in a common function?
Not sure if it's a once-off at this stage.

non blocking.

packages/shared/src/components/auth/LoginForm.tsx
```diff
@@ -52,28 +61,39 @@ function LoginForm({
...
+    };
     onPasswordLogin({ action, params });
   };
 
   return (
-    <AuthForm className="gap-2" onSubmit={onLogin} action="#" ref={formRef}>
+    <AuthForm
+      className="gap-2"
+      onSubmit={onLogin}
+      action="#"
```
rebelchris, line 76: Is this correct?

packages/shared/src/components/auth/LoginForm.tsx
```diff
@@ -52,28 +61,39 @@ function LoginForm({
...
+      <TokenInput token={getNodeValue('csrf_token', login?.ui?.nodes)} />
       <TextField
         leftIcon={<MailIcon />}
         inputId="identifier"
         name="identifier"
         label="Email"
         type="email"
+        saveHintSpace
+        hint={hint}
+        onChange={() => hint && setHint('')}
```
rebelchris, line 89: Just checking should it be empty string or null?
Not sure if it renders the UI now

packages/shared/src/components/fields/PasswordField.tsx
```diff
@@ -48,16 +53,22 @@ export function PasswordField({
       valueChanged={(value) =>
         setPasswordStrengthLevel(passwordStrength(value).id)
       }
-      hint={hint}
+      hint={showStrength ? hint : props.hint}
       validityChanged={setIsValid}
       valid={isValid}
       hintClassName={
-        hasUserAction && passwordStrengthStates[passwordStrengthLevel].className
+        showStrength &&
```
rebelchris, line 60: Seeing you use this multiple times you can actually abstract it to it's own variable in this file.
Will clean up the render bit.

packages/shared/src/hooks/useRegistration.ts
```diff
@@ -92,11 +92,11 @@ const useRegistration = ({
 
   const onValidateRegistration = async (values: RegistrationParameters) => {
     const { nodes, action } = registration.ui;
-    const csrfToken = getNodeByKey('csrf_token', nodes);
     const postData: RegistrationParameters = {
       ...values,
       method: values.method || 'password',
-      csrf_token: csrfToken.attributes.value,
+      csrf_token: getNodeValue('csrf_token', nodes),
+      'traits.image': values['traits.image'] || fallbackImages.avatar,
```
rebelchris, line 99: I think I removed this, because it was always pre-overwritting the social sign-in.
We only need to set this on the actual registration call (the last one)

packages/webapp/pages/reset-password.tsx
```diff
@@ -1,28 +1,97 @@
-import React, { ReactElement } from 'react';
+import React, { FormEvent, ReactElement, useState } from 'react';
```
rebelchris, line 1: General question, can we not abstract the reset password component to work both here and in the UI?
Or atleast the function?

rebelchris (changes requested):
Wow, so much work on this 🤯

packages/webapp/pages/reset-password.tsx
sshanzel: I haven't checked yet how the reset password would look like. I will check how it would be processed then see how we can uniform their calls. Thanks for the mention.

packages/shared/src/components/auth/ForgotPasswordForm.tsx
sshanzel: I think I haven't had the chance that I had to use this one. Though this one is specific for this email form.

rebelchris (approved):
Other than the noticed things nothing from my side.
I didn't really deep-dive in the test to be honest.

sshanzel:
Since I will be working on the refactor of reset password flow in a separate PR. I'll be merging this now, and just create another PR to hook up what Viktor has prepared for calling email existence on registration flow.

PR #1123, opened 2022-08-03
refactor: use registration hook

sshanzel:
- Changes are mostly improving types, cleanups, setting the right parameters for the required traits, some minor bug fix in logout, etc.
- The registration process will be used in different places, I created a hook to the registration auth service for easier use.
- I have also centralized the processing of data for both social provider and password method.
- Would love to hear feedback on areas we can improve 🚢 

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done


packages/shared/src/lib/auth.ts
```diff
@@ -97,18 +133,16 @@ export interface AuthSession extends LogoutSessionData {
...
-  created_at: '2022-07-21T05:18:27.975693Z';
-  updated_at: '2022-07-21T05:18:27.975693Z';
-  refresh: false;
-  requested_aal: 'aal1';
+export interface SuccessfulRegistrationData {
+  session: AuthSession;
+  identity: Identity;
 }
 
+const authUrl = 'http://127.0.0.1:4433';
```
rebelchris, line 141: Probably best to put this in a env actually.

rebelchris (approved):
Nice, thanks a lot for going through these.
I think we can still cleanup some of the usecases, but this start will already really help.

Perhaps we can go over some things in a realtime session at one point.
(non blocking for this)

packages/shared/src/lib/auth.ts
sshanzel: Totally agree. Let's just make this value a fallback if the env is empty. Thank you!

sshanzel:
@rebelchris totally agree. At some point we should do a sync session. Thank you for the review! 🚢

PR #1128, opened 2022-08-07
refactor: security page tabs individual components

sshanzel:
- Created multiple isolated components for easier management.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-301 #done

### DIFF
--- a/packages/webapp/components/layouts/AccountLayout/EmailSentSection.tsx
+++ b/packages/webapp/components/layouts/AccountLayout/EmailSentSection.tsx
@@ -1,0 +1,46 @@
+import { Button } from '@dailydotdev/shared/src/components/buttons/Button';
+import { Overlay } from '@dailydotdev/shared/src/components/utilities';
+import React, { ReactElement } from 'react';
+import { OverlayContainer } from './common';
+
+interface EmailSentSectionProps {
+  onResend?: (e: React.MouseEvent) => void;
+  onCancel?: (e: React.MouseEvent) => void;
+}
+
+function EmailSentSection({
+  onResend,
+  onCancel,
+}: EmailSentSectionProps): ReactElement {
+  return (
+    <OverlayContainer className="mt-6 border-theme-status-warning">
+      <Overlay className="bg-overlay-quaternary-bun" />
+      <p>
+        We sent an email to verify your account. Please check your spam folder
+        if you {`don't`} see the email.
+      </p>
+      <span className="flex flex-row gap-4 mt-4">
+        {onResend && (
+          <Button
+            onClick={onResend}
+            buttonSize="xsmall"
+            className="w-fit btn-primary"
+          >
+            Resend
+          </Button>
+        )}
+        {onCancel && (
+          <Button
+            onClick={onCancel}
+            buttonSize="xsmall"
+            className="w-fit btn-secondary"
+          >
+            Cancel Request
+          </Button>
+        )}
+      </span>
+    </OverlayContainer>
+  );
+}
+
+export default EmailSentSection;
--- a/packages/webapp/components/layouts/AccountLayout/EmailSentSection.tsx
+++ b/packages/webapp/components/layouts/AccountLayout/EmailSentSection.tsx
@@ -1,7 +1,6 @@
+import AlertContainer from '@dailydotdev/shared/src/components/alert/AlertContainer';
 import { Button } from '@dailydotdev/shared/src/components/buttons/Button';
-import { Overlay } from '@dailydotdev/shared/src/components/utilities';
 import React, { ReactElement } from 'react';
-import { OverlayContainer } from './common';
 
 interface EmailSentSectionProps {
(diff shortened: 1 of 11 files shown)


packages/webapp/components/layouts/AccountLayout/Security/index.tsx
```diff
@@ -0,0 +1,142 @@
...
+  isEmailSent?: boolean;
+  onSwitchDisplay: (display: Display) => void;
+}
+
+function AccountSecurityDefault({
+  isEmailSent,
+  onSwitchDisplay,
+}: AccountSecurityDefaultProps): ReactElement {
+  const { user } = useContext(AuthContext);
+  const [email, setEmail] = useState<string>();
```
kirillkurko, line 36: Here we have `string` type, but initial value is `undefined`, my suggestion is to set initial value to empty string, so only string are used for this state

kirillkurko (approved):
Looks good, just left a small comment about type-safety

packages/webapp/components/layouts/AccountLayout/Security/index.tsx
sshanzel: I had the same scenario back then. I am also into using an empty string. Though from a review before, we used null.

@rebelchris what do you think we should go for moving forward with most (if not all) situations like this?`null`, `undefined`, or an empty string?

image

packages/webapp/components/layouts/AccountLayout/EmailSentSection.tsx
```diff
@@ -0,0 +1,46 @@
...
+  onResend?: (e: React.MouseEvent) => void;
+  onCancel?: (e: React.MouseEvent) => void;
+}
+
+function EmailSentSection({
+  onResend,
+  onCancel,
+}: EmailSentSectionProps): ReactElement {
+  return (
+    <OverlayContainer className="mt-6 border-theme-status-warning">
```
rebelchris, line 16: Why don't we leverage modals for this?

packages/webapp/components/layouts/AccountLayout/Security/index.tsx
```diff
@@ -0,0 +1,142 @@
...
+        <Button
+          className="mt-6 w-fit btn-secondary"
+          onClick={() => setResetPasswordSent(true)}
+        >
+          Reset password
+        </Button>
+      </AccountContentSection>
+      {/* )} */}
+      {resetPasswordSent && (
+        <OverlayContainer className="mt-6">
```
rebelchris, line 122: This could be seperate/modal?

rebelchris (changes requested):
Just some questions around the overlay use?

packages/webapp/components/layouts/AccountLayout/Security/index.tsx
rebelchris: We do quite a lot of checks on nullable (most of the time) so I personally also think null is the way to go.
Especially when strong typed with the <string> before it.

But love to hear if there are more benefits to using empty string compared to null?

packages/webapp/components/layouts/AccountLayout/EmailSentSection.tsx
sshanzel: Ah, this is not a modal, this is the faded background overlay in the image below. Probably a poor naming, but I can't think of anything other than "Overlay".
image

packages/webapp/components/layouts/AccountLayout/Security/index.tsx
sshanzel: I only prefer the empty string to infer the type, but I am cool with using null as we've been using it in the past, and being `null` fits the context.

packages/webapp/components/layouts/AccountLayout/Security/index.tsx
sshanzel: image
This overlay portion is the one below.

packages/webapp/components/layouts/AccountLayout/EmailSentSection.tsx
rebelchris: Oh right, yeah overlay seems a tad off then.
Perhaps we can leverage `AlertContainer` we have something similar coming up for the community picks.
Although that has a border side to it.
(See here: https://dailydotdev.slack.com/archives/C03JJ8CBEH3/p1659967066043279)

Do you think we can make one generic "AlertContainer" that can handle all of these?

sshanzel: I think, yes. Alert container sounds about right since we are trying to get the user's attention with this. An alert fits the context. Will refactor this.

rebelchris (approved):
Sweet refactor!
Nice new component we can leverage 💖

PR #1131, opened 2022-08-12
feat: account recovery

sshanzel:
## Changes
Didn't realize this PR for integration wasn't created.

- The changing of password for when the user has clicked the recovery link

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-185 #done


packages/shared/src/lib/auth.ts
```diff
@@ -374,6 +374,41 @@ export const errorsToJson = <T extends string>(
...
+export const getSettingsFlow = async (
+  id: string,
+): Promise<InitializationData> => {
+  const res = await fetch(`${authUrl}/self-service/settings/flows?id=${id}`, {
+    credentials: 'include',
+    headers: { Accept: 'application/json' },
+  });
+
+  return res.json();
+};
```
rebelchris, line 386: This is basically the same as `getRegistrationFlow` note to clean this up.
(No need to change now as agreed, but we musn't forget.

packages/webapp/pages/reset-password.tsx
```diff
@@ -1,28 +1,90 @@
...
+  const onSubmit = (e: FormEvent<HTMLFormElement>) => {
+    e.preventDefault();
+    const { password } = formToJson<{ password: string }>(e.currentTarget);
+    const params: RegistrationParameters = {
+      method: 'password',
+      csrf_token: getNodeValue('csrf_token', data.ui.nodes),
+      'traits.email': getNodeValue('traits.email', data.ui.nodes),
+      'traits.fullname': getNodeValue('traits.fullname', data.ui.nodes),
+      'traits.username': getNodeValue('traits.username', data.ui.nodes),
+      'traits.image': getNodeValue('traits.image', data.ui.nodes),
```
rebelchris, line 54: Why does the reset password flow care about all these fields? that seems weird?

rebelchris (changes requested):
Just one small note.

packages/shared/src/lib/auth.ts
sshanzel: Ah, yes. I think the only difference is the endpoint, from `/registration` to `settings`. We can probably create an enum for available `AuthFlow`. We can probably start the refactoring on the ticket we have for kratos lib we have planned.

packages/webapp/pages/reset-password.tsx
sshanzel: The account recovery flow would allow you to update your profile details as well, basically, the settings flow, but with privileged access already. So to validate the transaction it requires all the necessary information if some of the fields are required.

packages/shared/src/lib/auth.ts
rebelchris: Good idea let's add a task for it.

rebelchris (approved):
Let's go 💪

packages/shared/src/lib/auth.ts
sshanzel: I think this ticket we created before should be suitable for the things to work on. I'll add a comment referencing to our conversation here.

https://dailydotdev.atlassian.net/jira/software/projects/WT/boards/11?selectedIssue=WT-344

PR #1135, opened 2022-08-15
feat: introducing new password guideline

rebelchris:
- New password textfield guidelines
- Decided to do most logic in the textfield as it's guidelined

## Events

## Manual Testing

https://user-images.githubusercontent.com/554874/184629527-2ba00e0a-ccef-4103-b4ad-287aad82ab4e.mp4

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-337 #done


idoshamun (commented):
Can we perhaps create a new component that will rely on the `TextField` component but just for password? By extending its functionality for example.
Given that password fields are rarely used, adding a new package to the `TextField` which is something very fundamental seems like too much to me. What do you think?

rebelchris:
@idoshamun Refactored to a password field.
The only thing I noticed is that we currently bubble programmtic events which all trigger the validation.

See here: no user interaction was done.
(Is this something we want?)
Screenshot 2022-08-16 at 15 58 34

idoshamun:
I think it has its reasons but I'm not sure why...
The only use case that comes to mind is when the field is already populated and we want to trigger validation in case this field is marked as required. But if initially the field is empty we don't have to bubble it

packages/shared/src/components/fields/PasswordField.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+export function PasswordField({
+  type,
+  ...props
+}: TextFieldProps): ReactElement {
+  const [useType, setUseType] = useState(type);
+  const [passwordStrengthLevel, setPasswordStrengthLevel] = useState<number>(0);
+  const [isValid, setIsValid] = useState<boolean>(null);
+  const hasUserAction = isValid !== null;
+  const hint = () => {
```
idoshamun, line 38: Why do we need it as a function?

rebelchris: We don't, changed it to conditions now :)

idoshamun (approved):

PR #1139, opened 2022-08-15
feat: accounts profile integration

sshanzel:
- Able to update the profile information in API service.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done


rebelchris (changes requested):
@sshanzel I'm pretty sure we only need to update the API for the profile information.
Only the email change (separate flow) should notify both.

packages/shared/src/lib/auth.ts
```diff
@@ -295,6 +307,42 @@ export const logoutSession = (
     headers: { Accept: 'application/json', 'X-CSRF-Token': csrf_token },
   });
 
+export const initializeSettings = async (): Promise<InitializationData> => {
```
rebelchris, line 310: Do we still need this?

rebelchris (changes requested):
Nice, quick fix 🙏

Just one note on the settings, unsure if we still need it?

packages/shared/src/lib/auth.ts
sshanzel: Should be not, I think. Let me remove this one quickly! Thanks for the fast review! 🚢

packages/shared/src/components/auth/RegistrationForm.tsx
```diff
@@ -84,7 +84,9 @@ export const RegistrationForm = ({
         ref={formRef}
         onSubmit={onSubmit}
       >
-        {socialAccount && <ImageInput initialValue={socialAccount.image} />}
+        {socialAccount && (
```
rebelchris, line 87: We need to ensure that this is non-editable for the social flow.
See WT-348, (just a reminder if you can pick that one up after this?)

rebelchris (approved):
Awesome work, really happy with this one so far

packages/shared/src/components/auth/RegistrationForm.tsx
sshanzel: Would be more than happy to take it 🚢

packages/shared/src/components/auth/AuthOptions.tsx
```diff
@@ -64,13 +64,9 @@ function AuthOptions({
...
     hasLoggedOut() ? Display.SignBack : defaultDisplay,
   );
-  const { onUpdateSession } = useContext(AuthContext);
   const { validateRegistration, onSocialRegistration } = useRegistration({
     key: 'registration_form',
-    onValidRegistration: (data) => {
-      onUpdateSession(data);
-      setActiveDisplay(Display.EmailSent);
-    },
+    onValidRegistration: () => setActiveDisplay(Display.EmailSent), // on valid registration get boot
```
rebelchris, line 69: Are we stilling missing this boot call?

packages/shared/src/components/auth/LoginForm.tsx
```diff
@@ -40,12 +37,7 @@ function LoginForm({
   );
   const { mutateAsync: onPasswordLogin } = useMutation(
     (params: ValidateLoginParams) => submitKratosFlow(params),
-    {
-      onSuccess: ({ data }) => {
-        const session = data as AuthSession;
-        onUpdateSession(session);
-      },
-    },
+    // { onSuccess: ({ data }) => {} }, // refetch boot after login
```
rebelchris, line 40: Same here, are you tackling this seperate?

rebelchris (changes requested):
Almost there 👀

packages/shared/src/components/auth/LoginForm.tsx
sshanzel: Yes, it will be part of the held off PR for login and registration.

PR #1140, opened 2022-08-16
refactor: uniform kratos endpoints

sshanzel:
- A kratos library that has 3 major functions, generic ones to initialize, get, and submit flow.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done


packages/shared/src/contexts/AuthContext.tsx
```diff
@@ -69,7 +70,14 @@ const appendLogoutParam = (link: string): URL => {
...
 const logout = async (logoutUrl: string, token: string): Promise<void> => {
-  await Promise.all([dispatchLogout(), logoutSession(logoutUrl, token)]); // temporary
+  await Promise.all([
+    dispatchLogout(),
+    submitKratosFlow({
+      action: logoutUrl,
+      params: { csrf_token: token },
+      method: 'GET',
+    }),
+  ]); // temporary
```
rebelchris, line 80: The API logout already handles Kratos logout, no need to extra call it.

packages/shared/src/lib/kratos.ts
```diff
@@ -0,0 +1,247 @@
...
+export enum AuthFlow {
+  Login = '/login',
+  Logout = '/logout',
+  Registration = '/registration',
+  Settings = '/settings',
+  Recovery = '/recovery',
+  Verification = '/verification',
+}
+
+const authUrl = process.env.NEXT_PUBLIC_AUTH_URL || 'http://127.0.0.1:4433';
```
rebelchris, line 158: We probably want to move this away to a const file

packages/shared/src/lib/kratos.ts
```diff
@@ -0,0 +1,247 @@
...
+}
+
+export interface SuccessfulRegistrationData {
+  session: AuthSession;
+  identity: Identity;
+}
+
+export enum AuthFlow {
+  Login = '/login',
+  Logout = '/logout',
```
rebelchris, line 151: Don't think we need a separate flow for this.

packages/shared/src/lib/kratos.ts
```diff
@@ -0,0 +1,247 @@
...
+  Registration = '/registration',
+  Settings = '/settings',
```
rebelchris, line 153: What falls under settings flow?

rebelchris (changes requested):
Nice work Lee!
Just some questions around things we probably no longer need.

packages/shared/src/lib/kratos.ts
sshanzel: The reset password should be under it, it is technically a settings flow.

packages/shared/src/contexts/AuthContext.tsx
sshanzel: Awesome. We should remove this then. Thank you for the confirmation.

packages/shared/src/lib/kratos.ts
sshanzel: Totally agree.

rebelchris (approved):
Awesome! 💖

PR #1146, opened 2022-08-18
feat: store last sign-in method

sshanzel:
- Store the last clicked provider used.
- When the user opens the modal that is only when it will be checked for which one was used the last one.
- Once the user has seen the "sign back in" form, it will remove the storage value so next signin will include allowing of registration.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-341 #done

### DIFF
--- a/packages/shared/src/components/auth/AuthDefault.tsx
+++ b/packages/shared/src/components/auth/AuthDefault.tsx
@@ -1,8 +1,10 @@
 import React, { ReactElement, ReactNode, useState } from 'react';
+import { storageWrapper as storage } from '../../lib/storageWrapper';
 import { CloseModalFunc } from '../modals/common';
 import AuthModalFooter from './AuthModalFooter';
 import AuthModalHeader from './AuthModalHeader';
 import AuthModalHeading from './AuthModalHeading';
+import { SIGNIN_METHOD_KEY } from './AuthSignBack';
 import { ColumnContainer, providers } from './common';
 import EmailSignupForm from './EmailSignupForm';
 import LoginForm from './LoginForm';
@@ -51,6 +53,11 @@ const AuthDefault = ({
     return onSignup(input.value.trim());
   };
 
+  const onSocialClick = (provider: string) => {
+    storage.setItem(SIGNIN_METHOD_KEY, provider);
+    onProviderClick(provider);
+  };
+
   return (
     <>
       <AuthModalHeader title="Sign up to daily.dev" onClose={onClose} />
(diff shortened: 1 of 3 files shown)


packages/shared/src/components/auth/AuthDefault.tsx
```diff
@@ -51,6 +53,11 @@ const AuthDefault = ({
     return onSignup(input.value.trim());
   };
 
+  const onSocialClick = (provider: string) => {
+    storage.setItem(SIGNIN_METHOD_KEY, provider);
+    onProviderClick(provider);
+  };
```
rebelchris, line 59: We also need this on the register calls, perhaps we can do it inside the final onRegister and onLogin?

I would also suggest if we can convert to cookie?
That way it will sync across apps

rebelchris: Thinking about it, can't we have Kratos set a cookie?

rebelchris (changes requested):
Some comments, but amazing work already

packages/shared/src/components/auth/AuthDefault.tsx
sshanzel: But for login, I think there should not be any difference in the UI since only the placement of the providers was arranged. What do you think?

sshanzel: > Thinking about it, can't we have Kratos set a cookie?

I don't want to involve Kratos now as much as possible, to be honest 🤣

rebelchris: Not sure I understand the login query.
To me it should be for register and login -> set cookie -> next login we'll show the different login UI?

sshanzel: Yes, that's correct. Though I thought it would only be when the user logged in with the social provider. Let get back to the specs again.

sshanzel: For this one, I'm not sure which social provider we would display if the user logged out from password method.

image

rebelchris: Let's raise it for product/design please 🙏

sshanzel: After discussing with Tomer, we can go with the default for if the last used method was login.

https://dailydotdev.slack.com/archives/C03KATZEKDW/p1660933534458229?thread_ts=1660900677.896609&cid=C03KATZEKDW

rebelchris: I think we had a misunderstanding here.
My point was more around the social flow, both register and signin with social should remember the state.
Not sure if that's the case now.

And still think we should aim for a cookie to do cross system?

sshanzel: Ah, yes. This should remember the state. The current flow is when the user clicked any provider it would already register what they chose even before finishing the login/registration.

rebelchris (approved):

PR #1151, opened 2022-08-24
feat: privileged session modal

sshanzel:
WIP: Tests

- New custom hook for login for reusability
- A hook for privileged sessions for future usage.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-362 #done

### DIFF
--- a/packages/shared/src/components/auth/AuthDefault.tsx
+++ b/packages/shared/src/components/auth/AuthDefault.tsx
@@ -23,6 +23,7 @@ interface AuthDefaultProps {
   title?: string;
   providers: Provider[];
   disableRegistration?: boolean;
+  disablePassword?: boolean;
 }
 
 const AuthDefault = ({
@@ -35,6 +36,7 @@ const AuthDefault = ({
   isV2,
   providers,
   disableRegistration,
+  disablePassword,
   title = 'Sign up to daily.dev',
 }: AuthDefaultProps): ReactElement => {
   const [shouldLogin, setShouldLogin] = useState(isV2);
@@ -68,10 +70,36 @@ const AuthDefault = ({
     onProviderClick(provider);
   };
 
+  const getForm = () => {
+    if (disablePassword && disableRegistration) {
+      return null;
+    }
+
+    if (shouldLogin || disableRegistration) {
+      return (
+        <LoginForm
+          loginHint={loginHint}
+          onPasswordLogin={onPasswordLogin}
+          onForgotPassword={onForgotPassword}
+        />
+      );
+    }
+
+    return <EmailSignupForm onSubmit={onEmailSignup} isV2={isV2} />;
+  };
+
+  const getOrDivider = () => {
+    if (isV2 || !providers.length || disablePassword) {
+      return null;
+    }
+
+    return <OrDivider />;
+  };
+
   return (
     <>
       <AuthModalHeader title={title} onClose={onClose} />
-      <ColumnContainer>
+      <ColumnContainer className={disableRegistration && 'mb-6'}>
         {isV2 && (
           <AuthModalHeading
             tag="h2"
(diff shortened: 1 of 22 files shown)


packages/shared/src/components/auth/AuthDefault.tsx
```diff
@@ -1,38 +1,48 @@
...
+  providers,
+  disableRegistration,
+  disablePassword,
+  title = 'Sign up to daily.dev',
 }: AuthDefaultProps): ReactElement => {
   const [shouldLogin, setShouldLogin] = useState(isV2);
-  const { mutateAsync: checkEmail } = useMutation((emailParam: string) =>
-    checkKratosEmail(emailParam),
-  );
+  // const { mutateAsync: checkEmail } = useMutation((emailParam: string) =>
```
sshanzel, line 43: Will uncomment this before merging to auth intermediate branch.

packages/shared/src/hooks/usePrivilegedSession.ts
```diff
@@ -0,0 +1,98 @@
...
+          const url = new URL(redirect);
+          const returnTo = new URL(url.searchParams.get('return_to'));
+          const flowId = returnTo.searchParams.get('flow');
+          return setVerifySessionId(flowId);
+        }
+
+        return onResetSuccessful?.();
+      },
+    },
+  );
```
rebelchris, line 60: I'm not a 100% sure this should be in this file.
It's a specific action and can work independently from the privileged session

packages/shared/src/hooks/usePrivilegedSession.ts
```diff
@@ -0,0 +1,98 @@
...
+
+  const { mutateAsync: resetPassword } = useMutation(
+    (params: ValidateResetPassword) => submitKratosFlow(params),
+    {
+      onSuccess: async ({ redirect }) => {
+        if (redirect) {
+          const url = new URL(redirect);
+          const returnTo = new URL(url.searchParams.get('return_to'));
+          const flowId = returnTo.searchParams.get('flow');
+          return setVerifySessionId(flowId);
```
rebelchris, line 54: I see we also mix two ways of keeping track of redirect, I think we should really stick to one way for clarity

rebelchris (changes requested):
I think in general this is a really great PR.
I would just like to see more distinguishment between the actions.

The privileged actions should work on their own, only when it's required should they come in play and hijack the flow I think.

Let me know what you think

packages/shared/src/hooks/usePrivilegedSession.ts
sshanzel: Totally agree, just thought this will probably be used in some areas as well and will definitely use the privileged session. But we can take this out of here, for sure.

packages/webapp/pages/account/security.tsx
```diff
@@ -2,16 +2,82 @@ import TabContainer, {
...
+    onSocialLogin,
+  } = usePrivilegedSession();
+
+  const { mutateAsync: resetPassword } = useMutation(
+    (params: ValidateResetPassword) => submitKratosFlow(params),
+    {
+      onSuccess: async ({ redirect }) => {
+        if (redirect) {
+          return initializePrivilegedSession(redirect);
+        }
```
rebelchris, line 66: I think we have to distinguish if it's a privileged or not.
You could still be valid not to have privileged,  but still need to redirect.

rebelchris (changes requested):
Almost there

packages/webapp/pages/account/security.tsx
sshanzel: Ah, good point. We need to check the given url if it is a login one.

sshanzel: We can still call this now, but I have placed a condition checker in the custom hook so we won't have to check it every implementation.

rebelchris (approved):
Ok this will work.
But you might get in trouble with the social linking, as it's privileged and uses a redirect regardless of it being privileged.

For now this is ok, but putting it out here as a concern.
Also don't forget to comment back the heimdall calls.

sshanzel:
> Ok this will work.
But you might get in trouble with the social linking, as it's privileged and uses a redirect regardless of it being privileged.

Okay, cool. I will look into the social linking process of how it happens after merging this (and doing the uncomment fixing the tests).

PR #1156, opened 2022-08-25
feat: add link provider error message

rebelchris:
- Provide error message if linked account already exists

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-305 #done


packages/shared/src/lib/kratos.ts
```diff
@@ -209,6 +209,17 @@ export const getKratosFlow = async (
...
+  flow: AuthFlow,
+  id: string,
+): Promise<InitializationData> => {
+  const res = await fetch(`${authUrl}/self-service${flow}/flows?id=${id}`, {
+    credentials: 'include',
+    headers: { Accept: 'application/json' },
+  });
+  return res.json();
+};
+
```
sshanzel, line 222: I think this is the same as the generic `getKratosFlow`

packages/webapp/components/layouts/AccountLayout/Security/index.tsx
```diff
@@ -65,13 +69,27 @@ function AccountSecurityDefault({
...
   const { data: userProviders } = useQuery('providers', getKratosProviders);
   const { data: settings } = useQuery('settings', () =>
     initializeKratosFlow(AuthFlow.Settings),
   );
 
+  useWindowEvents('message', async (e) => {
+    if (e.data?.flow) {
+      const flow = await getKratosSettingsFlow(AuthFlow.Settings, e.data.flow);
+      const { ui } = flow;
+      const error = ui.messages[0]?.id;
```
sshanzel, line 85: We should also check the type for the `message` property as it can be sometimes just an info message for a successful event.

sshanzel (approved):
Just two minor things, but it is so amazing to see the link/unlink come together! 🚢 🚢 🚢

packages/shared/src/lib/kratos.ts
rebelchris: Nope the format is different, really annoying.

packages/webapp/components/layouts/AccountLayout/Security/index.tsx
rebelchris: Oh good point, not sure what it should do.

sshanzel: Oh, I just realized you compared the ID already, and for info related ones, I think it will have a different ID. This should be good, I think.

packages/shared/src/lib/kratos.ts
sshanzel: Oh, wow. The only difference is the search parameter. Lol.

PR #1161, opened 2022-08-30
refactor: use window events to require key

sshanzel:
- Create a hook that wraps the usage of window event to utilize an event key in determining if the call should proceed.
- Refactor old usage of window event to use the new one.
- Enforce sending an event key on post message function.
- Fix persisting the session verification modal on reload by making it live through current session only.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-378 #done

### DIFF
--- a/packages/shared/src/hooks/useLogin.ts
+++ b/packages/shared/src/hooks/useLogin.ts
@@ -9,6 +9,7 @@ import {
   ValidateLoginParams,
 } from '../lib/auth';
 import {
+  AuthEvent,
   AuthFlow,
   EmptyObjectLiteral,
   getKratosSession,
@@ -16,7 +17,7 @@ import {
   initializeKratosFlow,
   submitKratosFlow,
 } from '../lib/kratos';
-import useWindowEvents from './useWindowEvents';
+import { useKeyedWindowEvent } from './useWindowEvents';
 
 interface UseLogin {
   loginFlowData: InitializationData;
@@ -97,12 +98,17 @@ const useLogin = ({
     onPasswordLogin({ action, params });
   };
 
-  useWindowEvents('message', async (e) => {
-    if (!e.data?.flow) {
+  useKeyedWindowEvent('message', AuthEvent.Login, async () => {
+    if (!session) {
       return;
     }
 
     const verified = await getKratosSession();
+
+    if (!verified) {
+      return;
+    }
+
     const hasRenewedSession =
       session.authenticated_at !== verified.authenticated_at;
 
--- a/packages/shared/src/hooks/useLogin.ts
+++ b/packages/shared/src/hooks/useLogin.ts
@@ -17,7 +17,7 @@ import {
   initializeKratosFlow,
   submitKratosFlow,
 } from '../lib/kratos';
-import { useKeyedWindowEvent } from './useWindowEvents';
+import useWindowEvents from './useWindowEvents';
 
 interface UseLogin {
   loginFlowData: InitializationData;
@@ -98,7 +98,7 @@ const useLogin = ({
     onPasswordLogin({ action, params });
   };
 
-  useKeyedWindowEvent('message', AuthEvent.Login, async () => {
+  useWindowEvents('message', AuthEvent.Login, async () => {
     if (!session) {
       return;
     }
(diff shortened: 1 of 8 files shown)


packages/shared/src/hooks/useWindowEvents.ts
```diff
@@ -1,14 +1,44 @@
...
+interface EventData {
+  eventKey?: AuthEvent | string;
+}
+
+type EventParameter = keyof WindowEventMap;
+type KeyedWindowEventHandler<T = unknown> = (e: MessageEvent<T>) => void;
+
+export interface SocialRegistrationFlow extends EventData {
+  flow: string;
+}
```
rebelchris, line 13: This should not be part of this hook as it's implementation orientated.

packages/shared/src/hooks/useWindowEvents.ts
```diff
@@ -1,14 +1,44 @@
...
+import { useCallback, useEffect } from 'react';
+import { AuthEvent } from '../lib/kratos';
 
-// to make intellisense work on the first parameter
-const useWindowEvents: typeof window.addEventListener = (
-  event: string,
-  func: (e: Event) => void,
+interface EventData {
+  eventKey?: AuthEvent | string;
+}
```
rebelchris, line 6: This one actually also is implementation wide.

packages/shared/src/hooks/useWindowEvents.ts
```diff
@@ -1,14 +1,44 @@
...
+  event: EventParameter,
+  func: KeyedWindowEventHandler<T>,
 ): void => {
   useEffect(() => {
     window.addEventListener(event, func);
     return () => window.removeEventListener(event, func);
   }, [event, func]);
 };
 
+export const useKeyedWindowEvent = <T extends EventData = EventData>(
```
rebelchris, line 25: Should we not just deprecate the above one in favor of making this one "useWindowEvent" just so we can sole rely on it?

packages/shared/src/hooks/useLogin.ts
```diff
@@ -97,12 +98,17 @@ const useLogin = ({
     onPasswordLogin({ action, params });
   };
 
-  useWindowEvents('message', async (e) => {
-    if (!e.data?.flow) {
+  useKeyedWindowEvent('message', AuthEvent.Login, async () => {
```
rebelchris, line 101: I'm still pretty sure this hook doesn't have to have this window event.

rebelchris (changes requested):
Awesome, really great work and would love to see this as our default actually.

packages/shared/src/hooks/useLogin.ts
sshanzel: Once you finish the social provider login it will redirect you to our callback which this is our way of letting the tab know it has finished.

packages/shared/src/hooks/useWindowEvents.ts
sshanzel: Good point. Let me simplify.

packages/shared/src/hooks/useLogin.ts
rebelchris: I haven't seen this being fired, is it only with the login flow?

sshanzel: The event is being called from here in the callback page.

https://github.com/dailydotdev/apps/blob/8b064d112a8471803b0ba87f437d9d14b6eb69b9/packages/webapp/pages/callback.tsx#L5-L17

sshanzel: Oh, wait. I haven't pushed a config in heimdall to make this work 🤦

packages/shared/src/hooks/useWindowEvents.ts
sshanzel: I think this one has to be here since this is the base requirement of this specific hook.

rebelchris (approved):
Alright, with this Heimdall fix it makes more sense.
Good thing we added the key check :)

PR #1175, opened 2022-09-05
fix: account pages responsiveness

sshanzel:
- Fixes the responsiveness of the account page container
- Fixes the image input component

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done

### DIFF
--- a/packages/shared/src/components/auth/AuthModal.tsx
+++ b/packages/shared/src/components/auth/AuthModal.tsx
@@ -49,6 +49,7 @@ export default function AuthModal({
       onRequestClose={onDiscardAttempt}
       className={classNames(styles.authModal, className)}
       contentClassName={classNames('auth', authVersion)}
+      style={{}}
     >
       {isV1 && (
         <>
--- a/packages/webapp/components/layouts/AccountLayout/common.tsx
+++ b/packages/webapp/components/layouts/AccountLayout/common.tsx
@@ -95,7 +95,7 @@ export const accountPage: Record<AccountPage, AccountPageProps> = {
 export const AccountPageContent = classed(
   'div',
   pageBorders,
-  'flex flex-col w-full max-w-[40rem] tablet:border-l mr-auto',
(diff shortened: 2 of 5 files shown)


packages/webapp/components/layouts/AccountLayout/common.tsx
```diff
@@ -95,7 +95,7 @@ export const accountPage: Record<AccountPage, AccountPageProps> = {
 export const AccountPageContent = classed(
   'div',
   pageBorders,
-  'flex flex-col w-full max-w-[40rem] tablet:border-l mr-auto',
+  'flex flex-col w-full max-w-[calc(100vw-19.75rem)] laptop:max-w-[calc(100vw-36rem)] laptopL:max-w-[40rem] tablet:border-l mr-auto',
```
rebelchris, line 98: Should this really grow as wide?
Perhaps double check with design on that.

rebelchris (approved):
Looks good, but unsure if it's intended from design perspective.

packages/shared/src/components/auth/AuthModal.tsx
```diff
@@ -49,6 +49,7 @@ export default function AuthModal({
       onRequestClose={onDiscardAttempt}
       className={classNames(styles.authModal, className)}
       contentClassName={classNames('auth', authVersion)}
+      style={{}}
```
sshanzel, line 52: Since we introduced a default styling from one of the recent PRs, we need to override it since we already apply the styling in a dedicated css module in this case.

packages/webapp/components/layouts/AccountLayout/common.tsx
sshanzel: I agree it should actually be 640px, but they insisted back then to have this new layout size. With that, I messaged in the channel to confirm with them.

sshanzel: The code `laptopL:max-w-[40rem]` should handle the larger screens setting a specific width.

rebelchris (approved):
Awesome!

PR #1176, opened 2022-09-05
Wt 399 new auth flow

rebelchris:
- New social flow

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-399 #done

### DIFF
--- a/packages/shared/src/lib/auth.ts
+++ b/packages/shared/src/lib/auth.ts
@@ -48,6 +48,7 @@ export interface RegistrationParameters {
   'traits.userId'?: string;
   'traits.name'?: string;
   'traits.username': string;
+  file?: string;
   'traits.image': string;
 }
 
--- a/packages/shared/src/lib/auth.ts
+++ b/packages/shared/src/lib/auth.ts
@@ -38,6 +38,12 @@ export interface ValidateEmailParameters extends AuthPostParams {
 
 export type VerificationParams = KratosFormParams<AccountRecoveryParameters>;
 
+export interface SocialRegistrationParameters {
+  name?: string;
+  username?: string;
+  file?: string;
+}
+
 export interface RegistrationParameters {
   csrf_token: string;
   provider?: string;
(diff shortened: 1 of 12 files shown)


packages/shared/src/lib/auth.ts
```diff
@@ -48,6 +48,7 @@ export interface RegistrationParameters {
   'traits.userId'?: string;
   'traits.name'?: string;
   'traits.username': string;
+  file?: string;
```
sshanzel, line 51: What is the file we send here?

packages/webapp/pages/_app.tsx
```diff
@@ -80,30 +80,6 @@ function InternalApp({ Component, pageProps, router }: AppProps): ReactElement {
 
   useTrackPageView();
 
-  useEffect(() => {
```
sshanzel, line 83: Finally removed this block of code ❤️

packages/shared/src/lib/kratos.ts
```diff
@@ -4,7 +4,7 @@ import { authUrl, heimdallUrl } from './constants';
 export type EmptyObjectLiteral = Record<string, never | string>;
 
 export interface SocialRegistrationFlow extends MessageEventData {
-  flow: string;
+  social_registration: boolean;
```
sshanzel, line 7: What happens to the linking of accounts? That goes through here as well and it requires the flow to be present for checking any error.

sshanzel (commented):
Just had a question around linking/unlinking. But the rest looks amazing to me 🚢

packages/shared/src/lib/kratos.ts
sshanzel: I think we can keep them both but be nullable properties.

sshanzel (approved):
Let's go! 🚢

PR #1180, opened 2022-09-07
feat: info confirmed false completion

sshanzel:
- Replace the redirect mechanism to display a modal to update missing information.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-405 #done


rebelchris (changes requested):
I guess this will work, but only on mainLayout.
Can we not capture it in the already opened AuthModal since that won't close?

sshanzel:
> Can we not capture it in the already opened AuthModal since that won't close?

The auth modal I believe closes when the user data is present. So I think we can't use the AuthModal.

packages/shared/src/components/MainLayout.tsx
```diff
@@ -235,6 +247,19 @@ export default function MainLayout({
           onRequestClose={() => setIsIncompleteRegistration(false)}
         />
       )}
+      {user && !user.infoConfirmed && (
+        <StyledModal
+          isOpen
+          onRequestClose={() => setIsIncompleteRegistration(false)}
```
rebelchris, line 253: You can actually simply log them out here. It's most likely a ghost user, and they should not be able to continue browsing.

rebelchris (changes requested):
Just one more thing.

packages/shared/src/components/MainLayout.tsx
sshanzel: Ah, true. We should just do that. Thanks for the note.

packages/shared/src/components/MainLayout.tsx
```diff
@@ -235,6 +247,16 @@ export default function MainLayout({
           onRequestClose={() => setIsIncompleteRegistration(false)}
         />
       )}
+      {user && !user.infoConfirmed && (
+        <StyledModal isOpen onRequestClose={() => logout()}>
+          <SocialRegistrationForm
+            provider=""
```
rebelchris, line 253: Maybe just best to make it optional then?

rebelchris (approved):
Just one little note :)

packages/shared/src/components/MainLayout.tsx
sshanzel: Totally agree with that.

PR #1193, opened 2022-09-11
refactor: styled modal class selector

sshanzel:
- The reason why some of the modal changes when another smaller popup shows is the overriding of styles and having the same class selector.
- To make it easier to override the base styles, we now instead utilize inline tailwind classes, since CSS module usage has heavier selectors.
- Probably one thing to do as well is to refactor the existing usages of `global` in our CSS modules that tend to conflict with one another.
- Would love to hear some thoughts on this.

Preview:
image

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-437 #done


packages/shared/src/components/modals/StyledModal.tsx
```diff
@@ -1,40 +1,31 @@
...
-        },
-      }}
-      className={classNames('focus:outline-none modal', contentClassName)}
+      portalClassName={className?.toString()}
+      overlayClassName={classNames(
+        'overlay flex fixed flex-col justify-center inset-0 max-h-[100vh] items-center px-5 bg-theme-overlay-quaternary z-[10]',
+        overlayClassName,
+      )}
+      className={classNames(
+        'focus:outline-none modal flex flex-col relative w-full max-w-[26.25rem] max-h-[100%] overflow-y-auto items-center bg-theme-bg-tertiary shadow-2 border border-theme-divider-secondary rounded-16',
```
sshanzel, line 25: Fully converted the content of the styled modal css module

rebelchris (commented):
Nice let's try the onboarding modal on iPhone first as this was where the fix was for initially.
I'll get back to you after testing it.

sshanzel:
> Nice let's try the onboarding modal on iPhone first as this was where the fix was for initially.
I'll get back to you after testing it.

Nice, thank you for mentioning as I only tried it on android. I will let you know soon as I test it in iOS.

rebelchris (approved):
@sshanzel Just tested the flows on desktop and iOS chrome + safari, they seem solid from this side! 🥳

PR #1225, opened 2022-09-15
fix: timezone modal design issues

sshanzel:

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-500 #done
WT-501 #done
WT-502 #done
WT-503 #done

### DIFF
--- a/packages/shared/src/components/auth/IncompleteRegistrationModal.tsx
+++ b/packages/shared/src/components/auth/IncompleteRegistrationModal.tsx
@@ -49,13 +49,14 @@ function IncompleteRegistrationModal({
       {...props}
       onRequestClose={onRequestClose}
       contentClassName="flex flex-col"
+      style={{ content: { overflow: 'hidden' } }}
     >
       <header className="flex relative justify-between items-center p-3 w-full border-b border-theme-divider-tertiary">
-        <span className="font-bold typo-body">Set your timezone</span>
+        <span className="ml-2 font-bold typo-body">Set your timezone</span>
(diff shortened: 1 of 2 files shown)


packages/shared/src/components/auth/IncompleteRegistrationModal.tsx
```diff
@@ -49,13 +49,14 @@ function IncompleteRegistrationModal({
       {...props}
       onRequestClose={onRequestClose}
       contentClassName="flex flex-col"
+      style={{ content: { overflow: 'hidden' } }}
     >
       <header className="flex relative justify-between items-center p-3 w-full border-b border-theme-divider-tertiary">
-        <span className="font-bold typo-body">Set your timezone</span>
+        <span className="ml-2 font-bold typo-body">Set your timezone</span>
```
rebelchris, line 55: I think we are stilling missing the height of the header to be 50px not?

rebelchris (changes requested):
Think we're only missing the height still, but maybe I missed it in a previous one.

packages/shared/src/components/auth/IncompleteRegistrationModal.tsx
sshanzel: Ah, nice catch. I totally forgot about which one to make it 50px. It is the height indeed.

rebelchris (approved):
Wohoo

PR #1246, opened 2022-09-20
Wt 439 safari bug

rebelchris:
- Safari has to open the window outside any async calls, this way the window is not considered a blocking popup

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-439 #done

### DIFF
--- a/packages/shared/src/components/auth/AuthOptions.tsx
+++ b/packages/shared/src/components/auth/AuthOptions.tsx
@@ -2,6 +2,7 @@ import React, {
   MutableRefObject,
   ReactElement,
   useContext,
+  useRef,
   useState,
 } from 'react';
 import classNames from 'classnames';
@@ -86,6 +87,7 @@ function AuthOptions({
     onUpdateHint,
     isLoading: isProfileUpdateLoading,
   } = useProfileForm();
+  const windowPopup = useRef<Window>(null);
 
   const { registration, validateRegistration, onSocialRegistration } =
     useRegistration({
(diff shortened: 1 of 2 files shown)


packages/shared/src/components/auth/AuthOptions.tsx
```diff
@@ -118,7 +123,6 @@ function AuthOptions({
           name: getNodeValue('traits.name', connected.ui.nodes),
           email: getNodeValue('traits.email', connected.ui.nodes),
           image: getNodeValue('traits.image', connected.ui.nodes),
-          flowId: connected.id,
```
sshanzel, line 121: I think this was not intended.

packages/shared/src/components/auth/AuthOptions.tsx
```diff
@@ -225,7 +229,7 @@ function AuthOptions({
         <Tab label={Display.SignBack}>
           <AuthSignBack
             onRegister={() => setActiveDisplay(Display.Default)}
-            onProviderClick={onSocialRegistration}
+            onProviderClick={onProviderClick}
```
sshanzel, line 232: Not sure if this was also part of the debugging phase.

sshanzel (approved):
Whoa! Can't believe this is all for fixing the Safari issue 🤯 

Other than the comments I have nothing to stop this amazing fix! 🚢

packages/shared/src/components/auth/AuthOptions.tsx
rebelchris: No this needs to exist for the fix on login to work (not even sure why it was a seperate action)

packages/shared/src/components/auth/AuthOptions.tsx
rebelchris: This should just be the same as the main branch, not sure why there was some change in my commit history

PR #1248, opened 2022-09-20
fix: accounts security

sshanzel:
- Fixes change the email button to be small
- Fixes set password button to be small
- Refactor the password field component to allow the option whether to automatically indicate password strength
- Fixes change email button should be hidden if the user doesn't have a password yet

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-571 #done
WT-572 #done
WT-573 #done
WT-574 #done


rebelchris (changes requested):
I think we should refactor this.
It should never start validating without input or submit.

sshanzel:
> I think we should refactor this.
It should never start validating without input or submit.

Totally agree with that.

sshanzel:
@rebelchris I have added a few more fixes, specifically the missing tooltip and the styling of removing a provider. Let me know what you think 🚢

packages/webapp/components/layouts/AccountLayout/Security/index.tsx
```diff
@@ -176,8 +204,13 @@ function AccountSecurityDefault({
             inputId="new_password"
             label="Password"
             name="password"
+            indicateStrengthAutomatically={false}
```
rebelchris, line 207: Why do we choose to turn this off here?

packages/shared/src/components/fields/PasswordField.tsx
```diff
@@ -45,20 +52,32 @@ export function PasswordField({
...
       {...props}
       autoComplete="off"
       type={useType}
-      valueChanged={(value) =>
-        setPasswordStrengthLevel(passwordStrength(value).id)
-      }
+      valueChanged={onChange}
       leftIcon={<LockIcon />}
-      hint={showStrength ? hint : props.hint}
+      hint={showStrength && indicateStrength ? hint : props.hint}
```
rebelchris, line 77: I think this is becoming a little bit messy with mulitple states indicating if it should show or not?
Or maybe I'm misreading it.

But do we really need to have it so complex?
Why don't we simply always show the strength once inputted or validating?

rebelchris (changes requested):
Some small questions

packages/webapp/components/layouts/AccountLayout/Security/index.tsx
sshanzel: The design team asked that we should not show any indicator until the user interacted with it. Well, I think this should not be needed anymore after we have decided to start checking the strength once the field has value.

packages/shared/src/components/fields/PasswordField.tsx
sshanzel: I agree, this was the previous work before we had that condition. Let me refactor this one.

packages/webapp/components/layouts/AccountLayout/Security/index.tsx
rebelchris: Ah yes perfect, so we can remove this whole indicateauto stuff that would make it way more readable.

packages/shared/src/components/fields/PasswordField.tsx
```diff
@@ -45,17 +46,20 @@ export function PasswordField({
     ? `Password need a minimum length of ${props.minLength}`
     : passwordStrengthStates[passwordStrengthLevel].label;
   const hint = !hasUserAction ? null : userActionHint;
-  const shouldShowStrength = showStrength && hasUserAction;
+  const shouldShowStrength = !!value && showStrength && hasUserAction;
```
rebelchris, line 49: It should always validate on submit though, just making sure that happens (even when input is empty)

rebelchris (changes requested):

packages/shared/src/components/fields/PasswordField.tsx
sshanzel: I think when it is empty, the hint should be different, that it is required/missing.

packages/shared/src/components/fields/PasswordField.tsx
```diff
@@ -45,17 +46,20 @@ export function PasswordField({
...
     <TextField
       {...props}
       autoComplete="off"
       type={useType}
-      valueChanged={(value) =>
-        setPasswordStrengthLevel(passwordStrength(value).id)
-      }
+      valueChanged={onChange}
       leftIcon={<LockIcon />}
       hint={showStrength ? hint : props.hint}
```
rebelchris, line 64: But then should this not be the shouldShowStrength comparison?

rebelchris (approved):
Awesome, just one question for the hint

packages/shared/src/components/fields/PasswordField.tsx
sshanzel: I thought there were some ideas behind why this was left out to not use it. I agree it is better to consolidate the conditions using a single parameter. Let me refactor this.

PR #1250, opened 2022-09-21
fix: logout flow

rebelchris:
- Removed the timezone modal
- Removed the incomplete profile modal
- Reset the social registration if user is not confirmed
- Auto log out users who are not confirmed
- Don't show login profile card if user not confirmed

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-562 #done


packages/shared/src/components/MainLayout.tsx
```diff
@@ -200,7 +199,6 @@ export default function MainLayout({
         )}
         {children}
       </main>
-      <AccountCompletionModals />
```
sshanzel, line 203: Why did we have to remove this and the file completely? I think there are some things inside of it that are needed.

packages/shared/src/components/auth/AuthModal.tsx
```diff
@@ -31,7 +31,11 @@ export default function AuthModal({
 }: AuthModalProps): ReactElement {
   const [showOptionsOnly, setShowOptionsOnly] = useState(false);
   const { authVersion } = useContext(FeaturesContext);
-  const { closeLogin } = useContext(AuthContext);
+  const { closeLogin, logout } = useContext(AuthContext);
+  const closeAndLogout = (e) => {
```
idoshamun, line 35: I think `closeLogin` should have this functionality of closing the modal and logging out by default so there will be no way to close it without logging out

rebelchris: Ah it's not always that you are already logged in, but we can default to it if you think it makes more sense.

packages/shared/src/components/MainLayout.tsx
rebelchris: Don't think we need any of those fallbacks anymore in this new flow.
But love to hear if I missed something.

packages/shared/src/contexts/AuthContext.tsx
```diff
@@ -87,11 +88,16 @@ export const AuthContextProvider = ({
   refetchBoot,
   visit,
   isLegacyLogout,
+  firstLoad,
 }: AuthContextProviderProps): ReactElement => {
   const [loginState, setLoginState] = useState<LoginState | null>(null);
   const endUser = user && 'providers' in user ? user : null;
   const referral = user?.referrer;
 
+  if (firstLoad === true && endUser && !endUser?.infoConfirmed) {
```
idoshamun, line 97: Shouldn't this be inside a `useEffect`?

idoshamun (changes requested):
Please see my comments

packages/shared/src/contexts/AuthContext.tsx
rebelchris: Don't think it has to, it's all passed down which will cause re-evaluation each time.
All my test cases seem to work, what would be the positive of going useEffect for this?

packages/shared/src/components/auth/AuthModal.tsx
idoshamun: You can add a check inside the hook. But it's more fail-safe this way

packages/shared/src/contexts/AuthContext.tsx
idoshamun: That it will happen only once

packages/shared/src/components/auth/AuthModal.tsx
rebelchris: Ah no wait, but closeLogin is also called to simply close the modal, and that should not log out the person.
It's particular when they discard the changes only.

packages/shared/src/contexts/AuthContext.tsx
rebelchris: I think that would actually give us a lot of headaches and complicates as it's always delayed then.
Worse case this would call twice, but because of the state also can't happen right? (The firstLoad state)

idoshamun: Ok, let's do it :)

packages/shared/src/components/auth/AuthModal.tsx
idoshamun: Hmm I see

idoshamun (approved):

packages/shared/src/components/auth/AccountCompletionModals.tsx
```diff
@@ -1,105 +0,0 @@
...
-        }
-
-        const error =
-          data.ui.messages.find(
-            (message) => message.id === KRATOS_ERROR.INVALID_TOKEN,
-          ) || data.ui.messages[0];
-        displayErrorMessage(error.text);
-      },
-    },
-  );
```
sshanzel, line 68: This one is used to handle error messages from expired recovery flow when they get redirected to the app.

packages/shared/src/components/MainLayout.tsx
sshanzel: I see. One question, does this mean the user will always have complete information after the social registration?

packages/shared/src/components/auth/AccountCompletionModals.tsx
```diff
@@ -1,105 +0,0 @@
...
-    }
-
-    const res = await getKratosError(e.data.id);
-
-    if (!res?.error) {
-      return;
-    }
-
-    displayToast(res.error.message);
-  });
```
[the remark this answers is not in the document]
rebelchris: Ah right, ok that didn't make sense to be in the account completion then.
Missed that one, but yeah let's rather move it to a isolated scope.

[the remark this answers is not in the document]
sshanzel: Yep. A hook or something should work, I think. Sorry I didn't want to create a separate hook back then, lol.

packages/shared/src/contexts/BootProvider.tsx
```diff
@@ -120,6 +121,11 @@ export const BootDataProvider = ({
 
   useEffect(() => {
     if (bootRemoteData) {
+      if (initialLoad === null) {
+        setInitialLoad(true);
+      } else {
+        setInitialLoad(false);
+      }
```
sshanzel, line 128: Just a simplification.
```suggestion
      setInitialLoad(initialLoad === null);
```

sshanzel (approved):
Just one very small thing but all looks good to me! 🚢

packages/shared/src/components/MainLayout.tsx
rebelchris: Yes if they cancel the modal they will be logged out.
If they refresh the page, they will be logged out.

rebelchris:
@sshanzel Thanks for the concerns, added two new generic hooks please let me know what you think.
I've tested them and they seem to work nicely.

packages/shared/src/contexts/BootProvider.tsx
sshanzel: Yes, I think that should work as that is the condition itself.

sshanzel (approved):
Looks so nice and clean! Amazing work! ✨

PR #1263, opened 2022-09-24
fix: missing account verification modal

sshanzel:
- Refactored the `useAuthRecovery` to manage verification as well (and renamed the file).
- Managed the verification messages including showing the modal if it is finished.

Preview:
image

There are two ways for you to see this, whether you are already logged in or not. Once the former occurred, it displays the below modal (without the left side and changing the header). Though I feel like it is a bit empty even after I made the icon bigger.
Also the text we will have to wait for the other team's response (I will send one for it in auth channel).
image

## Events

- [ ] Does this not break anything in a companion?

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-597 #done


packages/shared/src/hooks/useAuthVerificationRecovery.ts
```diff
@@ -0,0 +1,72 @@
...
+      onSuccess: checkErrorMessage,
+    },
+  );
+
+  useQuery(
+    [{ type: 'verification', flow: router?.query.flow as string }],
+    ({ queryKey: [{ flow }] }) => getKratosFlow(AuthFlow.Verification, flow),
+    {
+      ...disabledRefetch,
+      enabled: !!router?.query.flow && !router?.query.recovery,
```
rebelchris, line 68: Is recovery correct here?
Both queries fire for the same thing?

packages/shared/src/components/auth/EmailVerified.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+function EmailVerified({
+  hasUser,
+  children,
+  onClose,
+}: EmailVerifiedProps): ReactElement {
+  return (
+    <>
+      <AuthModalHeader
+        title={hasUser ? 'Email address verified' : 'Log in to daily.dev'}
```
rebelchris, line 21: Not quite sure why this title is dynamic as the text always implies you are logged in?

rebelchris (changes requested):
Nice, only some small remarks.

packages/shared/src/components/auth/EmailVerified.tsx
sshanzel: Yes. There are two states where you can click the verification link. Whether you are logged in or not is valid.

packages/shared/src/hooks/useAuthVerificationRecovery.ts
sshanzel: The other one is recovery, this one is for verification.

rebelchris (approved):

PR #1265, opened 2022-09-25
refactor: consolidate code blocks for text area and text input

sshanzel:
## Changes
There was a lot of copy and pasted code from the `TextField` to the `Textarea` component which led us to decide we should refactor to make it more uniform and easier in maintaining.

- One of the clear changes is the `className` property being an object, as there are a lot of other elements in the component, we put classes on some of them some times. We use tailwindcss and since the property is not just `className` anymore (ex. `hintClassName`) the IntelliSense for available class names is not showing up. But when using an object, it does.
- Also refactored the `classed` library to be able to accept an object. (We can also copy and paste the whole lib and just change the parameter to be solely an object, if you think it will be much cleaner)
- Checked out fields and all seems to look normal.
- Tests have passed as well including the one dedicated to Textarea.
- Created another custom hook to encapsulate events and side effects.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-586 #done


rebelchris (approved):
Wow nice work.

This does seems like super high impact though due to the `classed` changes and text fields everywhere.
Please do some extensive tests on different devices/screens and browser.

sshanzel:
> This does seems like super high impact though due to the classed changes and text fields everywhere.
Please do some extensive tests on different devices/screens and browser.

After running through the pages and existing inputs, I think everything looks normal as how they were. Will be merging after fixing the conflicts.

PR #1272, opened 2022-09-28
feat: stepped modal

sshanzel:
## Changes
** Pointing to an intermediate branch.

All these are yet to include the functions that will happen after auth which will be worked on in a separate PR. This is so that if anyone wants to work on any part here, the UI is available to move forward to which area to work on.

- New component for managing stepped modal.
- Feed filters modal component that includes step 1.
- Refactored the normal switch to match our branding color.
- Refactored the radio button to match our branding color.
- Refactored the `IconsSwitch` to be reusable more than just for icons.

Preview:

https://user-images.githubusercontent.com/13744167/192800408-19a3e7e8-7e28-4477-8af7-db390029c634.mov

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-535 #done
WT-536 #done
WT-537 #done
WT-540 #done
WT-541 #done
WT-542 #done
WT-595 #done


packages/shared/src/components/Settings.tsx
```diff
@@ -116,11 +116,11 @@ export default function Settings({
     <div className={classNames('flex', 'flex-col p-6', className)} {...props}>
       <Section className="mt-0">
         <SectionTitle>Layout</SectionTitle>
-        <IconsSwitch
+        <CustomSwitch
```
rebelchris, line 119: Just thinking maybe "TextToggle" is a better name for it?
Custom doesn't really describe what it is.

packages/shared/src/components/fields/CustomSwitch.tsx
```diff
@@ -0,0 +1,104 @@
...
+        type="checkbox"
+        className="absolute w-0 h-0 opacity-0"
+        checked={checked}
+        onChange={onToggle}
+      />
+      <ContentContainer
+        ref={leftRef}
+        className={checked && 'text-theme-label-tertiary'}
+      >
+        {typeof LeftContent === 'string' ? (
```
rebelchris, line 77: I see quite a lot of these typeof checks why do we need them? Is there anything else we render in here?

packages/shared/src/components/modals/FeedFiltersModal.tsx
```diff
@@ -0,0 +1,106 @@
...
+              background:
+                'linear-gradient(45deg, rgba(23, 25, 31, 1) 0%, rgba(23, 25, 31, 0) 100%)',
+            }}
+          />
+          {getFilterCardPreviews(6, isInsaneMode)}
+        </div>
+      </FeedFilterStep>
+      <FeedFilterStep
+        title="Your eyes don’t lie"
+        description="Dark mode will less emit ׳blue light' from your screen -  which can keep you awake if you use your device before you go to bed"
```
rebelchris, line 98: There seems to be a weird collon before blue?

packages/shared/src/components/modals/FeedFiltersModal.tsx
```diff
@@ -0,0 +1,106 @@
...
+          )}
+        >
+          <div
+            className="absolute inset-0 rounded-8"
+            style={{
+              background:
+                'linear-gradient(45deg, rgba(23, 25, 31, 1) 0%, rgba(23, 25, 31, 0) 100%)',
+            }}
+          />
+          {getFilterCardPreviews(6, isInsaneMode)}
```
rebelchris, line 93: Why do we use insaneMode here?
Just thinking for future reference is "list" or "grid" not a better naming?

Looking as a outsider insane mode doesn't say anything to me.

packages/shared/src/components/modals/SteppedModal.tsx
```diff
@@ -0,0 +1,116 @@
...
+  children,
+  isLastStepLogin,
+  onStepChange,
+  onFinish,
+  onRequestClose,
+  ...props
+}: SteppedModalProps): ReactElement {
+  const [step, setStep] = useState(0);
+  const getWidth = () => {
+    const count = Children.count(children);
```
rebelchris, line 29: Might be a bit confusing using `children` and `Children` together, but not sure if there's an alternative?

packages/shared/src/components/modals/SteppedModal.tsx
```diff
@@ -0,0 +1,116 @@
...
+  const getBackwardsLabel = () => {
+    if (step === 0) {
+      return 'Skip';
+    }
+
+    return step === 1 ? 'Close' : 'Back';
+  };
+
+  const onBack = async (e: MouseEvent | KeyboardEvent) => {
+    if (step === 0 || step === 1) {
```
rebelchris, line 41: Remember steps can be interchangeable later on

packages/shared/src/lib/featureManagement.ts
```diff
@@ -128,10 +128,17 @@ export class Features {
     ['control', 'v1', 'v2', 'v3'],
   );
 
-  static readonly MyFeedVersion = new Features('my_feed_version', 'v1', [
+  static readonly FeedFilterVersion = new Features(
```
rebelchris, line 131: Be aware the old version should still be active and is a separate AB test to see if that one is shown or the new one.

packages/shared/src/components/modals/FeedFiltersModal.tsx
```diff
@@ -0,0 +1,106 @@
...
+          container: 'items-center',
+          content: classNames(
+            'relative flex flex-col items-center w-4/5 mt-8',
+            isInsaneMode && 'px-8',
+          ),
+        }}
+      >
+        <img
+          className="absolute -top-4 -left-2"
+          src="/recommended.png"
```
rebelchris, line 69: We probably want to switch this to cloudinary, same with the other image.

rebelchris (changes requested):
Awesome work, you've had some massive progress already on this! 👏

packages/shared/src/components/Settings.tsx
sshanzel: This can also work on any kind of component that accepts `className` and texts. Can be an Icon any random component. WDYT of `GenericSwitch`?

packages/shared/src/components/fields/CustomSwitch.tsx
sshanzel: This is to easily accommodate string kind of parameter. How about creating two different switch components to handle text only and the other side of this condition? Would love to hear what you think 🚢

packages/shared/src/components/modals/FeedFiltersModal.tsx
sshanzel: Oh, probably from the copy paste. Wow your eyes are really sharp! 🔭

packages/shared/src/components/modals/FeedFiltersModal.tsx
sshanzel: This is about the feed layout, which we call the insane mode the list kind of it. Should we replace our terminology?

packages/shared/src/components/modals/SteppedModal.tsx
sshanzel: I can probably use `React.Children` instead of importing the `Children` directly. Will refactor this one.

packages/shared/src/components/modals/SteppedModal.tsx
sshanzel: Yep, this will be agnostic of the context of the content, the exact ordering of the children is where it will be based.

packages/shared/src/components/modals/FeedFiltersModal.tsx
sshanzel: Yep, just applied it like this to move forward. Will note this down to not forget.

packages/shared/src/lib/featureManagement.ts
sshanzel: Yes, this change is relevant from the intermediate branch, the original should have a different variable. Just thought of renaming the new one to avoid confusion since the original one is named `MyFeedOnboardingVersion` and both are for onboarding.

packages/shared/src/components/Settings.tsx
rebelchris: Maybe just Switch or Toggle even? 👀

sshanzel: We already have the `Switch` - the most basic one we have 😭

packages/shared/src/components/modals/FeedFiltersModal.tsx
sshanzel: Alrighty. Ido was probably mad creating the list type and named it for insane people 🤣

packages/shared/src/components/Settings.tsx
rebelchris: Alright let's keep CustomSwitch then :)

sshanzel:
@rebelchris would love to hear your thoughts after making the updates 🚢

packages/shared/src/components/MainFeedLayout.tsx
```diff
@@ -361,6 +362,7 @@ export default function MainFeedLayout({
           directlyOpenedTab="Manage tags"
         />
       )}
+      <FeedFitlersModal isOpen /> {/* temporary for development */}
```
rebelchris, line 365: We just have to be very careful with this.
Possibly in future not even commit it, but use it locally, or perhaps adding a TODO block?

More a remark for discussion at gathering.

packages/shared/src/components/fields/CustomSwitch.tsx
```diff
@@ -0,0 +1,104 @@
...
+import classed from '../../lib/classed';
+
+export interface ContentsSwitchProps {
+  className?: string;
+  inputId: string;
+  name: string;
+  checked?: boolean;
+  onToggle?: () => unknown;
+  leftContent: React.FC<{ className?: string }> | string;
+  rightContent: React.FC<{ className?: string }> | string;
```
rebelchris, line 13: Did we not say to not use FC? I remember some discussion about this a while ago, but happy to leave this for now and discuss in gathering

packages/shared/src/components/fields/RadioItem.module.css
```diff
@@ -18,23 +18,23 @@
 
 .item.checked:hover .checkmark, .checked input:focus-visible ~ .checkmark {
   /* stylelint-disable-next-line unit-no-unknown */
```
rebelchris, line 20: This should no longer be unkown color or is it because of the 3D?

rebelchris (approved):
Nice, also don't forget we should technically put these on cloudinary (the images)

packages/shared/src/components/fields/CustomSwitch.tsx
sshanzel: This was just from the original file itself, but yes I think this can be a good discussion in the Gathering.

image

packages/shared/src/components/fields/RadioItem.module.css
sshanzel: Looks like it is because of the 3D as you mentioned.

https://app.circleci.com/pipelines/github/dailydotdev/apps/6094/workflows/69549dd1-13b0-4431-8000-a658efefd36c/jobs/9219

sshanzel:
@rebelchris nothing in this PR reflects the changes for the onboarding of feed filters, do you think we should instead merge this to master instead of referring to an intermediate branch?

sshanzel:
Merged to the intermediate branch, we can easily merge that to master if we decide to do so. Just let me know 🚢

PR #1280, opened 2022-10-03
fix: invalid if required and empty

sshanzel:
- The hint turns red once submitted but only when it has value
- This should fix the requirement for when the hint should also turn red if empty but must be a required field.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-619 #done

### DIFF
--- a/packages/shared/src/components/auth/LoginForm.tsx
+++ b/packages/shared/src/components/auth/LoginForm.tsx
@@ -61,6 +61,7 @@ function LoginForm({
         hint={hint as string}
         onChange={() => hint && setHint(null)}
         valid={!hint}
+        required
       />
       <PasswordField
         data-testid="login_password"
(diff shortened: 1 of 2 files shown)


packages/shared/src/components/auth/LoginForm.tsx
```diff
@@ -61,6 +61,7 @@ function LoginForm({
         hint={hint as string}
         onChange={() => hint && setHint(null)}
         valid={!hint}
+        required
```
rebelchris, line 64: Just be aware this also introduces the native browser validation, which is another task where we want to disable that.

rebelchris (approved):

packages/shared/src/components/auth/LoginForm.tsx
sshanzel: Oh, cool. Thanks for the note! 🚢

PR #1287, opened 2022-10-03
fix: open login when needed instead of signup

idoshamun:

Change the `onShowLogin` callback to open the login modal instead of the sign up

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-626 #done


packages/shared/src/components/fields/ImageInput.tsx
```diff
@@ -44,7 +44,7 @@ function ImageInput({
 }: ImageInputProps): ReactElement {
   const inputRef = useRef<HTMLInputElement>();
   const [error, setError] = useState('');
-  const [image, setImage] = useState(initialValue);
+  const [image, setImage] = useState(initialValue || fallbackImage);
```
rebelchris, line 47: I don't think it should always fallback, because of the OIDC flow. That will then always use the fallback?

Not sure if it's still the case, but that user to be why this was static null.

rebelchris (changes requested):
Not sure why the idea files and locks are diffed?

idoshamun:
Lockfiles were out of sync so I updated it. CircleCI complained so I did a fresh install.
As for the idea files, not sure. I think we should even remove them from the repo. What do you think?

packages/shared/src/components/fields/ImageInput.tsx
idoshamun: The edge case here, if `initialValue` is empty no error will be thrown and the image will be broken. It happened to me while trying out this feature. This is why I applied this fix. We don't always use fallback only when `initialValue` is empty

rebelchris: Hmm ok, just to make sure this doesn't break the OIDC or email flow as it used to do.

rebelchris:
Ok, makes sense, uhm please keep the idea files though, else we get into another round of issues.

packages/shared/src/components/fields/ImageInput.tsx
idoshamun: I'll double check

rebelchris (approved):
Nice 🚢

PR #1307, opened 2022-10-05
fix: apple touch icon should be square according to apple guidelines

minomy13:
## Changes

Made "Apple Touch Icon" square.
According to [Apple Design Human Interface Guidelines](https://developer.apple.com/design/human-interface-guidelines/foundations/app-icons#app-icon-attributes) you should not add corner radius to mobile icons.

Bildschirmfoto 2022-10-05 um 20 16 56

IMG_5658

- [X] iOS (Chrome and Safari)


minomy13:
Hey! Ist something wrong with this Pull Request? Should I change something? :)

rebelchris:
Hi @minomy13 Nothing wrong it sparked a internal discussion about changing the icon regardless, as we just opted to go for the square one.
I will keep you posted on the result.

minomy13:
Hi @rebelchris, thank you for quick reply. I'm not sure, maybe you got me wrong. The Apple Touch Icon only affects the Icon on iPhone and iPad if you add the App to hoemescreen. If this Icon has rounded Corners, like yours at the moment, IOS nevertheless will add its own rounded corners. Than you get these ugly black corners, as you can see on the screenshot, left side. So you can simply take an image of your icon without rounded corners at all and let IOS do its thing. (Screenshot -> right side) This won't affect your Icon on the web.

If you got me right and I explained it to you in such detail, I'm sorry. 

Best regards

rebelchris:
@minomy13 Thanks again for this amazing PR.
We recently switched to a completely new icon (purple one) and we took your advice to heart and made it rounded as expected for the apple guidelines.

Closing this PR, but again thank you so much for the info 🙏

PR #1310, opened 2022-10-07
refactor: logged in feed filters 

sshanzel:
- It can be noticed that the `v2` [parameter](https://github.com/dailydotdev/apps/compare/refactor-logged-filters?expand=1#diff-0c1c992f716223715bd0340d8841b8c575529e420b5262736043e58ecbc82548R24) for the items in the modal is strongly typed. This is true since that modal is only available for logged-in users and we are going to replace the current sidebar totally. But then that specific component is used somewhere else (create my feed) and we will continue to run an AB test.
- Also did a cleanup on the `shouldShowMyFeed` AB test that has been concluded to be turned on for everyone.

https://user-images.githubusercontent.com/13744167/194583396-8f0b32ee-d5f8-4b18-8d2c-5689d6f39c1e.mov

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-539 #done
WT-544 #done
WT-592 #done
WT-594 #done


packages/shared/src/components/sidebar/Sidebar.spec.tsx
```diff
@@ -129,40 +121,6 @@ it('should not render create my feed button if user has alerts.filter as false',
   expect(createMyFeed).not.toBeInTheDocument();
 });
 
-it('should remove filter alert if the user has filters and opened feed filters', async () => {
```
sshanzel, line 132: Note: removal of this test case was intentional since the sidebar won't be able to open the "Feed filters" menu with the current changes as it will now be part of the header for `my-feed` page. And in order for you to have a `my-feed` page, you must go through the feed filters onboarding process which has its own test suite that handles this case.

Relevant test suite: `TagsFilter.spec.tsx`

packages/webapp/__tests__/PopularPage.tsx
```diff
@@ -104,22 +103,6 @@ const renderComponent = (
   );
 };
 
-it('should request user feed', async () => {
```
sshanzel, line 107: Removal of this test case was intentional, as we will now never request for personalized feeds on the popular page with the `my feed` feature being turned on for everyone.

sshanzel:
@rebelchris @el would love to hear what you both think of the PR.

Note: it can be noticed that the branch name doesn't have the right prefix (ticket number) as this PR has fixed a lot of tickets at the same time. Probably a good discussion in Gathering on what should be followed.

packages/shared/src/components/MainFeedLayout.tsx
```diff
@@ -44,39 +44,40 @@ const FeedEmptyScreen = dynamic(
   () => import(/* webpackChunkName: "feedEmpty" */ './FeedEmptyScreen'),
 );
 
+const FeedFilters = dynamic(() => import('./filters/FeedFilters'));
```
rebelchris, line 47: We could add the chunk name if it's used anywhere else

packages/shared/src/components/MainFeedLayout.tsx
```diff
@@ -257,16 +249,30 @@ export default function MainFeedLayout({
...
+          {feedTitles[feedName]}
+          {hasFiltered && (
+            <SimpleTooltip placement="right" content="Feed filters">
+              <Button
+                iconOnly
+                className="mx-3 btn-tertiary"
+                icon={<FilterIcon />}
+                onClick={() => setIsFeedFiltersOpen(true)}
+              />
+            </SimpleTooltip>
```
rebelchris, line 273: Perhaps we can extract this into `FeedFilterButton` (even in this file) for readability

packages/shared/src/components/filters/TagCategoryDropdown.tsx
```diff
@@ -1,35 +1,52 @@
...
 export default function TagCategoryDropdown({
+  version,
   tagCategory,
   followedTags,
   blockedTags,
   onFollowTags,
   onUnfollowTags,
   onUnblockTags,
 }: TagCategoryDropdownProps): ReactElement {
+  const isV2 = version === 'v2';
```
rebelchris, line 35: What exactly are the versions for this modal? 
Or is this related to when it renders inside the feed filter onboarding?

rebelchris (changes requested):
Some minor comments.
Also thanks a lot for revamping the Sidebar a bit! 💖

packages/shared/src/components/MainFeedLayout.tsx
sshanzel: Totally agree with that. Also, I will have to migrate the alert for my feed when the user has activated the feature. I will probably include that part in this new component.

packages/shared/src/components/filters/TagCategoryDropdown.tsx
sshanzel: This is a component used by the feed filter onboarding version 1, but then we can easily reuse it for the new feed filter settings when logged-in which is why I applied it here. Now that you mentioned it, I feel like it is more appropriate if we will name the variable `isOnboarding` as this is not something related to AB testing. WDYT?

Ref: onboarding version 1
image

Ref: feed filters menu (always like this, no AB testing)
image

sshanzel:
> I added some comments, but since this is my first pr review, please take them as a grain of salt. ❤️

@el thank you for the effort! Though I think it did not proceed (probably still a draft), as I am not seeing it.

packages/shared/src/components/MainFeedLayout.tsx
sshanzel: Not used anywhere else as of the moment, but I agree, I think it is a good practice to set the chunk name to be future-proof.

packages/shared/src/components/filters/TagItemList.tsx
```diff
@@ -1,10 +1,11 @@
 import React, { ReactElement } from 'react';
 import { Tag } from '../../graphql/feedSettings';
 import { TagActionArguments } from '../../hooks/useTagAndSource';
-import { FiltersList } from './common';
+import { FiltersList, FiltersGrid } from './common';
 import TagItemRow from './TagItemRow';
 
 type TagItemListProps = {
+  version?: string;
```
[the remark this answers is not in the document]
sshanzel: Good point! Let's see, we were thinking of replacing this with a boolean value as the `version` seems to be a bit unfit, but if in case, I agree, we should refactor to using enum.

packages/shared/src/components/MainFeedLayout.tsx
```diff
@@ -44,39 +44,40 @@ const FeedEmptyScreen = dynamic(
...
-    },
-    upvoted: {
-      query: MOST_UPVOTED_FEED_QUERY,
-    },
-    discussed: {
-      query: MOST_DISCUSSED_FEED_QUERY,
-    },
-  };
+const propsByFeed: Record<MainFeedPage, FeedQueryProps> = {
+  'my-feed': {
```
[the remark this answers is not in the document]
sshanzel: It is strongly-typed using `Record<MainFeedPage, FeedQueryProps>`, so typos should be covered and reported by Typescript in any case. I think this can be a good discussion about whether we use string for the key or use the defined enum for its type. Let me log it down.

packages/shared/src/components/MainFeedLayout.tsx
```diff
@@ -44,39 +44,40 @@ const FeedEmptyScreen = dynamic(
...
 type FeedQueryProps = {
   query: string;
   queryIfLogged?: string;
   variables?: Record<string, unknown>;
 };
 
-interface FeedOptionalParams {
-  shouldShowMyFeed?: boolean;
+enum MainFeedPage {
+  MyFeed = 'my-feed',
```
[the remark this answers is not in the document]
sshanzel: Good point! Though for keys in general, and on enums, we use PascalCase as much as possible. I think should also list this down for discussion when we want to use snake_case.

packages/shared/src/components/filters/TagCategoryDropdown.tsx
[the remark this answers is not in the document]
sshanzel: I really like this one! Let me see how I can incorporate it after the changes we will apply to the props for this component.

rebelchris: On another note, we need to remove the icons in this version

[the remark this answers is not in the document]
rebelchris: This is a great suggestion I agree it will make it way more manageable down the line 👍
Thanks for the suggestion @el

sshanzel: I think Ido has delete everything out in the DB as I don't see it in prod anymore.

packages/shared/src/components/MainFeedLayout.tsx
sshanzel: Refactored this in the PR related to the pointer 🚢

sshanzel:
Thank you both for the awesome review! I have now made the necessary refactors, love the maintainability comments! 🚢 
Let me know what you think 🚀

packages/shared/src/components/filters/FeedFilters.tsx
```diff
@@ -1,70 +1,109 @@
...
-      updateAlerts({ filter: false });
+  const onNavClick = (active: string) => {
+    setDisplay(active);
+    if (isNavOpen) {
+      setIsNavOpen(false);
     }
-  }, [isOpen, alerts, user, hasAnyFilter]);
+  };
+
+  const getHeight = () => {
```
rebelchris, line 49: Just a question more general:
Why do we need to calculate the height dynamic. I think in our new modals everything is either 100% or a fixed height right?

packages/shared/src/components/filters/TagsFilter.tsx
```diff
@@ -23,10 +24,24 @@ import { FilterMenuProps } from './common';
...
+const TagsContainer = classed('div', 'grid grid-cols-1 gap-4 mx-6');
+
+interface TagsFilterProps extends FilterMenuProps {
+  tagCategoryLayout?: TagCategoryLayout;
+}
+
+const Container: Record<TagCategoryLayout, HTMLElementComponent> = {
+  settings: TagsContainer,
+  default: React.Fragment,
+};
```
rebelchris, line 39: Is there anyway we can uniform it with the one above?
Looks very similar

rebelchris (changes requested):
amazing work Lee,
Really well done on uniforming everything and the record types.

Some small comments, not really breaking, but have a look at them. 🙏

packages/shared/src/components/MainFeedLayout.tsx
sshanzel: Looks like I have pushed it in a different PR. Though that PR is pointing here, I think we can mark this out. 

https://github.com/dailydotdev/apps/pull/1318/files#diff-dc8e96a8b7f7dbbeb512be9e39b798856c9749a3506621c9f1927f79091bc903R47-R50

packages/shared/src/components/filters/FeedFilters.tsx
sshanzel: This is so the scrollbar is not hitting the header. With that mentioned, let me check how it would look like when the `position` is `sticky` or `fixed` since in the design, it should be always on top when scrolling.

packages/shared/src/components/filters/TagsFilter.tsx
sshanzel: They have their own way of using their containers. I think it is better to keep them separated from their own files. Also that they are only used in their respective components/files. WDYT?

rebelchris: Sure no problem with that 🙏

packages/shared/src/components/filters/FeedFilters.tsx
sshanzel: Sticky worked beautifully! Thank you for questioning this! 🚢

sshanzel:
Have now made the necessary changes and the awesome suggestions. Let me know if there are any further concerns we can make better 🚢 🚢 🚢 cc @rebelchris @el

rebelchris (approved):
Nice work Lee, 

Just two request before you merge it.
- Can you ensure the extension keeps working as intended
- Can you verify the mobile versions please (on device if possible)

sshanzel:
Preview on mobile.

https://user-images.githubusercontent.com/13744167/195795230-04294955-0898-431b-afe0-eb3d0f6ccdf8.mp4

PR #1313, opened 2022-10-10
fix: don't recall browser login for logged users

rebelchris:
- We currently always recall the login flow, but this should not be the case for logged users. (Only in privileged actions)

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-650 #done

### DIFF
--- a/packages/shared/src/components/auth/LoginForm.spec.tsx
+++ b/packages/shared/src/components/auth/LoginForm.spec.tsx
@@ -23,7 +23,10 @@ import { AuthContextProvider } from '../../contexts/AuthContext';
 import { formToJson } from '../../lib/form';
 import AuthOptions, { AuthOptionsProps } from './AuthOptions';
 
+let user = null;
+
 beforeEach(() => {
+  nock.cleanAll();
   jest.clearAllMocks();
 });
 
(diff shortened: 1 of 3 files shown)


packages/shared/src/components/auth/LoginForm.spec.tsx
```diff
@@ -23,7 +23,10 @@ import { AuthContextProvider } from '../../contexts/AuthContext';
 import { formToJson } from '../../lib/form';
 import AuthOptions, { AuthOptionsProps } from './AuthOptions';
 
+let user = null;
+
 beforeEach(() => {
+  nock.cleanAll();
```
sshanzel, line 29: No waaaayyy. I could have missed this as well in our test suites for `AccountSecurityPage` which could be the reason for the flaky tests/difficulty to do re-work 😢

sshanzel (approved):
LGTM! Nice fix! 🚢

packages/shared/src/components/auth/LoginForm.spec.tsx
rebelchris: Hmm could actually be!

PR #1318, opened 2022-10-11
refactor: alert placement flexibility

sshanzel:
- For easier usage of the alert for future developments, I have created four basic placements. Top, right, left, bottom.
- Now, the positioning is just a matter of props to pass.

Note: Kindly don't mind the actual component being cut off, the sample element I used has its container overflow hidden. 
Previews (reference element is the filter button):

Bottom
image

Left
image

Top
image

Right
image

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-538 #done


packages/shared/src/components/sidebar/MyFeedAlert.tsx
```diff
@@ -1,48 +1,106 @@
...
 
-const alertCopy = {
-  migrated: `Psst, your feed has a new name! We've already applied your content filters to it.`,
-  created: `🎉 Your feed is ready! Click here to manage your feed's settings`,
+interface ClassName {
+  container?: string;
+  label?: string;
+}
+
+export enum AlertPlacement {
```
rebelchris, line 13: We can probably extract this into a `alertComponent` that the MyFeedAlert uses?
Let me know what you think.

packages/shared/src/components/sidebar/MyFeedAlert.tsx
```diff
@@ -1,48 +1,106 @@
...
   hideAlert: () => unknown;
+  children: ReactNode;
+  placement?: AlertPlacement;
 }
+
+const getStyle = (
+  placement: AlertPlacement,
+  { height = 0 }: Partial<DOMRect> = {},
+): CSSProperties => {
+  if (placement === AlertPlacement.Left) {
```
rebelchris, line 47: I feel like we can optimize this a little bit better, maybe by using a same Record approach as above?

rebelchris (commented):
Nice, glad to see this being enhanced, some small suggestions to make it even more flexible.

packages/shared/src/components/sidebar/MyFeedAlert.tsx
sshanzel: Oh, that is totally right. Didn't realize the name while refactoring. Let me work on it quickly.

packages/shared/src/components/sidebar/MyFeedAlert.tsx
sshanzel: Make sense, let me find a way to do it a much secure way like you suggested.

packages/shared/src/components/sidebar/MyFeedAlert.tsx
sshanzel: Oh, that is totally right. Didn't realize the name while refactoring. Let me work on it quickly.

What do you think of PointedAlert for the new name?

rebelchris: Yeah something like `AlertPointer` will probably do.

sshanzel:
@rebelchris thank you for the awesome review. With your suggestions, I think it is so much cleaner now. 
Also, I refactored the name of the `AlertContainer` as it seems quite vague.

packages/shared/src/components/alert/PointedAlert.tsx
```diff
@@ -0,0 +1,94 @@
...
+      />
+      <PointedAlertContainer
+        className={classNames(
+          'absolute font-normal',
+          messageContainerClasses[placement],
+          className.container,
+        )}
+        style={getStyle(rect?.height)[placement]}
+      >
+        <PointedAlertCopy className={className.label}>
```
rebelchris, line 79: Oh sorry my main intention for the seperate file was to make it become unrelated to this specific use-case.
However maybe we can simply accept the below as children?

sshanzel: That is correct, we can now just add a parameter for the message, a `ReactNode` since this is now agnostic of the context behind the consuming end. Let me work on it quickly.

sshanzel:
I have now applied the necessary changes and fixed the companion alert message as well. I think that also shows that it works as intended after being applied to two different areas. Also, I think the component is now much more flexible. Happy reviewing! 🚢

packages/shared/src/components/sidebar/MyFeedAlert.tsx
sshanzel: Awesome. Will rename the file before merging.

packages/extension/src/companion/CompanionMenu.tsx
```diff
@@ -186,41 +183,12 @@ export default function CompanionMenu({
...
-                    ? 'block rotate-90'
-                    : 'hidden group-hover:block -rotate-90',
-                )}
-              />
-            </>
-          }
-          onClick={toggleCompanion}
-        />
-      </SimpleTooltip>
+      <CompanionToggle
```
rebelchris, line 186: Yes! This is amazing exactly why the refactor was worth it 🥳

packages/shared/src/components/alert/PointedAlert.tsx
```diff
@@ -0,0 +1,134 @@
...
+};
+
+const pointerClasses = {
+  [AlertPlacement.Top]: 'rotate-180',
+  [AlertPlacement.Right]: '-rotate-90',
+  [AlertPlacement.Left]: 'rotate-90',
+};
+
+const alertContainerClasses: Record<AlertPlacement, string> = {
+  top: '-translate-y-full top-0 flex-col-reverse',
```
rebelchris, line 42: Would we not want to use `[AlertPlacement.Top]` here to generalize it?

packages/shared/src/components/alert/PointedAlert.tsx
```diff
@@ -0,0 +1,134 @@
...
+  right: 'translate-x-full right-0 flex-row',
+  left: '-translate-x-full left-0 flex-row-reverse',
+  bottom: 'translate-y-full bottom-0 flex-col',
+};
+
+const getContainerStyle = (
+  [xOffset = 0, yOffset = 0]: OffsetXY = [0, 0],
+): Record<AlertPlacement, CSSProperties> => ({
+  left: { left: xOffset },
```
rebelchris, line 51: Would we not want to use `[AlertPlacement.Top]` here to generalize it?

packages/shared/src/components/alert/PointedAlert.tsx
```diff
@@ -0,0 +1,134 @@
...
+  color?: PointerColor;
+  isAlertDisabled?: boolean;
+  onClose?: () => unknown;
+}
+
+const DEFAULT_POINTER_LENGTH = 6;
+const getMessageStyle = (
+  length = DEFAULT_POINTER_LENGTH,
+): Record<AlertPlacement, CSSProperties> => ({
+  left: { marginRight: length },
```
rebelchris, line 72: Would we not want to use `[AlertPlacement.Top]` here to generalize it?

packages/shared/src/components/alert/PointedAlert.tsx
```diff
@@ -0,0 +1,134 @@
...
+  offset,
+  isAlertDisabled,
+  onClose,
+}: PointedAlertProps): ReactElement {
+  const pointerRef = useRef<HTMLDivElement>();
+  const rect = pointerRef?.current?.getBoundingClientRect?.();
+
+  if (isAlertDisabled) {
+    return <>{children}</>;
+  }
```
rebelchris, line 93: In which scenario do we use this version?

rebelchris (approved):
Nice love the refactor.
No breaking changes, however some things you can consider.

packages/shared/src/components/alert/PointedAlert.tsx
sshanzel: Just to make it less verbose, on the other variable, it did not use the `Record` type since not all placements required a value (bottom placement). Though we can probably convert the variable I am mentioning, it will only have an empty string literal for its value `''`, which is probably much safer since it will require the introduction of new items in the enum to force a value and not be forgotten.

packages/shared/src/components/alert/PointedAlert.tsx
sshanzel: When we now hide the alert and just show the children we wrapped.

packages/shared/src/components/alert/PointedAlert.tsx
rebelchris: Yeah that sounds good, it will make it more uniform that way.
Resolving all the ones logged below in that case

PR #1327, opened 2022-10-17
fix: render issue with sign back in

sshanzel:
- Fixes the infamous issue of provider button showing differently.
- Used `useMemo` instead of the state as we don't actually need the second parameter from `useState`.
- There seems to be an issue with `React.Fragment` rendering the component. If you want to know more about it, just check with Chris or me as it seems to be an actual React issue and a bit complicated as we found.
- Will also probably raise an issue on React/NextJS after digging further.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-667 #done


rebelchris (approved):
Let's do it, but indeed investigate if it's a react issue seperatly.

PR #1333, opened 2022-10-24
refactor: sidebar broken down to multiple components

sshanzel:
## Changes
As changes and additions are applied on the sidebar, the code becomes very bloated. It also proves difficult to read and to easily add new items in the sidebar itself. I've decided to break down every part possible to be in separate components. Also did some refactoring on existing components

- Refactored some prop names.
- Refactored component/file name (the settings modal was named `FeedSettings` when it is for general settings).
- Extracted some components in the common file to have their file for better maintainability.
- Included some feature flags in the `FeaturesContext`.
- Extracted the "RenderSection" to have its file for better maintainability.
- Broken down the whole `Sidebar` component into multiple files based on sections.
- While testing out the dnd modal, I also noticed it doesn't behave as intended when clicking outside the modal, which should close itself.

**Note: Have thoroughly tested the items in the sidebar and all work as intended. Including the preview deployment and extension with production build**

## Events

- [x] MobileL (420px)
- [x] Tablet (656px)
- [x] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-691 #done


packages/shared/src/components/UserSettingsButton.tsx
```diff
@@ -4,14 +4,14 @@ import LayoutIcon from './icons/Layout';
 import { HeaderButton } from './buttons/common';
 import { SimpleTooltip } from './tooltips/SimpleTooltip';
 
-const FeedSettingsModal = dynamic(
+const UserSettingsModal = dynamic(
   () =>
     import(
-      /* webpackChunkName: "feedSettingsModal" */ './modals/FeedSettingsModal'
+      /* webpackChunkName: "UserSettingsModal" */ './modals/UserSettingsModal'
```
rebelchris, line 10: It's not documented, but we should keep camelCase webpack chunk names, as suggested best practice.
I noticed we use it intertwined with PascalCase, but let's set for camelCase.

packages/shared/src/components/sidebar/ContributeSection.tsx
```diff
@@ -0,0 +1,86 @@
...
+import FeaturesContext from '../../contexts/FeaturesContext';
+import EmbedIcon from '../icons/Embed';
+import LinkIcon from '../icons/Link';
+import { ListIcon, SidebarMenuItem } from './common';
+import { Section, SectionCommonProps } from './Section';
+
+const SubmitArticleModal = dynamic(
+  () =>
+    import(
+      /* webpackChunkName: "SubmitArticleModal" */ '../modals/SubmitArticleModal'
```
rebelchris, line 13: Suggested camelCase:
```suggestion
      /* webpackChunkName: "submitArticleModal" */ '../modals/SubmitArticleModal'
```

packages/shared/src/components/sidebar/ContributeSection.tsx
```diff
@@ -0,0 +1,86 @@
...
+    ),
+);
+
+const NewSourceModal = dynamic(
+  () =>
+    import(/* webpackChunkName: "NewSourceModal" */ '../modals/NewSourceModal'),
```
rebelchris, line 19:
```suggestion
    import(/* webpackChunkName: "newSourceModal" */ '../modals/NewSourceModal'),
```

packages/shared/src/components/sidebar/ContributeSection.tsx
```diff
@@ -0,0 +1,86 @@
...
+  const {
+    canSubmitArticle,
+    submitArticleOn,
+    submitArticleSidebarButton,
+    submitArticleModalButton,
+  } = useContext(FeaturesContext);
+  const { trackEvent } = useContext(AnalyticsContext);
+  const [showSubmitArticle, setShowSubmitArticle] = useState(false);
+  const [showNewSourceModal, setShowNewSourceModal] = useState(false);
+  const contributeMenuItems: SidebarMenuItem[] = [];
```
rebelchris, line 32: Let's actually set the suggest new source always as the default, we can then insert the submit article when it's existing at the right position.
No need to assign it later on.

packages/shared/src/components/sidebar/ContributeSection.tsx
```diff
@@ -0,0 +1,86 @@
...
+    };
+    contributeMenuItems.push(submitArticleMenuItem);
+  }
+
+  contributeMenuItems.push({
+    icon: () => <ListIcon Icon={() => <EmbedIcon />} />,
+    title: 'Suggest new source',
+    action: () => setShowNewSourceModal(true),
+    active: showNewSourceModal,
+  });
```
rebelchris, line 60: This can immediately be set.

packages/shared/src/components/sidebar/Sidebar.tsx
```diff
@@ -2,140 +2,34 @@ import React, { ReactElement, useContext, useState, useMemo } from 'react';
...
+import { UnlabeledBottomSection } from './UnlabeledBottomSection';
+import { DiscoverSection } from './DiscoverSection';
+import { ContributeSection } from './ContributeSection';
+import { ManageSection } from './ManageSection';
+import { MobileMenuIcon } from './MobileMenuIcon';
+
+const UserSettingsModal = dynamic(
+  () =>
+    import(
+      /* webpackChunkName: "UserSettingsModal" */ '../modals/UserSettingsModal'
```
rebelchris, line 26:
```suggestion
      /* webpackChunkName: "userSettingsModal" */ '../modals/UserSettingsModal'
```

packages/shared/src/components/sidebar/UnlabeledBottomSection.tsx
```diff
@@ -0,0 +1,62 @@
...
+import { ListIcon, Nav, SidebarMenuItem } from './common';
+import InvitePeople from './InvitePeople';
+import { Section, SectionCommonProps } from './Section';
+
+interface UnlabeledBottomSectionProps extends SectionCommonProps {
+  optOutWeeklyGoal: boolean;
+  showSettings: boolean;
+}
+
+export function UnlabeledBottomSection({
```
rebelchris, line 15: Not a big fan of the name, perhaps SidebarBottomSection would even be better?

packages/shared/src/components/sidebar/DiscoverSection.tsx
```diff
@@ -0,0 +1,68 @@
...
+  onNavTabClick?: (page: string) => unknown;
+}
+
+export function DiscoverSection({
+  isItemsButton,
+  enableSearch,
+  onNavTabClick,
+  ...defaultRenderSectionProps
+}: DiscoverSectionProps): ReactElement {
+  const { popularFeedCopy } = useContext(FeaturesContext);
```
rebelchris, line 22: One thing to consider is actually passing this down from the sidebar, we already initialise the FeaturesContext there so passing down would be neater?

packages/shared/src/components/sidebar/ContributeSection.tsx
```diff
@@ -0,0 +1,86 @@
...
+    import(/* webpackChunkName: "NewSourceModal" */ '../modals/NewSourceModal'),
+);
+
+export function ContributeSection(props: SectionCommonProps): ReactElement {
+  const {
+    canSubmitArticle,
+    submitArticleOn,
+    submitArticleSidebarButton,
+    submitArticleModalButton,
+  } = useContext(FeaturesContext);
```
rebelchris, line 28: I think we can pass these down from the Sidebar as the context is already loaded there, what do you think?

packages/shared/src/components/sidebar/ContributeSection.tsx
```diff
@@ -0,0 +1,86 @@
...
+  const { trackEvent } = useContext(AnalyticsContext);
```
rebelchris, line 29: Same with this AnalyticsContext, I see some components use it, but they all part of the sidebar anyway.
So maybe we can simply pass it down?

rebelchris (changes requested):
Nice work Lee,

Way more manageable like this.
Got some small remarks but open to discussion on them ofcourse.

packages/shared/src/components/UserSettingsButton.tsx
sshanzel: Ah, I see. This was copy-pasted, thought it was alright to have it in PascalCase. I shall use camelCase in the future then.

packages/shared/src/components/sidebar/ContributeSection.tsx
sshanzel: Was thinking this is one too many variables to pass, and just to keep the props slim, also as much as to take away from the `Sidebar`, but we can definitely do that.

sshanzel: Was thinking this is one too many variables to pass, and just to keep the props slim, also as much as to take away from the Sidebar, but we can definitely do that. WDYT?

rebelchris: I think anything outside of analyticsContext makes sense it will make the components more dynamic as well.
But eventually leaving it up to you to decide 🙏

sshanzel: Thanks for the thoughts. Let me apply some refactoring on this 🚢

sshanzel:
I Applied the refactor for features flags coming from the props as it is much cleaner, as you suggested, I was just probably madly refactoring the sidebar and did tear it down left and right 😆 

@rebelchris would love to hear what you think!

packages/shared/src/components/sidebar/MobileMenuIcon.tsx
```diff
@@ -0,0 +1,41 @@
...
+import AnalyticsContext from '../../contexts/AnalyticsContext';
+import SettingsContext from '../../contexts/SettingsContext';
+import { Button } from '../buttons/Button';
+import ArrowIcon from '../icons/Arrow';
+import { SimpleTooltip } from '../tooltips/SimpleTooltip';
+
+export function MobileMenuIcon(): ReactElement {
+  const { trackEvent } = useContext(AnalyticsContext);
+  const { sidebarExpanded, toggleSidebarExpanded } =
+    useContext(SettingsContext);
```
rebelchris, line 11: We can actually also pass these down form the sidebar as they already loaded there.

packages/shared/src/components/sidebar/Sidebar.tsx
```diff
@@ -338,84 +91,58 @@ export default function Sidebar({
...
+            <ContributeSection
               {...defaultRenderSectionProps}
-              title="Manage"
-              items={manageMenuItems}
-              useNavButtonsNotLinks={false}
+              canSubmitArticle={canSubmitArticle}
+              submitArticleOn={submitArticleOn}
+              submitArticleSidebarButton={submitArticleSidebarButton}
+              submitArticleModalButton={submitArticleModalButton}
+              onTrackEvent={trackEvent}
```
rebelchris, line 121: Sorry for messing you around, but this actually makes more sense closer to implementation as it's not re-used.

I think we can safely say tracking can happen inside component, all other should rather be drill-down?

rebelchris (approved):
Amazing work Lee! 🥳

Got some minor comments, but perhaps wait till you get a second set of eyes on this before merging.

packages/shared/src/components/sidebar/Sidebar.tsx
sshanzel: No worries, I shall make that happen! 🚢

rebelchris (approved):
Wohoo amazing 🥳

PR #1334, opened 2022-10-25
feat: new onboarding experiment

sshanzel:
## Changes

Missing: Tracking events. Planning to take it up in a separate PR once functionality wise are in place.

- Introduced a new onboarding version experiment.
- Created a new component for the kind of stepped modal (the one with the progress bar).
- The onboarding modal was created separately from the MyFeedFiltersModal for easier management. This just then decides which one to display based on the current onboarding version.
- Based on the Slack discussion, logged-in users shall also see the onboarding process if they haven't seen it yet. @rebelchris I would really appreciate it if you can give an intense review of this part in the `MainFeedLayout` as you were familiar with how it worked on the anonymous users. 
- Introduced multiple new components for the new onboarding experience.
- Made the `Feature` class type-strict so the build won't fail whenever we use the type of an `enum` for the data of `FeaturesContextProvider`

Tested out the variation for the new onboarding layout, and switched v1 and v2 for the onboarding version itself.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-670 #done


sshanzel:
Overriding feature flags should now be possible on preview deployment. I shall create a confluence page for it.
Ex:
Flag name: `onboarding_version`
To set it open the console in your Chrome browser. Type as below then hit enter:
`document.cookie = "onboardingVersion=v2";`
From snake_case in the flagsmith, just make it a camelCase in the cookie values (as how we normally name our variables). Or else I will have to make it accept snake_case and just turn it as camelCase in our code so it will be easier for them. Would love to hear what you think! 🚢

packages/shared/src/components/MainFeedLayout.tsx
```diff
@@ -37,6 +37,11 @@ import AnalyticsContext from '../contexts/AnalyticsContext';
 import useSidebarRendered from '../hooks/useSidebarRendered';
 import FeedFilterMenuButton from './filters/FeedFilterMenuButton';
 import SortIcon from './icons/Sort';
+import {
+  MyFeedOnboardingVersion,
+  OnboardingVersion,
+} from '../lib/featureValues';
+import OnboardingModal from './modals/OnboardingModal';
```
rebelchris, line 44: Should we not opt to load this dynamic?

packages/shared/src/components/MainFeedLayout.tsx
```diff
@@ -379,6 +365,45 @@ export default function MainFeedLayout({
     }
   }, [sortingEnabled, selectedAlgo, loadedSettings, loadedAlgo]);
 
+  useEffect(() => {
```
rebelchris, line 368: The following two useEffects are quite big and specific, could we perhaps move them to a custom hook instead?

It will make it more understandable what's happening.

packages/shared/src/components/MainFeedLayout.tsx
```diff
@@ -129,13 +134,16 @@ const getQueryBasedOnLogin = (
   return null;
 };
 
+type ModalCommand = Record<OnboardingVersion, (value: boolean) => unknown>;
```
rebelchris, line 137: This technically has nothing to do with the MainFeedLayout, can we move it away to another file?

packages/shared/src/components/MainFeedLayout.tsx
```diff
@@ -175,50 +185,28 @@ export default function MainFeedLayout({
...
-      feedName !== defaultFeed &&
-      !getShouldRedirect(isMyFeed, !!user)
-    ) {
-      updateDefaultFeed(feedName);
-    }
-  }, [defaultFeed, feedName]);
+  const modalCommand: ModalCommand = {
+    v1: setCreateMyFeed,
+    v2: setIsOnboardingOpen,
+  };
```
rebelchris, line 199: I found it hard to understand at first what the ModalCommand exactly did.
Is there anyway we can make it a bit more clear?

packages/shared/src/components/containers/FeedTopicCard.tsx
```diff
@@ -1,53 +1,93 @@
...
+  text?: string;
+}
+
 const BackgroundLayer = classed(
   'div',
   'group-hover:z-0 absolute inset-0 w-full h-full rounded-14 bg-theme-status-cabbage transition-transform',
 );
 
+const background: Record<OnboardingFiltersLayout, ReactNode> = {
+  grid: (
```
rebelchris, line 28: Did we not choose to always go for the Enums as well?
It will make it way easier to find and replace if they ever need changing?

But maybe I'm wrong.

packages/shared/src/components/modals/SteppedModal.tsx
```diff
@@ -3,37 +3,77 @@ import React, {
...
+  trigger: string;
+  skippable?: boolean;
   isLastStepLogin?: boolean;
+  isFirstStepIntroduction?: boolean;
   onStepChange?: (stepBefore: number, stepNow: number) => void | Promise<void>;
   onFinish?: () => void | Promise<void>;
+  onAuthSuccess?: (isLogin?: boolean) => void | Promise<void>;
 }
 
 function SteppedModal({
```
rebelchris, line 29: This is actually a really cool generic component, but we now tie it down to the auth, is it not better to push auth as the last step from outside?

Just thinking this would make a super easy adjustment and make this way more flexible?
But if it's a lot of work or hassle don't worry about it.

packages/shared/src/components/onboarding/FilterOnboarding.tsx
```diff
@@ -0,0 +1,54 @@
...
+  tagsCategories: TagCategory[];
+  selectedId: Record<string, boolean>;
+  topicLayout?: OnboardingFiltersLayout;
+  onSelectedChange: (id: string) => void;
+}
+
+const classes: Record<OnboardingFiltersLayout, string> = {
+  grid: 'grid-cols-3 gap-6',
+  list: 'grid-cols-1 gap-4',
+};
```
rebelchris, line 19: This can actually go in the common onboarding file, just to clean it up a bit.

packages/shared/src/components/onboarding/FilterOnboarding.tsx
```diff
@@ -0,0 +1,54 @@
...
+  tagsCategories,
+  onSelectedChange,
+}: FilterOnboardingProps): ReactElement {
+  return (
+    <OnboardingStep
+      topIcon={
+        <img
+          className="mx-auto mb-6 w-16"
+          src={cloudinary.feedFilters.supercharge}
+          alt="test"
```
rebelchris, line 33: Please change this before merging, else we'll forget about it.

packages/shared/src/lib/featureManagement.ts
```diff
@@ -122,8 +128,30 @@ export class Features {
 
   static readonly MyFeedOnboardingVersion = new Features(
     'my_feed_onboarding_version',
-    'control',
-    ['control', 'v1', 'v2', 'v3'],
+    MyFeedOnboardingVersionEnum.Control,
```
rebelchris, line 131: We recently decided to move away from control as a name.
Can you please move them all one up, but perhaps let product know about it so they know which version is which naming.

rebelchris: The idea behind this change is that `control` is a generic product term. Today it might reference v1, but in 3 months from now they might decide v2 becomes the control version. (Hope this makes sense)

packages/shared/src/lib/featureManagement.ts
```diff
@@ -122,8 +128,30 @@ export class Features {
...
+      MyFeedOnboardingVersionEnum.Control,
+      MyFeedOnboardingVersionEnum.V1,
+      MyFeedOnboardingVersionEnum.V2,
+      MyFeedOnboardingVersionEnum.V3,
+    ],
+  );
+
+  static readonly OnboardingSteps = new Features(
+    'onboarding_steps',
+    'topics/layout/theme',
```
rebelchris, line 142: Rather use the enums you defined for this 🙏

packages/shared/src/lib/featureManagement.ts
```diff
@@ -122,8 +128,30 @@ export class Features {
...
+  );
+
+  static readonly UserOnboardingVersion = new Features(
+    'onboarding_version',
+    OnboardingVersion.V1,
+    [OnboardingVersion.V1, OnboardingVersion.V2],
+  );
```
rebelchris, line 149: What makes this difference from the myfeed onboarding version?

packages/shared/src/lib/featureValues.ts
```diff
@@ -9,3 +9,20 @@ export enum AuthVersion {
...
+  V2 = 'v2',
+}
+
+export enum OnboardingFiltersLayout {
+  Grid = 'grid',
+  List = 'list',
+}
+
+export enum MyFeedOnboardingVersion {
+  Control = 'control',
```
rebelchris, line 24: Same here, we can move them up one.

rebelchris (changes requested):
Initial code review, thanks a lot for refactoring to such small components and enhancing the types for the features. 🥳

Going to do some device tests now.

packages/shared/src/components/containers/FeedTopicCard.tsx
sshanzel: I remember we decided to strictly use `enum` as `key` whenever we don't use `Record` as with this type, changes and missing keys will surface as well.

packages/shared/src/components/MainFeedLayout.tsx
sshanzel: True. Let's load this dynamically.

packages/shared/src/components/MainFeedLayout.tsx
sshanzel: Good point. Let me see how we can move out these code blocks that also involve the modal.

packages/shared/src/components/MainFeedLayout.tsx
sshanzel: Will refactor this to the new hook/component.

packages/shared/src/components/modals/SteppedModal.tsx
sshanzel: Yep, good point. My only worry was the generic modal would have no idea to overrule the footer of next and previous. I am actually torn if those things should be handled outside as well and just pass a state for the max and current step. WDYT?

Though the auth is optional, you will pass a certain parameter to when it should be the last step or not at all.

packages/shared/src/components/onboarding/FilterOnboarding.tsx
sshanzel: They are not used anywhere else outside this component, which is why I thought of keeping them here. Should we still store them in a common file?

packages/shared/src/components/onboarding/FilterOnboarding.tsx
sshanzel: Oh, no. Thank you for pointing that out. Didn't realize this mistake. Let me refactor it now.

packages/shared/src/components/modals/SteppedModal.tsx
rebelchris: If it's too much work just leave it, we'll deal with it once we need it :)

packages/shared/src/lib/featureManagement.ts
sshanzel: This is actually the combination of enums, as this will be coming from the flagsmith. But when we expose it to our code, this is converted to `enum[]`

packages/shared/src/lib/featureManagement.ts
sshanzel: My feed onboarding version is the old one, not sure if it is still going on. I believe some of its values will determine whether to show the skip button or not.

packages/shared/src/lib/featureManagement.ts
rebelchris: This is the default right? That's why I was thinking perhaps we can typecast this one a bit safer?

sshanzel: Yes, this is the default, but I'm quite not yet sure of what would be the values for the enum.

Currently, we do the typecasting before we pass it to our context provider as below.
image

sshanzel:
@rebelchris this now validates the number of minimum tags to select based on the feature flag.

Also refactored the way we allow to set feature flags. Just execute something like below:

`window.setFeature('onboardingVersion', 'v2')`

packages/shared/src/lib/featureManagement.ts
sshanzel: I have now tried to concatenate it this way. WDYT?

image

el (approved):

sshanzel:
@rebelchris would love to hear what you think about the refactoring made 🚢 🚢 🚢

packages/shared/src/contexts/FeaturesContext.tsx
```diff
@@ -36,11 +47,19 @@ export const FeaturesContextProvider = ({
...
+  const steps = getFeatureValue(Features.OnboardingSteps, flags)?.split?.('/');
+  const onboardingSteps = (steps as OnboardingStep[]) || [];
   const features = useMemo(
     () => ({
       flags,
-      feedFilterVersion: getFeatureValue(Features.FeedFilterVersion, flags),
-      feedFilterCardVersion: getFeatureValue(Features.FeedFilterVersion, flags),
+      onboardingSteps,
+      onboardingMinimumTopics:
+        getFeatureValue(Features.OnboardingMinimumTopics, flags) ?? 0,
```
rebelchris, line 57: Is zero going to be the default, remember from back in the discussions 2 was the magic number?

rebelchris (approved):
Nice work Lee, seeing this is quite a high impact component you might want to consider asking Ido for a review as well.

packages/shared/src/contexts/FeaturesContext.tsx
sshanzel: From the meeting we had about the experiments, it was mentioned to instead have 0. With that, I will send out a message in the channel for final confirmation. Thank you! 🚢

sshanzel:
> Nice work Lee, seeing this is quite a high impact component you might want to consider asking Ido for a review as well.

cc @idoshamun for another pair of eyes as suggested 🚢

packages/shared/src/contexts/FeaturesContext.tsx
rebelchris: Alright if you guys discussed it, it's fine 🙏

idoshamun: Yes. Zero is control at the moment :)

packages/shared/src/components/MainFeedLayout.tsx
```diff
@@ -371,6 +344,17 @@ export default function MainFeedLayout({
...
+  useEffect(() => {
+    if (
+      defaultFeed !== null &&
+      feedName !== null &&
+      feedName !== defaultFeed &&
+      !getShouldRedirect(isMyFeed, !!user)
+    ) {
+      updateDefaultFeed(feedName);
+    }
+  }, [defaultFeed, feedName]);
```
idoshamun, line 356: What's this?

packages/shared/src/components/MainFeedLayout.tsx
```diff
@@ -384,12 +368,18 @@ export default function MainFeedLayout({
           onRequestClose={() => setIsFeedFiltersOpen(false)}
         />
       )}
-      {createMyFeed && (
+      {isCreateMyFeedOpen && (
```
idoshamun, line 371: I find the names `CreateMyFeed` and `Onboarding` quite confusing. May I suggest `LegacyOnboarding` and `Onboarding`?

packages/shared/src/components/MainFeedLayout.tsx
sshanzel: I think this was to set the default feed if the user has my-feed to make it automatic. Though I just moved it downwards to make all use effects be at before rendering.

packages/shared/src/components/containers/FeedTopicCard.tsx
```diff
@@ -1,53 +1,93 @@
...
   isActive,
+  topicLayout = OnboardingFiltersLayout.Grid,
   onClick,
 }: FeedTopicCardProps): ReactElement {
-  const { feedFilterCardVersion } = useContext(FeaturesContext);
-  const isV1 = feedFilterCardVersion === 'v1';
-  const isV2 = feedFilterCardVersion === 'v2';
+  const backgroundLayers = background[topicLayout];
+
+  const getIcon = () => {
```
idoshamun, line 58: I think this can go outside the component

packages/shared/src/components/modals/OnboardingModal.tsx
```diff
@@ -0,0 +1,150 @@
...
+  const { registerLocalFilters } = useMyFeed();
+  const [selectedTopics, setSelectedTopics] = useState({});
+  const [invalidMessage, setInvalidMessage] = useState<string>(null);
+  const { onboardingSteps, onboardingFiltersLayout, onboardingMinimumTopics } =
+    useContext(FeaturesContext);
+  const { insaneMode, themeMode, setTheme, toggleInsaneMode } =
+    useContext(SettingsContext);
+  const { tagsCategories } = useFeedSettings();
+  const { onFollowTags, onUnfollowTags } = useTagAndSource({
+    origin: 'filter onboarding',
```
idoshamun, line 45: Please keep it consistent with whatever we have in the existing implementation

idoshamun, line 45: I think it will be much cleaner if every step imports the context for itself. The only thing the onboarding modal is propagate it down which makes it messy

packages/shared/src/components/modals/OnboardingModal.tsx
```diff
@@ -0,0 +1,150 @@
...
+  const [step, setStep] = useState(0);
+  const onStepChange = (beforeStep: number, stepNow: number) => {
+    setStep(stepNow);
+    const filterStep = onboardingSteps.indexOf(OnboardingStep.Topics);
+
+    if (beforeStep - 1 === filterStep) {
+      const key = getFeedSettingsQueryKey();
+      const { feedSettings } = client.getQueryData<AllTagCategoriesData>(key);
+      updateLocalFeedSettings(feedSettings);
+    }
```
idoshamun, line 56: What's this?

packages/shared/src/components/modals/OnboardingModal.tsx
```diff
@@ -0,0 +1,150 @@
...
+  const nextButtonValidations = (() => {
+    const validations = Array(STEPS_COUNT).fill(null);
+    const filterStep = onboardingSteps.indexOf(OnboardingStep.Topics);
+
+    if (filterStep !== -1) {
+      validations[filterStep + 1] = onValidateFilter;
+    }
+
+    return validations;
+  })();
```
idoshamun, line 114: What's this?

packages/shared/src/components/modals/OnboardingModal.tsx
```diff
@@ -0,0 +1,150 @@
...
+
+  const onFinishOnboarding = async (isLogin: boolean) => {
+    const key = getFeedSettingsQueryKey();
+    const { feedSettings } = client.getQueryData<AllTagCategoriesData>(key);
+    const { hasFilters } = await registerLocalFilters(feedSettings);
+    if (!hasFilters) {
+      return;
+    }
+
+    await refetchBoot();
```
idoshamun, line 124: Why is it required?

packages/shared/src/components/modals/SteppedModal.tsx
```diff
@@ -3,37 +3,86 @@ import React, {
...
+
+  const onCancelAuth = async () => {
+    await logout();
+    await onStepChange?.(step, step - 1);
+    setStep(step - 1);
+    trackEvent({
+      event_name: AuthEventNames.CloseSignUp,
+      extra: JSON.stringify({ trigger, screenValue }),
+    });
+  };
```
idoshamun, line 74: Doesn't the auth modal take care of it already?

packages/shared/src/components/modals/SteppedModal.tsx
```diff
@@ -3,37 +3,86 @@ import React, {
...
   isLastStepLogin,
+  isFirstStepIntroduction,
+  onValidateNext = [],
+  onInvalid,
   onStepChange,
   onFinish,
   onRequestClose,
+  onAuthSuccess,
   ...props
 }: SteppedModalProps): ReactElement {
```
idoshamun, line 49: `SteppedModal` sounds like a generic component but looking inside, it seems like it's not.
Should we change the name of the component?

packages/shared/src/components/onboarding/LayoutOnboarding.tsx
```diff
@@ -0,0 +1,61 @@
...
+}
+
+function LayoutOnboarding({
+  isListMode,
+  onListModeChange,
+}: LayoutOnboardingProps): ReactElement {
+  return (
+    <OnboardingStep
+      title="Cards or list?"
+      description="daily content can be presented in cards or a list, choose what works for you the best"
```
idoshamun, line 20: Is this the final copy? Sounds weird

packages/shared/src/components/modals/OnboardingModal.tsx
sshanzel: This is a validation mechanism depending on each step if you want to put one.

packages/shared/src/components/onboarding/ThemeOnboarding.tsx
```diff
@@ -0,0 +1,33 @@
...
+}
+
+function ThemeOnboarding({
+  selectedTheme,
+  onThemeChange,
+}: ThemeOnboardingProps): ReactElement {
+  return (
+    <OnboardingStep
+      title="Your eyes don’t lie"
+      description="Dark mode will emit less blue light from your screen - which can keep you awake if you use your device before you go to bed"
```
idoshamun, line 18: Again is this the final copy?

packages/shared/src/components/modals/OnboardingModal.tsx
sshanzel: This is setting the local feed filters, this will be soon deprecated since our new system now supports without redirection for authentication.

packages/shared/src/components/modals/SteppedModal.tsx
sshanzel: It is actually a generic component, you will have to put multiple children, and each will be considered a step, this has just abstracted the whole moving forward and backward steps.

packages/shared/src/hooks/useOnboardingModal.ts
```diff
@@ -0,0 +1,108 @@
...
+  onboardingVersion: OnboardingVersion;
+  onFeedPageChanged: (page: MainFeedPage) => unknown;
+}
+
+type ModalCommand = Record<OnboardingVersion, (value: boolean) => unknown>;
+
+const FIRST_TIME_SESSION = 'firstTimeSession';
+const LOGGED_USER_ONBOARDING = 'loggedUserOnboarding';
+
+export const useOnboardingModal = ({
```
idoshamun, line 31: The logic here is quite hard to follow. Eventually, what we want is simple:
Did the user dismiss onboarding or has done it already? Yes, don't show onboarding. Else, show it

packages/shared/src/lib/links.ts
```diff
@@ -1,2 +1,15 @@
 export const getTagPageLink = (tag: string): string =>
   `${process.env.NEXT_PUBLIC_WEBAPP_URL}tags/${encodeURIComponent(tag)}`;
+
+const previewDeployments = [
+  'https://preview.app.daily.dev',
+  'https://preview2.app.daily.dev',
+];
+
+export const checkIsPreviewDeployment = (): boolean => {
```
idoshamun, line 9: May I suggest using environment variables for this? We can set Vercel to embed an env var for every preview deployment. Much more flexible than checking the domains

idoshamun (changes requested):
That's massive amount of work! Thanks for that. I do have some comments and questions

packages/shared/src/components/modals/OnboardingModal.tsx
idoshamun: I think it requires further explanation or refactoring. It's hard to follow

packages/shared/src/components/MainFeedLayout.tsx
idoshamun: I advise to add proper explanation or refactor into its own hook with explicit name

packages/shared/src/components/modals/OnboardingModal.tsx
idoshamun: Ok so I'll wait for the new implementation

packages/shared/src/components/modals/SteppedModal.tsx
sshanzel: Just looked it up and yep, this should not be needed anymore 🚢

packages/shared/src/components/modals/OnboardingModal.tsx
sshanzel: Good point. I'll be more than happy to do it that way.

packages/shared/src/lib/links.ts
sshanzel: Surely. I will come up with a more descriptive variable name in the case.

packages/shared/src/hooks/useOnboardingModal.ts
sshanzel: This was just extracted from the MainFeedLayout to reduce noise on that component. Probably the reason it became complicated is there are now two options for which one to show. This involved data from the boot query, local storage, and requires being in sync.

With that, I'll see what I can do to refactor this one.

packages/shared/src/components/modals/SteppedModal.tsx
sshanzel: My bad. After checking again, it seems the code block for tracking the close event was available on `AuthModal`. So it is necessary for us to have this here.

packages/shared/src/components/modals/OnboardingModal.tsx
sshanzel: After logging in, the local state is not yet updated that the user has successfully logged in (with email and password). Either we will have to do it at the `AuthOptions` component or here.

sshanzel: Seems like there is just an issue with the local environment and `AuthOptions` already fetches the boot data when necessary.

packages/shared/src/hooks/useOnboardingModal.ts
sshanzel: After updating the variables, I think it is much clearer now that this is all about onboarding. With that, unfortunately, each part of the code is necessary. Each part of the use effect is necessary as there are two variables to depend on.

packages/shared/src/components/modals/SteppedModal.tsx
sshanzel: Have now refactored the component to be more loose so the consuming end will have better flexibility, which in turn will have less abstraction which I think was a good part.

packages/shared/src/hooks/useDefaultFeed.ts
```diff
@@ -1,14 +1,52 @@
...
+export default function useDefaultFeed({
+  hasFiltered,
+  hasUser,
+  feed,
+}: UseDefaultFeedProps = {}): string {
+  const isMyFeed = feed === MainFeedPage.MyFeed;
   const [defaultFeed, updateDefaultFeed] = usePersistentContext(
     'defaultFeed',
-    'my-feed',
+    !hasUser || !hasFiltered ? 'popular' : 'my-feed',
```
idoshamun, line 34: Just in case

```suggestion
    (!hasUser || !hasFiltered) ? 'popular' : 'my-feed',
```

packages/shared/src/hooks/useDefaultFeed.ts
```diff
@@ -1,14 +1,52 @@
...
     ['my-feed', 'popular', 'upvoted', 'discussed'],
+    !hasUser || !hasFiltered ? 'popular' : 'my-feed',
```
idoshamun, line 36: Just in case

```suggestion
    (!hasUser || !hasFiltered) ? 'popular' : 'my-feed',
```

packages/shared/src/hooks/useOnboardingModal.ts
```diff
@@ -0,0 +1,108 @@
...
+  user,
+  alerts,
+  isFirstVisit,
+  onboardingVersion,
+  onFeedPageChanged,
+}: UseOnboardingModalProps): UseOnboardingModal => {
+  const { trackEvent } = useContext(AnalyticsContext);
+  const [onboardingMode, setOnboardingMode] = useState(OnboardingMode.Manual);
+  const [isFirstSession, setIsFirstSession, isSessionLoaded] =
```
idoshamun, line 40: I'm not sure we need the first session tracking anymore. All we care about is whether the user has tried onboarding or has configured feed already

idoshamun (changes requested):
A few more comments but we're getting there

packages/shared/src/hooks/useOnboardingModal.ts
sshanzel: Cool. In that case, I will remove this part and use a single variable only.

packages/shared/src/hooks/useDefaultFeed.ts
sshanzel: Unfortunately, the `linting` is not accepting it.

sshanzel:
@idoshamun I have now refactored to only use a single variable in deciding whether to display the onboarding modal.

Also, I refactored the way the new onboarding register filters, it is now centralized so both onboardings will go through the same pipeline, and whenever we need to add a new kind of onboarding, it will have to go through the same process and ensure workability.

packages/shared/src/components/modals/OnboardingModal.tsx
```diff
@@ -0,0 +1,125 @@
...
+import { LoginTrigger } from '../../lib/analytics';
+
+const INTRODUCTION_ADDITIONAL_STEP = 1;
+
+interface OnboardingModalProps extends ModalProps {
+  trigger?: string;
+  onRegistrationSuccess?: () => void;
+}
+
+const backCopy = ['Skip', 'Close', 'Back', 'Back'];
```
rebelchris, line 27: How does this work? What happens when AB decided to only show one screen?

packages/shared/src/components/modals/SteppedModal.tsx
```diff
@@ -83,8 +159,9 @@ function SteppedModal({
   return (
```
rebelchris, line 159: For this component I'm not super happy that the auth is so baked in this thing.

Can't we have the stepped modal be plain generic, it simply steps some children regardless what they are?
Then we can have the FeedFilterModal decide what those children are?

CC: @idoshamun How do you feel about this one?

packages/shared/src/hooks/useDefaultFeed.ts
```diff
@@ -1,14 +1,53 @@
...
   );
 
-  return [defaultFeed, updateDefaultFeed];
+  useEffect(() => {
+    if (
+      defaultFeed !== null &&
+      feed !== null &&
+      feed !== defaultFeed &&
+      feed !== 'default' &&
+      feed !== '/' &&
```
rebelchris, line 45: Suggestion to make it more readable:

```
![null, defaultFeed, 'default', '/'].includes(feed)
```

packages/shared/src/hooks/useOnboardingModal.ts
```diff
@@ -0,0 +1,107 @@
...
+
+  useEffect(() => {
+    if (!user || !shouldUpdateFilters) {
+      return;
+    }
+
+    registerLocalFilters().then(() => {
+      setShouldUpdateFilters(false);
+      modalStateCommand[onboardingVersion]?.(false);
+    });
```
rebelchris, line 74: Can't we simply call this on the modal onFinish? That way we don't have to use useEffect for it?

rebelchris (approved):
Lee, amazing really impressed with how you tackled this PR and how much cleaner it looks.

I do have a couple small questions and remarks.
Happy to discuss in detail if need be.

But well done on this 👏

packages/shared/src/hooks/useOnboardingModal.ts
sshanzel: We can. But on v1 we can't. I wanted them to go through the same process so when we have to add a new kind of onboarding we can ensure the workability of the all the onboarding.

packages/shared/src/components/modals/OnboardingModal.tsx
sshanzel: This should still be as what is intended. First is skip and Close on start of actual onboarding.

packages/shared/src/hooks/useDefaultFeed.ts
sshanzel: I think we should use `every` as includes will make the condition act like an `OR`. Will refactor this one.

packages/shared/src/components/modals/SteppedModal.tsx
sshanzel: It was part of the requirement for the auth to be baked in the onboarding so I thought maybe in the future there will be another requirement with stepped modal to have auth as well.

It would actually be so much better if we can just use the `showLogin` (our generic auth modal) instead of creating this new auth UI.

idoshamun: I also didn't like the fact that it's baked in. But I think we can refactor it in the first moment we need the stepped modal to be more generic. Because it's used only in the feed filters at the moment it's not a critical point imo

packages/shared/src/contexts/FeaturesContext.tsx
```diff
@@ -44,61 +44,64 @@ export type FeaturesContextProviderProps = {
...
-      postCardVersion: getFeatureValue(Features.PostCardVersion, flags),
-      postCardShareVersion: getFeatureValue(
-        Features.PostCardShareVersion,
-        flags,
-      ) as ShareVersion,
-      authVersion: getFeatureValue(Features.AuthenticationVersion, flags),
-    }),
-    [flags],
-  );
+  const features = useMemo(() => getFeatures(flags), [flags]);
```
idoshamun, line 91: I think we can do one `useMemo` for both cases. It's not necessary to separate them

packages/shared/src/contexts/FeaturesContext.tsx
```diff
@@ -44,61 +44,64 @@ export type FeaturesContextProviderProps = {
   flags: IFlags | undefined;
 };
 
+const getFeatures = (flags: IFlags): FeaturesData => {
+  const steps = getFeatureValue(Features.OnboardingSteps, flags);
+  const onboardingSteps = (steps?.split?.('/') || []) as OnboardingStep[];
+
+  return {
+    flags,
```
idoshamun, line 52: Should we even provide the raw flags? I think we should consider ignoring them completely after we parse them. This will force us to parse and add proper support for each flag. cc @rebelchris

idoshamun (changes requested):

packages/shared/src/contexts/FeaturesContext.tsx
sshanzel: Yes, that is true. This was just included before for backwards compatibility. We can probably create a ticket to cleanup all legacy implementations.

idoshamun: cc @rebelchris

packages/shared/src/components/modals/SteppedModal.tsx
rebelchris: Sure let's leave it as is for now.
Whenever another tasks needs the stepped modal we can asses what changes we need.
Resolving for now.

packages/shared/src/components/modals/OnboardingModal.tsx
rebelchris: Not quite sure I understand the purpose then?
Is it not assigned to each step?

Perhaps we can turn this into a dictionary?

packages/shared/src/hooks/useOnboardingModal.ts
rebelchris: I see, ok thanks for the info resolving this.

packages/shared/src/components/modals/OnboardingModal.tsx
sshanzel: I think we will have to keep it an array, the index represents which step it will be reflected in.

The introduction will have index 0 which is the skip and so on.

The footer is abstracted inside the component but we have control over its labels and callbacks if any.

packages/shared/src/hooks/useOnboardingModal.ts
sshanzel: Ergonomics existed even before remote became the new norm.

idoshamun (approved):
I did QA on desktop only and it looks great except two issues I raised in the channel. The rest of the issues are fixed. Didn't check mobile though

el (approved):

sshanzel:
@rebelchris though you have approved and I requested for review again, I will be merging this to the intermediate branch. If you have further comments around this, please let me know on that PR so we can sort it there and close this one out.

PR #1347, opened 2022-11-01
fix: applying new headline button guideline

sshanzel:
- New guideline for button component.
- Applied the new component when the user has filtering applied.
- Tooltip is removed as it is no longer necessary since this is now a button with its own label.

Preview: 
image

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-538 #done
WT-539 #done


packages/shared/src/styles/components/buttons.css
```diff
@@ -31,6 +31,14 @@
     border-radius: 0.875rem;
   }
 
+  &.headline {
+    height: 3rem;
+    padding: 0 1.5rem;
+    font-size: 1.0625rem;
+    border-radius: 0.875rem;
+    color: var(--theme-label-primary);
+  }
```
rebelchris, line 40: Is this a recurring thing?
Why don't we set them as classNames?

packages/shared/src/components/MainFeedLayout.tsx
```diff
@@ -254,6 +240,23 @@ export default function MainFeedLayout({
 
   const hasFiltered = feedName === MainFeedPage.MyFeed && !alerts?.filter;
 
+  /* eslint-disable react/no-children-prop */
+  const feedHeading = {
+    [MainFeedPage.MyFeed]: !hasFiltered ? (
```
rebelchris, line 245: Optionally maybe we can create a MyFeedHeading component which will do the conditional rendering? (Just to clean this up some more)

rebelchris (approved):
Nice work! 

Just some small comments, no breaking from my side.

packages/shared/src/components/MainFeedLayout.tsx
sshanzel: Good point. Will make this part cleaner. Let me refactor this one.

packages/shared/src/styles/components/buttons.css
sshanzel: It is due to using `primary`, `secondary`, etc., automatically applying their own color of the label, this is to force it, regardless of which type of button was used.

rebelchris (approved):
Thanks for the refactor.

PR #1349, opened 2022-11-02
feat: new onboarding event analytics

sshanzel:
- All analytics for when the user has selected some topics/tags are inside the hook it uses to update the local data (meaning, no additional tracking needed).
- Applied tracking of impression for the new onboarding version modal.
- Also applied tracking when the auth modal opens from the generic component named `SteppedModal`, which parameters will be coming from the props.
- Will update the taxonomy sheet once we verify the correct tracking events.

Tagging @idoshamun for review to ensure all requirements for events are met since it was mentioned in the spec to there be no additional ones.
[Reference](https://dailydotdev.atlassian.net/wiki/spaces/HAN/pages/198049793/Feed+filters#New-events):
image

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-670 #done


packages/shared/src/components/modals/SteppedModal.tsx
```diff
@@ -117,6 +120,12 @@ function SteppedModal({
       return onFinish?.();
     }
 
+    trackEvent({
+      event_name: 'click',
+      target_type: trigger,
+      target_id: targetId,
+    });
```
rebelchris, line 127: Which one is this in the taxonomy?
I also don't see it in the changed/new flow

packages/shared/src/components/modals/OnboardingModal.tsx
```diff
@@ -127,6 +138,15 @@ function OnboardingModal({
     }
   };
 
+  useEffect(() => {
+    trackEvent({
+      event_name: 'impression',
+      target_type: 'onboarding modal',
+      target_id: 'v2',
+      extra: JSON.stringify({ origin: mode }),
+    });
```
rebelchris, line 147: This is not inline with the proposed taxonomy:

https://docs.google.com/spreadsheets/d/18Lv7zXges9QfVX5VYL1a-Hyl0e1sQ3sLr0OK8YZWKXI/edit#gid=0

(My feed onboarding)

rebelchris (changes requested):
I think you may have missed some properties.

I also see "my feed onboarding skip" in the taxonomy, could you ensure that's still there?

idoshamun:
@sshanzel I'll wait with my review until we conclude the other PR

packages/shared/src/components/modals/OnboardingModal.tsx
```diff
@@ -140,6 +160,7 @@ function OnboardingModal({
       invalidMessage={invalidMessage}
       isFirstStepIntroduction
       isLastStepLogin
+      targetId="onboarding-v2"
```
el, line 163: Can we use an enum here?

packages/shared/src/components/modals/OnboardingModal.tsx
```diff
@@ -127,6 +138,15 @@ function OnboardingModal({
     }
   };
 
+  useEffect(() => {
+    trackEvent({
+      event_name: 'impression',
```
el, line 143: I think we should use enums for the values

el (approved):

packages/shared/src/components/modals/OnboardingModal.tsx
sshanzel: Totally agree with both of you.

packages/shared/src/components/modals/OnboardingModal.tsx
sshanzel: Thank you for pointing this out, I have now a better understanding of the taxonomy sheet.

sshanzel:
I've accidentally merged the changes. Let me re-open a new PR. Keyboard issues 🤦

PR #1354, opened 2022-11-03
Wt 309 share article link

el:
## Changes

Overrides the share version when share sheet is available

### DIFF
--- a/packages/shared/src/components/Feed.tsx
+++ b/packages/shared/src/components/Feed.tsx
@@ -42,6 +42,7 @@ import {
 import { useSharePost } from '../hooks/useSharePost';
 import { Origin } from '../lib/analytics';
 import ShareOptionsMenu from './ShareOptionsMenu';
+import { ShareVersion } from '../lib/featureValues';
 
 export type FeedProps<T> = {
   feedName: string;
@@ -118,11 +119,12 @@ export default function Feed<T>({
 }: FeedProps<T>): ReactElement {
   const {
     postCardVersion,
-    postCardShareVersion,
+    postCardShareVersion: shareVersion,
     postModalByDefault,
     postEngagementNonClickable,
     showCommentPopover,
   } = useContext(FeaturesContext);
+  const postCardShareVersion = navigator.share ? ShareVersion.V2 : shareVersion;
   const { trackEvent } = useContext(AnalyticsContext);
   const currentSettings = useContext(FeedContext);
   const { user } = useContext(AuthContext);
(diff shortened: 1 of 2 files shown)


packages/shared/src/components/Feed.tsx
```diff
@@ -118,11 +119,12 @@ export default function Feed<T>({
 }: FeedProps<T>): ReactElement {
   const {
     postCardVersion,
-    postCardShareVersion,
+    postCardShareVersion: shareVersion,
     postModalByDefault,
     postEngagementNonClickable,
     showCommentPopover,
   } = useContext(FeaturesContext);
+  const postCardShareVersion = navigator.share ? ShareVersion.V2 : shareVersion;
```
idoshamun, line 127: Just in case :)

```suggestion
  const postCardShareVersion = navigator?.share ? ShareVersion.V2 : shareVersion;
```

el: Is there a possibility that navigator is not available? (apart from using in node.js)

packages/shared/src/components/ShareOptionsMenu.tsx
```diff
@@ -56,7 +56,7 @@ export default function ShareOptionsMenu({
   onHidden,
   contextId = 'share-context',
 }: ShareOptionsMenuProps): ReactElement {
-  const link = post && post?.permalink;
```
el, line 59: Why my previous commit here? It has already been merged to the main. And what can I do to avoid it? + @rebelchris @idoshamun

packages/shared/src/components/Feed.tsx
idoshamun: Not sure. Anyway it's not something major, we can keep it as it is :)

packages/shared/src/components/ShareOptionsMenu.tsx
idoshamun: Not sure actually. Maybe you merged the updates into your feature branch and this caused this problem

packages/shared/src/components/Feed.tsx
rebelchris: Test could potentially break, we normally do check if navigator exists in other places, but not sure if it was legacy.

packages/shared/src/components/ShareOptionsMenu.tsx
rebelchris: Probably if you repull from main it will change this :)

rebelchris (approved):

packages/shared/src/components/ShareOptionsMenu.tsx
el: this is the same previous branch. We used to use the same branch for small fixes. I guess I need to use a new branch for each pr because of squash?

rebelchris: Ah yes, we normally repull main -> then create a new individual branch for hotfixes.

PR #1356, opened 2022-11-03
feat: modal guidelines

el:
## Changes

- New modal parent component to be used by all modals in the app

## Events

No new events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [x] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-716


packages/shared/src/components/modals/UserSettingsModal.tsx
```diff
@@ -10,7 +11,19 @@ export function UserSettingsModal({
   ...modalProps
 }: ModalProps): ReactElement {
   useTrackModal({ isOpen: modalProps.isOpen, title: 'feed settings' });
-
+  // return (
```
el, line 14: @rebelchris @sshanzel This is how we will use the new modals

packages/shared/src/components/modals/common/Modal.tsx
```diff
@@ -0,0 +1,71 @@
...
+  const modalKindToClassName = {
+    [ModalKind.FixedCenter]: '',
+    [ModalKind.FlexibleCenter]: 'h-[40rem]',
+    [ModalKind.FlexibleTop]: classNames(
+      'mt-10',
+      size === ModalSize.Medium && 'mt-20',
+      size === ModalSize.Large && 'mt-14',
+    ),
+  };
+  const modalSizeToClassName = {
```
el, line 38: responsiveness support is missing, but the styles for the desktop sizes are here

packages/shared/src/components/modals/common/Modal.tsx
```diff
@@ -0,0 +1,71 @@
...
+        modalKindToOverlayClassName[kind],
+      )}
+      className={classNames(
+        'focus:outline-none modal flex flex-col relative w-full max-h-[calc(100vh-5rem)] overflow-y-auto items-center bg-theme-bg-tertiary shadow-2 border border-theme-divider-secondary rounded-16',
+        className,
+        modalKindToClassName[kind],
+        modalSizeToClassName[size],
+      )}
+    >
+      <ModalPropsContext.Provider value={{ size, kind, onRequestClose }}>
```
el, line 59: I passed the modal props to the children components using context to avoid requiring the user of the component to pass to the children components again (like request close to `<Modal.Header/>`

packages/shared/src/components/modals/UserSettingsModal.tsx
```diff
@@ -1,5 +1,6 @@
 import React, { ReactElement } from 'react';
 import { ResponsiveModal } from './ResponsiveModal';
+// import { Modal } from './common/Modal';
```
el, line 3: All we need to import is `{ Modal }` to use the component

packages/shared/src/components/modals/common/Modal.tsx
```diff
@@ -0,0 +1,71 @@
...
+    </ReactModal>
+  );
+}
+
+Modal.Size = ModalSize;
+Modal.Kind = ModalKind;
+Modal.Header = ModalHeader;
+Modal.Body = ModalBody;
+Modal.Footer = ModalFooter;
+Modal.Sidebar = ModalSidebar;
```
rebelchris, line 71: Why do we choose the dot notation?
Not against it, just something we haven't used anywhere else in favor of things like ModalSize.

packages/shared/src/components/modals/common/types.ts
```diff
@@ -0,0 +1,26 @@
...
+  onRequestClose: null | ((event: MouseEvent | KeyboardEvent) => void);
+  kind: ModalKind;
+  size: ModalSize;
+};
+
+export const ModalPropsContext = createContext<ModalContextProps>({
+  onRequestClose: null,
+  kind: ModalKind.FlexibleCenter,
+  size: ModalSize.Medium,
+});
```
rebelchris, line 26: Why do we favor context here instead of passing directly to the modal?

packages/shared/src/components/modals/common/ModalFooter.tsx
```diff
@@ -0,0 +1,9 @@
+import React, { ReactElement, ReactNode } from 'react';
+
+export type ModalFooterProps = {
+  children?: ReactNode;
+};
+
+export function ModalFooter({ children }: ModalFooterProps): ReactElement {
+  return <footer>{children}</footer>;
```
rebelchris, line 8: Does this one not have any styling?

packages/shared/src/components/modals/common/ModalBody.tsx
```diff
@@ -0,0 +1,26 @@
...
+import { ModalKind, ModalPropsContext, ModalSize } from './types';
+
+export type ModalBodyProps = {
+  children?: ReactNode;
+};
+
+export function ModalBody({ children }: ModalBodyProps): ReactElement {
+  return (
+    <ModalPropsContext.Consumer>
+      {({ kind, size }) => (
```
rebelchris, line 12: This comes from the context?
Might be a bit unclear at this stage.

packages/shared/src/components/modals/common/ModalBody.tsx
```diff
@@ -0,0 +1,26 @@
...
+        <section
+          className={classNames(
+            'overflow-auto relative w-full h-full shrink max-h-full',
+            kind === ModalKind.FlexibleTop && size === ModalSize.Large
```
rebelchris, line 16: This might be more readable to abstract outside the render as a condition?

rebelchris (changes requested):
Wow really well done, amazing setup.
My remarks are more around code styling and choices if you could answer those please.

packages/shared/src/components/Settings.tsx
```diff
@@ -114,7 +114,7 @@ export default function Settings({
   }, []);
 
   return (
-    <div className={classNames('flex', 'flex-col p-6', className)} {...props}>
```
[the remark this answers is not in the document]
el: Settings has its own padding, but instead it should be handled by the modal

packages/shared/src/components/modals/common/Modal.tsx
el: I am working on it, if we merge this PR, I will open a new one with that

packages/shared/src/components/modals/common/Modal.tsx
```diff
@@ -0,0 +1,71 @@
...
+  };
+  const modalSizeToClassName = {
+    [ModalSize.XSmall]: 'max-w-[21.25]',
+    [ModalSize.Small]: 'max-w-[26.25rem]',
+    [ModalSize.Medium]: 'max-w-[35rem]',
+    [ModalSize.Large]: 'max-w-[63.75rem]',
+  };
+  return (
+    <ReactModal
+      isOpen
```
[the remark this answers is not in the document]
el: We can use the conditional rendering in outer scope

packages/shared/src/components/modals/common/Modal.tsx
el: So that we can import and use everything modal related with the same class.

packages/shared/src/components/modals/common/ModalFooter.tsx
el: This is not done yet

packages/shared/src/components/modals/common/types.ts
el: We do pass it directly to the modal, the context is used inside the modal

el: Happy to jump on a call if it is not clear

packages/shared/src/components/modals/common/ModalBody.tsx
el: Done

packages/shared/src/components/modals/common/Modal.tsx
```diff
@@ -0,0 +1,71 @@
...
+      size === ModalSize.Medium && 'mt-20',
+      size === ModalSize.Large && 'mt-14',
+    ),
+  };
+  const modalSizeToClassName = {
+    [ModalSize.XSmall]: 'max-w-[21.25]',
+    [ModalSize.Small]: 'max-w-[26.25rem]',
+    [ModalSize.Medium]: 'max-w-[35rem]',
+    [ModalSize.Large]: 'max-w-[63.75rem]',
+  };
```
idoshamun, line 43: constants can go outside the component :)

packages/shared/src/components/modals/common/Modal.tsx
idoshamun: Would you not favor changing to `export ModalSize`?
You can still import everything from the same file.
Another common practice is to have an index.ts file in the directory that only exports whatever is necessary.
Something like this: https://github.com/dailydotdev/pulumi-common/blob/main/src/index.ts
Both ways are somewhat opinionated so whatever works for you works for me :)

packages/shared/src/components/modals/common/ModalHeader.tsx
```diff
@@ -0,0 +1,18 @@
...
+
+export type ModalHeaderProps = {
+  title?: ReactNode;
+};
+
+export function ModalHeader({ title }: ModalHeaderProps): ReactElement {
+  return (
+    <header className="flex justify-between items-center py-4 px-6 w-full h-14 border-b border-theme-divider-tertiary">
+      {!!title && <h3 className="font-bold typo-title3">{title}</h3>}
+      <ModalPropsContext.Consumer>
```
idoshamun, line 13: I'm more into `useContext` to keep the html part a bit more clean. My personal taste

packages/shared/src/components/modals/common/Prompt.tsx
```diff
@@ -0,0 +1,9 @@
+import React, { ReactElement } from 'react';
+
+export type PromptProps = {
+  title: string;
+};
+
+export function Prompt({ title }: PromptProps): ReactElement {
```
idoshamun, line 7: Seems like an unnecessary component, is it not?

packages/shared/src/components/modals/common/ModalSidebar.tsx
```diff
@@ -0,0 +1,8 @@
+import React, { ReactElement, ReactNode } from 'react';
+
+export type ModalSidebarProps = {
+  children?: ReactNode;
+};
+export function ModalSidebar({ children }: ModalSidebarProps): ReactElement {
```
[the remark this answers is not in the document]
el: I will update this component, (it is not done yet)

packages/shared/src/components/modals/common/Prompt.tsx
el: I will update this component, (it is not done yet)

packages/shared/src/components/modals/common/ModalHeader.tsx
el: Yeah I can change ^_^

packages/shared/src/components/modals/common/Modal.tsx
el: Done 👍

packages/shared/src/components/modals/common/Modal.tsx
el: Yeah, that was my first idea as well, but I went with this approach because it is much easier to use the component.
You just import the Modal and all the child components will be in your fingertips. We talked with @rebelchris about this and decided to give it a try, happy to jump on a call if you want more context or have any feedback about it.

idoshamun: This was just my two cents. You make the call 🤙
Thanks for the info

PR #1359, opened 2022-11-07
feat: support responsive design in modals

el:
- Support responsive designs for the modals

## Events
- [x] No new events

- [x] MobileL (420px)
- [x] Tablet (656px)
- [x] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-729 #done

### DIFF
--- a/packages/shared/src/components/modals/common/Modal.tsx
+++ b/packages/shared/src/components/modals/common/Modal.tsx
@@ -13,6 +13,39 @@ export type ModalProps = ReactModal.Props & {
   size?: ModalSize;
 };
 
+const modalKindToOverlayClassName: Record<ModalKind, string> = {
+  [ModalKind.FixedCenter]: 'mobileL:justify-center',
+  [ModalKind.FlexibleCenter]: 'justify-center',
+  [ModalKind.FlexibleTop]: 'm-0',
+};
+const modalKindAndSizeToOverlayClassName = {
+  [ModalKind.FlexibleTop]: {
+    [ModalSize.Medium]: 'mobileL:pt-10',
+  },
+};
+const modalKindToClassName: Record<ModalKind, string> = {
+  [ModalKind.FixedCenter]:
+    'h-full max-h-[calc(100vh-2.5rem)] mobileL:h-[40rem] mobileL:max-h-[calc(100vh-5rem)] mt-10 mobileL:mt-0',
+  [ModalKind.FlexibleCenter]:
+    'mx-4 max-w-[calc(100vw-2rem)] max-h-[min(calc(100vh),40rem)] mobileL:max-h-[min(calc(100vh-5rem),40rem)]',
+  [ModalKind.FlexibleTop]:
+    'm-0 mobileL:mt-10 mobileL:h-auto mobileL:max-h-[calc(100vh-2.5rem)]',
+};
+const modalKindAndSizeToClassName = {
+  [ModalKind.FlexibleTop]: {
+    [ModalSize.Medium]:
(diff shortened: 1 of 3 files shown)


packages/shared/src/components/modals/common/Modal.tsx
```diff
@@ -13,6 +13,43 @@ export type ModalProps = ReactModal.Props & {
...
+> = {
+  [ModalKind.FlexibleTop]: {
+    [ModalSize.Medium]: 'mobileL:pt-10',
+  },
+};
+const modalKindToClassName: Record<ModalKind, string> = {
+  [ModalKind.FixedCenter]:
+    'h-full max-h-[calc(100vh-2.5rem)] mobileL:h-[40rem] mobileL:max-h-[calc(100vh-5rem)] mt-10 mobileL:mt-0',
+  [ModalKind.FlexibleCenter]:
+    'mx-4 max-w-[calc(100vw-2rem)] max-h-[min(calc(100vh),40rem)] mobileL:max-h-[min(calc(100vh-5rem),40rem)]',
```
rebelchris, line 32: Nothing wrong with this, but when using vh and calc it's very important to solidly test it out on actual mobile devices.
I know safari on iPhone has some issues with vh.

packages/shared/src/components/modals/common/ModalBody.tsx
```diff
@@ -1,26 +1,16 @@
...
-            kind === ModalKind.FlexibleTop && size === ModalSize.Large
-              ? 'p-8'
-              : 'p-6',
-          )}
-        >
-          {children}
-        </section>
-      )}
-    </ModalPropsContext.Consumer>
+  const { kind, size } = useContext(ModalPropsContext);
```
rebelchris, line 10: Amazing, love this refactor ❤️

packages/shared/src/components/modals/common/ModalHeader.tsx
```diff
@@ -1,18 +1,27 @@
...
 import { ModalPropsContext } from './types';
 
 export type ModalHeaderProps = {
-  title?: ReactNode;
+  children?: ReactNode;
+  title?: string;
 };
 
-export function ModalHeader({ title }: ModalHeaderProps): ReactElement {
+const ModalHeaderTitle = classed('h3', 'font-bold typo-title3');
```
rebelchris, line 11: I'm all for these, but why do we used it for the title, but not the header?
Perhaps we should find a uniform for this?

rebelchris (approved):
Looks great!
I would love to give this a proper test on actual devices though.

Doesn't block merging, but please let me know which modal I can eventually test.

packages/shared/src/components/modals/common/Modal.tsx
el: Yeap, makes perfect sense, I left the mobile device tests out of this. Let me add a jira just for this :
https://dailydotdev.atlassian.net/browse/WT-734

packages/shared/src/components/modals/common/ModalHeader.tsx
el: I used for the header to be able to expose it outside of this component. Do we need it for everything?
My thought was to only use it for exposed components.

el:
> Looks great! I would love to give this a proper test on actual devices though.
> 
> Doesn't block merging, but please let me know which modal I can eventually test.
@rebelchris 


After todays PR, 
Most of the Modals can be moved to the new structure. 
We can have a quick call about this 👍

PR #1361, opened 2022-11-09
fix: cleanup navigator use with it's own function

rebelchris:
- It seems impossible to test the full static behaviour so I couldn't add specific test cases for this.
- However decided to create a new `getNavigator` function which can be used on the Next static side to get the navigator.
- This removes the need for the useEffect hook.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-727 #done

### DIFF
--- a/packages/shared/src/components/Feed.tsx
+++ b/packages/shared/src/components/Feed.tsx
@@ -4,7 +4,6 @@ import React, {
   ReactNode,
   useContext,
   useEffect,
-  useState,
 } from 'react';
 import classNames from 'classnames';
 import dynamic from 'next/dynamic';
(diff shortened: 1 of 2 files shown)


el (approved):
No comments. Thanks for taking care of this.

sshanzel (approved):
So happy to see we can always make sure the navigator works! Thank you!

Probably an additional abstraction in the future, we should have a file that exports everything `window` related (like how navigator is part of window),  and make use of getter functions.

Ex:
```
export const thisWindow = {
  get window(): Window {
    if (typeof window === 'undefined') {
      return null;
    }

    return window;
  },
  get navigator(): Navigator {
    if (!this.window) {
      return null;
    }

    return this.window.navigator;
  },
};

console.log(thisWindow.navigator?.something);
```

So whenever we're gonna need something window-related, it must be abstracted as above so we can be sure it is working and we will be able to utilize optional chaining.

PR #1362, opened 2022-11-09
feat: modal tabs and footer changes

el:
## Changes

- Add support for tabs to the modals

- [x] MobileL (420px)
- [x] Tablet (656px)
- [x] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-731 #done

### DIFF
--- a/packages/shared/src/components/modals/NewCommentModal.tsx
+++ b/packages/shared/src/components/modals/NewCommentModal.tsx
@@ -19,6 +19,7 @@ import { Post } from '../../graphql/posts';
 import { ModalCloseButton } from './ModalCloseButton';
 import DiscardActionModal from './DiscardActionModal';
 import { useRequestProtocol } from '../../hooks/useRequestProtocol';
+// import { Modal } from './common/Modal';
 
 interface CommentVariables {
   id: string;
@@ -147,6 +148,55 @@ export default function NewCommentModal({
   useEffect(() => {
     onInputChange?.(input);
   }, [input]);
+
+  // return (
+  //   <Modal
+  //     contentRef={modalRef}
+  //     onRequestClose={confirmClose}
+  //     kind={Modal.Kind.FlexibleCenter}
+  //     size={Modal.Size.Small}
+  //     tabs={['Write', 'Preview']}
+  //     {...props}
+  //   >
+  //     <Modal.Header />
+  //     <Modal.Body tab="Write">
+  //       <CommentBox
+  //         {...props}
+  //         onInput={setInput}
+  //         input={input}
+  //         editId={editId}
+  //         errorMessage={errorMessage}
+  //         sendingComment={sendingComment}
+  //         sendComment={sendComment}
+  //       />
+  //     </Modal.Body>
+  //     <Modal.Body tab="Preview">
+  //       {isPreview && previewContent?.preview && (
+  //         <Markdown
+  //           content={previewContent.preview}
+  //           appendTooltipTo={props.parentSelector}
+  //         />
+  //       )}
+  //     </Modal.Body>
+  //     <Modal.Footer>
+  //       <Button
+  //         disabled={!input?.trim().length || input === props.editContent}
+  //         loading={sendingComment}
+  //         onClick={sendComment}
+  //         className="mt-auto ml-auto btn-primary-avocado"
+  //       >
+  //         {editId ? 'Update' : 'Comment'}
+  //       </Button>
+  //     </Modal.Footer>
+  //     <DiscardActionModal
+  //       isOpen={showDiscardModal}
+  //       onRequestClose={() => setShowDiscardModal(false)}
+  //       rightButtonAction={onRequestClose}
+  //       shouldCloseOnOverlayClick={false}
+  //       parentSelector={props.parentSelector}
+  //     />
+  //   </Modal>
+  // );
 
   return (
     <ResponsiveModal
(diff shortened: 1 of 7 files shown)


packages/shared/src/components/modals/NewCommentModal.tsx
```diff
@@ -148,6 +149,55 @@ export default function NewCommentModal({
...
+  //   <Modal
+  //     contentRef={modalRef}
+  //     onRequestClose={confirmClose}
+  //     kind={Modal.Kind.FlexibleCenter}
+  //     size={Modal.Size.Small}
+  //     tabs={['Write', 'Preview']}
+  //     {...props}
+  //   >
+  //     <Modal.Header />
+  //     <Modal.Body tab="Write">
```
el, line 162: If we pass a tab to Modal.Body or Modal.Footer, it will only render contents if the active tab matches.

packages/shared/src/components/modals/common/Modal.tsx
```diff
@@ -51,13 +53,18 @@ const modalSizeToClassName: Record<ModalSize, string> = {
...
+  defaultTab,
   className,
   overlayClassName,
   children,
   kind = ModalKind.FlexibleCenter,
   size = ModalSize.Medium,
   onRequestClose,
+  tabs,
 }: ModalProps): ReactElement {
+  const [activeTab, onTabChange] = useState<string | undefined>(
```
el, line 65: we are making the tabs first class citizen of the modals, therefor no need for any workarounds

packages/shared/src/components/modals/common/ModalFooter.tsx
```diff
@@ -1,12 +1,30 @@
...
+  className?: string;
+  justify?: 'end' | 'center' | 'between';
+  tab?: string;
 };
 
-export function ModalFooter({ children }: ModalFooterProps): ReactElement {
+export function ModalFooter({
+  children,
+  className,
+  justify = 'end',
```
el, line 15: FYI, @rebelchris

packages/shared/src/components/modals/common/Modal.tsx
rebelchris: Like this approach that we have the option to overwrite if need be

packages/shared/src/components/modals/common/ModalFooter.tsx
rebelchris: Perfect ❤️

packages/shared/src/components/modals/common/ModalFooter.tsx
```diff
@@ -1,12 +1,30 @@
...
+  tab,
+}: ModalFooterProps): ReactElement {
+  const { activeTab } = useContext(ModalPropsContext);
+  if (tab && tab !== activeTab) return null;
```
rebelchris, line 19: The footer is not part of the tab right, why do we hide it?

packages/shared/src/components/modals/common/ModalTabs.tsx
```diff
@@ -0,0 +1,33 @@
...
+          onClick={() => onTabChange(tab)}
+          type="button"
+          role="menuitem"
+        >
+          {tab}
+          {tab === activeTab && (
+            <div
+              className="absolute mx-auto w-4 h-px bg-theme-label-primary"
+              style={{ bottom: '-0.75rem', left: 'calc(50% - 0.5rem)' }}
+            />
```
rebelchris, line 27: Is this not just the active state of a button?
Pretty sure it should be a class on the button above rather then a extra element.

rebelchris (changes requested):
Amazing!

Lovely to see how reusable this is.
Just some minor changes requested.

packages/shared/src/components/modals/common/ModalFooter.tsx
el: No, it can be, for example for the new comment modal, footer changes based on the tab.

packages/shared/src/components/modals/common/ModalTabs.tsx
el: The active state has a background and a border below it. Thats why this is a separate element
image


The design of the tabs has been changed slightly
https://www.sketch.com/s/2867ec63-0966-4647-b1d7-b6fe696ba251/a/agogLmZ

packages/shared/src/components/modals/common/ModalFooter.tsx
el: https://www.sketch.com/s/2867ec63-0966-4647-b1d7-b6fe696ba251/a/agogLmZ

rebelchris: How would it be set for modal with tabs that have 1 generic footer?

packages/shared/src/components/modals/common/ModalTabs.tsx
rebelchris: Oh right!
Sorry missed that thing indeed 🙏

packages/shared/src/components/modals/common/ModalFooter.tsx
el: If you dont pass a tab, it will render it

```tsx
<Footer> Generic footer that renders for all tabs</Footer>
```

```tsx
<Footer tab="Preview">Special footer that renders for only if the Preview tab is active</Footer>
```

PR #1363, opened 2022-11-09
feat: confirmation modal update for the new modal guidelines

el:
## Changes

- My first idea was to create whole new prompt component but it would change everything with not much added benefit, so just updating the wrapper component will be enough once we transition.

WT-733 #done


sshanzel (approved):
Code-wise looks good to me! 🚢

rebelchris (approved):
Perfect for me :)

PR #1364, opened 2022-11-10
feat: new onboarding event analytics

sshanzel:
I've accidentally merged the previous PR. I just raised this new one, which is why the comment history was not kept.

- All analytics for when the user has selected some topics/tags are inside the hook it uses to update the local data (meaning, no additional tracking needed).
- Applied tracking of impression for the new onboarding version modal.
- Also applied tracking when the auth modal opens from the generic component named SteppedModal, which parameters will be coming from the props.
Tagging @idoshamun for review to ensure all requirements for events are met since it was mentioned in the spec to there be no additional ones.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done


packages/shared/src/components/modals/LegacyOnboardingModal.tsx
```diff
@@ -58,7 +58,7 @@ const getFooterButton = ({
 };
 
 interface LegacyOnboardingModalProps extends ModalProps {
```
rebelchris, line 60: Perhaps also typecast the target_type here, just so we 100% it aligns

packages/shared/src/components/modals/OnboardingModal.tsx
```diff
@@ -118,6 +164,19 @@ function OnboardingModal({
     return validations;
   }, [onboardingSteps]);
 
+  useEffect(() => {
+    trackEvent({
+      event_name: AnalyticsEvent.Impression,
+      target_type: TargetType.MyFeedModal,
+      target_id: OnboardingVersion.V2,
```
rebelchris, line 170: This is not aligned with how we track target_id for the `my feed button` just double sure with @idoshamun what it should be for both.

packages/shared/src/hooks/useOnboardingModal.ts
```diff
@@ -69,6 +70,12 @@ export const useOnboardingModal = ({
     }
 
     registerLocalFilters().then(() => {
+      trackEvent({
+        event_name: AnalyticsEvent.CompleteOnboarding,
+        target_type: TargetType.MyFeedModal,
+        target_id: OnboardingVersion.V1,
+        extra: JSON.stringify({ origin: onboardingMode }),
+      });
```
rebelchris, line 78: @idoshamun this is empty in the taxonomy?

rebelchris (changes requested):
We have to double check behaviour on `target_id` I see we mix between actually placements and versions.

Also missing these events:
line 86 (manage tag)
line 88 (search tags)

packages/shared/src/components/modals/LegacyOnboardingModal.tsx
sshanzel: Oh, sorry. I think I did some cleanups and the `target_type` was not needed so I removed it already.

packages/shared/src/components/modals/OnboardingModal.tsx
sshanzel: We have two impressions, the button which has `my feed button` and the `my feed modal` which is for when the modals were seen.

sshanzel: Most of the `tracking_id` I am actually not 100% on their correctness. I think it would be amazing if we can have a guideline on how to identify it, but if the impact on PR is not too much, then probably it is not necessary.

sshanzel:
@rebelchris requested for review, though the `target_id` values are still subject to confirmation with Ido.

packages/shared/src/components/filters/MyFeedHeading.tsx
```diff
@@ -25,6 +27,16 @@ function MyFeedHeading({
     return <FeedHeading>My feed</FeedHeading>;
   }
 
+  const { trackEvent } = useContext(AnalyticsContext);
+
+  const onClick = () => {
+    trackEvent({
```
rebelchris, line 33: You are missing the event name here.
```suggestion
    trackEvent({
    event_name: 'click manage tags'
```

Preferably convert this into the enum that holds Impression though.

packages/shared/src/components/filters/TagsFilter.tsx
```diff
@@ -58,10 +62,29 @@ export default function TagsFilter({
...
+        `${apiUrl}/graphql`,
+        SEARCH_TAGS_QUERY,
+        { query },
+      );
+
+      if (!query) {
+        return data;
+      }
+
+      trackEvent({
```
rebelchris, line 76: Missing event_name: 'search tags'

rebelchris (approved):
Almost there!

And let's wait for Ido/Sab untill merging, feel free to ask Sab about the target_id's on Slack :)

rebelchris (changes requested):
Almost there.

packages/shared/src/components/filters/MyFeedHeading.tsx
idoshamun: Chris is right there can't be an event without event_name
Also, target_id and target_type are not required here. I left a comment for Sab on the taxonomy.

packages/shared/src/components/filters/TagsFilter.tsx
```diff
@@ -58,10 +62,29 @@ export default function TagsFilter({
...
+        target_type: TargetType.SearchTag,
+        target_id: TargetId.TagsFilter,
```
idoshamun, line 78: target_id and target_type are not required here. I left a comment for Sab on the taxonomy

packages/shared/src/components/modals/OnboardingModal.tsx
```diff
@@ -4,49 +4,79 @@ import React, {
...
 
-  const onCloseConfirm = (
-    e: MouseEvent | KeyboardEvent,
-    stepAtClose?: number,
-  ) => {
+  const onCloseConfirm = (e: MouseEvent | KeyboardEvent) => {
+    if (
+      mode === OnboardingMode.Auto &&
+      onboardingVersion === OnboardingVersion.V2
+    ) {
```
idoshamun, line 69: Why do you apply this condition for tracking?

packages/shared/src/components/modals/OnboardingModal.tsx
```diff
@@ -4,49 +4,79 @@ import React, {
...
+      const screen = getScreen(onboardingSteps, step);
+      trackEvent({
+        event_name: AnalyticsEvent.OnboardingSkip,
+        target_type: backCopy[step],
```
idoshamun, line 73: Do we experiment with the copy of the button?

packages/shared/src/components/modals/OnboardingModal.tsx
```diff
@@ -4,49 +4,79 @@ import React, {
...
+        target_id: OnboardingVersion.V2,
```
idoshamun, line 74: Not needed

packages/shared/src/components/modals/OnboardingModal.tsx
```diff
@@ -66,10 +96,30 @@ function OnboardingModal({
       onClose(e);
       return true;
     }
+
+    trackEvent({
+      event_name: AnalyticsEvent.ClickOnboardingBack,
+      target_type: backCopy[beforeStep],
+      target_id: OnboardingVersion.V2,
```
idoshamun, line 103: Not needed

packages/shared/src/components/modals/OnboardingModal.tsx
```diff
@@ -66,10 +96,30 @@ function OnboardingModal({
...
+      nextCopy[beforeStep] ||
+      getForwardLabel({
+        step: beforeStep,
+        length: onboardingSteps.length,
+      });
+    const screen = getScreen(onboardingSteps, beforeStep);
+    trackEvent({
+      event_name: AnalyticsEvent.ClickOnboardingNext,
+      target_type: label,
+      target_id: OnboardingVersion.V2,
```
idoshamun, line 120: Not needed

packages/shared/src/components/modals/OnboardingModal.tsx
idoshamun: @rebelchris can you elaborate on what's not aligned please? It seems good to me

packages/shared/src/hooks/useOnboardingModal.ts
```diff
@@ -51,10 +52,11 @@ export const useOnboardingModal = ({
...
   const onCloseOnboardingModal = () => {
-    if (onboardingMode === OnboardingMode.Auto) {
-      trackEvent({
-        event_name: 'my feed onboarding skip',
-      });
+    if (
+      onboardingMode === OnboardingMode.Auto &&
+      onboardingVersion === OnboardingVersion.V1
+    ) {
+      trackEvent({ event_name: AnalyticsEvent.OnboardingSkip });
```
idoshamun, line 59: Why do we track the onboarding skip event here as well? Isn't it better to track in a centralized place?

packages/shared/src/hooks/useOnboardingModal.ts
```diff
@@ -69,6 +71,12 @@ export const useOnboardingModal = ({
     }
 
     registerLocalFilters().then(() => {
+      trackEvent({
+        event_name: AnalyticsEvent.CompleteOnboarding,
+        target_type: TargetType.MyFeedModal,
+        target_id: `onboarding-${onboardingVersion}`,
+        extra: JSON.stringify({ origin: onboardingMode }),
```
idoshamun, line 78: Not needed

idoshamun (changes requested):
I added some comments and also my main concern is that sometimes I feel like some properties are "guessed". The events tracking is one of our crucial aspects in making smart decisions. We have to make sure we track everything properly.
If you have some concerns or not sure about certain events please raise it rather than guessing.

Sidenote, can you confirm that the existing events will continue working?
image

packages/shared/src/hooks/useOnboardingModal.ts
sshanzel: The version one doesn't have the same condition to show a confirmation modal when closing/skipping the onboarding. So every "kind" of closing was intentional.

packages/shared/src/components/modals/OnboardingModal.tsx
sshanzel: The first one is the same as the previous implementation, the second one is since we already have a tracking event in our common custom hook but only designed for v1 due to their differences in the behavior of closing the modal.

sshanzel:
> Sidenote, can you confirm that the existing events will continue working?

Confirmed "they" still work as intended. Though since in v2, the way we select filter is different from v1, the only application events will be:

- click
- follow all
- unfollow all

packages/shared/src/components/filters/MyFeedHeading.tsx
sshanzel: I have missed this completely. I have now updated our `AnalyticsEvent` type to have the property `event_name` required as it was previously optional to avoid missing this one.

packages/shared/src/components/modals/OnboardingModal.tsx
rebelchris: @idoshamun This comment is related to we normally don't track flag values and rely on what get's added on the default tracking object.
But since we now decided to track specific values rather you can ignore this comment.

Just wanted you to note we do have different ways of using "target_id" at this point.
New ones are flags, old ones are actual positions (hardcoded)

rebelchris (changes requested):
Some questions for taxonomy.

idoshamun:
> Confirmed "they" still work as intended. Though since in v2, the way we select filter is different from v1, the only applicable events will be:

@sshanzel some of the events in the image I shared should be applicable to the "my feed settings" page where you can block, follow, unfollow single tags, etc

packages/shared/src/components/modals/OnboardingModal.tsx
```diff
@@ -16,37 +17,64 @@ import LayoutOnboarding from '../onboarding/LayoutOnboarding';
...
-  ) => {
+  const onCloseConfirm = (e: MouseEvent | KeyboardEvent) => {
+    if (
+      mode === OnboardingMode.Auto &&
+      onboardingVersion === OnboardingVersion.V2
+    ) {
+      const screen = getScreen(onboardingSteps, step);
+      trackEvent({
+        event_name: AnalyticsEvent.OnboardingSkip,
+        extra: JSON.stringify({ screen_value: screen }),
```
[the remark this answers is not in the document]
idoshamun: If it's already there we can use it actually :)

packages/shared/src/components/modals/OnboardingModal.tsx
idoshamun: Thanks for noting it down. As we proceed, I believe it will be more flags with some unique cases that will use placement ids. It really depends on what we want to test at the end of the day

idoshamun (approved):
I'm approving but I expect one minor change in the legacy onboarding:
Please change the target_id to 'legacy' instead of 'v3' which is set using `MY_FEED_VERSION_WINNER`.

Also the most important events are the impression of the my feed modal (in both versions) and complete onboarding (in both versions). The test will rely mostly on them. So we have to make sure it works

sshanzel:
> @sshanzel some of the events in the image I shared should be applicable to the "my feed settings" page where you can block, follow, unfollow single tags, etc

For "my feed settings" everything is still the same in terms of internal processing which includes the tracking as we only changed its layout. Also verified that things are working as intended.

> Also the most important events are the impression of the my feed modal (in both versions) and complete onboarding (in both versions). The test will rely mostly on them. So we have to make sure it works

I will do thorough end-to-end testing again before merging.

packages/shared/src/components/modals/OnboardingModal.tsx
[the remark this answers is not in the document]
sshanzel: I think it can add value to see where the users usually skip.

sshanzel:
Did an end-to-end testing and in the preview deployment it shows the complete onboarding being sent 🚢 

image

@rebelchris would love to hear your thoughts on the changes 🚢

rebelchris (approved):
Love it! ❤️

PR #1365, opened 2022-11-11
feat: notification nav footer item and page

sshanzel:
- Base UI for the notifications page
- Nav bar change to include the notifs items.

Note: the number indication is not yet available until such integration exist.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-696 #done


packages/webapp/pages/history.tsx
```diff
@@ -50,24 +51,12 @@ const History = (): ReactElement => {
...
-        {seo}
-        <ReadingHistoryEmptyScreen />
-      </>
-    );
-  }
-
   return (
-    <>
-      {seo}
+    <ProtectedPage
```
rebelchris, line 55: Just wondering if the layouts won't be a better option for this?

rebelchris (approved):
Nice work on this 🥳

rebelchris:
@sshanzel I'm not sure you want this to go into main though?
Should we not opt for a intermediate branch, else this gets added in the frontend for users.

sshanzel:
> @sshanzel I'm not sure you want this to go into main though?
Should we not opt for a intermediate branch, else this gets added in the frontend for users.

Ahh, my bad. Yep I created an intermediate branch earlier but I forgot to point this one to that.

packages/webapp/pages/history.tsx
sshanzel: Are you referring to the function `getLayout`? Or the NextJS layout?

rebelchris: Basically the Next layouts indeed, we could opt to create a ProcetedPage layout that could wrap around everything?
But open to this one as well so leaving it up to you.

sshanzel: Ah, interesting. Will look into that, might be something quick we can implement, at least we can start now.

el (approved):

packages/webapp/pages/history.tsx
sshanzel: We are already utilizing the `getLayout` property, the current implementation should be fine with its own layout as the `getLayout` will provide the nesting layout.

PR #1367, opened 2022-11-11
feat: header bell icon

sshanzel:
- The primary goal in this PR is to add the bell icon on the header.
- It can be noticed that there seem to be a lot of changes that happened in this PR. I decided to refactor the `MainLayout` to extract all the `Header` related concerns. Everything related to it can be found in this commit https://github.com/dailydotdev/apps/commit/2144bab87e8febdfd150b4b047774a85946e0785
- To make it easier to review the changes for applying the new bell icon, please refer to this specific commit https://github.com/dailydotdev/apps/commit/93f7d402ce7fcdd10dedea5c93a922c64848418f

Note: as mentioned in the primary PR, the number indication shall be worked on once the integration has been created.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-695 #done


packages/shared/src/components/layout/MainLayoutHeader.tsx
```diff
@@ -0,0 +1,114 @@
...
+            )}
+            {shouldShowLogo({ mobileTitle, sidebarRendered }) && (
+              <HeaderLogo
+                user={user}
+                onLogoClick={onLogoClick}
+                greeting={greeting}
+              />
+            )}
+          </div>
+          {!hideButton && user && (
```
rebelchris, line 91: When do we want this one hidden?

rebelchris (approved):
Wow thanks for abstracting so much here 🙏

packages/shared/src/components/layout/MainLayoutHeader.tsx
sshanzel: When things are still being loaded and when we only show the logo parameter is true.

packages/shared/src/components/layout/MainLayoutHeader.tsx
```diff
@@ -0,0 +1,114 @@
...
+            <LinkWithTooltip
+              tooltip={{ placement: 'left', content: 'Notifications' }}
+              href={`${webappUrl}notifications`}
```
el, line 94: what happens to this link if it is used as an extension?

el (approved):
Just had a small question, other than that LGTM. Thanks ❤️

packages/shared/src/components/layout/MainLayoutHeader.tsx
sshanzel: The extension will redirect you to our webapp. A similar scenario would be from the extension and then clicking the bookmarks tab.

PR #1368, opened 2022-11-14
chore: upgrade chunk names to company standard

rebelchris:
- Modified all existing imports to use the company defined chunk name standard

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-693 #done

### DIFF
--- a/packages/shared/src/components/auth/AuthModal.tsx
+++ b/packages/shared/src/components/auth/AuthModal.tsx
@@ -16,7 +16,10 @@ import AnalyticsContext from '../../contexts/AnalyticsContext';
 export type AuthModalProps = { trigger?: AuthTriggersOrString } & ModalProps;
 
 const DiscardActionModal = dynamic(
-  () => import('../modals/DiscardActionModal'),
+  () =>
+    import(
+      /* webpackChunkName: "discardActionModal" */ '../modals/DiscardActionModal'
+    ),
 );
 
 const containerMargin = {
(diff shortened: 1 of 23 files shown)


packages/shared/src/components/auth/AuthModal.tsx
```diff
@@ -16,7 +16,10 @@ import AnalyticsContext from '../../contexts/AnalyticsContext';
 export type AuthModalProps = { trigger?: AuthTriggersOrString } & ModalProps;
 
 const DiscardActionModal = dynamic(
-  () => import('../modals/DiscardActionModal'),
+  () =>
+    import(
+      /* webpackChunkName: "discardActionModal" */ '../modals/DiscardActionModal'
```
el, line 21: Why dont we update webpack configuration to use camel case, and we can use the webpackChunkName for custom things

el (changes requested):
I think we should update the webpack configuration to match what we want instead of changing the custom names everywhere

packages/shared/src/components/auth/AuthModal.tsx
sshanzel: Didn't know that was possible 🤯

rebelchris: Ah didn't know that was an option, but it might prove difficult to sync between SWC (next) and webpack on the extension side?
Not sure if you have any idea about that?

rebelchris: @el I had a read through but it seems you would still rely on manually naming them right?
See [this article](https://www.codemzy.com/blog/how-to-name-webpack-chunk), although would love to hear if you know another way?

el (approved):

PR #1371, opened 2022-11-14
feat: sidebar support for the new modal structure

el:
- Sidebar support for modals

- [x] MobileL (420px)
- [x] Tablet (656px)
- [x] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-732 #done

### DIFF
--- a/packages/shared/src/components/filters/FeedFilters.tsx
+++ b/packages/shared/src/components/filters/FeedFilters.tsx
@@ -15,6 +15,7 @@ import ArrowIcon from '../icons/Arrow';
 import { Button } from '../buttons/Button';
 import { UnblockItem } from './FilterMenu';
 import UnblockModal from '../modals/UnblockModal';
+// import { Modal } from '../modals/common/Modal';
 
 enum FilterMenuTitle {
   Tags = 'Manage tags',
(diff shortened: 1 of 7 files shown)


packages/shared/src/components/filters/FeedFilters.tsx
```diff
@@ -56,6 +57,60 @@ export default function FeedFilters({
...
+  //     {
+  //       title: FilterMenuTitle.Advanced,
+  //       options: { icon: <FilterIcon /> },
+  //     },
+  //     {
+  //       title: FilterMenuTitle.Blocked,
+  //       options: { icon: <BlockIcon /> },
+  //     },
+  //   ];
+  //   return (
```
el, line 74: This is the updated usage for the new structure

packages/shared/src/components/modals/common/ModalTabs.tsx
```diff
@@ -1,33 +1,36 @@
...
-          {tab}
-          {tab === activeTab && (
-            <div
-              className="absolute mx-auto w-4 h-px bg-theme-label-primary"
-              style={{ bottom: '-0.75rem', left: 'calc(50% - 0.5rem)' }}
-            />
-          )}
-        </button>
-      ))}
+      {tabs.map((tab: string | ModalTabItem) => {
```
el, line 9: sidebar tabs are basically tabs on the left :)

packages/shared/src/components/filters/FeedFilters.tsx
```diff
@@ -56,6 +57,60 @@ export default function FeedFilters({
...
+  //     <Modal
+  //       className="h-full mb-8 flex-1"
```
rebelchris, line 76: should the h-full not be part of the guidelined modal in this case? (basically all these classes?)

rebelchris (approved):
Really cool approach, just one question around the generic styling.

packages/shared/src/components/filters/FeedFilters.tsx
el: Feed filters has a full screen height modal which is not part of the modal guidelines. Thats why I added the classes outside. But maybe we can check with `@ design` to see if it should be part of the guidelines

PR #1375, opened 2022-11-15
feat: [modal migration] submit article

el:
@rebelchris  @sshanzel Can you please check this locally? I was not able to test all scenarios.


packages/shared/src/components/modals/SubmitArticleModal.tsx
```diff
@@ -182,66 +181,78 @@ export default function SubmitArticleModal({
...
-    <ResponsiveModal
-      {...modalProps}
-      onRequestClose={onRequestClose}
-      padding={false}
-      style={{ content: { maxWidth: '35rem', width: '100%' } }}
-    >
-      <header className="flex justify-between items-center px-6 w-full h-14 border-b border-theme-divider-tertiary">
-        <div className="flex flex-row items-center">
-          <h3 className="font-bold typo-title3">{headerCopy}</h3>
+    <form ref={submitFormRef} aria-busy={isValidating} onSubmit={onSubmit}>
```
rebelchris, line 184: I'm not quite sure I like the form wrapping around the whole modal.
Can we not move it into the body and use form="id" on the button rather?

packages/shared/src/components/modals/SubmitArticleModal.tsx
```diff
@@ -182,66 +181,78 @@ export default function SubmitArticleModal({
...
+      <Modal
+        {...modalProps}
+        kind={Modal.Kind.FlexibleCenter}
+        size={Modal.Size.Medium}
+        onRequestClose={onRequestClose}
+      >
+        <Modal.Header>
+          <Modal.Header.Title>{headerCopy}</Modal.Header.Title>
           <BetaBadge />
```
rebelchris, line 193: We can remove this, after discussion on Slack:
https://dailydotdev.slack.com/archives/C02AA18RD1D/p1668588454326879

rebelchris (changes requested):
I'm not fully able to test UX wise, but tested all variations locally.

The only issue seems to be with the existing article see attached:
Screenshot 2022-11-16 at 12 00 52

packages/shared/src/components/modals/SubmitArticleModal.tsx
el: Sure thing, but what is the downside?

packages/shared/src/components/modals/SubmitArticleModal.tsx
el: Done 👍

packages/shared/src/components/modals/SubmitArticleModal.tsx
el: Done 👍

rebelchris: No technical downside more of a structural point of view.
It felt like abusing the wrapping.

el:
@rebelchris @sshanzel here is how it looks now:

image

el:
Updated article position:
image

rebelchris (changes requested):
Would you mind checking with a article with a super short name?
It seems that proves a issue. (might have been existing)

el:
Short title fix

image

rebelchris (approved):
Amazing, nice work 🥳

PR #1376, opened 2022-11-15
fix: await features flags availability

sshanzel:
- Issue was onboarding is loaded very quickly and does not wait for the boot data to finish yet.
- Fix is to include a variable to check whether the data was fetched already.

Tagged @idoshamun for raising the issue.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done

### DIFF
--- a/packages/shared/src/contexts/FeaturesContext.tsx
+++ b/packages/shared/src/contexts/FeaturesContext.tsx
@@ -35,6 +35,7 @@ interface Experiments {
   squadVersion?: SquadVersion;
   squadForm?: string;
   squadButton?: string;
+  isFeaturesLoaded?: boolean;
 }
 
 export interface FeaturesData extends Experiments {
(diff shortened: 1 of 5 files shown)


packages/shared/src/contexts/FeaturesContext.tsx
```diff
@@ -93,19 +98,20 @@ const getFeatures = (flags: IFlags): FeaturesData => {
 };
 
 export const FeaturesContextProvider = ({
+  isFeaturesLoaded,
   children,
   flags,
 }: FeaturesContextProviderProps): ReactElement => {
   const featuresFlags: FeaturesData = useMemo(() => {
-    const features = getFeatures(flags);
+    const features = getFeatures(flags, isFeaturesLoaded);
```
idoshamun, line 106: I would recommend adding the `isFeaturesLoaded` prop after the `getFeatures` function.
I think we can keep the `getFeatures` implementation as it is. Makes more sense in my opinion

PR #1377, opened 2022-11-16
feat: [modal migration] upvoted by

el:
## Changes

Also fixes issue with the close button size:

image


packages/shared/src/components/RecommendedMention.tsx
```diff
@@ -35,7 +35,7 @@ export function RecommendedMention({
           key={user.username}
           user={user}
           className={classNames(
-            'px-3 cursor-pointer',
+            'py-3 px-3 cursor-pointer',
```
rebelchris, line 38:
```suggestion
            'p3 cursor-pointer',
```

rebelchris (changes requested):
Some more generic to this one feedback:

- Pressing outside the modal doesn't close it for some reason
- When the list has the preloaders it actually shifts the layout (height of the modal)

el:
@rebelchris @sshanzel here is how it looks now:

image

packages/shared/src/components/RecommendedMention.tsx
el: Upps, it should be `p-3`

rebelchris (commented):
It almost looks too big now compared to the one we have?
Also can we prioritised the backdrop close before pushing this live? (I don't want this to go live as it has high impact on users)

PR #1378, opened 2022-11-16
feat: [modal migration] share post & comment

el:
## Changes

I used size=small for now but we can change it to medium if the design wants

image


rebelchris (changes requested):
Generic:
- Doesn't close on backdrop click

Specific:
Comment only has top spacing, not sure that's correct (might have been an issue in current one to be honest)
Screenshot 2022-11-16 at 12 26 35

Cut off text:
Screenshot 2022-11-16 at 12 25 07

el:
image

Close on backdrop is also fixed

rebelchris (approved):
Perfect, happy with this one 🥳

PR #1380, opened 2022-11-17
feat: [modal migration] upvoted by

el:
## Changes

Reopened PR for upvoted by

### DIFF
--- a/packages/shared/src/components/CloseButton.tsx
+++ b/packages/shared/src/components/CloseButton.tsx
@@ -1,10 +1,6 @@
 import classNames from 'classnames';
 import React, { forwardRef, ReactElement, Ref } from 'react';
-import {
-  Button,
-  StyledButtonProps,
-  ButtonProps,
-} from './buttons/Button';
+import { Button, StyledButtonProps, ButtonProps } from './buttons/Button';
 import CloseIcon from './icons/Close';
 
 type CloseButtonProps = StyledButtonProps & ButtonProps<'button'>;
--- a/packages/shared/src/components/CloseButton.tsx
+++ b/packages/shared/src/components/CloseButton.tsx
@@ -4,23 +4,19 @@ import {
   Button,
   StyledButtonProps,
   ButtonProps,
-  ButtonSize,
 } from './buttons/Button';
 import CloseIcon from './icons/Close';
 
-type CloseButtonProps = StyledButtonProps &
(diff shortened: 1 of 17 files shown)


rebelchris (changes requested):
Specific issues with this modal:

- Size seems to wide at the moment
- Backdrop (as handled in other task, but please merge that one before or pull that change in?)
- The user items have padding
- The user items don't show the user tooltip on hover

Old:
Screenshot 2022-11-17 at 10 03 22

New:
Screenshot 2022-11-17 at 10 03 29

packages/shared/src/components/CloseButton.tsx
```diff
@@ -1,15 +1,25 @@
...
+  Button,
+  StyledButtonProps,
+  ButtonProps,
+  ButtonSize,
+} from './buttons/Button';
 import CloseIcon from './icons/Close';
 
+type CloseButtonProps = StyledButtonProps &
+  ButtonProps<'button'> & {
+    size?: ButtonSize;
```
sshanzel, line 13: Should we not keep using the `buttonSize` as it is already available from the `ButtonProps`

packages/shared/src/components/modals/common/ModalClose.tsx
```diff
@@ -13,7 +13,8 @@ function ModalCloseComponent(
       {...props}
       onClick={onClick}
       ref={ref}
-      className={classNames('right-4 z-1', className)}
+      size="medium"
```
sshanzel, line 16: I think we can just use the `buttonSize` instead of introducing a new one.

sshanzel (commented):
Code-wise I only had one comment (around button size property), other than that. There is nothing more I can add. Chris has already mentioned the concerns. 

Thank you for taking this on. Implementations of modal looks so clean 🚢

packages/shared/src/components/CloseButton.tsx
el: Great catch. Fixed 👍

packages/shared/src/components/modals/common/ModalClose.tsx
el: Yeap, sharp eyes 👀

rebelchris (approved):
Tested and looks to work perfectly now 🥳

PR #1381, opened 2022-11-17
fix: onboarding trap

sshanzel:
- One issue could be a race condition where we were not awaiting the variable we use when updating them, then the useEffect gets triggered to open the modal again. This would ensure setting the value has happened first before closing them modal.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done

### DIFF
--- a/packages/shared/src/components/MainFeedLayout.tsx
+++ b/packages/shared/src/components/MainFeedLayout.tsx
@@ -183,7 +183,7 @@ export default function MainFeedLayout({
     hasFiltered: !alerts?.filter,
     hasUser: !!user,
   });
-  const { flags, popularFeedCopy, onboardingVersion, isFeaturesLoaded } =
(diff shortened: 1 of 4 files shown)


rebelchris (approved):
Wouldn't this delay the closing of the modal as well?
We can give it a try though.

sshanzel:
> We can give it a try though.

I noticed a slight delay compared to the subsequent closes of modal. WDYT?

sshanzel:
@rebelchris even waiting for local cache fails probably due to some inaccuracies with the update of the flags. Local flags may have a value of v1 but then now we released the `v2`, then waiting for local values is not correct.

We should instead wait for when the first boot data is fetched.

rebelchris (approved):
Alright seems good to me.

PR #1384, opened 2022-11-18
feat: [modal migration] suggest new source

el:
## Changes
Screenshot 2022-11-18 at 16 43 45
Screenshot 2022-11-18 at 16 43 54

WT-753 #done


packages/shared/src/components/Loader.tsx
```diff
@@ -12,7 +12,15 @@ export function Loader({
...
 }: LoaderProps): ReactElement {
   return (
-    <div {...props} className={classNames('w-5 h-5', className, styles.loader)}>
+    <div
+      {...props}
+      className={classNames(
+        'w-5 h-5',
+        className,
+        styles.loader,
+        invertColor && 'invert',
```
el, line 21: invert wasnt working, so I fixed that

packages/shared/src/components/modals/SubmitArticleModal.tsx
```diff
@@ -182,7 +182,7 @@ export default function SubmitArticleModal({
   return (
     <Modal
       {...modalProps}
-      kind={Modal.Kind.FlexibleCenter}
```
rebelchris, line 185: Why did this one change?

rebelchris (changes requested):
One remark for the other modal.

And for this one, it looks a bit weird with this placeholder.
Screenshot 2022-11-18 at 16 09 23

Maybe double check with design what they would like this to do?

packages/shared/src/components/modals/SubmitArticleModal.tsx
rebelchris: Pretty sure they should both be FlexibleCenter right?

el: Their height changes based on the content so it causes them to shift, that's why I changed them both. in the design one of they were stated as flexible top:

https://www.sketch.com/s/2867ec63-0966-4647-b1d7-b6fe696ba251/a/dlpl9AZ

rebelchris (approved):
Looks solid now and double checked with design on the flexibleTop which is right

Issue #1386, opened 2022-11-21
tailwind concatenating classes


sshanzel:
While reviewing [this PR](https://github.com/dailydotdev/apps/pull/1385/files) I have noticed an issue in the `ModalFooter` component, more of a common pitfall with using tailwindcss.

Concatenation of string to produce a specific class is not encouraged. I can't find the particular doc page from them but the concern with using this process is the classes will get tree-shaken.  Hence, it might not work in the production build.

https://github.com/dailydotdev/apps/blob/8d6ee1d856d289e82cb9082c0a6deefec623c6ee/packages/shared/src/components/modals/common/ModalFooter.tsx#L24

cc @rebelchris @el 

It was way back. Not sure if that has changed now though.

rebelchris:
Oh good find Lee, this is indeed a concern as we should only opt to use the fully qualified names, totally missed this one.

sshanzel:
After digging through the tailwindcss docs, I finally found the relevant page. It seems to be still the case as discussed here:

https://tailwindcss.com/docs/content-configuration#dynamic-class-names

rebelchris:
I'll add it to our board and pick it up in Eliz's absence.

sshanzel:
Oh, totally forgot about it. Sorry for tagging! 🙏

sshanzel:
We should probably include the link as well in our onboarding guide around getting to know more of our code base as it is really to fall into this trap if you haven't used tailwindcss in the past.

el:
@sshanzel Amazing find, I wasn't aware of this recommendation but it makes total sense.

PR #1388, opened 2022-11-21
fix: added fully qualified type for justify modal footers

rebelchris:
- Added Justify enum for fully qualified classnames and ease of import for the modal footer

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-771 #done #1386

### DIFF
--- a/packages/shared/src/components/utilities.tsx
+++ b/packages/shared/src/components/utilities.tsx
@@ -16,6 +16,12 @@ export enum Theme {
   Lettuce = 'lettuce',
 }
 
+export enum Justify {
+  End = 'justify-end',
+  Center = 'justify-center',
+  Between = 'justify-between',
+}
+
 export interface ThemeColor {
   border: string;
   shadow: string;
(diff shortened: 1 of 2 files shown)


packages/shared/src/components/utilities.tsx
```diff
@@ -16,6 +16,12 @@ export enum Theme {
   Lettuce = 'lettuce',
 }
 
+export enum Justify {
+  End = 'justify-end',
+  Center = 'justify-center',
+  Between = 'justify-between',
+}
```
sshanzel, line 23: Love this enum. We should probably do this as well on some other CSS properties that we tend to clash with sometimes. Probably when we face it on the road.

sshanzel (approved):
Amazing cleanup you did there! Love the implementation. Hoping to see more of this in the future!

packages/shared/src/components/utilities.tsx
rebelchris: Yep would be a great idea going forward, enums just make things more explicit and set in stone 🥳

el: I love this. We should use it on more utility classes (if they are passed as props)

el:
@rebelchris Thanks for taking care of this. ❤️

PR #1391, opened 2022-11-24
fix: focus mode modal

rebelchris:
- refactor DnD modal to new guidelines
Screenshot 2022-11-24 at 15 29 37
Screenshot 2022-11-24 at 15 29 41

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-750 #done


sshanzel (approved):
It is so satisfying to see inline styles getting removed! ✨

packages/extension/src/newtab/DndModal.tsx
```diff
@@ -113,36 +80,36 @@ export default function DndModal({
...
-          className={classnames(
-            'font-bold text-theme-label-primary',
-            titleSize,
-          )}
-        >
-          Pause new tab
-        </h3>
+      <Modal.Header title="Pause new tab" />
+      <Modal.Body>
         <p className="mt-1 typo-callout text-theme-label-secondary">
```
el, line 92: is mt-1 necessary? Modal.Body already has a padding

packages/extension/src/newtab/DndModal.tsx
```diff
@@ -113,36 +80,36 @@ export default function DndModal({
...
-  const titleSize = isActive ? 'typo-title3' : 'typo-title2';
-
   return (
-    <StyledModal
+    <Modal
+      kind={Modal.Kind.FlexibleTop}
+      size={Modal.Size.Medium}
       {...modalProps}
       onRequestClose={onRequestClose}
-      style={{
```
el, line 122: So good to see these go 😍

packages/extension/src/newtab/DndModal.tsx
```diff
@@ -113,36 +80,36 @@ export default function DndModal({
...
-        </h3>
+      <Modal.Header title="Pause new tab" />
+      <Modal.Body>
         <p className="mt-1 typo-callout text-theme-label-secondary">
           {getDescription()}
         </p>
-      </div>
-      {isActive ? turnedOnContent : turnedOffContent}
-    </StyledModal>
+        {!isActive ? turnedOffContent : null}
```
el, line 95: I think we should move the turnedoffcontent here as well, or convert it into a component.

el:
Great work, and thanks for taking care of this ❤️

packages/extension/src/newtab/DndModal.tsx
rebelchris: We could simply show it here, let me refactor :)

packages/extension/src/newtab/DndModal.tsx
```diff
@@ -113,36 +46,70 @@ export default function DndModal({
...
-        <p className="mt-1 typo-callout text-theme-label-secondary">
+      <Modal.Header title="Pause new tab" />
+      <Modal.Body>
+        <p className="typo-callout text-theme-label-secondary">
           {getDescription()}
         </p>
-      </div>
-      {isActive ? turnedOnContent : turnedOffContent}
-    </StyledModal>
+        {!isActive ? (
```
el, line 61:
```suggestion
        {!isActive && (
```

packages/extension/src/newtab/DndModal.tsx
```diff
@@ -113,36 +46,70 @@ export default function DndModal({
...
+                <Dropdown
+                  className={{ container: 'w-40' }}
+                  options={customTimeOptions}
+                  selectedIndex={customTimeIndex}
+                  onChange={(_, index) => setCustomTimeIndex(index)}
+                />
+              </div>
+            )}
+          </div>
+        ) : null}
```
el, line 95:
```suggestion
        )}
```

el (changes requested):
Requested a simple fix, otherwise perfect

packages/extension/src/newtab/DndModal.tsx
rebelchris: I recently found myself avoiding these for best practice as they can introduce null of false renders.
We can make a global web decision about it though.

As for reference:
https://twitter.com/TkDodo/status/1592276656261140480?s=20&t=pmNq6lGDpYM396S-amgkDw

el: You mean the case for numbers?

something like this?
`items.length && <Component />`

For that I always used (and we enforced in my prev company) to cast to boolean first.

`!!items.length && <Component />`

rebelchris: Hmm yes that the main concern, but apparently it can also happen with null values (useState) 
It's the safer way to be explicit, not sure how you feel about that?

I changed it to `&&` for now, but perhaps something to consider in the future, as explicit null is always safer.

el (approved):
🚢  🎊

PR #1397, opened 2022-11-29
feat: fix icon issues provider by design

rebelchris:
- Fixed the icon issues design provider us with.
- @sshanzel special request to look at the customSwitch change, not sure if there's a better alternative for now.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-789 #done


packages/shared/src/components/fields/CustomSwitch.tsx
```diff
@@ -9,8 +9,8 @@ export interface ContentsSwitchProps {
   name: string;
   checked?: boolean;
   onToggle?: () => unknown;
-  leftContent: React.FC<{ className?: string }> | string;
-  rightContent: React.FC<{ className?: string }> | string;
+  leftContent: React.FC<{ className?: string; secondary?: boolean }> | string;
```
sshanzel, line 12: I think a better property to pass would be `isChecked` since the content can either be something else entirely and not an icon but might still require the value if it is checked or not.

sshanzel (commented):
Just had one comment about a more accurate name for the prop. Thanks for fixing outdated icons!

packages/shared/src/components/fields/CustomSwitch.tsx
rebelchris: Not quite sure I follow, the prop for the icons is actually `secondary` how would we transpile it to that?

rebelchris: @sshanzel I changed to ReactNode this way we can keep native implementation on the input side.
Let me know what you think please.

packages/shared/src/components/sidebar/SidebarListItem.tsx
```diff
@@ -28,7 +28,7 @@ function SidebarListItem({
   );
   const content = (
     <>
-      {icon}
+      {React.cloneElement(icon as ReactElement, { secondary: isActive })}
```
el, line 31: Why did you clone it here?

rebelchris: To add the props to the render, basically because we have neater accessibility to the isActive.
Would you say it's neater to add it higher up but manually check with strings?

packages/shared/src/components/fields/CustomSwitch.tsx
sshanzel: That works too. This was a legacy implementation, I'm more than happy to get the FC removed. Thanks!

packages/shared/src/components/fields/CustomSwitch.tsx
```diff
@@ -79,21 +85,21 @@ export function CustomSwitch({
...
       >
-        {typeof LeftContent === 'string' ? (
-          LeftContent
-        ) : (
-          <LeftContent className={classNames(leftClasses)} />
-        )}
+        {typeof leftContent === 'string'
+          ? leftContent
+          : React.cloneElement(leftContent as ReactElement, {
+              className: leftClasses,
```
sshanzel, line 91: Just one question, would this not override the current node's `className`?

sshanzel (approved):
LGTM! 🚀 

Just had a question but nothing blocking! 🚢

packages/shared/src/components/fields/CustomSwitch.tsx
rebelchris: Yeah this will set it always to this value, but so would the original one right?
We don't pass any existing classNames, we could extend with `classNames` to take the original if you think that's needed?

sshanzel: The original one though passes it on a function component, so the consuming end will be able to combine the passed className by the CustomSwitch.

sshanzel: The original one though passes it on a function component, so the consuming end will be able to combine the passed className by the CustomSwitch.

rebelchris: Hmm, we can maybe deal with it as we get a component that takes className?
Don't think there is one, but let me try and add classNames think that will do the trick 🙏

sshanzel: I'll leave it up to you, I am cool on taking it on as we go as we don't have anything that requires it now or any upcoming component.

packages/shared/src/components/sidebar/SidebarListItem.tsx
el: Ahh, this is neater ❤️

el (approved):

PR #1398, opened 2022-11-29
feat: [modal migration] comments modal

el:
- Comments modal

- [x] MobileL (420px)
- [x] Tablet (656px)
- [x] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-747 #done

### DIFF
--- a/packages/shared/src/components/modals/CommentBox.tsx
+++ b/packages/shared/src/components/modals/CommentBox.tsx
@@ -10,10 +10,8 @@ import classNames from 'classnames';
 import AuthContext from '../../contexts/AuthContext';
 import { commentBoxClassNames } from '../comments/common';
 import { commentDateFormat } from '../../lib/dateFormat';
-import { Button } from '../buttons/Button';
 import styles from './CommentBox.module.css';
 import { ProfilePicture } from '../ProfilePicture';
-import { ClickableText } from '../buttons/ClickableText';
 import Markdown from '../Markdown';
 import { RecommendedMentionTooltip } from '../tooltips/RecommendedMentionTooltip';
 import {
@@ -22,9 +20,7 @@ import {
   useUserMention,
 } from '../../hooks/useUserMention';
 import { Post } from '../../graphql/posts';
-import AtIcon from '../icons/At';
 import { cleanupEmptySpaces } from '../../lib/strings';
-import { markdownGuide } from '../../lib/constants';
 
 export interface CommentBoxProps {
   authorName: string;
@@ -32,10 +28,8 @@ export interface CommentBoxProps {
   publishDate: Date | string;
   contentHtml: string;
   editContent?: string;
-  editId?: string;
   input?: string;
   errorMessage?: string;
-  sendingComment?: boolean;
   parentSelector?: () => HTMLElement;
   sendComment: (event: MouseEvent | KeyboardEvent) => Promise<void>;
   onInput?: (value: string) => unknown;
@@ -48,10 +42,8 @@ function CommentBox({
   publishDate,
   contentHtml,
   editContent,
-  editId,
   input,
   errorMessage,
-  sendingComment,
   onInput,
   sendComment,
   parentSelector,
@@ -66,7 +58,6 @@ function CommentBox({
     mentionQuery,
     selected,
     commentRef,
-    onInitializeMention,
     onInputClick,
   } = useUserMention({
     postId: post.id,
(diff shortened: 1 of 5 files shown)


packages/shared/src/components/modals/NewCommentModal.tsx
```diff
@@ -144,114 +149,79 @@ export default function NewCommentModal({
       setSendingComment(false);
     }
   };
+  const { onInitializeMention } = useUserMention({
```
el, line 152: @rebelchris @sshanzel This does not seem to work and I had a brain freeze 🥶  trying to make it work. Can you help me fix it 🙏

el: Thanks for the suggestion @rebelchris

packages/shared/src/components/modals/CommentBox.tsx
```diff
@@ -10,29 +10,24 @@ import classNames from 'classnames';
...
 import styles from './CommentBox.module.css';
 import { ProfilePicture } from '../ProfilePicture';
-import { ClickableText } from '../buttons/ClickableText';
 import Markdown from '../Markdown';
 import { RecommendedMentionTooltip } from '../tooltips/RecommendedMentionTooltip';
 import {
   fixHeight,
   UPDOWN_ARROW_KEYS,
-  useUserMention,
+  UseUserMention,
```
rebelchris, line 20: Can we maybe make the props separate as `UseUserMentionProps`

packages/shared/src/hooks/useUserMention.ts
```diff
@@ -213,7 +213,7 @@ export function useUserMention({
   };
 
   const onInitializeMentionButtonClick = () => {
-    if (typeof query !== 'undefined') {
+    if (typeof query !== 'undefined' || !commentRef.current) {
```
rebelchris, line 216: I think we should still have this in case the ref for comments is not available right?
What happens when you simply type `@T` (without the button)

packages/shared/src/hooks/useUserMention.ts
```diff
@@ -246,10 +246,10 @@ export function useUserMention({
     };
 
     const dom = commentRef.current?.getRootNode();
-    dom.addEventListener('mousedown', userClicked);
+    dom?.addEventListener('mousedown', userClicked);
```
rebelchris, line 249: When would this be empty?

rebelchris (changes requested):
Looks great just some minor questions before we can push it live :)

packages/shared/src/hooks/useUserMention.ts
rebelchris: I tested this and it seems to work fine :)

packages/shared/src/hooks/useUserMention.ts
el: Ok it is because of `strict:false` in tsconfig

packages/shared/src/components/modals/CommentBox.tsx
el: done

packages/shared/src/hooks/useUserMention.ts
rebelchris: Oh interesting that it didn't fail before, I wonder why 🤯

packages/shared/src/hooks/useUserMention.ts
```diff
@@ -31,7 +31,7 @@ import {
 import { nextTick } from '../lib/func';
 import { useRequestProtocol } from './useRequestProtocol';
 
-interface UseUserMention {
+export interface UseUserMention {
```
rebelchris, line 34: I'm sorry I meant can't we just export this as `UseUserMentionProps` it's a bit vague now with two times the same export `useUserMention` and `UseUserMention`

rebelchris (changes requested):

packages/shared/src/hooks/useUserMention.ts
el: we already have a type as `UseUserMentionProps` it is the input types `useUserMention` function takes. this is the output.

```

function useUserMention(props: UseUserMentionProps): UseUserMention { ... }

```

el: Renamed to `UseUserMentionOptions`

PR #1399, opened 2022-11-30
feat: [modal migration] weekly reading goal

el:
### Design:
https://www.sketch.com/s/2867ec63-0966-4647-b1d7-b6fe696ba251/a/ZOmO1Z7

### Previous: 
image

### New:
image

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-756 #done

### DIFF
--- a/packages/shared/src/components/modals/common/Modal.tsx
+++ b/packages/shared/src/components/modals/common/Modal.tsx
@@ -105,7 +105,6 @@ export function Modal({
 
 export const ModalSubtitle = classed('strong', 'typo-headline mb-2');
 export const ModalText = classed('p', 'typo-callout text-theme-label-tertiary');
-// export const ModalSection = classed('section', 'flex flex-col p-6');
 
 Modal.Size = ModalSize;
 Modal.Kind = ModalKind;
--- a/packages/shared/src/components/modals/common/Modal.tsx
+++ b/packages/shared/src/components/modals/common/Modal.tsx
@@ -12,6 +12,7 @@ import {
   ModalTabItem,
   modalTabTitle,
 } from './types';
+import classed from '../../../lib/classed';
 
 export type ModalProps = ReactModal.Props & {
   children: React.ReactNode;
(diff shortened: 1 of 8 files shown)


packages/shared/src/components/modals/common/Modal.tsx
```diff
@@ -102,9 +103,15 @@ export function Modal({
   );
 }
 
+export const ModalSubtitle = classed('strong', 'typo-headline mb-2');
+export const ModalText = classed('p', 'typo-callout text-theme-label-tertiary');
+// export const ModalSection = classed('section', 'flex flex-col p-6');
```
rebelchris, line 108: This can be removed?

packages/shared/src/components/modals/common/Modal.tsx
```diff
@@ -102,9 +103,15 @@ export function Modal({
   );
 }
 
+export const ModalSubtitle = classed('strong', 'typo-headline mb-2');
+export const ModalText = classed('p', 'typo-callout text-theme-label-tertiary');
```
rebelchris, line 107: Are these generic for each modal, else it's maybe best to keep it in the specific modal?
Just not sure if it's recurring or not else we'll have to apply it to other modals as well that use this format.

rebelchris (approved):
Happy with this, just one question around the generic classes used.

Another request:
Could you locally force test the last rank to see the offset still works?

packages/shared/src/components/modals/common/Modal.tsx
el: Upps, removed

packages/shared/src/components/modals/common/Modal.tsx
el: I have seen at least 1 more usage, actually let me remove migrate those as well

el: I think they are being used in other places, I can try to migrate those as well after this PR

el:
@rebelchris last rank:
image

packages/shared/src/components/modals/common/Modal.tsx
rebelchris: Sure it can be separate :)
Just nice to have everything from one source instead of re-inventing the wheel.

rebelchris (approved):

PR #1405, opened 2022-12-02
feat: feed filter anonymous

rebelchris:
- Added post card feed filter versions

## Visual tests

|  | Modal | Post |
|-------|-------|-------|
| Off | Screenshot 2022-12-07 at 16 06 37 | Screenshot 2022-12-07 at 15 23 47 |
| V1 | Screenshot 2022-12-07 at 15 20 33 | Screenshot 2022-12-07 at 15 20 52 |
| V2 | Screenshot 2022-12-07 at 15 21 23 | Screenshot 2022-12-07 at 15 21 38 |
| V3 | Screenshot 2022-12-07 at 15 22 01 | Screenshot 2022-12-07 at 15 22 17 |

Responsive tested each version same outcome:
Screenshot 2022-12-07 at 15 23 52

## Events

| New | impression  | article anonymous cta  |
| New | click article anonymous cta  | extra: { origin } |

- [x] MobileL (420px)
- [x] Tablet (656px)
- [x] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-793 #done


rebelchris:
Generic note: Need to test the extensions feed redirections still 🥲 (This is the only potential blocker)

el (approved):
Great work, just had a small comment

packages/shared/src/components/modals/ResponsiveModal.module.css
```diff
@@ -2,9 +2,7 @@
   & :global(.overlay) {
     position: relative;
     min-height: 100vh;
-    min-height: -moz-available;
-    min-height: -webkit-fill-available;
-    min-height: fill-available;
+    min-height: stretch;
```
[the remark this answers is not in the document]
rebelchris: You are right!

Removed it now 🥳

rebelchris:
@el I switched to the new modals as well, this introduces a bit of a weird style, as the post modal actually styles the content, not the modal itself.
Didn't want to redo the whole scroll effect so left it as such and for this modal turned the border for the modal off.

el (changes requested):
Some small issues:

There seems to be an extra top padding, and an alignment issue:
image

Not sure if it is a/b related but I don't see the right side buttons:
image

image

rebelchris:
@el Nice find, I messed up the rendering there, fixed it now :)

el (approved):
Great work :shipit:

sshanzel:
I have deployed this in the `preview2` to easily how it looks like in actual data. Will provide review afterwards.

sshanzel:
Currently reviewing the UI. I think the bottom margin is now missing.
Preview2:
image


image

packages/shared/src/components/post/PostContent.tsx
```diff
@@ -204,6 +217,10 @@ export function PostContent({
   const hasNavigation = !!onPreviousPost || !!onNextPost;
   const Wrapper = hasNavigation ? BodyContainer : PageBodyContainer;
 
+  const { onInitializeOnboarding } = useOnboardingModal({
+    onFeedPageChanged: () => {},
```
sshanzel, line 221: Sorry for the additional comment. Just tagging this to be optional.

packages/shared/src/components/post/PostContent.tsx
```diff
@@ -253,119 +270,170 @@ export function PostContent({
...
-            imgSrc={postById.post.image}
-            imgAlt="Post cover image"
-            ratio="49%"
-            eager
-            fallbackSrc="https://res.cloudinary.com/daily-now/image/upload/f_auto/v1/placeholders/1"
-          />
-        </a>
-        {postById.post?.toc?.length > 0 && (
-          <PostToc
+        <>
```
sshanzel, line 322: I think we can remove the fragment here.

packages/shared/src/components/post/PostNavigation.tsx
```diff
@@ -54,49 +38,58 @@ export function PostNavigation({
...
-          />
-        </SimpleTooltip>
-      )}
-      {onNextPost && (
-        <SimpleTooltip content="Next">
-          <Button
-            className="rotate-90 btn-secondary"
-            icon={<ArrowIcon />}
-            onClick={onNextPost}
+    <>
```
sshanzel, line 41: I think we can remove the Fragment here.

packages/shared/src/components/post/PostPreviousNext.tsx
```diff
@@ -0,0 +1,42 @@
...
+const SimpleTooltip = dynamic(
+  () =>
+    import(/* webpackChunkName: "simpleTooltip" */ '../tooltips/SimpleTooltip'),
+);
+
+type PostPreviousNextProps = Pick<
+  PostNavigationProps,
+  'onPreviousPost' | 'onNextPost'
+>;
+export function PostPreviousNext({
```
sshanzel, line 16: Nice. Thank you for extracting this for readability 🙇

packages/shared/src/hooks/useOnboardingModal.ts
```diff
@@ -83,7 +82,7 @@ export const useOnboardingModal = ({
     registerLocalFilters().then(() => {
       trackEvent({ event_name: AnalyticsEvent.CompleteOnboarding });
       setShouldUpdateFilters(false);
-      modalStateCommand[onboardingVersion]?.(false);
+      setIsOnboardingOpen(false);
```
sshanzel, line 85: Now that this hook will be instantiated in multiple areas, all of them will be monitoring the `shouldUpdateFilters` and the `registerLocalFilters` will be triggered on all those channels as well. We might need to add a props (like `trackLocalFilters`) in this hook that we need to explicitly pass `true` so when it is `false`, this `useEffect` will return early. Else we will take this out of here and add it in a single-instantiated component (MainLayout or somewhere else).

sshanzel (changes requested):
Thank you so much for doing the cleanups along the way, especially extracting existing codes into components 🙏 

I just had a few comments, but I can't imagine the work you've gone through will all their differences. Just amazing 💪

packages/shared/src/components/MainLayout.tsx
```diff
@@ -129,6 +137,11 @@ export default function MainLayout({
 
   const hasBanner = !!bannerData?.banner || !!customBanner;
 
+  const { isOnboardingOpen, onCloseOnboardingModal, onShouldUpdateFilters } =
+    useOnboardingModal({
+      onFeedPageChanged: () => {},
```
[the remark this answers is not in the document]
rebelchris: Good point, can do that :)

packages/shared/src/hooks/useOnboardingModal.ts
rebelchris: That's a good point, let me check, it should be possible based on the function as it can only be called from one of them.
Another option is to set a query state to state wether it's done?

packages/shared/src/components/modals/PostModal.module.css
```diff
@@ -42,8 +42,6 @@
       max-width: 63.75rem;
       border-radius: 1rem;
       border: 1px solid var(--theme-divider-secondary);
-      margin-top: 3.5rem;
-      margin-bottom: 3.5rem;
```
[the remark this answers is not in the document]
rebelchris: Think it might be better to revert the bottom one here as it will be replaced in the end.

packages/shared/src/components/post/PostContent.tsx
rebelchris: Ah technically yes, but the conditionalWrapper expects valid ReactElements which can only be one parent.
This one includes the PostWidgets as a separate element, thus being wrapped in a fragment.

Hope that makes sense?

packages/shared/src/components/post/PostNavigation.tsx
rebelchris: Same comment here, it includes two elements on a root level, thus needs to be wrapped to be a valid ReactElement.

packages/shared/src/hooks/useOnboardingModal.ts
rebelchris: @sshanzel added a state for it, I think that should already catch it.
Let me know what you think.

packages/shared/src/components/post/PostContent.tsx
sshanzel: Is it not possible to make the `children` a `ReactNode` as we commonly use?

If not, then, in that case, this is really necessary. Thanks for letting me know! 🚢

sshanzel (changes requested):
Amazing idea on how to fix the trigger for registering the filters.

Just the content shift on scroll, then I think this should be good to go 🚢 🚢 🚢

idoshamun:
Issue: v2 is shown only in the article page not in the modal

sshanzel:
> Issue: v2 is shown only in the article page not in the modal

I think the issue could be from the `setFeatureFlags` I experienced earlier the same and that I had to use `window.clearFeatures()` before setting it again to work.
image

idoshamun:
> > Issue: v2 is shown only in the article page not in the modal
> 
> @idoshamun I think the issue could be from the `setFeatureFlags` I experienced earlier the same and that I had to use `window.clearFeatures()` before setting it again to work.
> 
> Not sure why I had to set it multiple times sometimes to work, and refreshes as well. Probably needs an investigation ticket. cc @rebelchris
> 

@sshanzel you're right! Thanks :)

packages/shared/src/components/post/PostContent.tsx
rebelchris: This will give the error that ReactNode is not a valid JSX Element.
Think this is quite what we would expect when passing multi-high level conditions right.

Let's leave this for now, we can always modify later?

packages/shared/src/hooks/useOnboardingModal.ts
rebelchris: Idea now is to actually split this hook into two
1 will inject the listeners and one the modifiers that way we can ensure the listener is only injected once.

In future mods we should probably move to context, but it's out of scope for now.

sshanzel: That's a great idea to isolate what's happening.

Yep, context is always like the best way to go, just feels like overkill sometimes. *redux is waving* lol.

packages/shared/src/components/post/PostContent.tsx
sshanzel: I agree. Let's keep this as the current implementation 🚢

packages/shared/src/contexts/FeaturesContext.tsx
```diff
@@ -91,6 +93,10 @@ const getFeatures = (flags: IFlags): FeaturesData => {
     squadVersion: getFeatureValue(Features.SquadVersion, flags),
     squadForm: getFeatureValue(Features.SquadForm, flags),
     squadButton: getFeatureValue(Features.SquadButton, flags),
+    articleOnboardingVersion: getFeatureValue(
```
sshanzel, line 96: Love the shorter name. Thanks for doing this! 🚀

packages/shared/src/contexts/OnboardingContext.tsx
```diff
@@ -0,0 +1,120 @@
...
+      onCloseOnboardingModal,
+      onShouldUpdateFilters: setShouldUpdateFilters,
+      onInitializeOnboarding: (onFeedPageChangedFn) => {
+        if (onFeedPageChangedFn) {
+          onFeedPageChanged.current = onFeedPageChangedFn;
+        }
+        setIsOnboarding(true);
+      },
+    }),
+    [isOnboarding],
```
sshanzel, line 110: Just the missing deps.
```suggestion
    [isOnboarding, user, alerts, onboardingMode, shouldUpdateFilters],
```

sshanzel (approved):
Thank you so much for making it easier to maintain. I agree that context is the best way to go.

Just had non-blocking comments, and leaving the decision on those areas up to you as those are not major.

Kindly ensure the content shift is also fixed before merging. Thank you for the patience 🚢 🚢 🚢

packages/shared/src/contexts/OnboardingContext.tsx
rebelchris: Oh dang thanks! 
Totally forgot I had it in mind and then forgot to execute so thank you 🙏

packages/shared/src/components/MainLayout.tsx
```diff
@@ -48,6 +49,13 @@ const Greeting = dynamic(
   () => import(/* webpackChunkName: "greeting" */ './Greeting'),
 );
 
+const OnboardingModal = dynamic(
+  () =>
+    import(
+      /* webpackChunkName: "onboardingModal" */ './modals/OnboardingModal'
+    ),
+);
+
```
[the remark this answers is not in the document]
rebelchris: Interesting one!
I moved it now, definitely something we could do even for auth modal in the future? 👀

Let me know what you think.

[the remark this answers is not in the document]
sshanzel: Oh, that is true. I just realized that. We should definitely do that.

sshanzel (approved):
Amazing! ✨ ✨ ✨

sshanzel (approved):
LGTM! 🚢 🚢 🚢

packages/shared/src/components/modals/common/Modal.tsx
```diff
@@ -48,8 +48,7 @@ const modalKindAndSizeToClassName: Partial<
   [ModalKind.FlexibleTop]: {
     [ModalSize.Medium]:
       'mobileL:max-h-[calc(100vh-7.5rem)] max-h-[calc(100vh-2.5rem)] h-auto',
-    [ModalSize.Large]:
-      'mobileL:mt-14 mobileL:max-h-[calc(100vh-5rem)] max-h-[100vh]',
+    [ModalSize.Large]: 'laptop:mt-14 laptop:mb-10',
```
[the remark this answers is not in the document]
rebelchris: The large one is so far only used for the PostModal. This was mainly some misalignment from spec (which I had to discuss with Tsahi and Tomer)
So should not break any other modals.

PR #1406, opened 2022-12-02
fix: [modal migration] sign in modal

el:
## Changes

Updates the sign-in modals to use the new modal component

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done

# Extensive tests are required before it can be merged since it affects the most important part of the app


packages/shared/src/components/auth/AuthModalHeader.tsx
```diff
@@ -1,41 +1,31 @@
...
+    <Modal.Header className={className}>
       {onBack && (
         <Button
           icon={<ArrowIcon className="-rotate-90" />}
           className="mr-2 btn-tertiary"
           onClick={onBack}
         />
       )}
-      <h3 className="font-bold typo-body">{title}</h3>
-      {onClose && <ModalCloseButton onClick={onClose} />}
```
rebelchris, line 37: With this removed does it show everywhere now?
The main idea was that certain AB test don't show the close button, is that still effective?

rebelchris (approved):
Just one question I couldn't test on preview.
Otherwise it looks solid.

packages/shared/src/components/auth/AuthModalHeader.tsx
el: Yes, it shows everywhere. Should we ask product if that's ok to be consistent?

el: Seems like we have ok from the product

PR #1407, opened 2022-12-04
feat: first notification

sshanzel:
- Notifications context to be included in the boot provider
- Notifications page showing the enable settings banner
- First notification
- Generic component to handle notifications

image

@rebelchris can you help me upload the asset in the Cloudinary? It is about the preview of notification. Link to the Sketch can be found below the screen shot.

image

https://www.sketch.com/s/2867ec63-0966-4647-b1d7-b6fe696ba251/a/m1Q273L#Inspect

Preview:
image

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-702 #done


packages/shared/src/lib/styling.ts
```diff
@@ -0,0 +1,16 @@
+import classNames from 'classnames';
+
+const fadedBackground =
```
rebelchris, line 3: This seems like a lot of overhead I think we can simply get a new color match preferably?
I have the same issue now with feed filters and think we should rather ask design for a solid color?

packages/shared/src/components/notifications/NotificationItem.tsx
```diff
@@ -0,0 +1,84 @@
...
+  isUnread,
+  description,
+  avatars,
+  attachments,
+}: NotificationItemProps): ReactElement {
+  const DOMPurify = useRef<DOMPurify.DOMPurifyI>();
+
+  useEffect(() => {
+    DOMPurify.current = createDOMPurify(window);
+  }, []);
```
rebelchris, line 33: We already use this somewhere correct?
Perhaps we can convert it into a hook (just for clean code)

packages/shared/src/components/notifications/NotificationItemAttachment.tsx
```diff
@@ -0,0 +1,24 @@
...
+function NotificationItemAttachment({
+  image,
+  title,
+}: NotificationItemAttachmentProps): ReactElement {
+  return (
+    <div className="flex flex-row items-center p-4 mt-2 rounded-16 border border-theme-divider-tertiary">
+      <img
+        className="object-cover w-24 h-16 rounded-16"
+        src={image}
+        alt="Attachment preview"
```
rebelchris, line 17: This should likely be the title rather?

packages/shared/src/components/notifications/NotificationItemAvatar.tsx
```diff
@@ -0,0 +1,31 @@
...
+  type,
+  image,
+  name,
+  referenceId,
+}: NotificationItemAvatarProps): ReactElement {
+  if (type === NotificationAvatarType.Source) {
+    return <SourceButton source={{ id: referenceId, name, image }} />;
+  }
+
+  return <ProfilePicture user={{ image, id: referenceId }} />;
```
rebelchris, line 28: Generic question:

Should we maybe hard cast a type === here as well?
In case we introduce a new one as we go, rather have that render as empty?

packages/shared/src/lib/styling.ts
rebelchris: To iterate, perhaps it's better to add a new scope of colors "faded" and add all the new faded colors?
But suggestion is to ask design for all of them :)

rebelchris (changes requested):
Love it!

Really nice setup with the generic notification.
Have some suggestions for the backgrounds though

packages/shared/src/components/utilities.tsx
```diff
@@ -118,11 +119,14 @@ export const pageBorders =
   'laptop:border-r laptop:border-l border-theme-divider-tertiary';
 const pagePaddings = 'px-4 tablet:px-8';
 
+export const pageContainerClassnames = classNames(
```
el, line 122: pageContainerClassNames

packages/shared/src/lib/styling.ts
```diff
@@ -0,0 +1,16 @@
...
+  'relative before:absolute before:inset-0 before:w-full before:h-full before:opacity-24';
+
+export const getFadedBackground = (className: string): string => {
+  const isBefore = className.includes('before:bg-theme-');
+
+  if (!isBefore) {
+    throw new Error(
+      'You must need apply the background with a `before:` class',
```
el, line 11:
```
'You need to apply the background with a \'before:\' class'
```

packages/shared/src/lib/styling.ts
```diff
@@ -0,0 +1,16 @@
+import classNames from 'classnames';
+
+const fadedBackground =
+  'relative before:absolute before:inset-0 before:w-full before:h-full before:opacity-24';
+
+export const getFadedBackground = (className: string): string => {
```
el, line 6: I think a better approach would be something similar to this:
```
enum Color {
  OverlayCabbage: 'overlay-cabbage'
  ...
}
function generateFadedBackground: (color: Color): string {
  return `before:bg-theme-${color}`;
}
```

el (approved):
Great work, other than small comments, looks perfect 👏

packages/shared/src/components/notifications/NotificationItemAvatar.tsx
sshanzel: Very good point. We shall do that.

packages/shared/src/lib/styling.ts
rebelchris: Yeah let's rather ask design.

They can give us 8 digit hex colors so we can adjust them in our guidelines rather.
I prefer having a className for it, compared to a function that needs to execute.

packages/shared/src/components/notifications/NotificationItem.tsx
rebelchris: If it's reused a hook might still be best, but it's not a hard requirement, so leaving it up to you.

sshanzel: Cool. Was just checking if you are cool too with creating a hook to abstract this single useState. In my opinion, we should go with your suggestion so I will do that before moving forward. Thank you!

packages/shared/src/components/utilities.tsx
sshanzel: Sharp eyes! 🔭

packages/shared/src/lib/styling.ts
sshanzel: The lint gave an error. Let me make the parent a double quotation as the lint suggested. Thanks for mentioning!

sshanzel:
@rebelchris feel free to run a review, kindly disregard the `fadedBackground` thing until we receive a reply from the message I sent to the design team. Thank you!

packages/shared/src/components/notifications/NotificationItem.spec.tsx
```diff
@@ -0,0 +1,99 @@
...
+    const img = await screen.findByAltText(
+      `Preview cover of the article: ${attachment.title}`,
+    );
+    expect(img).toHaveAttribute('src', attachment.image);
+  });
+
+  it('should have a title', async () => {
+    const [attachment] = sampleNotification.attachments;
+    render(<NotificationItem {...sampleNotification} />);
+    await screen.findByText(attachment.title);
```
rebelchris, line 53: You await but don't expect anything here?

packages/shared/src/components/notifications/NotificationItem.spec.tsx
```diff
@@ -0,0 +1,99 @@
...
+  it('should display the avatar of the user', async () => {
+    const [, user] = sampleNotification.avatars;
+    render(<NotificationItem {...sampleNotification} />);
+    const img = await screen.findByAltText(`${user.referenceId}'s profile`);
+    expect(img).toHaveAttribute('src', user.image);
+  });
+
+  it('should not display anything if the type is unknown', async () => {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const notif = { ...sampleNotification } as any;
```
rebelchris, line 74: Should it not cast as `Notification` or something?

packages/shared/src/components/utilities.tsx
```diff
@@ -118,11 +119,14 @@ export const pageBorders =
   'laptop:border-r laptop:border-l border-theme-divider-tertiary';
 const pagePaddings = 'px-4 tablet:px-8';
 
+export const pageContainerClassNames = classNames(
+  styles.pageContainer,
```
rebelchris, line 123: Just a question:
Do we still needs these or can we potentially move them to tailwind classes now?

packages/shared/src/contexts/NotificationsContext.tsx
```diff
@@ -0,0 +1,54 @@
...
+    if (!user) return 'default';
+
+    const result = await globalThis.window.Notification.requestPermission();
+    setHasPermission(result === 'granted');
+    return result;
+  };
+
+  const data = useMemo(() => {
+    return {
+      unreadCount: 0,
```
rebelchris, line 43: This is going to be another PR?

rebelchris (changes requested):
Some minor questions and leaving it at changes until we have clarity on the faded background 🙏

packages/shared/src/components/notifications/NotificationItem.spec.tsx
sshanzel: The await will fail if there is no found element.

packages/shared/src/components/notifications/NotificationItem.spec.tsx
sshanzel: This is the part where the type of notification is not found in our system. I used any so we can put values outside of the current `enum NotificationType`.

packages/shared/src/components/utilities.tsx
sshanzel: I just checked they are pretty simple, we can probably convert them, though in a separate PR as it affects a lot of things and requires rigorous manual testing.

packages/shared/src/contexts/NotificationsContext.tsx
sshanzel: Yes, the other PR fixes this to come from the props.

packages/shared/src/lib/styling.ts
sshanzel: Now that we have concluded the usage of opacity, I think we will have to move forward with the function to generate it until we upgrade our tailwind version.

We probably need to put some comments on it to make sure it gets updated. WDYT?

rebelchris: As far as I was concerned the solution was actually to generate classes for it right?
I feel like we might be misaligned here from what feedback we received let's rather discuss it quickly today.

sshanzel: Let's jump in Gather once you are available.

sshanzel:
@rebelchris I have now applied the discussed classes for the different states.

packages/shared/src/components/notifications/NotificationItem.tsx
```diff
@@ -0,0 +1,79 @@
...
+      <NotificationItemAvatar key={avatar.referenceId} {...avatar} />
+    )) ?? [];
+  const hasAvatar = avatarComponents.some((component) => !!component);
+
+  return (
+    <button
+      type="button"
+      className={classNames(
+        'flex flex-row py-4 pl-6 pr-4 hover:bg-theme-hover focus:bg-theme-active',
+        isUnread && 'bg-theme-float',
```
rebelchris, line 39: What's the fallback for read messages in this case?

rebelchris (approved):
Amazing!

Sorry for dragging it on so long, but it looks really good now 🥳

PR #1408, opened 2022-12-04
feat: notifications page

sshanzel:
- Infinite scrolling component
- Refactored the upvoted popup modal to implement the new component (tested at preview2 and still works as intended)
- Notifications page to support infinite scrolling
- Dispatch mutation only once there is any unread notification
- Refactored the types to be placed in a single file that is all related to notifications

Can you help me upload the notification bell icon our cloudinary as part of the notification types icons.
https://www.sketch.com/s/aca0d743-919a-4f68-9411-3c155f9f85a0/a/v8LGA33#Inspect

Note: test cases are still WIP but review is doable.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-702 #done


packages/shared/src/components/profile/UpvoterList.tsx
```diff
@@ -16,19 +16,11 @@ export function UpvoterList({
...
-    queryResult.data.pages.length > 0;
-
-  const infiniteScrollRef = useFeedInfiniteScroll({
-    fetchPage: queryResult.fetchNextPage,
-    canFetchMore,
-  });
-
   return (
-    <div className="flex relative flex-col">
+    <InfiniteScrolling
```
rebelchris, line 20: Perhaps we can also add a "amount" to this hook?

packages/shared/src/lib/image.ts
```diff
@@ -11,6 +11,7 @@ export const cloudinary = {
       'https://daily-now-res.cloudinary.com/image/upload/v1665563545/public/supercharge.png',
   },
   notifications: {
+    bell: '', // TODO: update when image is uploaded to cloudinary
```
rebelchris, line 14: The bell icon is a icon right? Why do you want that added to cloudinary?

packages/webapp/pages/notifications.tsx
```diff
@@ -1,19 +1,56 @@
...
+  const { mutateAsync: readNotifications } = useMutation(() =>
+    request(`${apiUrl}/graphql`, READ_NOTIFICATIONS_MUTATION),
+  );
+  const queryResult = useInfiniteQuery<NotificationsData>(
+    ['notifications'],
+    ({ pageParam }) =>
+      request(`${apiUrl}/graphql`, NOTIFICATIONS_QUERY, {
+        first: 100,
+        after: pageParam,
+      }),
```
rebelchris, line 40: Oh sorry, see this now, that's amazing 🥳

rebelchris (changes requested):
Love it!

Really amazing work on the infinite function.
Question regarding the icon though?

packages/shared/src/components/profile/UpvoterList.tsx
sshanzel: Not sure where we are going to use the `amount` property.

packages/shared/src/lib/image.ts
sshanzel: I would also prefer the icon to be decided in the frontend based on which type belongs to which icon. But from the discussion we've had, it would be ideal to use a link for the icon as well to make our notifications super flexible.
cc @idoshamun

idoshamun: I think we referred only to the icons inside the notification component itself, not the bell that shows the notifications

sshanzel: Yes, this is that. We will be using those icons to be populated in our data, and this one is used for the first notification which is implemented in the frontend to not create a row in our table for notifications for all users.

packages/shared/src/components/notifications/Toast.spec.tsx
```diff
@@ -49,8 +53,9 @@ it('should display a toast notification', async () => {
   fireEvent.click(button);
   await screen.findByRole('alert');
   await screen.findByText('Sample Notification');
-  await act(() => new Promise((resolve) => setTimeout(resolve, 1500))); // wait for the toast to expire
-  expect(screen.queryByRole('alert')).not.toBeInTheDocument();
+  await waitForElementToBeRemoved(() => screen.queryByRole('alert'), {
```
sshanzel, line 56: This is amazing, I never knew this was supported.

rebelchris: Yep, preferred way actually as it will deal with it's own cleanup :)

packages/webapp/pages/notifications.tsx
```diff
@@ -1,20 +1,58 @@
...
-import { NotificationType } from '@dailydotdev/shared/src/graphql/notifications';
-import { getLayout as getFooterNavBarLayout } from '../components/layouts/FooterNavBarLayout';
-import { getLayout } from '../components/layouts/MainLayout';
+import {
+  getLayout,
+  getLayout as getFooterNavBarLayout,
+} from '../components/layouts/FooterNavBarLayout';
 import ProtectedPage from '../components/ProtectedPage';
 
+const checkHasUnread = (data: InfiniteData<NotificationsData>) =>
```
rebelchris, line 26:
```suggestion
const hasUnread = (data: InfiniteData<NotificationsData>) =>
```

packages/webapp/pages/notifications.tsx
```diff
@@ -1,20 +1,58 @@
...
+      request(`${apiUrl}/graphql`, NOTIFICATIONS_QUERY, {
+        first: 100,
+        after: pageParam,
+      }),
+    {
+      getNextPageParam: (lastPage) =>
+        lastPage?.notifications?.pageInfo?.hasNextPage &&
+        lastPage?.notifications?.pageInfo?.endCursor,
+      onSuccess: (data) => {
+        if (checkHasUnread(data)) {
```
rebelchris, line 48:
```suggestion
        if (hasUnread(data)) {
```

packages/webapp/pages/notifications.tsx
rebelchris: Or `hasUnreadNotifications`

rebelchris (approved):
Minor suggestion otherwise happy with the PR

packages/webapp/pages/notifications.tsx
sshanzel: I believe we use `has` as the first word on a variable, but to let it be known as a function we put some kind of prefix for it. Though I am happy to change this as well if it is actually not a usual thing in our codebase.

rebelchris (approved):
Nice work. 🥳

PR #1412, opened 2022-12-05
feat: add json schema to the manifest jsons

el:
## Changes

Added the schema properties to make sure the changes we make to these files are correct

### DIFF
--- a/packages/extension/src/manifest.json
+++ b/packages/extension/src/manifest.json
@@ -1,4 +1,5 @@
 {
+  "$schema": "https://json.schemastore.org/chrome-manifest",
   "manifest_version": 2,
   "version": "0.0.0",
   "name": "daily.dev | The Homepage Developers Deserve",
--- a/packages/webapp/public/manifest.json
+++ b/packages/webapp/public/manifest.json
@@ -1,4 +1,5 @@
 {
+  "$schema": "https://json.schemastore.org/web-manifest",
   "name": "daily.dev",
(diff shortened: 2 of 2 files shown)


packages/extension/src/manifest.json
```diff
@@ -1,4 +1,5 @@
 {
+  "$schema": "https://json.schemastore.org/chrome-manifest",
```
rebelchris, line 2: I'm a bit hesitant to make any changes to the manifest.
We don't know the full impact and it seems to be working fully so far.

rebelchris (changes requested):

packages/extension/src/manifest.json
idoshamun: The result is what matters. Let's try to build it and see if the `$schema` field gets dropped out.
If it gets dropped, then we have nothing to worry about. If it remains, let's think what to do

el: What would be the downside if the `$schema` field does not drop?

idoshamun: Manifest is something that gets reviewed very carefully by the stores. They are very sensitive to changes there, which can delay the review by days. We can't keep it in the final version of the manifest. For development version that's fine of course

el: ohh, ok. lets not complicate things then. I am closing this PR for now since it is not something super important

PR #1420, opened 2022-12-12
fix: enabled message temporary state

sshanzel:
## Changes

I didn't realize I was not able to create the PR last week.

- The design was if the user accepts, the banner should temporarily still be visible, and when they go back it should be gone. This is on "enabled" process.

Note: will fix merge conflict once in a while. Just doing some setups.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done

### DIFF
--- a/packages/shared/src/components/notifications/EnableNotification.tsx
+++ b/packages/shared/src/components/notifications/EnableNotification.tsx
@@ -1,34 +1,56 @@
-import React, { ReactElement, useContext } from 'react';
+import React, { ReactElement, useContext, useState } from 'react';
 import usePersistentContext from '../../hooks/usePersistentContext';
 import { Button } from '../buttons/Button';
 import NotificationsContext from '../../contexts/NotificationsContext';
 import { cloudinary } from '../../lib/image';
+import VIcon from '../icons/V';
+import { webappUrl } from '../../lib/constants';
 
 const DISMISS_BROWSER_PERMISSION = 'dismissBrowserPermission';
 
 function EnableNotification(): ReactElement {
+  const [isEnabled, setIsEnabled] = useState(false);
   const { hasPermission, requestPermission } = useContext(NotificationsContext);
   const [dismissed, setDismissed, isLoaded] = usePersistentContext(
     DISMISS_BROWSER_PERMISSION,
     false,
   );
 
-  if (!isLoaded || dismissed || hasPermission) {
+  const onEnable = async () => {
+    const permission = await requestPermission();
+
+    setIsEnabled(permission === 'granted');
+  };
+
+  if (!isLoaded || dismissed || (hasPermission && !isEnabled)) {
     return null;
   }
 
   return (
     <div className="relative py-4 px-6 w-full bg-theme-float border-l typo-callout border-l-theme-color-cabbage">
-      <span className="font-bold">Push notifications</span>
-      <p className="mt-2 w-3/5 text-theme-label-tertiary">
-        Stay in the loop whenever you get a mention, reply and other important
-        updates.
-      </p>
+      <span className="flex flex-row font-bold">
+        {isEnabled && <VIcon className="mr-2" />}
+        {`Push notifications${isEnabled ? ' successfully enabled' : ''}`}
+      </span>
+      {isEnabled ? (
+        <p>
+          Changing your{' '}
+          <a href={`${webappUrl}account/notifications`}>
+            notification settings
+          </a>{' '}
+          can be done anytime through your account details
+        </p>
+      ) : (
+        <p className="mt-2 w-3/5 text-theme-label-tertiary">
+          Stay in the loop whenever you get a mention, reply and other important
+          updates.
+        </p>
+      )}
       <span className="flex flex-row gap-4 mt-4">
         <Button
           buttonSize="small"
           className="w-28 btn-primary-cabbage"
-          onClick={requestPermission}
+          onClick={onEnable}
         >
           Enable
         </Button>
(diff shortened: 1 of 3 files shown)


rebelchris (approved):
Looks good, just be sure to keep Eliz's implementations in mind when testing :)

el (changes requested):

packages/shared/src/components/notifications/EnableNotification.tsx
```diff
@@ -1,34 +1,56 @@
...
 import { Button } from '../buttons/Button';
 import NotificationsContext from '../../contexts/NotificationsContext';
 import { cloudinary } from '../../lib/image';
+import VIcon from '../icons/V';
+import { webappUrl } from '../../lib/constants';
 
 const DISMISS_BROWSER_PERMISSION = 'dismissBrowserPermission';
 
 function EnableNotification(): ReactElement {
+  const [isEnabled, setIsEnabled] = useState(false);
```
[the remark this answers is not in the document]
sshanzel: Will do this first thing tomorrow. Thank you! 💯

packages/shared/src/components/notifications/EnableNotification.tsx
```diff
@@ -13,11 +16,7 @@ export enum NotificationPromptSource {
 
 const DISMISS_BROWSER_PERMISSION = 'dismissBrowserPermission';
 
-type DismissBrowserPermissions = {
-  [NotificationPromptSource.NotificationList]: boolean;
-  [NotificationPromptSource.NewComment]: boolean;
-  [NotificationPromptSource.NewSource]: boolean;
-};
+type DismissBrowserPermissions = Record<NotificationPromptSource, boolean>;
```
sshanzel, line 19: @el just made it use a record so new introduction of `NotificationPromptSource` will automatically result to error if the dev won't add the new property in the implementation.

packages/shared/src/components/notifications/EnableNotification.tsx
```diff
@@ -56,49 +70,48 @@ function EnableNotification({
...
+              href={`${webappUrl}account/notifications`}
+            >
+              notification settings
+            </a>{' '}
+            can be done anytime through your account details
+          </>
+        ) : (
+          message
+        )}
+      </p>
```
sshanzel, line 103: @el this just looks like there's a lot of changes, but the only difference is the padding/margin/border classes around the container. Also the conditional rendering of the "Push notifications` title.
Then the `p` element content.

packages/shared/src/components/notifications/EnableNotification.tsx
rebelchris: Why did you abstract it into a type and not directly in the const?
We don't really have a rule for it, but seeing as it's one singular type it might make sense to keep it inline?

Let me know what you think.

packages/shared/src/components/notifications/EnableNotification.tsx
```diff
@@ -30,10 +29,17 @@ type EnableNotificationProps = {
   parentCommentAuthorName?: string;
 };
 
+const containerClassName: Record<NotificationPromptSource, string> = {
+  notificationList: 'px-6 w-full border-l border-theme-color-cabbage',
```
rebelchris, line 33: It seems the border-theme-color-cabbage can actually be set on the main attribute as it's recurring for all 3 versions.

packages/shared/src/components/notifications/EnableNotification.tsx
```diff
@@ -56,49 +70,48 @@ function EnableNotification({
...
+        ) : (
+          message
+        )}
+      </p>
       <span className="flex flex-row gap-4 mt-4">
         <Button
           buttonSize="small"
-          className="text-white btn-primary-cabbage"
-          onClick={requestPermission}
+          className="text-white min-w-[7rem] btn-primary-cabbage"
```
rebelchris, line 107: Is this a button spec, or specific for this view?
Might be worthwhile checking with design if it's going to be guidelined

rebelchris (changes requested):
Looks great, got some minor suggestions otherwise happy with it.

packages/shared/src/components/notifications/EnableNotification.tsx
sshanzel: It was already a type, I just updated it to be a record to make it type-strict.

packages/shared/src/components/notifications/EnableNotification.tsx
sshanzel: I believe the contrast issue is still ongoing with regards to explicitly setting the `text-white`. The width is specific to this to have a minimum value for shorter content.

sshanzel: I believe they've decided to keep every cabbage button with white text.

rebelchris: Ah sorry I meant to highlight the width specifically.

sshanzel: Ah, this was specific to one of the buttons with small content which resulted in small width. And in the spec, it already has a specific size but the width is different.

rebelchris (approved):

sshanzel:
@el I have now merged the PR after fixing the conflict and applying the changes you have made. I have asked Chris to help in reviewing to ensure it is what is intended. Thank you both! 🙇

PR #1423, opened 2022-12-16
feat: introducing feed filter scroll experiment

rebelchris:
- Introduces the new scroll feed filter experiment
- In the current implementation it technically means it always shows directly on the feed (due to how we assign the new load)

Version 1:
Screenshot 2022-12-16 at 13 13 09

Version 2:
Screenshot 2022-12-16 at 13 13 32

## Events

| New | impression  | target_type: scroll block, target_id: [V1, V2] |
| New | click scroll block  | target_id: [V1, V2] |

## Manual Testing

> Note: Only shows up on desktop sizes

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-806 #done

### DIFF
--- a/packages/shared/src/components/ScrollFeedFiltersOnboarding.tsx
+++ b/packages/shared/src/components/ScrollFeedFiltersOnboarding.tsx
@@ -50,7 +50,7 @@ export default function scrollFeedFiltersOnboarding({
       )}
       style={{
         background:
-          'linear-gradient(90deg, rgba(4, 19, 30, 1) 0%, rgba(4, 19, 30, 0.605797) 100%)',
+          'linear-gradient(180deg, #04131e00 0%, #04131eb3 20%, #0e1217 100%)',
       }}
     >
       <div className="flex flex-col items-center">
--- a/packages/shared/src/components/ScrollFeedFiltersOnboarding.tsx
+++ b/packages/shared/src/components/ScrollFeedFiltersOnboarding.tsx
@@ -1,0 +1,95 @@
+import React, { ReactElement, useContext, useEffect } from 'react';
+import classNames from 'classnames';
+import { Button } from './buttons/Button';
+import { ScrollOnboardingVersion } from '../lib/featureValues';
+import { cloudinary } from '../lib/image';
+import AnalyticsContext from '../contexts/AnalyticsContext';
+import { AnalyticsEvent, TargetType } from '../lib/analytics';
+
+const GaussianBlur = (): ReactElement => {
+  return (
+    <div className="absolute w-52 h-52 rounded-full opacity-32 blur-2xl bg-theme-color-cabbage" />
+  );
+};
+
+const versionToContainerClassName: Record<ScrollOnboardingVersion, string> = {
+  [ScrollOnboardingVersion.V1]: 'h-60',
+  [ScrollOnboardingVersion.V2]: 'h-[36rem]',
+};
+const versionToButtonClassName: Record<ScrollOnboardingVersion, string> = {
+  [ScrollOnboardingVersion.V1]: 'w-[16.25rem]',
+  [ScrollOnboardingVersion.V2]: 'w-40',
+};
+const versionToButtonText: Record<ScrollOnboardingVersion, string> = {
+  [ScrollOnboardingVersion.V1]: 'Customize your feed',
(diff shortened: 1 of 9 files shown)


packages/shared/src/components/ScrollFeedFiltersOnboarding.tsx
```diff
@@ -0,0 +1,95 @@
...
+    });
+  }, []);
+
+  return (
+    <div
+      className={classNames(
+        'flex fixed bottom-0 left-60 z-3 gap-16 justify-center items-center w-full pr-40',
+        versionToContainerClassName[version],
+      )}
+      style={{
```
el, line 51: There seems to be an issue with the background color. It should not have a hard stop
image

packages/shared/src/components/ScrollFeedFiltersOnboarding.tsx
```diff
@@ -0,0 +1,95 @@
...
+        background:
+          'linear-gradient(90deg, rgba(4, 19, 30, 1) 0%, rgba(4, 19, 30, 0.605797) 100%)',
+      }}
```
el, line 54: Same issue on v2
image

packages/shared/src/components/ScrollFeedFiltersOnboarding.tsx
```diff
@@ -0,0 +1,95 @@
...
+            versionToButtonClassName[version],
+          )}
+          buttonSize="large"
+          onClick={onInitializeOnboarding}
+        >
+          {versionToButtonText[version]}
+        </Button>
+      </div>
+      {version === ScrollOnboardingVersion.V2 && (
+        <img
```
el, line 87: Do we really need an image here?

There seems to be an issue with the font 
image

packages/shared/src/components/ScrollFeedFiltersOnboarding.tsx
```diff
@@ -0,0 +1,95 @@
...
+          src={cloudinary.feedFilters.scroll}
```
el, line 88: We can hide the right image if the screen width is small
image

packages/shared/src/components/ScrollFeedFiltersOnboarding.tsx
```diff
@@ -0,0 +1,95 @@
...
+            'text-white btn-primary-cabbage ',
+            versionToButtonClassName[version],
+          )}
+          buttonSize="large"
+          onClick={onInitializeOnboarding}
+        >
+          {versionToButtonText[version]}
+        </Button>
+      </div>
+      {version === ScrollOnboardingVersion.V2 && (
```
el, line 86: V2 covers the whole page if the screen height is small
image

packages/shared/src/components/ScrollFeedFiltersOnboarding.tsx
```diff
@@ -0,0 +1,95 @@
...
+            />
+          </div>
+        )}
+        {version === ScrollOnboardingVersion.V2 && (
+          <h2 className="mb-6 font-bold typo-title2">
+            Supercharge your feed with
+            <br /> content you&apos;ll love reading!
+          </h2>
+        )}
+        <Button
```
el, line 75: I think the user should be able to dismiss as well :/

el (changes requested):

packages/shared/src/components/ScrollFeedFiltersOnboarding.tsx
rebelchris: It's per design, the font issue is known awaiting a new export from design.

packages/shared/src/components/ScrollFeedFiltersOnboarding.tsx
rebelchris: Ah that might be a good one for now.

packages/shared/src/components/ScrollFeedFiltersOnboarding.tsx
rebelchris: For now that's intended for the experiment.

el (approved):
Great work :shipit:

PR #1425, opened 2022-12-19
fix: notifications page integration

sshanzel:
- Notifications page updated integration API requirements.
- Notification Icon component based on the discussion in slack.
- Fix dom purify hook - I noticed components are not rendered properly when not using hook

Preview: 
image

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-858 #done


packages/shared/src/graphql/notifications.ts
```diff
@@ -41,6 +47,33 @@ export interface NotificationsData {
...
+  Upvote = 'Upvote',
+  Bell = 'Bell',
+}
+
+export const notificationIcon: Record<
+  NotificationIcon,
+  ComponentType<IconProps>
+> = {
+  [NotificationIcon.DailyDev]: DailyIcon,
+  [NotificationIcon.CommunityPicks]: DailyIcon,
```
idoshamun, line 63: Why this points to `DailyIcon`?

sshanzel: Ah my bad. Development stuff. I was supposed to download the asset for community picks and left it.

idoshamun (approved):
Looks great! I have one comment and one suggestion:
We might want to consider memoizing the purified title and description.

packages/shared/src/components/notifications/NotificationIcon.tsx
```diff
@@ -0,0 +1,35 @@
...
+interface NotificationItemIconProps {
+  icon: NotificationIcon;
+}
+
+function NotificationItemIcon({
+  icon,
+}: NotificationItemIconProps): ReactElement {
+  const Icon = notificationIcon[icon] ?? notificationIcon.DailyDev;
+
+  if (!notificationIcon[icon] || noBackgroundIcons.indexOf(icon) !== -1) {
```
el, line 22: `noBackgroundIcons.includes(icon)` is easier to read

packages/shared/src/components/notifications/NotificationIcon.tsx
```diff
@@ -0,0 +1,35 @@
...
+];
+
+interface NotificationItemIconProps {
+  icon: NotificationIcon;
+}
+
+function NotificationItemIcon({
+  icon,
+}: NotificationItemIconProps): ReactElement {
+  const Icon = notificationIcon[icon] ?? notificationIcon.DailyDev;
```
el, line 20: If you are defaulting to a value, why not use the default value above:
```
function NotificationItemIcon({
  icon = NotificationIcon.DailyDev,
}: NotificationItemIconProps): ReactElement {
```

packages/shared/src/graphql/notifications.ts
```diff
@@ -41,6 +47,33 @@ export interface NotificationsData {
   notifications: Connection<Notification>;
 }
 
+export enum NotificationIcon {
```
el, line 50: I dont think these should be inside the graphql directory, can you move them to a utilities directory?

packages/shared/src/graphql/notifications.ts
```diff
@@ -41,6 +47,33 @@ export interface NotificationsData {
...
+  ComponentType<IconProps>
+> = {
+  [NotificationIcon.DailyDev]: DailyIcon,
+  [NotificationIcon.CommunityPicks]: DailyIcon,
+  [NotificationIcon.Comment]: DiscussIcon,
+  [NotificationIcon.Upvote]: UpvoteIcon,
+  [NotificationIcon.Bell]: BellIcon,
+};
+
+export const notificationDefaultTheme: Record<NotificationIcon, string> = {
```
el, line 69: Lets move these to a utilities directory

el (changes requested):
Some small changes ❤️

packages/shared/src/components/notifications/NotificationIcon.tsx
sshanzel: Good point. Let's do this. I kinda forgot this API all the time 😆

packages/shared/src/components/notifications/NotificationIcon.tsx
sshanzel: That is just the props, we are defaulting to the value given by the `notificationIcon` object.

packages/shared/src/graphql/notifications.ts
sshanzel: This is part of the returned value of the GraphQL query. Do you think we should take this out?

packages/shared/src/graphql/notifications.ts
sshanzel: I agree, this is no way related to the graphql query. Let's move this.

packages/shared/src/graphql/notifications.ts
idoshamun: I agree with Eliz. We use it throughout. Let's take it to somewhere else

sshanzel: Alright. In that case, let me turn the icon property into just a string as it might result to a dependency cycle if the utils file will ever need anything inside from this file.

idoshamun: I think enum is ok for this purpose. Better than a string actually. I wouldn't worry about cycle at this point

packages/shared/src/components/notifications/NotificationItem.tsx
```diff
@@ -24,6 +24,17 @@ function NotificationItem({
...
+    useMemo(() => {
+      if (!purify?.sanitize) {
+        return { title: '', description: '' };
+      }
+
+      return {
+        title: purify.sanitize(title),
+        description: purify.sanitize(description),
+      };
+    }, [purify]);
```
idoshamun, line 37:
```suggestion
    }, [purify, title, description]);
```

sshanzel: Oh, just saw this now. The most recent push should have this covered. Thank you!

idoshamun (approved):

el (approved):
LGTM. :shipit:

PR #1426, opened 2022-12-19
feat: onesignal integration

sshanzel:
- Pulling changes on this PR should allow you to see push notifications from one signal.
- Tested locally and notifications are sent.
- To be able to test it out from scratch, first, clear your cache then go to the applications tab and click "Delete database" like below.
image

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done


sshanzel:
@idoshamun I think I am having cache issues with the build. Do you know a quick way to fix this? I will write the solution on a confluence page afterward.

idoshamun:
@sshanzel I'll look at it later if that's ok :)

sshanzel:
> @sshanzel I'll look at it later if that's ok :)

No rush. We can still set up a test environment without this yet. Thank you!

packages/extension/.env
```diff
@@ -7,3 +7,5 @@ NEXT_PUBLIC_WEBAPP_URL=http://webapp.local.com:5002
 NEXT_PUBLIC_AUTH_URL=http://sso.local.com
 NEXT_PUBLIC_HEIMDALL_URL=http://sso.local.com
 #NEXT_PUBLIC_AUTH_URL=https://sso.daily.dev
+
+NEXT_PUBLIC_ONESIGNAL_APP_ID=04cddacf-5920-47a6-b567-9ae334d26681
```
idoshamun, line 11: We need to set production `.env` file as well

packages/webapp/.env
```diff
@@ -7,5 +7,5 @@ NEXT_PUBLIC_WEBAPP_URL=/
 NEXT_PUBLIC_AUTH_URL=http://sso.local.com
 NEXT_PUBLIC_HEIMDALL_URL=http://sso.local.com
 #NEXT_PUBLIC_AUTH_URL=https://sso.daily.dev
-NEXT_PUBLIC_AUTH_URL=http://sso.apps.local.app.garden
-NEXT_PUBLIC_HEIMDALL_URL=http://heimdall.apps.local.app.garden
+
+NEXT_PUBLIC_ONESIGNAL_APP_ID=04cddacf-5920-47a6-b567-9ae334d26681
```
idoshamun, line 11: We need to set Vercel as well with the production setting

packages/webapp/public/OneSignalSDKWorker.js
```diff
@@ -0,0 +1 @@
+importScripts('https://cdn.onesignal.com/sdks/OneSignalSDKWorker.js');
```
idoshamun, line 1: We also need the updater right? At least this is what the docs mention

packages/shared/src/contexts/NotificationsContext.tsx
```diff
@@ -28,31 +32,57 @@ export const NotificationsContextProvider = ({
...
 
-  const requestPermission = async (): Promise<NotificationPermission> => {
+  const onTogglePermission = async (): Promise<NotificationPermission> => {
     if (!user) return 'default';
 
+    if (hasPermission) {
+      await OneSignalReact.setSubscription(false);
+      setHasPermission(false);
+      return 'denied';
+    }
```
idoshamun, line 47: What's the purpose of this snippet?

packages/shared/src/contexts/NotificationsContext.tsx
```diff
@@ -2,16 +2,20 @@ import React, {
   ReactElement,
   ReactNode,
   useContext,
+  useEffect,
   useMemo,
   useState,
 } from 'react';
+import OneSignalReact from 'react-onesignal';
```
idoshamun, line 9: We need to change the import of this library to dynamic import to reduce the bundle size.
Let me know if you want more information

packages/shared/src/contexts/NotificationsContext.tsx
```diff
@@ -28,31 +32,57 @@ export const NotificationsContextProvider = ({
...
+    setHasPermission(isGranted);
+    OneSignalReact.setSubscription(isGranted);
+
     return result;
   };
 
-  const data = useMemo(() => {
-    return {
+  useEffect(() => {
+    if (isInitialized || !user || isExtension) {
```
idoshamun, line 58: I'm not sure, but possibly `user` might always be defined even when you are not logged in. Can you please confirm? If so, we might need to change this condition.

idoshamun, line 58: Another point, before merging this PR we have to decide what to do with the extension. Let's discuss the possibilities. We can have a meeting about it or you can suggest

packages/shared/src/contexts/NotificationsContext.tsx
```diff
@@ -28,31 +32,57 @@ export const NotificationsContextProvider = ({
...
+    () => ({
+      isInitialized,
       unreadCount,
       hasPermission,
-      notificationsAvailable,
-      requestPermission,
-    };
-  }, [hasPermission, user]);
+      onTogglePermission,
+      get isNotificationAvailable() {
```
idoshamun, line 80: I think "available" might be confusing with whether the user approved it or not

```suggestion
      get isNotificationSupported() {
```

idoshamun (changes requested):
Please see my comments. I think we have more work to do before we can merge it.
Overall I love the structure of how you decided to implement it

packages/shared/src/contexts/NotificationsContext.tsx
sshanzel: I agree. Let's do this.

packages/shared/src/contexts/NotificationsContext.tsx
sshanzel: This used to be an abstraction of the actual `requestPermission`, there are areas where it waits for a response on what actually happened but we can turn this though into a boolean to keep it simple.

packages/webapp/public/OneSignalSDKWorker.js
sshanzel: For new implementations, like us, we shouldn't need it anymore. Also tested locally if the notification still works, and thankfully it still does.
image

packages/shared/src/contexts/NotificationsContext.tsx
sshanzel: Surely. I am planning to finish off the webapp first, so I applied some conditions to ensure the extension won't be affected in any way after webapp implementation. Once done, I will go back to the extension and see what can work.

idoshamun: So I'll keep it as a blocker. We can't merge it before the extension support

packages/shared/src/contexts/NotificationsContext.tsx
sshanzel: This is for unsubscribing. Let me turn this to a boolean value to return so it will be clearer for the consuming end.

packages/shared/src/contexts/NotificationsContext.tsx
sshanzel: This is currently the condition in our `AuthContext` when deciding whether `user` should be null or not.

image

packages/webapp/.env
sshanzel: Alright. Let me add this now for production setting.

packages/shared/src/contexts/NotificationsContext.tsx
sshanzel: Cool. Would that be a blocker for this PR only or as a whole in the notifications intermediate branch?

packages/webapp/.env
sshanzel: We might need another OneSignal App for our staging/preview environment as the URL is different from the webapp production URL.

idoshamun: I'll create one for our preview deployment and send you so you can set it on Vercel

packages/shared/src/contexts/NotificationsContext.tsx
idoshamun: Looks great!

idoshamun: For this PR

packages/webapp/.env
sshanzel: Thank you! 🚢

idoshamun: Thinking about it, we can't set a different one for the preview as it works against our production backend.
My suggestion would be to temporarily change the url of production for the testing purposes and then revert. How does it sound?

sshanzel: Sounds good to me. This will be a one time testing anyways as there will be not much things to change once setup is successful.

el (approved):

sshanzel:
@idoshamun feel free to review this PR as well, the missing piece here was the extension-related config which is in a separate PR, but the comments from the initial review should have been fixed here. Thank you for tirelessly reviewing! 🚢

idoshamun (approved):
🤩

PR #1427, opened 2022-12-20
refactor: boot data type to include notifications data

sshanzel:
- Boot provider to pass down unread count.
- Bubble component and usage on bell icon for the unread count.
- Unread count local data mutation.
- Fixed the first notification to be readable state.

Currently deployed at preview 1. You should be able to see the bubble text on header.
https://preview.app.daily.dev/

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done

### DIFF
--- a/packages/shared/src/components/buttons/Button.tsx
+++ b/packages/shared/src/components/buttons/Button.tsx
@@ -74,11 +74,12 @@ function ButtonComponent<TagName extends AllowedTags>(
     position = 'relative',
     textPosition = 'justify-center',
     readOnly,
+    iconOnly: showIconOnly,
     ...props
   }: StyledButtonProps & ButtonProps<TagName>,
   ref?: Ref<ButtonElementType<TagName>>,
 ): ReactElement {
-  const iconOnly = icon && !children && !rightIcon;
+  const iconOnly = (icon && !children && !rightIcon) || showIconOnly;
 
   const getIconWithSize = useGetIconWithSize(buttonSize, iconOnly);
 
--- a/packages/shared/src/components/layout/MainLayoutHeader.tsx
+++ b/packages/shared/src/components/layout/MainLayoutHeader.tsx
@@ -101,7 +101,7 @@ function MainLayoutHeader({
               href={`${webappUrl}/notifications`}
             >
               <Button
-                className="mr-4 btn-tertiary"
+                className="hidden laptop:flex mr-4 btn-tertiary"
                 buttonSize="small"
                 iconOnly
                 icon={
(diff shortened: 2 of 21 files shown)


packages/shared/src/components/buttons/Button.tsx
```diff
@@ -74,11 +74,12 @@ function ButtonComponent<TagName extends AllowedTags>(
     position = 'relative',
     textPosition = 'justify-center',
     readOnly,
+    iconOnly: showIconOnly,
     ...props
   }: StyledButtonProps & ButtonProps<TagName>,
   ref?: Ref<ButtonElementType<TagName>>,
 ): ReactElement {
-  const iconOnly = icon && !children && !rightIcon;
+  const iconOnly = (icon && !children && !rightIcon) || showIconOnly;
```
idoshamun, line 82: Why do we need this addition?

packages/shared/src/components/layout/MainLayoutHeader.tsx
```diff
@@ -59,6 +63,9 @@ function MainLayoutHeader({
     return <LoginButton className="hidden laptop:block" />;
   })();
 
+  const hasNotification = !!unreadCount;
+  const atNotificationsPage = checkAtNotificationsPage();
```
idoshamun, line 67: Why do we care about this?

packages/shared/src/components/notifications/utils.ts
```diff
@@ -50,3 +51,8 @@ export const notificationDefaultTheme: Record<NotificationIcon, string> = {
   [NotificationIcon.Upvote]: 'text-theme-color-blueCheese',
   [NotificationIcon.Bell]: '',
 };
+
+const notificationsUrl = `${webappUrl}/notifications`;
+
+export const checkAtNotificationsPage = (): boolean =>
```
idoshamun, line 57: This will not work in the extension

idoshamun (changes requested):
Looks great! Some comments though, also seems like you ignored the extension

packages/shared/src/components/notifications/utils.ts
sshanzel: That's alright, there is no notifications page in the extension, only in webapp so we will never have to use the comparison.

idoshamun: Oh so the bell button opens the notifications page in the webapp always?

sshanzel: Yes. Should we create a new page in the extension? Or should I ask product?

idoshamun: That's fine. Let's keep it as it is

idoshamun (approved):

PR #1429, opened 2022-12-20
fix: notifications page inaccuracies

sshanzel:
- [x] notification permission banner is missing its BG
- [x] avatars are bigger than expected and the tooltip doesn’t work
- [x] the notification item is supposed to be a link to be middle clickable
- [x] border between notification items

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done

### DIFF
--- a/packages/shared/src/components/notifications/FirstNotification.tsx
+++ b/packages/shared/src/components/notifications/FirstNotification.tsx
@@ -20,6 +20,7 @@ function FirstNotification(): ReactElement {
       isUnread={isUnread}
       type={NotificationType.System}
       icon={NotificationIcon.Bell}
+      targetUrl="#"
       title="Welcome to your new notification center!"
       description="The notification system notifies you of important events such as replies, mentions, updates etc."
     />
--- a/packages/shared/src/components/notifications/NotificationItem.tsx
+++ b/packages/shared/src/components/notifications/NotificationItem.tsx
@@ -60,7 +60,7 @@ function NotificationItem({
       type="button"
       onClick={onClick}
       className={classNames(
-        'flex flex-row py-4 pl-6 pr-4 hover:bg-theme-hover focus:bg-theme-active',
+        'flex flex-row py-4 pl-6 pr-4 hover:bg-theme-hover focus:bg-theme-active border-y border-theme-bg-primary',
         isUnread && 'bg-theme-float',
       )}
     >
--- a/packages/shared/src/components/notifications/NotificationItem.tsx
+++ b/packages/shared/src/components/notifications/NotificationItem.tsx
@@ -1,6 +1,6 @@
-import classNames from 'classnames';
-import { useRouter } from 'next/router';
 import React, { ReactElement, useMemo } from 'react';
+import classNames from 'classnames';
+import Link from 'next/link';
 import { Notification } from '../../graphql/notifications';
 import { useDomPurify } from '../../hooks/useDomPurify';
 import NotificationItemIcon from './NotificationIcon';
@@ -13,7 +13,8 @@ export interface NotificationItemProps
     'type' | 'icon' | 'title' | 'description' | 'avatars' | 'attachments'
   > {
   isUnread?: boolean;
-  targetUrl?: string;
+  targetUrl: string;
+  onClick?: (e: React.MouseEvent<HTMLAnchorElement>) => void;
 }
 
 function NotificationItem({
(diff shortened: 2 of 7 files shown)


packages/shared/src/components/notifications/NotificationItem.tsx
```diff
@@ -51,47 +52,45 @@ function NotificationItem({
...
-        isUnread && 'bg-theme-float',
-      )}
-    >
-      <NotificationItemIcon icon={icon} />
-      <div className="flex flex-col flex-1 ml-4 w-full text-left typo-callout">
-        {hasAvatar && (
-          <span className="flex flex-row gap-2 mb-4">{avatarComponents}</span>
+    <Link href={targetUrl} passHref>
+      <a
+        href="#"
```
idoshamun, line 58: I think this is a mistake right?

packages/shared/src/components/notifications/NotificationItemAvatar.tsx
```diff
@@ -18,7 +20,16 @@ function NotificationItemAvatar({
 
   if (type === NotificationAvatarType.User) {
     return (
-      <ProfilePicture user={{ image, id: referenceId }} nativeLazyLoading />
+      <ProfileTooltip
+        link={{ href: `${webappUrl}/${referenceId}` }}
```
idoshamun, line 24: Avatar has the property `targetUrl` please use it instead

idoshamun (changes requested):
Few changes

packages/shared/src/components/notifications/FirstNotification.tsx
```diff
@@ -20,6 +20,7 @@ function FirstNotification(): ReactElement {
       isUnread={isUnread}
       type={NotificationType.System}
       icon={NotificationIcon.Bell}
+      targetUrl="#"
```
el, line 23: I dont think you should do that

packages/shared/src/components/notifications/NotificationItem.tsx
```diff
@@ -51,47 +52,45 @@ function NotificationItem({
...
-        'flex flex-row py-4 pl-6 pr-4 hover:bg-theme-hover focus:bg-theme-active',
-        isUnread && 'bg-theme-float',
-      )}
-    >
-      <NotificationItemIcon icon={icon} />
-      <div className="flex flex-col flex-1 ml-4 w-full text-left typo-callout">
-        {hasAvatar && (
-          <span className="flex flex-row gap-2 mb-4">{avatarComponents}</span>
+    <Link href={targetUrl} passHref>
+      <a
```
el, line 57: Please use `button` if it doesnt have a target URL, and use `a` if it does with the href

el (changes requested):

packages/shared/src/components/notifications/FirstNotification.tsx
sshanzel: There is nothing to do here in the first notification. Not sure if there should be any action here.

packages/shared/src/components/notifications/NotificationItem.tsx
sshanzel: Every notification will have a `targetUrl`.

packages/shared/src/components/notifications/NotificationItem.tsx
sshanzel: We pass the href, from the `Link` component, so this gets overwritten. Removing it makes the linter go crazy.

packages/shared/src/components/notifications/NotificationItemAvatar.tsx
sshanzel: Oh, yep. That is much better. Let me change this.

packages/shared/src/components/notifications/FirstNotification.tsx
sshanzel: We still don't have the rebrandly link from the product, let's add it in the intermediate branch so we can close out this PR.

el: Yeah, ok for me 👍

el (approved):

packages/shared/src/components/notifications/FirstNotification.tsx
sshanzel: Thank you for the review! 🚢

PR #1430, opened 2022-12-20
feat: in-app notifications

el:
## Changes

This feature is not complete, but I wanted to create the PR before it is done to make reviews easier.

image

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android


idoshamun:
@el can you please elaborate on what's ready and what's not so I can review it accordingly?

el:
> @el can you please elaborate on what's ready and what's not so I can review it accordingly?

@idoshamun 

- The notifications are not connected to the web socket. 
- The display and hide animations are missing
- Alternative notification designs are missing

packages/shared/src/graphql/notifications.ts
```diff
@@ -83,3 +83,11 @@ export const READ_NOTIFICATIONS_MUTATION = gql`
     }
   }
 `;
+
+export const NEW_NOTIFICATIONS_SUBSCRIPTION = gql`
```
el, line 87: @idoshamun Can you check if this is correct? I got this from here:
https://github.com/dailydotdev/daily-api/blob/2bc1d3602f264f544f5b51d2a2aa31309bae5b21/src/schema/notifications.ts#L266

packages/shared/src/graphql/notifications.ts
```diff
@@ -83,3 +83,11 @@ export const READ_NOTIFICATIONS_MUTATION = gql`
...
+  subscription NewNotification {
+    newNotifications {
```
el, line 89: @idoshamun I think this should be singular, it only returns 1 notification

idoshamun: I was also considering whether it should be singular or plural. The stream will send you all the new notifications, but you will get them one by one. So it depends on the perspective. Anyway, I'll be happy to change it if it makes more sense from your perspective.

el: I usually name functions based on their return types. If it returns an array, it is plural, otherwise, it is singular.

idoshamun: Let me change that quickly

idoshamun: Deploying now

el: @idoshamun I have updated the function here to `newNotification`

idoshamun: Let me know when it's ready for review :)

packages/shared/src/components/notifications/InAppNotification.module.css
```diff
@@ -0,0 +1,32 @@
...
+
+    svg {
+        height: 1.5rem;
+        width: 1.5rem;
+    }
+}
+
+@keyframes inAppNotificationAnimationEntry {
+  from {
+    transform: translateX(10rem);
```
idoshamun, line 16: Perhaps it's better to use percentage here?

packages/shared/src/components/notifications/InAppNotification.module.css
```diff
@@ -0,0 +1,32 @@
...
+        border-radius: 6px;
+        height: 2rem;
+        width: 2rem;
+    }
+
+    svg {
+        height: 1.5rem;
+        width: 1.5rem;
+    }
+}
```
idoshamun, line 12: Shouldn't we use tailwind for that?

el: These are inside other components, it was much easier to set them like this. But I can update the inner components to pass the size if you like.

packages/shared/src/components/notifications/InAppNotification.module.css
el: Fair point let me update that.

packages/shared/src/components/notifications/InAppNotification.tsx
```diff
@@ -0,0 +1,97 @@
...
+  const closeNotification = () => {
+    setExit(true);
+    setTimeout(() => {
+      setExit(false);
+      client.setQueryData(IN_APP_NOTIFICATION_KEY, null);
+    }, 150);
+  };
+  const stopTimer = () => clearTimeout(timeoutId);
+  const startTimer = (timer: number) => {
+    timeoutId = setTimeout(closeNotification, timer);
```
idoshamun, line 39: It's a good practice to clear the timeout before we set anyone just in case it was set for some odd reason

packages/shared/src/components/notifications/InAppNotification.module.css
el: done

packages/shared/src/components/notifications/InAppNotification.tsx
el: Done 👍

packages/shared/src/hooks/useInAppNotification.ts
```diff
@@ -0,0 +1,60 @@
...
+
+interface UseInAppNotification {
+  displayNotification: (
+    notification: NewNotification,
+    params?: NotifyOptionalProps,
+  ) => void;
+  dismissNotification: () => void;
+}
+
+export interface IInAppNotification {
```
idoshamun, line 18: The double I at the beginning can be confusing

el: It was a common practice to prepend the interface name with an `I` in my previous company, I can rename it if you like

packages/shared/src/components/notifications/InAppNotification.module.css
idoshamun: I see. Let's see what Lee thinks, but I don't want to block the merge. So we can ask him to review this part on the intermediate branch

packages/shared/src/hooks/useInAppNotification.ts
el: Updated 👍

idoshamun: We usually don't this practice at this point. For me the only concern is the double I. I didn't realize for a bit that they are different types. Felt like a typo

idoshamun (approved):
Awesome! Let's merge 🤩

packages/shared/src/components/notifications/InAppNotification.module.css
sshanzel: As much as possible, we would want to sway away from using `module.css`, only on certain cases where we won't be able to avoid it such as the keyframes below.

I'll note this in the other PR so we can convert it when fixing bugs/finishing touches.

PR #1435, opened 2022-12-23
feat: update tab title based on unread count

sshanzel:
- Update tab title based on the unread count.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done


sshanzel:
@el @idoshamun @rebelchris I have tried doing `document.title` but it doesn't work on NextJS it seems. So what I did is, we will have to manually include it per page to keep it most updated whenever the unread count changes. The issue is, the title comes from the 

I have only made the change on a single page but I wanted to get others' opinions if you have a more efficient way of doing this.

Another option is, we will render the `NextSeo` component used by each page and just rather pass the variable through the `Component.layoutProps` and modify the title in our `MainLayout`. This will reduce the chance of it getting missed as you won't have to explicitly pull the context, and use the utility function for text as things will only happen in one place. Though we will have to update all the page files, still.

packages/extension/src/newtab/App.tsx
```diff
@@ -94,6 +97,12 @@ function InternalApp({
     }
   }, [contentScriptGranted]);
 
+  useEffect(() => {
+    document.title = unreadCount
+      ? `(${unreadCount}) ${DEFAULT_TAB_TITLE} `
```
idoshamun, line 102: Removed extra space

```suggestion
      ? `(${unreadCount}) ${DEFAULT_TAB_TITLE}`
```

packages/shared/src/components/notifications/utils.ts
```diff
@@ -60,3 +60,6 @@ const MAX_UNREAD_DISPLAY = 20;
 
 export const getUnreadText = (unread: number): string =>
   unread > MAX_UNREAD_DISPLAY ? `${MAX_UNREAD_DISPLAY}+` : unread.toString();
+
+export const getUnreadTitle = (count: number, title = ''): string =>
+  count ? `(${getUnreadText(count)}) ${title}` : '';
```
idoshamun, line 65: Why do we return empty string if there's no unread notifs?

idoshamun:
I'd like to suggest a different approach that will make it easier to implement.
On every change to the `unreadCount` we will use `document.title` to update the title with the relevant number of unread.
We can build the title like this:
* Fetch current title using `document.title`
* Remove existing unread notifs from the title by matching the pattern `(some number)` at the beginning
* Concat the new pattern
* Set the window title
This way you don't have to know what is the current title and so on.
It's one `useEffect` and that's it

sshanzel:
> I'd like to suggest a different approach that will make it easier to implement.
> 
> On every change to the `unreadCount` we will use `document.title` to update the title with the relevant number of unread.
> 
> We can build the title like this:
> 
> * Fetch current title using `document.title`
> 
> * Remove existing unread notifs from the title by matching the pattern `(some number)` at the beginning
> 
> * Concat the new pattern
> 
> * Set the window title
> 
> This way you don't have to know what is the current title and so on.
> 
> It's one `useEffect` and that's it

I actually did that first thing but unfortunately it was not working on nextjs app. But in extension it is working.

idoshamun:
Oh that's weird. @sshanzel did setting `document.title` never worked?

sshanzel:
> Oh that's weird. @sshanzel did setting `document.title` never worked?

On webapp only. On the extension it works as intended that's why at the extension/newtab I had to implement a different approach which is using the document.title itself.

idoshamun:
@sshanzel I see thanks. Please check my open comment and we'll proceed from there

packages/shared/src/components/notifications/utils.ts
sshanzel: This has evolved into something that should require the title, I think this was a mistake and needs to return the original title if it is passed.

sshanzel:
@idoshamun I have now applied the fix now on the utility function. For more context, on the webapp, for the `document.title` to work, we must set a long timeout before updating the title. I have noticed that anything less than 700ms doesn't work (that is on local env, on prod it could prove to be longer). But if we wait a bit more, then the header will get overridden. Then we will have to cover some edge cases such as when you navigate, NextJS will update the title again so we will have to set a timeout to update it with the correct unread count.
I think with the current process it is much more predictable and we won't have to worry about edge cases.

idoshamun:
@sshanzel thanks for checking! I agree with your observation. But I have another idea to suggest, we can utilize the `DefaultSeo` component (we already use it in the `InternalApp`) and set the `titleTemplate` property accordingly. So this way we don't need to know the title upfront and change the pages themself. Can you try that please?

packages/webapp/components/WebSeo.tsx
```diff
@@ -0,0 +1,11 @@
+import { getUnreadTitle } from '@dailydotdev/shared/src/components/notifications/utils';
+import { useNotificationContext } from '@dailydotdev/shared/src/contexts/NotificationsContext';
+import { NextSeo, NextSeoProps } from 'next-seo';
+import React, { ReactElement } from 'react';
+
+function WebSeo({ title, ...props }: NextSeoProps): ReactElement {
+  const { unreadCount } = useNotificationContext();
+  return <NextSeo {...props} title={getUnreadTitle(unreadCount, title)} />;
```
el, line 8: Did you check `titleTemplate` ? Maybe it would be cleaner?

packages/extension/src/newtab/App.tsx
```diff
@@ -94,6 +97,12 @@ function InternalApp({
     }
   }, [contentScriptGranted]);
 
+  useEffect(() => {
+    document.title = unreadCount
```
el, line 101: Why do we need to set it manually here?

sshanzel: The other implementations are only for the webapp, the extension is not affected by those changes using `<NextSeo />` - this one is specific for the extension as they work a bit differently.

packages/webapp/components/WebSeo.tsx
sshanzel: Not yet. Will check the link Ido gave then will make changes here. Thank you!

sshanzel:
@idoshamun @el thank you both for recommending the usage of `titleTemplate`, the implementation became so much cleaner.

There were some pages where it was actually utilizing the `titleTemplate`, we have two options. The first is to recreate the template by pulling the notifications context and concatenating them all of them together. The second is the current implementation, to just use a utility function for generating the template so we will use the `titleTemplate` property in one place only.

idoshamun (approved):
Looks great!

packages/webapp/components/WebSeo.tsx
el: Ahh, didnt see Ido mentioned the same. He beat me to it ^_^

PR #1438, opened 2022-12-23
feat: extension popup permission for web push

sshanzel:
- Fixed the issue of the package getting imported/initialized twice.
- We are storing a local cache to identify whether the user has subscribed from the webapp push notifications.
- I realized there could be numerous popups in the future (one candidate is scout) and could have different functionalities aside from subscribing, so I created a directory for it named `popup` and then another folder named for its context, in this case, `notifications` and then the purpose which is the name of the page `subscribe`. The endpoint becomes `http://webapp/popup/notifications/subscribe`

Preview:

https://user-images.githubusercontent.com/13744167/209546829-be9d45d1-93cc-426a-88ce-04c42cee545b.mov

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-865 #done


packages/shared/src/components/notifications/EnableNotification.tsx
```diff
@@ -39,7 +47,14 @@ function EnableNotification({
   source = NotificationPromptSource.NotificationList,
   parentCommentAuthorName,
 }: EnableNotificationProps): ReactElement {
+  const isExtension = !!process.env.TARGET_BROWSER;
```
idoshamun, line 50: I think we have a utility function for this

packages/shared/src/components/notifications/EnableNotification.tsx
```diff
@@ -39,7 +47,14 @@ function EnableNotification({
...
   const [isEnabled, setIsEnabled] = useState(false);
+  const [permissionCache, setPermissionCache] =
```
idoshamun, line 52: I think it will be better to add all the new logic here to the notifications context. I think the context can encapsulate it better, and then no matter when we want to use the logic it will already be there.
What do you think?

packages/shared/src/components/notifications/EnableNotification.tsx
```diff
@@ -52,17 +67,44 @@ function EnableNotification({
     setDismissedCache({ ...dismissedCache, [source]: value });
 
   const onEnable = async () => {
+    if (isExtension) {
+      const params = `scrollbars=no,resizable=no,status=no,location=no,toolbar=no,menubar=no,width=728,height=756,left=100,top=100`;
+      window.open(
+        `${webappUrl}popup/notifications/subscribe`,
+        'notifications subscribe popup',
```
idoshamun, line 74:
```suggestion
        'Enable notifications popup',
```

packages/webapp/pages/popup/notifications/subscribe.tsx
```diff
@@ -0,0 +1,92 @@
+import classNames from 'classnames';
```
idoshamun, line 1: Instead of `subscribe`, can we name it `enable` please?

idoshamun (changes requested):
Really love what you did there!
Great implementation and PR

packages/shared/src/components/notifications/EnableNotification.tsx
sshanzel: I agree with that. It is better to encapsulate everything we can so we will only have to make changes in one place if necessary. Let me work on that quickly.

packages/webapp/pages/popup/notifications/subscribe.tsx
sshanzel: Definitely. Let's do it 🚢

packages/extension/.env
```diff
@@ -3,7 +3,7 @@ NEXT_PUBLIC_API_URL=http://api.local.com
 NEXT_PUBLIC_SUBS_URL=ws://api.local.com/graphql
 NEXT_PUBLIC_DOMAIN=local.com
 
-NEXT_PUBLIC_WEBAPP_URL=http://webapp.local.com:5002
+NEXT_PUBLIC_WEBAPP_URL=http://webapp.local.com:5002/
```
el, line 6: Is this intentional ?

el (approved):
Great work ❤️

packages/extension/.env
sshanzel: Yes. The variable originally had the slash at the end, but with the recent move we had, made this incompatible with some current implementations, but only on local env.

sshanzel:
@el thank you for the awesome review! 🚢 

@idoshamun I have now made it part of the `NotificationsContext` but instead of copying and pasting it, I created a hook to minimize the noise inside the context itself.

packages/shared/src/components/notifications/EnableNotification.tsx
sshanzel: I instead utilized the `app` variable passed to the boot provider as it is much more accurate 🚢

packages/shared/src/contexts/NotificationsContext.tsx
```diff
@@ -100,7 +142,14 @@ export const NotificationsContextProvider = ({
...
       },
     }),
-    [hasPermission, currentUnreadCount, isInitialized, user],
+    [
+      hasPermission,
+      currentUnreadCount,
+      isInitialized,
+      user,
+      hasEnabledPermission,
+      hasPermissionCache,
```
idoshamun, line 151: I find the naming here somewhat confusing. Not sure what's the difference between `hasPermission`, `hasEnabledPermission`, `hasPermissionCache`

packages/shared/src/components/notifications/EnableNotification.tsx
```diff
@@ -53,16 +58,30 @@ function EnableNotification({
...
   };
 
+  const hasEnabled = (hasPermission || hasPermissionCache) && isEnabled;
+
+  useEffect(
+    () => () => {
+      onHasEnabledPermission(false);
+    },
+    [],
+  );
```
idoshamun, line 78: This is a bit confusing, we should favor the `return` statement in this case to make it clearer

idoshamun (changes requested):
Naming issue and that's it!

packages/shared/src/components/notifications/EnableNotification.tsx
sshanzel: Ah, I totally agree with that 🔭

sshanzel:
@idoshamun hopefully the differences are distinguishable now. Super exciting! Thank you for reviewing! 🚢 🚢 🚢

idoshamun (approved):

PR #1439, opened 2022-12-27
feat: list squads in the sidebar

el:
## Changes

image

image

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done


packages/shared/src/components/sidebar/SquadsList.tsx
```diff
@@ -0,0 +1,61 @@
...
+    path: '/new-squad',
+  };
+
+  return (
+    <>
+      {squads.map((squad) => {
+        const { handle, name, permalink, image } = squad;
+        const menuItem: SidebarMenuItem = {
+          icon: () =>
+            image ? <SquadImage {...squad} /> : <DefaultSquadIcon />,
```
idoshamun, line 41: Image should always be defined. Server side handles the fallback image.
So if user didn't provide an image for the squad, the server will take care.
But anyway we can keep this implementation as another fail-safe. Up to you

packages/shared/src/components/sidebar/SquadsList.tsx
```diff
@@ -0,0 +1,61 @@
...
+          title: name,
+          path: permalink,
+          action: () => onNavTabClick?.(`dsadsquads/${handle}`),
```
idoshamun, line 44: What's this? `dsadsquads/${handle}`

packages/shared/src/contexts/SquadsContext.tsx
```diff
@@ -0,0 +1,46 @@
...
+  loadedSquads: false,
+});
+
+interface SquadsContextProviderProps {
+  children: ReactNode;
+  squads: Squads;
+  loadedSquads?: boolean;
+}
+
+export const SquadsContextProvider = ({
```
idoshamun, line 26: Do we need another context for that or can we utilize an existing context?

packages/shared/src/graphql/squads.ts
```diff
@@ -0,0 +1,14 @@
+import { Source } from './sources';
+
+export interface Squad extends Source {
+  active: boolean;
+  handle: string;
+  image: string | null;
```
idoshamun, line 6: This cannot be null

packages/shared/src/lib/image.ts
```diff
@@ -16,4 +16,8 @@ export const cloudinary = {
     scroll:
       'https://daily-now-res.cloudinary.com/image/upload/v1671199935/public/scroll_feed_filters.svg',
   },
+  squads: {
+    imageFallback:
+      'https://res.cloudinary.com/daily-now/image/upload/f_auto/v1/placeholders/1',
```
idoshamun, line 21:
```suggestion
      'https://daily-now-res.cloudinary.com/image/upload/v1672041320/squads/squad_placeholder.jpg',
```

idoshamun, line 21:
```suggestion
      'https://daily-now-res.cloudinary.com/image/upload/v1672041320/squads/squad_placeholder.jpg',
```

idoshamun (changes requested):
Few changes 🙏

packages/shared/src/components/sidebar/SquadsList.tsx
el: Upps, fixed

packages/shared/src/contexts/SquadsContext.tsx
el: I didnt like adding another context but non of the current ones seemed related

packages/shared/src/graphql/squads.ts
el: If it is empty in the db, it returns null currently. Thats why I set it like this

packages/shared/src/lib/image.ts
el: thanks 🙏

packages/shared/src/components/sidebar/SquadsList.tsx
el: Lets leave it as it is just to be on the safe side

packages/shared/src/graphql/squads.ts
idoshamun: You're right! Eventually, I didn't set it to a required field. Sorry!

packages/shared/src/contexts/SquadsContext.tsx
idoshamun: Perhaps we can use the user context?

packages/shared/src/components/sidebar/SquadsList.tsx
```diff
@@ -0,0 +1,61 @@
...
+  return (
+    <>
+      {squads.map((squad) => {
+        const { handle, name, permalink, image } = squad;
+        const menuItem: SidebarMenuItem = {
+          icon: () =>
+            image ? <SquadImage {...squad} /> : <DefaultSquadIcon />,
+          title: name,
+          path: permalink,
+          action: () => onNavTabClick?.(`squads/${handle}`),
```
idoshamun, line 44: Do you know what this callback (`onNavTabClick`) does?
Because I'm not sure what's the right parameter to pass here

packages/shared/src/contexts/SquadsContext.tsx
el: Done 👍

packages/shared/src/components/sidebar/SquadsList.tsx
el: I think it changes the page to webapp if it is clicked in the extension

idoshamun: In this case, we should use `permalink` don't we?
The backend already returns the exact link we should use

el: onNavTabClick is used with relative URLs everywhere else. That's why I created a relative URI instead of using the permalink which is absolute
image

idoshamun: Ok, let's keep it for the final review, then. I'd love Chris to share his thought. I'm not 100% sure what's going on behind the scenes

idoshamun (approved):
Alright! Looks good

packages/shared/src/components/sidebar/SquadsList.tsx
rebelchris: I think onNavClick is only for items that are in the extension.
This always goes to webapp so think indeed permalink should be used.

You can double check how we link reading history :)

idoshamun: @rebelchris thanks for jumping in! 🙏

el: Ok, removed onnavclick 👍 (we already have path)

PR #1442, opened 2022-12-30
feat: create squad modal

el:
## Changes

Screenshot 2022-12-30 at 14 39 33
Screenshot 2022-12-30 at 14 39 42
Screenshot 2022-12-30 at 14 40 00

Screenshot 2022-12-30 at 18 32 01
Screenshot 2022-12-30 at 18 32 13

Screenshot 2022-12-30 at 14 40 22

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done


packages/shared/src/components/squads/utils.ts
```diff
@@ -0,0 +1,42 @@
...
+import { Squad } from '../../graphql/squads';
+
+export async function checkSourceExists(id: string): Promise<boolean> {
+  try {
+    const data = await request<SourceData>(`${apiUrl}/graphql`, SOURCE_QUERY, {
+      id,
+    });
+    return !!data.source;
+  } catch (err) {
+    if (err?.response?.errors?.[0].extensions?.code === 'NOT_FOUND') {
```
el, line 14: @idoshamun 

This is the error I get when we check for a source with no image:
```
Error: Cannot return null for non-nullable field Source.image.: 
```
```json

{
    "request": {
        "query": "\n  query Source($id: ID!) {\n    source(id: $id) {\n      id\n      image\n      name\n    }\n  }\n",
        "variables": {
            "id": "squad"
        }
    },
    "response": {
        "data": {
            "source": null
        },
        "status": 200,
        "errors": [
            {
                "message": "Cannot return null for non-nullable field Source.image.",
                "locations": [
                    {
                        "line": 5,
                        "column": 7
                    }
                ],
                "path": [
                    "source",
                    "image"
                ]
            }
        ],
        "headers": {
            "map": {
                "content-length": "167",
                "origin-agent-cluster": "?1",
                "content-security-policy": "default-src 'self';base-uri 'self';font-src 'self' https: data:;form-action 'self';frame-ancestors 'self';img-src 'self' data:;object-src 'none';script-src 'self';script-src-attr 'none';style-src 'self' https: 'unsafe-inline';upgrade-insecure-requests",
                "strict-transport-security": "max-age=15552000; includeSubDomains",
                "x-content-type-options": "nosniff",
                "cross-origin-embedder-policy": "require-corp",
                "date": "Fri, 30 Dec 2022 11:44:30 GMT",
                "referrer-policy": "no-referrer",
                "x-permitted-cross-domain-policies": "none",
                "content-type": "application/json; charset=utf-8",
                "cross-origin-opener-policy": "same-origin",
                "x-dns-prefetch-control": "off",
                "access-control-allow-origin": "http://webapp.local.com:5002",
                "x-download-options": "noopen",
                "vary": "Origin, Accept-Encoding",
                "access-control-allow-credentials": "true",
                "x-frame-options": "SAMEORIGIN",
                "connection": "close",
                "x-xss-protection": "0",
                "cross-origin-resource-policy": "same-origin"
            }
        }
    }
}
```

idoshamun: @el it's not possible to have a source without an image. I need to fix the db constraint

el: Ok, I am not going to do for this then 👍

el:
Sorry this turned out to be a pretty big PR ❤️

packages/shared/src/components/modals/NewSquadModal.tsx
```diff
@@ -0,0 +1,92 @@
...
+    setSquad(newSquad);
+  };
+  const stateProps: SquadStateProps = {
+    modalState,
+    form,
+    setForm,
+    onNext,
+  };
+  return (
+    <Modal
```
rebelchris, line 63: Should this not be the SteppedModal we already have?

packages/shared/src/components/modals/SquadsBetaModal.tsx
```diff
@@ -0,0 +1,92 @@
...
+        <Button className="btn-secondary" onClick={onRequestClose}>
+          Cancel
+        </Button>
+        <Button className="btn-primary-ketchup" onClick={onContinue}>
+          Continue
+        </Button>
+      </ConfirmationButtons>
+    </ConfirmationModal>
+  );
+}
```
rebelchris, line 47: Perhaps we can abstract this into it's own component (or the function you once suggested?)

packages/shared/src/components/modals/SquadsBetaModal.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+function SquadsBetaModal({
+  onRequestClose,
+  onNext,
+}: SquadsBetaModalProps): ReactElement {
+  const [showConfirmation, setShowConfirmation] = useState(false);
+
+  return (
+    <>
+      <Modal isOpen kind={Modal.Kind.FixedCenter} size={Modal.Size.Small}>
```
rebelchris, line 57: Another alternative is to incorporate this screen in the NewSquadModal, that way you will not see the flicker between modals.
It might also help with the abstraction on the sidebar part.
We have an example in the feedFilterOnboarding. 

Would love to hear what you think of that?

packages/shared/src/svg/SquadInviteMembers.tsx
```diff
@@ -0,0 +1,119 @@
...
+    <svg
+      width="420px"
+      height="215px"
+      viewBox="0 0 420 215"
+      version="1.1"
+      xmlns="http://www.w3.org/2000/svg"
+      xmlnsXlink="http://www.w3.org/1999/xlink"
+      {...props}
+    >
+      <title>Group</title>
```
rebelchris, line 16: I think the markup of these SVGs can be minified, Tsahi used to assist with that.
Perhaps we can check with design how to optimize the SVGs

rebelchris (changes requested):
Got some high level remarks around the modal use-cases.
Let me know what you think.

packages/shared/src/components/modals/NewSquadModal.tsx
el: SteppedModal does not use the modal guidelines and still has not been updated. And it is only being used in Onboarding. I didn't want to use the old method

packages/shared/src/components/modals/SquadsBetaModal.tsx
el: Lets move this to its own file for now, and create a task for the function.

packages/shared/src/components/modals/SquadsBetaModal.tsx
el: I thought about that, but the reason I went this way was to separate this since it seemed like a beta-only screen. I am also open to moving it inside if you like.

packages/shared/src/svg/SquadInviteMembers.tsx
el: Makes sense, let me take a look.

packages/shared/src/components/modals/SquadsBetaModal.tsx
rebelchris: If it doesn't have any UX issues I'm fine with this.
Just remembered it was an issue before, so we'll need to verify.

packages/shared/src/components/modals/NewSquadModal.tsx
rebelchris: Ah alright, thanks for explaining.
Is there anyway we can create a "new" stepped modal version?

I assume we'll be getting more and more "stepped" modals.

el: Sure thing, let me work on that

packages/shared/src/svg/SquadInviteMembers.tsx
el: Done 👍 

Used this app: https://jakearchibald.github.io/svgomg/

packages/shared/src/components/modals/SquadsBetaModal.tsx
el: The reason for the flicker is that the new squad modal is loaded asynchronously after the beta screen, I can fix that

packages/shared/src/components/modals/SquadsBetaModal.tsx
el: Done: https://dailydotdev.atlassian.net/browse/WT-888

packages/shared/src/graphql/squads.ts
```diff
@@ -1,14 +1,72 @@
...
+  createSquad: Squad;
+};
+
+/*
+
+SharePostInput
+      id: ID!
+      commentary: String!
+      sourceId: ID!
+*/
```
rebelchris, line 36: What is this for?

packages/shared/src/components/squads/SelectArticle.tsx
```diff
@@ -0,0 +1,61 @@
...
+        Select one <SquadTitleColor>article.</SquadTitleColor>
+      </SquadTitle>
+      <p className="py-4 text-center">
+        Make your squad aware of your reading history by sharing one article
+        with them
+      </p>
+
+      {data?.pages.map((page) =>
+        page.readHistory.edges.map((edge) => (
+          <div // eslint-disable-line jsx-a11y/click-events-have-key-events,jsx-a11y/no-static-element-interactions
```
rebelchris, line 43: Would it not be better to make this a button for the A11y linter?
Leaving it open, non urgent

rebelchris (changes requested):
Some extra remarks, otherwise happy with this.

rebelchris (approved):

packages/shared/src/components/squads/SelectArticle.tsx
el: Done

PR #1453, opened 2023-01-04
feat: stepped modal in create squad

el:

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done


packages/shared/src/components/squads/Details.tsx
```diff
@@ -18,102 +18,109 @@ import { cloudinary } from '../../lib/image';
...
-            className={{
-              container: 'w-full',
-            }}
-            valueChanged={(text) => setForm({ ...form, name: text })}
-          />
-          <TextField
-            label="Handle"
-            inputId="handle"
-            hint={!handleValid ? 'The handle is already exists' : undefined}
-            valid={handleValid}
```
el, line 82: @idoshamun @rebelchris 
Do we need any validation for the handle here? The only validation right now is if it exists or not.

Should we only allow `[a-z0-9-]` ?

rebelchris (approved):
Looks great 🙏

packages/shared/src/components/squads/Details.tsx
el: I think we should add the check on both backend and the frontend

PR #1458, opened 2023-01-05
feat: moved to code password reset flow

rebelchris:
- Moved from link password reset to code password reset

**IMPORTANT**
- Merge has to happen with a plan to support legacy extensions 
- Merge has to happen icm with product (kratos)
- Email template is still missing
- Open question around tracking events is still to come

https://user-images.githubusercontent.com/554874/210780873-630540fd-cd10-4fbb-9246-c85c243e08a5.mp4

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-816 #done


rebelchris:
@sshanzel / @el 

Not fully completed, but you should be able to review the changes already.
We moved from a link password reset to a verification code email reset.

el:
Screenshot 2023-01-05 at 15 56 53


Can you fix this background issue?

rebelchris:
@el That's my password manager auto fill option.
Not really a code-base issue I think.

packages/shared/src/hooks/useAccountEmailFlow.ts
```diff
@@ -2,44 +2,58 @@ import { useMemo, useState } from 'react';
...
     [{ type: flow, sentCount: resentCount }],
-    ({ queryKey: [{ type }] }) => initializeKratosFlow(type),
-    { enabled: isQueryEnabled },
+    ({ queryKey: [{ type }] }) =>
+      flowId
+        ? getKratosFlow(AuthFlow.Recovery, flowId)
+        : initializeKratosFlow(type),
+    {
+      ...(flowId
+        ? { ...disabledRefetch, retry: false, enabled: !!flowId }
```
sshanzel, line 54: No need for the `enabled` now since the starting condition checks if it has a value.
```suggestion
        ? { ...disabledRefetch, retry: false }
```

packages/shared/src/hooks/useAccountEmailFlow.ts
```diff
@@ -48,12 +62,12 @@ function useAccountEmailFlow(
         action: emailFlow.ui.action,
         params: {
           email,
-          method: 'link',
+          method: 'code',
```
sshanzel, line 65: We will need to make this come from the parameters. This is also used for resending the account verification email.

packages/shared/src/components/auth/CodeVerificationForm.tsx
```diff
@@ -0,0 +1,101 @@
...
+        onSubmit();
+      },
+    },
+  );
+
+  const onCodeVerification = async (e) => {
+    e.preventDefault();
+    // trackEvent({
+    //   event_name: AuthEventNames.SubmitForgotPassword,
+    // });
```
sshanzel, line 50: Just tagging this for visibility if ever this needs to be uncommented.

packages/webapp/components/layouts/AccountLayout/Security/index.tsx
```diff
@@ -100,10 +100,7 @@ function AccountSecurityDefault({
     AuthEvent.SocialRegistration,
     async (e) => {
       if (e.data?.flow) {
-        const flow = await getKratosSettingsFlow(
-          AuthFlow.Settings,
-          e.data.flow,
-        );
+        const flow = await getKratosFlow(AuthFlow.Settings, e.data.flow);
```
sshanzel, line 103: Can you verify if this still works as intended? The issue here was, they have identical calls but the difference was the query parameters.

sshanzel (changes requested):
Amazing work on this! Finally we've come to a solution for the reset password issue.

Just had a few comments about the verification as it uses the same hook that the previous reset password flow use

packages/shared/src/hooks/useAccountEmailFlow.ts
rebelchris: Still would be good to wait for the flow to be available no?
Else this will simply call and fail?

packages/shared/src/hooks/useAccountEmailFlow.ts
rebelchris: Hmm does it use the same function?
I thought that was send automatically via Kratos?

packages/webapp/components/layouts/AccountLayout/Security/index.tsx
rebelchris: Hmm weird, exact same type and call, but indeed made a difference!

rebelchris:
@sshanzel Added a version on: https://preview.app.daily.dev/

packages/shared/src/hooks/useAccountEmailFlow.ts
sshanzel: Then I think we need to have the `enabled: false` on the `else` value in the ternary operator.

packages/shared/src/hooks/useAccountEmailFlow.ts
sshanzel: It sends automatically, but if you want to re-send the email (once the timer ends after registration, or in the profile->security page), we do it manually in this flow. Just not sure if this will work the same.

rebelchris: Adjusted it, it's technically a different flow, but made sure both can live together.

packages/shared/src/hooks/useAccountEmailFlow.ts
rebelchris: That one fires based on `isQueryEnabled`?
I'm just not a 100% sure I'm following your concern here, perhaps you can explain it in our 1:1?

sshanzel: Oh, it is just that the value for `enabled` will always be true since in the ternary operator the first object won't be created unless the `flowId` is true. So omitting it should be fine. Sure, let's talk about it in our 1:1 if it is not clear.

sshanzel:
@rebelchris I tried the registraiton, resending the link when the timer runs out and resending through the account->security page. All of them work!

Though I tried the forgot password in the login form and it gave me the error below:
image

rebelchris:
@sshanzel Correct as the current Heimdall is not deployed yet.

sshanzel:
Ah, cool. In that case, this will be held off until we deploy the new one on `heimdall`? Code-wise though, everything looks amazing to me 🚢

sshanzel (approved):
Nice work on converting this flow! This will ensure the reset password will work until the code in front of them expires than visiting a link! 🚀

packages/shared/src/components/auth/AuthOptions.tsx
```diff
@@ -192,6 +197,12 @@ function AuthOptions({
     onSocialRegistration(provider);
   };
 
+  const onForgotPasswordSubmit = (_email: string, _flow: string) => {
```
el, line 200:
```suggestion
  const onForgotPasswordSubmit = (email: string, flow: string) => {
```

I think _ should only be used for `un-used` parameters

packages/shared/src/components/auth/ChangePasswordForm.tsx
```diff
@@ -0,0 +1,110 @@
...
+    (params: ValidateRegistrationParams) => submitKratosFlow(params),
+    {
+      onSuccess: ({ error, data: success }) => {
+        if (success) {
+          displayToast('Password changed successfully!');
+          return onSubmit();
+        }
+
+        if (!error?.ui) {
+          return setHint('Session could have expired!');
```
el, line 47:
```suggestion
          return setHint('Session might have expired!');
```

packages/shared/src/components/auth/CodeVerificationForm.tsx
```diff
@@ -0,0 +1,101 @@
...
+        <AuthModalText className="text-center">
+          We just sent the verification code to {initialEmail}
+        </AuthModalText>
+        <TextField
+          className={{ container: 'mt-6 w-full' }}
+          name="code"
+          type="code"
+          inputId="code"
+          label="Code"
+          hint={hint as string}
```
el, line 80: I dont think we need the type assertion here ?

packages/shared/src/components/auth/CodeVerificationForm.tsx
```diff
@@ -0,0 +1,101 @@
...
+          valid={!hint}
+          onChange={() => hint && setHint('')}
+          leftIcon={<KeyIcon />}
+        />
+        <Button className="mt-6 bg-theme-color-cabbage" type="submit">
+          Verify
+        </Button>
+        <Button
+          className="mx-auto mt-6 btn-secondary w-30"
+          tag="a"
```
el, line 90: I dont think we should use 'a' for non-link buttons

el (approved):
Some small remarks that can be addressed, otherwise great work ❤️

packages/shared/src/components/auth/CodeVerificationForm.tsx
rebelchris: Unfurtionally hint has 2 types, but our textfield expects one.
Quite impactful to retro change all of them.

packages/shared/src/components/auth/CodeVerificationForm.tsx
rebelchris: Sharp eye! Must have missed that one in a refactor.

packages/shared/src/components/auth/AuthOptions.tsx
rebelchris: Made a suggestion 🙏

PR #1464, opened 2023-01-09
refactor: post layout for pre-squad

sshanzel:
Previously we were morphing the `PostNavigation` depending on the current state and applying padding (including a lot of classes) to fix the jumpiness as the element breaks down. It feels a bit hacky so instead of breaking off any element, we will instead keep two elements and hide the `FixedPostNavigation` whenever necessary.

I think it is now much simpler to maintain vs to the numerous conditional to decide what classes to apply and where.

- Fixes the following:
  - Post skeleton on loading.
  - Post jumpy scroll.
  - Post stuck after scrolling.
  - Post content getting pushed.
- Experiment cleanup for article onboarding.
- Pre-squad tasks
  - New comment box design.
  - New comment section design (including companion) - flat layout.
  - Post layout size changes.
  - Updated spacing between elements.

image

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done


packages/shared/src/lib/featureValues.ts
```diff
@@ -22,13 +22,11 @@ export enum OnboardingFiltersLayout {
...
   V2 = 'v2',
-  V3 = 'v3',
 }
 
-export enum ScrollOnboardingVersion {
-  V1 = 'v1',
-  V2 = 'v2',
+export enum ExperimentWinner {
+  ArticleOnboarding = 'v3',
 }
```
rebelchris, line 32: Like this approach, but perhaps we should always aim to fully cleanup.
In which case is this value still used?

packages/shared/src/components/post/PostFeedFiltersOnboarding.tsx
```diff
@@ -1,145 +1,47 @@
...
-  postPreviousNext,
   onInitializeOnboarding,
 }: PostFeedFiltersOnboardingProps): ReactElement {
   const { trackEvent } = useContext(AnalyticsContext);
   useEffect(() => {
     trackEvent({
       event_name: AnalyticsEvent.Impression,
       target_type: TargetType.ArticleAnonymousCTA,
-      target_id: version,
+      target_id: ExperimentWinner.ArticleOnboarding,
```
rebelchris, line 22: I'll follow up with Sab and Ido how they wish to proceed with these things in the future. 🙏

rebelchris (changes requested):
Early review,
Let me know if you want me to test it locally (UI wise)

packages/shared/src/lib/featureValues.ts
rebelchris: NVM, see next comment 🙏

sshanzel:
> Early review,
Let me know if you want me to test it locally (UI wise)

Thank you for the early review! This PR should now be complete with the pre-squad requirements. This PR also includes the cleanup of the post onboarding experiment and fixes some UI shifts.

This is currently deployed at https://preview.app.daily.dev so you won't have to pull the changes locally, but feel free to do so if there is anything you are unable to replicate.

rebelchris (changes requested):
Looks great, just saw some copy differences with the Sketch is that intentional?

Screenshot 2023-01-09 at 11 19 59

sshanzel:
> Looks great, just saw some copy differences with the Sketch is that intentional?

The copy is actually inconsistent across the boards but I've connected with Sab and he mentioned we will be doing an AB test in this.

rebelchris:
> > Looks great, just saw some copy differences with the Sketch is that intentional?
> 
> The copy is actually inconsistent across the boards but I've connected with Sab and he mentioned we will be doing an AB test in this.

In that case, maybe we should add the flags so they are available quickly?

sshanzel:
> In that case, maybe we should add the flags so they are available quickly?

I agree. Let me add that quickly.

packages/shared/src/components/post/NewComment.tsx
```diff
@@ -1,49 +1,60 @@
...
-            user={user}
-            size="small"
-            className="mr-3 -ml-2"
-            nativeLazyLoading
-          />
-        )}
-        Start the discussion...
-      </button>
-    </NewCommentContainer>
+        Reply
```
rebelchris, line 56: I would also make this one flagged, just in case.
as this is also different in the design now.

rebelchris (changes requested):
Nice, I'd suggest one more, and then we are good to go.

packages/shared/src/components/post/NewComment.tsx
sshanzel: Good point 🔭  this should be added on the latest commit now.

rebelchris (approved):
Great work Lee!

el (approved):
Nice work, LGTM :shipit:

PR #1467, opened 2023-01-09
feat: use prompt for confirmation modals

el:
## Changes

image

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done


packages/webapp/components/layouts/AccountLayout/Security/index.tsx
```diff
@@ -90,10 +92,45 @@ function AccountSecurityDefault({
...
+  const { showPrompt } = usePrompt();
+
+  const manageSocialProviders = async ({
+    type,
+    provider,
+  }: ManageSocialProvidersProps) => {
+    setLinkProvider(provider);
+    onUpdateProviders({ [type]: provider });
+  };
+  const alreadyLinkedProvider = async (provider: string) => {
```
el, line 106: I have updated 2 modals to the new structure as a start

packages/shared/src/components/modals/Prompt.tsx
```diff
@@ -0,0 +1,47 @@
...
+      <ConfirmationButtons>
+        {cancelButton !== null && (
+          <Button
+            className="btn-secondary"
+            onClick={onFail}
+            {...(cancelButton || {})}
+          >
+            {cancelButton?.title ?? 'Cancel'}
+          </Button>
+        )}
```
rebelchris, line 34: Maybe we should be less explicit here?
In the past we had some AB test where left and right actions swap around (not sure if they are still available)

Let me know what you think.

packages/shared/src/hooks/usePrompt.ts
```diff
@@ -0,0 +1,53 @@
...
+
+type UsePromptRet = {
+  showPrompt: (options: PromptOptions) => Promise<boolean>;
+  prompt: Prompt;
+};
+
+export function usePrompt(): UsePromptRet {
+  const client = useQueryClient();
+  const { data: prompt } = useQuery<Prompt>(PROMPT_KEY);
+  const setPrompt = (data: Prompt) => client.setQueryData(PROMPT_KEY, data);
```
rebelchris, line 33: Can you test what happens if there is a case where there are 2 different prompts after each other?
So prompt 1 will prompt another one?

rebelchris (changes requested):
Nice work, love the generic approach.
Got some minor validation questions.

packages/shared/src/hooks/usePrompt.ts
el: It would replace prompt 1, I can add a stack if there is such a use-case. I never saw 2 consequent prompts at the same time.

packages/shared/src/components/modals/Prompt.tsx
el: I can add a `reverseButtons` parameter. WDYT?

packages/webapp/components/layouts/AccountLayout/Security/index.tsx
```diff
@@ -247,13 +266,6 @@ function AccountSecurityDefault({
           onRequestClose={() => setShowDeleteAccount(false)}
         />
       )}
-      {alreadyLinkedProvider && (
```
sshanzel, line 250: Nice to see these modals go. Thank you for the clean up! 🚢

sshanzel (commented):
Amazing to see the modals getting cleaned up ✨ 

Just had some non-blocking comments but everything looks great! 💪

packages/shared/src/hooks/usePrompt.ts
rebelchris: Alright, not at the same time, but in consecutive order, was wondering if react query handled it correct with the same key, but no worries don't think we have it at the moment.

packages/shared/src/components/modals/Prompt.tsx
rebelchris: That sounds like a plan, however it might be the onClose header button still always needs the close action?

el: You mean we should expose the close handler as well? I think we can do that if we need.

rebelchris: Yep, let's tackle it if it comes to that.

packages/shared/src/components/modals/Prompt.tsx
```diff
@@ -0,0 +1,47 @@
...
+  ConfirmationButtons,
+} from './ConfirmationModal';
+
+export function PromptElement(): ReactElement {
+  const { prompt } = usePrompt();
+  if (!prompt) return null;
+  const {
+    onFail,
+    onSuccess,
+    options: { title, description, cancelButton, okButton },
```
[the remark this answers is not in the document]
el: Done 👍 great suggestion

packages/shared/src/hooks/usePrompt.ts
```diff
@@ -0,0 +1,53 @@
...
+      };
+      const newPrompt: Prompt = {
+        options: promptOptions,
+        onFail: () => closeWith(false),
+        onSuccess: () => closeWith(true),
+      };
+      setPrompt(newPrompt);
+    });
+
+  return {
```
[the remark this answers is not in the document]
el: Done ❤️

rebelchris (approved):
Nice work 🥳

packages/shared/src/components/modals/Prompt.tsx
```diff
@@ -0,0 +1,62 @@
...
+    >
+      <ConfirmationHeading className={className.title}>
+        {title}
+      </ConfirmationHeading>
+      {!!description && (
+        <ConfirmationDescription className={className.description}>
+          {description}
+        </ConfirmationDescription>
+      )}
+      <ConfirmationButtons className={className.buttons}>
```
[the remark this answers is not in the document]
rebelchris: Awesome, that's a great one 🤘

PR #1468, opened 2023-01-10
fix: comment popup size

sshanzel:
Preview:
image

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done


packages/shared/src/components/modals/NewCommentModal.tsx
```diff
@@ -173,14 +173,15 @@ export default function NewCommentModal({
       contentRef={modalRef}
       onRequestClose={confirmClose}
       kind={Modal.Kind.FlexibleCenter}
-      size={Modal.Size.Small}
+      size={Modal.Size.Medium}
       tabs={Object.values(CommentTabs)}
       {...props}
+      className="min-h-[25rem]"
```
rebelchris, line 179: I'm not really sure about this one, should it not rather be something the guidelines handle instead of customizing a modal just for this?

packages/shared/src/components/modals/NewCommentModal.tsx
```diff
@@ -173,14 +173,15 @@ export default function NewCommentModal({
...
     >
       <Modal.Header.Tabs
         disabledTab={(tab) => tab === CommentTabs.Preview && disabled}
       />
-      <Modal.Body view={CommentTabs.Write}>
+      <Modal.Body className="min-h-[15rem]" view={CommentTabs.Write}>
```
rebelchris, line 184: Same here, @el what do you think?

rebelchris (changes requested):
The sizing looks great, just have some concerns about starting to customize modals again.

packages/shared/src/components/modals/NewCommentModal.tsx
sshanzel: The `min-h-[25rem]` on the container is the desired output. Kindly advise how you think we should proceed.

packages/shared/src/components/modals/NewCommentModal.tsx
el: I agree, why do we set a custom size here? You have already changed to `Modal.Size.Medium` if there is a height issue with that `Modal.Size.Medium` it should be handled in the modal file. Is there a comment modal-specific height requirement?

sshanzel: @el yes, there is a height requirement due to the UX. So when there is not much comment to preview, the modal wouldn't shrink. I was just worried that the other Modals using `Medium` might not have a height requirement.

el: Can you please check this with design? If the height requirement is comment specific, we can leave it like this, if it applies to all the modals with the size & kind, you can move the `min-height` to `modal.tsx`.

sshanzel: After discussing with Tomer. The height requirement is only specific to this, but having the `640px` height will be then a requirement when the modal is a `FixedCenter - Medium` but since this is a `FlexibleCenter - Medium` the said requirement does not apply here.

We should be good to go with the changes here. Thanks for taking a look at this 🚢

sshanzel: After Tomer's recent feedback. I will implement a new kind of modal. Kindly wait for it to get implemented before reviewing. Thanks!

sshanzel:
@rebelchris @el feel free to review the changes! 🚢 

Preview:
image

rebelchris (approved):
Let's go 🥳

PR #1471, opened 2023-01-11
feat: move confirmation modals to prompt

el:

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done


packages/shared/src/request/comment.ts
```diff
@@ -0,0 +1,28 @@
+import { useMutation } from 'react-query';
+import { DELETE_COMMENT_MUTATION } from '../graphql/comments';
+import { EmptyResponse } from '../graphql/emptyResponse';
+import { Post } from '../graphql/posts';
+import { usePostComment } from '../hooks/usePostComment';
+import { useRequestProtocol } from '../hooks/useRequestProtocol';
+import { apiUrl } from '../lib/config';
+
+export const deleteComment = ({
```
el, line 9: I didnt want to make the main components bloated, so I created this folder/files to keep the request helpers

packages/shared/src/request/post.ts
```diff
@@ -0,0 +1,16 @@
+import request from 'graphql-request';
+import { EmptyResponse } from '../graphql/emptyResponse';
+import { BAN_POST_MUTATION, DELETE_POST_MUTATION } from '../graphql/posts';
+import { apiUrl } from '../lib/config';
+
+export const banPost = (id: string): Promise<EmptyResponse> => {
```
el, line 6: I didnt want to make the main components bloated, so I created this folder/files to keep the request helpers

packages/shared/src/components/post/PostComments.tsx
```diff
@@ -116,6 +112,23 @@ export function PostComments({
...
+      description:
+        'Are you sure you want to delete your comment? This action cannot be undone.',
+      okButton: {
+        title: 'Delete',
+        className: 'btn-primary-ketchup',
+      },
+    };
+    if (await showPrompt(options)) {
+      await deleteComment({ commentId, parentId, post });
+    }
```
rebelchris, line 130: Would it not be neater to attach this to the actually button?
The cancel doesn't go to this resolve?

packages/shared/src/request/comment.ts
```diff
@@ -0,0 +1,28 @@
...
+  parentId,
+  post,
+}: {
+  commentId: string;
+  parentId: string | null;
+  post: Post;
+}): Promise<EmptyResponse> => {
+  const { deleteCommentCache } = usePostComment(post);
+  const { requestMethod } = useRequestProtocol();
+  const { mutateAsync: deleteCommentRequest } = useMutation<EmptyResponse>(
```
rebelchris, line 20: Perhaps we can move this to the `graphQL` `comments` file like you did previously to have everything uniform?

packages/shared/src/request/post.ts
rebelchris: Technically also could sit in the graphQL layer right?

rebelchris (changes requested):
Did you migrate all the existing test to the new flow?

packages/shared/src/request/post.ts
el: Yeap. I am fine with both. Either here or in graphql. Just dont want to bloat the component

rebelchris: Quite liked the fact graphQL had all the functions related to it.
So maybe we should just standardize it 🙏

el: Makes sense, done 👍

packages/shared/src/components/post/PostComments.tsx
el: this is not a callback. `showPrompt` returns `true` for okButton, and `false` for the cancel.

rebelchris (approved):
Thanks for the cleanup 🥳

PR #1477, opened 2023-01-12
feat: remove discard action modal

el:
## Changes

@rebelchris @sshanzel Please test this thoroughly. I have updated https://preview.app.daily.dev/ to point to this branch for easy testing.


packages/shared/src/components/modals/NewCommentModal.tsx
```diff
@@ -78,15 +78,26 @@ export default function NewCommentModal({
...
+      title: 'Discard comment',
+      description: 'Are you sure you want to close and discard your comment?',
+      okButton: {
+        className: 'btn-primary-ketchup',
+        title: 'Discard',
+      },
+      cancelButton: {
+        title: 'Stay',
+      },
+    };
```
sshanzel, line 92: I think we can take this outside of the component and be declared just a const.

packages/shared/src/components/modals/OnboardingModal.tsx
```diff
@@ -70,8 +70,23 @@ function OnboardingModal({
...
+      description: 'If you leave your changes will not be saved',
+      okButton: {
+        className: 'btn-primary-ketchup',
+        title: 'Discard',
+      },
+      cancelButton: {
+        title: 'Stay',
+        className: 'btn-primary-cabbage',
+      },
+    };
```
sshanzel, line 84: Tagging for visibility.

packages/shared/src/hooks/useAuthForms.ts
```diff
@@ -20,31 +19,45 @@ interface UseAuthFormsProps {
...
+        className: 'btn-secondary',
+      },
+      cancelButton: {
+        title: 'Stay',
+        className: 'btn-primary-cabbage',
+      },
+      className: {
+        buttons: 'flex-row-reverse',
+      },
+    };
```
sshanzel, line 40: Tagging for visibility.

packages/webapp/pages/signup.tsx
```diff
@@ -11,7 +11,7 @@ function Signup(): ReactElement {
   const { formRef } = useAuthForms();
   const router = useRouter();
 
-  const onClose = () => {
+  const onClose = async (): Promise<void> => {
```
sshanzel, line 14: I think we can revert this change as there seem to be no `await` call was added

sshanzel (commented):
I have reviewed code-wise (will test the preview deployment after this) - everything looks amazing and it is so much cleaner now. Thank you for removing the numerous generic discard modals. Really a sight to see 🙏 

I just had one general comment with regards to the prompt options. I think we can those outside the component itself. Probably just a regular const inside the file.

sshanzel:
Not sure they should be both primary buttons.
image


Everything else, I have checked works as intended! Great work! 🚢

packages/shared/src/components/modals/OnboardingModal.tsx
el: Fixed, thank you 🙏

packages/shared/src/hooks/useAuthForms.ts
```diff
@@ -20,31 +19,45 @@ interface UseAuthFormsProps {
...
     if (forceClose || !formRef?.current) {
       return onDiscard(e, forceClose);
     }
-    return setIsDiscardOpen(true);
-  };
-
-  const onDiscardCanceled: CloseAuthModalFunc = (e) => {
-    e.stopPropagation();
-    setIsDiscardOpen(false);
+    return openPrompt(e);
```
el, line 50: @sshanzel I dont think this prompt will be shown anywhere. The only place that uses `onDiscardAttempt` function is already passing forceClose=true in AuthOptions.tsx: 
```
  const onProfileSuccess = async () => {
    await refetchBoot();
    onSuccessfulRegistration?.();
    onClose(null, true);
  };
```

Can we clear up this or am I missing something?

sshanzel: I believe this is used in AuthModal.

rebelchris: Yeah it's a edge case for the callback on the authForm

rebelchris (commented):
Think Lee covered all my points already.
Nice cleanup 🙏

packages/shared/src/components/modals/NewCommentModal.tsx
el: Makes sense. Done.

packages/shared/src/hooks/useAuthForms.ts
el: Thanks, should be fine now.

packages/webapp/pages/signup.tsx
el: Fixed ❤️

packages/shared/src/components/modals/OnboardingModal.tsx
```diff
@@ -70,8 +70,26 @@ function OnboardingModal({
     if (forceClose) {
       return onRequestClose(e);
     }
-
-    return setIsClosing(true);
+    const options: PromptOptions = {
```
rebelchris, line 73: These can also go outside the component.

rebelchris (changes requested):
One more to go 🥳

sshanzel (approved):
Your work on the generic prompt made this cleanup possible. Can't say nothing but great work! ✨ ✨ ✨

packages/shared/src/components/modals/OnboardingModal.tsx
el: Done

rebelchris (approved):
Awesome!
Just make sure to double check once it hits production 🙏

PR #1483, opened 2023-01-13
feat: invitation page

sshanzel:
## Changes

**Note:** I will still have to write the test cases but feel free to do an review!

TODO:
- [ ] Join squad on button click.
- [ ] Anonymous/Logged in user scenarios.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done

### DIFF
--- a/packages/shared/src/components/utilities.tsx
+++ b/packages/shared/src/components/utilities.tsx
@@ -1,4 +1,9 @@
-import React, { HTMLAttributes, ReactElement, ReactHTML } from 'react';
+import React, {
+  CSSProperties,
+  HTMLAttributes,
+  ReactElement,
+  ReactHTML,
+} from 'react';
 import classNames from 'classnames';
 import classed, { ClassedHTML } from '../lib/classed';
 import styles from './utilities.module.css';
@@ -248,4 +253,11 @@ export const getContextBottomPosition = (
   const { right, bottom } = e.currentTarget.getBoundingClientRect();
 
   return { x: right, y: bottom + 4 };
+};
+
+export const fadedBackgroundStyle: CSSProperties = {
+  background: 'radial-gradient(ellipse, #c029f088 0%, #c029f000 400px)',
+  backgroundSize: '1200px 500px',
+  backgroundPosition: 'center -270px',
+  backgroundRepeat: 'no-repeat',
 };
--- a/packages/shared/src/graphql/squads.ts
+++ b/packages/shared/src/graphql/squads.ts
@@ -1,7 +1,8 @@
 import request, { gql } from 'graphql-request';
+import { USER_SHORT_INFO_FRAGMENT } from './users';
 import { apiUrl } from '../lib/config';
 import { UserShortProfile } from '../lib/user';
-import { Edge } from './common';
+import { Connection, Edge } from './common';
 import { Source } from './sources';
 
 export interface Squad extends Source {
@@ -12,6 +13,7 @@ export interface Squad extends Source {
   type: 'squad';
   description?: string;
   membersCount: number;
+  members: Connection<SquadMember>;
 }
 
 export type Squads = {
@@ -85,6 +87,21 @@ export const CREATE_SQUAD_MUTATION = gql`
   }
 `;
 
+const SOURCE_BASE_FRAGMENT = gql`
+  fragment SourceBaseFragment on Source {
+    id
+    active
+    handle
+    name
+    permalink
+    public
+    type
+    description
+    image
+    membersCount
+  }
+`;
+
 export const ADD_POST_TO_SQUAD_MUTATION = gql`
   mutation AddPostToSquad($id: ID!, $sourceId: ID!, $commentary: String!) {
     sharePost(id: $id, sourceId: $sourceId, commentary: $commentary) {
@@ -96,18 +113,10 @@ export const ADD_POST_TO_SQUAD_MUTATION = gql`
 export const SQUAD_QUERY = gql`
   query Source($handle: ID!) {
     source(id: $handle) {
-      id
-      active
-      handle
-      name
-      permalink
-      public
-      type
-      description
-      image
-      membersCount
+      ...SourceBaseFragment
     }
   }
+  ${SOURCE_BASE_FRAGMENT}
 `;
 
 export const SQUAD_MEMBERS_QUERY = gql`
@@ -117,17 +126,37 @@ export const SQUAD_MEMBERS_QUERY = gql`
         node {
           role
           user {
-            id
-            name
-            image
-            permalink
-            username
-            bio
+            ...UserShortInfoFragment
           }
         }
       }
     }
   }
+  ${USER_SHORT_INFO_FRAGMENT}
+`;
+
+export const SQUAD_INVITATION_QUERY = gql`
+  query SourceInvitationQuery($handle: ID!, $token: String!) {
+    member: sourceMemberByToken(token: $token) {
+      user {
+        ...UserShortInfoFragment
+      }
+    }
+    source(id: $handle) {
+      ...SourceBaseFragment
+      members {
+        edges {
+          node {
+            user {
+              ...UserShortInfoFragment
+            }
+          }
+        }
+      }
+    }
+  }
+  ${SOURCE_BASE_FRAGMENT}
+  ${USER_SHORT_INFO_FRAGMENT}
 `;
 
 export type SquadData = {
@@ -165,3 +194,18 @@ export async function getSquadMembers(id: string): Promise<SquadMember[]> {
   );
   return res.sourceMembers.edges?.map((edge) => edge.node);
 }
+
+interface SquadInvitation {
+  member: SquadMember;
+  source: Squad;
+}
+
+export interface GetSquadInvitationProps {
+  token?: string;
+  handle?: string;
+}
+
+export const getSquadInvitation = (
+  params: GetSquadInvitationProps,
+): Promise<SquadInvitation> =>
+  request<SquadInvitation>(`${apiUrl}/graphql`, SQUAD_INVITATION_QUERY, params);
--- a/packages/webapp/pages/squads/[handle]/[token].tsx
+++ b/packages/webapp/pages/squads/[handle]/[token].tsx
@@ -1,0 +1,107 @@
+import {
+  fadedBackgroundStyle,
+  PageContainer,
+} from '@dailydotdev/shared/src/components/utilities';
+import { useQuery } from 'react-query';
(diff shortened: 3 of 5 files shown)


packages/shared/src/components/utilities.tsx
```diff
@@ -249,3 +254,10 @@ export const getContextBottomPosition = (
 
   return { x: right, y: bottom + 4 };
 };
+
+export const fadedBackgroundStyle: CSSProperties = {
+  background: 'radial-gradient(ellipse, #c029f088 0%, #c029f000 400px)',
+  backgroundSize: '1200px 500px',
+  backgroundPosition: 'center -270px',
+  backgroundRepeat: 'no-repeat',
+};
```
rebelchris, line 263: Nice one!
Noticed the bg-repeat is also a tailwind class perhaps we can use simply that one?

rebelchris (approved):
Alright looks good so far 🙏
Just some minor remarks

packages/shared/src/components/utilities.tsx
sshanzel: Sharp eyes! 🔭  Let me simplify this quickly! 🚢

sshanzel: I tried to refactor but it seems the class is getting overpowered.

Screenshot 2023-01-15 at 10 58 17 PM

packages/shared/src/graphql/squads.ts
```diff
@@ -85,6 +87,21 @@ export const CREATE_SQUAD_MUTATION = gql`
   }
 `;
 
+const SOURCE_BASE_FRAGMENT = gql`
```
el, line 90: Nice work utilizing the fragment

packages/shared/src/components/utilities.tsx
```diff
@@ -249,3 +254,9 @@ export const getContextBottomPosition = (
 
   return { x: right, y: bottom + 4 };
 };
+
+export const fadedBackgroundStyle: CSSProperties = {
```
el, line 258: Can you move this to a css class instead?

el (approved):
Just a simple request. Other than that, looks great

packages/shared/src/components/utilities.tsx
sshanzel: Good point. Let me find a global CSS file we use.

sshanzel:
@rebelchris @el I have made some changes with regard to the data to send and how we validate the token locally. Would really appreciate if you can take a re-peek at the changes made. Thank you!

packages/shared/src/components/modals/common/Modal.tsx
```diff
@@ -29,7 +29,7 @@ export interface ModalProps extends ReactModal.Props {
 const modalKindToOverlayClassName: Record<ModalKind, string> = {
   [ModalKind.FixedCenter]: 'mobileL:justify-center pt-10 mobileL:pt-0',
   [ModalKind.FlexibleCenter]: 'justify-center',
-  [ModalKind.FlexibleTop]: 'pt-10',
+  [ModalKind.FlexibleTop]: 'pt-5 mobileL:pt-10',
```
rebelchris, line 32: This should have been pulled from main?

packages/shared/src/components/utilities.tsx
rebelchris: Is this still open?

packages/shared/src/graphql/squads.ts
```diff
@@ -82,6 +86,31 @@ export const CREATE_SQUAD_MUTATION = gql`
...
       description
+      members {
+        edges {
+          node {
+            referralToken
+          }
+        }
+      }
+    }
+  }
```
rebelchris, line 97: Why do you retrieve the token on createSquad, do we need it at that point?

packages/shared/src/graphql/squads.ts
```diff
@@ -167,3 +217,24 @@ export async function getSquadMembers(id: string): Promise<SquadMember[]> {
...
+export const getSquadInvitation = (token: string): Promise<SquadInvitation> =>
+  request(`${apiUrl}/graphql`, SQUAD_INVITATION_QUERY, { token });
+
+export const joinSquadInvitation = ({
+  token,
+  sourceId,
+}: SquadInvitationProps): Promise<SquadInvitation> =>
+  request(`${apiUrl}/graphql`, SQUAD_JOIN_MUTATION, {
+    token,
+    sourceId: sourceId.toLowerCase(),
```
rebelchris, line 239: This is always lowercase already, right?

rebelchris (changes requested):
Nice work, got some generic questions.

packages/shared/src/graphql/squads.ts
sshanzel: Ah, yes, this is source id which is UUID, so no need to lowercase.

packages/shared/src/graphql/squads.ts
sshanzel: Yes, this is when the user will be copying the invitation link.

packages/shared/src/components/utilities.tsx
sshanzel: Yep. My bad. I must have stashed it somewhere. Should be fixed now. Thank you for the notice! 🚢

sshanzel:
I have also included the tests on the latest commit. Let me know what you think! 🚢

packages/shared/__tests__/fixture/settings.ts
```diff
@@ -0,0 +1,39 @@
+import {
+  SettingsContextData,
+  ThemeMode,
+} from '../../src/contexts/SettingsContext';
+
+export const createTestSettings = (
```
rebelchris, line 6: Wohoo finally! 🥳

packages/shared/src/components/MainLayout.tsx
```diff
@@ -128,7 +130,7 @@ export default function MainLayout({
           hasBanner ? 'laptop:pt-22' : 'laptop:pt-14',
         )}
       >
-        {!showOnlyLogo && sidebarRendered !== null && (
+        {showSidebar && !showOnlyLogo && sidebarRendered !== null && (
```
rebelchris, line 133: Are you sure this is right?
Not 100% sure this should fire as it has the mobile hidden sidebar?

packages/shared/src/components/cards/SourceButton.tsx
```diff
@@ -33,7 +33,7 @@ export default function SourceButton({
           name: source.name,
           image: source.image,
           permalink: getSourcePermalink(source.id),
-          username: source.id,
+          username: source.handle,
```
rebelchris, line 36: Maybe fallback to id, not sure we set handle for all existing sources?

packages/webapp/__tests__/SquadTokenPage.tsx
```diff
@@ -0,0 +1,294 @@
...
+  jest.clearAllMocks();
+  nock.cleanAll();
+  replaced = '';
+});
+
+let client: QueryClient;
+
+const defaultToken = 'ki3YLcxvSZ2Q6KgMBZvMbly1gnrZ6JnIrhTpUML-Hua';
+const generateMember = (i: string | number): SquadMember => ({
+  user: {
```
rebelchris, line 55: Consider creating a fixture rather, it will help with future tests.

packages/webapp/pages/squads/[handle]/[token].tsx
```diff
@@ -0,0 +1,172 @@
...
+  const { user, source } = member;
+  const others = source.members.edges.filter(
+    ({ node }) => node.user.id !== user.id,
+  );
+  const othersLabel = getOthers(
+    source.members.edges.filter(({ node }) => node.user.id !== user.id),
+  );
+
+  return (
+    <PageContainer className="relative justify-center items-center min-h-[calc(100vh-3.5rem)]">
```
Comment (rebelchris) [inline packages/webapp/pages/squads/[handle]/[token].tsx:104]: Is the min height calc really needed?
Seems like we don't use it anywhere else at the moment

rebelchris (changes requested):
Nice work Lee,
Small suggestion on the test cases.

packages/shared/src/components/cards/SourceButton.tsx
sshanzel: Our DB column for `source.handle` is `NOT NULL` - this should be safe, I think.

packages/webapp/__tests__/SquadTokenPage.tsx
sshanzel: I agree, though I think we can keep it a function, I had some issues earlier keeping it an object (modifying properties running across tests). Let me store this in a generic file.

packages/webapp/pages/squads/[handle]/[token].tsx
Comment (sshanzel) [reply packages/webapp/pages/squads/[handle]/[token].tsx]: I have used padding instead of min height. WDYT?

Comment (rebelchris) [reply packages/webapp/pages/squads/[handle]/[token].tsx]: Yep! Nice one

packages/shared/src/components/MainLayout.tsx
sshanzel: Nice catch! Just tested it now, the sidebar is now available on mobile and only hidden on desktop.

sshanzel: Good point, I agree, better to ask them as this is a new behavior.

sshanzel: Tsahi has confirmed, this is indeed a new behavior where we show header buttons but sidebar is hidden 🚢

packages/extension/src/companion/CompanionContextMenu.tsx
```diff
@@ -90,7 +91,10 @@ export default function CompanionContextMenu({
         <Item onClick={onBookmark}>
           <BookmarkIcon
             size="medium"
-            className="mr-2"
+            className={classNames(
+              'mr-2',
+              postData?.bookmarked && 'text-theme-color-bun',
+            )}
```
rebelchris, line 97: Is this a merge conflict?
We should aim to get this live on the main branch before squads if possible.

rebelchris (approved):
Nice work Lee,

All looks good, but can we please make sure the color changes are actually active on main branch?
We need to publish a new extension version outside of squads context.

packages/extension/src/companion/CompanionContextMenu.tsx
sshanzel: I think so too. I just pulled main and this seems to be gone

PR #1484, opened 2023-01-15
feat: squad member modal

rebelchris:
- Modal for squad members
- Modified existing upvoter modal a bit for it.
- Report is a seperate ticket

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-849 #done

### DIFF
--- a/packages/shared/src/components/modals/SquadMemberModal.tsx
+++ b/packages/shared/src/components/modals/SquadMemberModal.tsx
@@ -1,42 +1,45 @@
 import React, { ReactElement, useRef, useState } from 'react';
 import { useInfiniteQuery } from 'react-query';
-import { UpvoterList } from '../profile/UpvoterList';
+import request from 'graphql-request';
 import {
   UserShortInfoPlaceholder,
   UserShortInfoPlaceholderProps,
 } from '../profile/UserShortInfoPlaceholder';
 import { apiUrl } from '../../lib/config';
-import { RequestQuery, UpvotesData } from '../../graphql/common';
-import { useRequestProtocol } from '../../hooks/useRequestProtocol';
 import { Modal, ModalProps } from './common/Modal';
+import {
+  Squad,
+  SQUAD_MEMBERS_QUERY,
+  SquadEdgesData,
+} from '../../graphql/squads';
+import { SquadMemberList } from '../profile/SquadMemberList';
 
 export interface UpvotedPopupModalProps extends ModalProps {
+  squad: Squad;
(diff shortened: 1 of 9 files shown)


packages/shared/src/components/profile/SquadMemberList.tsx
```diff
@@ -0,0 +1,60 @@
+import React, { ReactElement } from 'react';
```
sshanzel, line 1: I think the only difference here with the `UpvoterList` is the property to access `sourceMembers`. We can probably create a ticket to create a generic component to be used by both of the 🚢

sshanzel (approved):
Love how small the footprint is with using a modal with the new system. Amazing refactor too on making the `UserShortInfo` to be more generic.

packages/shared/src/components/profile/SquadMemberList.tsx
rebelchris: Yeah agreed, would be a nice win.
Let me add it quickly.

rebelchris: Well the one does have a extra button on the right.
But maybe it can be a generic list component?
Added a ticket now.

packages/shared/src/components/squads/SquadHeaderBar.tsx
```diff
@@ -23,13 +25,24 @@ export function SquadHeaderBar({
   ...props
 }: SquadHeaderBarProps): ReactElement {
   const { onMenuClick } = useSquadMenu();
+  const { openModal } = useLazyModal();
```
el, line 28: Really cool how it is being used.

packages/shared/src/components/modals/SquadMemberModal.tsx
```diff
@@ -0,0 +1,78 @@
...
+      size={Modal.Size.Small}
+      {...modalProps}
+    >
+      <Modal.Header title="Squad members" />
+      <Modal.Body
+        className="py-2 px-0"
+        data-testid={`List of ${queryKey[0]} with ID ${queryKey[1]}`}
+        ref={container}
+      >
+        {page && page.sourceMembers.edges.length > 0 ? (
```
el, line 64:
```
        {page?.sourceMembers.edges.length > 0 ? (
```

rebelchris: Ah good one, I'll retro fix it :)

PR #1486, opened 2023-01-17
refactor: user lists into a generic component

sshanzel:
- Generic list and modal for user short info.
- Since I was already working on something related to infinite scrolling, I have decided to take on the fix for applying infinite scrolling on reading history too.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-936 #done
WT-925 #done


packages/shared/src/components/profile/SquadMemberList.tsx
```diff
@@ -1,60 +0,0 @@
...
-                  user={user}
-                  className="flex-1"
-                />
-              </Link>
-              <Button
-                buttonSize="small"
-                className="btn-tertiary"
-                iconOnly
-                icon={<MenuIcon />}
-              />
```
rebelchris, line 52: We are now missing this option for the squad list.

packages/shared/src/components/profile/UserList.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+export interface UserListProps {
+  scrollingContainer?: HTMLElement;
+  appendTooltipTo?: HTMLElement;
+  scrollingProps: Omit<InfiniteScrollingProps, 'children'>;
+  users: UserShortProfile[];
+  placeholderAmount?: number;
+}
+
+function UserList({
```
rebelchris, line 18: Perhaps you can add rightButton here, with the conditionalWrapper?

rebelchris (changes requested):
Nice really cool how you made it super generic!
I did notice one removal which we have to support.

My advice would be to look at introducing a optional right button that can get conditional wrapped?
*(also note that the classes differ a little bit)

packages/shared/src/components/profile/SquadMemberList.tsx
```diff
@@ -1,60 +0,0 @@
...
-                'hover:bg-theme-hover hover:cursor-pointer',
-              )}
-            >
-              <Link href={user.permalink}>
-                <UserShortInfo
-                  {...props}
-                  tag="a"
-                  href={user.permalink}
-                  user={user}
-                  className="flex-1"
```
rebelchris, line 44: Here is the different class

rebelchris (changes requested):
Another advice for future PRs would be to abstract per ticket.
This now entails 2 tickets, it's not urgent, but for future reference it makes more sense to keep PRs to the bare minimum

sshanzel:
> Another advice for future PRs would be to abstract per ticket.

Sorry I forgot to mention that I contained the changes for article history in this commit https://github.com/dailydotdev/apps/pull/1486/files - though I agree, it would probably better to have them in two different PRs. Will do that in the future.

sshanzel:
I have now refactored it to make it possible to include an options button which parent component which has the capacity to decide what to do when clicked.

image

rebelchris (changes requested):

packages/shared/src/components/modals/UserListModal.tsx
```diff
@@ -0,0 +1,54 @@
...
+  ...props
+}: UserListModalProps): ReactElement {
+  const container = useRef<HTMLElement>();
+  const [modalRef, setModalRef] = useState<HTMLElement>();
+
+  return (
+    <Modal
+      contentRef={(e) => setModalRef(e)}
+      kind={Modal.Kind.FlexibleCenter}
+      size={Modal.Size.Small}
```
rebelchris, line 32: Also this is dynamic actually, currently the squad one is bigger then upvotes.
But you might want to check with design what's the idea there.

rebelchris (changes requested):

packages/shared/src/components/profile/UserList.tsx
```diff
@@ -0,0 +1,58 @@
...
+        <Link key={user.username} href={user.permalink}>
+          <UserShortInfo {...props} tag="a" href={user.permalink} user={user}>
+            {onOptionsClick && (
+              <Button
+                buttonSize="small"
+                className="m-auto mr-0 btn-tertiary"
+                iconOnly
+                icon={<MenuIcon />}
+                onClick={(e) => onOptionsClick(e, user, i)}
+              />
```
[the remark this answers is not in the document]
sshanzel: Oh, let me take this out then and make it a ReactNode but a function like our Moda.StepWrapper so we can get the info from the inside.

packages/shared/src/components/modals/UserListModal.tsx
sshanzel: Oh, the previous implementation had the props below. Let me check the accurate value in the Sketch then.
```
size={Modal.Size.Small}
```

packages/shared/src/components/profile/UserList.tsx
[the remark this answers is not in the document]
sshanzel: What do you think of the changes in the commit below? This way we can compare the values if the user is the owner. I have commented some sample code we can use later.

https://github.com/dailydotdev/apps/pull/1486/commits/556ad4831f7de19f85a94edb5a0546aee806308b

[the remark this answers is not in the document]
rebelchris: Yeah that will work, not 100% sure on naming as it's not the full content.
Perhaps something as extra or additionalContent would make more sense?

But the idea is amazing!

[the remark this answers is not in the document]
sshanzel: I totally agree with the `additionalContent`. I kinda ran out of variable names in my head 😆

packages/shared/src/components/modals/UserListModal.tsx
sshanzel: Should now have the accurate value for size 🚢

packages/shared/src/components/modals/UserListModal.tsx
```diff
@@ -0,0 +1,54 @@
...
```
rebelchris, line 32: Should it not accept this as props now?

rebelchris (changes requested):
Awesome, just one question around the passing of the size.

packages/shared/src/components/modals/UserListModal.tsx
sshanzel: The spread operator below overrides this property since the prop extended from `ModalProps`.

sshanzel: Oh, wait. Yep, removing this will be better and make it come from the props.

rebelchris: Yeah accept it from props, but maybe default to Small or something like that.

sshanzel:
@rebelchris thank you so much for the effort! Really indeed a deep review 🙏

rebelchris (approved):
Nice work Lee!
Was worth the upgrade here 🥳

PR #1495, opened 2023-01-18
feat: unauthorized component

sshanzel:
## Changes

- Unauthorized components whose contents can be overwritten.
- As I noticed new pages with centered content are being introduced more and more, I created a dedicated container for it.
- This includes introducing a new tailwind CSS for min-height that considers the page header for computing the height itself.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-821 #done

### DIFF
--- a/packages/shared/src/components/Icon.tsx
+++ b/packages/shared/src/components/Icon.tsx
@@ -9,7 +9,8 @@ export type Size =
   | 'xlarge'
   | 'xxlarge'
   | 'xxxlarge'
-  | 'xxxxlarge';
+  | 'xxxxlarge'
+  | 'heading';
 
 const IconSize: Record<Size, string> = {
   xsmall: 'w-3 h-3',
@@ -20,6 +21,7 @@ const IconSize: Record<Size, string> = {
   xxlarge: 'w-10 h-10',
   xxxlarge: 'w-14 h-14',
   xxxxlarge: 'w-16 h-16',
+  heading: 'w-20 h-20',
 };
 
 type ItemType = React.ComponentType<{ className }>;
--- a/packages/shared/src/components/Icon.tsx
+++ b/packages/shared/src/components/Icon.tsx
@@ -9,8 +9,7 @@ export type Size =
   | 'xlarge'
   | 'xxlarge'
   | 'xxxlarge'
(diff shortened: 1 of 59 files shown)


packages/shared/src/components/Icon.tsx
```diff
@@ -9,7 +9,8 @@ export type Size =
   | 'xlarge'
   | 'xxlarge'
   | 'xxxlarge'
-  | 'xxxxlarge';
+  | 'xxxxlarge'
+  | 'heading';
```
rebelchris, line 13: Maybe rather check with Design on naming as it needs to match with their system.
Not 100% sure if the sizing is still on point.

rebelchris (changes requested):
Nice one!

One small remark on naming.

packages/shared/src/components/Icon.tsx
sshanzel: Tomer just responded, he reverted the icon to be at `xxxlarge` size. Let me remove this new addition quickly.

packages/shared/src/components/errors/SquadLoading.tsx
```diff
@@ -0,0 +1,25 @@
+import React, { ReactElement } from 'react';
+import classed from '../../lib/classed';
+
+const PlaceholderElement = classed('div', 'bg-theme-float');
+const RectangleElement = classed(PlaceholderElement, 'rounded-12');
+
+function SquadLoading(): ReactElement {
+  return (
+    <div className="flex overflow-hidden flex-col items-center pt-6 w-full max-h-page">
```
rebelchris, line 9: Something I only thought about now, did you make sure this also works neatly on mobile?

rebelchris (approved):
Nice just one non blocking question 🥳

packages/shared/src/components/errors/SquadLoading.tsx
sshanzel: Good catch! I have now pushed a commit that displays like below:

mobile:
image

tablet:
image

laptop:
image

For the Unauthorized, it works very well on mobile devices.

PR #1501, opened 2023-01-19
feat: squad referral

sshanzel:
## Changes

- Refactored `LoginState` to include optional properties such as `referral`.
- Since `referral` was already built, there are no further changes required in the backend/auth system.
- After checking the taxonomy, I think it would be better to do it in one go for the analytics.

image

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-846 #done

### DIFF
--- a/packages/shared/src/contexts/AuthContext.tsx
+++ b/packages/shared/src/contexts/AuthContext.tsx
@@ -19,6 +19,7 @@ import { Squad } from '../graphql/squads';
 export interface LoginState {
   trigger: AuthTriggersOrString;
   referral?: string;
+  onRegistrationSuccess?: () => void;
 }
 
 type LoginOptions = Omit<LoginState, 'trigger'>;
--- a/packages/shared/src/contexts/AuthContext.tsx
+++ b/packages/shared/src/contexts/AuthContext.tsx
@@ -16,14 +16,19 @@ import { isCompanionActivated } from '../lib/element';
 import { AuthTriggers, AuthTriggersOrString } from '../lib/auth';
 import { Squad } from '../graphql/squads';
 
-export type LoginState = { trigger: AuthTriggersOrString };
+export interface LoginState {
+  trigger: AuthTriggersOrString;
+  referral?: string;
+}
+
+type LoginOptions = Omit<LoginState, 'trigger'>;
 
 export interface AuthContextData {
   user?: LoggedUser;
   referral?: string;
   trackingId?: string;
   shouldShowLogin: boolean;
-  showLogin: (trigger: AuthTriggersOrString) => void;
+  showLogin: (trigger: AuthTriggersOrString, options?: LoginOptions) => void;
   closeLogin: () => void;
   loginState?: LoginState;
   logout: () => Promise<void>;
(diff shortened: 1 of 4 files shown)


packages/shared/src/contexts/AuthContext.tsx
```diff
@@ -16,14 +16,19 @@ import { isCompanionActivated } from '../lib/element';
...
+
+type LoginOptions = Omit<LoginState, 'trigger'>;
 
 export interface AuthContextData {
   user?: LoggedUser;
   referral?: string;
   trackingId?: string;
   shouldShowLogin: boolean;
-  showLogin: (trigger: AuthTriggersOrString) => void;
+  showLogin: (trigger: AuthTriggersOrString, options?: LoginOptions) => void;
```
rebelchris, line 31: Alternatively would it not be better to make this a Object? Then we can simply pass LoginState?

packages/webapp/pages/squads/[handle]/[token].tsx
```diff
@@ -83,9 +83,9 @@ const SquadReferral = ({ token, handle }: SquadReferralProps): ReactElement => {
   );
 
   const onJoinClick = async () => {
-    if (!loggedUser) return showLogin('join squad');
+    if (loggedUser) return onJoinSquad();
 
-    return onJoinSquad();
+    return showLogin('join squad', { referral: member.user.id });
```
Comment (rebelchris) [inline packages/webapp/pages/squads/[handle]/[token].tsx:88]: Don't think the order is right, this was just a test I think?

rebelchris (changes requested):
Nice, got some small suggestions

packages/webapp/pages/squads/[handle]/[token].tsx
Comment (sshanzel) [reply packages/webapp/pages/squads/[handle]/[token].tsx:88]: This should be right. I switched as I think there could be additional things we do under this object in the future. We can revert though.

packages/shared/src/contexts/AuthContext.tsx
sshanzel: There would be too many changes with the current implementations and I didn't want to put too much stress on the review. Probably let's do it in a separate ticket. WDYT?

sshanzel: There would be too many changes with the current implementations and I didn't want to put too much stress on the review. Probably let's do it in a separate ticket. WDYT?

idoshamun:
Can you please tag when it's ready?

packages/webapp/pages/squads/[handle]/[token].tsx
Comment (rebelchris) [reply packages/webapp/pages/squads/[handle]/[token].tsx:88]: Hmm but now once you login, it has no callback right?
Where before it logged in, then do whatever you invoked.

Or am I understanding it wrong?

packages/shared/src/contexts/AuthContext.tsx
rebelchris: Yeah makes sense, we can do it separate, just lodge a ticket please.

sshanzel:
> Can you please tag when it's ready?

Yes, this is now ready. When you register by clicking the "Join Squad", it adds the referral to the user information. The tracking events will be part of the task to manage all tracking.

packages/webapp/pages/squads/[handle]/[token].tsx
Comment (sshanzel) [reply packages/webapp/pages/squads/[handle]/[token].tsx:88]: Oh, the current setup would make the user have to click the "Join Squad" again. Let me make it automatic.

rebelchris (approved):
@idoshamun Ready for your eyes 👀

el (approved):
LGTM :shipit:

idoshamun (approved):
Perfect!

PR #1502, opened 2023-01-20
feat: display the feed for the squad source

el:
## Changes

image

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done

### DIFF
--- a/packages/shared/src/graphql/feed.ts
+++ b/packages/shared/src/graphql/feed.ts
@@ -22,6 +22,11 @@ export const FEED_POST_FRAGMENT = gql`
       name
       image
     }
+    sharedPost {
+      id
+      title
+      image
+    }
     permalink
     numComments
     numUpvotes
@@ -41,6 +46,7 @@ export const FEED_POST_FRAGMENT = gql`
     }
     trending
     tags
+    type
   }
 `;
 
@@ -161,12 +167,14 @@ export const SOURCE_FEED_QUERY = gql`
     $first: Int
     $after: String
     $ranking: Ranking
+    $supportedTypes: [String!] = ["article"]
   ) {
     page: sourceFeed(
       source: $source
       first: $first
       after: $after
       ranking: $ranking
+      supportedTypes: $supportedTypes
     ) {
       ...FeedPostConnection
     }
--- a/packages/shared/src/hooks/useFeed.ts
+++ b/packages/shared/src/hooks/useFeed.ts
@@ -161,20 +161,10 @@ export default function useFeed<T>(
       newItems = feedQuery.data.pages.flatMap(
         ({ page }, pageIndex): FeedItem[] => {
           const posts: FeedItem[] = [];
-          if (onNewSquadPost) {
+          if (onNewSquadPost && !pageIndex) {
             posts.push({
               type: 'new_squad_post',
               action: onNewSquadPost,
-            });
-          }
-          if (adsQuery.data?.pages[pageIndex]) {
-            posts.push({
-              type: 'ad',
-              ad: adsQuery.data?.pages[pageIndex],
-            });
-          } else if (!onNewSquadPost) {
-            posts.push({
-              type: 'placeholder',
             });
           }
           page.edges.map(({ node }, index) =>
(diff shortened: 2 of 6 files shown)


packages/shared/src/graphql/feed.ts
```diff
@@ -22,6 +22,11 @@ export const FEED_POST_FRAGMENT = gql`
       name
       image
     }
+    sharedPost {
```
el, line 25: This does not cause any issues on other source feeds and just returns null. But we can also make it conditional if you want.

packages/shared/src/hooks/useFeed.ts
```diff
@@ -159,22 +160,31 @@ export default function useFeed<T>(
...
+          const posts: FeedItem[] = [];
+          if (onNewSquadPost) {
+            posts.push({
+              type: 'new_squad_post',
+              action: onNewSquadPost,
+            });
+          }
           if (adsQuery.data?.pages[pageIndex]) {
-            posts.splice(adSpot, 0, {
+            posts.push({
```
rebelchris, line 171: Not sure if this is correct.
It should inject ads in certain spots.

packages/webapp/pages/squads/[handle]/index.tsx
```diff
@@ -49,7 +49,11 @@ const SquadPage = ({ handle }: SourcePageProps): ReactElement => {
   const { user } = useContext(AuthContext);
   // Must be memoized to prevent refreshing the feed
   const queryVariables = useMemo(
-    () => ({ source: squadId, ranking: 'TIME' }),
+    () => ({
+      source: squadId,
+      ranking: 'TIME',
+      supportedTypes: ['share', 'article'],
```
Comment (rebelchris) [inline packages/webapp/pages/squads/[handle]/index.tsx:55]: This one should actually, only accept `share` articles correct?
The regular source should then accept `article` only.

rebelchris (changes requested):
Some comments, looks like a great setup

packages/shared/src/hooks/useFeed.ts
el: ahh, my bad. Fixed

packages/webapp/pages/squads/[handle]/index.tsx
Comment (el) [reply packages/webapp/pages/squads/[handle]/index.tsx:55]: removed 👍

packages/shared/src/hooks/useFeed.ts
```diff
@@ -185,6 +175,19 @@ export default function useFeed<T>(
               index,
             }),
           );
+          const ad = adsQuery.data?.pages[pageIndex];
```
rebelchris, line 178: Also have some concerns here.
I'm not quite sure this will work as the ads might be loading or returning 0.

We want to show the placeholder while they are loading (post + ads)

rebelchris (changes requested):
One more concern

packages/shared/src/hooks/useFeed.ts
el: But this is how it worked before, I just refactored it. Is there a mistake?

el: This is how it was before:
```
          if (adsQuery.data?.pages[pageIndex]) {	
            posts.splice(adSpot, 0, {	
              type: 'ad',	
              ad: adsQuery.data?.pages[pageIndex],	
            });	
          } else {	
            posts.splice(adSpot, 0, {	
              type: 'placeholder',	
            });
          }
```

And this is how it is now:
```
          const ad = adsQuery.data?.pages[pageIndex];
          posts.splice(
            adSpot,
            0,
            ad
              ? {
                  type: 'ad',
                  ad,
                }
              : {
                  type: 'placeholder',
                },
          );

```

el: NVM reverted

rebelchris (approved):
Looks good!
Wasn't sure about the ads, but we've had some issues in the past so rather play it safe there.

PR #1506, opened 2023-01-23
fix: delete squad fix and update squad list after operations (add/edit/delete)

el:
## Changes

Just a quick fix for delete source mutation and updating the sidebar list after an update

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done

### DIFF
--- a/packages/shared/src/components/modals/LockedSquadModal.tsx
+++ b/packages/shared/src/components/modals/LockedSquadModal.tsx
@@ -1,4 +1,5 @@
 import React, { ReactElement } from 'react';
+import { useQuery } from 'react-query';
 import { Button } from '../buttons/Button';
 import { Justify } from '../utilities';
 import { Modal, ModalProps } from './common/Modal';
@@ -6,7 +7,7 @@ import { TextField } from '../fields/TextField';
 import CopyIcon from '../icons/Copy';
 import Alert, { AlertType } from '../widgets/Alert';
 import LinkIcon from '../icons/Link';
-import { deleteSquad, Squad } from '../../graphql/squads';
+import { deleteSquad, getSquad, Squad } from '../../graphql/squads';
 import { useCopyLink } from '../../hooks/useCopyLink';
 import { Image } from '../image/Image';
 import { cloudinary } from '../../lib/image';
@@ -33,16 +34,24 @@ const options: PromptOptions = {
 };
 
 type LockedSquadModalProps = {
-  squad: Squad;
+  initialSquad: Squad;
 } & ModalProps;
+
 function LockedSquadModal({
-  squad,
+  initialSquad,
   onRequestClose,
 }: LockedSquadModalProps): ReactElement {
   const { updateSquads } = useBoot();
   const { showPrompt } = usePrompt();
   const { displayToast } = useToastNotification();
-  const [copying, copyLink] = useCopyLink(() => squad.permalink);
+  const { data: squad, isLoading } = useQuery(
+    [{ type: 'squad_locked', id: initialSquad.id }],
+    ({ queryKey: [{ id }] }) => getSquad(id),
+    { initialData: initialSquad },
+  );
+  const token = squad?.currentMember?.referralToken ?? '';
+  const invitation = `${squad?.permalink}/${token}`;
+  const [copying, copyLink] = useCopyLink(() => invitation);
   const onCopy = () => {
     copyLink();
   };
@@ -85,10 +94,11 @@ function LockedSquadModal({
         <h3 className="font-bold typo-title2">{squad.name}</h3>
         <h4 className="text-theme-label-tertiary">@{squad.handle}</h4>
         <TextField
+          aria-busy={isLoading}
           className={{ container: 'w-full mt-10' }}
           name="permalink"
           inputId="permalink"
-          label={squad.permalink}
+          label={invitation}
           type="url"
           fieldType="tertiary"
           disabled
@@ -101,7 +111,7 @@ function LockedSquadModal({
               data-testid="textfield-action-icon"
             />
           }
-          value={squad.permalink}
+          value={invitation}
           readOnly
         />
         <Alert
--- a/packages/shared/src/components/modals/LockedSquadModal.tsx
+++ b/packages/shared/src/components/modals/LockedSquadModal.tsx
@@ -41,7 +41,7 @@ function LockedSquadModal({
   initialSquad,
   onRequestClose,
 }: LockedSquadModalProps): ReactElement {
-  const { updateSquads } = useBoot();
+  const { deleteSquad: deleteCachedSquad } = useBoot();
   const { showPrompt } = usePrompt();
   const { displayToast } = useToastNotification();
   const { data: squad, isLoading } = useQuery(
@@ -62,7 +62,7 @@ function LockedSquadModal({
     }
     try {
       await deleteSquad(squad.id);
-      await updateSquads();
+      deleteCachedSquad(squad.id);
       onRequestClose(e);
     } catch (error) {
       displayToast('An error occurred.');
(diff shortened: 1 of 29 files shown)


packages/shared/src/components/modals/LockedSquadModal.tsx
```diff
@@ -38,15 +41,23 @@ function LockedSquadModal({
...
   };
 
   const onDeleteSquad = async (e) => {
-    if (await showPrompt(options)) {
+    if (!(await showPrompt(options))) {
+      return;
+    }
+    try {
       await deleteSquad(squad.id);
+      await client.invalidateQueries(BOOT_QUERY_KEY);
```
rebelchris, line 57: Not sure if manipulating the boot query directly is the best idea.
Alternatively we can manually set different data, or refetch the list of active squads to push over it?

rebelchris: @idoshamun / @sshanzel Would love to hear how you think about invalidating it.

rebelchris (commented):
Small concern from my side, but let's see what the rest says first 🙏

packages/shared/src/components/modals/LockedSquadModal.tsx
sshanzel: IIRC invalidating the cache would trigger `refetch`, which I think should be fine, but the concern would be is right after the async call of `deleteSquad`, the `boot` data from `gateway` would already be updated (literally split second after the call). I think locally modifying here can also be an option, but if the current procedure works, then I'm cool with it.

idoshamun: If there's a way to prevent the unnecessary call, it would be better I think. I agree we can simply patch the data ourselves.

el: I have added a new utility function to fetch and update only the squads

packages/shared/src/hooks/useBoot.ts
```diff
@@ -0,0 +1,20 @@
...
+import { Boot } from '../lib/boot';
+
+type UseBoot = {
+  updateSquads: () => Promise<void>;
+};
+
+export const useBoot = (): UseBoot => {
+  const client = useQueryClient();
+  const updateSquads = async () => {
+    const squads = await squadMemberships();
```
rebelchris, line 13: Alternatively as we know which one gets deleted we can manually delete is from the cache.
Thinks upvotes and bookmark use this concept, takes away a API call where we know the result.

deleteCommentCache is also a neat function that uses this approach a bit more.

packages/shared/src/graphql/squads.ts
```diff
@@ -65,6 +65,16 @@ type EditSquadOutput = {
   editSquad: Squad;
 };
 
+type SquadMembershipsOutput = {
+  mySourceMemberships: {
+    sources: {
+      node: {
+        source: Squad;
+      };
+    }[];
```
sshanzel, line 74: I think we have an existing type for this.
```suggestion
    sources: Edge<SquadMember>[];
```

packages/shared/src/graphql/squads.ts
```diff
@@ -189,6 +199,27 @@ export const SQUAD_HANDE_AVAILABILITY_QUERY = gql`
   }
 `;
 
+export const SQUAD_MEMBERSHIPS_QUERY = gql`
+  query Source {
+    mySourceMemberships {
+      sources: edges {
```
sshanzel, line 205: I think `members` is much more accurate.
```suggestion
      members: edges {
```

sshanzel (commented):
Just some minor non-blocking comments aside from Chris' concern 🚢 🚢 🚢 
Other than that, everything looks amazing to me ✨

packages/shared/src/graphql/squads.ts
rebelchris: I think it returns sources, not members @sshanzel it's just the sources this user is part of

packages/shared/src/graphql/squads.ts
rebelchris: Think this also would be:
```
    sources: Edge<Squad>[];
```

packages/shared/src/graphql/squads.ts
sshanzel: I think it is members, where it has a `source` property.

sshanzel: I think this is members, where it has the source property.

sshanzel: Just confirmed, it is indeed members.

image

rebelchris: Oh, not sure I like the name `mySourceMemberships` that much in that case 😅

idoshamun: It returns `SourceMember` objects hence `SourceMemberships`. This was the logic I followed.
Usually the best practice would be to return the complete object which in this case is `SourceMember` rather than a child directly. Because `SourceMember` contains relevant data

packages/shared/src/hooks/useBoot.ts
el: No worries, it makes sense. Done 👍

packages/shared/src/hooks/useBoot.ts
```diff
@@ -0,0 +1,27 @@
...
+
+export const useBoot = (): UseBoot => {
+  const client = useQueryClient();
+  const deleteSquad = (squadId: string) => {
+    const bootData = client.getQueryData<Boot>(BOOT_QUERY_KEY);
+    const squads = bootData.squads?.filter((squad) => squad.id !== squadId);
+    client.setQueryData<Boot>(BOOT_QUERY_KEY, { ...bootData, squads });
+  };
+  const updateSquads = async () => {
+    const squads = await squadMemberships();
```
rebelchris, line 19: For this one, can't we also simply push the squad we just got in return from the create callback?

rebelchris (changes requested):
Amazing, got one more suggestion if its possible.

packages/shared/src/hooks/useBoot.ts
el: What will be the order? Is it going to be at the end?

rebelchris: I'd say where you get the response from `createSquad` (which includes the squad) we can simply call `updateSquads` and pass that single squad right?

Let me know if I overlooked something.

el: No, I meant once we have the squad on our hand, what will be its position on the squad list. I think this is ordered by something:
image

el: ok, looks like it is ordered by squad name

rebelchris: Oh sorry yes by name `.orderBy('s.name', 'ASC')`

packages/shared/src/hooks/useBoot.ts
```diff
@@ -0,0 +1,44 @@
...
+    client.setQueryData<Boot>(BOOT_QUERY_KEY, { ...bootData, squads });
+  };
+  const updateSquad = (squad: Squad) => {
+    const bootData = getBootData();
+    const squads = bootData.squads?.map((bootSquad) =>
+      squad.id !== bootSquad.id ? bootSquad : squad,
+    );
+    client.setQueryData<Boot>(BOOT_QUERY_KEY, { ...bootData, squads });
+  };
+  const updateSquads = async () => {
```
rebelchris, line 33: This one can be removed now right?

rebelchris (changes requested):

packages/shared/src/hooks/useBoot.ts
el: Yeah, I have also deleted the squad membership since it is not used anywhere else

rebelchris (approved):
Nice, love this new hook! 
We can address the order change when we get answer retro-active

PR #1507, opened 2023-01-23
refactor: usage of new query to check handle availability

sshanzel:
- Changed the query being used to check whether the source exists or not.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-956 #done

### DIFF
--- a/packages/shared/src/graphql/squads.ts
+++ b/packages/shared/src/graphql/squads.ts
@@ -183,6 +183,12 @@ export const SQUAD_QUERY = gql`
   ${SOURCE_BASE_FRAGMENT}
 `;
 
+export const SQUAD_HANDE_AVAILABILITY_QUERY = gql`
+  query SourceHandleTaken($handle: String!) {
+    sourceHandleTaken(handle: $handle)
+  }
+`;
+
 export const SQUAD_MEMBERS_QUERY = gql`
   query Source($id: ID!) {
     sourceMembers(first: 5, sourceId: $id) {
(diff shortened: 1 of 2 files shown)


rebelchris (commented):
Actually checking this, the source is called with view permission so what is the exact issue?

packages/shared/src/graphql/squads.ts
```diff
@@ -294,6 +300,17 @@ export const joinSquadInvitation = (
   params: SquadInvitationProps,
 ): Promise<SquadInvitation> =>
   request(`${apiUrl}/graphql`, SQUAD_JOIN_MUTATION, params);
+
+export const checkIsHandleTaken = async (handle: string): Promise<boolean> => {
```
rebelchris, line 304: Do we use this taken tense?
Not quite sure we have other examples wether we prefer positive or negative tense.

(eg: handleAvailable vs handleTaken) or isExistingHandle) 
not blocking, but if we have something similar let's keep it in line.

rebelchris (approved):
Nice work!
Have 2 non blocking request let me know what you think.

packages/shared/src/graphql/squads.ts
sshanzel: Since we have use-cases for `exists` already, let's keep it to that. Will refactor this once the API is ready for merging. Thanks for noting! 🚢

PR #1508, opened 2023-01-24
fix: missing referral token on locked squad for invitation link

sshanzel:
- The referral token was missing on the copy field for invitation link.
- We instead fetch the Squad data on LockedModal open.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-958 #done

### DIFF
--- a/packages/shared/src/components/modals/LockedSquadModal.tsx
+++ b/packages/shared/src/components/modals/LockedSquadModal.tsx
@@ -32,17 +32,17 @@ const options: PromptOptions = {
 };
 
 type LockedSquadModalProps = {
-  squad: Squad;
+  initialSqual: Squad;
 } & ModalProps;
 
 function LockedSquadModal({
-  squad: squadProp,
+  initialSqual,
   onRequestClose,
 }: LockedSquadModalProps): ReactElement {
   const { data: squad, isLoading } = useQuery(
-    [{ type: 'squad_locked', id: squadProp.id }],
+    [{ type: 'squad_locked', id: initialSqual.id }],
     ({ queryKey: [{ id }] }) => getSquad(id),
-    { initialData: squadProp },
+    { initialData: initialSqual },
   );
   const { showPrompt } = usePrompt();
   const token = squad?.currentMember?.referralToken ?? '';
@@ -58,8 +58,6 @@ function LockedSquadModal({
       onRequestClose(e);
     }
   };
-
-  if (isLoading && !squad) return <></>;
 
   return (
     <Modal
@@ -86,6 +84,7 @@ function LockedSquadModal({
         <h3 className="font-bold typo-title2">{squad.name}</h3>
         <h4 className="text-theme-label-tertiary">@{squad.handle}</h4>
         <TextField
+          aria-busy={isLoading}
           className={{ container: 'w-full mt-10' }}
           name="permalink"
           inputId="permalink"
--- a/packages/shared/src/components/modals/LockedSquadModal.tsx
+++ b/packages/shared/src/components/modals/LockedSquadModal.tsx
@@ -1,4 +1,5 @@
 import React, { ReactElement } from 'react';
+import { useQuery } from 'react-query';
 import { Button } from '../buttons/Button';
 import { Justify } from '../utilities';
 import { Modal, ModalProps } from './common/Modal';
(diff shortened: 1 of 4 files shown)


packages/shared/src/components/modals/LockedSquadModal.tsx
```diff
@@ -31,14 +32,21 @@ const options: PromptOptions = {
...
+
 function LockedSquadModal({
-  squad,
+  squadId,
   onRequestClose,
 }: LockedSquadModalProps): ReactElement {
+  const { data: squad, isLoading } = useQuery(
+    [{ type: 'squad_locked', squadId }],
+    ({ queryKey: [{ squadId: id }] }) => getSquad(id),
+  );
```
rebelchris, line 45: Potentially you can set initialData to the squad you received actually.
Might make a slight difference in rendering for slow connections.

Let me know what you think.

packages/shared/src/components/modals/LockedSquadModal.tsx
```diff
@@ -31,14 +32,21 @@ const options: PromptOptions = {
...
   const { showPrompt } = usePrompt();
-  const [copying, copyLink] = useCopyLink(() => squad.permalink);
+  const token = squad?.currentMember?.referralToken ?? '';
+  const invitation = `${squad?.permalink}/${token}`;
```
rebelchris, line 48: This would look weird in the case the currentMember is null right?
Is it even possible?

rebelchris: Asking because you have the conditional after currentMember above.

rebelchris (approved):
Minor suggestion, but also happy with this.

packages/shared/src/components/modals/LockedSquadModal.tsx
sshanzel: I am not fully familiar with how the `currentMember` is implemented at the moment, but being part of data that is coming from an API, I thought it could be nullable. If you think we can remove it, I'll refactor it.

packages/shared/src/components/modals/LockedSquadModal.tsx
sshanzel: Great idea!

packages/shared/src/components/modals/LockedSquadModal.tsx
rebelchris: Yeah, it should always return something if you are part of the squad, based on logged in user.

el (changes requested):
SCR-20230124-nqi


Can you also update the referral link in this screen?

sshanzel:
@el I believe that has been fixed in the change below:
https://github.com/dailydotdev/apps/pull/1508/files#diff-e7d5ed4a3f359f1db4a16b844f8b6daa4f3faa8b4b7eb2a2cc45881b73c681caR66

packages/shared/src/components/modals/LockedSquadModal.tsx
sshanzel: In the case we set the `initialData` for the query, we will have to keep the optional chaining since the passed `squad` from the props will have a null `currentMember`.

packages/shared/src/components/modals/LockedSquadModal.tsx
```diff
@@ -33,12 +34,20 @@ const options: PromptOptions = {
 type LockedSquadModalProps = {
   squad: Squad;
 } & ModalProps;
+
 function LockedSquadModal({
-  squad,
+  squad: squadProp,
```
rebelchris, line 39: Maybe `initialSquad` is a better argument here?

packages/shared/src/components/modals/LockedSquadModal.tsx
rebelchris: Ah good point yes

rebelchris: In that case, should we maybe set loading on the textfield?
That way the user know it's still busy

rebelchris (changes requested):
Awesome!
Got 2 more suggestions let me know what you think

packages/shared/src/components/modals/LockedSquadModal.tsx
sshanzel: I agree. Let's remove the empty fragment for loading and put it in the textfield now. Thank you for the suggestion! 🚢

rebelchris (approved):
Love it! ❤️

el (approved):
I didnt realize it fixes it, looks great ❤️

PR #1509, opened 2023-01-24
fix: hidden options button when moderator

sshanzel:
- We made the delete option inside a context menu which should also be shown when the user is a moderator.
- Also added a test case for it.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done


rebelchris (approved):
Awesome, weird we missed it in the initial review, so thanks for adding the test case 🙏

rebelchris:
@sshanzel Could you double check we maybe removed any other implementation for mods?

sshanzel:
> @sshanzel Could you double check we maybe removed any other implementation for mods?

The other place where we use moderator access is post options. After checking, they should not be affected and still works as intended.

PR #1512, opened 2023-01-24
fix: introduced a combined click handler

rebelchris:
- Introducing a generic onClick handler that applies onClick + onMouseUp (for auxiliary input)
## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-961 #done

### DIFF
--- a/packages/shared/src/lib/click.ts
+++ b/packages/shared/src/lib/click.ts
@@ -1,6 +1,6 @@
 export const combinedClicks = (func) => {
   return {
-    onMouseUp: (event) => event.button === 1 && func,
+    onMouseUp: (event) => event.button === 1 && func(event),
     onClick: func,
   };
 };
--- a/packages/shared/src/lib/click.ts
+++ b/packages/shared/src/lib/click.ts
@@ -1,0 +1,6 @@
+export const combinedClicks = (func) => {
+  return {
+    onMouseUp: (event) => event.button === 1 && func,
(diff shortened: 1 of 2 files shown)


packages/shared/src/lib/click.ts
```diff
@@ -0,0 +1,6 @@
+export const combinedClicks = (func) => {
+  return {
+    onMouseUp: (event) => event.button === 1 && func,
```
rebelchris, line 3: @idoshamun My concern is with this one and options where we use (e) => preventdefault passthrough.

idoshamun: You forgot to call the function. Also I don't see the input types.
Not sure about your comment, can you elaborate please?

```suggestion
    onMouseUp: (event) => event.button === 1 && func(event),
```

rebelchris: Well the issue is often the function takes multiple arguments.
So onClick=(e, something, fuu) => callFunc(e, something, fuu)

idoshamun: If you want, you can use also (event, ...args) => event.button === 1 && func(event, ...args)

idoshamun:
If possible just add types

idoshamun (approved):

PR #1513, opened 2023-01-25
refactor: user mention to be restricted inside squad members only

sshanzel:
- Include `sourceId` in the query to validate whether to restrict mentions to members.

Previews:

There are three accounts with name `Lee Hansel Solevilla` and the recommendation only displayed one since this is the only valid member.
Screenshot 2023-01-24 at 7 29 17 PM

Screenshot 2023-01-24 at 9 48 02 PM

Proof that it still works the same on our regular feeds.

Screenshot 2023-01-24 at 9 48 22 PM

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-928 #done


packages/shared/src/components/FeedLayout.tsx
```diff
@@ -163,6 +163,7 @@ const determineFeedSettings = ({
 
 export default function FeedLayout({
   children,
+  squadId,
```
rebelchris, line 166: Hmm, is there not alternative way of passing data, not sure FeedLayout needs to have any knowledge of squads.
Technically it should be available under the page already right?

rebelchris (changes requested):
It feels a bit "ducktaped" to the wrong component for some reason.
Perhaps we can somehow abstract it into the feed, or even abstract it from the post? (post->source_id?)

Maybe we should also abstract to `source_id` just to future proof public sources as well? (leaving this open for discussion)

packages/shared/src/components/FeedLayout.tsx
sshanzel: We have different kinds of feed which is why I thought it is fitting to include it here since this is feed-related information.

It would be available to the post yes, though we will have to drill the props on multiple components, which using a Context solves.

sshanzel: @idoshamun WDYT?

idoshamun: Agree with Chris. I'm not sure why the feed layout has to change.
The mention feature is available only in the article modal and the article modal has access to source id already. Feed has nothing to do with this information.
Mentions need to be restricted in a general regardless if I'm in a squad feed or my feed but I opened a squad post

sshanzel: > Mentions need to be restricted in a general regardless if I'm in a squad feed or my feed but I opened a squad post

Good point. So we won't have to do double work when we support squad posts in my-feed.

sshanzel:
Now that we have made the BE to be able to handle cases of public Squad (Source), we should not worry anymore whether we are passing a `squadId` or a `sourceId` - meaning, no additional verification is needed.

packages/shared/src/components/modals/NewCommentModal.tsx
```diff
@@ -163,6 +165,7 @@ export default function NewCommentModal({
   };
   const useUserMentionOptions = useUserMention({
     postId: props.post.id,
+    sourceId,
```
rebelchris, line 168: We actually already pass the post right?
Perhaps it's easier to simply pass that if it exists?

rebelchris (changes requested):
Just had one suggestion to avoid extra passthrough

packages/shared/src/components/modals/NewCommentModal.tsx
sshanzel: I think it would be better to minimize the object we're trying to pass as we only need its id (and source now). WDYT?

rebelchris: Agree with that, let's go back a step and attribute bare minimum then :)

sshanzel:
@rebelchris let me know if you think there should be additional changes! 🚢 🚢 🚢

packages/shared/src/hooks/useUserMention.ts
```diff
@@ -79,7 +81,7 @@ export function useUserMention({
         requestMethod(
           `${apiUrl}/graphql`,
           RECOMMEND_MENTIONS_QUERY,
-          { postId, query },
+          { postId, query, sourceId },
```
rebelchris, line 84: I'm not sure if we need to invalidate the "old" key manually maybe?

packages/shared/src/components/modals/NewCommentModal.tsx
```diff
@@ -163,6 +165,7 @@ export default function NewCommentModal({
   };
   const useUserMentionOptions = useUserMention({
     postId: props.post.id,
```
rebelchris, line 167: But here we still have the post as option, so why can't we abstract the `sourceId` from this?

rebelchris (changes requested):
Just 2 more questions

packages/shared/src/hooks/useUserMention.ts
sshanzel: It is a good practice to keep the parameter of the query inside the key itself. This is to ensure cache is accurate as it can get.

packages/shared/src/components/modals/NewCommentModal.tsx
sshanzel: This is legacy code. Let me refactor it quickly.

packages/shared/src/hooks/useUserMention.ts
rebelchris: I know, but you changed a existing query key.
Wondering if it has any impact on the existing apps cache?

packages/shared/src/components/modals/NewCommentModal.tsx
sshanzel: Oh, I thought you initially wanted to pass the whole `post` to the `useUserMention` - let me change the `sourceId` to come from here itself.

sshanzel:
@rebelchris let me know what you think.

rebelchris (approved):
Clean! 🥳

PR #1516, opened 2023-01-25
fix: do not display when the notification request again when dismissed

sshanzel:
- Note: when you dismiss the enable permission widget on comments (and other places), you won't see this too. Goes the same when you dismissed this popup.
- Updated the copy based on the discussion in Slack: https://dailydotdev.slack.com/archives/C01L0QXQJNB/p1674036243515409?thread_ts=1673854401.136469&cid=C01L0QXQJNB

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-960 #done

### DIFF
--- a/packages/shared/src/components/modals/NewSourceModal.tsx
+++ b/packages/shared/src/components/modals/NewSourceModal.tsx
@@ -19,6 +19,8 @@ import { AuthTriggers } from '../../lib/auth';
 import { Modal, ModalProps } from './common/Modal';
 import NotificationsContext from '../../contexts/NotificationsContext';
 import PushNotificationModal from './PushNotificationModal';
+import usePersistentContext from '../../hooks/usePersistentContext';
+import { DISMISS_PERMISSION_BANNER } from '../notifications/EnableNotification';
 
 interface RSS {
   url: string;
@@ -62,6 +64,10 @@ export default function NewSourceModal(props: ModalProps): ReactElement {
   const { user, loginState, showLogin } = useContext(AuthContext);
   const loginTrigger = AuthTriggers.SubmitNewSource;
   const { onRequestClose } = props;
+  const [isDismissed, setIsDismissed] = usePersistentContext(
+    DISMISS_PERMISSION_BANNER,
+    false,
+  );
 
   const failedToScrape = () => {
     setShowContact(true);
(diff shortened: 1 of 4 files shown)


el (approved):
Looks great ❤️

packages/shared/src/components/modals/NewSourceModal.tsx
```diff
@@ -62,6 +64,10 @@ export default function NewSourceModal(props: ModalProps): ReactElement {
   const { user, loginState, showLogin } = useContext(AuthContext);
   const loginTrigger = AuthTriggers.SubmitNewSource;
   const { onRequestClose } = props;
+  const [isDismissed, setIsDismissed] = usePersistentContext(
+    DISMISS_PERMISSION_BANNER,
+    false,
+  );
```
rebelchris, line 70: So does this only work for the `NewSourceModal` permissions?
Or for all permission prompts?

rebelchris (approved):
Looks great, just have one question.
Please check before merging.

packages/shared/src/components/modals/NewSourceModal.tsx
sshanzel: This is for all prompts. Sorry I had merged already, there was a split second where the comment wasn't shown yet.

rebelchris: Ah I see, maybe a hook would be better as there are 2 places it's shared now, could introduce some weird edge-cases.

sshanzel: The behavior will be the same even if we introduce a hook as we instantiate it in two places. I think this should be fine. Introducing a hook would probably look like below:

```
const useDismissedPermission = () => usePersistentContext(
    DISMISS_PERMISSION_BANNER,
    false,
  );
  
// implementation
const [isDismissed, setIsDismissed] = useDismissedPermission();
```

Should I create a hotfix PR for refactor?

rebelchris: Hmm, nah let's leave it.
If we introduce yet another one, let's add a "usePermissions" hook or something.

sshanzel: I agree, something we can generalize permissions 🚢 
Thanks for the consideration! 🚀

PR #1520, opened 2023-01-26
feat: squad article modal

sshanzel:
- Abstracted most of the logic inside an article page/modal in numerous hooks.
- Broken down some big components to make contents reusable.
- Introduction of Squad article modal.
- Most of the things here (function-wise) just came from how we did on `PostModal` - new introductions are just UI-wise. So they technically work the same.

Note: this does not include Squad article page yet.

Preview:

https://user-images.githubusercontent.com/13744167/214835105-164738d9-4d8e-4d57-9e33-f887fd97627d.mov

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-850 #done


packages/shared/src/components/Feed.tsx
```diff
@@ -132,6 +137,7 @@ export default function Feed<T>({
   variables,
   className,
   header,
+  isFeedSquad,
```
rebelchris, line 140: Remember the main feed can contain both normal articles as well as share articles.
Think the evaluation has to be per item?

packages/shared/src/components/cards/SourceButton.tsx
```diff
@@ -21,7 +21,7 @@ export default function SourceButton({
 }: SourceButtonProps): ReactElement {
   return source ? (
     <LinkWithTooltip
-      href={getSourcePermalink(source.id)}
+      href={getSourcePermalink(source.handle)}
```
rebelchris, line 24: Does this still work for existing sources?
Also I just changes this so be aware of that as well :)

packages/shared/src/components/modals/PostModal.tsx
```diff
@@ -32,85 +15,43 @@ interface PostModalProps
...
-
-      parent?.scrollTo?.(0, 0);
-      setPosition('relative');
-    }
-  }, [isLoading]);
+    isDisplayed: props.isOpen,
+  });
 
   return (
-    <Modal
```
rebelchris, line 90: Also don't forget to change the `FeedFilters` modal, it used the "old" large

packages/shared/src/components/modals/PostModal.tsx
```diff
@@ -32,85 +15,43 @@ interface PostModalProps
...
-        postById={postById}
+        post={post}
         onPreviousPost={onPreviousPost}
         onNextPost={onNextPost}
         inlineActions
         className={{
           container: 'post-content',
+          fixedNavigation: {
+            container: modalSizeToClassName[ModalSize.XLarge],
+          },
```
rebelchris, line 48: Can this not be part of the modal guidelines rather?

packages/shared/src/components/modals/SquadPostModal.tsx
```diff
@@ -0,0 +1,226 @@
...
+}
+
+interface SourceInfoProps {
+  source: Source;
+  date?: string;
+  size?: ProfileImageSize;
+  typo?: string;
+}
+
+const SourceInfo = ({
```
rebelchris, line 44: Suggestion to make this a seperate component?

packages/shared/src/components/modals/SquadPostModal.tsx
```diff
@@ -0,0 +1,226 @@
...
+              <LazyImage
+                imgSrc={post.image}
+                imgAlt="Post cover image"
+                ratio="52%"
+                eager
+                fallbackSrc="https://res.cloudinary.com/daily-now/image/upload/f_auto/v1/placeholders/1"
+              />
+            </a>
+          </span>
+          {/* {post.summary && showSumamry && ( */}
```
rebelchris, line 184: Can be removed?

packages/shared/src/components/modals/SquadPostModal.tsx
```diff
@@ -0,0 +1,226 @@
...
+          <PostSummary
+            ref={(el) => {
+              if (!el?.offsetHeight || height !== null) return;
+
+              setHeight(el.offsetHeight);
```
rebelchris, line 189: Maybe make this a const above?

packages/shared/src/components/modals/SquadPostModal.tsx
```diff
@@ -0,0 +1,226 @@
...
+            }}
+            style={{ height: getHeight() }}
+            className="m-4 mt-0 transition-all duration-150 ease-in-out"
+            summary="The awesome-pages plugin allows you to customize how your pages show up the navigation of your MkDocs. It gives you detailed control using a small configuration file directly placed in the relevant directory of your documentation. This package requires Python >=3.5 and Mk Docs version 1.0 or higher."
+          />
+          {/* )} */}
```
rebelchris, line 195: Remove?

packages/shared/src/components/modals/SquadPostModal.tsx
```diff
@@ -0,0 +1,226 @@
...
+    onBookmark: onToggleBookmark,
+    onPreviousPost,
+    onNextPost,
+    onReadArticle,
+    onClose: onRequestClose,
+    onShare,
+    inlineActions: true,
+  };
+
+  const postLinkProps = {
```
rebelchris, line 101: Is this a a tag? (important for tracking purpose)

packages/shared/src/components/modals/SquadPostModal.tsx
```diff
@@ -0,0 +1,226 @@
...
+    href: post.sharedPost.permalink,
+    title: 'Go to article',
+    target: '_blank',
+    rel: 'noopener',
+    onClick: onReadArticle,
+    onMouseUp: (event: React.MouseEvent) =>
+      event.button === 1 && onReadArticle(),
```
rebelchris, line 108: This is now on the Button component, so shouldn't be seperate needed, but good to check?

packages/shared/src/components/post/PostContentContainer.tsx
```diff
@@ -0,0 +1,27 @@
...
+);
+
+function PostContentContainer({
+  hasNavigation,
+  className,
+  children,
+}: PostContentContainerProps): ReactElement {
+  const Wrapper = hasNavigation ? BodyContainer : PageBodyContainer;
+
+  return <Wrapper className={className}>{children}</Wrapper>;
```
rebelchris, line 24: You can also use ConditionalWrapper, it's an existing component, but up to you :)

packages/shared/src/components/profile/UserShortInfo.tsx
```diff
@@ -48,29 +54,41 @@ export function UserShortInfo<Tag extends AnyTag>({
     appendTo: appendTooltipTo || document?.body || 'parent',
     visible: disableTooltip ? false : undefined,
   };
+  const TooltipComponent = ({ children: child }: ElemehtChildren) => (
```
rebelchris, line 57:
```suggestion
  const TooltipComponent = ({ children: child }: ElementChildren) => (
```

rebelchris (changes requested):
Wow amazing work Lee!
This was a big one, high impact though, so did a first round of review.
Let me know once that's done and I'll do hands-on testing locally.

packages/shared/src/components/Feed.tsx
el: I think we should use the post type

packages/shared/src/components/Feed.tsx
```diff
@@ -459,7 +468,7 @@ export default function Feed<T>({
           onHidden={onShareOptionsHidden}
         />
         {selectedPost && (
```
el, line 470: You can check based on `selectedPost.type`

packages/shared/src/components/cards/SourceButton.tsx
```diff
@@ -32,7 +32,7 @@ export default function SourceButton({
           id: source.id,
           name: source.name,
           image: source.image,
-          permalink: getSourcePermalink(source.id),
+          permalink: getSourcePermalink(source.handle),
```
el, line 35: Shouldnt we use squad for the link?
```
export const getSourcePermalink = (id: string): string =>
  `${process.env.NEXT_PUBLIC_WEBAPP_URL}sources/${id}`;

```

packages/shared/src/components/modals/SquadPostModal.tsx
```diff
@@ -0,0 +1,226 @@
...
+  };
+
+  const getHeight = () => {
+    if (height === null) return 'auto';
+
+    return shoudShowSumamry ? height : 0;
```
el, line 114: shouldShowSummary

packages/shared/src/components/modals/SquadPostModal.tsx
```diff
@@ -0,0 +1,226 @@
...
+          date={postDateFormat(post.createdAt)}
+          source={post.source}
+        />
+        <UserShortInfo
+          className="items-center pt-3"
+          imageSize="xxlarge"
+          user={post.author}
+        />
+        <p className="mt-6 typo-title3">
+          They breezed through simple tasks. I threw more complex tasks at them
```
el, line 150: Is this sample text?

packages/shared/src/components/modals/SquadPostModal.tsx
```diff
@@ -0,0 +1,226 @@
...
+          {/* {post.summary && showSumamry && ( */}
+          <PostSummary
+            ref={(el) => {
+              if (!el?.offsetHeight || height !== null) return;
+
+              setHeight(el.offsetHeight);
+            }}
+            style={{ height: getHeight() }}
+            className="m-4 mt-0 transition-all duration-150 ease-in-out"
+            summary="The awesome-pages plugin allows you to customize how your pages show up the navigation of your MkDocs. It gives you detailed control using a small configuration file directly placed in the relevant directory of your documentation. This package requires Python >=3.5 and Mk Docs version 1.0 or higher."
```
el, line 193: sample text?

packages/shared/src/components/post/PostEngagements.tsx
```diff
@@ -0,0 +1,164 @@
...
+import { useShareComment } from '../../hooks/useShareComment';
+import { useUpvoteQuery } from '../../hooks/useUpvoteQuery';
+import { AuthTriggers } from '../../lib/auth';
+import { AuthorOnboarding } from './AuthorOnboarding';
+import { NewComment } from './NewComment';
+import { PostActions, ShareBookmarkProps } from './PostActions';
+import { PostComments } from './PostComments';
+import { PostUpvotesCommentsCount } from './PostUpvotesCommentsCount';
+
+const UpvotedPopupModal = dynamic(
```
el, line 17: Maybe we can use the lazymodal for these?

el (changes requested):
This is a fantastic start @sshanzel Great work

packages/shared/src/components/cards/SourceButton.tsx
rebelchris: We should base it on squad or source indeed.

packages/shared/src/components/Feed.tsx
sshanzel: Oh, in that case, we should then use `post.type`.

packages/shared/src/components/cards/SourceButton.tsx
sshanzel: Yes, existing ones have id as their handle.

packages/shared/src/components/modals/PostModal.tsx
sshanzel: Good catch!

packages/shared/src/components/modals/PostModal.tsx
sshanzel: Not sure about that. This is highly unlikely any other place will use the post navigation. I mean, even on article page it is not used.

packages/shared/src/components/modals/SquadPostModal.tsx
sshanzel: We can't. Because we will be overriding the value based on the initial height. We set it to `0` when it is hidden.

packages/shared/src/components/modals/SquadPostModal.tsx
sshanzel: Ah, correct! Development debugging 😅

packages/shared/src/components/modals/SquadPostModal.tsx
sshanzel: This is the image in the post itself enclosed by an anchor element.

packages/shared/src/components/post/PostContentContainer.tsx
sshanzel: I think this is different since the wrapper here changes not the content itself which the `ConditionalWrapper` handles.

packages/shared/src/components/cards/SourceButton.tsx
sshanzel: Ah, let me refactor this function to accept type as well. Good catch!

packages/shared/src/components/modals/SquadPostModal.tsx
sshanzel: I am plagued by typos. I think I need to get my eyes checked again 😆

packages/shared/src/components/modals/SquadPostModal.tsx
sshanzel: Oh my god, yes 😅

packages/shared/src/components/post/PostEngagements.tsx
sshanzel: I was able to migrate the `UpvotedPopupModal` but for the others, I tried and it produced dependency cycle that is not easy to fix, I think we should create a ticket dedicated to migrating these legacy implementations. WDYT?

packages/shared/src/components/notifications/NotificationItemAvatar.tsx
```diff
@@ -16,10 +17,22 @@ function NotificationItemAvatar({
 }: NotificationAvatar): ReactElement {
   if (type === NotificationAvatarType.Source) {
     return (
-      <SourceButton
-        link={targetUrl}
-        source={{ id: referenceId, handle: referenceId, name, image }}
-      />
+      <LinkWithTooltip
```
sshanzel, line 20: The type will always be missing and is a required property for getting permalink. Making it not suitable for the `SourceButton` anymore so I had to build an element of its own for `Notifications`.

packages/shared/src/components/modals/SquadPostModal.tsx
sshanzel: Should we create a custom anchor element to handle these easily? Like `TrackedAnchorElement` component? WDYT?

packages/shared/src/components/notifications/NotificationItemAvatar.tsx
rebelchris: I added the targetURL there that solves the notification issues :)

packages/shared/src/components/modals/SquadPostModal.tsx
rebelchris: Let me rather explain this one quickly in web meeting 👍

packages/shared/src/components/notifications/NotificationItemAvatar.tsx
sshanzel: Yes, but the type will produce an error now as I have updated the `Source` to have a `type` property. So the build fails.

packages/shared/src/components/Feed.tsx
```diff
@@ -125,6 +129,11 @@ const getStyle = (useList: boolean, spaciness: Spaciness): CSSProperties => {
   return {};
 };
 
+const PostModalMap = {
+  [PostType.Article]: PostModal,
+  [PostType.Share]: SquadPostModal,
+};
```
rebelchris, line 135: Maybe just to be safe make it a Record type?

packages/shared/src/components/cards/SourceButton.tsx
```diff
@@ -16,14 +15,15 @@ interface SourceButtonProps {
 
 export default function SourceButton({
   source,
-  link,
```
rebelchris, line 19: This was introduced for the notification avatar, It should always prefer a direct link (`targetURL`) so think we should keep it.

packages/shared/src/components/modals/PostModal.tsx
rebelchris: Hmm yeah exactly so could we not simply attach the classes directly to that?

packages/shared/src/components/modals/common.tsx
```diff
@@ -44,7 +49,7 @@ type GetComponentProps<T> = T extends
   ? P
   : never;
 
-type ModalsType = typeof modals;
+export type ModalsType = typeof modals;
```
rebelchris, line 52: This shouldn't need an export, as the LazyModal types already have their own?

packages/shared/src/components/modals/common/Modal.tsx
```diff
@@ -58,13 +58,15 @@ const modalKindAndSizeToClassName: Partial<
     [ModalSize.Medium]:
       'mobileL:max-h-[calc(100vh-7.5rem)] max-h-[calc(100vh-2.5rem)] h-auto min-h-[25rem] mt-6',
     [ModalSize.Large]: 'laptop:mt-14 laptop:mb-10',
+    [ModalSize.XLarge]: 'laptop:mt-14 laptop:mb-10',
   },
 };
-const modalSizeToClassName: Record<ModalSize, string> = {
+export const modalSizeToClassName: Record<ModalSize, string> = {
```
rebelchris, line 64: If the page using this already imports from `modal` you can also add it to the main export here?
It will be one less line.

packages/shared/src/components/notifications/NotificationItemAvatar.tsx
rebelchris: I'd really go back to what it was, and optionally create the URL if passed link is empty.
It takes away overhead

packages/shared/src/graphql/sources.ts
```diff
@@ -1,18 +1,64 @@
...
 }
 
-export const getSourcePermalink = (id: string): string =>
-  `${process.env.NEXT_PUBLIC_WEBAPP_URL}sources/${id}`;
+const SOURCE_PATH: Record<SourceType, string> = {
+  machine: 'source',
+  squad: 'squads',
+};
+
+export const getSourcePermalink = (id: string, type: SourceType): string => {
```
rebelchris, line 23: I think if we just default it to `machine` type we support the notification concern right?

rebelchris (changes requested):
wow, amazing work!
Got some small suggestion, let me know what you think.

packages/shared/src/components/Feed.tsx
sshanzel: Totally agree with that.

packages/shared/src/components/notifications/NotificationItemAvatar.tsx
sshanzel: Me as well, the typescript is just not accepting it because the `SourceButton` property for `source` has a `type`. And from the notification, we won't know what `Source` type it is.

packages/shared/src/graphql/sources.ts
sshanzel: Yes, but what if the notification is `squad` - we won't have any reference to say it is `squad`

packages/shared/src/components/modals/common.tsx
sshanzel: We need it in the `LazyModal` to make typescript with `data` in the `useQuery` work. Currently the `data` when accessed returns unknown and make any attempt to use its property, not work.

packages/shared/src/components/notifications/NotificationItemAvatar.tsx
rebelchris: That's why I said we can make the type default to machine for backwards compatability.

packages/shared/src/components/modals/common.tsx
rebelchris: Hmm but what if you use the initial type we already export?

packages/shared/src/graphql/sources.ts
rebelchris: We don't need it if it's unknown right?
Only notifications have that, which is fine as they default to targetURL anyway

packages/shared/src/components/modals/PostModal.tsx
sshanzel: Ah right. I applied it now to components where we can directly to it. Thank you!

packages/shared/src/components/notifications/NotificationItemAvatar.tsx
sshanzel: Backwards compatibility is fine, the issue is the future notifications, the source type will always be `machine` if we do that since we don't have idea if it was from a notification source type of squad.
```
export interface Source {
  __typename?: string;
  id?: string;
  name: string;
  image: string;
  handle: string;
  type: SourceType;
  permalink: string;
}
```

Our notification avatar
```
export interface NotificationAvatar {
  type: NotificationAvatarType;
  image: string;
  name: string;
  targetUrl: string;
  referenceId: string;
}
```

packages/shared/src/components/modals/common.tsx
sshanzel: Yes, that is what it is using initially. Had to make it accept a type parameter.

packages/shared/src/components/notifications/NotificationItemAvatar.tsx
rebelchris: Yes, the notification never needs to know all the source information.
It always has a absolute URL which takes priority.

I'm not sure I'm following the issue.

sshanzel: Yes, we are using the targetUrl. The problem becomes typescript. It is not accepting the `source` parameter because we don't have a `type` property. It is the build itself that rejects.

rebelchris: Alright so we need to fix that, rather then just removing it.
Make it optional when it comes from notifications maybe?

sshanzel: There is one option to keep the same component, it is if we are always sure permalink is going to have a value from the API. In that case, we will instead make it receive a custom object than `Source` object itself. WDYT?

packages/shared/src/components/modals/common/Modal.tsx
sshanzel: Right. Have updated it now.

packages/shared/src/components/notifications/NotificationItemAvatar.tsx
sshanzel: Not sure if it is possible to change props based on a single implementation. We will have to explicitly change its prop.

sshanzel:
@rebelchris sorry for doing another refactor, I have now made it ready for the article page to use the same components as well. It should be ready though on all ends.

packages/shared/src/components/MainFeedLayout.tsx
```diff
@@ -234,7 +234,9 @@ export default function MainFeedLayout({
     <LayoutHeader className="flex-col">
       {alerts?.filter && (
         <CreateMyFeedButton
-          action={() => onInitializeOnboarding(onFeedPageChanged)}
+          action={() =>
+            onInitializeOnboarding(() => onFeedPageChanged(MainFeedPage.MyFeed))
+          }
```
rebelchris, line 239: Why did we need to change something here?
Noting this so we can make sure to test it properly.

packages/shared/src/components/ShareOptionsMenu.tsx
```diff
@@ -66,7 +66,7 @@ export default function ShareOptionsMenu({
     {
       icon: <MenuIcon Icon={ShareIcon} />,
       text: 'Share article via...',
-      action: onShare,
+      action: () => onShare(post),
```
rebelchris, line 69: Did onShare not already always take the post?

packages/shared/src/components/cards/SourceButton.tsx
```diff
@@ -1,13 +1,12 @@
 import React, { CSSProperties, ReactElement } from 'react';
 import { TooltipPosition } from '../tooltips/BaseTooltipContainer';
 import { LinkWithTooltip } from '../tooltips/LinkWithTooltip';
-import { getSourcePermalink, Source } from '../../graphql/sources';
```
rebelchris, line 4: is this function `getSourcePermalink` used anywhere else?
Else we can clean it up perhaps.

packages/shared/src/components/post/SquadPostContent.tsx
```diff
@@ -0,0 +1,165 @@
...
+    onShare: onSharePost,
+    onClose,
+    inlineActions,
+  };
+
+  const getHeight = () => {
+    if (height === null) return 'auto';
+
+    return shoudShowSummary ? height : 0;
+  };
```
rebelchris, line 55: Might have to memoize this as it get's called from the FE.

packages/shared/src/components/post/SquadPostContent.tsx
```diff
@@ -0,0 +1,165 @@
...
+                rel="noopener"
+                onClick={onReadArticle}
+                className="block overflow-hidden w-70 rounded-2xl cursor-pointer"
+              >
+                <LazyImage
+                  imgSrc={post.image}
+                  imgAlt="Post cover image"
+                  ratio="52%"
+                  eager
+                  fallbackSrc="https://res.cloudinary.com/daily-now/image/upload/f_auto/v1/placeholders/1"
```
rebelchris, line 129: Did we not have a const export for this somewhere?

packages/shared/src/components/widgets/PostUsersHighlights.tsx
```diff
@@ -122,7 +122,7 @@ export function PostUsersHighlights({ post }: PostAuthorProps): ReactElement {
     <WidgetContainer className="flex flex-col">
       <UserHighlight
         {...source}
-        permalink={getSourcePermalink(source.id)}
+        permalink={getSourcePermalink(source.handle, source.type)}
```
rebelchris, line 125: Is this not available under `source.permalink` here?

packages/shared/src/contexts/OnboardingContext.tsx
```diff
@@ -62,6 +63,23 @@ export const OnboardingContextProvider = ({
   const [hasTriedOnboarding, setHasTriedOnboarding, hasOnboardingLoaded] =
     usePersistentContext<boolean>(LOGGED_USER_ONBOARDING, !alerts.filter);
 
+  const onInitializeOnboarding = (onFinish: () => void) => {
+    if (onFinish) {
+      onFeedPageChanged.current = onFinish;
+    }
+    setIsOnboarding(true);
+  };
```
rebelchris, line 71: Since we changed this, it's highly important we test the important flows 🙏

packages/webapp/pages/posts/[id].tsx
```diff
@@ -50,30 +45,27 @@ export interface Props {
...
   const [position, setPosition] =
     useState<CSSProperties['position']>('relative');
-  const { tokenRefreshed } = useContext(AuthContext);
   const router = useRouter();
   const { isFallback } = router;
 
-  if (!isFallback && !id) {
-    return <Custom404 />;
-  }
+  if (!isFallback && !id) return <Custom404 />;
```
Comment (rebelchris) [inline packages/webapp/pages/posts/[id].tsx:63]: Just be aware this will clash with my changes, but shouldn't be breaking to much.

packages/webapp/pages/posts/[id].tsx
```diff
@@ -50,30 +45,27 @@ export interface Props {
...
-    isFetched,
-  } = useQuery<PostData>(
-    ['post', id],
-    () => request(`${apiUrl}/graphql`, POST_BY_ID_QUERY, { id }),
-    { initialData: postData, enabled: !!id && tokenRefreshed },
-  );
+  const { post, isLoading } = usePostById({
+    id,
+    options: { initialData: postData },
+  });
```
Comment (rebelchris) [inline packages/webapp/pages/posts/[id].tsx:68]: This might make a difference, compared to what I just did.
Perhaps you can check that out.

packages/webapp/pages/posts/[id].tsx
```diff
@@ -94,26 +86,32 @@ const PostPage = ({ id, postData }: Props): ReactElement => {
   });
 
   useEffect(() => {
-    window.addEventListener('popstate', () => {
+    globalThis.window?.addEventListener('popstate', () => {
       router.reload();
     });
   }, []);
```
Comment (rebelchris) [inline packages/webapp/pages/posts/[id].tsx:92]: We actually need a cleanup here.
Did we not introduce a `useWindow` kind of hook?

Comment (rebelchris) [reply packages/webapp/pages/posts/[id].tsx:92]: Or useEventListener, not sure what we called it.

rebelchris (changes requested):
Awesome work Lee,
Almost there it feels like.

With this one, make sure you get at least 2 approvals before merging 🙏

packages/shared/src/components/ShareOptionsMenu.tsx
sshanzel: It was not. From the type it came from, it was empty, but it probably worked as the post came from the parent component where it was triggered and handled. Now that we are reusing the type, it is better to be explicit with the function than handling it from the parent's implementation.

packages/shared/src/components/cards/SourceButton.tsx
sshanzel: Right, let's clean all implementations.

packages/shared/src/components/post/SquadPostContent.tsx
sshanzel: Let me check quickly.

sshanzel: After checking, it was not yet organized and implemented on numerous places. Will update all soon.

packages/webapp/pages/posts/[id].tsx
Comment (sshanzel) [reply packages/webapp/pages/posts/[id].tsx:80]: I think we should that PR to this branch as there are a lot of changes involved here so it will be easier to fix merge issues. But let's see what fits.

sshanzel:
@el @rebelchris let me know what you think! This is currently deployed at `https://preview.app.daily.dev/`.

packages/webapp/pages/posts/[id].tsx
Comment (rebelchris) [reply packages/webapp/pages/posts/[id].tsx:80]: Yeah let's do this one first, and i'll adjust from merged :)

packages/shared/src/components/ShareOptionsMenu.tsx
rebelchris: Oh good find then! Lucky this didn't break randomly

packages/webapp/pages/posts/[id].tsx
```diff
@@ -94,31 +103,34 @@ const PostPage = ({ id, postData }: Props): ReactElement => {
...
```
Comment (rebelchris) [inline packages/webapp/pages/posts/[id].tsx:109]: this can be removed right?

rebelchris (changes requested):
One small issue, then good from my side.

rebelchris:
@sshanzel There is a `loading` text + layout shift.
See video:

https://user-images.githubusercontent.com/554874/215085930-5d0fe233-7d30-4f89-9a3e-ed152b2cfd31.mp4

sshanzel:
@rebelchris we don't have yet a loading skeleton in Sketch so I did put some placeholder for the moment. Let me check with design team.

For the layout shift, that definitely needs to be fixed.

sshanzel:
@rebelchris the layout shift should be fixed now 🚢

el:
F37E79A8-C3CD-46A1-9FF0-B0FEE614FEDF


Maybe a nicer loading screen?

el:
BC1F7BB7-EF2B-49CD-B29C-937E5B9A6E42

missing images for the articles

packages/shared/src/components/post/PostModalActions.tsx
```diff
@@ -89,22 +89,24 @@ export function PostModalActions({
...
-          onClick={onReadArticle}
+      {onReadArticle && (
+        <SimpleTooltip
+          placement="bottom"
+          content="Read article"
+          disabled={!inlineActions}
         >
-          {!inlineActions && 'Read article'}
-        </Button>
-      </SimpleTooltip>
```
el, line 107: Read article button does not open the post in a new tab

rebelchris:
Regular modal on mobile seems to be broken now.

image

packages/shared/src/components/post/PostSourceInfo.tsx
```diff
@@ -0,0 +1,36 @@
...
+import SourceButton from '../cards/SourceButton';
+import { ProfileImageSize } from '../ProfilePicture';
+
+interface SourceInfoProps {
+  source: Source;
+  date?: string;
+  size?: ProfileImageSize;
+  typo?: string;
+}
+
```
el, line 14: Clicking the link arrow on the shared post modal header makes it “download” 
7C6D5C53-2DDE-4488-A524-662E7C66F520
BF9268C4-AEC1-4117-A1A4-A5F1073F7FF4

el (changes requested):
Have some change requests

packages/shared/src/components/post/PostModalActions.tsx
sshanzel: I think it should be based on your preference in the settings.

sshanzel:
> Maybe a nicer loading screen?

There is no design yet. We will implement once available.

sshanzel:
> Regular modal on mobile seems to be broken now.

@rebelchris should be fixed by now.

packages/shared/src/components/post/PostModalActions.tsx
el: No, it does open it on regular posts, just not on shared post

el: Looks like you have fixed it ❤️

packages/shared/src/components/post/PostModalActions.tsx
```diff
@@ -89,22 +91,24 @@ export function PostModalActions({
...
+          <Button
+            className={inlineActions ? 'btn-tertiary' : 'btn-secondary'}
+            tag="a"
+            href={post.permalink}
```
el, line 103:
```suggestion
            href={post.sharedPost.permalink}
```

el (changes requested):

packages/shared/src/components/post/PostModalActions.tsx
sshanzel: Good catch! Let me add a fallback to the original post permalink as this is used in regular post modal as well.

rebelchris:
@sshanzel / @el The loading skeleton we can simply make right?
It's just mimicking the placeholders based on which elements we have. you can even look at the post one and simply remove the sidebar for now.

I think it's important to do it before this get's merged.

rebelchris (approved):
Nice one, code wise happy with this 🥳
I'd suggest we fix the placeholder loading though, we can simply mock what it needs to render based on the article one.

sshanzel:
@rebelchris fixed the user hover tooltip issue and applied the loading placeholder we have and removed the widgets part as suggested. Thank you for the review! 🚢

rebelchris (approved):
Amazing @sshanzel good to go from my side 🙏
But let's wait for 1 more review :)

packages/shared/src/components/post/FixedPostNavigation.tsx
```diff
@@ -22,15 +22,16 @@ function FixedPostNavigation({
...
+          'laptop:left-[unset] max-w-full',
           className?.container,
         ),
         actions: className?.actions,
       }}
       onPreviousPost={onPreviousPost}
       onNextPost={onNextPost}
     >
-      <div className="overflow-hidden flex-col flex-1 ml-2">
+      <div className="hidden tablet:flex overflow-hidden flex-col flex-1 ml-2">
```
el, line 34: The buttons should be right alligned
image

sshanzel: Yep, didn't push the fix yet as I was working on something else. Should be committed now. Thank you!

packages/shared/src/components/post/PostContentContainer.tsx
```diff
@@ -0,0 +1,27 @@
+import React, { ReactElement, ReactNode } from 'react';
```
el, line 1: There seems to be an issue with the shared post modal (no image and incorrect link)
SCR-20230130-ew0

packages/shared/src/components/post/PostContentContainer.tsx
```diff
@@ -0,0 +1,27 @@
...
+import classed from '../../lib/classed';
+
+interface PostContentContainerProps {
```
el, line 4: I dont think we should have the option to hide articles from the squad here (you can create a task and skip this if it is not related to the modal)

SCR-20230130-exc

el (changes requested):
Just some minor requests

packages/shared/src/components/post/PostContentContainer.tsx
el: The link should be to the article, not the post page imo

packages/shared/src/components/post/PostContentContainer.tsx
rebelchris: I already discussed this and product does not want to change any behaviour here at closed beta.

packages/shared/src/components/post/PostContentContainer.tsx
sshanzel: I think this is a question suited for the product team.

el (approved):
Wohoo. I think it is ready to be merged 😍

PR #1528, opened 2023-01-27
fix: share post loading

rebelchris:
- Changed loading state for posts.
- If Server side succeeds (normal posts) we simply show that as initialData
- If Server side fails (auth posts, squads) we fallback to client side rendering
- Modified how we show the data merge

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-979 #done

### DIFF
--- a/packages/shared/src/hooks/usePostById.ts
+++ b/packages/shared/src/hooks/usePostById.ts
@@ -32,9 +32,11 @@ const usePostById = ({
     { ...options, enabled: !!id && tokenRefreshed },
   );
 
+  const post = postById || (options?.initialData as PostData);
+
   return useMemo(
     () => ({
-      post: postById?.post,
+      post: post?.post,
       isLoading: isLoading || !isFetched || isFetchingNextPage,
     }),
     [postById, isLoading],
--- a/packages/webapp/pages/posts/[id].tsx
+++ b/packages/webapp/pages/posts/[id].tsx
@@ -1,6 +1,7 @@
 import React, {
   CSSProperties,
   ReactElement,
+  useCallback,
   useState,
 } from 'react';
 import { useRouter } from 'next/router';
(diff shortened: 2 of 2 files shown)


packages/webapp/pages/posts/[id].tsx
```diff
@@ -47,42 +47,49 @@ export const getSeoDescription = (post: Post): string => {
...
     ['post', id],
     () => request(`${apiUrl}/graphql`, POST_BY_ID_QUERY, { id }),
-    { initialData: postData, enabled: !!id && tokenRefreshed },
+    { initialData, enabled: !!id && tokenRefreshed, retry: false },
   );
 
+  // Need this as sometimes client side might still be loading, but we already have initial data.
+  const post = fetchedPost ?? initialData;
+
+  if ((!isFallback && !id) || (isFetched && !post?.post.title)) {
```
Comment (el) [inline packages/webapp/pages/posts/[id].tsx:77]: Why do we check the title?

el (approved):
Just a minor question, otherwise looks good

packages/webapp/pages/posts/[id].tsx
Comment (rebelchris) [reply packages/webapp/pages/posts/[id].tsx:77]: Because we always have the ID from the param, which could technically be a false one.

idoshamun (approved):
This looks great! Another idea for improvement, maybe we can return 403 in case the post is private and the request cannot be authenticated then the flow will be as follows:
* Server side sends requests to API
* If data returns, render the initial page and then fetch again (like now)
* If 403, render a skeleton then fetch again from client and decide accordingly
* If 404, render custom 404.
Just a suggestion, I'm also happy with current implementation

rebelchris:
@idoshamun With that I'm afraid the user query (authenticated) will fire first and fail.
But let me add it as a task so we can look at that later on regardless.

idoshamun:
@rebelchris ok, not sure I understand why but let's keep it as a discussion topic once needed

sshanzel (approved):
The fix looks good to me! If you need help with merging this, let me know, I'd be glad to help.

rebelchris:
@sshanzel Just double check this one please 🙏

packages/shared/src/hooks/usePostById.ts
```diff
@@ -32,9 +32,11 @@ const usePostById = ({
     { ...options, enabled: !!id && tokenRefreshed },
   );
 
+  const post = postById || (options?.initialData as PostData);
+
   return useMemo(
     () => ({
-      post: postById?.post,
+      post: post?.post,
```
sshanzel, line 39: I think no changes are required here since the `options` was passed and the `postById` will get the value of it on the first render.

sshanzel (approved):
LGTM! Let's go!! 🚢🚢🚢

PR #1531, opened 2023-01-28
fix: locked squad access

rebelchris:
- Show 404 if squad is still locked
- Also decided to pre show it so we don't waste other queries

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-993 #done

### DIFF
--- a/packages/webapp/pages/squads/[handle]/index.tsx
+++ b/packages/webapp/pages/squads/[handle]/index.tsx
@@ -40,6 +40,11 @@ const SquadPage = ({ handle }: SourcePageProps): ReactElement => {
       enabled: !!handle,
     },
   );
+
+  if (!squad && isLoading) return <SquadLoading />;
+
+  if (isFallback || !squad.active) return <Unauthorized />;
+
   const squadId = squad?.id;
 
   const { data: squadMembers } = useQuery<SquadMember[]>(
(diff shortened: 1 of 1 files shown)


packages/webapp/pages/squads/[handle]/index.tsx
```diff
@@ -40,6 +40,11 @@ const SquadPage = ({ handle }: SourcePageProps): ReactElement => {
       enabled: !!handle,
     },
   );
+
+  if (!squad && isLoading) return <SquadLoading />;
+
+  if (isFallback || !squad.active) return <Unauthorized />;
+
```
Comment (sshanzel) [inline packages/webapp/pages/squads/[handle]/index.tsx:47]: I think the correct way to go is to set the `enabled` property to `true` until the conditions above are reversely met. From the docs it is ideal to not return early when there are hooks declared under the condition.
> Instead, always use Hooks at the top level of your React function, before any early returns. 

Reference: https://reactjs.org/docs/hooks-rules.html#only-call-hooks-at-the-top-level

Comment (rebelchris) [reply packages/webapp/pages/squads/[handle]/index.tsx:47]: Ah I see, well we can modify the order.
I assumed we wouldn't need it, but yes good point.

rebelchris:
@sshanzel Moved it away, that should work :)

sshanzel (approved):
Nice work!

PR #1540, opened 2023-01-31
fix: share modal content

sshanzel:
- As discussed, when the user tries to share a Squad post, the link should be directed towards the original article.
- I have also refactored the fragments to be contained in a single `fragments` file as I was having some deps issues.
- I have also checked previous implementations of `Fragments` and the word `Fragment` itself should not be part of the name.
- Note: still waiting for Sab's response on which title should we display in the reading history. Title of the actual article or the user's commentary but feel free to review.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-999 #done

### DIFF
--- a/packages/shared/src/components/modals/ShareModal.tsx
+++ b/packages/shared/src/components/modals/ShareModal.tsx
@@ -31,9 +31,12 @@ export default function ShareModal({
   ...props
 }: ShareModalProps): ReactElement {
   const isComment = !!comment;
+  const permalink = isComment
+    ? post?.commentsPermalink
+    : post?.sharedPost?.commentsPermalink ?? post?.commentsPermalink;
   const link = isComment
-    ? `${post?.commentsPermalink}${getCommentHash(comment.id)}`
-    : post?.commentsPermalink;
+    ? `${permalink}${getCommentHash(comment.id)}`
+    : permalink;
   const { trackEvent } = useContext(AnalyticsContext);
   const [, copyUrl] = useCopyLink(() => link);
 
@@ -63,13 +66,15 @@ export default function ShareModal({
     return () => baseTrackingEvent('close share');
   }, []);
 
+  const article = post?.sharedPost ?? post;
+
   return (
     <Modal size={Modal.Size.Small} kind={Modal.Kind.FlexibleCenter} {...props}>
       <Modal.Header title={isComment ? 'Share comment' : 'Share article'} />
       {!isComment && (
         <PostItemCard
           className="mt-2"
-          postItem={{ post }}
+          postItem={{ post: article }}
           showButtons={false}
           clickable={false}
         />
@@ -96,7 +101,7 @@ export default function ShareModal({
         />
         <p className="py-2.5 font-bold typo-callout">Share via</p>
         <SocialShare
-          post={post}
+          post={article}
           comment={comment}
           origin={origin}
           columns={columns}
--- a/packages/shared/src/components/modals/ShareModal.tsx
+++ b/packages/shared/src/components/modals/ShareModal.tsx
@@ -74,7 +74,7 @@ export default function ShareModal({
       {!isComment && (
         <PostItemCard
           className="mt-2"
-          postItem={{ post: article }}
+          postItem={{ post }}
           showButtons={false}
           clickable={false}
         />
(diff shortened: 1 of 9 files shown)


packages/shared/src/components/modals/ShareModal.tsx
```diff
@@ -96,7 +101,7 @@ export default function ShareModal({
         />
         <p className="py-2.5 font-bold typo-callout">Share via</p>
         <SocialShare
-          post={post}
+          post={article}
```
rebelchris, line 104: Why do we moved to article here?
Would sharedPost not be more descriptive?

packages/shared/src/components/post/PostItemCard.tsx
```diff
@@ -56,7 +57,7 @@ export default function PostItemCard({
         )}
       >
         <Image
-          src={post.image}
+          src={article.image}
```
rebelchris, line 60: This one is also a bit confusing to read what it actually is

rebelchris (commented):
Small remarks

packages/shared/src/components/modals/ShareModal.tsx
sshanzel: Sab has decided when you are sharing an article from the feed, it must be pointed to the original article. Relevant discussion: https://dailydotdev.slack.com/archives/C03QZBF46JC/p1675073160566779?thread_ts=1675068777.979039&cid=C03QZBF46JC

The variable article can either be the original post as the fallback value. `sharedPost` is reserved for posts inside a Squad and this can be either of the two.

packages/shared/src/components/post/PostItemCard.tsx
sshanzel: When we say, we usually mean the actual post. And `article` is both true for `post` and `sharedPost` as the value can be either of the two.

packages/shared/src/components/modals/ShareModal.tsx
el: I am with @sshanzel on this one. Article makes more sense since it would mean the post for the regular posts, and sharedpost for the squad posts. Which is the article for both cases.

packages/shared/src/components/modals/ShareModal.tsx
```diff
@@ -31,9 +31,12 @@ export default function ShareModal({
   ...props
 }: ShareModalProps): ReactElement {
   const isComment = !!comment;
+  const permalink = isComment
+    ? post?.commentsPermalink
+    : post?.sharedPost?.commentsPermalink ?? post?.commentsPermalink;
```
el, line 36: Why do you need this one? ?? `post?.commentsPermalink`

packages/shared/src/graphql/fragments.ts
```diff
@@ -0,0 +1,59 @@
+import { gql } from 'graphql-request';
```
el, line 1: This is nice

packages/shared/src/components/modals/ShareModal.tsx
sshanzel: When `isComment` is true, it would mean the user clicked the comment's share button. I think the intent is most likely to share the comment itself not the article this time.

packages/shared/src/graphql/sources.ts
```diff
@@ -17,37 +17,6 @@ export interface Source {
 
 export type SourceData = { source: Source };
```
el, line 19: @sshanzel this refactor is mostly unrelated to the PR. I recommend separating big tasks like this in a separate PR in the future. It is ok for this PR though.

el (approved):
Just some minor comments, but good to go :shipit:

packages/shared/src/graphql/sources.ts
sshanzel: I was having some dependencies cycle issue which is why I had to move some of them. Then I realized why not move everything 😅 

Will try to separate the PRs next time. Thanks for the review!

packages/shared/src/components/modals/ShareModal.tsx
el: Did you want to have something like this?
```
  const permalink = isComment || !post?.sharedPost
    ? post?.commentsPermalink
    : post?.sharedPost?.commentsPermalink;
```

sshanzel: I think the current setup is fine. That case should only be for `isComment` to true. When false, we always want to return the `sharedPost` link when available (to prioritize the actual post link).

PR #1542, opened 2023-01-31
fix: comment replies avatar z-index

sshanzel:
- Fixes the issue where the timeline is showing on top of comment replies avatars.
- Also fixes the alignment of timeline as I noticed it not being centered of the author avatar.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-997 #done

### DIFF
--- a/packages/shared/src/components/comments/CommentBox.tsx
+++ b/packages/shared/src/components/comments/CommentBox.tsx
@@ -71,7 +71,7 @@ function CommentBox({
           user={comment.author}
           tooltip={{ appendTo: appendTooltipTo }}
         >
-          <ProfileImageLink user={comment.author} />
+          <ProfileImageLink className="z-1" user={comment.author} />
         </ProfileTooltip>
         <div className="flex flex-col ml-3 typo-callout">
           <FlexRow>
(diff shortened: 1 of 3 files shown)


packages/shared/src/components/comments/CommentBox.tsx
```diff
@@ -71,7 +71,7 @@ function CommentBox({
           user={comment.author}
           tooltip={{ appendTo: appendTooltipTo }}
         >
-          <ProfileImageLink user={comment.author} />
+          <ProfileImageLink className="z-1" user={comment.author} />
```
rebelchris, line 74: This you could even always give it z-1 right? Instead of condtional bht up to you 👍

rebelchris (approved):

packages/shared/src/components/comments/CommentBox.tsx
sshanzel: Good point. I was just thinking it might affect different areas and wanted to keep positioning related properties to be given by the implementing component. Let's see what other think of this too 🚢

PR #1547, opened 2023-01-31
fix: join squad race condition

sshanzel:
- Don't wait for the boot to refresh its data but locally add the squad to boot manually.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1005 #done

### DIFF
--- a/packages/webapp/pages/squads/[handle]/[token].tsx
+++ b/packages/webapp/pages/squads/[handle]/[token].tsx
@@ -94,7 +94,7 @@ const SquadReferral = ({ token, handle }: SquadReferralProps): ReactElement => {
 
     return showLogin('join squad', {
       referral: member.user.id,
-      onLoginSuccess: onJoinClick,
+      onLoginSuccess: onJoinSquad,
       onRegistrationSuccess: onJoinSquad,
     });
   };
--- a/packages/webapp/pages/squads/[handle]/[token].tsx
+++ b/packages/webapp/pages/squads/[handle]/[token].tsx
@@ -56,6 +56,7 @@ const SquadReferral = ({ token, handle }: SquadReferralProps): ReactElement => {
     ['squad_referral', token, loggedUser?.id],
     () => getSquadInvitation(token),
     {
+      keepPreviousData: true,
       enabled: !!token,
       onSuccess: (response) => {
         if (!loggedUser) return null;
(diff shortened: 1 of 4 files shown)


packages/webapp/pages/squads/[handle]/[token].tsx
```diff
@@ -74,10 +76,17 @@ const SquadReferral = ({ token, handle }: SquadReferralProps): ReactElement => {
     },
   );
   const sourceId = member?.source?.id;
-  const squadsUrl = `/squads/${sourceId}`;
   const { mutateAsync: onJoinSquad } = useMutation(
     () => joinSquadInvitation({ sourceId, token }),
-    { onSuccess: () => router.replace(squadsUrl) },
+    {
```
Comment (el) [inline packages/webapp/pages/squads/[handle]/[token].tsx:81]: Can we also handle the error here? What if the token is invalid?

el (approved):

packages/webapp/pages/squads/[handle]/[token].tsx
Comment (el) [reply packages/webapp/pages/squads/[handle]/[token].tsx:81]: I tried this link https://preview2.app.daily.dev/squads/google_club/faketoken I didnt see any errors?

Comment (sshanzel) [reply packages/webapp/pages/squads/[handle]/[token].tsx:81]: Oh. Let me check that one then. Thanks for the quick link.

packages/webapp/pages/squads/[handle]/[token].tsx
```diff
@@ -74,17 +77,25 @@ const SquadReferral = ({ token, handle }: SquadReferralProps): ReactElement => {
...
+      onSuccess: (data) => {
+        const squad = squads.find(({ id }) => id === data.id);
+        if (squad) return router.replace(squad.permalink);
```
Comment (idoshamun) [inline packages/webapp/pages/squads/[handle]/[token].tsx:85]: Is this really needed? Not sure...

Comment (sshanzel) [reply packages/webapp/pages/squads/[handle]/[token].tsx:85]: So you won't have to add locally. It would be duplicated in the list if we add it and if in your boot it exists already.

packages/webapp/pages/squads/[handle]/[token].tsx
Comment (sshanzel) [reply packages/webapp/pages/squads/[handle]/[token].tsx:82]: I have updated the function call to fetch the token details and it is now handled correctly. Good catch!

PR #1557, opened 2023-02-02
fix: flicker when switching squad modals

sshanzel:

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done


packages/shared/src/components/modals/NewSquadModal.tsx
```diff
@@ -77,6 +81,16 @@ function NewSquadModal({
...
   };
+
+  if (isIntroShown) {
+    return (
+      <SquadsBetaModal
+        onRequestClose={handleClose}
+        onNext={() => setIsIntroShown(false)}
+      />
+    );
+  }
```
rebelchris, line 92: Perhaps it's better to simply put the body as a conditional render on the modal?

packages/shared/src/hooks/useCreateSquadModal.ts
```diff
@@ -34,9 +34,10 @@ export const useCreateSquadModal = ({
     });
   const openSquadBetaModal = () =>
     openModal({
-      type: LazyModal.BetaSquad,
+      type: LazyModal.NewSquad,
```
rebelchris, line 37: We should also remove the betaSquad from the useModal props.

rebelchris (changes requested):
I suggest we change it around a bit.

packages/shared/src/hooks/useCreateSquadModal.ts
sshanzel: If it is not going to be used anywhere, then I agree, we should remove it.

el: You can just delete it since it was only used here anyways. And if that changes in the future for some reason, we can re-add it

sshanzel: Awesome, I will do that then. Thanks for the confirmation.

packages/shared/src/components/modals/NewSquadModal.tsx
```diff
@@ -77,6 +81,16 @@ function NewSquadModal({
     if (shouldQuit) onRequestClose();
     return null;
   };
+
+  if (isIntroShown) {
```
el, line 85: showBetaModal

el (changes requested):
If you can remove the `LazyModal.BetaSquad`

packages/shared/src/components/modals/NewSquadModal.tsx
el: If it is going to be refactored, it can be converted to a step with a conditional to use it or not.

packages/shared/src/components/modals/NewSquadModal.tsx
```diff
@@ -77,6 +84,37 @@ function NewSquadModal({
     if (shouldQuit) onRequestClose();
     return null;
   };
+
+  if (shouldShowBetaModal) {
+    return (
+      <Modal isOpen kind={Modal.Kind.FixedCenter} size={Modal.Size.Small}>
```
rebelchris, line 90: Just double check how this behaves with bg click?

rebelchris (approved):
Just one question, but should be fine.

sshanzel:
@el let me know what you think 🥳

sshanzel:
@el let me know what you think 🥳

packages/shared/src/components/modals/NewSquadModal.tsx
```diff
@@ -77,6 +84,42 @@ function NewSquadModal({
...
+
+  if (shouldShowBetaModal) {
+    return (
+      <Modal
+        isOpen
+        kind={Modal.Kind.FixedCenter}
+        size={Modal.Size.Small}
+        onRequestClose={handleClose}
+      >
+        <Modal.Body>
```
el, line 96: I think you should keep `Modal.Body` in a separate component here and use the same Modal wrapper. Then, you can either use it as a step or render it conditionally.

sshanzel: Done. I think it would be easier to have a separate component since the modal type of the main component is different from this one.

el (approved):
Great work, :shipit:

PR #1566, opened 2023-02-02
fix: stale post when navigating from reading history

sshanzel:
- For some reason, the post is always considered `stale`.
- This is being fixed by refetching the data since the sole purpose of that is to check the cache if it is stale.
- Since we have decided to bring back refetching, I have fixed it here.
- With that in mind, I have also ensured to fix the issue we have that repeatedly loads NOT_FOUND article by disabling the fetch itself when the `initialData` is not found.
- The `initialData` is always null when an error has occurred (including NOT_FOUND), so I believe that is a right fix.
 
## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1037 #done

### DIFF
--- a/packages/shared/src/hooks/usePostById.ts
+++ b/packages/shared/src/hooks/usePostById.ts
@@ -4,6 +4,7 @@ import { QueryObserverOptions, useQuery } from 'react-query';
 import { graphqlUrl } from '../lib/config';
 import { useAuthContext } from '../contexts/AuthContext';
 import { Post, PostData, POST_BY_ID_QUERY } from '../graphql/posts';
+import { isNullOrUndefined } from '../lib/func';
 
 interface UsePostByIdProps {
   id: string;
@@ -22,6 +23,7 @@ const usePostById = ({
   options = {},
 }: UsePostByIdProps): UsePostById => {
   const { tokenRefreshed } = useAuthContext();
+  const isEnabled = !!id && tokenRefreshed;
   const {
     data: postById,
     isLoading,
@@ -29,7 +31,12 @@ const usePostById = ({
   } = useQuery<PostData>(
     ['post', id],
     () => request(graphqlUrl, POST_BY_ID_QUERY, { id }),
-    { ...options, enabled: !!id && tokenRefreshed },
+    {
+      ...options,
+      enabled: isNullOrUndefined(options.enabled)
+        ? isEnabled
+        : isEnabled && options.enabled,
+    },
   );
 
   const post = postById || (options?.initialData as PostData);
@@ -39,7 +46,7 @@ const usePostById = ({
       post: post?.post,
       isLoading: isLoading || !isFetched || isFetchingNextPage,
     }),
-    [postById, isLoading],
+    [id, postById, isLoading, options, isFetchingNextPage],
   );
 };
 
--- a/packages/shared/src/hooks/usePostById.ts
+++ b/packages/shared/src/hooks/usePostById.ts
@@ -4,7 +4,6 @@ import { QueryObserverOptions, useQuery } from 'react-query';
 import { graphqlUrl } from '../lib/config';
 import { useAuthContext } from '../contexts/AuthContext';
 import { Post, PostData, POST_BY_ID_QUERY } from '../graphql/posts';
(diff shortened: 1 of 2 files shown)


idoshamun:
@sshanzel this PR will break the support in shared posts. We need to find another way

sshanzel:
> @sshanzel this PR will break the support in shared posts. We need to find another way

Oh, cool. Let me think of something else to support it.

packages/webapp/pages/posts/[id].tsx
```diff
@@ -100,8 +99,8 @@ const PostPage = ({ id, initialData }: Props): ReactElement => {
     scrollProperty: 'scrollY',
   });
 
-  if (!initialData && (isFallback || isLoading)) {
```
Comment (idoshamun) [inline packages/webapp/pages/posts/[id].tsx:103]: I think that the solution to the problem should be this:
`if (!initialData && (isFallback || !isFetched))`

Comment (sshanzel) [reply packages/webapp/pages/posts/[id].tsx:103]: Yep, worked like a charm! Let me push it quickly!

packages/shared/src/hooks/usePostById.ts
```diff
@@ -22,24 +24,31 @@ const usePostById = ({
...
     isFetched,
   } = useQuery<PostData>(
     ['post', id],
     () => request(graphqlUrl, POST_BY_ID_QUERY, { id }),
-    { ...options, enabled: !!id && tokenRefreshed },
+    {
+      ...options,
+      enabled: isNullOrUndefined(options.enabled)
+        ? isEnabled
+        : isEnabled && options.enabled,
```
idoshamun, line 39: I think we can leave it as it was right?

sshanzel: Yes, was just thinking in the future, but let's remove it as it is not too complex to add in the future.

packages/webapp/pages/posts/[id].tsx
```diff
@@ -127,7 +126,7 @@ const PostPage = ({ id, initialData }: Props): ReactElement => {
         position={position}
         post={post}
         isFallback={isFallback}
-        isLoading={isLoading}
+        isLoading={isLoading || !isFetched}
```
Comment (idoshamun) [inline packages/webapp/pages/posts/[id].tsx:129]: I think this is the right one

```suggestion
        isLoading={!isFetched}
```

packages/shared/src/hooks/usePostById.ts
```diff
@@ -22,24 +24,31 @@ const usePostById = ({
...
 
   return useMemo(
     () => ({
       post: post?.post,
-      isLoading: isLoading || !isFetched || isFetchingNextPage,
+      isFetched,
+      isLoading: isLoading || isFetchingNextPage,
     }),
-    [postById, isLoading],
+    [id, postById, isLoading, options, isFetched, isFetchingNextPage],
```
idoshamun, line 47: Why do we need all of that in the deps list?

sshanzel: I think they are necessary as those values are changing, and we decide some things based on them. This is to keep the values accurate.

idoshamun: How come? I see inside the use memo you use only `post, isFetched, isLoading, isFetchingNextPage`

sshanzel: Yes, but the `post` is based on the `options`, for when it changes, the value of `post` might change too.

packages/webapp/pages/posts/[id].tsx
```diff
@@ -115,7 +114,7 @@ const PostPage = ({ id, initialData }: Props): ReactElement => {
   };
   const customNavigation = navigation[post?.type] ?? navigation.article;
 
-  if (!Content && (!isFallback || !isLoading)) return <Custom404 />;
+  if (!Content && (!isFallback || !isFetched)) return <Custom404 />;
```
Comment (idoshamun) [inline packages/webapp/pages/posts/[id].tsx:117]:
```suggestion
  if (!Content && (!isFallback || isFetched)) return <Custom404 />;
```

idoshamun (approved):

PR #1567, opened 2023-02-03
refactor: share options to include squad

sshanzel:
- Refactored the `ProfileLink` component to ask for the href only instead of the whole user object which only needs the permalink.
- Conditions when to the squad options:
  - When it is not a comment (as the user's intent is to share the comment and shared items in Squad are posts only.
  - When the post is public.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-977 #done

### DIFF
--- a/packages/extension/src/companion/useRawBackgroundRequest.ts
+++ b/packages/extension/src/companion/useRawBackgroundRequest.ts
@@ -1,8 +1,9 @@
+import { EmptyObjectLiteral } from '@dailydotdev/shared/src/lib/kratos';
 import { useEffect } from 'react';
 import { browser } from 'webextension-polyfill-ts';
 
 export const useRawBackgroundRequest = (
-  command: (params: unknown) => void,
+  command: (params: EmptyObjectLiteral) => void,
 ): void => {
   useEffect(() => {
     const handler = ({ key, ...args }) => {
--- a/packages/shared/src/components/widgets/SocialShare.tsx
+++ b/packages/shared/src/components/widgets/SocialShare.tsx
@@ -123,7 +123,7 @@ export const SocialShare = ({
         squads.map((squad) => (
           <button
             type="button"
-            className="flex flex-col"
+            className="flex flex-col items-center"
             key={squad.id}
             onClick={() => onShareToSquad(squad)}
           >
--- a/packages/shared/src/components/widgets/SocialShare.tsx
+++ b/packages/shared/src/components/widgets/SocialShare.tsx
@@ -9,7 +9,7 @@ import {
   getWhatsappShareLink,
   ShareProvider,
 } from '../../lib/share';
-import { SocialShareIcon } from './SocialShareIcon';
+import { ShareText, SocialShareIcon } from './SocialShareIcon';
 import { Post } from '../../graphql/posts';
 import MailIcon from '../icons/Mail';
 import TwitterIcon from '../icons/Twitter';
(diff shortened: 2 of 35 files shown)


packages/extension/src/companion/useRawBackgroundRequest.ts
```diff
@@ -1,8 +1,9 @@
+import { EmptyObjectLiteral } from '@dailydotdev/shared/src/lib/kratos';
 import { useEffect } from 'react';
 import { browser } from 'webextension-polyfill-ts';
 
 export const useRawBackgroundRequest = (
-  command: (params: unknown) => void,
+  command: (params: EmptyObjectLiteral) => void,
```
sshanzel, line 6: I just noticed an error while having the extension projects opened.

rebelchris (changes requested):
No remarks, but wait with merging please as there is an ongoing discussion here:
https://dailydotdev.slack.com/archives/C03QZBF46JC/p1675365375994899

I'll request changes for now, until we get a resolution sorry for that.

sshanzel:
No worries, we can wait til decision is made 🚢

rebelchris (approved):
Alright good to go 🙏

packages/shared/src/components/widgets/SocialShare.tsx
```diff
@@ -102,6 +118,19 @@ export const SocialShare = ({
         onClick={() => trackClick(ShareProvider.Email)}
         label="Email"
       />
+      {!isComment &&
```
el, line 121: There is an alignment issue
SCR-20230203-of5

packages/shared/src/components/widgets/SocialShare.tsx
```diff
@@ -102,6 +118,19 @@ export const SocialShare = ({
...
+        !post.private &&
+        squads.map((squad) => (
+          <button
+            type="button"
+            className="flex flex-col items-center"
+            key={squad.id}
+            onClick={() => onShareToSquad(squad)}
+          >
+            <SourceProfilePicture source={squad} />
+            <ShareText className="mt-2">@{squad.handle}</ShareText>
```
el, line 131: Shouldnt we use the name?

el (changes requested):

packages/shared/src/components/widgets/SocialShare.tsx
sshanzel: Names are often so much longer and can be the same with other Squads. I think going with the handle has its benefits. WDYT?

packages/shared/src/components/widgets/SocialShare.tsx
sshanzel: Oh, good catch! Let me check this quickly! 🚢

sshanzel: Context behind the solution: the WhatsApp icon has the correct size but the label below made its total width expand, making it unequal to the email share, which led to making it look like the Squads weren't in the same alignment. I turned it into a grid to make them have the same total width sizes across the board.

Normal size:
image

Mobile:
image

Where there are long labels:
image

Let me know what you think!

sshanzel:
I have also refactored the order based on the discussion in Slack.
https://dailydotdev.slack.com/archives/C03QZBF46JC/p1675571154715349?thread_ts=1675525512.721859&cid=C03QZBF46JC

packages/shared/src/components/widgets/SocialShare.tsx
el: Yeap, looks great, nice work. Just make sure it also works well on narrower phones.

el (approved):
Great work ❤️

PR #1569, opened 2023-02-03
fix: squad form displaying error message

sshanzel:
- Used try catch to check if the request failed and display the error message.
- Ensured to lowercase the handle sent to the server.
- Did some small cleanups.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1015 #done

### DIFF
--- a/packages/shared/src/components/squads/Details.tsx
+++ b/packages/shared/src/components/squads/Details.tsx
@@ -1,5 +1,7 @@
 import React, { FormEvent, ReactElement, useState } from 'react';
 import classNames from 'classnames';
+import { ClientError } from 'graphql-request';
+import { useMutation } from 'react-query';
 import { Button } from '../buttons/Button';
 import { TextField } from '../fields/TextField';
 import UserIcon from '../icons/User';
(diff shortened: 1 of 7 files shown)


packages/shared/src/components/squads/SteppedComment.tsx
```diff
@@ -1,25 +1,51 @@
...
+        const newSquad = await createSquad({ ...form, ...data });
+        if (!newSquad) return;
+        onCreate(newSquad);
+        nextStep(e);
+      } catch (err) {
+        const clientError = err as ClientError;
+        const message = clientError?.response?.errors?.[0]?.message;
+        if (!message) return;
+        const error = JSON.parse(message);
+        displayToast(error?.handle ?? DEFAULT_ERROR);
```
rebelchris, line 42: Should we not display the "basic" handle error in the hint field?
I think that makes more sense, the fallback could always be toast though for anything besides that.

rebelchris (changes requested):
Thanks for this, just one remark I think we can improve the flow on.

packages/shared/src/components/squads/SteppedComment.tsx
sshanzel: That is all the way back to the first step. I think we should have a more comprehensive handle checker mutation so we won't have to wait the end of the process before knowing there could be invalid formats.

I tried a handle "asdas asd" and it produced an error in the end but not on the handle check. WDYT?

rebelchris: Hmm yeah it should be inline on the first step from the user perspective right?
I mean makes more sense if you preliminary know something is wrong, so maybe we should try and fix that handle check indeed.

Happy to help or debug it together if you need me 🙏

sshanzel: Amazing. I will try to introduce some changes to the handle checker to see if it will be valid.

Thanks! I'll let you know if I hit a bumpy road 🚗

rebelchris (approved):
Amazing! That will do the trick 🥳

packages/shared/src/components/squads/Details.tsx
```diff
@@ -13,39 +15,70 @@ import { formToJson } from '../../lib/form';
...
-      onSubmit(e, formJson);
+      return null;
     }
+
+    const data = await getFormData(formJson, imageChanged);
+
+    if (!createMode) return onSubmit(e, data);
+
+    const handleExists = await onValidateHandle(formJson.handle);
+    setHandleHint(handleExists ? 'The handle is already exists' : null);
```
el, line 77: The handle already exists

packages/shared/src/components/squads/Details.tsx
```diff
@@ -13,39 +15,70 @@ import { formToJson } from '../../lib/form';
...
+
+    if (!handleExists) return onSubmit(e, data);
+
+    return null;
```
el, line 81: I think this is unnecessary

el (changes requested):
Small change request

packages/shared/src/components/squads/Details.tsx
sshanzel: The linter is not accepting the function not having a return statement for having conditional returns.

packages/shared/src/components/squads/Details.tsx
sshanzel: I don't understand. Can you elaborate?

sshanzel: Oh, this was copy-pasted. Nice catch! Didn't know this existed before.

PR #1573, opened 2023-02-03
feat: disable event pings during development

capJavert:

Talking to Chris I noticed that during development there are lot of periodical or navigation based pings that will always fail with `503` because analytics service is not available in development. Specifically it disables these pings:
- active ping inside useFeed hook
- sendEvents and sendBeacon pings from useAnalyticsQueue hook (replaced by console.log)

All of this results in a lot of console.error logs and failed network requests (in chrome network tab) which IMO clutter those places when looking for stuff that actually matters.

I am aware that this could also be solved with filters in console/network tab but those are not so easy to persist across tabs/urls etc.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done

### DIFF
--- a/packages/shared/src/hooks/analytics/useAnalyticsQueue.ts
+++ b/packages/shared/src/hooks/analytics/useAnalyticsQueue.ts
@@ -32,6 +32,13 @@ export default function useAnalyticsQueue({
   const enabledRef = useRef(false);
   const { mutateAsync: sendEvents } = useMutation(
     async (events: AnalyticsEvent[]) => {
+      if (process.env.NODE_ENV === 'development') {
+        // eslint-disable-next-line no-console
+        console.log('useAnalyticsQueue.sendEvents', { events });
+
+        return;
+      }
+
       const res = await fetchMethod(ANALYTICS_ENDPOINT, {
         method: 'POST',
         body: JSON.stringify({ events }),
@@ -78,6 +85,14 @@ export default function useAnalyticsQueue({
           const blob = new Blob([JSON.stringify({ events })], {
             type: 'application/json',
           });
+
+          if (process.env.NODE_ENV === 'development') {
+            // eslint-disable-next-line no-console
+            console.log('useAnalyticsQueue.sendBeacon', { events });
+
+            return;
+          }
+
           if (backgroundMethod) {
             backgroundMethod?.({
               url: ANALYTICS_ENDPOINT,
--- a/packages/shared/src/hooks/useFeed.ts
+++ b/packages/shared/src/hooks/useFeed.ts
@@ -139,7 +139,8 @@ export default function useFeed<T>(
     },
     {
       getNextPageParam: () => Date.now(),
-      enabled: query && tokenRefreshed,
+      enabled:
+        process.env.NODE_ENV !== 'development' && query && tokenRefreshed,
       refetchOnMount: false,
       refetchOnReconnect: false,
       refetchOnWindowFocus: false,


packages/shared/src/hooks/analytics/useAnalyticsQueue.ts
```diff
@@ -32,6 +32,13 @@ export default function useAnalyticsQueue({
   const enabledRef = useRef(false);
   const { mutateAsync: sendEvents } = useMutation(
     async (events: AnalyticsEvent[]) => {
+      if (process.env.NODE_ENV === 'development') {
+        // eslint-disable-next-line no-console
+        console.log('useAnalyticsQueue.sendEvents', { events });
```
capJavert, line 37: I don't know if logging is the best way here but I wanted to leave some trace in development because event debugging should not be affected.

packages/shared/src/hooks/analytics/useAnalyticsQueue.ts
```diff
@@ -78,6 +85,14 @@ export default function useAnalyticsQueue({
           const blob = new Blob([JSON.stringify({ events })], {
             type: 'application/json',
           });
+
+          if (process.env.NODE_ENV === 'development') {
+            // eslint-disable-next-line no-console
+            console.log('useAnalyticsQueue.sendBeacon', { events });
```
capJavert, line 91: same as https://github.com/dailydotdev/apps/pull/1573/files#r1095795079

packages/shared/src/hooks/analytics/useAnalyticsQueue.ts
capJavert: Also thought about replacing `useAnalyticsQueue` hook with "empty" one in `development` but after looking at the rest of the code some of the functions as sendBeacon are required to function on other places.

packages/shared/src/hooks/analytics/useAnalyticsQueue.ts
```diff
@@ -32,6 +32,13 @@ export default function useAnalyticsQueue({
   const enabledRef = useRef(false);
   const { mutateAsync: sendEvents } = useMutation(
     async (events: AnalyticsEvent[]) => {
+      if (process.env.NODE_ENV === 'development') {
```
el, line 35: I dont think we should have development-related checks on production. Maybe we can mock the backend if we want to get rid of the network errors.
@rebelchris @sshanzel @idoshamun

idoshamun: Yes I agree this might be a better solution. Pretty sure there's an "echo" service we can spin up to plot all the requests it receives. It will help us debug a lot of analytics issues

capJavert: I agree that the mocked service would be a better solution long term 👍 

In terms of that service I don't know if you have something specific in mind but as `pino` is already used in api project it has a great support for logging network requests so maybe a quick endpoint for `/api/e` would be a quick win to improve debugging and just reduce the clutter of all those request errors https://github.com/pinojs/pino-http#pinohttpstdserializersreq

I am up to adding this but my concern would be where would be the best place to add `/api/e` inside daily-api to avoid interfering with production deployment. 

FYI @idoshamun If there is someone better to handle this instead of me that is of course an option.

idoshamun: Probably the `daily-api` would be the right project. @rebelchris should prioritize :)

rebelchris: Hmm yeah we can look at this at a later stage.
Not to dismiss it for now, but we have a couple urgent issues to deal with first.

Let's put this on the backlog and we can add it for a sprint with some more breathing room?

capJavert: @idoshamun already did it after all of the gateway changes 👌

capJavert:
Solved https://github.com/dailydotdev/apps/pull/1573#discussion_r1113302520

PR #1574, opened 2023-02-03
fix: missing queryFn console error in useQuery usages

capJavert:

While trying to cleanup console clutter during development in #1573 I noticed another error that happens passively while clicking through the app `Error: Missing queryFn` coming from `useQuery` usages.

This happens in pattern when using useQuery as state hook. Valid fix should be to add queryFn that just loads up the current data of the query (if exists). I looked at the rest of the code and I can see that on some other places it is using the same pattern (eg. `packages/shared/src/components/notifications/Toast.tsx`) so I think these were just missed accidentally.

As far as I can see this does not break anything because all of the hooks fallback and handle when data is present or `undefined`.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done

### DIFF
--- a/packages/shared/src/hooks/useInAppNotification.ts
+++ b/packages/shared/src/hooks/useInAppNotification.ts
@@ -41,6 +41,7 @@ export const useInAppNotification = (): UseInAppNotification => {
   const { incrementUnreadCount } = useNotificationContext();
   const { data: notification } = useQuery<InAppNotification>(
     IN_APP_NOTIFICATION_KEY,
+    () => client.getQueryData<InAppNotification>(IN_APP_NOTIFICATION_KEY),
   );
   const hasNotification = (): boolean =>
     !!client.getQueryData(IN_APP_NOTIFICATION_KEY);
--- a/packages/shared/src/hooks/useLazyModal.ts
+++ b/packages/shared/src/hooks/useLazyModal.ts
@@ -15,7 +15,9 @@ export function useLazyModal<
   T extends LazyModalType<K> = LazyModalType<K>,
 >(): UseLazyModal<K, T> {
   const client = useQueryClient();
-  const { data: modal } = useQuery<T>(MODAL_KEY);
+  const { data: modal } = useQuery<T>(MODAL_KEY, () =>
+    client.getQueryData<T>(MODAL_KEY),
+  );
   const openModal = (data: T) => client.setQueryData(MODAL_KEY, data);
   const closeModal = () => client.setQueryData(MODAL_KEY, null);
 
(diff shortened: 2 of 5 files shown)


capJavert:
Not sure why the tests fail, I don't think it is connected to the changes here, noticed it also in #1573

rebelchris (approved):
Tested this and seems to work fine, I think we might have just missed these.
@sshanzel Do you know if this had any specific reason?

sshanzel:
> Tested this and seems to work fine, I think we might have just missed these. @sshanzel Do you know if this had any specific reason?

I believe what Ante said is correct, in longer terms, we are trying to use `react-query` as a state to be accessed across the app in numerous areas. The implementation works as the only issue would be the initial state which is not a concern in those areas. Probably an anti-pattern but, yep, ensuring the function call is definitely safer by doing as what in this PR has displayed.

This is why I was suggesting the introduction of some state-management libraries like zustand/redux. The current implementations are a bit unpredictable.

rebelchris:
@sshanzel I still think React query will work for now as state manager.
Maybe we should rather introduce some stricter linting.

But fully open to look at another state management solution in the meantime with future aspect in mind.
So thanks for bringing that to attention.

For time being how do you feel about this PR?

sshanzel:
@rebelchris yep, I agree, current ones are just fine, probably when we get more complex requirements then we might actually need it.

Oh, sorry, I forgot to click the "Submit review" button 😅 
Everything looks amazing to me though! 🚢
Great work here @capJavert

sshanzel (approved):
Thanks for the cleanup! 🙏

capJavert:
> > Tested this and seems to work fine, I think we might have just missed these. @sshanzel Do you know if this had any specific reason?
> 
> I believe what Ante said is correct, in longer terms, we are trying to use `react-query` as a state to be accessed across the app in numerous areas. The implementation works as the only issue would be the initial state which is not a concern in those areas. Probably an anti-pattern. Though I agree ensuring the function call is definitely safer by doing as what in this PR has displayed.
> 
> This is why in the past I was suggesting the introduction of some state-management libraries like zustand/redux. The current implementations are a bit unpredictable.

One additional thing that is maybe good to note when using `useQuery` as state manager.

The `query.stale` and `query.cache` options could be set to `Infinity`, this would mitigate any mismatches in data for empty query functions because those would never get invalidated and would behave more as true global state. 

Again this is not relevant for cases in this PR but in general.

sshanzel:
> The query.stale and query.cache options could be set to Infinity, this would mitigate any mismatches in data for empty query functions because those would never get invalidated and would behave more as true global state.

Oh, that is good to know. Thank you for the tip!

PR #1575, opened 2023-02-03
feat: disable next/prev buttons on post modal

el:
## Changes
first post:
SCR-20230203-nwn

last post:
SCR-20230203-nwu

only post:
SCR-20230203-nx3

middle post:
SCR-20230203-nxd

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done

### DIFF
--- a/packages/shared/src/components/post/PostNavigation.tsx
+++ b/packages/shared/src/components/post/PostNavigation.tsx
@@ -4,6 +4,7 @@ import { Button } from '../buttons/Button';
 import { SimpleTooltip } from '../tooltips/SimpleTooltip';
 import ArrowIcon from '../icons/Arrow';
 import { PostModalActions, PostModalActionsProps } from './PostModalActions';
+import { PostPosition } from '../../hooks/usePostModalNavigation';
 
 type PostActions = Pick<
   PostModalActionsProps,
@@ -22,6 +23,7 @@ export interface PostNavigationClassName {
 }
 
 export interface PostNavigationProps extends PostActions {
+  postPosition?: PostPosition;
   onPreviousPost?: () => unknown;
   onNextPost?: () => unknown;
   className?: PostNavigationClassName;
@@ -29,6 +31,7 @@ export interface PostNavigationProps extends PostActions {
 }
 
 function PostNavigation({
+  postPosition,
   onPreviousPost,
   onNextPost,
   className = {},
@@ -49,6 +52,9 @@ function PostNavigation({
             className="-rotate-90 btn-secondary"
             icon={<ArrowIcon />}
             onClick={onPreviousPost}
+            disabled={[PostPosition.First, PostPosition.Only].includes(
+              postPosition,
+            )}
           />
         </SimpleTooltip>
       )}
@@ -58,6 +64,9 @@ function PostNavigation({
             className="rotate-90 btn-secondary"
             icon={<ArrowIcon />}
             onClick={onNextPost}
+            disabled={[PostPosition.Last, PostPosition.Only].includes(
+              postPosition,
+            )}
           />
         </SimpleTooltip>
       )}
--- a/packages/shared/src/hooks/usePostModalNavigation.ts
+++ b/packages/shared/src/hooks/usePostModalNavigation.ts
@@ -7,7 +7,15 @@ import { FeedItem, PostItem, UpdateFeedPost } from './useFeed';
 import { useKeyboardNavigation } from './useKeyboardNavigation';
 import { Origin } from '../lib/analytics';
 
+export enum PostPosition {
+  First = 'first',
+  Only = 'only',
+  Middle = 'middle',
+  Last = 'last',
+}
+
 interface UsePostModalNavigation {
+  postPosition: PostPosition;
   onPrevious: () => void;
   onNext: () => Promise<void>;
   onOpenModal: (index: number, fromPopState?: boolean) => void;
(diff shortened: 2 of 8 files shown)


packages/shared/src/hooks/usePostModalNavigation.ts
```diff
@@ -7,7 +7,15 @@ import { FeedItem, PostItem, UpdateFeedPost } from './useFeed';
 import { useKeyboardNavigation } from './useKeyboardNavigation';
 import { Origin } from '../lib/analytics';
 
+export enum PostPosition {
+  First = 'first',
+  Only = 'only',
+  Middle = 'middle',
```
rebelchris, line 13: Technically this one is not used right?

packages/shared/src/hooks/usePostModalNavigation.ts
```diff
@@ -106,8 +114,18 @@ export const usePostModalNavigation = (
...
 
+  const getPostPosition = () => {
+    const isPost = (item: FeedItem) => item.type === 'post';
+    const firstPost = items.findIndex(isPost);
+    const lastPost = [...items].reverse().findIndex(isPost);
+    const isLast = items.length - lastPost - 1 === openedPostIndex;
+    if (firstPost === openedPostIndex)
+      return isLast ? PostPosition.Only : PostPosition.First;
+    return isLast ? PostPosition.Last : PostPosition.Middle;
+  };
```
rebelchris, line 125: Not sure if this will become heavy on people who scroll a lot, seems like it would calculate every new item, and with a long list it might be expensive?

packages/shared/src/hooks/usePostModalNavigation.ts
```diff
@@ -106,8 +114,18 @@ export const usePostModalNavigation = (
     };
   }, [items]);
 
+  const getPostPosition = () => {
+    const isPost = (item: FeedItem) => item.type === 'post';
```
rebelchris, line 118: Maybe we should enum this too?
We already have `SourceType` so maybe `PostType`?

Although thinking about it, does it matter if it's post only, the sharedpost should also have the same behaviour right?

rebelchris (changes requested):
Nice!
Have two smaller concerns maybe we can verify locally?

packages/shared/src/hooks/usePostModalNavigation.ts
el: Shared post is still a post. This is not checking the post.type. This is checking feeditem.type

packages/shared/src/hooks/usePostModalNavigation.ts
el: You are right. I think this is O(n²)
I will try to improve it

packages/shared/src/hooks/usePostModalNavigation.ts
rebelchris: Got it 🙏

packages/shared/src/hooks/usePostModalNavigation.ts
el: Yeap, but we should still keep it imo

packages/shared/src/components/post/PostNavigation.tsx
```diff
@@ -58,6 +64,9 @@ function PostNavigation({
             className="rotate-90 btn-secondary"
             icon={<ArrowIcon />}
             onClick={onNextPost}
+            disabled={[PostPosition.Last, PostPosition.Only].includes(
+              postPosition,
+            )}
```
sshanzel, line 69: Making this arrow (pointing towards the right) disabled must also consider whether it is the last item and whether you can still fetch more pages as these are paginated values. I think the variable is named `canFetchMore`.

sshanzel (changes requested):
One comment with regard to deciding whether to disable the right button.

When the right arrow is disabled, we should also disable the keyboard navigation going left and right.

In the current implementation, the right arrow is disabled when you reach item 30 - but since the `onNext` function is built to fetch the next page when it is the `last` item displayed and then when you hit the keyboard right arrow, the navigation would work, which makes it look like it fixed the issue as it fetches the next page.

You will notice in the video below the arrow got disabled and I am unable to navigate to the next page, then suddenly the modal went to the next page, it is because I clicked the keyboard's right arrow key.
https://user-images.githubusercontent.com/13744167/216778743-37ec496c-90ca-4632-afb8-b76ac554d405.mov

packages/shared/src/hooks/usePostModalNavigation.ts
```diff
@@ -7,7 +7,15 @@ import { FeedItem, PostItem, UpdateFeedPost } from './useFeed';
 import { useKeyboardNavigation } from './useKeyboardNavigation';
 import { Origin } from '../lib/analytics';
 
+export enum PostPosition {
+  First = 'first',
+  Only = 'only',
```
sshanzel, line 12: I think the only should not be necessary.

el: It happens when there is only 1 post

packages/shared/src/components/post/PostNavigation.tsx
el: Makes sense

rebelchris (approved):
Alright happy with this for now, we can actually get the last, based on wether there is more to fetch I think?

packages/shared/src/hooks/usePostModalNavigation.ts
sshanzel: Ah, right. I am incorrect. Let me resolve this comment 🚢

sshanzel:
This was not a release blocker, I believe it would have been better to hold this off instead of merging. There is a good chance the process to manage the disabling of buttons at the same time with keyboard events will change from the proposed change here.

PR #1581, opened 2023-02-06
fix: new comment latency updating local data

sshanzel:
- The issue from the new comment latency is not actually the request time.
- After further investigation, we don't show the comments list component when the `post.numComments` is at 0.
- It happens to update (after some time, thinking it is after mutation latency) due to our real-time updates through `useSubscription`. That is why it only happens on a post's very first comment.
- To fix the issue, I have created a utility function to update the post cache.
- Also created a function to generate a post query key as it is used in a lot of areas but didn't want to update everything in this PR.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1021 #done

### DIFF
--- a/packages/shared/src/hooks/usePostById.ts
+++ b/packages/shared/src/hooks/usePostById.ts
@@ -1,6 +1,11 @@
 import request from 'graphql-request';
 import { useMemo } from 'react';
-import { QueryObserverOptions, useQuery } from 'react-query';
+import {
+  QueryClient,
+  QueryKey,
+  QueryObserverOptions,
+  useQuery,
+} from 'react-query';
 import { graphqlUrl } from '../lib/config';
 import { useAuthContext } from '../contexts/AuthContext';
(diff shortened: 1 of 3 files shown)


rebelchris (approved):
Nice one, we can eventually look at consolidating these into generic functions for any type of updates.
Also really need to look into proper query key functions 😂

el (approved):
LGTM :shipit:

sshanzel:
> Nice one, we can eventually look at consolidating these into generic functions for any type of updates.

Great point! This will become easy I think once we have the function to generate correct query keys. Will be a tough battle but for sure worth it!

> Also really need to look into proper query key functions 😂

Hopefully, we get the time soon with all the amazing things coming up 🤣 - but the team is getting bigger, and we are now having more and more resources ❤️ 
Probably we can start with what TkDodo had in their team. We probably talk about it in the near web team meeting.

PR #1583, opened 2023-02-06
fix: squad feed not updating after sharing new post

sshanzel:
- Since I was testing out some theory of the `refetch` could be the issue (but turns out to be not), I took the liberty of reverting the `refetch` as we always tend to prefer it enabled.
- I also tested out and it shows the loading skeleton as opposed to the current behavior of flickering 404 page.
- I checked to access it on incognito and no flicker whatsoever.
- I updated the `useFeed` hook to accept the options for enabling refetch on mount. This is unwanted for regular feeds, but feeds like `Source` should be okay.
- Automatically closes the Share modal once the transaction for sharing a post to a squad was successful.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1035 #done

### DIFF
--- a/packages/shared/src/components/Feed.tsx
+++ b/packages/shared/src/components/Feed.tsx
@@ -8,7 +8,7 @@ import React, {
 } from 'react';
 import classNames from 'classnames';
 import dynamic from 'next/dynamic';
-import useFeed, { PostItem } from '../hooks/useFeed';
+import useFeed, { PostItem, UseFeedOptionalParams } from '../hooks/useFeed';
 import { Ad, Post, PostType } from '../graphql/posts';
 import AuthContext from '../contexts/AuthContext';
 import FeedContext from '../contexts/FeedContext';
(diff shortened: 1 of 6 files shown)


sshanzel:
> I updated the useFeed hook to accept the options for enabling refetch on mount. This is unwanted for regular feeds, but feeds like Source should be okay.

@idoshamun would love to hear your thoughts in the section above. We can manually push the item in the query cache for squad feed posts, but it is a paginated infinite query and I haven't had any chance in the past to work on something similar, there could be some conflict that might arise. But if not, we can actually do that. WDYT?

packages/shared/src/components/Feed.tsx
```diff
@@ -174,8 +174,7 @@ export default function Feed<T>({
       currentSettings.adSpot,
       numCards,
       showOnlyUnreadPosts,
-      query,
-      variables,
+      { query, variables, options: { refetchOnMount: true } },
```
rebelchris, line 177: Is this not a default?

rebelchris (approved):
Hmm yeah good point, pretty sure this refetch was the issue as we always reset the feed on creation.
No blocker just a question, feel free to merge.

packages/shared/src/components/Feed.tsx
sshanzel: I was also surprised to see this actually. Probably the mechanism for real-time updates in our feed is different, I could be wrong tho.

image

sshanzel:
I will be merging this for the moment since the change here is only configured for the Squad feed and won't affect production existing feeds. But if any, I will quickly update any inputs.

PR #1588, opened 2023-02-07
refactor: list priority for squad member list

sshanzel:
- Created a priority to easily identify the order in whom to display.
- Member is set at `0` since this is the most basic level, to make it backwards compatible (extension-wise), I have introduced a default value of new introductions as `1`. Most likely we will introduce moderators/admins, which are a level above members (0). The owner is the highest (10) - it is highly unlikely we will get 10 levels with roles.
- Deleted an empty file.

Preview:
image

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1034 #done

### DIFF
--- a/packages/shared/src/components/modals/SquadMemberModal.tsx
+++ b/packages/shared/src/components/modals/SquadMemberModal.tsx
@@ -1,4 +1,4 @@
-import React, { ReactElement, useState } from 'react';
+import React, { ReactElement, useMemo, useState } from 'react';
 import { useInfiniteQuery } from 'react-query';
 import request from 'graphql-request';
 import { graphqlUrl } from '../../lib/config';
@@ -8,6 +8,7 @@ import {
   SQUAD_MEMBERS_QUERY,
   SquadEdgesData,
   SquadMemberRole,
+  getRolePriority,
 } from '../../graphql/squads';
 import UserListModal from './UserListModal';
 import { checkFetchMore } from '../containers/InfiniteScrolling';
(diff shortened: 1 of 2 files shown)


packages/shared/src/components/modals/SquadMemberModal.tsx
```diff
@@ -52,6 +53,23 @@ export function SquadMemberModal({
     onMenuClick(e);
   };
 
+  const members = useMemo(
```
rebelchris, line 56: Should we not rather do this on the backend?
Seems to be more memory efficient actually.

rebelchris (changes requested):

packages/shared/src/components/modals/SquadMemberModal.tsx
sshanzel: Good point. We can save resources if we instead use the client's processing power (FE) for simple list manipulation than in BE as sorting is expensive in DB query. But definitely an option. WDYT? @idoshamun

sshanzel: @rebelchris since the sorting won't probably be too heavy in the backend as we are at closed beta, I will open up a PR in the backend for that and close this one.

el: I aggree with @rebelchris   this should be done in the be

idoshamun: Yap agree. It's backend responsibility. Especially when paging involved (not now but probably in the future)

PR #1592, opened 2023-02-07
feat: introduces squad tracking events

rebelchris:
- Introduces new tracking events for the squads beta
- Introduced a new way to track optional prev/next on stepped modals, would love to hear your thoughts. After this PR we can modify the other implementation as well.

## Events

There are a bunch of new events, without noting them down here, I refer to the taxonomy sheet:
https://docs.google.com/spreadsheets/d/18Lv7zXges9QfVX5VYL1a-Hyl0e1sQ3sLr0OK8YZWKXI/edit#gid=2069206456

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-951 #done

### DIFF
--- a/packages/shared/src/components/ShareOptionsMenu.tsx
+++ b/packages/shared/src/components/ShareOptionsMenu.tsx
@@ -8,7 +8,7 @@ import { postAnalyticsEvent } from '../lib/feed';
 import { MenuIcon } from './MenuIcon';
 import { ShareBookmarkProps } from './post/PostActions';
 import BookmarkIcon from './icons/Bookmark';
-import { Origin } from '../lib/analytics';
+import { AnalyticsEvent, Origin } from '../lib/analytics';
 import { ShareProvider } from '../lib/share';
 import { useCopyPostLink } from '../hooks/useCopyPostLink';
 import LinkIcon from './icons/Link';
@@ -109,14 +109,19 @@ export default function ShareOptionsMenu({
             <MenuIcon Icon={DefaultSquadIcon} />
           ),
           text: `Share to ${squad.name}`,
-          action: () =>
+          action: () => {
+            trackEvent({ event_name: AnalyticsEvent.StartShareToSquad });
             openModal({
               type: LazyModal.PostToSquad,
               props: {
                 squad,
                 post,
+                onSharedSuccessfully: () => {
+                  trackEvent({ event_name: AnalyticsEvent.ShareToSquad });
+                },
               },
-            }),
+            });
+          },
         }),
     );
   }
--- a/packages/shared/src/components/ShareOptionsMenu.tsx
+++ b/packages/shared/src/components/ShareOptionsMenu.tsx
@@ -94,7 +94,7 @@ export default function ShareOptionsMenu({
         action: () =>
           openModal({
             type: LazyModal.NewSquad,
-            props: { post },
+            props: { post, origin: Origin.Share },
           }),
       });
     }
--- a/packages/shared/src/components/ShareOptionsMenu.tsx
+++ b/packages/shared/src/components/ShareOptionsMenu.tsx
@@ -110,14 +110,18 @@ export default function ShareOptionsMenu({
           ),
           text: `Share to ${squad.name}`,
           action: () => {
-            trackEvent({ event_name: AnalyticsEvent.StartShareToSquad });
+            trackEvent(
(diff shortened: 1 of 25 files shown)


idoshamun:
I haven't yet reviewed carefully but skimmed through and noticed this PR misses two new properties we added to post events
image

And also two new post events
image

They are not in the Squads tab because they belong to the posts section but we tagged you in these changes

rebelchris:
@idoshamun Maybe we should avoid these tabs, I totally forgot about those.
And for me it doesn't make things quicker or easier.

packages/shared/src/lib/analytics.ts
```diff
@@ -35,13 +44,26 @@ export enum AnalyticsEvent {
   ClickNotificationDismiss = 'click notification dismiss',
   DisableNotification = 'disable notification',
   // notifications - end
+  // squads - start
+  ViewSquadInvitation = 'view squad invitation',
+  ClickJoinSquad = 'click join squad',
+  CompleteJoiningSquad = 'complete joining the squad',
```
idoshamun, line 50:
```suggestion
  CompleteJoiningSquad = 'complete joining squad',
```

packages/webapp/pages/squads/[handle]/[token].tsx
```diff
@@ -79,11 +84,37 @@ const SquadReferral = ({ token, handle }: SquadReferralProps): ReactElement => {
       },
     },
   );
+
+  useEffect(() => {
+    if (trackedImpression || !member) return;
+
+    trackEvent({
+      event_name: AnalyticsEvent.ViewSquadInvitation,
+      extra: JSON.stringify({
```
Comment (idoshamun) [inline packages/webapp/pages/squads/[handle]/[token].tsx:93]: Why don't you use `joinSquadAnalyticsExtra` here?

idoshamun:
@rebelchris from a data analysis perspective, it's much easier to handle. But we need to find better ways to organize these changes.

idoshamun:
The rest looks great

idoshamun:
@rebelchris one more point, this is a tricky one. When I'm clicking "read article" on a share post, I think that current implementation will provide the details of the share post, and I think we need to provide the details post that was shared. It's very important from feed perspective as well. We can add an extra field to indicate the id of the share post so we can know that it was triggered due to this. wdyt?

packages/shared/src/components/ShareOptionsMenu.tsx
```diff
@@ -109,14 +109,19 @@ export default function ShareOptionsMenu({
             <MenuIcon Icon={DefaultSquadIcon} />
           ),
           text: `Share to ${squad.name}`,
-          action: () =>
+          action: () => {
+            trackEvent({ event_name: AnalyticsEvent.StartShareToSquad });
```
idoshamun, line 113: This is a post event and should include all the properties we include in other post events (see for example click, or upvote post)

packages/shared/src/components/ShareOptionsMenu.tsx
```diff
@@ -109,14 +109,19 @@ export default function ShareOptionsMenu({
...
             openModal({
               type: LazyModal.PostToSquad,
               props: {
                 squad,
                 post,
+                onSharedSuccessfully: () => {
+                  trackEvent({ event_name: AnalyticsEvent.ShareToSquad });
```
idoshamun, line 120: Same as above here

packages/shared/src/components/widgets/SocialShare.tsx
```diff
@@ -61,15 +61,20 @@ export const SocialShare = ({
       }),
     );
 
-  const onShareToSquad = (squad: Squad) =>
+  const onShareToSquad = (squad: Squad) => {
+    trackEvent({ event_name: AnalyticsEvent.StartShareToSquad });
```
idoshamun, line 65: Same here

packages/shared/src/components/widgets/SocialShare.tsx
```diff
@@ -61,15 +61,20 @@ export const SocialShare = ({
...
     openModal({
       type: LazyModal.PostToSquad,
       props: {
         squad,
         post,
-        onSharedSuccessfully: onSquadShare,
+        onSharedSuccessfully: () => {
+          trackEvent({ event_name: AnalyticsEvent.ShareToSquad });
```
idoshamun, line 72: Same here

idoshamun (approved):

PR #1595, opened 2023-02-08
fix: missing private property in reading history query

sshanzel:
- The reading history query was missing the private property.
- Also fixed the selections of Squad to be the active ones only.
- Let's open a ticket to consolidating our queries to use a Fragment for better maintenance I think.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done

### DIFF
--- a/packages/shared/src/graphql/users.ts
+++ b/packages/shared/src/graphql/users.ts
@@ -181,6 +181,7 @@ const READING_HISTORY_FRAGMENT = gql`
       numUpvotes
       bookmarked
       numComments
+      private
       sharedPost {
         ...SharedPostInfo
       }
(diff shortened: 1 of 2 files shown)


packages/shared/src/graphql/users.ts
```diff
@@ -181,6 +181,7 @@ const READING_HISTORY_FRAGMENT = gql`
       numUpvotes
       bookmarked
       numComments
+      private
```
rebelchris, line 184: This is not great, it should still show in your actual reading history, just not the ones that are used for sharing I think.

rebelchris (changes requested):

packages/shared/src/graphql/users.ts
sshanzel: We already have the condition in the ShareModal to not show private posts. In reading history, the context menu also have that logic.

rebelchris: Sorry my bad, I was thinking you introduced "isPublic" 😅

rebelchris (approved):
Good to go, thanks!

PR #1599, opened 2023-02-09
fix: squad join page responsiveness

sshanzel:
- Responsiveness of the Squad join page was missed. There should be a button under the description when it is mobile.
- This also fixes the wrap (overflow) issue on the "members waiting" list.
- I also fixed the missing left margin on the shared post content's cover image.

Preview:
image

image

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1058 #done

### DIFF
--- a/packages/webapp/pages/squads/[handle]/[token].tsx
+++ b/packages/webapp/pages/squads/[handle]/[token].tsx
@@ -149,7 +149,7 @@ const SquadReferral = ({ token, handle }: SquadReferralProps): ReactElement => {
   );
 
   return (
-    <PageContainer className="relative justify-center items-center pt-24">
+    <PageContainer className="overflow-hidden relative justify-center items-center pt-24">
       <NextSeo
         title={`Invitation to ${source.name}`}
         titleTemplate="%s | daily.dev"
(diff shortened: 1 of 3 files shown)


packages/webapp/pages/squads/[handle]/[token].tsx
```diff
@@ -179,21 +179,30 @@ const SquadReferral = ({ token, handle }: SquadReferralProps): ReactElement => {
...
+            {source.description}
+          </BodyParagraph>
+        )}
+        <Button
+          className="flex tablet:hidden mt-4 ml-auto w-full btn-primary"
+          buttonSize="large"
+          onClick={onJoinClick}
+        >
+          Join Squad
+        </Button>
```
Comment (rebelchris) [inline packages/webapp/pages/squads/[handle]/[token].tsx:200]: Not sure if it's possible, but can't we put the bodyParagraph inside this element?
And then flex-reverse it on mobile?

Beats having two buttons that do the same in my opinion.

rebelchris (changes requested):

packages/webapp/pages/squads/[handle]/[token].tsx
Comment (sshanzel) [reply packages/webapp/pages/squads/[handle]/[token].tsx:200]: Great suggestion, but not possible, unfortunately, the first button is inside a flex-row where the paragraph is outside of it and must be under those components.

Comment (rebelchris) [reply packages/webapp/pages/squads/[handle]/[token].tsx:200]: But looking at both mobile and desktop designs it's all inside that one box right?
Not quite sure why a flex solution would not work, seems to even simply push the button down that's all.

Comment (sshanzel) [reply packages/webapp/pages/squads/[handle]/[token].tsx:200]: Yes, but the box below, this must be contained, so if we add inside the description, it will be placed on the right side of the button, or if we will contain the two of them, the placement of the description will only be in the same column of the button.

Another option would be to contain the name, handle and description in the same column but then the description won't take the space on the right which is under the join button.

image

Comment (rebelchris) [reply packages/webapp/pages/squads/[handle]/[token].tsx:200]: Oh right, didn't take the description into account.
I don't have a issue with it, but never like multiple buttons, danger lies in changing behaviour/text/etc.

Comment (sshanzel) [reply packages/webapp/pages/squads/[handle]/[token].tsx:200]: I agree, me too, just had no other choice 😭

Comment (rebelchris) [reply packages/webapp/pages/squads/[handle]/[token].tsx:200]: Suggestion for that: Can we maybe make it 1 shared component on top of the file? or outside the component so we limit the risk.

rebelchris (changes requested):

packages/webapp/pages/squads/[handle]/[token].tsx
Comment (sshanzel) [reply packages/webapp/pages/squads/[handle]/[token].tsx:200]: Good suggestion! Let me add that, probably good to have test cases as well to ensure when they are both displayed.

Comment (sshanzel) [reply packages/webapp/pages/squads/[handle]/[token].tsx]: Made the function and applied tests as well.

rebelchris (approved):
Amazing work Lee 🙌

PR #1600, opened 2023-02-09
fix: feedback button overlapping with title

sshanzel:
- Apply 24px padding when it is on mobile.

Preview:
image

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1052 #done

### DIFF
--- a/packages/shared/src/components/squads/SquadHeaderMenu.tsx
+++ b/packages/shared/src/components/squads/SquadHeaderMenu.tsx
@@ -1,6 +1,5 @@
 import React, { ReactElement, useContext } from 'react';
 import dynamic from 'next/dynamic';
-import { Item } from '@dailydotdev/react-contexify';
 import { useRouter } from 'next/router';
 import AuthContext from '../../contexts/AuthContext';
 import EditIcon from '../icons/Edit';
@@ -12,6 +11,11 @@ import { useLazyModal } from '../../hooks/useLazyModal';
 import { LazyModal } from '../modals/common/types';
 import { useDeleteSquad } from '../../hooks/useDeleteSquad';
 import { useLeaveSquad } from '../../hooks/useLeaveSquad';
+import FeedbackIcon from '../icons/Feedback';
+import { squadFeedback } from '../../lib/constants';
+import ContextMenuItem, {
+  ContextMenuItemProps,
+} from '../tooltips/ContextMenuItem';
 
 const PortalMenu = dynamic(
   () => import(/* webpackChunkName: "portalMenu" */ '../fields/PortalMenu'),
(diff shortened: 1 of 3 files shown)


rebelchris:
@sshanzel In the design it actually states this becomes a context button action.
Screenshot 2023-02-09 at 07 18 39

rebelchris (changes requested):
@sshanzel It's not inline with the mobile version actually, missed that in my initial review.

So apparently that feedback button and the invite button become part of the context menu.

sshanzel:
> @sshanzel It's not inline with the mobile version actually, missed that in my initial review.
> 
> So apparently that feedback button and the invite button become part of the context menu.

Oh, sure, let's update those missing items as well in this PR then, since we're already moving the feedback button.

rebelchris:
@sshanzel I raised a request about it on Slack, I personally don't feel that invite button is a good idea.
So maybe let's await that part.

sshanzel:
> So maybe let's await that part.

Cool. I'll adjust the feedback only then. Thanks!

packages/shared/src/components/squads/SquadHeaderMenu.tsx
```diff
@@ -59,34 +69,24 @@ export default function SquadHeaderMenu({
...
-        </Item>
-      ) : (
-        <Item className="typo-callout" onClick={onLeaveSquad}>
-          <span className="flex items-center w-full typo-callout">
-            <ExitIcon size="medium" secondary={false} className="mr-2" /> Leave
-            Squad
-          </span>
-        </Item>
-      )}
+      <ContextMenuItem
```
el, line 84: this should be hidden on non-mobile i think

el (changes requested):
small change

packages/shared/src/components/squads/SquadHeaderMenu.tsx
rebelchris: Agreed.

rebelchris (changes requested):
Nice addition for this component.
Optional you could even have a array that loops the potential items (like we do in sidebar, but it's up to you)

rebelchris (approved):

sshanzel:
> Optional you could even have a array that loops the potential items (like we do in sidebar, but it's up to you)

Oh, right. Let me do that to keep things consistent. Thanks!

sshanzel:
@rebelchris have updated it. Next we can create an options menu component to just accept props and these items to make it easier in the future.

PR #1603, opened 2023-02-09
fix: refetch of bookmarks to update

sshanzel:
## Changes

We had two options, update the data locally or allow refetch when the cache is stale.

The reason we didn't go for the first one is, there could be instances when you bookmark a post, you might not have all the necessary information when you toggled the bookmark (i.e. in reading history) which will then update the cache (of bookmarks feed), we will have to fetch to ensure the data is complete, else, the bookmarks feed might malfunction.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done

### DIFF
--- a/packages/shared/src/components/BookmarkFeedLayout.tsx
+++ b/packages/shared/src/components/BookmarkFeedLayout.tsx
@@ -14,6 +14,7 @@ import Feed, { FeedProps } from './Feed';
 import BookmarkEmptyScreen from './BookmarkEmptyScreen';
 import { Button } from './buttons/Button';
 import ShareIcon from './icons/Share';
+import { generateQueryKey, RequestKey } from '../lib/query';
 
 export type BookmarkFeedLayoutProps = {
   searchQuery?: string;
@@ -37,12 +38,12 @@ export default function BookmarkFeedLayout({
   const { user, tokenRefreshed } = useContext(AuthContext);
   const [showEmptyScreen, setShowEmptyScreen] = useState(false);
   const [showSharedBookmarks, setShowSharedBookmarks] = useState(false);
-
+  const defaultKey = generateQueryKey(RequestKey.Bookmarks, user);
   const feedProps = useMemo<FeedProps<unknown>>(() => {
     if (searchQuery) {
       return {
         feedName: 'search-bookmarks',
-        feedQueryKey: ['bookmarks', user?.id ?? 'anonymous', searchQuery],
+        feedQueryKey: defaultKey.concat(searchQuery),
         query: SEARCH_BOOKMARKS_QUERY,
         variables: { query: searchQuery },
         emptyScreen: <SearchEmptyScreen />,
@@ -50,7 +51,7 @@ export default function BookmarkFeedLayout({
     }
     return {
       feedName: 'bookmarks',
-      feedQueryKey: ['bookmarks', user?.id ?? 'anonymous'],
+      feedQueryKey: defaultKey,
       query: BOOKMARKS_FEED_QUERY,
       onEmptyFeed: () => setShowEmptyScreen(true),
       options: { refetchOnMount: true },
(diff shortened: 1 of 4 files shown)


rebelchris (approved):
Not sure when we would miss information, but think the refetchOnMount is a good way to go in this case.
It would allow for background fetch regardless which the user would expect anyway.

sshanzel:
I agree, I think it is always best to refetch on mount to refresh the data.

> Not sure when we would miss information, but think the refetchOnMount is a good way to go in this case.

On the reading history page, we only fetch only minimal amount of information about the post. So I was thinking its query is probably not appropriate with the feed query's requirement.

sshanzel:
Oh, wait removal scenario, I must update the local cache, this time by deleting as it can be easy as just filtering and there will be no onmount trigger.

rebelchris:
@sshanzel Yes, think we actually already support bookmark local cache manipulation right?

sshanzel:
@rebelchris updated it now. Even though we may take a different path in the future with regard to generating a query key, I think the existing generic function can make things a bit cleaner today and will make the refactor easy. WDYT?

sshanzel:
I checked the codebase, it was for updating local reading history post to not have the `bookmarked` property to true/false.

packages/shared/src/components/BookmarkFeedLayout.tsx
```diff
@@ -37,22 +38,23 @@ export default function BookmarkFeedLayout({
...
   const [showEmptyScreen, setShowEmptyScreen] = useState(false);
   const [showSharedBookmarks, setShowSharedBookmarks] = useState(false);
-
+  const defaultKey = generateQueryKey(RequestKey.Bookmarks, user);
   const feedProps = useMemo<FeedProps<unknown>>(() => {
     if (searchQuery) {
       return {
         feedName: 'search-bookmarks',
-        feedQueryKey: ['bookmarks', user?.id ?? 'anonymous', searchQuery],
+        feedQueryKey: defaultKey.concat(searchQuery),
```
el, line 46: I find array spread clearer 
```suggestion
        feedQueryKey: [...defaultKey, ...searchQuery],
```

sshanzel: Good point. I just think the former feels like reading an actual English sentence and making it intuitive. Though this is probably just a personal preference. WDYT? @rebelchris

packages/shared/src/hooks/useBookmarkPost.ts
```diff
@@ -64,9 +68,26 @@ export default function useBookmarkPost<
...
-        trackEvent(onRemoveBookmarkTrackObject()),
+      onSuccess: (_, { id }) => {
+        if (onRemoveBookmarkTrackObject)
+          trackEvent(onRemoveBookmarkTrackObject());
+
+        const key = generateQueryKey(RequestKey.Bookmarks, user);
+        client.setQueryData<InfiniteData<FeedData>>(key, (feed) => {
+          if (!feed) return null;
+
+          return {
```
el, line 79: Maybe you can move this logic to a utility function, since it looks like a useful thing to have 
```ts

removeNode(pages: Page[], id: string): {
   return feed?.pages?.map((edge) => ({
              ...edge,
              page: {
                ...edge.page,
                edges: edge.page.edges.filter(({ node }) => node.id !== id),
              },
            })),
}

```

el (approved):
Other than low-priority suggestions, It looks good :shipit:

packages/shared/src/hooks/useBookmarkPost.ts
sshanzel: Oh, nice. That is true. Let me create one for update and filter like this.

packages/shared/src/components/BookmarkFeedLayout.tsx
rebelchris: I personally also like spread more, but it's really optional.
Do think we tend to use spread more than concat already.

rebelchris (approved):
Awesome, besides Eliz optional comment nothing major

PR #733, opened 2021-12-11
build(extension): upgrading manifest v2 to v3

sshanzel:
The changes made in the manifest were purely based on the [guide in this document](https://developer.chrome.com/docs/extensions/mv3/intro/mv3-migration/#declarativenetrequest-conditional-perms).

I have separated each commit to isolate the changes required from the document.

- [Host permissions: ](https://developer.chrome.com/docs/extensions/mv3/intro/mv3-migration/#host-permissions) 082d8f08f51ade0971d2a6a1b76e9385242d3fae
- [Content security policy:](https://developer.chrome.com/docs/extensions/mv3/intro/mv3-migration/#content-security-policy) 0da511e943ca1f824002e6e5717c29c892f25a41
- [Action API unification:](https://developer.chrome.com/docs/extensions/mv3/intro/mv3-migration/#action-api-unification) dc077ca860f5190c3a46b3e3442b964d5ed0130e
- [WIP - web accessible resources:](https://developer.chrome.com/docs/extensions/mv3/intro/mv3-migration/#web-accessible-resources) dc077ca860f5190c3a46b3e3442b964d5ed0130e
- [Remotely hosted code:](https://developer.chrome.com/docs/extensions/mv3/intro/mv3-migration/#remotely-hosted-code) Not applicable
- [Executing arbitrary strings:](https://developer.chrome.com/docs/extensions/mv3/intro/mv3-migration/#executing-arbitrary-strings) Not applicable
- [Background service workers:](https://developer.chrome.com/docs/extensions/mv3/intro/mv3-migration/#background-service-workers) Not applicable
- [Modifying network requests:](https://developer.chrome.com/docs/extensions/mv3/intro/mv3-migration/#background-service-workers) Not applicable

[Sunset for deprecated APIs](https://developer.chrome.com/docs/extensions/mv3/intro/mv3-migration/#background-service-workers) - we only had to change one API- [`chrome.extension.getURL()`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/extension/getURL) -> [`runtime.getURL()`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/runtime/getURL)
chrome.extension.getURL

[...]

PR #1605, opened 2023-02-09
feat: changelog pop-up

capJavert:
- Implement new change log pop-up

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1045 #done


packages/shared/src/components/sidebar/common.tsx
```diff
@@ -137,12 +142,15 @@ export const ItemInner = ({
...
         className={classNames(
-          'flex-1 text-left transition-opacity truncate',
+          'flex flex-1 text-left transition-opacity truncate flex-row items-center',
           shouldShowLabel ? 'opacity-100 delay-150' : 'opacity-0',
           item?.className?.text,
         )}
       >
         {item.title}
+        {item.badge?.active === true && (
+          <span className="block ml-auto w-2 h-2 rounded-lg bg-theme-color-cabbage" />
```
rebelchris, line 152: Maybe we can use the `AlertDot` component for this?

packages/shared/src/components/sidebar/SidebarBottomSection.tsx
```diff
@@ -29,6 +44,9 @@ export function SidebarBottomSectionSection({
       icon: () => <ListIcon Icon={() => <TerminalIcon />} />,
       title: 'Changelog',
       path: `${process.env.NEXT_PUBLIC_WEBAPP_URL}sources/daily_updates`,
+      badge: {
+        active: isChangelogModalActive,
+      },
```
rebelchris, line 49: We actually support `alert` already on the item.

packages/shared/src/components/sidebar/SidebarBottomSection.tsx
```diff
@@ -45,6 +63,7 @@ export function SidebarBottomSectionSection({
       {props.sidebarExpanded && !optOutWeeklyGoal && (
         <SidebarRankProgress {...props} disableNewRankPopup={showSettings} />
       )}
+      <ChangelogModal isOpen={isChangelogModalActive} />
```
rebelchris, line 66: Since we dynamically import the modal (loading performance) it's best to also conditionally render this item only when `isChangelogModalActive` is true.

rebelchris (commented):
Some early feedback 🙏

packages/shared/src/components/modals/ChangelogModal.tsx
```diff
@@ -0,0 +1,24 @@
...
+function ChangelogModal({
+  className,
+  children,
+  ...props
+}: ModalProps): ReactElement {
+  return (
+    <Modal
+      kind={Modal.Kind.FixedCenter}
+      size={Modal.Size.Small}
+      overlayClassName={styles.changelogModal}
```
el, line 15: Please use the tailwind classes instead `bg-none`

packages/shared/src/components/sidebar/SidebarBottomSection.tsx
```diff
@@ -8,6 +10,11 @@ import InvitePeople from './InvitePeople';
 import { Section, SectionCommonProps } from './Section';
 import { docs, feedback } from '../../lib/constants';
 
+const ChangelogModal = dynamic(
```
el, line 13: Can you please use LazyModal instead ?

el (changes requested):
This is a great start @capJavert 

Please make sure to target this PR to an epic branch instead since it is not ready for production.

packages/shared/src/components/tooltips/ChangelogTooltip.module.css
```diff
@@ -0,0 +1,17 @@
+.changelogTooltipArrow {
```
sshanzel, line 1: As much as possible we wanted to sway away from using module CSS since these values are available on tailwind (let me write this down too on confluence). Also, if you need help navigating how to use this tooltip component, as there is no guideline to it yet, I'd be glad to have a coding session with you.

packages/shared/src/components/tooltips/ChangelogTooltip.tsx
```diff
@@ -0,0 +1,118 @@
...
+            </section>
+            <footer className="flex gap-3 items-center py-3 px-4 w-full h-16 border-t border-theme-divider-tertiary">
+              <Button className="btn-tertiary">Release notes</Button>
+              {isExtension && (
+                <Button className="bg-cabbage-40 btn-primary">
+                  Update extension
+                </Button>
+              )}
+            </footer>
+          </div>
```
sshanzel, line 98: This is something we would most likely be using on the modal version as well. I think we can take this out and name a new component for this.

packages/shared/src/components/tooltips/ChangelogTooltip.tsx
```diff
@@ -0,0 +1,118 @@
...
+        className: 'shadow',
+        paddingClassName: 'p-0',
+        roundedClassName: 'rounded-16',
+        bgClassName: 'bg-cabbage-40',
+        arrowClassName: styles.changelogTooltipArrow,
+      }}
+      reference={elementRef}
+      arrow
+      placement="right-end"
+      visible={visible && !!post}
```
sshanzel, line 113: Wow, amazing to see you be able to understand how to use this component quickly. Great work! 🚢

sshanzel (commented):
Just had some comments about usage of CSS module and reusable components, but you are definitely on the right track! Great work! 🚄

packages/shared/src/components/sidebar/SidebarBottomSection.tsx
capJavert: Moved away from modal on this component but gonna keep in mind for other.

packages/shared/src/components/sidebar/SidebarBottomSection.tsx
capJavert: Of course 👍

packages/shared/src/components/sidebar/SidebarBottomSection.tsx
capJavert: As far as I can see it alert is used on the icon in this case. (also positioned in the icon container) So I think new prop is required for this, I will however copy the signature by using element type instead of object, I'll like it more in this case and because render props is obviously used in the rest of the component.

packages/shared/src/components/tooltips/ChangelogTooltip.tsx
capJavert: What modal version?

sshanzel: From the Sketch link, it looks like there will be a modal equivalent of this for the experiment.

Screenshot 2023-02-13 at 8 04 31 PM

capJavert: From what I know the modal was the initial design, but one with a tooltip is the one we are going for. 

[Task page](https://dailydotdev.atlassian.net/wiki/spaces/HAN/pages/443351051/Change+log+popup#Parameters-for-A%2FB) does not mention any ab experiments, also the [referenced design](  https://www.sketch.com/s/2867ec63-0966-4647-b1d7-b6fe696ba251/p/9895C739-4D4B-4E24-BB62-E34B62CF6086/canvas) on ticket in Jira only shows the tooltip variant.

Let me know if I am missing something 🙏

sshanzel: Oh, cool. I never knew about the path taken, and as you mentioned, there wasn't anything mentioned in the task about the AB experiment. Let me check with the team as it is a bit confusing.

capJavert: After checking with Seb there is no a/b test. Tnx @sshanzel for checking

packages/extension/src/background/index.ts
```diff
@@ -131,6 +131,25 @@ async function handleMessages(message, sender: Runtime.MessageSender) {
       },
     });
   }
+
+  if (message.type === 'REQUEST_UPDATE') {
+    // check if requestUpdateCheck is available
+    // @ts-expect-error Property 'requestUpdateCheck' does not exist on type 'Static'
```
capJavert, line 137: Had to add this ts override because `webextension-polyfill-ts` is so outdated that it does not have the correct types. I think it is fine for now because I do not want to update packages in this PR and also there is a #733

packages/extension/src/background/index.ts
```diff
@@ -152,3 +171,11 @@ browser.runtime.onInstalled.addListener(async (details) => {
 browser.runtime.onStartup.addListener(async () => {
   await getContentScriptPermissionAndRegister();
 });
+
+// only add update listener if requestUpdateCheck is not available
+// @ts-expect-error Property 'requestUpdateCheck' does not exist on type 'Static'
```
capJavert, line 176: Same as https://github.com/dailydotdev/apps/pull/1605/files#r1104724992

packages/shared/src/components/tooltips/ChangelogTooltip.tsx
```diff
@@ -0,0 +1,136 @@
...
+                href={post.permalink}
+              >
+                Release notes
+              </Button>
+              {isExtension && (
+                <Button
+                  className="bg-cabbage-40 btn-primary"
+                  onClick={() => {
+                    if (isExtension) {
+                      // @ts-expect-error Property 'browser' does not exist on type 'Window & typeof globalThis'
```
capJavert, line 101: related to https://github.com/dailydotdev/apps/pull/1605/files#r1104724992

if correct types were available I could just use this behind `isExtension` check.

sshanzel:
@capJavert let me know once this is ready for another round of review, I'd be glad to go through it 🚢

packages/extension/src/background/index.ts
sshanzel: Yes, it is really a pain to keep this outdated package. Hopefully, we can focus soon on migrating the Manifest since FF has started receiving entries now iirc.

packages/shared/src/components/tooltips/ChangelogTooltip.tsx
```diff
@@ -0,0 +1,128 @@
...
+        )
+      }
+      offset={[6 * 16, 2.5 * 16]}
+      interactive
+      container={{
+        className: 'shadow',
+        paddingClassName: 'p-0',
+        roundedClassName: 'rounded-16',
+        bgClassName: 'bg-cabbage-40',
+        arrowClassName: styles.changelogTooltipArrow,
```
sshanzel, line 118: It would actually be much preferred if we won't be using CSS module. What we can probably do is, make the container accept an `arrow: ReactNode` property, so we can just use it if it has value to override the existing arrow.

packages/shared/src/hooks/useChangelog.spec.tsx
```diff
@@ -0,0 +1,96 @@
...
+      ['changelog', 'latest-post', { loggedIn: false }],
+      defaultPost,
+    );
+    const lastChangelog = new Date(defaultAlerts.lastChangelog);
+    lastChangelog.setMonth(lastChangelog.getMonth() - 1);
+    defaultAlerts.lastChangelog = lastChangelog.toISOString();
+
+    const { result } = renderHook(() => useChangelog(), {
+      wrapper: Wrapper,
+    });
```
sshanzel, line 56: This is so cool. I have never tried testing a hook before. We should do this more often. Nice work!

sshanzel (changes requested):
The way you built this feature was really great! 💪 

Just had some minor comments, and the concern with regards to not using a CSS module. Other than that, this PR looks amazing! Especially with the numerous tests you have written.

packages/shared/src/components/sidebar/ClickableNavItem.tsx
```diff
@@ -32,17 +37,20 @@ export function ClickableNavItem({
...
+            {...(item.action && { onClick: item.action })}
+            {...props}
+            target={item?.target}
+            className={navBtnClass}
+            rel="noopener noreferrer"
+          >
+            {children}
+          </a>
+        </Link>
+        {isValidElement(item.badge) && item.badge}
```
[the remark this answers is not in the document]
capJavert: If it's inside the anchor element then the whole tooltip body will be a link to the `https://app.daily.dev/sources/daily_updates` which is bad because we have other CTA and links inside the changelog tooltip.

packages/extension/src/background/index.ts
```diff
@@ -131,6 +131,25 @@ async function handleMessages(message, sender: Runtime.MessageSender) {
       },
     });
   }
+
+  if (message.type === 'REQUEST_UPDATE') {
```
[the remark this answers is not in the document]
capJavert: Sure thing, did not want to make a change because its better to have a PR changes focused on the features, I can make a separate PR with enum adjustment and merge it before this one?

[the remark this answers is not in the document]
sshanzel: Definitely, the right thing to do. So long as it will be remembered, you can even create a ticket so it won't be forgotten.

packages/shared/src/components/sidebar/ClickableNavItem.tsx
```diff
@@ -32,17 +37,20 @@ export function ClickableNavItem({
...
-          {...props}
-          target={item?.target}
-          className={navBtnClass}
-          rel="noopener noreferrer"
-        >
-          {children}
-        </a>
-      </Link>
+      <>
+        <Link href={item.path} passHref prefetch={false}>
```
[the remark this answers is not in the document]
capJavert: It was already set before I just copied the code because of the added `Fragment`. But yeah `prefetch={false}` here is good because you probably don't want to prefetch all the menu links after user lands on the page. This still prefetches on link hover which is good but IMO bad DX from next team because this should (by naming) probably disable `prefetch` all together.

packages/shared/src/components/sidebar/common.tsx
```diff
@@ -26,6 +27,11 @@ export interface SidebarUserButtonProps {
   sidebarRendered?: boolean;
 }
 
+export interface SidebarMenuItemBadge {
+  active: boolean;
+  ref?: MutableRefObject<HTMLElement>;
+}
+
```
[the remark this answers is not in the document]
capJavert: Good catch, linter did not catch it though because of `export`.

packages/extension/src/background/index.ts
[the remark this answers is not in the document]
capJavert: Added in #1613

packages/extension/src/background/index.ts
capJavert: Solve with #1613

packages/shared/src/components/sidebar/SidebarBottomSection.tsx
```diff
@@ -29,6 +42,13 @@ export function SidebarBottomSectionSection({
       icon: () => <ListIcon Icon={() => <TerminalIcon />} />,
       title: 'Changelog',
       path: `${process.env.NEXT_PUBLIC_WEBAPP_URL}sources/daily_updates`,
+      badge: changelog.isAvailable && (
+        <AlertDot
+          className="right-2"
+          ref={changelogBadgeRef}
+          color={AlertColor.Cabbage}
+        />
+      ),
```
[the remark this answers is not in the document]
sshanzel: I think you can work around that by setting the property `appendTo={globalThis?.document?.body}

packages/shared/src/components/tooltips/ChangelogTooltip.module.css
capJavert: Replace with tailwind classes

packages/shared/src/components/sidebar/SidebarBottomSection.tsx
capJavert: after some though using existing `rightIcon` prop

packages/shared/src/components/sidebar/SidebarBottomSection.tsx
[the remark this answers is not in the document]
capJavert: @sshanzel this worked 👌 

One thing I needed to adjust though was the tab index for accessibility. When `appendTo` is `document.body` then if you tab around the page the order will be:
- header
- header items
- menu
- menu items
- rest of the page

So tabbing over menu items does not tab to the changelog tooltip which means that people would miss that completely.

In this commit https://github.com/dailydotdev/apps/pull/1605/commits/463ec2712382913fe8d50c9c0d1b2eb1f34f9d13 I passed the ref for navItem around so I can then append to the `li` tag. This makes the tab index order as following:
- header
- header items
- menu
- menu items
  - changelog tooltip ✅ 
- rest of the page

I think this should be ok now implementation wise.

packages/shared/src/components/tooltips/ChangelogTooltip.tsx
```diff
@@ -0,0 +1,167 @@
...
+                          case 'update_available':
+                            toastMessage = 'Browser extension updated';
+                            break;
+                          default:
+                            break;
+                        }
+
+                        if (toastMessage) {
+                          toast.displayToast(toastMessage);
+                        }
```
capJavert, line 136: @idoshamun added toast as a response to "update extension" button click, just checking if it looks good to you?

capJavert: `update_available` case will not currently even be shown because when `background` script triggers `runtime.reload` the extension will just close up in the current tab and the user will get the new version when they open next tab.

packages/shared/src/components/tooltips/ChangelogTooltip.tsx
```diff
@@ -0,0 +1,167 @@
...
+        roundedClassName: 'rounded-16',
+        bgClassName: 'bg-cabbage-40',
+        arrowClassName:
+          'bg-cabbage-40 bottom-24 !left-0 !h-2.5 !w-0 flex items-center justify-end before:!w-10 before:!h-px before:!transform-none',
+      }}
+      reference={elementRef}
+      arrow
+      placement="right-end"
+      visible={!!post}
+      zIndex={9}
```
capJavert, line 162: @sshanzel FYI add this so it is below settings and other modals, as far as I can see the modals are `z-index: 10`

packages/shared/src/components/sidebar/SidebarBottomSection.tsx
```diff
@@ -45,6 +71,12 @@ export function SidebarBottomSectionSection({
       {props.sidebarExpanded && !optOutWeeklyGoal && (
         <SidebarRankProgress {...props} disableNewRankPopup={showSettings} />
       )}
+      {changelog.isAvailable && (
+        <ChangelogTooltip
+          elementRef={changelogBadgeRef}
+          appendTo={() => navItemRef.current}
```
sshanzel, line 77: Way to go! 🚢 🚢 🚢

packages/shared/src/hooks/useChangelog.ts
```diff
@@ -0,0 +1,59 @@
...
+type UseChangelog = {
+  isAvailable: boolean;
+  latestPost?: Post;
+  dismiss: () => Promise<void>;
+};
+
+export function useChangelog(): UseChangelog {
+  const { alerts, updateAlerts } = useContext(AlertContext);
+  const { user } = useContext(AuthContext);
+  const isDesktop = useMedia([laptop.replace('@media ', '')], [true], false);
```
sshanzel, line 18: We have an existing hook called `useSidebarRendered`, it returns a property named `sidebarRendered` with a boolean value. If it is true, then it is desktop. We can probably utilize it.

packages/shared/src/components/tooltips/ChangelogTooltip.tsx
sshanzel: Great catch! 🚀

packages/shared/src/components/tooltips/ChangelogTooltip.tsx
```diff
@@ -0,0 +1,167 @@
...
+                          case 'no_update':
+                            toastMessage =
+                              'You are already on the latest available version';
+                            break;
+                          case 'update_available':
+                            toastMessage = 'Browser extension updated';
+                            break;
+                          default:
+                            break;
+                        }
```
sshanzel, line 132: I think we can create an object with key as the `status` and the value will be what we set for the toast message. It would also probably a good idea to turn it into enum since we will be creating a map for the message to be type-strict. WDYT?

Then we can make it like below or something similar:
```
const toastMessage = updateStatusMessage[updateResponse.status]
```

sshanzel (commented):
I am really amazed with how you tackled this feature! Great work indeed 👏 

We are almost there! Just had some very minor stuff but this is already looking spectacular ✨

packages/shared/src/components/tooltips/ChangelogTooltip.tsx
capJavert: I though about that but in the end I don't have the correct types because of the outdated `webextension-polyfill` and I think it would be contra productive to add our own `status` type. When #733 is merged I think it would be a right time to then just use https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/runtime/RequestUpdateCheckStatus 

I did however refactor to object map, it is more readable like this 👍

sshanzel (approved):
Can't believe we are about to ship this feature now 🥳 🥳 🥳 

Everything looks good to me, but since Chris will be back by Monday, it would be nice to get a round of review from him as his eyes are as sharp as 🦅 

cc @rebelchris

packages/extension/src/background/index.ts
rebelchris: So we can change this right?

packages/shared/src/components/tooltips/ChangelogTooltip.tsx
```diff
@@ -0,0 +1,160 @@
...
+                  }}
+                >
+                  Update extension
+                </Button>
+              )}
+            </footer>
+          </div>
+        )
+      }
+      offset={[6 * 16, 2.5 * 16]}
```
rebelchris, line 141: What is this?
Looks like magic numbers, maybe we can abstract it rather, if it's even needed.

rebelchris (changes requested):
Wow amazing work,
got some questions regarding some elements.

Also is there a way to mimic it in preview deployment?

packages/shared/src/components/tooltips/ChangelogTooltip.tsx
capJavert: > Wow amazing work, got some questions regarding some elements.
> 
> Also is there a way to mimic it in preview deployment?

Honestly locally I would just update the `alerts.lastChangelog` for my user to some previous date (year 2017 or something). Alternatively  some flagsmith flag to always show latest `daily_updates` post but I don't know about that.

capJavert: > What is this? Looks like magic numbers, maybe we can abstract it rather, if it's even needed.

It's just to position the tooltip as per design. I agree with magic numbers point, probably I could just extract it to the top and do something like `const tooltipArrowOffset = [6 * 16, 2.5 * 16]`

rebelchris: That last one would work for me :)

packages/extension/src/background/index.ts
capJavert: It should be solved with https://github.com/dailydotdev/apps/pull/733, #1613 only added `message.type`. Unresolving this for reference.

packages/shared/src/components/tooltips/ChangelogTooltip.tsx
capJavert: Done

packages/shared/src/components/sidebar/SidebarBottomSection.tsx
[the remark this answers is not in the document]
sshanzel: That is super amazing. Thank you for giving great thought to it. This should do very well for the initial implementation

packages/shared/src/graphql/posts.ts
```diff
@@ -390,3 +390,15 @@ export const VIEW_POST_MUTATION = gql`
...
   request(graphqlUrl, VIEW_POST_MUTATION, { id });
+
+export const getLatestChangelogPost = async (
+  loggedIn: boolean,
+): Promise<Post> => {
+  const feedData = await request<FeedData>(graphqlUrl, SOURCE_FEED_QUERY, {
+    source: 'daily_updates',
+    first: 1,
+    loggedIn,
+  });
```
rebelchris, line 401: I think you need to include `ranking: RankingAlgorithm.Time` here.

rebelchris (changes requested):
Only found one issue that might have caused the flicker as well.

rebelchris:
@capJavert I found another issue.

If you are anonymous and get the changelog you can't dismiss it, since you can't invoke "updateAlerts"

capJavert:
> @capJavert I found another issue.
> 
> If you are anonymous and get the changelog you can't dismiss it, since you can't invoke "updateAlerts"

Interesting, you should not be getting it right if you are anonymous? Since `lastChangelog` is set to `false` always https://dailydotdev.slack.com/archives/C02AA18RD1D/p1676033578757279?thread_ts=1675948105.488369&cid=C02AA18RD1D

packages/shared/src/components/sidebar/SidebarBottomSection.spec.tsx
```diff
@@ -0,0 +1,112 @@
...
+      ['changelog', 'latest-post', { loggedIn: false }],
+      defaultPost,
+    );
+    const lastChangelog = new Date(defaultAlerts.lastChangelog);
+    lastChangelog.setMonth(lastChangelog.getMonth() - 1);
+    defaultAlerts.lastChangelog = lastChangelog.toISOString();
+
+    let changelog: HTMLElement;
+
+    await act(async () => {
```
rebelchris, line 82: Just general remark, preferably it's best to avoid act, there are some other solution as to test changing elements, was it giving issues in this context?

packages/shared/src/components/sidebar/SidebarBottomSection.spec.tsx
```diff
@@ -0,0 +1,112 @@
...
+  it('should NOT render changelog and badge if changelog NOT available', async () => {
+    client.setQueryData(
+      ['changelog', 'latest-post', { loggedIn: false }],
+      defaultPost,
+    );
+    const lastChangelog = new Date(defaultAlerts.lastChangelog);
+    lastChangelog.setMonth(lastChangelog.getMonth() + 1);
+    defaultAlerts.lastChangelog = lastChangelog.toISOString();
+
+    act(() => {
```
rebelchris, line 102: This should not be needed

packages/shared/src/components/tooltips/ChangelogTooltip.tsx
```diff
@@ -0,0 +1,162 @@
...
+                data-testid="changelogReleaseNotesBtn"
+              >
+                Release notes
+              </Button>
+              {isExtension && (
+                <Button
+                  tag={isFirefoxExtension ? 'a' : undefined}
+                  href={
+                    isFirefoxExtension
+                      ? 'https://r.daily.dev/firefoxupdate'
```
rebelchris, line 127: This should actually be logged `lib/constants`

packages/shared/src/components/tooltips/ChangelogTooltip.tsx
```diff
@@ -0,0 +1,162 @@
...
+      }
+      offset={tooltipArrowOffset}
+      interactive
+      container={{
+        className: 'shadow',
+        paddingClassName: 'p-0',
+        roundedClassName: 'rounded-16',
+        bgClassName: 'bg-cabbage-40',
+        arrowClassName:
+          'bg-cabbage-40 bottom-24 !left-0 !h-2.5 !w-0 flex items-center justify-end before:!w-10 before:!h-px before:!transform-none',
```
rebelchris, line 151: Just a question, are all the important overrides needed here?

rebelchris (changes requested):
Some last small comments and questions.
Almost there!

packages/shared/src/components/tooltips/ChangelogTooltip.tsx
capJavert: Yeah, because of the tipsy tooltip.

sshanzel: In this case, it would probably be better to turn the `arrow` property to a `ReactNode` so we can pass this custom component instead.But will leave it up to you to decide.

capJavert: I am not a fan of props doing two things or having union types like this because it gets messy real fast. But refactoring `arrow` to ReactNode exclusively  would make sense. I would also prefer to do it outside of this task because it is not required for any core functionality. 

If both of you think we should do it here I am up for it also. 👐

rebelchris: Happy for another task :)

sshanzel: I agree with that. I think we only had one place where we introduced something similar, I just thought of backward compatibility. Making it `ReactNode` exclusively as you mentioned would be the best option. You will just have to refactor existing implementations with `arrow={false}` to `arrow={<></>}`. So the ones using default would still yield the expected result for deciding whether to display the prop value or the original one.

sshanzel: @rebelchris would love to hear what you think.

rebelchris: Think it's a minor thing for another task really.
Don't want to overcomplicate this PR in particular.

sshanzel: Just my two cents, I believe overrides are okay for one or two classes, but when it becomes many, I think it will defeat the purpose of the original component/element, but the problem is, where do we draw the line where it is still considered acceptable.

sshanzel: > Think it's a minor thing for another task really. Don't want to overcomplicate this PR in particular.

True. We can probably do that in a separate PR/task.

packages/shared/src/components/sidebar/SidebarBottomSection.spec.tsx
capJavert: Just a leftover from copy pasting and initial playing with tests setup.. I will remove `act` from places where it's not needed.

packages/shared/src/components/tooltips/ChangelogTooltip.tsx
capJavert: Added to backlog.

packages/shared/src/components/AlertDot.tsx
```diff
@@ -16,10 +17,13 @@ interface AlertDotProps {
...
-}: AlertDotProps): ReactElement {
-  return <Dot {...props} className={classNames(color, className)} />;
-}
+export const AlertDot = forwardRef<HTMLElement, AlertDotProps>(
+  function AlertDot(
+    { color, className, ...props }: AlertDotProps,
+    ref,
+  ): ReactElement {
+    return (
+      <Dot ref={ref} {...props} className={classNames(color, className)} />
```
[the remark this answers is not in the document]
capJavert: This is to anchor changelog tooltip to the dot in the sidebar. You can see it in `SidebarBottomSectionSection`

capJavert:
Tests failing due to https://github.com/dailydotdev/daily-api/pull/1189#issuecomment-1442124835

packages/shared/src/components/tooltips/ChangelogTooltip.tsx
```diff
@@ -0,0 +1,161 @@
...
+      if (typeof sendMessage === 'function') {
+        const updateResponse: { status: string } = await sendMessage({
+          type: ExtensionMessageType.RequestUpdate,
+        });
+
+        const toastMessageMap = {
+          throttled: 'There is no update available, try again later',
+          no_update: 'You are already on the latest available version',
+          update_available: 'Browser extension updated',
+        };
```
rebelchris, line 48: This can actually be on top of the file, outside the component.

rebelchris (changes requested):
Nice work, guessing you are still working on the changelog data change right?

packages/shared/src/components/tooltips/ChangelogTooltip.tsx
```diff
@@ -0,0 +1,161 @@
...
+                <Button
+                  tag={isFirefoxExtension ? 'a' : undefined}
+                  href={
+                    isFirefoxExtension ? updateFirefoxExtensionLink : undefined
+                  }
+                  className="bg-cabbage-40 btn-primary"
+                  data-testid="changelogExtensionBtn"
+                  onClick={async () => {
+                    await updateExtension();
+                  }}
```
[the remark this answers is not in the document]
capJavert: Just TS complaining if I just passed mutationAsync function.

packages/shared/src/components/tooltips/ChangelogTooltip.tsx
```diff
@@ -0,0 +1,161 @@
...
+
+  const { mutateAsync: updateExtension } = useMutation(async () => {
+    if (isFirefoxExtension) {
+      return;
+    }
+
+    if (isExtension) {
+      const sendMessage = globalThis?.browser?.runtime?.sendMessage;
+
+      if (typeof sendMessage === 'function') {
```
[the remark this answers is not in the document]
capJavert: There is a check for rendering that we are in the extension context and in extension context `runtime` will be available.

packages/shared/src/components/tooltips/ChangelogTooltip.tsx
```diff
@@ -0,0 +1,161 @@
...
+                New release
+              </Button>
+              <ModalClose
+                onClick={(event) => {
+                  if (typeof onRequestClose === 'function') {
+                    onRequestClose(event);
+                  }
+
+                  dismissChangelog();
+                }}
```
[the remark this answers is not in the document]
capJavert: If you guys prefer named functions above JSX I can move it. Seen both cases throughout the codebase.

packages/shared/src/components/tooltips/ChangelogTooltip.tsx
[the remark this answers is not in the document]
capJavert: Looked at it agian, since I do check if function exists i'll add fallback.

packages/shared/src/graphql/alerts.ts
```diff
@@ -6,6 +6,8 @@ export type Alerts = {
   myFeed?: string;
   squadTour?: boolean;
   companionHelper?: boolean;
+  lastChangelog?: string;
+  changelog?: boolean; // read only, comes from boot as computed property
```
capJavert, line 10: Note: add UpdateAlerts interface

capJavert:
Have to check the tests, for some reason they are failing on CI and locally they are passing 🤔 , looks like a race condition of some sorts.

capJavert:
> Have to check the tests, for some reason they are failing on CI and locally they are passing 🤔 , looks like a race condition of some sorts.

Tests were failing because of the shared `QueryClient` and `client.clear` call in `afterEach`. Looks like in CircleCI even though jest is set to `runInBand` there were still issues because when the client queries were cleared `useQuery` would try to fetch data and would get cancelled (since test runtime was destroyed).

rebelchris (approved):
Wohoo! Happy to say it looks to be complete! 🥳

PR #1610, opened 2023-02-13
feat: enable react query devtools

capJavert:
Just enabling react query devtools. Devtools are automatically excluded from production bundle by `react-query`. 
- https://tanstack.com/query/v4/docs/react/devtools#install-and-import-the-devtools
- https://github.com/TanStack/query/blob/main/packages/react-query-devtools/src/index.ts

Also, there is no need to `dynamic` import because in production devtools component is empty so import breaks.

 I think they are really useful to just see all the active queries on the page and then working back and finding them inside the codebase. Especially when someone is just starting.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done


idoshamun (approved):
I think the dynamic import was for disabling it to work on SSR but if you validated that it works then we should be cool!

sshanzel (approved):
Thanks for taking this on! 🚢

PR #1615, opened 2023-02-16
feat: squad tour

sshanzel:
- Introduced a new component named `Carousel`.
- Introduced a new component named `InteractivePopup` for popups like on desktop mode. Can also be used on the likes of CookieBanner, etc.
- The slide event only happens on touch devices (tablet/mobile).
- ~~Squad tour should only be displayed on the Squad page only (still in discussion)~~.

Currently deployed at preview2.
Quick link: https://preview2.app.daily.dev/squads/dailydev

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-950 #done


packages/shared/src/styles/base.css
```diff
@@ -32,6 +32,8 @@ body {
       padding-right: 0.5rem;
     }
   }
+  
+  overflow-x: hidden;
```
sshanzel, line 36: We never had a horizontal display that requires the user to scroll the page. Applying this property because without this the horizontal scroll is resulting in unnecessary jumpyness.

Preview without this:

https://user-images.githubusercontent.com/13744167/219273409-92f03934-2f8a-4706-8060-cd02233621fd.mov

sshanzel:
I have merged the two tickets relating to the Squad tour. Though the only commit related to [WT-910](https://dailydotdev.atlassian.net/browse/WT-910) is this commit: https://github.com/dailydotdev/apps/pull/1615/commits/97867903b151f7c4b44935fb9ac604987ec8ff4f

packages/shared/src/lib/image.ts
```diff
@@ -37,5 +37,17 @@ export const cloudinary = {
   squads: {
     imageFallback:
       'https://daily-now-res.cloudinary.com/image/upload/v1672041320/squads/squad_placeholder.jpg',
+    tour: {
```
rebelchris, line 40: In preview these all seem low quality?

packages/shared/src/lib/storage.ts
```diff
@@ -0,0 +1,9 @@
+export enum StorageTopic {
+  Squad = 'squad',
+}
+
+export const generateStorageKey = (
+  topic: StorageTopic,
+  key: string,
+  identifier = 'anonymous',
+): string => `${topic}:${key}:${identifier}`;
```
rebelchris, line 9: Nice do we just document this new behaviour somewhere?
I see our indexDB has 4 different variants now.

packages/webapp/pages/squads/[handle]/index.tsx
```diff
@@ -114,6 +130,7 @@ const SquadPage = ({ handle }: SourcePageProps): ReactElement => {
 
   return (
     <ProtectedPage seo={seo} fallback={<></>} shouldFallback={!user}>
+      {isPopupOpen && <SquadTourPopup onClose={onClosePopup} />}
```
Comment (rebelchris) [inline packages/webapp/pages/squads/[handle]/index.tsx:133]: Why don't you use the useModal here?

rebelchris (changes requested):
Could you explain to me how you do the switching from the bottom right to center?

packages/shared/src/lib/storage.ts
sshanzel: I agree, I also found it hard to think which one to follow, so instead, I introduced this to keep it consistent in the future. Let me write this down somewhere on our Docs page.

packages/shared/src/lib/image.ts
sshanzel: I think that is only on the preview. They look pretty alright on normal browsing.

packages/webapp/pages/squads/[handle]/index.tsx
Comment (sshanzel) [reply packages/webapp/pages/squads/[handle]/index.tsx:133]: There was a long discussion about whether should this be modal or not. The conclusion was to introduce a new popup since the behavior is different. The user should be able to interact with the components outside this popup and keep it open until the user explicitly closes it which our modal doesn't allow due to having an overlay.

sshanzel:
> Could you explain to me how you do the switching from the bottom right to center?

On mobile devices, this will use the modal, which is given by the `sidebarRendered`. On larger screen sizes, this will use the popup. There is currently no active mechanism to check whether the user was from a desktop and resized the browser size to mobile. Do you think we should add something to catch it?

packages/webapp/pages/squads/[handle]/index.tsx
Comment (rebelchris) [reply packages/webapp/pages/squads/[handle]/index.tsx:133]: Oh I see, but it's different from the SquadTour modal?

Comment (sshanzel) [reply packages/webapp/pages/squads/[handle]/index.tsx:133]: Yes, they will be two different components but using the same content as `SquadTour`.

Comment (rebelchris) [reply packages/webapp/pages/squads/[handle]/index.tsx:133]: Got it then, makes sense.

Seems to work fine while testing it locally.

rebelchris (approved):
Please do wait for second review on this one.

sshanzel:
@idoshamun would love to hear what you think of this PR 🚢

idoshamun:
@sshanzel if you got green light from Chris you can go ahead unless there's something specific you want me to review :)

sshanzel:
> @sshanzel if you got green light from Chris you can go ahead unless there's something specific you want me to review :)

Awesome. Cool.

@rebelchris should we now merge?

rebelchris:
@sshanzel Yeah feel free to merge :)

sshanzel:
Glad to see the tests we've implemented worked, so far.

@rebelchris I just fixed the test and I will be merging this in a few minutes from now. Will also give product a heads-up.

PR #1620, opened 2023-02-17
fix: sidebar spacing between discover and squads

sshanzel:
- There was a missing spacing between the discover section and the list of squads

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1071 #done


rebelchris (changes requested):
Not sure this is the right fix Lee,
The issue only happens on the mobile menu, this fix seems to affect both right?

sshanzel:
I am not yet aware of the issue on mobile, but the fix here should be on the desktop size which was discussed here:

https://dailydotdev.slack.com/archives/C01L0QXQJNB/p1675951039937079

sshanzel:
@rebelchris yes, this should fix both cases. On mobile and on desktop there was a missing margin.

image

image

rebelchris:
The design is only mobile ;)
It should not change the desktop size as far as I can tell. that was already good.

sshanzel:
> The design is only mobile ;) It should not change the desktop size as far as I can tell. that was already good.

Oh, from the design here, there seems to be a margin in-between.

image

rebelchris (approved):
Hmm seems changed since last viewed, so good to go.

sshanzel:
> Hmm seems changed since last viewed, so good to go.

Let me confirm with the design team just to be 101% sure. This can probably be for introducing custom feeds in the future, but they can better answer it. I will get back to this.

sshanzel:
@rebelchris confirmed! it seems they have made a recent change to the spacing.

PR #1622, opened 2023-02-17
refactor: author fragments usage and cleanup

sshanzel:
- We've had some bugs in the past where some features don't work. Such as the tooltip that requires the user's id, when no other information is available.
- The root cause was, the id was also missing in the data from the server by not having the `id` in the query itself.
- To make sure every author tooltip will work, we created a fragment to be used across the queries.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-77 #done


rebelchris (approved):
Thanks for fixing this, this will make a huge difference in consistency.

PR #1629, opened 2023-02-21
feat: feed anonymous users scroll experiment

capJavert:
Adjustments from the latest design updates

### Todo
- [x] Check event triggers that Ido mentioned

## Events

| None| eligible scroll block| |

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1069 #done


packages/shared/src/components/ScrollFeedFiltersOnboarding.tsx
```diff
@@ -4,48 +4,69 @@ import { Button } from './buttons/Button';
...
+}): ReactElement => {
+  return (
+    <div
+      className={classNames(
+        'absolute w-48 h-48 rounded-full blur-2xl opacity-[0.2] bg-theme-color-cabbage',
+        versionToGaussianBlurClassName[version],
+      )}
+    />
+  );
+};
```
sshanzel, line 44: This looks to be something we can reuse in other places. Let's take this out and dedicate a file. Let's just remove some of the custom classes and make it accept a `className` property for modifications.

packages/shared/src/components/ScrollFeedFiltersOnboarding.tsx
```diff
@@ -4,48 +4,69 @@ import { Button } from './buttons/Button';
...
+import { AnalyticsEvent } from '../lib/analytics';
 
-const GaussianBlur = (): ReactElement => {
-  return (
-    <div className="absolute w-52 h-52 rounded-full opacity-32 blur-2xl bg-theme-color-cabbage" />
-  );
-};
+type ScrollOnboardingVersionMap = Partial<
+  Record<ScrollOnboardingVersion, string>
+>;
```
sshanzel, line 11: I think we can remove the partial.

```suggestion
type ScrollOnboardingVersionMap =  Record<ScrollOnboardingVersion, string>;
```

packages/shared/src/components/Feed.tsx
```diff
@@ -199,6 +199,7 @@ export default function Feed<T>({
   const showScrollOnboardingVersion =
     sidebarRendered &&
     alerts?.filter &&
+    !user?.id &&
     Object.values(ScrollOnboardingVersion).includes(scrollOnboardingVersion);
   const fetchPageFn = showScrollOnboardingVersion
```
sshanzel, line 204: I think we will need this since the new default value is string as well which would be true in this case.
```suggestion
  const fetchPageFn = showScrollOnboardingVersion === ScrollOnboardingVersion.Control
```

sshanzel: I think it would be better to apply the condition in the `canFetchMore` property as well - should be along the lines below: https://github.com/dailydotdev/apps/blob/e8ce0cdd2083dae4ff6ca1f950bbdd9b00063a35/packages/shared/src/components/Feed.tsx#L204-L206

Because the `useInfiniteScroll` would probably think you are currently at the place where it should refetch, and might be executing the function endlessly.

```
const infiniteScrollRef = useFeedInfiniteScroll({
    fetchPage: fetchPageFn,
    canFetchMore: canFetchMore && showScrollOnboardingVersion !== ScrollOnboardingVersion.Control,
  });
```

sshanzel (changes requested):
Great start on this feed filters experiment! 🚢 

Since we have changed the value for the control, we should also check existing implementations on where conditions are still applicable.

packages/shared/src/components/Feed.tsx
```diff
@@ -199,6 +199,7 @@ export default function Feed<T>({
   const showScrollOnboardingVersion =
     sidebarRendered &&
     alerts?.filter &&
+    !user?.id &&
     Object.values(ScrollOnboardingVersion).includes(scrollOnboardingVersion);
```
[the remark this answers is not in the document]
idoshamun: The trigger looks good! Once the PR is concluded, I'd like to do a round of manual tests to see that everything is fine

packages/shared/src/components/ScrollFeedFiltersOnboarding.tsx
```diff
@@ -4,48 +4,69 @@ import { Button } from './buttons/Button';
...
-      target_type: TargetType.ScrollBlock,
+      event_name: AnalyticsEvent.EligibleScrollBlock,
       target_id: version,
     });
-  }, []);
+  }, [version]);
+
+  if (version === ScrollOnboardingVersion.Control) {
+    return null;
+  }
```
rebelchris, line 64: We don't need to define control if it's a empty version.
People will simply not get it active.

packages/shared/src/lib/featureManagement.ts
```diff
@@ -141,9 +141,13 @@ export class Features<T extends FeatureValue = string> {
   static readonly ShowCommentPopover = new Features('show_comment_popover');
 
   static readonly ScrollOnboardingVersion = new Features(
-    'scroll_onboarding_version',
-    null,
-    [ScrollOnboardingVersion.V1, ScrollOnboardingVersion.V2],
+    'anonymous_users_scroll_home_page',
+    ScrollOnboardingVersion.Control,
```
rebelchris, line 145: Think we can revert it back to null, people will simply not get it.

rebelchris (changes requested):
Almost there, doing some last checks locally.
Do check Lee's comments as well.

packages/shared/src/components/ScrollFeedFiltersOnboarding.tsx
capJavert: From the analytics events excel sheet it was defined that `control`, `v1` or `v2`  should be sent.

rebelchris: @idoshamun Pretty sure we should not fire event if user doesn't even get it right?

packages/shared/src/components/ScrollFeedFiltersOnboarding.tsx
capJavert: This was already here from the previous PR I just adjusted it. With exposed className only I think it is not reusable in this state and for sure not giving any extra value if we just add `className`.

Some of my ideas for `GaussianBlur` component:
- should in this case be called something like `GaussianBlurRadius` since this one is specific to that version of blur
- expose radius as `prop` (so people do not need to add classes and it feels more like a component and not just `classed`/`classNames` wrapper
- expose width/height
- expose oppacity
- expose color

I think this effort is better suited for another PR.

packages/shared/src/components/ScrollFeedFiltersOnboarding.tsx
capJavert: I think this is due to `control` also being logged with the new instructions for `eligible scroll block` event.

capJavert: Had a chat with @idoshamun about it yesterday (you can look at the description of the event in the sheet).

packages/shared/src/components/ScrollFeedFiltersOnboarding.tsx
capJavert: WDYT?

packages/shared/src/components/ScrollFeedFiltersOnboarding.tsx
idoshamun: @rebelchris we should fire the event even if the user doesn't get it. This is required for an accurate analysis.

rebelchris: Alright, just something we normally don't do so it was out of the ordinary.

packages/shared/src/components/ScrollFeedFiltersOnboarding.tsx
rebelchris: My bad probably, I do think these would work great as individual components, maybe for now simply allow classnames option?
Eventually we can abstract with the options you mention.

packages/shared/src/components/ScrollFeedFiltersOnboarding.tsx
capJavert: Added the `Partial` because of `ScrollOnboardingVersion.Control`, did not want to add `''` mappings to each map object. Related to this thread https://github.com/dailydotdev/apps/pull/1629#discussion_r1113918285 `classNames` will take care of `undefined` below.

packages/shared/src/components/ScrollFeedFiltersOnboarding.tsx
capJavert: Since I have time I can add it today and cut our backlog some slack 😆

packages/shared/src/lib/featureManagement.ts
capJavert: Resolved here https://github.com/dailydotdev/apps/pull/1629#discussion_r1114001422

packages/shared/src/components/Feed.tsx
capJavert: Didn't really look at this part of the code, but good catch.

capJavert: Adjusted it even further without `useState`. https://github.com/dailydotdev/apps/pull/1629/commits/d9a87c94125721999cc85e2cd048a3da7fff4ba6

packages/shared/src/components/Feed.tsx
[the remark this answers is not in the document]
capJavert: @idoshamun you can test this manually when you get the chance, as most of the other comments are resolved and will not interfere with the core implementation.

packages/shared/src/components/ScrollFeedFiltersOnboarding.tsx
capJavert: Had some issues with typescript while implementing this, I am gonna show you the details maybe in the web team meet so we can discuss, I think its a curious case which I think we don't have in the codebase yet. Probably stumbled upon it since I am used to working with JSS libraries instead of tailwind.

packages/shared/src/components/Feed.tsx
sshanzel: Yeah, just remembered it, sorry for having it requested now. Thank you for understanding 🙏

capJavert: No worries, it's definitely better this way.

packages/shared/src/components/ScrollFeedFiltersOnboarding.tsx
```diff
@@ -4,48 +4,69 @@ import { Button } from './buttons/Button';
...
-  );
-};
+type ScrollOnboardingVersionMap = Partial<
+  Record<ScrollOnboardingVersion, string>
+>;
 
-const versionToContainerClassName: Record<ScrollOnboardingVersion, string> = {
-  [ScrollOnboardingVersion.V1]: 'h-60',
-  [ScrollOnboardingVersion.V2]: 'h-[36rem]',
+const versionToContainerClassName = {
```
rebelchris, line 13: Think we are missing `: ScrollOnboardingVersionMap` here.

rebelchris (changes requested):
One small remark.

packages/shared/src/components/Feed.tsx
```diff
@@ -199,14 +197,12 @@ export default function Feed<T>({
...
-  const fetchPageFn = showScrollOnboardingVersion
-    ? () => setShowFeedFilters(true)
-    : fetchPage;
 
   const infiniteScrollRef = useFeedInfiniteScroll({
-    fetchPage: fetchPageFn,
-    canFetchMore,
+    fetchPage,
+    canFetchMore: !showScrollOnboardingVersion && canFetchMore,
```
sshanzel, line 205: On `Control` they should be able to scroll, still.
```suggestion
    canFetchMore: (!showScrollOnboardingVersion || scrollOnboardingVersion === ScrollOnboardingVersion.Control) && canFetchMore,
```

packages/shared/src/components/ScrollFeedFiltersOnboarding.tsx
sshanzel: Good point. Thought of making it consistent since on line 28 we had an empty string too. But this is nothing blocking, this comment should be resolved! 💯

idoshamun:
@capJavert reporting issues from qa:
* If I'm on control, I still can't scroll past the first page. This is not the intended behavior. Control should behave the same.
* Scroll block should affect only the popular feed (please verify with Sab)
* When the feed is loading I can see the scroll block element above the fold
* The image below is blurry 
image
* I know you asked design about the topics the look clickable. Can you point me to the discussion please? I couldn't find it

sshanzel (changes requested):
Great work on having one-less usage of state! Way to go! 🙌 

Just had one last comment, and everything should be cool from my end. Thank you for the extra mile! 🚢

packages/shared/src/components/Feed.tsx
capJavert: Just realised this from the Ido's comment. Will fix.

sshanzel: Oh, just saw it. Yep, that should fix what he mentioned.

packages/shared/src/components/Feed.tsx
```diff
@@ -198,15 +203,17 @@ export default function Feed<T>({
 
   const showScrollOnboardingVersion =
     sidebarRendered &&
+    feedName === 'popular' &&
```
capJavert, line 206: I know this is hardcoded but currently it's like that in the rest of the codebase. Could be beneficial to add some kind of enum for feed types we have?

packages/shared/src/hooks/useFeed.ts
```diff
@@ -235,5 +236,6 @@ export default function useFeed<T>(
     canFetchMore: feedQuery.hasNextPage,
     emptyFeed:
       !feedQuery?.data?.pages[0]?.page.edges.length && !feedQuery.isFetching,
+    isLoading: feedQuery.isLoading,
```
capJavert, line 239: Exposed `isLoading` from hook. Needed it to check initial load state in the `Feed`. This is not breaking anything else.

packages/shared/src/components/Feed.tsx
sshanzel: I think we have an we introduced one recently, it should be named `enum FeedPage`.

rebelchris: Correct. `MainFeedPage.Popular`

packages/shared/src/components/Feed.tsx
```diff
@@ -198,15 +203,17 @@ export default function Feed<T>({
...
+    !isLoading &&
     alerts?.filter &&
+    !user?.id &&
     Object.values(ScrollOnboardingVersion).includes(scrollOnboardingVersion);
-  const fetchPageFn = showScrollOnboardingVersion
-    ? () => setShowFeedFilters(true)
-    : fetchPage;
+  const shouldScrollBlock =
+    scrollOnboardingVersion !== ScrollOnboardingVersion.Control;
```
rebelchris, line 212: What will this evaluate if it's `null` or `undefined`?
Just to make sure when we might clear it at one stage.

packages/shared/src/components/ScrollFeedFiltersOnboarding.tsx
```diff
@@ -4,66 +4,95 @@ import { Button } from './buttons/Button';
...
   const { trackEvent } = useContext(AnalyticsContext);
   useEffect(() => {
     trackEvent({
-      event_name: AnalyticsEvent.Impression,
-      target_type: TargetType.ScrollBlock,
+      event_name: AnalyticsEvent.EligibleScrollBlock,
       target_id: version,
     });
-  }, []);
+  }, [version]);
```
rebelchris, line 61: Thinking about this, we might need to add a state to ensure it only triggers once, not on re-render.
There are some examples of tracking where we do that.

packages/shared/src/lib/image.ts
```diff
@@ -11,8 +11,6 @@ export const cloudinary = {
       'https://daily-now-res.cloudinary.com/image/upload/v1665224467/public/recommended.png',
     yourFeed:
       'https://daily-now-res.cloudinary.com/image/upload/v1667998924/public/your_feed.png',
-    supercharge:
-      'https://daily-now-res.cloudinary.com/image/upload/v1665563545/public/supercharge.png',
```
rebelchris, line 15: I think it might be best to load the SVG from cloudinary as well, instead of in the files.
CC: @sshanzel / @idoshamun what do you guys think?

rebelchris (changes requested):
Amazing work, almost there on this one.

packages/shared/src/lib/image.ts
idoshamun: The good thing about cloudinary is the image optimizations which are not applied on svg. So for me it doesn't matter. Your call :)

packages/shared/src/components/Feed.tsx
capJavert: `Control` is defiend as default in `src/lib/featureManagement.ts`, so only if we change that then this would not work. But since `Control` is here what `null` usually is for other tests I think its fine.

packages/shared/src/components/ScrollFeedFiltersOnboarding.tsx
capJavert: As long as component does not remount this won't trigger again. This will trigger if user goes to other page and then back to the homepage (because of the unmount of course), but I think in that case it should trigger since users encounters the scroll block again.

If we want to trigger it only once per session let me know.

packages/shared/src/components/Feed.tsx
rebelchris: Yep agreed, but we have to make sure if we ever remove the flag (because there is a winner maybe) what will happen.
I guess we can then also modify this, but maybe it's safer to do the evaluation the otherway around?

To check if [V1, V2].includes(scrollOnboardingVersion).

Just a next level of security for someone who might join later, and is not aware of it.

packages/shared/src/components/ScrollFeedFiltersOnboarding.tsx
capJavert: also `version` currently does not change until next full page reload so this is just a case if in future we have a dynamic experiments added so it actually responds to the version change, more of a future proofing.

packages/shared/src/lib/image.ts
capJavert: I think generally bundling svg with code is better. Since we lazy load by default it is even better for that case.

packages/shared/src/components/ScrollFeedFiltersOnboarding.tsx
rebelchris: Ah yep for this type of component it's actually alright, never mind this comment.

rebelchris:
@capJavert On mobile it simply doesn't show the scroll block, but blocks scrolling.
On tablet it has a different version.

packages/shared/src/components/Feed.tsx
capJavert: Fine with me and it will help you sleep easier 😴 😆

packages/shared/src/contexts/OnboardingContext.tsx
```diff
@@ -138,10 +148,26 @@ export const OnboardingContextProvider = ({
...
+    ],
   );
 
+  // if onboarding is done/closed reset
+  // state values to default
+  useEffect(() => {
+    if (!isOnboarding) {
+      setSkipIntro(false);
+    }
+  }, [isOnboarding]);
```
sshanzel, line 169: We should avoid the usage of `useEffect` whenever possible. Let's probably create a function to wrap the calling of `setIsOnboarding(false)` to also execute `setSkipIntro(false)` at the same time. This would also benefit us once we move to React 18.

sshanzel (changes requested):
Looks amazing with all the changes! Just the `useEffect` that we can avoid and the blur component that we will talk about soon 🚢

packages/shared/src/contexts/OnboardingContext.tsx
capJavert: For state syncing like this, moving it to event handler is better I agree.

packages/shared/src/components/Feed.tsx
```diff
@@ -198,23 +204,28 @@ export default function Feed<T>({
...
+    canFetchMore: canFetchMore && !shouldScrollBlock,
   });
 
   const onInitializeOnboardingClick = () => {
     trackEvent({
       event_name: AnalyticsEvent.ClickScrollBlock,
       target_id: scrollOnboardingVersion,
     });
-    onInitializeOnboarding();
+    onInitializeOnboarding(undefined, true);
```
rebelchris, line 228: I actually think we might want to redirect the user to `my-feed` if they onboarded via this way.
There is an example of that already, but maybe double check with product.

rebelchris (approved):
One small remark, just something to verify :)

packages/shared/src/components/Feed.tsx
capJavert: Tested it and It already does that, not sure from where but I can check in detail.

capJavert: checked with Seb, and this already does that so nothing is needed.

rebelchris: Hmm yeah the first param is the onFinish right?
I think in other implementations we do something with it.
Just to ensure that's how it works

capJavert: yeah and its undefined by default, something other redirects though.. I can find what exactly if you want?

sshanzel (approved):
Massive work on your very first experiment! 🚢 🚢 🚢

PR #1644, opened 2023-03-01
feat: changelog design adjustments

capJavert:
- properly show upvotes comments in changelog
- position changelog and badge properly when sidebar is closed
- some other style/sizing adjustments mentioned by Tomer

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done


capJavert:
@sshanzel I think `@tippyjs/react` has some issues with reference handling and positioning. Partially due to ref being available and then again not available but mainly if anchor for the tooltip changes position looks like it breaks the position of the tooltip. Did you encounter anything like this? 

https://user-images.githubusercontent.com/9803078/222295780-ecc73cda-aa2b-4fe4-8e24-6f21f3fc526f.mov


I tried so far:
- making sure ref to anchor is always available eg. element that holds the ref never unmounts
  - it gets better but position is still not updated instantly, but it does update if you scroll or do other interactions -> video below

https://user-images.githubusercontent.com/9803078/222295832-ec84c9c1-f297-4916-b4a4-4a473cb9d004.mov

Other things I tried:
- enable https://atomiks.github.io/tippyjs/v6/all-props/#sticky -> did not help
- try to force remount of `BaseTooltip` component (and tippy instance under the hood) basically hacky way with `useState` and `useEffect` and `setTimeout` but that was just to confirm that when tooltip actually remounts it gets the correct position -> which it does
  - additionally here I noticed that based on `ms` I put in `setTimeout` to delay remount of the tooltip it gets stuck on different positions 🤣

sshanzel:
@capJavert thank you for elaborating on what you have tried and researched. Have you tried using the reference of the actual item (`li`)? So that the reference point will never change and always be shown. Then position the tooltip to the right and apply some negative X offset if necessary.

capJavert:
> @capJavert thank you for elaborating on what you have tried and researched. Have you tried using the reference of the actual item (`li`)? So that the reference element will never change/go missing and always be shown. Then position the tooltip to the right and apply some negative X offset if necessary.

@sshanzel yeah, but I get the same problem where it breaks and then fixes on scroll or if I force remount of tooltip (kinda like second video). That fact that it fixes on scroll is really weird, feels like tippy has some logic for adjusting position but it does not work correctly 🤔 

But I think I found the solution. Basically combining `sticky` plugin with anchoring element to the `li` (stable element that is always present) somehow makes it so that position updates and everything works. 🚀 I did have to add some more offset adjustments but I guess it is what it is. 

I do think that we should be on the lookout for other tooltip use cases in the future since obviously it has some issues when anchored element is moving.

packages/shared/package.json
```diff
@@ -117,6 +117,7 @@
     "lodash.clonedeep": "^4.5.0",
     "node-fetch": "^2.6.6",
     "react-onesignal": "^2.0.4",
+    "tippy.js": "^6.3.1",
```
capJavert, line 120: Added as dependency just so we can import inside `ChangelogTooltip` to access `sticky` plugin. It does not add anything to bundle size because `tippy.js` was already a dependency of `@tippyjs/react`, it was just a matter of linter complaining.

sshanzel:
> I do think that we should be on the lookout for other tooltip use cases in the future since obviously it has some issues when anchored element is moving.

Really agree with this, once we find the root cause on how it happens, we can write something up to explain what we can do against it.

> But I think I found the solution. Basically combining sticky plugin with anchoring element to the li (stable element that is always present) somehow makes it so that position updates and everything works. 🚀 I did have to add some more offset adjustments but I guess it is what it is.

Nice find! Let me know once this is ready for review and I'd be glad to take another peek 🚢

capJavert:
@sshanzel @rebelchris this should be good to go, there is a one small thing left that can be done in a separate PR, but. I should show `summary` in the changelog pop-up but from what I can see `SOURCE_FEED_QUERY` does not return that field.. it makes sense since it's a listing query and summary is not showed.

Having said that, the simple stupid way of fetching a summary for the changelog pop-up would be to take a post `id` from existing `getLatestChangelogPost` function and then do another call to `Post` query which includes `summary`.

Is there some query already better suited for this?

rebelchris:
@capJavert I would suggest we event modify the `getLatestChangelogPost` so it includes a new query.
Which uses the existing `SOURCE_FEED_QUERY`. but extends it with summary?

packages/shared/src/components/sidebar/common.tsx
```diff
@@ -135,18 +137,22 @@ export const ItemInner = ({
...
-      <span
-        className={classNames(
-          'flex flex-1 text-left transition-opacity truncate flex-row items-center',
-          shouldShowLabel ? 'opacity-100 delay-150' : 'opacity-0',
-          item?.className?.text,
-        )}
-      >
-        {item.title}
-      </span>
+      {sidebarExpanded && (
```
sshanzel, line 145: I think this will take away the current animation on our Sidebar. Though not sure if it is high priority to be kept. cc @rebelchris 

The `shouldShowLabel` should handle whether this label is actually showed, so we might want to pass this `sidebarExpanded` property in the items though I believe it is already included. I could be wrong though.

packages/shared/src/components/AlertDot.tsx
```diff
@@ -11,19 +11,33 @@ export enum AlertColor {
...
-  ): ReactElement {
-    return (
-      <Dot ref={ref} {...props} className={classNames(color, className)} />
-    );
-  },
-);
+const alertSizeMap: Record<AlertDotSize, string> = {
+  [AlertDotSize.XSmall]: 'w-2 h-2',
+  [AlertDotSize.Small]: 'w-2.5 h-2.5',
+};
```
rebelchris, line 29: Might be worthwhile checking with design if it's intended there are 2 different sizes or a design glitch.

rebelchris (changes requested):
Nice just a generic question.
And do think we need to sort the summary in this PR just because it's quite important to nail before launching.

packages/shared/src/components/tooltips/ChangelogTooltip.tsx
```diff
@@ -122,8 +138,29 @@ function ChangelogTooltip<TRef extends HTMLElement>({
...
+                  <UpvoteIcon secondary />
+                  <InteractionCounter value={post.numUpvotes} />
+                </span>
+                <span
+                  className="flex gap-1.5"
+                  data-testid="changelogCommentsCounter"
+                >
+                  <CommentIcon secondary />
+                  <InteractionCounter value={post.numComments} />
+                </span>
```
sshanzel, line 163: We should avoid using the `.5` values in our tailwindcss as our spacing should only be divisible by 4. When these cases arise, the design team mentioned to tag them in Slack for visibility and to be able for them to correct it.

sshanzel (changes requested):
Looks amazing with the fix! 💪 

Just a general concern with using values not being divisible by four.

packages/shared/src/components/sidebar/common.tsx
rebelchris: Oh well spotted, yes please make sure this doesn't remove the animation.

sshanzel: We used to have a hook to manage these scenarios and allow animation on dynamically loaded content. Kindly check the link below if it would be of any use to you:
packages/shared/src/hooks/useDynamicLoadAnimated.ts

packages/shared/src/graphql/posts.ts
```diff
@@ -392,15 +392,39 @@ export const VIEW_POST_MUTATION = gql`
...
+          image
+          permalink
+          numComments
+          numUpvotes
+          summary
+        }
+      }
+    }
+  }
+`;
```
capJavert, line 416: @rebelchris let me know if this is good

packages/shared/src/components/sidebar/common.tsx
capJavert: Looked into this.. honestly, with `150ms` delay for that opacity animation it is almost invisible to the human eye and also not worth it. Even with this hook I don't think including that much JavaScript is worth it. 

The main reason I did this change is so I can continue to use `rightIcon` prop for the changelog popup anchor. If we wanna retain this animation I think I'll have to add the dot somewhere else because obviously `rightIcon` is not a good choice since it is not visible when sidebar is hidden. 

image

sshanzel: I agree we must put the dot somewhere else. I think we don't even have to use the `rightIcon` anymore since our reference point is the list item. I highly suggest making the `BaseTooltipContainer` accept an `arrow` as `ReactNode` it would be easier on every part as the dot will always be attached to the arrow, and we won't have to worry about whether it is going to be displayed or not since it will be part of the popup that is always on top of the screen.

sshanzel: I think animations are those small things that are not talked about but are well appreciated by the users. For me, I'd love to keep it, but feel free to talk about it with @rebelchris.

capJavert: Agree in general, of course wanna try and do minimal impact on existing features, it is tricky when this kind of elements are introduced.

packages/shared/src/components/modals/common/ModalClose.tsx
```diff
@@ -4,14 +4,20 @@ import { ButtonProps } from '../../buttons/Button';
 import CloseButton from '../../CloseButton';
 
 function ModalCloseComponent(
-  { className, style, onClick, ...props }: ButtonProps<'button'>,
+  {
+    className,
+    style,
+    onClick,
+    buttonSize = 'medium',
```
rebelchris, line 11: This will have changed with my latest guideline changes

packages/shared/src/components/sidebar/common.tsx
rebelchris: I think it's quite noticeable to have and we should keep that behaviour.
Note on opening you see if quite well.

https://user-images.githubusercontent.com/554874/223675861-5335ef4e-612e-484a-b185-56ab8d9a4573.mp4

packages/shared/src/components/tooltips/ChangelogTooltip.tsx
```diff
@@ -84,20 +98,22 @@ function ChangelogTooltip<TRef extends HTMLElement>({
...
+                className="font-normal text-white bg-theme-color-water btn-primary small"
                 data-testid="changelogNewReleaseTag"
               >
                 New release
               </Button>
               <ModalClose
+                className="right-4"
                 onClick={onModalCloseClick}
                 data-testid="changelogModalClose"
+                buttonSize="xsmall"
```
rebelchris, line 116: This will have changed with latest changes

packages/shared/src/graphql/posts.spec.ts
```diff
@@ -67,11 +68,10 @@ it('should return latest changelog post', async () => {
   let queryCalled = false;
   mockGraphQL<MockFeedData>({
     request: {
-      query: SOURCE_FEED_QUERY,
+      query: LATEST_CHANGELOG_POST_QUERY,
       variables: {
         source: 'daily_updates',
         first: 1,
-        loggedIn: false,
         ranking: RankingAlgorithm.Time,
```
rebelchris, line 75: Think we can simply hard set these for the changelog query so nobody will change it down the line.

packages/shared/src/graphql/posts.ts
rebelchris: Looks good to me, just see my suggestion for the variables

packages/shared/src/hooks/useChangelog.ts
```diff
@@ -13,15 +12,12 @@ type UseChangelog = {
...
-  const { user } = useContext(AuthContext);
   const { sidebarRendered } = useSidebarRendered();
 
   const { data: latestPost } = useQuery(
-    ['changelog', 'latest-post', { loggedIn: !!user?.id }] as const,
-    async ({ queryKey }) => {
-      const [, , variables] = queryKey;
-
-      return getLatestChangelogPost(variables.loggedIn);
+    ['changelog', 'latest-post'] as const,
```
rebelchris, line 18: The `as const` is not needed here I don't think?

packages/shared/src/hooks/useChangelog.ts
```diff
@@ -13,15 +12,12 @@ type UseChangelog = {
...
+    async () => {
+      return getLatestChangelogPost();
```
rebelchris, line 20:
```suggestion
    () => getLatestChangelogPost()
```

rebelchris (approved):
Nice, have some small suggestions.
Do think we need to sort that animation.

rebelchris (changes requested):
Some small remarks, and do think we need to sort the animation out.

packages/shared/src/hooks/useChangelog.ts
capJavert: It does not anymore, it was required when queryKey had dynamic parts.

packages/shared/src/graphql/posts.spec.ts
capJavert: Sure, thought about it, thanks for confirming!

packages/shared/src/components/sidebar/common.tsx
capJavert: @sshanzel @rebelchris This should be good now, made changes to BaseTooltip and refactored other usages in c4076a411f3f7fce13b33df28129bf1186a0747c:
* added ArrowComponent for custom arrows
* remove arrowClassname, we don't need it because we can just add classes to custom ArrowComponent
* renamed `arrow` to `showArrow` - I think it better represents what prop means
  * it could have been done with passing `null` to `ArrowComponent` and having only one prop but I think it is more clear this way

rebelchris (approved):
Nice from code point of view, amazing work!
I don't see any blockers, but let's wait for Lee as well.

sshanzel (approved):
This looks so amazing! Thank you for all the effort. Turned out really well 🚢 🚢 🚢

PR #1650, opened 2023-03-02
feat: personal digest

capJavert:
- Frontend implementation for personal digest
- Also adding more general FeedSlider implementation for different feed items
- Thought about adding Slider primitive but I think that `react-swipeable` already has primitive business logic that Slider needs and adding any kind of UI would limit further implementations (maybe later when we think about Sliders in general)

### FeedSlider implementation
- Accepts misc items from `items: FeedSliderItem`
- It does contain logic for feed gap/sizes/numCards automatically, this is why it is called `FeedSlider`
- Allows customisation through `Item` prop which is FunctionalComponent. For example in slider implementation in this PR it handles Intro card on the first index differently (kinda like render handler in native apps), see `DigestPostItem`
- Because all different items are passed in the same Array it is was easier to create a slider logic inside `FeedSlider` component
- It also helps because we can eg. extend `FeedItem` to send intro card content (the one with "Start reading") from the API and control those values there
- All the different states (like opacity change for articles that user read) can be done inside custom `Item` component
- Things like sorting items can of course be done through `items` in the parent component 

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{1109} #done


capJavert:
@sshanzel @rebelchris for this to be merge-able I think for sure I should remove the example implementation from `Feed` component. There are TODO comments as well so if everything else looks good to you I'll clean those up.

capJavert:
I also do need to sort out the loading of the slider, because it jumps quite a bit, probably due to data not being there. Will be easier once API is connected.

capJavert:
Outdated.

PR #1651, opened 2023-03-03
fix: button and icon guidelines

rebelchris:
- Added icon and button size enums
- Made some small fixes according to spec

Review notes:
Old icons where 1 size bigger in naming
So "large" => "medium" in new spec

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1100 #done


packages/shared/src/components/buttons/Button.tsx
```diff
@@ -50,7 +51,7 @@ export interface BaseButtonProps {
 const useGetIconWithSize = (size: ButtonSize, iconOnly: boolean) => {
   return (icon: React.ReactElement<IconProps>) =>
     React.cloneElement(icon, {
-      size: IconSize[size],
+      size: icon.props?.size ?? buttonSizeToIconSize[size],
```
rebelchris, line 54: Had to add this as we have 1 icon (tags) being smaller then guidelines, asked design if this is correct.

packages/shared/src/components/Settings.tsx
```diff
@@ -116,7 +117,9 @@ export default function Settings({
         <CustomSwitch
           inputId="layout-switch"
           name="insaneMode"
-          leftContent={<CardIcon secondary={!insaneMode} />}
+          leftContent={
+            <CardIcon size={IconSize.XSmall} secondary={!insaneMode} />
```
sshanzel, line 121: Was it intended to intorudce the `XSmall` to this one?

packages/shared/src/components/buttons/Button.tsx
sshanzel: Oh, nice catch. Though it seems to be inaccurate in that case. Hope they will rectify it so we won't have to check this.

packages/shared/src/components/fields/PasswordField.tsx
```diff
@@ -61,7 +62,7 @@ export function PasswordField({
       autoComplete="off"
       type={useType}
       valueChanged={onChange}
-      leftIcon={<LockIcon />}
+      leftIcon={<LockIcon size={IconSize.Small} />}
```
sshanzel, line 65: Should this be not needed as the default is `Small` or is it getting overridden by the component?

packages/shared/src/components/squads/SquadHeaderBar.tsx
```diff
@@ -41,7 +42,7 @@ export function SquadHeaderBar({
       <SimpleTooltip placement="top" content="Squad options">
         <Button
           className="tablet:order-2 btn btn-secondary"
-          icon={<MenuIcon size="medium" />}
+          icon={<MenuIcon size={IconSize.Small} />}
```
sshanzel, line 45: Not sure if this is still needed since default is `Small`.

packages/shared/src/components/squads/SquadMemberMenu.tsx
```diff
@@ -36,8 +37,12 @@ export default function SquadMemberMenu({
           rel="noopener noreferrer"
         >
           <span className="flex items-center w-full typo-callout">
-            <FlagIcon size="medium" secondary={false} className="mr-2" /> Report
-            member
+            <FlagIcon
+              size={IconSize.Small}
```
sshanzel, line 41: Same here, we can probably remove the size since the default is `Small`.

packages/shared/src/components/widgets/Alert.tsx
```diff
@@ -58,7 +59,10 @@ function Alert({
       )}
     >
       <span className="flex flex-row items-center">
-        <Icon size="medium" className={classNames('mr-2', fontColor[type])} />
+        <Icon
+          size={IconSize.Small}
```
sshanzel, line 63: We can probably remove this `size` here as well.

packages/webapp/components/FooterNavBar.tsx
```diff
@@ -40,20 +41,24 @@ export const tabs: Tab[] = [
   {
     path: '/',
     title: 'Home',
-    icon: (active: boolean) => <HomeIcon secondary={active} size="xxlarge" />,
+    icon: (active: boolean) => (
+      <HomeIcon secondary={active} size={IconSize.Medium} />
```
sshanzel, line 45: Is this intended to be `Medium` than `Large`? Just commenting for visibility.

sshanzel (changes requested):
Thank you very much for putting in the effort to clean these up. 

Just had a few comments but nothing major. Really helps a lot with maintaining the repo.

packages/shared/src/components/Settings.tsx
rebelchris: Ah yes forgot to mention that, it was way to big originally (the non active one) so adjusted how we scale them in this new fix.

packages/shared/src/components/fields/PasswordField.tsx
rebelchris: You're right, I migrated them all as they where 1 bigger so missed these now being the default value, let me adjust.

rebelchris: Actually default is `IconSize.XSmall` now so this is still right.

packages/webapp/components/FooterNavBar.tsx
rebelchris: Yes thanks, they where way bigger then the design intended :)

packages/shared/src/components/buttons/Button.tsx
rebelchris: Apparently the "Tag" is not a button from design perspective, checking with Tomer now how they want to proceed.

packages/shared/src/components/fields/PasswordField.tsx
sshanzel: Oh, right. So the default previously was small. Then kindly disregard my comments pointing to default values.

sshanzel (approved):
Everything looks good to me!! Feel free to merge once you are finished with the concern involving the design team. Thank you! 🚢

packages/shared/src/components/buttons/Button.tsx
rebelchris: Design needs to look into this and will later on let us know the right decision.
For now let's merge this and retro fix this later.

PR #1657, opened 2023-03-09
feat: submit private links

sshanzel:
## Changes

Will be able to test once the [PR for the relevant API](https://github.com/dailydotdev/daily-api/pull/1203) is merged.

- Introduced a new component for the Squad Post Bar.
- Made the Modal to add a Squad post accept a link and render conditionally.
- Refactored the Modal to use a different endpoint if link was included in the parameters.
- Refactored the Squad header bar.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-WT-1132 #done


packages/shared/src/components/squads/SharePostBar.tsx
```diff
@@ -0,0 +1,88 @@
...
+  onNewSquadPost,
+}: SharePostBarProps): ReactElement {
+  const [url, setUrl] = useState('');
+  const { mutateAsync: getPost } = useMutation(getPostByUrl, {
+    onSuccess: (post) => {
+      onNewSquadPost({ post });
+    },
+    onError: () => {
+      onNewSquadPost({ url });
+    },
```
rebelchris, line 31: Depending on the error we might want to show the error already here.
So example being invalid URL or deleted post.

packages/shared/src/components/modals/PostToSquadModal.tsx
```diff
@@ -91,8 +107,12 @@ function PostToSquadModal({
       steps={post ? undefined : modalSteps}
     >
       <Modal.Header title="Share post" kind={ModalHeaderKind.Tertiary} />
-      {post ? (
-        <SquadComment form={form} onSubmit={onSubmit} isLoading={isLoading} />
+      {shouldSkipHistory ? (
```
rebelchris, line 110: This seems to cause a flicker locally atleast.

packages/shared/src/components/squads/SharePostBar.tsx
```diff
@@ -0,0 +1,88 @@
...
+        size="large"
+        nativeLazyLoading
+      />
+      <input
+        type="url"
+        name="share-post-bar"
+        className="flex-1 pl-1 outline-none bg-theme-bg-transparent text-theme-label-primary focus:placeholder-theme-label-quaternary hover:placeholder-theme-label-primary typo-callout"
+        placeholder="Enter link to share"
+        onInput={(e) => setUrl(e.currentTarget.value)}
+        required
```
rebelchris, line 61: Should be non-required as user can simply press "Post" as empty field.

rebelchris (changes requested):
Nice start!
Some small comments mainly for styling now.

Missing the other header adjustments:
Screenshot 2023-03-09 at 09 48 05

packages/shared/src/components/squads/SharePostBar.tsx
sshanzel: Right. Let me catch it.

packages/shared/src/components/squads/SharePostBar.tsx
sshanzel: Good catch!

sshanzel:
> @sshanzel I also noticed the getPostByUrl seems to throw an error underwater "could not find post" think we need to catch it and proceed as "ok" in those cases.

It only shows as prompt in development, but that is actually caught by the `useMutation` where the `onError` is triggered.

packages/shared/src/components/modals/PostToSquadModal.tsx
sshanzel: Should be okay now. There were two areas triggering the show of the modal.

rebelchris (changes requested):
Some generic remarks:

- Can just click on "Post" (should be possible)
- It validates URL in the header bar already, (browser validation)
- If I remove the URL in the second screen nothing happens (should show that it's required)
- If I paste new URL, then remove it and paste existing article it should show that article on the comment modal.

Let me know if you want me to show you certain things.

sshanzel:
> Let me know if you want me to show you certain things.

I will finish first the concern in the BE then I will get back to this.

sshanzel:
> Can just click on "Post" (should be possible)

Fixed! Thank you!

> It validates URL in the header bar already, (browser validation)

Do we not want to use the browser's native validation? I think it is very useful.

> If I remove the URL in the second screen nothing happens (should show that it's required)

It turns the left border to red as part of being required. Do you want to add an error text below?

image

> If I paste new URL, then remove it and paste existing article it should show that article on the comment modal.

Have finished this by will do more local tests tomorrow first thing.

sshanzel:
@rebelchris private links submission should now work as intended. I have also included a hint for when there is an error.

sshanzel:
Currently deployed at preview2. Quick link: https://preview2.app.daily.dev/squads/dailydotdev

packages/shared/src/components/squads/SquadHeaderBar.tsx
```diff
@@ -39,20 +51,28 @@ export function SquadHeaderBar({
...
-          onClick={onMenuClick}
-          aria-label="Squad options"
+          tag="a"
+          target="_blank"
+          rel="noopener noreferrer"
+          href={`${squadFeedback}#user_id=${user.id}&squad_id=${squad.id}`}
+          className="btn-secondary"
+          icon={<FeedbackIcon aria-label="squad-feedback-icon" />}
+          aria-label="squad-feedback"
         />
```
rebelchris, line 66: We should probably wrap this in a tooltip now as it's not super descriptive what the button does.

rebelchris (changes requested):
Wow, almost there!

Initial state has margin below url with no error:
Screenshot 2023-03-10 at 13 12 42

Error is not red:
Screenshot 2023-03-10 at 13 12 50

packages/shared/src/components/squads/SquadHeaderBar.tsx
sshanzel: Totally agree with that.

sshanzel:
> Initial state has margin below url with no error:

That was actually intended, built-in with our input field, to avoid content shift on responsive error message. Probably I should reduce the bottom padding of the container (if there was any).

sshanzel:
> @sshanzel The only issue I still see is the text no always being red.
It seems to only happen on re-focus.
> Maybe just always make it red, regardless of state?
It's an error so won't ever show differently right.

That is actually part of our design guideline. The red text would only happen on validation.

image

rebelchris (approved):
Maybe also ask ido to review this one.
From that point if looks complete though.

Thanks for the guideline, wasn't aware of that part.

sshanzel:
@rebelchris thank you for the awesome review! 🙇 

@idoshamun as Chris suggested, kindly review if you have some spare time. This is currently deployed at preview2.

Quick link: https://preview2.app.daily.dev/squads/dailydotdev

rebelchris:
@idoshamun One concern I have is that `postByUrl` now throws an error, not sure it's bad but could potentially be improved, so maybe something you can evaluate if it's ok for now?

idoshamun:
@rebelchris @sshanzel what are the cases the user can see an error?

sshanzel:
> @rebelchris @sshanzel what are the cases the user can see an error?

When the URL was invalid or when the URL is empty. Both cases happen after when you focus out which triggers form validation / submitting the form.

@rebelchris let me know if you have seen any responsive issues with the current updates.

PR #1664, opened 2023-03-12
feat(companion): show bookmark button and support share to squad

idoshamun:

* Put bookmark button in the companion widget between comment and share
* Add support for sharing posts to squads from companion
* Bonus: Submit commenatry on CMD+Enter like we have for comments

Everything was tested locally.

Sidenote: Some sites block our images from loading but it's a known issue as far as I recall. It's relevant for squad images and user images

image
image

### DIFF
--- a/packages/extension/src/background/index.ts
+++ b/packages/extension/src/background/index.ts
@@ -47,7 +47,7 @@ const sendBootData = async (_, tab: Tabs.Tab) => {
 
   const [
     deviceId,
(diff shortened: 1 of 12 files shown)


sshanzel (approved):
I am just thinking of a way to make creating requests in the companion easier in the future but this looks pretty good to me! 🚢 🚢 🚢

rebelchris (approved):
Looks good for now
We should probably add a task to uniform request as one object.

PR #1672, opened 2023-03-20
fix: changelog browser polyfill in production extension bundle

capJavert:
- Fixes issue where extension update button would always result in error
- Browser polyfill was not loaded in production extension bundle (while it is loaded during development and exposed in `globalThis`)
- While I could not find core reason for this (looking at webpack configs and build script), I think it is due to this being the only place where polyfill is required inside the shared package
- To fix I added polyfill as shared dependency and lazy load it when extension update is requested
- I also checked other places and all of them are from extension package and already use polyfill so this bug is not present

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done

### DIFF
--- a/packages/shared/src/components/tooltips/ChangelogTooltip.tsx
+++ b/packages/shared/src/components/tooltips/ChangelogTooltip.tsx
@@ -52,32 +52,39 @@ function ChangelogTooltip<TRef extends HTMLElement>({
   }, [sidebarExpanded]);
 
   const { mutateAsync: updateExtension, isLoading: isExtensionUpdating } =
-    useMutation(async () => {
-      if (isFirefoxExtension) {
-        return;
-      }
+    useMutation(
+      async () => {
+        if (isFirefoxExtension) {
+          return;
+        }
 
-      if (!isExtension) {
-        toast.displayToast(toastMessageMap.error);
+        if (!isExtension) {
+          throw new Error(
+            'updateExtension can only be used inside the extension runtime',
+          );
+        }
 
-        return;
-      }
+        const browser = await import('webextension-polyfill-ts').then(
+          (mod) => mod.browser,
+        );
 
-      const browser = await import('webextension-polyfill-ts').then(
-        (mod) => mod.browser,
-      );
+        const updateResponse: { status: string } =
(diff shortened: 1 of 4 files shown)


packages/shared/src/components/tooltips/ChangelogTooltip.tsx
```diff
@@ -62,17 +63,14 @@ function ChangelogTooltip<TRef extends HTMLElement>({
         return;
       }
 
-      const sendMessage = globalThis?.browser?.runtime?.sendMessage;
```
sshanzel, line 65: Great find on this. We never knew beforehand that the polyfill should be prioritized **all the time** on the extension package.

sshanzel (approved):
Love the fact that you remembered to use dynamic import. Great find and thanks for checking through the codebase. Amazing work all throughout! 🚢

packages/shared/src/components/tooltips/ChangelogTooltip.tsx
```diff
@@ -62,17 +63,14 @@ function ChangelogTooltip<TRef extends HTMLElement>({
...
+      const browser = await import('webextension-polyfill-ts').then(
+        (mod) => mod.browser,
+      );
 
-      if (typeof sendMessage !== 'function') {
-        toast.displayToast(toastMessageMap.error);
-
-        return;
-      }
```
rebelchris, line 71: We no longer need to support this in case of the import?
What happens if it's can't polyfill somewhere?

rebelchris (approved):
Great find, just one generic question.

packages/shared/src/components/tooltips/ChangelogTooltip.tsx
rebelchris: Yeah makes sense to catch the "Generic" error for now.
Message can be the one we used already for when it would fail.

capJavert: Added in 0b6c2aa4be4925f1ee57f770b6987a5972662ed2

PR #1674, opened 2023-03-20
feat(squads): add invite item to member list

capJavert:
- Add invite action to squad member list

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1158 #done


packages/shared/src/components/profile/UserList.tsx
```diff
@@ -14,15 +20,31 @@ export interface UserListProps {
...
+    const token = squad?.currentMember?.referralToken;
+
+    if (!permalink || !token) {
+      return undefined;
+    }
+
+    const invitation = `${permalink}/${token}`;
+
+    return invitation;
+  });
```
rebelchris, line 46: I think we should maybe refactor this to the top level and simply pass down the function to relevant components.

OR - create a hook for this (useCopySquadLink) since it will be a uniform hook that's used by multiple components.

Let me know what you think.

packages/shared/src/components/profile/UserList.tsx
```diff
@@ -31,6 +53,29 @@ function UserList({
         <UserShortInfoPlaceholder placeholderAmount={placeholderAmount} />
       }
     >
+      {!!squad && (
```
rebelchris, line 56: Thinking about it, maybe we can simply add a "initialItem" as it's a very generic component and we now narrow it down with a squad specific feature.

rebelchris (changes requested):
Some minor flow suggestions

packages/shared/src/components/profile/UserList.tsx
capJavert: Also thought about this. It can be done for sure.

packages/shared/src/components/profile/UserList.tsx
capJavert: Thought about this. It can be done for sure.

packages/shared/src/hooks/useCopyLink.ts
```diff
@@ -9,6 +9,7 @@ type CopyNotifyFunctionProps = NotifyOptionalProps & {
 };
 
 const defaultMessage = '✅ Copied link to clipboard';
+const noLinkErrorMessage = '❌ Could not copy, link is missing';
```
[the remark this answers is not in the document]
capJavert: There is a bunch of string concatenations around the project for the links eg. `${post?.commentsPermalink}${getCommentHash(commentId)}` where if something is `undefined` the link would end up `undefined/bCrZekKfq` or `undefined/undefined`. 

I think the correct way to handle this is to return `undefined` from the `getLink` method like I did in this PR. And then this is why I added additional handling so it can be caught in development and production.

packages/shared/src/components/modals/SquadMemberModal.tsx
```diff
@@ -91,6 +101,29 @@ export function SquadMemberModal({
...
+                  squad: squad.id,
+                }),
+              });
+
+              copySquadInvite();
+            }}
+          >
+            <div className="flex justify-center items-center mr-4 w-12 h-12 bg-theme-float rounded-10">
+              <LinkIcon size={IconSize.Large} />
+            </div>
```
sshanzel, line 123:
```suggestion
            <div className="flex justify-center items-center mr-4 w-12 h-12 bg-theme-float rounded-10">
              <LinkIcon className="" size={IconSize.Large} />
            </div>
```

sshanzel, line 123: I think we have a utility classed component like below.
```suggestion
            <FlexCentered className="mr-4 w-12 h-12 bg-theme-float rounded-10">
              <LinkIcon size={IconSize.Large} />
            </FlexCentered>
```

packages/shared/src/components/profile/UserList.tsx
sshanzel: I have created a hook in a different PR for copying. It will be able to utilize what you have here if we would want to keep the copying and the invitation separate.

sshanzel (approved):
Great work! Just had some non-blocking comments since you have already covered Chris' comments. Thank you!

packages/shared/src/components/modals/SquadMemberModal.tsx
```diff
@@ -91,6 +101,29 @@ export function SquadMemberModal({
             </SimpleTooltip>
           );
         }}
+        initialItem={
```
rebelchris, line 104: non-breaking suggestion:
I would suggest defining the item outside as a const and only passing that, think it will keep the overall component better structured.

packages/shared/src/components/modals/SquadMemberModal.tsx
```diff
@@ -91,6 +101,29 @@ export function SquadMemberModal({
...
+            disabled={copying}
+            className="flex justify-start items-center py-3 px-6 hover:bg-theme-hover"
+            onClick={() => {
+              trackEvent({
+                event_name: AnalyticsEvent.ShareSquadInvitation,
+                extra: JSON.stringify({
+                  origin: Origin.SquadMembersList,
+                  squad: squad.id,
+                }),
+              });
```
rebelchris, line 116: Ideally this should happen in the one hook accepting origin and squad id as props.
WDYT? (Might already be in @sshanzel hook?

rebelchris (approved):
Looks good, just see my suggestion and note on Web team.
We need to align the hook to be uniform across all actions 🙏

packages/shared/src/components/modals/SquadMemberModal.tsx
capJavert: @rebelchris I was wondering on which would be prefferd. I am fine with top level but maybe moving it to separate component is also good since many components in the same file also look messy. I guess it comes down to once it needs to be resed in multiple things we then move it outside.

rebelchris: I think it will be something we need to document as it's "vague" at best now.

Normally how I look at it:
- small component can be inline
- large components (30+ lines?) can be own components
- reusable small components can be own components.

So think for now inside this components function is fine, just outside of the render for now.
Buit love to hear your thoughts on this as well.

capJavert: Yeah, let's hard document this. it would be better to linter it up, but option though.

PR #1675, opened 2023-03-20
feat(squads): new squad button tooltip in sidebar

capJavert:
- Add sidebar tooltip to new squads button
- Add generic `useTutorial` hook

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1159 #done


packages/shared/src/components/sidebar/Sidebar.tsx
```diff
@@ -133,22 +139,38 @@ export default function Sidebar({
...
+                <SimpleTooltip
+                  content={
+                    <>
+                      You can always
+                      <br /> open a squad here
+                    </>
                   }
-                  onClick={() => openSquadBetaModal({ origin: Origin.Sidebar })}
+                  placement="right"
+                  offset={newSquadTooltipOffset}
```
rebelchris, line 151: The offset should not really be needed for this, it's not custom right?
Think the placement alone should work.

rebelchris (changes requested):

packages/shared/src/components/sidebar/Sidebar.tsx
capJavert: It is different by the design.

capJavert:
@rebelchris if you can take a look at `useTutorial`, I can add tests after that. 

Some additional things I though about:
- `state` is generic because for some tutorials we might want to have different states and `boolean` would not be sufficent
- added `reset`, maybe in some cases we would need to reshow tutorial to the user
- `complete` and `reset` are async for future proofing to if tutorials are tracked server side in the future
- thought about saving everything to storage but in the end decided not to because only case where that would be applicable is if user reloads the page and I don't think we support that for other things as well (like modals)

packages/shared/src/components/sidebar/Sidebar.tsx
```diff
@@ -133,22 +143,42 @@ export default function Sidebar({
...
+                  sticky
+                  plugins={[sticky]}
+                  visible={newSquadTooltipTutorial.isActive}
```
rebelchris, line 158: This could be used to render it as well right?

packages/shared/src/hooks/useTutorial.ts
```diff
@@ -0,0 +1,72 @@
...
+  key,
+  defaultState = false,
+}: UseTutorialProps): UseTutorial => {
+  const client = useQueryClient();
+  const [isCompleted, setCompleted, isFetched] = usePersistentContext(
+    `tutorial-${key}`,
+    false,
+  );
+
+  const queryKey = ['tutorial', 'query', key];
```
rebelchris, line 33: Why do we mix two keys though?
Can't we leverage the persistent context to keep track of the `isActive` and `setActive` as well?

rebelchris (changes requested):
I love the approach and think this is a great flow!
Some minor concerns around duplicate keys for 1 function.

Would also love to hear what @sshanzel thinks of this one 👍

packages/shared/src/components/sidebar/Sidebar.tsx
capJavert: @rebelchris you mean `newSquadTooltipTutorial.isActive && <SimpleTooltip />`?

packages/shared/src/hooks/useTutorial.ts
rebelchris: Hmm good question can you raise it in the `#feat-squads-closedbetav3` channel for Sab?

packages/shared/src/components/sidebar/Sidebar.tsx
rebelchris: Yeah exactly, would that not be better then using a seperate state?

sshanzel (commented):
Overall, I love the abstraction of the state for the current tutorial. Well thought out.

Though would it be not better of we would also manage the state of the sequence? Or that will be part of the next steps here? Assuming as this is still a draft.

Just had some minor things, but, this is well on the right track 🚂

capJavert:
> Overall, I love the abstraction of the state for the current tutorial. Well thought out.
> 
> Though would it be not better of we would also manage the state of the sequence? Or that will be part of the next steps here? Assuming as this is still a draft.
> 
> Just had some minor things, but, this is well on the right track 🚂

@sshanzel some explanation of usage [here](https://dailydotdev.slack.com/archives/C04UED2504R/p1679490131404769)

I don't think currently we need to track state of the sequence since tutorials are atomic and if you need two different on the same screen you would just use two hook instances. I did have `isActive` as a generic before in a2f884ccf321e0ec418fe4546c10866ca506c855 but decided against it.

If you have specific case where you think this is needed point it out, it would be helpful, I looked at other examples from squad v3 design and I think all of those should be supported by current functionality 🤞

packages/shared/src/hooks/useTutorial.ts
capJavert: Going without persisting active state for now, [thread here](https://dailydotdev.slack.com/archives/C04UED2504R/p1679490131404769)

packages/shared/src/components/modals/NewSquadModal.tsx
```diff
@@ -113,11 +114,18 @@ function NewSquadModal({
...
+  });
+
   const handleClose = async () => {
     if (activeView === ModalState.Ready) return onRequestClose();
 
     const shouldQuit = await showPrompt(quitSquadModal);
-    if (shouldQuit) onRequestClose();
+    if (shouldQuit) {
+      onRequestClose();
+      newSquadTutorial.complete(!newSquadTutorial.isCompleted);
```
rebelchris, line 127: Should we not simply always complete it?
Not sure when it should not complete.

packages/shared/src/hooks/useTutorial.ts
rebelchris: Ok, but would it not be simpler from a engineering side of view?
Now you have 2 queries basically doing the exact same thing already.

But maybe I'm missing something here.
Happy to chat about it.

rebelchris: I mean can't we simply have default = false, and if it's true it means it's active?
Then turn it to false once it's completed again?

Not sure we need the extra state with react query if that makes sense.

packages/shared/src/components/sidebar/Sidebar.tsx
capJavert: It should actually be show for 4 seconds, so additional logic is required. Will update.

packages/shared/src/hooks/useTutorial.ts
capJavert: I did think about it, how would you differentiate between activated and completed then?

Let's take a look at carousel example. So, we need to add animation around post field when we are on one of the carousel slides. If we go with two states we have (I am gonna show the state in each step as `{ isActive, isCompleted }`:
- users opens the carousel for the first time, state is `{ isActive: false, isCompleted: false }`
- user opens the target carousel slide
  - we activate the slide because tutorial is not yet completed `{ isActive: true, isCompleted: false }` 
- user closes the carousel
  - we finish the tutorial `{ isActive: false, isCompleted: true }`
- user opens the carousel again `{ isActive: false, isCompleted: true }`
- user opens to target carousel slide
  - we know that tutorial has been completed so we do not activate again  `{ isActive: false, isCompleted: true }`
- user closes the carousel
  - we know that tutorial has been completed so we don't need to complete again  `{ isActive: false, isCompleted: true }`
  
If we would to have single state eg. `{ isActive }`
- users opens the carousel for the first time, state is `{ isActive: false }`
- user opens the target carousel slide
  - we activate the slide because tutorial is not yet completed `{ isActive: true }` 
- user closes the carousel
  - we finish the tutorial `{ isActive: false }`
- user opens the carousel again `{ isActive: false }`
- user opens to target carousel slide
  - we just now carousel is not active so we activate it again  `{ isActive: true }`
- user closes the carousel
  -  we complete it again  `{ isActive: false }`

No matter how many times we run it we would always show the tutorial in this case.
  
So basically I think 2 states are needed just because we have an arbitrary completion conditions. If complete conditions was something that we could express in another state eg. we don't show animation anymore if user created at least one post inside squad that would eliminate the need for two states. Does this make sense, if I am missing something I would be glad to simplify this 🙏

sshanzel:
> If you have specific case where you think this is needed point it out, it would be helpful 🙏 , I looked at other examples from squad v3 design and I think all of those should be supported by current functionality 🤞

Atomicity would always be the best option whenever possible, and if the current setup supports all the requirements we have, then we can go with the current implementation.

Tracking of steps should probably be done only when necessary as it will complicate the process (as I initially thought was needed). Great job! 🚢

packages/shared/src/components/sidebar/Sidebar.tsx
```diff
@@ -72,6 +81,27 @@ export default function Sidebar({
...
+    }
+
+    const timeout = setTimeout(() => {
+      newSquadTooltipTutorial.complete();
+    }, 4 * 1000);
+
+    return () => {
+      clearTimeout(timeout);
+    };
+  }, [newSquadTooltipTutorial.isActive]);
```
rebelchris, line 103: We have some hooks I think that deal with timers already would suggest we stick to those.

Either `useTimedAnimation` or `useTimer` if they don't fit let's see how we can abstract into a better hook maybe?

rebelchris (changes requested):
Just one more comment for a hook use-case otherwise happy.

packages/shared/src/hooks/useTutorial.spec.tsx
```diff
@@ -0,0 +1,117 @@
...
+
+  it('should return the tutorial state', async () => {
+    const client = new QueryClient();
+    const { result } = renderHook(
+      () => useTutorial({ key: TutorialKey.SEEN_NEW_SQUAD_TOOLTIP_KEY }),
+      { wrapper: createWrapper({ client }) },
+    );
+
+    expect(result.current.isActive).toBe(false);
+    await waitForCompletedValue({ client, result, toBe: false });
```
capJavert, line 50: Had to use this because `usePersistentContext` and I think the `idb-keyval` lib under the hood had some weird promise handling and they would leak after the tests have completed. `await new Promise(process.nextTick)` did not work either.

rebelchris (approved):
Let's go.

PR #1676, opened 2023-03-21
fix: recreated the new squad flow

rebelchris:
- Simplify the create new squad flow
- Remove beta modal
- Still using the hook as it can be easy to adjust later on.

## Events

Removed:

  ClickSquadCreationNext = 'click squad creation next',
  ClickSquadCreationBack = 'click squad creation back',

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1152 #done


packages/shared/src/components/modals/NewSquadModal.tsx
```diff
@@ -121,39 +64,25 @@ function NewSquadModal({
...
+        src={cloudinary.squads.createSquad}
+        alt="Create your Squad"
+        className="h-60"
+      />
+      <SquadDetails
+        form={form}
+        onRequestClose={handleClose}
+        onSubmit={onCreate}
+        createMode
+      />
```
sshanzel, line 85: OMG. I love the changes here. It is so much simple now.

packages/shared/src/components/squads/Details.tsx
```diff
@@ -43,11 +46,13 @@ export function SquadDetails({
   onSubmit,
   form,
   createMode = true,
+  onRequestClose,
 }: SquadDetailsProps): ReactElement {
   const { name, handle, description } = form;
+  const [useHandle, setUseHandle] = useState(handle);
```
sshanzel, line 52: I think it would be best to reserve the verb/prefix `use` when working with custom hooks. WDYT?
```suggestion
  const [shouldUseHandle, setShouldUseHandle] = useState(handle);
```

packages/shared/src/components/squads/Details.tsx
```diff
@@ -96,29 +107,35 @@ export function SquadDetails({
...
+              id={squadImageId}
+              fallbackImage={cloudinary.squads.imageFallback}
+              className={{
+                container: '!rounded-full border-0 my-1',
+                img: 'object-cover',
+              }}
+              hoverIcon={<CameraIcon size={IconSize.Large} />}
+              alwayShowHover={!createMode && !imageChanged}
+              onChange={() => setImageChanged(true)}
+              size={createMode ? 'medium' : 'large'}
```
sshanzel, line 130: I think we can now omit the condition.
```suggestion
              size="large"
```

packages/shared/src/components/squads/Details.tsx
```diff
@@ -96,29 +107,35 @@ export function SquadDetails({
...
+            <ImageInput
+              initialValue={form.image ?? form.file}
+              id={squadImageId}
+              fallbackImage={cloudinary.squads.imageFallback}
+              className={{
+                container: '!rounded-full border-0 my-1',
+                img: 'object-cover',
+              }}
+              hoverIcon={<CameraIcon size={IconSize.Large} />}
+              alwayShowHover={!createMode && !imageChanged}
```
sshanzel, line 128:
```suggestion
              alwayShowHover={!imageChanged}
```

packages/shared/src/graphql/squads.ts
```diff
@@ -51,10 +51,7 @@ type EditSquadInput = SharedSquadInput & {
   sourceId: string;
 };
 
-type CreateSquadInput = SharedSquadInput & {
-  postId: string;
-  commentary: string;
-};
+type CreateSquadInput = SharedSquadInput;
```
sshanzel, line 55: Just a question. Is it not possible to create a Squad without a post? Like for when they clicked the Share button and then selected "Create Squad"?

sshanzel (changes requested):
Wow that was a lot of work and testing! Just had some minor comments, also wanted to point out that there could be some challenge when merging due to the v2 being introduced. Let me know if you need any help on that side 🚢 🚢 🚢

packages/shared/src/components/squads/Details.tsx
rebelchris: It's not really a should evaluation.
It accepts a incoming handle and then one it actually uses.

Wi'll think about another rename :)

packages/shared/src/graphql/squads.ts
rebelchris: Think we remove this in favor of simpler squad creation, but will check with product :)

packages/shared/src/components/squads/Details.tsx
sshanzel: Oh, cool. Though it was boolean a boolean from the name. Agreed, let's rename.

rebelchris: @sshanzel Went with `activeHandle` does that make more sense?

sshanzel: That sounds amazing and intuitive. Great suggestion! 🚢

packages/shared/src/graphql/squads.ts
sshanzel: Oh, cool. Thank you for checking with him! 🚢

sshanzel (approved):
Great work on this! Getting closer and closer by the minute! 🚢

PR #1689, opened 2023-03-24
fix: add tutorial for notification prompt

rebelchris:
- Prompt notifications only once tour is closed.
- @capJavert will cause some merge conflict with your one :)

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1175 #done

### DIFF
--- a/packages/shared/src/components/squads/SquadPageHeader.tsx
+++ b/packages/shared/src/components/squads/SquadPageHeader.tsx
@@ -26,8 +26,8 @@ export function SquadPageHeader({
   onNewSquadPost,
 }: SquadPageHeaderProps): ReactElement {
   const { sidebarRendered } = useSidebarRendered();
-  const squadEnableNotifications = useTutorial({
-    key: TutorialKey.SQUAD_ENABLE_NOTIFICATIONS_KEY,
+  const enableNotifications = useTutorial({
+    key: TutorialKey.SQUAD_ENABLE_NOTIFICATIONS,
   });
 
   const sharePostTutorial = useTutorial({
@@ -75,7 +75,7 @@ export function SquadPageHeader({
         onNewSquadPost={onNewSquadPost}
         className={sidebarRendered && 'absolute top-0 right-[4.5rem]'}
       />
-      {squadEnableNotifications.isActive && (
+      {enableNotifications.isActive && (
         <EnableNotification
           contentName={squad.name}
           source={NotificationPromptSource.SquadPage}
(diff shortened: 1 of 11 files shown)


capJavert (approved):
@rebelchris just one thing, I see you do not call `complete` for `SQUAD_ENABLE_NOTIFICATIONS`, It works fine because this notification box has it's own storage key `DISMISS_PERMISSION_BANNER`, but maybe to be consistent we can call `complete` also.

packages/shared/src/hooks/useTutorial.ts
```diff
@@ -3,7 +3,8 @@ import { useQuery, useQueryClient } from 'react-query';
 import usePersistentContext from './usePersistentContext';
 
 export enum TutorialKey {
-  SEEN_NEW_SQUAD_TOOLTIP_KEY = 'seenNewSquadTooltip',
+  SEEN_NEW_SQUAD_TOOLTIP = 'seenNewSquadTooltip',
+  SQUAD_ENABLE_NOTIFICATIONS = 'squadEnableNotifications',
   SHARE_SQUAD_POST = 'shareSquadPost',
   COPY_SQUAD_LINK = 'copySquadLink',
```
sshanzel, line 9: I believe for enums' properties, we use PascalCase.

```suggestion
  SeenNewSquadTooltip = 'seenNewSquadTooltip',
  SquadEnableNotifications = 'squadEnableNotifications',
  ShareSquadPost = 'shareSquadPost',
  CopySquadLink = 'copySquadLink',
```

packages/shared/src/components/squads/SquadPageHeader.tsx
```diff
@@ -74,7 +75,7 @@ export function SquadPageHeader({
         onNewSquadPost={onNewSquadPost}
         className={sidebarRendered && 'absolute top-0 right-[4.5rem]'}
       />
-      {hasTriedOnboarding && (
+      {enableNotifications.isActive && (
```
sshanzel, line 78: This would mean the user would only see the notification banner after closing the popup. Which is correct on the first time, but what about on the second time? Should we now keep it persistent to be displayed since they already saw the SquadTour?

sshanzel (approved):
Just had one concern with the key convention for enums.

Also, my question around being persistent for the notifications banner is more of a product-related one. I think we should it with Sab.

packages/shared/src/hooks/useTutorial.ts
rebelchris: We mix it now, but agree let's stick with PascalCase as it's most common actually.

rebelchris:
@sshanzel The hasTriedOnboarding trigger when you start onboarding, we only want to show it once you close it.
It's not the same update.

Before we added a new state, but this handles it for us now.

rebelchris:
@sshanzel You where right! Adjusted the format :D

sshanzel (approved):
As clean as it can get! 🚢

PR #1688, opened 2023-03-23
feat: squad page tutorial guides

capJavert:
### Describe what this PR does
- Add tutorial guides for posting and inviting members (also implements WT-1157) # done
- Add TutorialGuide component (it also has arrow logic for all 4 positions)

## Events

Did you introduce any new tracking events?
Don't forget to update the [Analytics Taxonomy sheet](https://docs.google.com/spreadsheets/d/18Lv7zXges9QfVX5VYL1a-Hyl0e1sQ3sLr0OK8YZWKXI/edit#gid=0)

Log the new/changed events below:

| Type   | event_name  | value |
|--------|-------------|-------|
| Change/New | event name  | extra: { ... } |

### On those affected packages:
- [ ] Have you done sanity checks in the webapp?
- [ ] Have you done sanity checks in the extension?
- [ ] Does this not break anything in companion?

### Did you test the modified components media queries?
- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

#### Did you test on actual mobile devices?
- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1156 #done

PR #1690, opened 2023-03-24
feat: highlight animation pulse

sshanzel:
- Just use the `highlight-pulse` class for it to work.
- At first, I tried to use the `transform: scale(n)` but ironically, it was not scalable as shown below.

https://user-images.githubusercontent.com/13744167/227490645-abc3a055-dc8f-4647-b88b-36528dc0df6c.mov

It works on smaller ones, but on the SharePostBar, it doesn't look right.

- But using width and height, it does the job on all kinds.

https://user-images.githubusercontent.com/13744167/227490817-54a290f2-c5ad-42d0-964a-98737c70321f.mov

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1154 #done


rebelchris (changes requested):
Looks great, can you possible also add the call from the relevant tour screens in this PR so we can test it out?

See for instance this PR how it works:
https://github.com/dailydotdev/apps/pull/1688

sshanzel:
> Looks great, can you possible also add the call from the relevant tour screens in this PR so we can test it out?

I have instead based this branch on that PR where the tutorial was implemented so I can just tag the areas where it is active.

To activate:
- Pull this branch.
- Open the Squads tour.
- Once closed you should see the highlight on the share post bar.

packages/shared/src/components/squads/SquadHeaderBar.tsx
```diff
@@ -54,7 +54,10 @@ export function SquadHeaderBar({
         )}
       >
         <Button
-          className="btn-primary"
+          className={classNames(
+            'btn-primary',
+            copyLinkTutorial.isActive && 'highlight-pulse',
```
rebelchris, line 59: This is the wrong prompt.
These pulse animations are seperate from the other CTA.

Basically the pulse should show DURING the tour, and the other ones AFTER the tour.
Let me know if you want me to show you quick?

rebelchris (changes requested):

packages/shared/src/components/squads/SquadHeaderBar.tsx
sshanzel: Oh, cool, when they are at that stage. Let me update it.

sshanzel:
@rebelchris let me know what you think! 🚢

packages/shared/src/components/squads/SquadPageHeader.tsx
```diff
@@ -84,6 +90,7 @@ export function SquadPageHeader({
       <div
         className={classNames(
           'absolute bottom-0 w-full translate-y-1/2 px-6 laptop:px-0 bg-theme-bg-primary',
+          tourIndex === TourScreenIndex.Post && 'highlight-pulse',
```
rebelchris, line 93: Was it not easier to use the useTutorial for this?
It would need less drilldown right

rebelchris (changes requested):
More generic question on this setup

rebelchris (changes requested):
Is there anyway we can convert this to a hook like `useSquadTour` that will keep the state instead of passing it down?
It's very prone to error as it has to many passthroughs i'm afraid.

packages/shared/src/components/modals/SquadTourModal.tsx
```diff
@@ -1,16 +1,17 @@
...
-  onScreenIndexChange,
   ...props
-}: ModalProps & CarouselMinimalProps): ReactElement {
+}: ModalProps): ReactElement {
+  const { onTourIndexChange } = useSquadTour();
   const onModalClose: typeof onRequestClose = (param) => {
-    onScreenIndexChange?.(-1);
+    onTourIndexChange(-1);
     onRequestClose(param);
   };
```
rebelchris: @sshanzel Note I made a new hook for this in favor of individual implementations 👯

sshanzel: Whoa. That would make it much cleaner! I will wait for that PR to get merged before proceeding here.

rebelchris (approved):
Love this uniform hook now, well done Lee 🥳

PR #1693, opened 2023-03-25
fix: multiple fixes to "share post" page

idoshamun:

* Not sure why we had this snippet there but there's no reason to redirect the user to the permalink of the post when they land on the article page. We should either show 404 or show the article page.
* Clicking on the referred post section in the "share post" page should lead to another post page unless it's from the unknown source
* We shouldn't show the image and handle of the unknown source

### DIFF
--- a/packages/shared/src/components/post/PostSourceInfo.tsx
+++ b/packages/shared/src/components/post/PostSourceInfo.tsx
@@ -18,6 +18,7 @@ function PostSourceInfo({
   size = 'medium',
   className,
 }: SourceInfoProps): ReactElement {
+  const isUnknown = source.id === 'unknown';
   return (
     <span
       className={classNames(
@@ -25,11 +26,15 @@ function PostSourceInfo({
         className,
       )}
     >
-      <SourceButton source={source} size={size} />
-      <h3 className="ml-3">{source.handle}</h3>
+      {!isUnknown && (
+        <>
+          <SourceButton source={source} size={size} />
+          <h3 className="ml-3">{source.handle}</h3>
+        </>
+      )}
       {!!date && (
         <>
-          <Separator />
+          {!isUnknown && <Separator />}
           <time dateTime={date}>{date}</time>
         </>
       )}
--- a/packages/webapp/pages/posts/[id].tsx
+++ b/packages/webapp/pages/posts/[id].tsx
@@ -4,7 +4,6 @@ import React, {
   ReactNode,
   useCallback,
   useContext,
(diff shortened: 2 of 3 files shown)


packages/webapp/pages/posts/[id].tsx
```diff
@@ -87,12 +86,6 @@ const PostPage = ({ id, initialData }: Props): ReactElement => {
     options: { initialData, retry: false },
   });
 
-  useEffect(() => {
-    if (!post?.private || !post?.permalink) return;
-
-    globalThis?.location.replace(post.permalink);
-  }, [post]);
-
```
Comment (sshanzel) [inline packages/webapp/pages/posts/[id].tsx:95]: If we point to the permalink when clicking the post content summary box, not sure why we would not redirect from here.

https://github.com/dailydotdev/apps/blob/f148977b74d46f972f7e0abd43f3102db72e8aff/packages/shared/src/components/post/SquadPostContent.tsx#L148-L153

Comment (idoshamun) [reply packages/webapp/pages/posts/[id].tsx:95]: Oh now I understand the notion of this snippet, but if this is the case than the original condition is completely not aligned with external link case. Anyway, I suggest for simplicity to remove the redirect. There's no way for a user to land on an external link post page unless they manually build the url

Comment (sshanzel) [reply packages/webapp/pages/posts/[id].tsx:95]: I agree, landing here is highly unlikely. The condition for the "safety-net" just came out wrong on the implementation. Thanks for the fix!

sshanzel (approved):
Thanks for the fix! 🚢 🚢 🚢

packages/shared/src/components/post/PostSourceInfo.tsx
```diff
@@ -18,18 +18,23 @@ function PostSourceInfo({
...
   return (
     <span
       className={classNames(
         'flex flex-row items-center text-theme-label-tertiary typo-footnote',
         className,
       )}
     >
-      <SourceButton source={source} size={size} />
-      <h3 className="ml-3">{source.handle}</h3>
+      {!isUnknown && (
```
rebelchris, line 29: This was a decision on slack though to simply show unknown

packages/webapp/pages/posts/[id].tsx
Comment (rebelchris) [reply packages/webapp/pages/posts/[id].tsx:95]: The concern was that links could get shared which is now possible and it should rather show 404 but decisions where made to redirect on slack before.

rebelchris (commented):
Ok with me just leaving comments as to why certain things where made.

We need clearer alignment in the future i feel like

PR #1694, opened 2023-03-27
fix: activated textfield has input

sshanzel:
- From the snippet below, we already pass the `hasInput` if it is available, but it seems to be coming from a state where it is not getting updated in some instances. So now when it detects a change that is not coming from an `input` change event, we would be able to make the state up-to date.
https://github.com/dailydotdev/apps/blob/main/packages/shared/src/components/fields/TextField.tsx#L144-L174

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1180 #done

### DIFF
--- a/packages/shared/src/hooks/useInputField.ts
+++ b/packages/shared/src/hooks/useInputField.ts
@@ -35,6 +35,7 @@ export function useInputField<T extends ValidInputElement = HTMLInputElement>(
   useEffect(() => {
     if (value !== undefined) {
       setInput(value?.toString() || '');
+      setHasInput(!!inputRef?.current?.value?.length);
     }
   }, [value]);
 


rebelchris (approved):
Ah nice find on this :)
Think it was only the programmatic error so this should fix that.

sshanzel:
> Think it was only the programmatic error so this should fix that.

I agree, I also tested it locally. Seems to look okay now.

sshanzel:
> Think it was only the programmatic error so this should fix that.

I agree, I also tested it locally. Seems to look okay now.

Preview:


https://user-images.githubusercontent.com/13744167/227942900-130a4763-5afc-4642-9f0e-fb12694b74f4.mov

PR #1698, opened 2023-03-28
fix: companion popups

sshanzel:
- Usage of `Portal` must allow to which place in the DOM it must be appended.
- LazyModalElement was not included in the Companion's rendered components.
- After long hours of battling with the LazyLoading, the issue is, the resource fetch is being pointed to the site we are currently at, instead of pointing to the resources of the `chrome-extension://id`. Haven't found a way to manually point the calls to that.
- Refactored the LazyModalElement to accept a list of modals to use to have more options.
- Fixed the upvoted modal usage on Companion where the initial state is the first node being null.
- I have also verified the tooltips/popups in the Companion and all seem to be covered now.

@capJavert just some early background info while I am still on my way to writing the doc page for the Companion, in order for our styles to work, we must append the popup/tooltip to the base container of our shadow root. We have a function to fetch the said element, it is named `getCompanionWrapper` as utilized in this PR.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1177 #done

### DIFF
--- a/packages/extension/src/companion/App.tsx
+++ b/packages/extension/src/companion/App.tsx
@@ -1,4 +1,4 @@
-import React, { ComponentType, ReactElement, useMemo, useState } from 'react';
+import React, { ReactElement, useMemo, useState } from 'react';
 import { QueryClient, QueryClientProvider } from 'react-query';
 import { browser } from 'webextension-polyfill-ts';
 import { Boot, BootApp } from '@dailydotdev/shared/src/lib/boot';
--- a/packages/extension/src/companion/App.tsx
+++ b/packages/extension/src/companion/App.tsx
@@ -1,4 +1,4 @@
-import React, { ReactElement, useMemo, useState } from 'react';
+import React, { ComponentType, ReactElement, useMemo, useState } from 'react';
 import { QueryClient, QueryClientProvider } from 'react-query';
 import { browser } from 'webextension-polyfill-ts';
 import { Boot, BootApp } from '@dailydotdev/shared/src/lib/boot';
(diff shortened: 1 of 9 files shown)


packages/extension/src/companion/App.tsx
```diff
@@ -41,6 +44,10 @@ export type CompanionData = { url: string; deviceId: string } & Pick<
 
 const refreshTokenKey = 'refresh_token';
 
+const companionModals = {
+  [LazyModal.UpvotedPopup]: UpvotedPopupModal,
+};
```
rebelchris, line 49: I don't think we should start doing this, basically companion needs to support all modals down the line.

packages/extension/src/companion/CompanionContent.tsx
```diff
@@ -19,6 +19,7 @@ import AnalyticsContext from '@dailydotdev/shared/src/contexts/AnalyticsContext'
 import { CompanionEngagements } from './CompanionEngagements';
 import { CompanionDiscussion } from './CompanionDiscussion';
 import { useBackgroundPaginatedRequest } from './useBackgroundPaginatedRequest';
+import { getCompanionWrapper } from './common';
```
rebelchris, line 22: Have you considered this context we chatted about?
What where the implications, would love to hear about this.

packages/shared/src/components/modals/LazyModalElement.tsx
```diff
@@ -1,12 +1,18 @@
-import React, { ReactElement } from 'react';
+import React, { ComponentType, ReactElement } from 'react';
```
rebelchris, line 1: This file should not introduce the switch between apps, it will become a big pain down the line.

rebelchris (changes requested):
Have some bigger concerns by changing the lazy modals, don't think it's needed even.

packages/extension/src/companion/CompanionContent.tsx
rebelchris: Yes, but we raised it here: https://dailydotdev.slack.com/archives/C02E2C3C13R/p1679929905819559?thread_ts=1679929715.768539&cid=C02E2C3C13R

sshanzel: After debugging the changes, there is not much difference between the usages. Both use cases must still live under a React component.
Current:
```
```
With Context:
```
  const { requestMethod } = useContext(RequestContext);
```

With the changes above there seem to be no impact on performance too and when an where we can use it.

rebelchris: Let's maybe take some time today to chat about the implications and what we are trying to solve with this maybe?

capJavert (changes requested):
Just some issue with memoization.

sshanzel:
Way too outdated. Closing now. Something to note though: the ultimate goal is to make lazy modal (dynamic import) work on companion, the current flow is when you load, the system is thinking you are trying to pull resources from the host page rather than the extension's asset and resources. If we will be able to put a prefix on the request for the asset/files, we should point it to our extension's resources.

PR #1701, opened 2023-03-28
fix: reading history grouping

sshanzel:
## Changes
- When the user's timezone is above over the current country/timezone they are at, it groups the day for tomorrow even though it happened technically "Today".
- So to avoid confusion, since it happened "Today", we will list the item as "Today" instead of "Tomorrow".
- Though it was discussed that we can use this mechanism to identify whether the user must change their timezone and we must show a popup, but that is still in progress under the product team.

Before:
image

After:
image

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-369 #done

### DIFF
--- a/packages/shared/src/components/history/ReadingHistoryList.tsx
+++ b/packages/shared/src/components/history/ReadingHistoryList.tsx
@@ -54,6 +54,8 @@ export default function ReadHistoryList({
 
   const renderList = useCallback(() => {
     let currentDate: Date;
+    let previousGroup: string;
+    const now = new Date();
 
     return data?.pages.map((page, pageIndex) =>
       page.readHistory.edges.reduce((dom, { node: history }, edgeIndex) => {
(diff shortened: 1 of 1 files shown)


packages/shared/src/components/history/ReadingHistoryList.tsx
```diff
@@ -62,7 +64,12 @@ export default function ReadHistoryList({
 
         if (!currentDate || !isDateOnlyEqual(currentDate, date)) {
           currentDate = date;
-          dom.push(getDateGroup(date));
+          const dateToGroup = currentDate > now ? now : date;
```
capJavert, line 67: This will technically only happen for the first group "Today" since all other will fall under their regular groups?

Maybe just a naming wise we could use something like `mismatchedTimezoneGroup`.

Also maybe add this handling to separate if statement, I know it will add some code repetition but I think it will be more verbose in the long run. Like this it takes a minute to see what is actually going on.

capJavert (commented):
Non blocking, but something to think about since changing the functionality for specific case.

packages/shared/src/components/history/ReadingHistoryList.tsx
sshanzel: I think that's a great idea, I would rather have a separate scope to manage this specific use case than merge them together as you suggested. Will work that out. Thanks for pointing it out! 🚢

PR #1703, opened 2023-03-28
hotfix: missed typo

rebelchris:
### Describe what this PR does
- Small typo we missed in copy-paste action

## Events

Did you introduce any new tracking events?
Don't forget to update the [Analytics Taxonomy sheet](https://docs.google.com/spreadsheets/d/18Lv7zXges9QfVX5VYL1a-Hyl0e1sQ3sLr0OK8YZWKXI/edit#gid=0)

Log the new/changed events below:

| Type   | event_name  | value |
|--------|-------------|-------|
| Change/New | event name  | extra: { ... } |

### On those affected packages:
- [ ] Have you done sanity checks in the webapp?
- [ ] Have you done sanity checks in the extension?
- [ ] Does this not break anything in companion?

### Did you test the modified components media queries?
- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

#### Did you test on actual mobile devices?
- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done

PR #1704, opened 2023-03-28
fix: processing toast only show if private link submitted

capJavert:
- As mentioned here https://github.com/dailydotdev/apps/pull/1703#discussion_r1150726244
- this should only show toast for processing if link is being shared

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done

### DIFF
--- a/packages/shared/src/components/modals/PostToSquadModal.tsx
+++ b/packages/shared/src/components/modals/PostToSquadModal.tsx
@@ -56,11 +56,11 @@ function PostToSquadModal({
   const onPostSuccess = async (squadPost?: Post) => {
     if (squadPost) onSharedSuccessfully?.(squadPost);
 
-    if (isLink) {
-      displayToast(
-        'This post is being processed and will be shared with your Squad shortly',
-      );
-    }
+    displayToast(
+      isLink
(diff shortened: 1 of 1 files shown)


packages/shared/src/components/modals/PostToSquadModal.tsx
```diff
@@ -56,9 +56,12 @@ function PostToSquadModal({
...
     if (squadPost) onSharedSuccessfully?.(squadPost);
 
-    displayToast(
-      'This post is being processed and will be shared with your Squad shortly',
-    );
+    if (isLink) {
+      displayToast(
+        'This post is being processed and will be shared with your Squad shortly',
+      );
+    }
```
rebelchris, line 63: Ah missed that variable when checking 😂
Let's do it something like this instead:

```suggestion
      displayToast(isLink ? 'This post is being processed and will be shared with your Squad shortly' : 'This post has been shared to your squad');
```

rebelchris (changes requested):
Just a small addition

rebelchris (approved):
Thanks 🥳

PR #1705, opened 2023-03-28
feat: add permission options to create squad modal

capJavert:
- add permission options to create squad modal
- this one only does the FE work it can be connected once API is done through `WT-1197`
- Minor adjustments to modal design/copy
- Note: currently showing permissions form fields only in `createMode`, in another PR I will continue implementation for squad settings

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1198 #done

### DIFF
--- a/packages/shared/src/components/squads/SquadMemberMenu.tsx
+++ b/packages/shared/src/components/squads/SquadMemberMenu.tsx
@@ -67,7 +67,7 @@ const getUpdateRoleOptions = (
     action: getUpdateRoleFn(SourceMemberRole.Owner, MenuItemTitle.AddAsOwner),
   };
   const promoteToModerator = {
-    text: MenuItemTitle.DemoteToModerator,
+    text: MenuItemTitle.PromoteToModerator,
     icon: <UserIcon size={IconSize.Small} />,
     action: getUpdateRoleFn(
       SourceMemberRole.Moderator,


sshanzel:
We will have to update the receiving mutation to accept this parameter, we will have to make changes to the database structure too.

packages/shared/src/components/squads/Details.tsx
```diff
@@ -170,6 +171,28 @@ export function SquadDetails({
...
+            <h4 className="mb-2 font-bold typo-headline text-theme-label-primary">
+              Post permissions
+            </h4>
+            <p className="mb-6 text-salt-90 typo-callout">
+              Choose who is allowed to post new content in this Squad.
+            </p>
+            <RadioItem
+              name="permissionLevel"
+              value="all-members"
+              checked={false}
```
rebelchris, line 184: This is not needed right?

packages/shared/src/components/squads/Details.tsx
```diff
@@ -170,6 +171,28 @@ export function SquadDetails({
               }}
             />
           )}
+          <div className="flex flex-col justify-start">
+            <h4 className="mb-2 font-bold typo-headline text-theme-label-primary">
+              Post permissions
+            </h4>
+            <p className="mb-6 text-salt-90 typo-callout">
```
rebelchris, line 178: Ah we shouldn't use color versions directly, did linting not complain about it?
And margin-bottom shows as 16 to the box so 4.

```suggestion
            <p className="mb-4 text-theme-label-tertiary typo-callout">
```

packages/shared/src/components/squads/Details.tsx
```diff
@@ -170,6 +171,28 @@ export function SquadDetails({
...
+              checked={false}
+            >
+              All members
+            </RadioItem>
+            <RadioItem
+              name="permissionLevel"
+              value="only-moderators"
+              checked={false}
+            >
+              Only Moderators
```
rebelchris, line 193: I think this is a design issue we don't use secondary capitalization unless it's a "brand" like Squad.

packages/shared/src/components/squads/Details.tsx
```diff
@@ -170,6 +171,28 @@ export function SquadDetails({
...
+            <RadioItem
+              name="permissionLevel"
+              value="all-members"
+              checked={false}
+            >
+              All members
+            </RadioItem>
+            <RadioItem
+              name="permissionLevel"
+              value="only-moderators"
```
rebelchris, line 190: I think @sshanzel suggested making this a boolean for instance as `restrictMemberPosting`.
So with `onlyModerators` it could be "true" and for all member "false"

Think it makes sense, let me know what you think.

rebelchris (changes requested):
Nice, have some minor suggestions.

packages/shared/src/components/squads/Details.tsx
capJavert: Nope, I did see it sometimes complains. Will keep this in mind but still little hard to follow this colors since they are not directly referenced in design.

sshanzel (commented):
Nothing more I can add! Amazing work here! 🚢 🚢 🚢

capJavert:
Just general comment, I think we really need a form library/standardisation, `RadioItem` with boolean shenanigans and also whole mess with `form` and separate states in `Details` are really error prone.

packages/shared/src/components/squads/Details.tsx
```diff
@@ -170,6 +182,30 @@ export function SquadDetails({
...
           )}
+          {createMode && (
+            <div className="flex flex-col justify-start">
+              <h4 className="mb-2 font-bold typo-headline text-theme-label-primary">
+                Post permissions
+              </h4>
+              <p className="mb-4 text-theme-label-tertiary typo-callout">
+                Choose who is allowed to post new content in this Squad.
+              </p>
+              <RadioItem
```
rebelchris, line 193: Probably favor using `<Radio>` component, this wraps the items and you can simply pass an array of options to it.

packages/shared/src/graphql/sources.ts
```diff
@@ -46,6 +46,7 @@ export interface Squad extends Source {
   members?: Connection<SourceMember>;
   membersCount: number;
   description: string;
+  memberPostingRank: number;
```
rebelchris, line 49: Should this not be `memberPostingRole`?

rebelchris (changes requested):
Two minor comments, but way to go

packages/shared/src/graphql/sources.ts
rebelchris: Yeah probably we can do that to uniform the edit function, but let's see once we tackle that one.

rebelchris (approved):
Nice one!

capJavert:
Merging so I can continue with `WT-1196`

PR #1706, opened 2023-03-28
feat: add indication for roles in squad

capJavert:
- Add indication for roles in squad
  - comment and sub comments
  - post view/content

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1201 #done


packages/shared/src/components/comments/MainComment.spec.tsx
```diff
@@ -20,6 +21,7 @@ const renderLayout = (
   user: LoggedUser = null,
 ): RenderResult => {
   const defaultProps: MainCommentProps = {
+    post,
```
capJavert, line 24: Noticed tests were failing because `post` was missing. `post` prop is required by `MainCommentProps` so I think this was just a test error.

packages/shared/src/graphql/posts.ts
```diff
@@ -153,6 +153,12 @@ export const POST_BY_ID_QUERY = gql`
           permissions
           role
         }
+        privilegedMembers {
+          user {
+            id
+          }
+          role
+        }
```
capJavert, line 161: Did not want to include all fields on purpose to avoid big array payloads.

capJavert: Also we mentioned to potentially only query this field if `source.type` is `squad`, any examples of queries that do something similar?

packages/shared/src/components/icons/Start/index.tsx
```diff
@@ -0,0 +1,10 @@
+import React, { ReactElement } from 'react';
+import Icon, { IconProps } from '../../Icon';
+import OutlinedIcon from './outlined.svg';
+import FilledIcon from './filled.svg';
+
+const StarIcon = (props: IconProps): ReactElement => (
```
rebelchris, line 6: The main file is called `Start` with a T think it's a typo.

rebelchris (changes requested):
Minor typo

packages/shared/src/graphql/posts.ts
idoshamun: Don't think it's possible

rebelchris (approved):
Nice! 🙌

packages/shared/src/components/squads/SquadMemberRoleBadge.tsx
```diff
@@ -0,0 +1,48 @@
...
+  return (
+    <span
+      className={classNames(
+        'flex items-center ml-2 text-cabbage-40 typo-footnote font-bold capitalize',
+        className,
+      )}
+    >
+      <RoleIcon secondary className="mx-0.5" />
+      {role}
+    </span>
```
sshanzel, line 44: I think we should instead create a simple component that will accept an icon and its label. I see this kind of component used when the user is the Author or Scout. Maybe it is time to create a single component to be used for badges like this.

Then we will simplify the `CommentAuthor` component to just receive an array of badges we will then map to return the newly introduced component. Technically the comment doesn't really need to know about/get the `source` object.

sshanzel (changes requested):
Great work on integrating the BE work here!

Just has some comments regarding how we can reuse a component for existing implementations.

Also the Star/Filled is not accurate, I think you downloaded the one for Marketing.

packages/shared/src/components/squads/SquadMemberRoleBadge.tsx
capJavert: I see where you are coming from, for sure we can generalize badges view component. 

Still I think we would need components like `SquadMemberRoleBadge` as business logic wrappers to contain logic for extracting role since it would be clunky to have that logic spread into all of the components where badges are used. 

I'll look into this today.

sshanzel: Totally agree with that. I always love seeing hooks getting introduced. We can probably use that to wrap the logic outside this component as you mentioned. Way to go! 🚢

packages/shared/src/components/squads/SquadMemberRoleBadge.tsx
```diff
@@ -0,0 +1,39 @@
...
+const SquadMemberRoleBadge = ({
+  author,
+  source,
+}: SquadMemberRoleBadgeProps): ReactElement => {
+  const { role } = useMemberRoleForSource({ source, user: author });
+  const RoleIcon = RoleToIconMap[role];
+
+  if (!RoleIcon) {
+    return null;
+  }
```
capJavert, line 28: Still left this mapping logic to component level, I don't think its a good pattern to import and returns icons/components from hooks. View related stuff is best left in components.

packages/shared/src/components/UserBadge.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+const UserBadge = ({
+  className,
+  content,
+  icon,
+}: UserBadgeProps): ReactElement => {
+  return (
+    <span
+      className={classNames(
+        'flex items-center ml-2 typo-footnote font-bold capitalize',
```
sshanzel, line 19: I think there is always a gap between the icon and the label itself, not sure of the exact spacing, but I think it can be a good idea to support it OOTB.

```suggestion
        'flex gap-2 items-center ml-2 typo-footnote font-bold capitalize',
```

packages/shared/src/components/icons/Star/outlined.svg
sshanzel, line 1: I think this icon is for marketing, kindly download the one for the app.

packages/shared/src/components/squads/SquadMemberRoleBadge.tsx
sshanzel: Totally agree with that. No point in importing something you can't render. Basically in a hook. Well done.

packages/shared/src/hooks/useMemberRoleForSource.ts
```diff
@@ -0,0 +1,25 @@
...
+export type UseMemberRoleForSourceProps = {
+  source: Source;
+  user?: Pick<PublicProfile, 'id'>;
+};
+
+export type UseMemberRoleForSourceResult = {
+  role?: SourceMemberRole;
+};
+
+export const useMemberRoleForSource = ({
```
sshanzel, line 14: Oh, I love this.

sshanzel (approved):
It looks so much cleaner and simpler. Thank you for the extra effort on making the components and the additional hooks.

Just had some non-blocking comments, but everything looks amazing!!! 💪 💪 💪

packages/shared/src/components/icons/Star/outlined.svg
capJavert: @sshanzel from what I can see this is the one linked in sketch, could you please reference the one you are talking about?

packages/shared/src/components/squads/SquadMemberRoleBadge.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+export type SquadMemberRoleBadgeProps = {
+  source?: Source;
+  author?: Author;
+};
+
+const RoleToIconMap = {
+  [SourceMemberRole.Moderator]: UserIcon,
+  [SourceMemberRole.Owner]: StarIcon,
+};
```
rebelchris, line 17: Would suggest adding a record type here, just for safety down the line.

packages/shared/src/components/comments/CommentBox.tsx
```diff
@@ -76,11 +79,32 @@ function CommentBox({
...
+                    content="Author"
+                    icon={<FeatherIcon secondary className="mx-0.5" />}
+                  />
+                ),
+                comment.author.id === postScoutId && (
+                  <UserBadge
+                    key="scout"
+                    className="text-theme-color-bun"
+                    content="Scout"
+                    icon={<ScoutIcon className="mr-0.5" />}
```
rebelchris, line 103: Why do these icons have different classes should they not be similar? (seeing they conditional render)

packages/shared/src/components/comments/CommentAuthor.tsx
```diff
@@ -2,21 +2,21 @@ import React, { ReactElement } from 'react';
...
-import FeatherIcon from '../icons/Feather';
 import { ProfileTooltip } from '../profile/ProfileTooltip';
+import { UserBadgeProps } from '../UserBadge';
 
 export interface CommentAuthorProps {
-  postAuthorId: string | null;
   author: Author;
   className?: string;
   appendTooltipTo?: () => HTMLElement;
+  badges?: ReactElement<UserBadgeProps>[];
```
rebelchris, line 12: The squadMemberRoleBadge is not a UserBadgeProps right?
Not sure if that's correct now

rebelchris (changes requested):
Minor type concerns, otherwise happy.

packages/shared/src/components/comments/CommentAuthor.tsx
capJavert: I guess our typescript does not work 😆

packages/shared/src/components/icons/Star/outlined.svg
sshanzel: The ones under the `Marketing` folder were used here. Probably was a mistake when they used which icon to apply. Though kindly confirm with the design team just so they will be aware of it.

Link to the assets: https://www.sketch.com/s/aca0d743-919a-4f68-9411-3c155f9f85a0/symbols?g=Icon&search=Star

image

capJavert:
Small adjustments, moving from `icon` to `Icon` as component to allow better customisation.

rebelchris (approved):
Looks great

packages/shared/src/components/UserBadge.tsx
```diff
@@ -0,0 +1,37 @@
+import classNames from 'classnames';
+import React, { FunctionComponent, ReactElement, ReactNode } from 'react';
+import { IconProps, IconSize } from './Icon';
+import useMedia from '../hooks/useMedia';
+import { tablet } from '../styles/media';
+
+export type UserBadgeProps = {
+  className?: string;
+  content: ReactNode;
+  Icon: FunctionComponent<IconProps>;
```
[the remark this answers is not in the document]
capJavert: I think it is cleaner then to put bunch of props on react element. With component we can also inject custom props more easily eg. in SquadMemberRoleBadge without cloning. It would be great for all of us to sit down one time and discuss different pros/cons since they are minor but it would be cool to standardize. 👌

packages/shared/src/components/comments/CommentBox.tsx
```diff
@@ -76,11 +79,38 @@ function CommentBox({
...
+                    key="scout"
+                    className="text-theme-color-bun"
+                    content="Scout"
+                    Icon={ScoutIcon}
+                    iconProps={{
+                      secondary: true,
+                    }}
+                  />
+                ),
+              ].filter(Boolean)}
```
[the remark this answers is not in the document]
capJavert: Because in the first iteration I was using functions here and rendering them inside CommentBox this was just to avoid having false values and need to do type checking inside CommentBox. For this `false` is just fine as it does not render by React. good catch 👌

PR #1707, opened 2023-03-29
refactor: source and squad entity to share current member

sshanzel:
- Reorganized the placement of the types to avoid a dependency cycle and update the names of the types to the more accurate ones.
- This is in preparation for what will be available and on the types that is needed soonest.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done


packages/shared/src/graphql/sources.ts
```diff
@@ -1,4 +1,18 @@
 import { gql } from 'graphql-request';
+import { UserShortProfile } from '../lib/user';
+import { Connection } from './common';
+
+export enum SourceMemberRole {
+  Member = 'member',
+  Owner = 'owner',
```
rebelchris, line 7: Maybe already introduce moderator here?

packages/shared/src/graphql/sources.ts
```diff
@@ -13,6 +27,12 @@ export interface Source {
   handle: string;
   type: SourceType;
   permalink: string;
+  membersCount: number;
+  description?: string;
+  members?: Connection<SourceMember>;
+  currentMember?: SourceMember;
+  owners?: string[];
+  moderators?: string[];
```
rebelchris, line 35: Not sure if this is right now?
A non squad source would never have owners and moderators so why would we add it there?

A squad is technically a "extended source"

sshanzel: Oh, good point. Though technically at some point it might happen (to manage the source by some employee from the source), but yes, it is not in our roadmap. Let me move these two.

rebelchris: Yes that makes sense it could take a Squad or Source possibly?
And in case of a type of source being squad then run those extra fields?

sshanzel: I just thought we can probably keep it `Source | Squad` and just typecast to Squad whenever necessary. Like only when checking some properties.

rebelchris: Yes exactly, same thoughts 🙏

packages/shared/src/graphql/sources.ts
capJavert: Agreed, I already added it in one of my PRs.

packages/shared/src/graphql/sources.ts
capJavert: Maybe in the long term unify through different queries so that proper types are returned. But `Source | Squad` is ok since most of squad properties are optional and we need to have checks even now.

sshanzel:
> Maybe ask Ido if he can help with the deps cycle, he seemed to solve it last time.

We will have to move a lot of types/files to certain folders basically. I don't want to block this as much as possible since those changes can be done in isolation and some PRs will be depending on the intial changes here.

packages/shared/src/graphql/sources.ts
```diff
@@ -13,6 +28,10 @@ export interface Source {
   handle: string;
   type: SourceType;
   permalink: string;
+  membersCount: number;
+  description?: string;
+  members?: Connection<SourceMember>;
+  currentMember?: SourceMember;
```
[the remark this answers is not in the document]
sshanzel: The build will fail on token page. The issue is, some parts were using the source member's source property, then its description. We can move all the types from `/squads` file to `/sources` file altogether if want this to pass and remove those properties back to Squad type. WDYT?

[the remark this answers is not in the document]
sshanzel: Wait, let me try to mix some things up.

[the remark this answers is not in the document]
capJavert: I don't think it is possible to fix it since literally the one file is referencing the other. Bundler will probably be able to resolve it but linter just forbids it. 

I think it is fine to refactor to move used types to common or something.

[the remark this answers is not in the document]
rebelchris: Yep probably move them to a common is actually a neater solution down the line.
But yes let's keep types where they belong and not make shortcuts because lint/type issues for now.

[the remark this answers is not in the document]
sshanzel: I have simply moved the Squad to the Source file since it basically extends Source. Build should be fixed now.

rebelchris (approved):
Nice work, besides lint/test issues that might show up good to go from my side.
Thanks Lee 🙏

PR #1709, opened 2023-03-30
fix: blocked squad member

rebelchris:
- Introduce blocked role
- Refactor evaluation on token page

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1215 #done

### DIFF
--- a/packages/webapp/__tests__/SquadTokenPage.tsx
+++ b/packages/webapp/__tests__/SquadTokenPage.tsx
@@ -256,7 +256,7 @@ describe('invalid token', () => {
     const owner = generateTestOwner();
     const member = generateTestMember(1);
     member.user.id = defaultUser.id;
-    owner.source.members.edges.push({ node: member });
+    owner.source.currentMember = member;
     renderComponent([createInvitationMock(defaultToken, owner)]);
     await waitForNock();
     expect(replaced).toEqual(`/squads/${owner.source.handle}`);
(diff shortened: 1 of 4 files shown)


packages/webapp/pages/squads/[handle]/[token].tsx
```diff
@@ -133,6 +141,11 @@ const SquadReferral = ({
   );
 
   const onJoinClick = async () => {
+    if (isBlockedUser) {
```
Comment (sshanzel) [inline packages/webapp/pages/squads/[handle]/[token].tsx:144]: We already have the current member's role, I don't think we need an additional state.
```suggestion
    if (member.source.currentMember.role === SquadMemberRole.Blocked) {
```

packages/webapp/__tests__/SquadTokenPage.tsx
```diff
@@ -256,7 +256,7 @@ describe('invalid token', () => {
     const owner = generateTestOwner();
     const member = generateTestMember(1);
     member.user.id = defaultUser.id;
-    owner.source.members.edges.push({ node: member });
+    owner.source.currentMember = member;
```
sshanzel, line 259: There should now be a case if the user is blocked we should not redirect. This test is now inaccurate since the member used here is not blocked and did not redirect.

sshanzel (changes requested):
Looking great so far! Just some cleanups and more tests.

sshanzel (approved):
Nice work! Thank you for fixing an issue along the way! 💯

PR #1712, opened 2023-03-31
feat: update roles

sshanzel:
- Allow owners to promote/demote/remove members.
- Allow moderators to remove members.
- Introduced a hook where we will include all actions we need for Squad.
- Prompt size to allow `Size.Small`.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1189 #done
WT-1193 #done


packages/shared/src/components/squads/SquadMemberMenu.tsx
```diff
@@ -1,51 +1,158 @@
...
+  RemoveMember = 'Remove member',
+}
+
+const promptDescription: Record<
+  MenuItemTitle,
+  (memberName: string, squadName: string) => string
+> = {
+  [MenuItemTitle.AddAsOwner]: (memberName, squadName) =>
+    `${memberName} will get the same permissions as you have. Adding as co-owner will not replace you as the owner of ${squadName}`,
+  [MenuItemTitle.PromoteToModerator]: () => '', // TODO: awaiting product's response
```
rebelchris, line 41: Blocks approval for now

packages/shared/src/components/squads/SquadMemberMenu.tsx
```diff
@@ -1,51 +1,158 @@
...
+  [MenuItemTitle.DemoteToModerator]: (memberName) =>
+    `${memberName} will no longer have owner permissions. You can always reverse this decision later`,
+  [MenuItemTitle.DemoteToMember]: () => '', // TODO: awaiting product's response
```
rebelchris, line 44: Blocks approval for now

packages/shared/src/components/squads/SquadMemberMenu.tsx
```diff
@@ -1,51 +1,158 @@
...
+  [MenuItemTitle.RemoveMember]: (memberName, squadName) =>
+    `${memberName} will be a Blocked members and will no longer have access to ${squadName}. They will not be able to rejoin unless you unblock them.`,
+};
 
 export default function SquadMemberMenu({
```
rebelchris, line 49: I find this file a bit overwhelming at the moment.
Is there anyway we can clean it out or re-use more parts?

Not sure if it's just me, but just looks a bit messy code-wise.

packages/shared/src/graphql/squads.ts
```diff
@@ -194,6 +203,17 @@ export interface SquadEdgesData {
   sourceMembers: Connection<SourceMember>;
 }
 
+interface UpdateSquadMemberRoleArgs {
```
rebelchris, line 206: Why do we opt Args here, think most codebase uses Props right?
Just want to align the codebase where possible.

packages/shared/src/hooks/squads/useSquadActions.ts
```diff
@@ -0,0 +1,17 @@
...
+import { useMutation, useQueryClient } from 'react-query';
+import { updateSquadMemberRole } from '../../graphql/squads';
+
+interface UseSquadActions {
+  onUpdateRole?: typeof updateSquadMemberRole;
+}
+
+export const useSquadActions = (): UseSquadActions => {
+  const client = useQueryClient();
+  const { mutateAsync: onUpdateRole } = useMutation(updateSquadMemberRole, {
```
rebelchris, line 11: Should it not just be `updateRole` as it actually executes and not just "After" handles it?

rebelchris (changes requested):
Also note you are pointing to main, think this should point to intermediate right?

packages/shared/src/components/modals/SquadMemberModal.tsx
```diff
@@ -103,21 +113,23 @@ export function SquadMemberModal({
...
+          if (loggedUser.id === user.id) return null;
+
           return (
             <SimpleTooltip content="Member options">
               <Button
                 buttonSize={ButtonSize.Small}
                 className="m-auto mr-0 btn-tertiary"
                 iconOnly
-                onClick={(e) => onReportClick(e, user.id)}
+                onClick={(e) => onReportClick(e, members[index])}
```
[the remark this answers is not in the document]
sshanzel: User doesn't have other necessary properties such as role, etc.

packages/shared/src/components/squads/SquadMemberMenu.tsx
sshanzel: Not sure which parts actually. Can you specify which ones you think requires refactoring?

rebelchris: Just in general there is a lot of arrays being structured and think we can clean them up a bit?
Maybe we can have a look together if you want.

packages/shared/src/components/modals/SquadMemberModal.tsx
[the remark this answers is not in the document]
rebelchris: But is user not exactly the members looped index?
Not sure I understand what the difference between looped element and index array is then

packages/shared/src/components/squads/SquadMemberMenu.tsx
sshanzel: I mean, those arrays are necessary as those are the options. Sure, let's meet after the web team meeting.

packages/shared/src/components/modals/SquadMemberModal.tsx
```diff
@@ -72,11 +74,21 @@ export function SquadMemberModal({
...
   );
 
-  const onReportClick = (e, userId) => {
-    setMemberId(userId);
+  const onReportClick = (e: React.MouseEvent, clickedMember: SourceMember) => {
+    setMember(clickedMember);
     onMenuClick(e);
   };
 
+  const members = useMemo(
```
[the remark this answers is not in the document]
sshanzel: That makes sense! Let's keep it all inside that hook.

packages/shared/src/graphql/squads.ts
sshanzel: Too much BE work 😅 
But since we are in apps repo, I agree, let's keep it to props.

packages/shared/src/hooks/squads/useSquadActions.ts
sshanzel: Our exposed functions have a prefix `on`, just keeping it in line with the code base. I think what you mean is about present tense vs past tense for which we use present tense as it is appropriate as you explained.

packages/shared/src/components/modals/SquadMemberModal.tsx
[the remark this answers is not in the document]
sshanzel: > But is user not exactly the members looped index?

No, it is the user only.

> Not sure I understand what the difference between looped element and index array is then

The abstracting component requires `user` entity, which we have a `source member` entity, we mapped it to be just a `user`, so to get the member, we use the index. Hope that clears it.

packages/shared/src/hooks/squads/useSquadActions.ts
```diff
@@ -0,0 +1,17 @@
...
+    onSuccess: (_, { sourceId }) =>
+      client.invalidateQueries(['squadMembers', sourceId]),
+  });
+
+  return useMemo(() => ({ onUpdateRole }), []);
```
capJavert, line 16: `onUpdateRole` missing in dependency array could lead to stale function references.

packages/shared/src/components/squads/SquadMemberMenu.tsx
```diff
@@ -1,51 +1,158 @@
...
+    const demoteToMember = {
+      text: MenuItemTitle.DemoteToMember,
+      icon: <SquadIcon size={IconSize.Small} />,
+      action: getUpdateRoleFn(
+        SourceMemberRole.Member,
+        MenuItemTitle.DemoteToMember,
+      ),
+    };
+
+    const canUpdateRole = squad.currentMember?.permissions.includes(
```
capJavert, line 111: Maybe we can extract `permissions.includes` into utility function, I see even now being called on few places, it will make more sense if we add additional logic also. We have similar utility functions on the API side.

capJavert (changes requested):
Solid so far 🤘, just a few notes in addition to what Chris said.

packages/shared/src/components/squads/SquadMemberMenu.tsx
sshanzel: Do you mean the arrays? I think if we re-create the one we have in API, it might lead to discrepancies and more maintenance work. Though I agree, checking in FE is a bit verbose now, we need to come up with something to lessen it. I'll try to create some functions.

packages/shared/src/hooks/squads/useSquadActions.ts
sshanzel: Right. But that function doesn't really depend on anything since it just accepts a parameter and executes functions. 

Having it on the dependency list and not being a memoized callback would just re-render it every time and make the use memo not fruitful. Or I could be wrong. WDYT?

packages/shared/src/components/squads/SquadMemberMenu.tsx
capJavert: Yeah, array iteration for now, API was just an example to where we do have those kind of utility functions. 👍

packages/shared/src/hooks/squads/useSquadActions.ts
capJavert: Yeah but like this you get the false sense of stable function which could then be stale due to inner workings of a hook. In that case it is better to skip useMemo all together. 

Also I do think that mutation reference might be stable in `useMutation`, but you would need to check that. If it is not then I would just avoid useMemo in this case.

sshanzel: >  In that case it is better to skip useMemo all together.

Totally agree with this. It has just become a thing that our hooks are memoized. Probably we can talk more about this in our next web team meeting to reconsider when we should use it.

sshanzel:
> Also note you are pointing to main, think this should point to intermediate right?

Good point. Was thinking we can merge this once the "Blocked" list becomes available, but we can point this to the intermediate branch as you suggested 🚢

sshanzel:
@rebelchris @capJavert would love to hear what you both think! 🚢

packages/shared/src/components/squads/SquadMemberMenu.tsx
```diff
@@ -1,51 +1,163 @@
...
   const { user } = useContext(AuthContext);
+  const { showPrompt } = usePrompt();
+  const onUpdateMember = async (
+    role: SourceMemberRole,
+    title: MenuItemTitle,
+  ) => {
+    const hasConfirmed = await showPrompt({
+      title: `${title}?`,
+      description: promptDescription[title](member.user.name, squad.name),
+      okButton: { title, className: 'btn-primary-ketchup' },
```
rebelchris, line 67: This should be the purple one right?

packages/shared/src/components/squads/SquadMemberMenu.tsx
```diff
@@ -1,51 +1,163 @@
...
+      content: <UserShortInfo user={member.user} />,
```
rebelchris, line 68: The remove role currently doesn't show this, is there any way we can dynamic map this as well?

rebelchris (changes requested):
Some minor comments, but great work altogether.

packages/shared/src/components/squads/SquadMemberMenu.tsx
sshanzel: Oh, right. Totally missed that. Thank you! 🔭

packages/shared/src/components/squads/SquadMemberMenu.tsx
sshanzel: Good catch! Coming right up!

sshanzel: After checking with the product/design team, they now want to keep the preview all the time.

Quick link to discussion: https://dailydotdev.slack.com/archives/C04UQUBDWKD/p1680526000089179?thread_ts=1680522693.874399&cid=C04UQUBDWKD

packages/shared/src/components/squads/SquadMemberMenu.tsx
sshanzel: I found a way to at least take out the arrays to be coming from a function for easier reading. Hope it helps.

packages/shared/src/hooks/squads/useSquadActions.ts
```diff
@@ -0,0 +1,78 @@
...
+
+  return useMemo(
+    () => ({
+      onUpdateRole,
+      membersQueryResult,
+      members,
+      verifyPermission: checkHasPermission,
+    }),
+    [onUpdateRole, membersQueryResult, squad, membersQueryEnabled],
+  );
```
capJavert, line 77: Still some minor inconsistencies for variables inside dependency array:

Missing from array:
- members
- checkHasPermission

Can be removed from array:
- membersQueryEnabled
- squad

capJavert (changes requested):
Much more readable 👌 single inconsistency on React side

packages/shared/src/hooks/squads/useSquadActions.ts
sshanzel: > membersQueryEnabled

Oh, right. This can be removed. Thanks!

> members

The `members` have a dependency from `membersQueryResult` which is already included here, when that changes, both `members` and the function in this `useMemo` will get executed to return the updated value.

> checkHasPermission

This is not a stable dependency since this is a regular function unless we use `useCallback` which I think is unnecessary, that's why we need the `squad` as a dependency, so when the `squad` changes, this `useMemo` will fire the function and then get the newly re-created `checkHasPermission` function.

packages/shared/src/hooks/squads/useSquadActions.ts
```diff
@@ -0,0 +1,78 @@
...
+      membersQueryResult.data?.pages
+        .map((page) =>
+          page.sourceMembers.edges.map(({ node }) => ({ ...node, page })),
+        )
+        .flat() ?? [],
+    [membersQueryResult],
+  );
+
+  const checkHasPermission = (permission: SourcePermissions) =>
+    verifyPermission(squad, permission);
```
rebelchris, line 67: Just wondering why we don't do the fn right here, it's only 1 line so not sure if the import is worth it.
Minor -  non blocking

rebelchris (approved):
Looks great 👍

packages/shared/src/hooks/squads/useSquadActions.ts
capJavert: > The members have a dependency from membersQueryResult which is already included here, when that dependency changes, both members and the function in this useMemo will get executed to return the updated value.

Hmm there is no point in not including `members` here then, since it will act the same as you said. 😃 

Imagine down the line members is no longer connected to membersQueryResult, what then? We would get bugs where members has stale values due to it not being included in deps array. If you want we can also talk in gather about this if you find it faster to close the discussion.

sshanzel: Good point, that makes devex better, but the concern is, `members` will always be based on the members query result.

Probably, what we can do is remove the variable declaration completely since this is not used anywhere inside this hook and just make it in-line with the return of `useMemo`. That way we can reduce the places we check. WDYT?

capJavert: @sshanzel yep, this is the way 👌

PR #1715, opened 2023-04-03
feat: blocked squad members list

sshanzel:
- Blocked members list
- Update the UserListModal to accept a header than just a title.
- Add the new parameter for fetching members by role
- Updated the modal tabs component to accept on click callback.
- Introduce usage of `roleRank`.

Preview:

image

image

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1188 #done


packages/shared/src/components/squads/SquadMemberActions.tsx
```diff
@@ -0,0 +1,68 @@
...
+  const isLoggedUser = loggedUser.id === user.id;
+
+  if (role !== SourceMemberRole.Member) {
+    return (
+      <>
+        <span
+          className="flex gap-1 items-center mr-2 font-bold typo-footnote text-theme-color-cabbage"
+          data-testvalue={user.username}
+        >
+          <SquadIcon secondary /> {capitalize(role)}
```
sshanzel, line 56: Subject to change once the indication PR is merged.

packages/shared/src/components/modals/SquadMemberModal.tsx
```diff
@@ -49,21 +49,23 @@ const InitialItem = ({ squad }: { squad: Squad }) => {
...
   const { onMenuClick } = useContextMenu({ id: 'squad-member-menu-context' });
   const {
     members,
     membersQueryResult: queryResult,
+    onUnblock,
     onUpdateRole,
     verifyPermission,
   } = useSquadActions({
     squad,
+    membersQueryParams: { role: roleFilter },
```
rebelchris, line 63: One side of me likes this, other side feels like it's a bit hacky around the modal guideline.
Not sure if we should change or not, maybe @capJavert has an opinion here?

packages/shared/src/components/squads/SquadMemberActions.tsx
```diff
@@ -0,0 +1,68 @@
...
+  member,
+  onUnblockClick,
+  onOptionsClick,
+}: SquadMemberActionsProps): ReactElement {
+  const { user: loggedUser } = useAuthContext();
+  const { role, user } = member;
+
+  if (role === SourceMemberRole.Blocked) {
+    return (
+      <Button
```
rebelchris, line 27: Has a tooltip in sketch

rebelchris (changes requested):
Just noting it's blocked until the other one is merged I guess so we can complete that code section.

packages/shared/src/components/modals/SquadMemberModal.tsx
sshanzel: I agree with your sentiment. One part makes it quick and working, on the other side, we are not really "switching" the UI per tab as the guideline would suggest. I am willing to refactor if we decide to do two separate components for the list.

packages/shared/src/components/modals/SquadMemberModal.tsx
```diff
@@ -73,39 +75,35 @@ export function SquadMemberModal({
       <UserListModal
         {...props}
         title="Squad members"
+        tabs={Object.values(SquadMemberTab)}
+        header={
+          squad?.currentMember?.roleRank > 0 ? (
```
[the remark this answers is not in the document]
sshanzel: Yes, we do, we keep the `title` property on line 77 🥳

packages/shared/src/components/squads/SquadMemberActions.tsx
sshanzel: Great catch! Thank you!

packages/shared/src/components/squads/SquadMemberActions.tsx
```diff
@@ -0,0 +1,88 @@
...
+  onOptionsClick,
+}: SquadMemberActionsProps): ReactElement {
+  const { showPrompt } = usePrompt();
+  const { user: loggedUser } = useAuthContext();
+  const { role, user } = member;
+
+  const onConfirmUnblock = async (e: React.MouseEvent) => {
+    e.preventDefault();
+
+    const hasUnblocked = await showPrompt({
```
rebelchris, line 31: Minor thing, but think now we mainly do something like this:
```
if (await showPrompt(promptOptions)) {
```

rebelchris (changes requested):
Amazing work, have some minor suggestions

packages/shared/src/components/modals/SquadMemberModal.tsx
```diff
@@ -71,41 +74,49 @@ export function SquadMemberModal({
...
+            <SquadMemberActions
+              member={members[index]}
+              onUnblock={() =>
+                onUnblock({ sourceId: squad.id, memberId: user.id })
+              }
+              onOptionsClick={(e) => onOptionsClick(e, members[index])}
+            />
+          ),
+          emptyPlaceholder: (
+            <FlexCentered className="flex-col pb-10 h-full">
```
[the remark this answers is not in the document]
sshanzel: It is centered but it looks a bit off and a bit better with the additional spacing. WDYT?

Without bottom padding:

image

packages/shared/src/components/squads/SquadMemberActions.tsx
```diff
@@ -0,0 +1,88 @@
...
+      title: 'Unblock member?',
+      description: `${member.user.name} will now have access to join your Squad and can then post, upvote and comment`,
```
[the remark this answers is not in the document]
sshanzel: This should be correct. We are referring to the selected user in the member's blocked list not the logged in user.

packages/shared/src/components/squads/SquadMemberRoleBadge.tsx
```diff
@@ -41,12 +44,22 @@ const SquadMemberRoleBadge = ({
 
   return (
     <UserBadge
-      className={classNames('text-cabbage-40', className)}
+      className={classNames('text-theme-color-cabbage', className)}
```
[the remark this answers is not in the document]
sshanzel: I was also shocked this got past through the lint. Probably happens when using at `classNames` function?

packages/shared/src/components/squads/SquadMemberRoleBadge.tsx
```diff
@@ -22,13 +27,11 @@ const RoleToIconMap: Partial<
   [SourceMemberRole.Owner]: StarIcon,
 };
 
-const SquadMemberRoleBadge = ({
+export const SquadMemberBadge = ({
```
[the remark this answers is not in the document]
rebelchris: Hmm is there any way we can uniform them into 1 component that can handle both cases?
EG > one can detect role early?

packages/shared/src/components/squads/SquadMemberActions.tsx
sshanzel: Ah, just my personal preference, I kinda don't like seeing `await` inside `if` conditions.

Though I think we do this when we are reusing the options, but to keep consistency let me update this one.

packages/shared/src/components/modals/SquadMemberModal.tsx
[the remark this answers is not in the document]
rebelchris: Rather consult with Tomer before changing it :)

packages/shared/src/components/squads/SquadMemberActions.tsx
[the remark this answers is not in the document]
rebelchris: But you deconstruct this right above no?

packages/shared/src/components/squads/SquadMemberRoleBadge.tsx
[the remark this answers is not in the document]
sshanzel: They have different prop requirements. The other one requires a lot of objects to determine the role, the new one requires just the role which is used by the old one.

packages/shared/src/components/squads/SquadMemberActions.tsx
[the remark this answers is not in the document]
sshanzel: Yes, though that deconstructed value came from context, and the member came from the prop.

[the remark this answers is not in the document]
sshanzel: Oh, wait. I just checked the code, yep, it seems to be accurate using `user`. Good catch!

packages/shared/src/components/squads/SquadMemberActions.tsx
rebelchris: Yeah either way has it's charm, maybe something we can create a rule for?

packages/shared/src/components/squads/SquadMemberRoleBadge.tsx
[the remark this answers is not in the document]
rebelchris: Hmm, but you could abstract the role in component above and render `SquadMemberBadge` with the role directly?
It feels a bit weird having 2 exports now and the naming is a bit confusing.

Thinking about looking at this 3 months down the line people might use the wrong component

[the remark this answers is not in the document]
sshanzel: The reason for the 2 exports is to support both use cases. One for simply showing the appropriate badge for anyone that can provide a role. The other one is for displaying it in the comment section where it uses a hook that abstracted something and has more requirements.

Probably we should rename the old component to be more accurate that it was meant for comment section.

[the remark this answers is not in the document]
capJavert: I like the wrapper solution but I think it is better to split those two into two separate files, I don't think we use these kind of mixed exports in other parts of the project?

packages/shared/src/graphql/fragments.ts
```diff
@@ -65,6 +65,7 @@ export const SOURCE_BASE_FRAGMENT = gql`
     membersCount
     currentMember {
       role
+      roleRank
       referralToken
       permissions
     }
```
rebelchris, line 71: Will have some issue with current fragment changes

packages/shared/src/graphql/sources.ts
```diff
@@ -25,6 +25,7 @@ export enum SourcePermissions {
 
 export interface SourceMember {
   role: SourceMemberRole;
+  roleRank: number;
```
rebelchris, line 28: I think @capJavert removed this number exposure now.
Maybe should align that?

packages/shared/src/components/squads/SquadMemberItemAdditionalContent.tsx
```diff
@@ -0,0 +1,88 @@
...
+  onOptionsClick,
+}: SquadMemberActionsProps): ReactElement {
+  const { showPrompt } = usePrompt();
+  const { user: loggedUser } = useAuthContext();
+  const { role, user } = member;
+
+  const onConfirmUnblock = async (e: React.MouseEvent) => {
+    e.preventDefault();
+
+    const options: PromptOptions = {
```
rebelchris, line 31: This can be outside the function (optional comment)

packages/shared/src/components/squads/SquadMemberAuthorBadge.tsx
```diff
@@ -41,12 +44,22 @@ const SquadMemberRoleBadge = ({
...
+      className={classNames('text-theme-color-cabbage', className)}
       content={role}
       Icon={Icon}
       iconProps={iconProps}
     />
   );
 };
 
-export default SquadMemberRoleBadge;
+const SquadMemberAuthorBadge = ({
```
rebelchris, line 55: Think we should abstract to 2 component, and not 100% sure on naming here.
but might be because I'm missing context of this change now.

rebelchris (changes requested):

packages/shared/src/graphql/sources.ts
sshanzel: Oh, why did we remove it?

packages/shared/src/components/squads/SquadMemberItemAdditionalContent.tsx
sshanzel: I think we can keep this here, so JS won't have to create this object until necessary.

packages/shared/src/graphql/fragments.ts
sshanzel: I'll check if there is any conflict and coordinate with Ante if necessary.

packages/shared/src/components/squads/SquadMemberAuthorBadge.tsx
sshanzel: What would we want to be abstracted into 2 components? I mean the content of this component is just 2 lines of code. Do you want me to delete this component and just call the hook on where this was rendered?

packages/shared/src/graphql/sources.ts
capJavert: I removed the one for posting permission (different field on Source). But generally I think frontend does not use numeric ranks anywhere, all checks are done through permissions so maybe this is not needed too.

packages/shared/src/components/modals/SquadMemberModal.tsx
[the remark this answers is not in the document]
capJavert: Should we not check `verifyPermission(squad, SourcePermissions.ViewBlockedMembers)` instead?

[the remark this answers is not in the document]
capJavert: This would also remove the need for frontend to know about role ranks.

packages/shared/src/components/modals/SquadMemberModal.tsx
capJavert: Sorry, missed this, what guideline are we talking about specifically?

packages/shared/src/components/squads/SquadMemberAuthorBadge.tsx
rebelchris: Let's go with the second option in that case.
Just to avoid confusion couple months down the line.

packages/shared/src/components/modals/SquadMemberModal.tsx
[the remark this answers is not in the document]
rebelchris: Good point, sounds like a good plan to me, wdyt @sshanzel ?

packages/shared/src/graphql/sources.ts
sshanzel: Oh, cool. Now that we have a use case here, we can now probably keep it.

rebelchris: @sshanzel but is it no better to evaluate actual permissions like Ante suggested to check if someone can view blocked members? (Seems more inline with what we do)

sshanzel: Oh, right, we have a new permission for blocked members. We can probably use that.

packages/shared/src/components/modals/SquadMemberModal.tsx
rebelchris: Mainly modal guideline, normally the stepped modal expect 2 seperate views.
Here we technically "hacked" a workaround for updates outside the modal.

rebelchris (approved):
Yeah, thinks this is more explicit and less error prone.
If we see more use for this role badge we can abstract into custom component.

packages/shared/src/components/modals/SquadMemberModal.tsx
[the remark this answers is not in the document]
sshanzel: Oh, just saw this now. Yes this should be the most updated implementation. Thank you both!

packages/shared/src/components/modals/SquadMemberModal.tsx
capJavert: If in all other places we do it with two views then this one should be no exception. But don't have the hard preference because above is simple logic wise.

capJavert (approved):

PR #1719, opened 2023-04-04
feat(squads) update posting permission

capJavert:
- Add permissions to edit squad form
- API implementation here https://github.com/dailydotdev/daily-api/pull/1228

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1196 #done

### DIFF
--- a/packages/shared/src/components/modals/EditSquadModal.tsx
+++ b/packages/shared/src/components/modals/EditSquadModal.tsx
@@ -47,7 +47,7 @@ function EditSquadModal({
       size={Modal.Size.Small}
       onRequestClose={onRequestClose}
     >
-      <Modal.Header title="Squad details" kind={ModalHeaderKind.Tertiary} />
+      <Modal.Header title="Squad settings" kind={ModalHeaderKind.Primary} />
       <SquadDetails form={squad} onSubmit={onSubmit} createMode={false} />
     </Modal>
   );
--- a/packages/shared/src/components/modals/EditSquadModal.tsx
+++ b/packages/shared/src/components/modals/EditSquadModal.tsx
@@ -4,7 +4,6 @@ import { LazyModalCommonProps, Modal } from './common/Modal';
 import { editSquad } from '../../graphql/squads';
 import { Squad } from '../../graphql/sources';
 import { SquadDetails } from '../squads/Details';
-import { ModalHeaderKind } from './common/types';
 import { useToastNotification } from '../../hooks/useToastNotification';
 import { useBoot } from '../../hooks/useBoot';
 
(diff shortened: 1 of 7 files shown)


packages/shared/src/components/modals/EditSquadModal.tsx
```diff
@@ -46,7 +47,7 @@ function EditSquadModal({
       size={Modal.Size.Small}
       onRequestClose={onRequestClose}
     >
-      <Modal.Header title="Squad details" kind={ModalHeaderKind.Tertiary} />
+      <Modal.Header title="Squad settings" kind={ModalHeaderKind.Primary} />
```
rebelchris, line 50: Primary is the default so we can omit it here.

packages/shared/src/components/squads/Details.tsx
```diff
@@ -191,24 +191,22 @@ export function SquadDetails({
...
-                name="memberPostingRole"
-                options={memberPostingRoleOptions}
-                value={memberPostingRole}
-                onChange={(value) =>
-                  setMemberPostingRole(value as SourceMemberRole)
-                }
-              />
-            </div>
-          )}
+          <div className="flex flex-col justify-start">
```
rebelchris, line 194: Don't think this adds anything.

```suggestion
          <div className="flex flex-col">
```

packages/shared/src/components/squads/Details.tsx
```diff
@@ -191,24 +191,22 @@ export function SquadDetails({
...
+            <h4 className="mb-2 font-bold typo-headline text-theme-label-primary">
```
rebelchris, line 195: This should be enough:
```suggestion
            <h4 className="mb-2 typo-headline">
```

packages/shared/src/components/squads/utils.tsx
```diff
@@ -4,7 +4,7 @@ import { PromptOptions } from '../../hooks/usePrompt';
 import { SquadForm } from '../../graphql/squads';
 
 export enum ModalState {
-  Details = 'Squad details',
```
rebelchris, line 7: Is this actually still used? Not quite sure we use this flow anymore.

rebelchris (changes requested):
Some minor comments, but otherwise happy.

packages/shared/src/components/squads/utils.tsx
capJavert: It is, but below it `modalStateToScreenValue`, left it because I did not know the context.

rebelchris: Ah let's leave it might be needed for typesafety or something

rebelchris (approved):
Amazing, nice work 🙏

PR #1721, opened 2023-04-05
fix: missing context menu items on feed cards

sshanzel:
## Changes

The context menu for post modal and the feed card was not unified in terms of behavior but using the same component.

Note: pointed to main as removing post is already available.

- Since both areas were using the same context menu component, we've updated how we identify allowing deletion of post.
- Also refactored the fragment to be unified so nothing will be missed out when looking for the information needed to identify whether the user can delete a post in the regular feed.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1218 #done

### DIFF
--- a/packages/shared/src/graphql/fragments.ts
+++ b/packages/shared/src/graphql/fragments.ts
@@ -24,33 +24,6 @@ export const SOURCE_SHORT_INFO_FRAGMENT = gql`
   }
 `;
 
-export const SHARED_POST_INFO_FRAGMENT = gql`
-  fragment SharedPostInfo on Post {
-    id
-    title
-    image
-    readTime
-    permalink
-    commentsPermalink
-    summary
-    createdAt
-    private
-    scout {
-      ...UserShortInfo
-    }
-    author {
-      ...UserShortInfo
-    }
-    type
-    tags
-    source {
-      ...SourceShortInfo
-    }
-  }
-  ${SOURCE_SHORT_INFO_FRAGMENT}
-  ${USER_SHORT_INFO_FRAGMENT}
-`;
-
 export const SOURCE_BASE_FRAGMENT = gql`
   fragment SourceBaseInfo on Source {
     id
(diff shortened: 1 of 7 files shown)


sshanzel (commented):
Have checked all fragments and every property was kept. Also tested all the feeds on preview deployment.

packages/shared/src/graphql/fragments.ts
```diff
@@ -24,6 +24,26 @@ export const SOURCE_SHORT_INFO_FRAGMENT = gql`
...
+    public
+    type
+    description
+    image
+    membersCount
+    currentMember {
+      role
+      referralToken
+      permissions
+    }
```
rebelchris, line 43: Just be aware this will need the new fragment

packages/shared/src/hooks/usePostMenuActions.ts
```diff
@@ -2,14 +2,23 @@ import { useMutation } from 'react-query';
...
+}
+
 interface UsePostMenuActionsProps {
   post: Post;
-  onPostDeleted?: () => void;
+  postIndex?: number;
+  onPostDeleted?: (args: DeletePostProps) => void;
 }
 
 const deletePromptOptions: PromptOptions = {
```
rebelchris, line 24: title should be: `Delete post?`
description: `Are you sure you want to delete this post?`
button color should be cabbage.

rebelchris (changes requested):
Nice work on uniforming this.
Have some sketch related changes

packages/shared/src/graphql/fragments.ts
sshanzel: Not sure that is already available here. We should be able to update this once we make the Squads v4 PR ready.

sshanzel: Have fixed merge conflict. Should be updated now.

rebelchris (approved):
Awesome!

sshanzel:
Have confirmed everything works as intended 🥳

PR #1722, opened 2023-04-05
feat: indicate when comment author has no access

sshanzel:
## Changes
- Request a query to check the user's membership.
- The type for the comments area is a bit messy, so I've taken the liberty to do some code cleanups. There were a lot of spread properties where some are used in other places and some are not. Taking the advice Ante has suggested in the past, I have introduced a single param for the whole object.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1202 #done


packages/shared/src/hooks/usePostComment.ts
```diff
@@ -40,6 +40,45 @@ const getParentAndCurrentIndex = (
   return [parent, current];
 };
 
+interface SharedData {
+  editContent?: string;
+  editId?: string;
+}
+
+export const getParentComment = (
```
rebelchris, line 48: This actually feels like we can uniform the return based on type maybe?
(optional comment)

packages/shared/src/graphql/squads.ts
```diff
@@ -183,6 +183,14 @@ export const SQUAD_JOIN_MUTATION = gql`
   ${SOURCE_BASE_FRAGMENT}
 `;
 
+export const CHECK_USER_MEMBERSHIP = gql`
+  query CheckUserMembership($userId: ID!, $sourceId: ID!) {
```
rebelchris, line 187: See my comment on API side, think we can make it more easy.

rebelchris (changes requested):
Wow thanks for the massive cleanup.
Be sure to spend some time testing it as it has high impact if something stops working here.

packages/shared/src/hooks/usePostComment.ts
sshanzel: I think we already have a return type of `ParentComment`. Do you mean something else?

rebelchris: Ah yes, I mean the both returns are basically the same was wondering if there was a way to uniform it into 1?

sshanzel: Hmm. Not really sure how since each property that would be coming from either a post or comment was explicitly defined. Do you have any suggestions?

rebelchris: Really not urgent, and not sure it's even better.
But something like

```
const commentProps = { author: name, content, etc } = comment;
const postProps = { source: name, title, etc } = post;
props = com ? comment ? commentProps : postProps;
```

As mentioned really not sure if it even makes sense, the fragment of code just looked as it could be optimized for this.

sshanzel: I agree, the code block looks a bit not looking good. Perhaps we can go back here once we think of something to replace this entirely. WDYT?

rebelchris: Yeah 100% I'll log it for future discussion :)

sshanzel:
@rebelchris so far this should still support whether the API will throw an error or not since we check for Blocked members.

rebelchris (approved):
Tested locally, was one issue where the key was firing for all queries, added authorID to make it unique per author :).

rebelchris:
@capJavert When you get a moment, could you please also review it.

packages/shared/src/components/modals/NewCommentModal.tsx
```diff
@@ -107,19 +108,29 @@ const PreviewTab = ({ input, sourceId, parentSelector }: PreviewTabProps) => {
...
 }: NewCommentModalProps): ReactElement {
+  const { handle, commentId, editId, editContent, replyTo, authorId } =
+    parentComment;
   const [input, setInput] = useState<string>(editContent || replyTo);
   const [sendingComment, setSendingComment] = useState<boolean>(false);
   const [errorMessage, setErrorMessage] = useState<string>(null);
   const { showPrompt } = usePrompt();
   const { requestMethod } = useRequestProtocol();
+  const { data: member, isFetched: isMembershipFetched } = useQuery(
+    generateQueryKey(RequestKey.PostComments, null, authorId),
```
capJavert, line 126: @rebelchris we should add `post.source.id` also to the query key.

Data would potentially be invalid if looking at the comment from the same author on different source (squad) because it would already be populated from previous comment membership.

packages/shared/src/lib/query.ts
```diff
@@ -12,14 +12,15 @@ export type MutateFunc<T> = (variables: T) => Promise<(() => void) | undefined>;
 
 export const generateQueryKey = (
   name: string | RequestKey,
-  user: Pick<LoggedUser, 'id'> | null,
+  user: Pick<LoggedUser, 'id'> = null,
```
capJavert, line 15: Not blocking but do we need default null` here, since there is a check where no user always fallbacks to `anonymous`

capJavert (changes requested):
Query key still missing source id, other then that 👍

packages/shared/src/components/modals/NewCommentModal.tsx
rebelchris: Ah good point indeed! I'll add that :)

packages/shared/src/lib/query.ts
rebelchris: Seems unneeded indeed, moved back to old style.

capJavert (approved):

packages/shared/src/lib/query.ts
sshanzel: I mean, that's the point. Instead of doing `generateQueryKey(key, null)` we can just do `generateQueryKey(key)`.

PR #1723, opened 2023-04-05
feat(squads): disabled post bar state

capJavert:
- Disabled post bar state when posting to squad is not allowed

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done


packages/shared/src/graphql/squads.ts
```diff
@@ -303,3 +349,26 @@ export async function editSquad(
...
+    if (membersCanPost) {
+      return true;
+    }
+
+    return [SourceMemberRole.Owner, SourceMemberRole.Moderator].includes(
+      squad.currentMember.role,
+    );
+  }
+
+  return hasPermission;
```
capJavert, line 373: @rebelchris so what I want to do is to check `memberPostingRole` in addition to `SourcePermissions.Post` permission. Problem is this will need to be updated if we add more permission options for users in squad settings eg. only owners can post. 

Now it works because by default everyone can post and you can only set moderator or above in squad creation/settings.

I was thinking maybe we can expose array of roles which would then be used for the check. It would be simple filter on the API side which would then only return roles that are higher then `memberPostingRank` and we could use that instead of hardcoded array above to check.

Other thing is that I might have been wrong and exposing `squad.memberRoleRank` and `currentMember.roleRank` (numeric values) to the frontend would be useful and then we can just check if current rank value is higher then the one set for a squad. 

WDYT?

In hindsight it would be awesome if we could somehow just omit the `SourcePermissions.Post` permission form `currentMember.permissions` depending on the `sourceMemberRole` squad setting and current user on the API side. Not sure how performant that query would be.

rebelchris: I agree the later is actually best case.
Dependant on squad `memberPostingRank` we evaluate if permissions should include `Post`.

Happy to include that change as it makes things more explicit and clear, let's do that?

capJavert: @rebelchris so the last one on the API side?

capJavert: Implementation here https://github.com/dailydotdev/daily-api/pull/1232

rebelchris (approved):
Wow this is so clean now, way to go.
Was worth the extra effort I think.

capJavert:
@rebelchris also added implementation for `WT-1225`.

rebelchris (approved):
Let's go 👍

packages/shared/src/components/squads/SharePostBar.tsx
```diff
@@ -17,13 +22,15 @@ export type NewSquadPostProps = Pick<
 interface SharePostBarProps {
   className?: string;
   onNewSquadPost?: (props?: NewSquadPostProps) => void;
+  squad: Squad;
```
sshanzel, line 25: Since this is like an input component for us as a whole, we can probably just accept a `disabled` boolean property than a whole object. Then we can proceed with checking whether to return the placeholder or not. WDYT?

sshanzel (approved):
Great work incorporating the boot squads! Just had one non-blocking comment for simpler checking.

packages/shared/src/components/squads/SharePostBar.tsx
capJavert: Great idea ✅

PR #1726, opened 2023-04-07
feat(dev): alert clipboard content to avoid app crash

capJavert:
In `WT-792` I noticed that navigator.clipboard is also undefined in development since our custom domain `webapp.local.com` is not served over HTTPS

Info from here [Unblocking clipboard access](https://web.dev/async-clipboard) 
> As with many new APIs, the Clipboard API is only supported for pages served over HTTPS. To help prevent abuse, clipboard access is only allowed when a page is the active tab. Pages in active tabs can write to the clipboard without requesting permission, but reading from the clipboard always requires permission.

serving on localhost  is the only way to have `navigator.clipboard` available without https. Since we we need our custom domain for local authentication to work we need an alternative.

To not crash the app in development and allow us to test those features easily I added development only check to `alert` the clipboard contents instead. I think it is an ok alternative for local DX.
image

Also, since it is wrapped in development check bundler will automatically remove it from production bundle so no overhead there.

Interested to see what you think or have other ideas.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done

### DIFF
--- a/packages/shared/src/hooks/useCopyLink.ts
+++ b/packages/shared/src/hooks/useCopyLink.ts
@@ -3,6 +3,7 @@ import {
   NotifyOptionalProps,
   useToastNotification,
 } from './useToastNotification';
+import { isDevelopment } from '../lib/constants';
 
 type CopyNotifyFunctionProps = NotifyOptionalProps & {
   message?: string;
@@ -25,10 +26,7 @@ export function useCopyLink(
     const link = getLink();
 
     if (link) {
-      if (
-        process.env.NODE_ENV === 'development' &&
-        !window.location.origin.includes('localhost')
-      ) {
+      if (isDevelopment) {
         // since we are using custom domain in dev, we can't use the clipboard API
         // this is a workaround to not crash the app in development and still be able
         // to test other stuff while working with clipboard
--- a/packages/shared/src/hooks/useCopyLink.ts
+++ b/packages/shared/src/hooks/useCopyLink.ts
@@ -25,7 +25,20 @@ export function useCopyLink(
     const link = getLink();
 
     if (link) {
-      await navigator.clipboard.writeText(link);
+      if (
+        process.env.NODE_ENV === 'development' &&
+        !window.location.origin.includes('localhost')
+      ) {
+        // since we are using custom domain in dev, we can't use the clipboard API
+        // this is a workaround to not crash the app in development and still be able
+        // to test other stuff while working with clipboard
+        // more info here https://web.dev/async-clipboard
+        // eslint-disable-next-line no-alert
+        alert(`Dev copied: ${link}`);
+      } else {
+        await navigator.clipboard.writeText(link);
+      }
+
       displayToast(props.message || defaultMessage, props);
     } else {
       displayToast(noLinkErrorMessage, props);


sshanzel:
I never had issues with the navigator in the past. Probably we can just try catch it for local development so we can keep where it works. WDYT?

cc @rebelchris

packages/shared/src/hooks/useCopyLink.ts
```diff
@@ -25,7 +25,20 @@ export function useCopyLink(
     const link = getLink();
 
     if (link) {
-      await navigator.clipboard.writeText(link);
+      if (
+        process.env.NODE_ENV === 'development' &&
+        !window.location.origin.includes('localhost')
```
rebelchris, line 30: Why do we need the localhost check here?
We can simply exclude all dev right?

Also we do have a `isDevelopment` export already :)

rebelchris (changes requested):

packages/shared/src/hooks/useCopyLink.ts
capJavert: as noted in description about `navigator.clipboard`

> serving on localhost  is the only way to have `navigator.clipboard` available without https. Since we we need our custom domain for local authentication to work we need an alternative.

capJavert: I wanted to add it as progressive enhancement so if in any way clipboard is accessible it would work, eg. our frontend does work on `localhost` aside from auth features.

My greater concern is to see if there is a better way to handle this aside from `alert`.

sshanzel (changes requested):
I'm not really sure we should always opt for this to alert whenever in development, can become quite annoying with the all the alerts. On my end, it works just fine on webapp. As suggested, we should probably have this in a try-catch and then check if it is in development env.

sshanzel:
Yes, I am using `http://webapp.local.com:5002` but have no custom SSL certificate nor any special setup. I mean I don't remember anything, as far as setting my unit, and also that I am using a fairly new unit.

capJavert:
> Yes, I am using `http://webapp.local.com:5002` but have no custom SSL certificate nor any special setup. I mean I don't remember anything, as far as setting my unit, and also that I am using a fairly new unit.

😮 I would be curious to see how this works on your machine because by spec `navigator.clipboard` should not be available in Chrome on non https websites (except for `localhost`). 

So if you click on copy squad invite in development you don't get runtime error and link is copied to your clipboard?
image

sshanzel:
> So if you click on copy squad invite in development you don't get runtime error and link is copied to your clipboard?

No, I don't 😛 

Result: http://webapp.local.com:5002/squads/sample/_NbO3CwS2COjh9MboIDY7IgvbdOq270UBqe8AmmXBdc

image

capJavert:
I notice you do not have `Not secure` next to URL bar.. Curious..

image

sshanzel:
When I click the icon, it shows me this.

image

capJavert:
@sshanzel you have this enabled I think in `chrome://flags`
image

sshanzel:
> @sshanzel you have this enabled I think in `chrome://flags`

Just checked now, and that is right! Then I think we can just specify our origins here and avoid the development check altogether.

image

capJavert:
> Just checked now, and that is right! Then I think we can just specify our origins here and avoid the development check altogether.

Yep 🤝

PR #1727, opened 2023-04-07
fix: feed upvote counter now syncs with single post counter

capJavert:
- Leveraged mutationKeys to identify upvote mutations
- Subscribe to upvote mutations inside feed upvote hook
- Listening to any mutation that was not coming from feed I update the upvote counter in feed to keep them in sync
- Also works with cancel upvote

Closes https://github.com/dailydotdev/daily/issues/830

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1135 #done


sshanzel:
@capJavert can you publish this on some unused preview deployments? Just want to test it out on prod.

packages/shared/src/hooks/feed/useFeedUpvotePost.ts
```diff
@@ -27,24 +49,72 @@ export default function useFeedUpvotePost(
...
+        mutationHandlers[upvotePostKey],
+      ),
+      onCancelPostUpvoteMutate: optimisticPostUpdateInFeed(
+        items,
+        updatePost,
+        mutationHandlers[cancelUpvotePostKey],
+      ),
+    });
+
+  useEffect(() => {
```
rebelchris, line 70: How come the above mutation doesn't catch this?
Seems like a lot of overhead to handle a optimistic update, but might miss the context in this PR

packages/shared/src/hooks/feed/useFeedUpvotePost.ts
```diff
@@ -27,24 +49,72 @@ export default function useFeedUpvotePost(
...
+
+        if (index === -1) {
+          return;
+        }
+
+        const mutationHandler = mutationHandlers[mutationKey];
+
+        if (!mutationHandler) {
+          return;
+        }
```
rebelchris, line 104: This could be outside the if even, as it would be early return right?

rebelchris (commented):
Would need some more background on the function I think to evaluate properly.

packages/shared/src/hooks/feed/useFeedUpvotePost.ts
capJavert: So basically the core issue is that upvote data is stored in two (or more) different queries. 

1. For single posts inside `['post', id]`
2. For different feeds (every feed has it's own query keys)

Issue is that `useUpvotePost` mutation's `onSuccess` handlers only modify/optimistically update data inside `1.` query key. 

As for `2.` updating data inside all feed queries would not really be possible since each feed has completely different query key.

Ultimately we would have to refactor our query keys and data on frontend to have a separate query to hold all the upvote data or additional query for each post upvote state. I actually tried going with first idea initially but realised this does not match our API data structure so I found it inefficient to remap all of this for this one user journey.

What I did instead was to inverse the logic and instead of manually trying to update query keys of all related queries I simply subscribe to mutations from active feed itself. Cool bonus is that since squad rely on the feed this automatically works there too.

This way:
- only active/rendered feed actually listens for data and optimistically updates
- no overhead in `useUpvotePost` upvote hook
- we can keep our current data structure and solve this pretty common user journey with minimal changes

FYI @sshanzel I've put this on preview https://preview.app.daily.dev

capJavert: > Moving forward, would you be able to create a reusable hook for such cases? I think we have the same case for bookmarks and comments too.

I would do this on the next task that touches this use case, not sure what would be needed to make the hook reusable at this time. The real fix should be revamping our query keys though.

> This got me thinking, would it be not easier to just pass a callback on the using modal (since it came from the feed) for when we upvote/comment/bookmark the post?

I personally love that this is contained inside a hook that is already specialised for feed upvotes without any additional logic sprinkled through callbacks on other components. This highlights power of global mutations which we get for free in react-query.

packages/shared/src/hooks/feed/useFeedUpvotePost.ts
capJavert: Oh yeah this could actually replace `matchedMutations.includes(mutationKey)` on the top. 👍

packages/shared/src/hooks/feed/useFeedUpvotePost.ts
```diff
@@ -27,24 +49,61 @@ export default function useFeedUpvotePost(
...
+      const { id, index: indexFromEvent } = event.options
+        .variables as unknown as UseFeedUpvotePostVariables;
+      const mutationFromFeedHook = typeof indexFromEvent !== 'undefined';
+
+      if (mutationFromFeedHook) {
+        return;
+      }
+
+      const index = items.findIndex(
+        (item) => item.type === 'post' && item.post.id === id,
```
rebelchris, line 93: What other types do we support?
Not sure that check is needed.

rebelchris (approved):
I personally agree with Lee, think a generic hook would work.
But for now also happy with this, we can note that on the backlog.

packages/shared/src/hooks/feed/useFeedUpvotePost.ts
rebelchris: Alright fine :)
Was just surprised to see this 😂

capJavert:
@idoshamun @sshanzel would like to hear your final yay or nay 😆 for this PR. Just so we can decide if this direction is ok or we wanna take a step back and look at other implementations mentioned in the comments above. 

Personally I think it brings enough values for our users to be merged with current fix.

idoshamun (approved):
Love the overall approach didn't go into details.
Great thinking!

capJavert:
Thanks for the feedback merging now 👌

PR #1730, opened 2023-04-11
fix: optional data for when modal loading

rebelchris:
- With new later return we need to make props optional to not-break rendering.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1250 #done

### DIFF
--- a/packages/shared/src/components/post/SquadPostContent.tsx
+++ b/packages/shared/src/components/post/SquadPostContent.tsx
@@ -57,8 +57,8 @@ function SquadPostContent({
   const engagementActions = usePostContent({ origin, post });
   const { onReadArticle, onSharePost, onToggleBookmark } = engagementActions;
   const { role } = useMemberRoleForSource({
-    source: post.source,
-    user: post.author,
+    source: post?.source,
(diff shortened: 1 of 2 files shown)


capJavert (changes requested):
Nice forward thinking.

I would also just add `source?.` inside `useMemberRoleForSource` and make it optional since right now `source` is expected to always be present. Not sure why linter does not flag this as error.

sshanzel (approved):

rebelchris:
@sshanzel Holding off till we have that one in place and will re-look how it performs or is needed :)

sshanzel:
I think the fix for the loading placeholder worked alright. But we can probably keep this just to be on the safe side always. Either way sounds good to me 🚢

capJavert (changes requested):
We still need to make `source` optional in `UseMemberRoleForSourceProps` type, don't know why typescript is not screaming for this.

capJavert (approved):

PR #1732, opened 2023-04-11
fix: forbidden page throwing error

sshanzel:
## Changes
- Fixes the issue where forbidden throws an error instead of showing the 404 page. Since we now support requesting the static information to fill the page, our previous condition doesn't work anymore.
- Also saw some discrepancies on the loading page when the error happens. Since there are two kinds (page and modal) of when to display it, I have refactored it to render from where it needs to happen (page and modal) than fitting to all 4 cases (all combinations).
- Fixed some typos.
- Have tested all four cases thoroughly. Will test it in preview once a deployment gets free tomorrow.
- Code-wise though this should be reviewable. Feedback would be highly appreciated, had to test all 4 cases every change, so I might have missed something.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1238 #done

### DIFF
--- a/packages/shared/src/components/modals/ArticlePostModal.tsx
+++ b/packages/shared/src/components/modals/ArticlePostModal.tsx
@@ -14,7 +14,6 @@ import { PostType } from '../../graphql/posts';
 
 interface ArticlePostModalProps extends ModalProps, PassedPostNavigationProps {
   id: string;
-  isFetchingNextPage?: boolean;
 }
 
 const containerClass = 'border border-theme-divider-tertiary rounded-16';
@@ -22,7 +21,6 @@ const containerClass = 'border border-theme-divider-tertiary rounded-16';
 export default function ArticlePostModal({
   id,
   className,
-  isFetchingNextPage,
   onRequestClose,
   onPreviousPost,
   onNextPost,
--- a/packages/shared/src/components/modals/ArticlePostModal.tsx
+++ b/packages/shared/src/components/modals/ArticlePostModal.tsx
@@ -10,6 +10,7 @@ import usePostNavigationPosition from '../../hooks/usePostNavigationPosition';
 import usePostById from '../../hooks/usePostById';
 import BasePostModal from './BasePostModal';
 import OnboardingContext from '../../contexts/OnboardingContext';
+import { PostType } from '../../graphql/posts';
 
 interface ArticlePostModalProps extends ModalProps, PassedPostNavigationProps {
   id: string;
@@ -28,15 +29,26 @@ export default function ArticlePostModal({
   ...props
 }: ArticlePostModalProps): ReactElement {
   const { showArticleOnboarding } = useContext(OnboardingContext);
-  const { post, isLoading } = usePostById({ id, isFetchingNextPage });
+  const {
+    post,
+    query: { isLoading, isFetching },
+  } = usePostById({ id });
+  const isPostLoading = isLoading || isFetchingNextPage || isFetching;
   const position = usePostNavigationPosition({
-    isLoading,
+    isLoading: isPostLoading,
     isDisplayed: props.isOpen,
     offset: showArticleOnboarding ? ONBOARDING_OFFSET : 0,
   });
+  const containerClass = 'border border-theme-divider-tertiary rounded-16';
 
   return (
-    <BasePostModal {...props} onRequestClose={onRequestClose}>
+    <BasePostModal
+      {...props}
+      onRequestClose={onRequestClose}
+      loadingClassName={containerClass}
+      postType={PostType.Article}
+      isLoading={isPostLoading}
+    >
       <PostContent
         position={position}
         post={post}
@@ -46,7 +58,7 @@ export default function ArticlePostModal({
         inlineActions
         className={{
           onboarding: 'mt-8',
-          container: 'border border-theme-divider-tertiary rounded-16',
+          container: containerClass,
           navigation: { actions: 'tablet:hidden ml-auto' },
           fixedNavigation: {
             container: modalSizeToClassName[Modal.Size.XLarge],
@@ -54,7 +66,6 @@ export default function ArticlePostModal({
           },
         }}
         onClose={onRequestClose}
-        isLoading={isLoading}
         origin={Origin.ArticleModal}
         onRemovePost={onRemovePost}
       />
--- a/packages/shared/src/components/modals/ArticlePostModal.tsx
+++ b/packages/shared/src/components/modals/ArticlePostModal.tsx
@@ -31,13 +31,9 @@ export default function ArticlePostModal({
   ...props
 }: ArticlePostModalProps): ReactElement {
   const { showArticleOnboarding } = useContext(OnboardingContext);
-  const {
-    post,
-    query: { isLoading, isFetching },
-  } = usePostById({ id });
-  const isPostLoading = isLoading || isFetchingNextPage || isFetching;
+  const { post, isPostLoadingOrFetching } = usePostById({ id });
   const position = usePostNavigationPosition({
-    isLoading: isPostLoading,
+    isLoading: isPostLoadingOrFetching,
     isDisplayed: props.isOpen,
     offset: showArticleOnboarding ? ONBOARDING_OFFSET : 0,
   });
@@ -48,7 +44,7 @@ export default function ArticlePostModal({
       onRequestClose={onRequestClose}
       loadingClassName={containerClass}
       postType={PostType.Article}
-      isLoading={isPostLoading}
+      isLoading={isPostLoadingOrFetching}
     >
       <PostContent
         position={position}
--- a/packages/shared/src/components/modals/ArticlePostModal.tsx
+++ b/packages/shared/src/components/modals/ArticlePostModal.tsx
@@ -17,6 +17,8 @@ interface ArticlePostModalProps extends ModalProps, PassedPostNavigationProps {
   isFetchingNextPage?: boolean;
 }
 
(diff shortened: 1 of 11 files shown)


packages/shared/src/components/modals/ArticlePostModal.tsx
```diff
@@ -28,15 +29,26 @@ export default function ArticlePostModal({
...
+    query: { isLoading, isFetching },
+  } = usePostById({ id });
+  const isPostLoading = isLoading || isFetchingNextPage || isFetching;
   const position = usePostNavigationPosition({
-    isLoading,
+    isLoading: isPostLoading,
     isDisplayed: props.isOpen,
     offset: showArticleOnboarding ? ONBOARDING_OFFSET : 0,
   });
+  const containerClass = 'border border-theme-divider-tertiary rounded-16';
```
rebelchris, line 42: Why did we decide to place this outside the direct implementation?
This case we could even place it outside the component.

packages/shared/src/components/modals/ArticlePostModal.tsx
```diff
@@ -28,15 +29,26 @@ export default function ArticlePostModal({
   ...props
 }: ArticlePostModalProps): ReactElement {
   const { showArticleOnboarding } = useContext(OnboardingContext);
-  const { post, isLoading } = usePostById({ id, isFetchingNextPage });
+  const {
+    post,
+    query: { isLoading, isFetching },
+  } = usePostById({ id });
+  const isPostLoading = isLoading || isFetchingNextPage || isFetching;
```
rebelchris, line 36: Just for my clarity the `isFetchingNextPage` can happen when the modal is open and you press next?

packages/shared/src/components/modals/SharePostModal.tsx
```diff
@@ -26,18 +27,26 @@ export default function PostModal({
   ...props
 }: PostModalProps): ReactElement {
   const { showArticleOnboarding } = useContext(OnboardingContext);
-  const { post, isLoading } = usePostById({ id, isFetchingNextPage });
+  const {
+    post,
+    query: { isLoading, isFetching },
+  } = usePostById({ id });
+  const isPostLoading = isLoading || isFetchingNextPage || isFetching;
```
rebelchris, line 34: Just a thought is there no way we can directly export this from the `usePostbyId` fn as we reuse it two times like this? 
Just to avoid changes in one place down the line.

packages/shared/src/hooks/usePostById.ts
```diff
@@ -38,36 +37,21 @@ export const updatePostCache = (
...
 
-  return useMemo(
-    () => ({
-      post: post?.post,
-      isFetched,
-      isLoading: isLoading || isFetchingNextPage,
-    }),
-    [id, postById, isLoading, options, isFetched, isFetchingNextPage],
-  );
+  return useMemo(() => ({ post: post?.post, query }), [query]);
```
rebelchris, line 54: Think you will always need `id` here to reset memo no?
Not sure query alone will fetch that.

rebelchris (changes requested):
Wow nice cleanup, some minor comments.
And would like to ask you to really take some time to test use-cases with this, it seems to have high impact as some of these elements previously broke certain aspects of the app.

packages/shared/src/components/modals/ArticlePostModal.tsx
sshanzel: We used it both at the `BasePostModal` to apply the same class on the container for the loading placeholder.

packages/shared/src/components/modals/ArticlePostModal.tsx
sshanzel: Yes. That's right! 🚀

packages/shared/src/components/modals/SharePostModal.tsx
sshanzel: I agree with your thoughts. I tried that first, I extracted some of these properties and exposed it down from the hook, then another one. Thought of using `Pick<UseQueryResult, 'property'>` for the returned type but as it grows it just becomes too much to do it one by one every time, IMHO.
It made me realize to just return the query itself so if we will have another thing to destructure from the query, we won't have to update the hook by what is already available from the query itself.

But that is definitely an option we can refactor to. WDYT?

packages/shared/src/hooks/usePostById.ts
sshanzel: The id is part of the query key, so when its key changes, it re-fires which will then update the `query`, and since the `query` gets updated, this gets reset. But I can definitely add the `id` for dependency to avoid any weird edge cases.

packages/shared/src/components/modals/ArticlePostModal.tsx
rebelchris: But this is not exported? 👀
So it's just duplicated, not quite sure I understand why it's abstracted to a const then?

packages/shared/src/components/modals/SharePostModal.tsx
rebelchris: Hmm, well the concern is that it makes it very opinionated.
You know today that you need all 3 to evaluate, but what happens if Ante needs to work on this next week, he  might have no clue about that constraint or why it's needed.

Thus I think the more verbose the hook is, the better it will be.
CC: @capJavert Would love to get your thoughts on this as well.

packages/shared/src/components/modals/ArticlePostModal.tsx
sshanzel: Oh, right we can take it outside this component and be a const in the file. It is not abstracted because the article modal will have a different container class for the post page.

packages/shared/src/components/modals/SharePostModal.tsx
sshanzel: No, that's a good point. There are pros and cons in either of the implementations, this would probably boil down to what we prefer more in the long run.

I am fine with any of the two.

capJavert: I agree with points Chris made to make it more verbose.

On another note we might be able to just use `isLoading` here. From what I see we need to make sure post data is available, `isLoading` will only be `false` until initial fetch is done, after that it is going to be `true`. So I am curios if we even need to check `isFetching` and `isFetchingNextPage` in addition to `isLoading` in all of the usages?

packages/shared/src/hooks/usePostById.ts
capJavert, line 54: Missing `post?.post` in dependency array.

capJavert (changes requested):
One issue with useMemo, it was present even before the changes made in this PR so whichever path this takes https://github.com/dailydotdev/apps/pull/1732/files#r1163651914 we should make sure dependency array contains only what is referenced inside the callback.

packages/shared/src/components/modals/SharePostModal.tsx
sshanzel: >  So I am curios if we even need to check isFetching and isFetchingNextPage in addition to isLoading in all of the usages?

The `isLoading` is always `false` from the beginning until the initial fetch is done, and we need to show a loading state on the initial load. That's the reason why we need the `isFetching`. We will need `isFetchingNextPage` when the user is switching left and right, though we can probably omit this just haven't tried yet.

Logging `isLoading`:
image

Alright. Thank you both for your thoughts. Let's go with the verbose implementation.

packages/shared/src/hooks/usePostById.ts
sshanzel: I mean, we removed redundant dependencies in the past, not sure we would want this one now. But I'm okay with it. Now I'm confused about when we want redundant dependencies and when we do not.

packages/shared/src/components/modals/SharePostModal.tsx
capJavert: Thanks for explaining cases for why fetching is needed. I would advise to then call `isPostLoading` something different and also verbose so we know it does two things. This is more of a problem of react-query nomenclature I think.

capJavert: `isPostLoadingOrFetching` is fine with me, but not everyone like that kind of verbosity.

sshanzel: > isPostLoadingOrFetching is fine with me, but not everyone like that kind of verbosity.

We can do this. Better to be clear than miss it. Thanks!

packages/shared/src/hooks/usePostById.ts
capJavert: > Now I'm confused about when we want redundant dependencies and when we do not.

We simply need to put all variables that are referenced inside the memo callback inside dependency array. `exhaustive-deps` rule would help with this since it would automatically say which ones are missing.

sshanzel: Right, it would probably be best to enable the linting to have a general rule. Though it would be a tedious one so we will have to do it in one go for the existing implementations.

rebelchris: We have a task already for the linting, but only after this work is done :)

rebelchris (approved):
Looks great, just the one optional comment we could probably put PostLoadingSkeleton in one return, but optional.

capJavert (approved):

sshanzel:
> Looks great, just the one optional comment we could probably put PostLoadingSkeleton in one return, but optional.

I think we might still end up with this once when we introduce new types of post.

PR #1737, opened 2023-04-12
fix: remove only members

rebelchris:
- Ensure moderators can only remove members and not other moderators or admins

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1249 #done

### DIFF
--- a/packages/shared/src/components/squads/SquadMemberMenu.tsx
+++ b/packages/shared/src/components/squads/SquadMemberMenu.tsx
@@ -174,7 +174,9 @@ export default function SquadMemberMenu({
     });
 
     const canRemoveMember =
-      canUpdateRole || verifyPermission(squad, SourcePermissions.MemberRemove);
+      canUpdateRole ||
+      (verifyPermission(squad, SourcePermissions.MemberRemove) &&
+        member.role === SourceMemberRole.Member);
 
     if (canRemoveMember) {
       menu.push({


packages/shared/src/components/squads/SquadMemberMenu.tsx
```diff
@@ -174,7 +174,9 @@ export default function SquadMemberMenu({
     });
 
     const canRemoveMember =
-      canUpdateRole || verifyPermission(squad, SourcePermissions.MemberRemove);
+      canUpdateRole ||
+      (verifyPermission(squad, SourcePermissions.MemberRemove) &&
+        member.role === SourceMemberRole.Member);
```
capJavert, line 179: Just a note, from what I can see this is covered on the API with `requireGreaterAccessPrivilege` which kinda spoils our clean usage of `verifyPermission`. As soon as we are actually checking roles and not permissions it could lead to API vs UI permission inconsistencies. 

I think right think here would be to have two separate permissions for our current implementation but I do know we are deep in v4 so might wanna leave it for later.

capJavert (approved):
Just a general concern, sorry for raising here, seen it before but I think this requirement shows why it could be a problem.

packages/shared/src/components/squads/SquadMemberMenu.tsx
rebelchris: Hmm yeah it's difficult as that would need to evaluate PER member and become very expensive to do API side as well.
Think for now this would be fine as it's the exact requirement. we can always optimize down the line.

Also tagging @sshanzel for his opinion as I'm open to adding a fn on FE side as well.

sshanzel: The only thing I can think of is the `roleRank`. If the user has a greater numerical value, we show the option.

rebelchris: The specs say you can only do members though, so not sure that would be valid.

sshanzel: But as a moderator, you have greater access over members only, right? So the condition won't have equal. For owners though, this probably should always be shown.

rebelchris: Ah yes, but then it's less explicit when we add new roles right?
Thus this being the more safe evaluation for now.

But still happy to promote to a function if that deems needed.

capJavert: I think core issue is from the API side so for frontend implementation I would leave it as it is.

sshanzel: If we introduce a new role and if you have greater access privileges, you should have power over it, right? Though would probably depend on the product's decision. So having it explicit now should work with the requirements. We can update if necessary.

rebelchris: Depends on product indeed, for now it's not set but could cause to break if product decided not to have that option :)

PR #1738, opened 2023-04-12
feat: align current member role to other roles in list

capJavert:

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1242 #done

### DIFF
--- a/packages/shared/src/components/squads/SquadMemberItemAdditionalContent.tsx
+++ b/packages/shared/src/components/squads/SquadMemberItemAdditionalContent.tsx
@@ -74,12 +74,6 @@ function SquadMemberItemAdditionalContent({
       />
     </SimpleTooltip>
   );
-  const emptyOption = (
-    <Button
-      buttonSize={ButtonSize.Small}
-      className="invisible m-auto mr-2 btn-tertiary"
-    />
-  );
 
   const isLoggedUser = loggedUser.id === user.id;
 
@@ -90,7 +84,7 @@ function SquadMemberItemAdditionalContent({
           className={!isLoggedUser && 'my-auto mr-2'}
           role={member.role}
         />
-        {isLoggedUser ? emptyOption : option}
+        {isLoggedUser ? null : option}
       </>
     );
   }
--- a/packages/shared/src/components/squads/SquadMemberItemAdditionalContent.tsx
+++ b/packages/shared/src/components/squads/SquadMemberItemAdditionalContent.tsx
@@ -74,6 +74,12 @@ function SquadMemberItemAdditionalContent({
       />
     </SimpleTooltip>
   );
+  const emptyOption = (
+    <Button
+      buttonSize={ButtonSize.Small}
+      className="invisible m-auto mr-2 btn-tertiary"
+    />
+  );
 
   const isLoggedUser = loggedUser.id === user.id;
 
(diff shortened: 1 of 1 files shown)


packages/shared/src/components/squads/SquadMemberItemAdditionalContent.tsx
```diff
@@ -74,6 +74,12 @@ function SquadMemberItemAdditionalContent({
       />
     </SimpleTooltip>
   );
+  const emptyOption = (
```
rebelchris, line 77: Ah my initial thoughts where to simply give MR-{amount} to the element when it was loggedUser?
Personally not a big fan of rendering things for padding purposes 🙈

rebelchris (changes requested):
Alternative idea to simply put margin right on the element?

packages/shared/src/components/squads/SquadMemberItemAdditionalContent.tsx
capJavert: Was just writing comment about that but you were too fast to review. I think this implementation with Button hidden is less error prone vs margin because if button styles change we get misalignment. WDYT?

capJavert: No strong feelings about it though, UIs by themself are error prone in the nature 😆

sshanzel: I think the number of items on the right side will be constant as we can keep everything under the options menu. So margin would probably be an alright solution for this case.

sshanzel: UIs are meant to be broken 😂

rebelchris: Ah I see where you coming from.
Yeah this type of UI rather gets broken I think.

The issue with this is it might become unclear for people why it's there and cause confusion down the line.
But something to be said for both 😂

packages/shared/src/components/squads/SquadMemberItemAdditionalContent.tsx
```diff
@@ -81,7 +81,7 @@ function SquadMemberItemAdditionalContent({
     return (
       <>
         <SquadMemberBadge
-          className={!isLoggedUser && 'my-auto mr-2'}
+          className={isLoggedUser ? 'mr-10' : 'mr-2'}
```
rebelchris, line 84: Do we not still need the my-auto for vertical alignment?

rebelchris (changes requested):
Just small check, not sure it's needed.

packages/shared/src/components/squads/SquadMemberItemAdditionalContent.tsx
capJavert: From what I see it does nothing.

rebelchris: Ah it has flex on parent indeed. :)

rebelchris (approved):
Let's go

PR #1739, opened 2023-04-12
fix: refactor owner to admin

rebelchris:
- Change all owner texts to admin

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1236 #done

### DIFF
--- a/packages/shared/src/graphql/sources.ts
+++ b/packages/shared/src/graphql/sources.ts
@@ -5,7 +5,7 @@ import { Connection } from './common';
 export enum SourceMemberRole {
   Member = 'member',
   Moderator = 'moderator',
-  Owner = 'owner',
+  Admin = 'admin',
   Blocked = 'blocked',
 }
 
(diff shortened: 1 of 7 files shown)


packages/shared/src/graphql/sources.ts
```diff
@@ -42,7 +42,7 @@ export interface Squad extends Source {
   permalink: string;
   public: boolean;
   type: SourceType.Squad;
-  owners?: string[];
+  admins?: string[];
```
capJavert, line 45: I think this is not used, `owners` and now `admins` does not exist as column or in `query Source` on API.

capJavert (changes requested):
Lots of renaming, one thing caught my eye, other then that 👏

packages/shared/src/graphql/sources.ts
rebelchris: Ah good point removed it now.

capJavert (approved):
🚀

PR #1742, opened 2023-04-13
feat: add global listener to history change to close lazy modal

capJavert:
- Added global listener per comments here https://dailydotdev.slack.com/archives/C02E2C3C13R/p1681299753422359
- Added it to _app due to router not being available for extension and we do not really have any route changes that happen in extensions (anything that changes route goes to app.daily.dev)
  - `popstate` event does not really work since it is not triggered for route changes
- My concern with this is that we are adding this for all lazy modals, we might want specific behaviour of not closing for specific modals
  - I would alternatively implement calling `closeModal` inside event handlers eg. when clicking item `UserListModal`

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done

### DIFF
--- a/packages/shared/src/hooks/useLazyModal.ts
+++ b/packages/shared/src/hooks/useLazyModal.ts
@@ -1,4 +1,4 @@
-import { useMemo } from 'react';
+import { useCallback, useMemo } from 'react';
 import { useQuery, useQueryClient } from 'react-query';
 import { LazyModalType, ModalsType } from '../components/modals/common';
 
@@ -18,8 +18,14 @@ export function useLazyModal<
   const { data: modal } = useQuery<T>(MODAL_KEY, () =>
     client.getQueryData<T>(MODAL_KEY),
   );
-  const openModal = (data: T) => client.setQueryData(MODAL_KEY, data);
-  const closeModal = () => client.setQueryData(MODAL_KEY, null);
+  const openModal = useCallback(
+    (data: T) => client.setQueryData(MODAL_KEY, data),
+    [client],
+  );
+  const closeModal = useCallback(
+    () => client.setQueryData(MODAL_KEY, null),
+    [client],
+  );
 
   return useMemo(
     () => ({
@@ -27,6 +33,6 @@ export function useLazyModal<
       closeModal,
       modal,
     }),
-    [modal],
+    [openModal, closeModal, modal],
   );
 }
--- a/packages/webapp/pages/_app.tsx
+++ b/packages/webapp/pages/_app.tsx
@@ -80,11 +80,27 @@ function InternalApp({ Component, pageProps, router }: AppProps): ReactElement {
 
   useTrackPageView();
   useInAppNotification();
-  useLazyModal();
+  const { modal, closeModal } = useLazyModal();
   usePrompt();
   useEffect(() => {
     updateCookieBanner(user);
   }, [user]);
+
+  useEffect(() => {
+    if (!modal) {
+      return undefined;
+    }
+
+    const onBeforeHistoryChange = () => {
+      closeModal();
+    };
+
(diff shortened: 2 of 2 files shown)


sshanzel:
We can only open modals on top of one another when one of them is not using lazymodal. This only happens LazyModal since it is rendered globally so it doesn't close whenever we switch page where the initialization of the implementation is gone. I think this is the way to go?

sshanzel:
When we switch page, it is definitely an intention to close everything though. So I see no issue why we should not close all modals when leaving.

capJavert:
> When we switch page, it is definitely an intention to close everything though. So I see no issue why we should not close all modals when leaving.

Agree on this, but not 100% if any edge cases. I can maybe take a look at all of the usages and write them up back here so we can evaluate? @rebelchris @sshanzel

sshanzel:
One example is once we migrated the `PostModal` to LazyModal, when we click the commenter/author/source icon, the modal should be closed where we will have to manually do it again.

sshanzel:
Probably in the future, we can add a property to pass to the lazy modal where it will ignore closing on navigation if it would be necessary.

rebelchris:
Some edge cases with the current implementation if becomes app specific.
So for instance now it won't work on extension and companion as you only implemented it in the webapp.

Is there anyway we can uniform it into the lazy modal in the `useLazyModal` to load the router there?
Not sure it's fully supported cross apps, but it makes it easier to maintain in my opinion.

capJavert:
Lazt modal usages:
| Component/Page | Modals | Page navigation |
|--------|--------|--------|
| pages/squads/[handle]/index.tsx | PostToSquad | No |
| src/components/ShareOptionsMenu.tsx | NewSquad, PostToSquad | No |
| src/components/squads/SquadHeaderMenu.tsx | EditSquad, SquadTour | No |
| src/components/squads/SquadMemberShortList.tsx | SquadMember | Yes, the user profile links |
| src/components/widgets/SocialShare.tsx | PostToSquad | No |
| src/hooks/useCreateSquadModal.ts | NewSquad | No |
| src/hooks/useSquadOnboarding.ts | NewSquad | No |
| src/hooks/useUpvoteQuery.ts | UpvotedPopup | Yes, the user profile links |

From the above the only case (currently) where we have clickable links are on the upvotes and squad member list, which use the same `UserListModal`. Granted, new cases could arise as Lee mentioned for the `PostModal`.

Because of that I am up for doing minimal change and refactoring to pass `onClick` to `UserListModal`. 

> Is there anyway we can uniform it into the lazy modal in the useLazyModal to load the router there?
Not sure it's fully supported cross apps, but it makes it easier to maintain in my opinion.

I Initially tried doing it there, but with many instances of `useLazyModal` hook it registers multiple listeners and also we can not use `next/router` in extension context and there are places where lazy modal is used inside of it

sshanzel:
> So for instance now it won't work on extension and companion as you only implemented it in the webapp.

But we don't need it in the extension and companion, right? When we navigate from those apps towards some page it will to the webapp. The extension only has feed pages where you can only navigate when everything is closed. Or I could have missed something.

rebelchris:
@sshanzel Guess your right indeed.
There is the modal for onboarding that can redirect to a feed, not sure it uses lazy now.

Just mentioning it for reference.
I don't have the direct implications at hand and it could cause some disturbances later on.

For now let's just deal with it like this and we can always optimise later when need be?

capJavert:
@sshanzel @rebelchris good to merge?

rebelchris (approved):
Let's do it

sshanzel (approved):
LGTM! 🚀

PR #1744, opened 2023-04-13
feat: support getting preview and loading state

sshanzel:
## Changes

**Blocked until we get an API to return a link's title and preview**

Though blocked, this should be ready for review, the only missing piece is the integration of fetching title and image.

The form should now accept a private link object which consists of title, image, and url.
Send the said parameters to the submitExternalLink mutation.

Preview:

image

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1251 #done


packages/shared/src/components/squads/Comment.tsx
```diff
@@ -70,21 +82,22 @@ export function SquadComment({
...
         ) {
-          onUpdateForm(null);
+          onUpdateForm({ post: { post: null } });
+          getPrivateLink({ url, sourceId });
         }
       },
     },
   );
-  const [checkUrl] = useDebounce(getPost, 200);
+  const [checkUrl] = useDebounce(getPost, 1000);
```
sshanzel, line 100: The request is rather expensive I believe. Do you think we should keep it at 200ms or 1000ms?

capJavert: Good point, I think test it out on preview to see how it feels and go from there.

sshanzel: Good idea. Let's do that once merged.

packages/shared/src/components/squads/Comment.tsx
```diff
@@ -159,24 +191,9 @@ export function SquadComment({
               onInput={onInputChange}
               hint={linkHint}
               saveHintSpace
+              disabled={isCheckingUrl}
```
rebelchris, line 194: what if it fails to load?

packages/shared/src/components/squads/Comment.tsx
```diff
@@ -70,21 +82,22 @@ export function SquadComment({
...
         }
       },
-      onError: (err: ApiErrorResult) => {
+      onError: (err: ApiErrorResult, url) => {
         if (
           err?.response?.errors?.[0].extensions.code === ApiError.NotFound &&
           !isNullOrUndefined(postItem)
         ) {
-          onUpdateForm(null);
+          onUpdateForm({ post: { post: null } });
```
rebelchris, line 94: Should we not show an error if user hits rate-limit or the url is not an actual URL or something like that?

packages/shared/src/graphql/posts.ts
```diff
@@ -444,8 +448,38 @@ export const SUBMIT_EXTERNAL_LINK_MUTATION = gql`
...
+  url: string;
+  sourceId: string;
+}
+
+export const getExternalLinkPreview = async (
+  params: ExternalLinkPreviewProps,
+): Promise<ExternalLinkPreview> => {
+  const res = await request(graphqlUrl, PREVIEW_LINK_MUTATION, params);
+
+  return res.checkLinkPreview;
```
rebelchris, line 479: What happens if this doesn't exist?

rebelchris (changes requested):
Mainly have some questions around the errors that could happen 🙏

packages/shared/src/components/squads/Comment.tsx
sshanzel: This won't be disabled then. The `isLoading` is handled properly inside react-query whether the request succeeded or not.

packages/shared/src/components/squads/Comment.tsx
sshanzel: Yes, that is not yet implemented as we're still waiting for the API PR.

sshanzel: Since we know what to expect, whenever it fails we just remove the preview.

packages/shared/src/graphql/posts.ts
sshanzel: Should throw an error and the process would continue without preview.

packages/shared/src/components/squads/Comment.tsx
```diff
@@ -54,13 +60,20 @@ export function SquadComment({
...
-  const [link, setLink] = useState(url);
-  const [linkHint, setLinkHint] = useState(url);
+  const [link, setLink] = useState(privateLink?.url);
+  const [linkHint, setLinkHint] = useState(privateLink?.url);
+  const { mutateAsync: getPrivateLink } = useMutation(getExternalLinkPreview, {
+    onSuccess: (preview, url) => {
+      onUpdateForm({ privateLink: { url, ...preview } });
+    },
+    onError: (_, url) => {
+      onUpdateForm({ privateLink: { url } });
```
sshanzel, line 74: Should we show a toast notification for when the limit has been reached?

idoshamun: I think so. Make sure to validate that it's indeed this error, if it's another one ask the user to try again

sshanzel:
Verified the rate limiting and toast works as intended 🥳 

image

packages/shared/src/components/squads/Comment.tsx
capJavert: Current feels good to me 👍

packages/shared/src/components/modals/PostToSquadModal.tsx
```diff
@@ -51,19 +46,22 @@ function PostToSquadModal({
...
   });
 
   const onPostSuccess = async (squadPost?: Post) => {
+    const syncSubmission = !!form?.privateLink?.title || !!form?.post?.post;
+
     if (squadPost) onSharedSuccessfully?.(squadPost);
 
     displayToast(
-      isLink
+      !syncSubmission
```
rebelchris, line 58: In what case can it be non sync submission?
Pretty sure we block anything that's not sync now right.

packages/shared/src/components/modals/PostToSquadModal.tsx
```diff
@@ -79,13 +77,20 @@ function PostToSquadModal({
...
 
-  const onSubmit: SubmitSharePostFunc = async (e, commentary, postLink) => {
+  const onSubmit: SubmitSharePostFunc = async (e, commentary) => {
     e?.preventDefault();
 
     if (isLoading) return null;
 
-    if (isLink) {
-      return onSubmitLink({ url: postLink, sourceId: squad.id, commentary });
+    if (form?.privateLink) {
```
rebelchris, line 85: Got a bit confused as to why this privateLink has multiple fields, but maybe it's just the naming to me personally.

capJavert (approved):
Minor comments, I the end implementation looks simple and works great 👌

packages/shared/src/components/post/PostPreview.tsx
```diff
@@ -0,0 +1,72 @@
...
+        <LoadingPlaceholder className={imageClassName} />
+      ) : (
+        <Image
+          src={image}
+          className={classNames('object-cover', imageClassName)}
+          loading="lazy"
+          fallbackSrc={cloudinary.post.imageCoverPlaceholder}
+        />
+      )}
+      <a href={url} target="_blank">
```
rebelchris, line 62: Seeing this is _blank, perhaps we need to consider adding rel options to avoid exploitation.

packages/shared/src/graphql/common.ts
```diff
@@ -82,14 +82,22 @@ export const isQueryKeySame = (left: QueryKey, right: QueryKey): boolean => {
 export enum ApiError {
   Forbidden = 'FORBIDDEN',
   NotFound = 'NOT_FOUND',
+  RateLimited = 'RATE_LIMITED',
 }
 
+export const hasApiError = (
```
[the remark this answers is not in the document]
sshanzel: Right. I changed it on the later part 😂

packages/shared/src/components/squads/Comment.tsx
```diff
@@ -53,38 +59,56 @@ export function SquadComment({
...
+    getExternalLinkPreview,
+    {
+      onSuccess: (preview, url) => {
+        onUpdateForm({ privateLink: { url, ...preview } });
+      },
+      onError: (err: ApiErrorResult, url) => {
+        const rateLimited = hasApiError(err, ApiError.RateLimited);
+        if (rateLimited) {
+          displayToast(rateLimited.message);
+        }
```
rebelchris, line 80: What about any other errors? (Generic URL does not exist etc)

packages/shared/src/components/squads/Comment.tsx
```diff
@@ -203,7 +231,7 @@ export function SquadComment({
               className="btn-primary-cabbage"
               type="submit"
               loading={isLoading}
-              disabled={!commentary || isLoading || isCheckingUrl}
+              disabled={!commentary || isLoading || isPreviewLoading}
```
rebelchris, line 234: I'm also pretty sure we block it if preview would fail for whatever reason. not sure this captures that (just want to make sure)

packages/shared/src/components/squads/SharePostBar.tsx
```diff
@@ -30,23 +31,42 @@ function SharePostBar({
...
-    onError: (err: ApiErrorResult, link) => {
-      if (
-        link === '' ||
-        allowedSubmissionErrors.includes(
-          err?.response?.errors?.[0].extensions.code,
-        )
-      ) {
-        onNewSquadPost({ url, onSharedSuccessfully });
-      }
+  const { mutateAsync: getPrivateLink, isLoading: isLoadingPreview } =
```
rebelchris, line 37: This code is now duplicate right?
Consider moving it to a hook maybe for uniform behaviour.

packages/shared/src/components/squads/SharePostBar.tsx
```diff
@@ -30,23 +31,42 @@ function SharePostBar({
...
+      onError: (err: ApiErrorResult, link) => {
+        const rateLimited = hasApiError(err, ApiError.RateLimited);
+        if (rateLimited) {
+          displayToast(rateLimited.message);
+        }
+
+        onNewSquadPost({ privateLink: { url: link }, onSharedSuccessfully });
+      },
+    });
+  const { mutateAsync: getPost, isLoading: isCheckingUrl } = useMutation(
```
rebelchris, line 52: This is actually also re-used, maybe can go in the same hook or the existing squad hook?

rebelchris (changes requested):
Small feedback and optimisations I see.
Overall looks solid

packages/shared/src/components/modals/PostToSquadModal.tsx
sshanzel: If the title is not found from the API where we fetch the preview.

packages/shared/src/components/squads/Comment.tsx
sshanzel: Oh, are not gonna allow anymore if there is no title and image from the preview API?

rebelchris: I believe so, if this one can't return the actual scraper won't be able to.
But maybe triple check with Sab on this.

packages/shared/src/components/squads/SharePostBar.tsx
sshanzel: I have thought of doing it but refrained while on it as it felt not that much difficult, but now that you mentioned it, let's see how it will look like.

sshanzel:
@rebelchris would love to hear what you think 🚢

packages/shared/src/components/squads/Comment.tsx
```diff
@@ -203,7 +207,12 @@ export function SquadComment({
               className="btn-primary-cabbage"
               type="submit"
               loading={isLoading}
-              disabled={!commentary || isLoading || isCheckingUrl}
+              disabled={
+                !commentary ||
+                isLoading ||
+                isPreviewLoading ||
+                (!form?.post?.post && !form?.externalLink?.title)
+              }
```
rebelchris, line 215: Nitpicking, but maybe move this to a const above just for clarity

rebelchris (changes requested):
Wow, looks a lot cleaner.
Minor question to something not super clear from the code point.

And would also love to get @idoshamun quick review on this if possible 🙏

idoshamun:
@rebelchris looking now

packages/shared/src/components/squads/SharePostBar.tsx
```diff
@@ -1,52 +1,48 @@
...
   onNewSquadPost,
   disabled = false,
 }: SharePostBarProps): ReactElement {
   const [url, setUrl] = useState('');
   const onSharedSuccessfully = () => setUrl('');
-  const { mutateAsync: getPost } = useMutation(getPostByUrl, {
-    onSuccess: (post) => {
-      onNewSquadPost({ post, onSharedSuccessfully });
+  const { getPost, isLoadingPreview, isCheckingPost } = usePostToSquad({
+    onEmptyUrl: (link) => onNewSquadPost({ externalLink: { url: link } }),
```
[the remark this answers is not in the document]
sshanzel: When you want to start posting and you don't have any URL.

packages/shared/src/components/squads/Comment.tsx
idoshamun: Agree

packages/shared/src/components/squads/SharePostBar.tsx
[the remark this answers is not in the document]
idoshamun: I think it's when clicking Post without adding a link, but more like a guess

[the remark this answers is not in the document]
sshanzel: Yes. That is when clicking the button "Post" without a link.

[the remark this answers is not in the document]
rebelchris: But then where does this `link` come from? 😅
Sorry didn't really look in depth but doesn't feel super clear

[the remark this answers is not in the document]
sshanzel: That's just the parameter passed, but probably redundant keeping the usage of the parameter itself.

sshanzel:
**Note:** would only be able to test once the API is merged :shipit: 

Since our main goal here is to display the preview, not necessarily to interact with a post when it is found in our DB (by utilizing the post properties), I simplified the process by unifying the requirements of just the title and image. Then if the preview has `id`, it means it can be found in our DB.

@capJavert I have made quite a few changes from the last you saw this PR, feel free to do another round if you have the time. Thank you!

capJavert (approved):
@sshanzel looks good to me code wise, I would like to test again once API is merged.

rebelchris (approved):
No further comments from my side.

capJavert:
After testing I think this is a go 🚀 gg @sshanzel

sshanzel:
> After testing I think this is a go 🚀 gg @sshanzel

Thank you for all the support! I will just test it once again based on Nimrod's feedback (once the refactor in the API is merged) then we will ship this 🥳

sshanzel:
@rebelchris let me know if we can merge now 🚢

PR #1745, opened 2023-04-13
feat: add react hooks eslint plugin

capJavert:
## Changes

**See for latest decision https://dailydotdev.atlassian.net/wiki/spaces/HAN/pages/512491748/WT-1235+ESlint+Rules+of+Hooks**

Shared lint result:
`172 problems (21 errors, 151 warnings)` - bunch of issues 

Extension lint result (most of the issues):
`23 problems (11 errors, 12 warnings)`

At the end it is important to note that with these kind of changes (especially `useEffect` fixes) we need to be careful not to introduce bugs since some of this issues contain anti patterns that went below the radar during development of our applications up until this point.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1235 #done

### DIFF
--- a/packages/eslint-config/index.js
+++ b/packages/eslint-config/index.js
@@ -8,8 +8,9 @@ module.exports = {
     'plugin:@typescript-eslint/recommended',
     'prettier',
     'plugin:prettier/recommended',
+    'plugin:react-hooks/recommended',
   ],
-  plugins: ['prettier', 'tailwindcss', '@dailydotdev/daily-dev-eslint-rules'],
+  plugins: ['prettier', 'tailwindcss', '@dailydotdev/daily-dev-eslint-rules', 'react-hooks'],
   parserOptions: {
     project: './tsconfig.eslint.json',
     ecmaVersion: 2020,
--- a/packages/extension/src/companion/App.tsx
+++ b/packages/extension/src/companion/App.tsx
@@ -63,16 +63,24 @@ export default function App({
     return <></>;
   }
 
+  // @NOTE see https://dailydotdev.atlassian.net/l/cp/dK9h1zoM
+  // eslint-disable-next-line react-hooks/rules-of-hooks
   const memoizedFlags = useMemo(() => flags, [flags]);
   const refetchData = async () =>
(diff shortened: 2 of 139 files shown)


capJavert:
@rebelchris there are still a lot of linter issues, `shared` has most of course since the most codebase is there. I do agree that this should be done in one go but from my points in description I think if we blindly follow lint warnings we will not get any better codebase and maybe even have worst rendering performance.

I can do this incrementally day by day (for 30minutes in parallel with squads v5 tasks) and clear up the issues as I did for web app, but from review point of view I think it would be better to look at it in chunks. If you look at the commits already in the PR there are different cases and types of fixes to solve issues with react-hooks. Some issues are not really solvable without changing implementation like this one https://github.com/dailydotdev/apps/pull/1745/commits/f0177886ba984fb7ca4a9fe228a0c7ef0d644332

Curious if you are fine with this ending up being large PR or would favor incremental adoption (which we would have to find the best way to do it). FYI @sshanzel 

Alternative is that we could ignore `useEffect` errors for now (since those are most complex but might actually have some "side effects" pun intended in them that actually cause bugs we don't see), and strive to only solve memoization (useMemo and useCallback) which are easy fixes 99% of the time.

rebelchris:
@capJavert I see I actually never commented on here.
But my suggestion try to introduce this as just a info state ([if possible](https://eslint.org/docs/latest/use/configure/rules) warn maybe?)
and we can do small PR's over time fixing the issues.

Until we get to a point where there are none, where we'll set it to strict.

capJavert:
@rebelchris @sshanzel from team meeting last week I took some time yesterday to add this, would appreciate if you have any more comments, most of the stuff is now explained in [the DR](https://dailydotdev.atlassian.net/wiki/spaces/HAN/pages/512491748/WT-1235+ESlint+Rules+of+Hooks) which is also linked in description.

rebelchris (approved):
wow nice find :D

idoshamun (approved):

PR #1755, opened 2023-04-19
feat: hide invite option based on squad permission

capJavert:
- Hide invitation button from squad header
- Hide invitation CTAa from members modal
- Hide invitation CTA from modals (unused but just to be sure)
- API implementation https://github.com/dailydotdev/daily-api/pull/1240

Also implements WT-1267

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1266 #done


packages/shared/src/components/modals/SquadInviteModal.tsx
```diff
@@ -50,15 +50,17 @@ function SquadInviteModal({
...
-      <Modal.Footer>
-        <Button
-          icon={<LinkIcon />}
-          className="flex-1 mx-4 btn-primary-cabbage"
-          onClick={() => inviteTextRef.current?.copyLink()}
-        >
-          Copy invitation link
-        </Button>
-      </Modal.Footer>
+      {verifyPermission(squad, SourcePermissions.Invite) && (
```
rebelchris, line 53: Wait, but we shouldn't even show this modal if you don't have the permission right?
Not sure this step is needed here.

packages/shared/src/components/modals/SquadMemberModal.tsx
```diff
@@ -36,6 +36,10 @@ const InitialItem = ({ squad }: { squad: Squad }) => {
     origin: Origin.SquadMembersList,
   });
 
+  if (!verifyPermission(squad, SourcePermissions.Invite)) {
+    return null;
+  }
```
rebelchris, line 41: should this not go above the hook, just so we don't load it unneeded?
Or is that a anti-pattern?

rebelchris (changes requested):
Nice, got some minor cleanup things I think we can do.

packages/shared/src/components/modals/SquadMemberModal.tsx
capJavert: Yes

packages/shared/src/components/modals/SquadInviteModal.tsx
capJavert: @rebelchris It is actually unused so we can remove component altogether?

packages/shared/src/components/modals/SquadMemberModal.tsx
rebelchris: Ah yes, we don't have locked squad anymore, so please do a cleanup yes 🙏

packages/shared/src/components/modals/SquadInviteModal.tsx
rebelchris: Ah yes, we don't have locked squad anymore, so please do a cleanup yes 🙏

rebelchris (approved):
Nice work! 🙏

PR #1761, opened 2023-04-21
feat: handle error when joining squad

capJavert:
- API changes here https://github.com/dailydotdev/daily-api/pull/1245
- Decided on showing toast [here](https://dailydotdev.slack.com/archives/C0506KU4RH9/p1682067432660349)

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1277 #done


packages/webapp/pages/squads/[handle]/[token].tsx
```diff
@@ -137,6 +141,15 @@ const SquadReferral = ({
         addSquad(data);
         return router.replace(data.permalink);
       },
+      onError: (error: ApiErrorResult) => {
+        const errorMessage = error?.response?.errors?.[0]?.message;
+
+        if (errorMessage === ApiErrorMessage.SOURCE_INVITE_NOT_VALID) {
```
Comment (capJavert) [inline packages/webapp/pages/squads/[handle]/[token].tsx:147]: @rebelchris @sshanzel interested what you think about this, I am returning this specific error message from API. I though about checking the `extension.code` but since `FORBIDDEN` is used for other stuff it would not work here since we want to do specific message for this case where invite is no longer valid. My concern is that if we for some reason message is changed on the API we need to update enum here too. 

You can also check API implementation linked in this PR, I thought as alternative to maybe create specific error code eg. `PERMISSION_DENIED` on API but not really know on how far we want to expand the error codes. If introducing new code we should probably add that one in other places we check for permissions on API.

Comment (rebelchris) [reply packages/webapp/pages/squads/[handle]/[token].tsx:147]: My suggestion would be to add a type to `errors.ts` on the API.
Something under UserFail maybe? `SourceInviteForbidden`.

Which would always throw this enum even.
FE can then focus on showing human readable version of it?

Comment (capJavert) [reply packages/webapp/pages/squads/[handle]/[token].tsx:147]: Hmm `UserFailErrorKeys` is `type`. I would rather add something like `SubmissionFailErrorMessage` maybe `PermissionErrorMessage` record or enum?

Then I could use it as error message value.

> My concern is that if for some reason message is changed on the API we need to update enum here too.

I would still had to match on the frontend eg. add those errors to `ApiErrorMessage` enum but I guess that will always be the case since we have two separate repos.

Comment (rebelchris) [reply packages/webapp/pages/squads/[handle]/[token].tsx:147]: If you make it an enum there should really be no need to change it.
And if we do it should trigger someone to look at why it's an enum.

You can go ahead and add something new though, more a suggestion of that scope example.

Comment (capJavert) [reply packages/webapp/pages/squads/[handle]/[token].tsx:147]: Oh ok, didn't catch that you are ok with refactoring that to enum. That makes sense.

rebelchris (approved):
Nice, super neat 🙏

PR #1749, opened 2023-04-15
refactor: removal of unnecessary use effects for hiding scrollbar

sshanzel:
## Changes
- One reason why we are losing the scroll position is that we hide it from the user when we open the article modal.
- I noticed the Modal overlay was also not taking up the whole screen's width so it kinda shifts the whole UI when the background has its own scroll.
- Also fixed the overscroll behavior that should have solved the issue of stopping the scroll on the body when the modal is opened than having the previous hack of `overflow: hidden` which we removed here.

### Describe what this PR does
- Short and concise, bullet points can help
- Screenshots if applicable can also help

## Events

Did you introduce any new tracking events?
Don't forget to update the [Analytics Taxonomy sheet](https://docs.google.com/spreadsheets/d/18Lv7zXges9QfVX5VYL1a-Hyl0e1sQ3sLr0OK8YZWKXI/edit#gid=0)

Log the new/changed events below:

| Type   | event_name  | value |
|--------|-------------|-------|
| Change/New | event name  | extra: { ... } |

### On those affected packages:
- [ ] Have you done sanity checks in the webapp?
- [ ] Have you done sanity checks in the extension?
- [ ] Does this not break anything in companion?

### Did you test the modified components media queries?
- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

#### Did you test on actual mobile devices?
- [ ] iOS (Chrome and Safari)
- [ ] Android

I believe the root cause of both these issues is the same. Hence, we will be able to close it once done.
WT-310 #done
WT-812 #done

PR #1763, opened 2023-04-21
feat: remember scroll position and restore after post modal close

capJavert:
- I could not really get anything concrete on this so wanted to propose a different approach
  - I also checked and this does not happen on Android phones I have on me Android 12 and 13
  - As said in tasks and reports seems like iOS only issue when clicking "X" close button
  - Also, if user just uses back button (instead of "X") scroll restoration works and they do not get jumped to top 
- To fix this I implemented a ref to save scroll position on feed when post modal gets opened, I keep it and when modal is closed by clicking on "X" I just restore that position and that works
  - also it does not break desktop/other platforms since they just retain scroll normally
  
I know this does not fix the core issue but if iOS browser was conforming to the spec this would not even be the problem so I think if this works and brings a lot of value to our users it should be enough.
  
It can be tested on [preview2](https://preview2.app.daily.dev)

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-812 #done

### DIFF
--- a/packages/shared/src/hooks/usePostModalNavigation.ts
+++ b/packages/shared/src/hooks/usePostModalNavigation.ts
@@ -1,5 +1,5 @@
 import { useRouter } from 'next/router';
-import { useContext, useEffect, useMemo, useState } from 'react';
+import { useContext, useEffect, useMemo, useRef, useState } from 'react';
 import AnalyticsContext from '../contexts/AnalyticsContext';
 import { Post, PostType } from '../graphql/posts';
 import { postAnalyticsEvent } from '../lib/feed';
@@ -37,6 +37,7 @@ export const usePostModalNavigation = (
   const [isFetchingNextPage, setIsFetchingNextPage] = useState(false);
   const { trackEvent } = useContext(AnalyticsContext);
   const router = useRouter();
+  const scrollPositionOnFeed = useRef(0);
 
   const changeHistory = (data: unknown, title: string, url: string) => {
     if (!isExtension) {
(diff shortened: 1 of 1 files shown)


capJavert:
Related to https://github.com/dailydotdev/apps/pull/1749

rebelchris:
@capJavert The only issue is still other modals though.
So the article modal now works, but for instance opening the customize window and closing that will always reset scroll position. (Maybe it's a edge-case we can live with, but logging it here)

capJavert:
> What is the performance impact of this proposal?

I don't think there is any performance impact. 

> So the article modal now works, but for instance opening the customize window and closing that will always reset scroll position. (Maybe it's a edge-case we can live with, but logging it here)

We can re-apply this logic to lazy modals I would think. We just need to hook into open/close events and add the ref to save the scroll position. This open post -> back to post is probably most used so it gives users the best gain.

idoshamun:
I can't test (no iOS device) but code wise it looks great.
Should we consider adding as Ante suggested to the modals hooks instead of this specific one?
This way we'll make sure it won't happen with any other modal

idoshamun (approved):

capJavert:
@rebelchris should I merge this then and add a backlog item to update/test this on lazy modal too? 

Don't wanna block this since we focus on onboarding and squad v5 this week.

rebelchris:
@capJavert Yes go ahead and merge this, can you take care of the backlog ticket for this for now?

PR #1769, opened 2023-04-25
feat(checklist): initial implementation of components

capJavert:

This PR implements:
- basic squad checklist with dynamic steps based on role
- adds current copy to steps from design
- creates step components for specific steps (missing logic and completion logic, wanted to split those up in sub tickets since this PR does enough already)
- integrates `useActions` hook to build steps for checklist

Implementation detail of checklist components:
- Create checklist components that will be used to display and assemble squad checklists needed for v5
- Idea is to have the base components that can be used and composed into specific implementation to satisfy different needs for checklist in scope of v5 and in the future
- `ChecklistCard` is the base component that accepts steps and creates a card
  - each step contains an action and additional props that are needed to render it (eg. additional component, classes etc.)
  - Checklist card can also composed into parent component to have its own specific logic on how to map actions to steps
  - see `SquadChecklistCard` - this one contains all the logic to map steps and actions per current member role
  - in the future when we have different checklist for some other parts of the app we can follow the same pattern
- `ChecklistStep` is a default component that is used to render steps inside the checklist
  - it can be composed into new components with additional views that are needed for some checklist step or some generic ones to be reused across different checklists (eg. install extension step)
  - see `InstallExtensionChecklistStep` and other `*ChecklistStep` components

You can see all of this in action on squad page.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1293 #done


packages/shared/src/components/checklist/ChecklistCard.tsx
```diff
@@ -0,0 +1,90 @@
...
+}: ChecklistCardProps): ReactElement => {
+  const activeStep = useMemo(
+    () => steps.find((item) => !item.action.dateCompleted)?.action.type,
+    [steps],
+  );
+  const [checkedStep, setCheckedStep] = useState<string>(activeStep);
+  const isDone = useMemo(
+    () => steps.every((item) => !!item.action.dateCompleted),
+    [steps],
+  );
```
rebelchris, line 24: This almost sounds like a perfect fit for a hook, especially since a lot of this will be controlled by external placed actions.
What do you think?

packages/shared/src/components/checklist/ChecklistCard.tsx
```diff
@@ -0,0 +1,90 @@
...
+        return undefined;
+      }
+
+      return action.type;
+    });
+  };
+
+  return (
+    <div className={className}>
+      <Card className="p-0 rounded-14 border-cabbage-40 hover:!border-cabbage-40 w-[340px]">
```
rebelchris, line 38: I don't think the color options here are good, please refrain from using colors directly as it doesn't include light/dark mode.
There should be a named version of this available. (Let me know if you can't find it)

packages/shared/src/components/checklist/ChecklistCard.tsx
```diff
@@ -0,0 +1,90 @@
...
+        <div className="p-4 rounded-t-14 bg-theme-bg-cabbage-opacity-24">
+          <p className="mb-1 font-bold typo-body text-theme-label-primary">
+            {title}
+          </p>
+          <p className="text-salt-90 typo-callout">{description}</p>
```
rebelchris, line 43: Also here the color can't be right.

packages/shared/src/components/checklist/ChecklistCard.tsx
```diff
@@ -0,0 +1,90 @@
...
+          {typeof onRequestClose === 'function' && (
+            <button
+              className="absolute top-4 right-4 text-salt-90"
```
rebelchris, line 46: Same note

packages/shared/src/components/checklist/ChecklistCard.tsx
```diff
@@ -0,0 +1,90 @@
...
+
+            return (
+              <StepComponent
+                key={step.action.type}
+                step={step}
+                checked={checkedStep === step.action.type}
+                active={activeStep === step.action.type}
+                onToggle={onToggleStep}
+                className={{
+                  checkmark: isDone && 'text-cabbage-40',
```
rebelchris, line 79: Here as well color issue

packages/shared/src/components/checklist/ChecklistStep.tsx
```diff
@@ -0,0 +1,94 @@
...
+          event.preventDefault();
+
+          onToggle(step.action);
+        }}
+      >
+        <div className="flex items-center flex-start">
+          <QuaternaryButton
+            className={classNames(
+              '-ml-2 text-salt-90',
+              isCompleted && 'text-pepper-10',
```
rebelchris, line 36: Color is not right here.

packages/shared/src/components/checklist/ChecklistStep.tsx
```diff
@@ -0,0 +1,94 @@
...
+            )}
+            id={step.action.type}
+            disabled={isCompleted}
+            icon={
+              <div
+                className={classNames(
+                  active && 'p-1 rounded-full bg-theme-bg-cabbage-opacity-24',
+                )}
+              >
+                <ChecklistAIcon
```
rebelchris, line 46: What is Checklist A? (not sure the name is a typo or not?)

packages/shared/src/components/checklist/ChecklistStep.tsx
```diff
@@ -0,0 +1,94 @@
...
+                  className={classNames(
+                    active && 'text-cabbage-40',
```
rebelchris, line 48: Color issue

packages/shared/src/components/checklist/ChecklistStep.tsx
```diff
@@ -0,0 +1,94 @@
...
+                  secondary={!isCompleted}
+                />
+              </div>
+            }
+          >
+            <p
+              className={classNames(
+                'typo-callout',
+                active ? 'font-bold text-theme-label-primary' : 'font-normal',
+                isCompleted && 'text-pepper-10',
```
rebelchris, line 61: Color

packages/shared/src/components/checklist/ChecklistStep.tsx
```diff
@@ -0,0 +1,94 @@
...
+                className.title,
+              )}
+            >
+              {step.title}
+            </p>
+          </QuaternaryButton>
+        </div>
+        <ArrowIcon
+          className={classNames(
+            !isOpen && 'rotate-180 text-salt-90',
```
rebelchris, line 71: Color

packages/shared/src/components/checklist/ChecklistStep.tsx
```diff
@@ -0,0 +1,94 @@
...
+            isCompleted && 'opacity-32',
+          )}
+          size={IconSize.Small}
+        />
+      </button>
+      {isOpen && (
+        <div className="my-2 ml-9">
+          <p
+            className={classNames(
+              'text-salt-90 typo-callout',
```
rebelchris, line 81: Color

packages/shared/src/components/checklist/NewSquadAndSiteMemberChecklistCard.tsx
```diff
@@ -0,0 +1,109 @@
...
+import { Squad } from '../../graphql/sources';
+import { ChecklistCard } from './ChecklistCard';
+import { InstallExtensionChecklistStep } from './InstallExtensionChecklistStep';
+import { ChecklistStep } from './ChecklistStep';
+import { Button } from '../buttons/Button';
+import { ChecklistStepProps } from '../../lib/checklist';
+import BellIcon from '../icons/Bell';
+import { FlexRow } from '../utilities';
+
+const NewSquadAndSiteMemberChecklistCard = ({
```
rebelchris, line 11: I'm not 100% sure this format is correct, in the latest spec we decided that steps are fully optional based on role + status.
So someone could be a multiple or roles rendering different checklist.
Did you take that into account?

packages/shared/src/styles/base.css
```diff
@@ -52,6 +52,7 @@ body {
   --theme-background-bun: theme('colors.bun.80');
   --theme-background-onion: theme('colors.onion.80');
   --theme-background-pepper: theme('colors.pepper.20');
+  --theme-background-cabbage-opacity-24: theme('colors.cabbage.50')24;
```
rebelchris, line 55: Are you sure we need to introduce a new color, would think it should already exist.

rebelchris (changes requested):
A generic note, the colors you use are not correct and technically the eslint should pick up on this. Might have to re-evaluate that.

Also noticed you choose per "state+role" checklist, in the last decision we choose a generic checklist, did you consider that case?

packages/shared/src/components/checklist/ChecklistCard.tsx
capJavert: Yes, as written in description, some of the logic should go to useActions hook, its here until we see how that one turns out.

packages/shared/src/components/checklist/NewSquadAndSiteMemberChecklistCard.tsx
capJavert: Yeah, steps creation is completely optional and can be what ever, this is just an example of one implementation.

packages/shared/src/components/checklist/ChecklistCard.tsx
capJavert: Yeah got it, It is clunky because in design colors are written like this and then finding those in theme is hell and they often mismatch 💦 . I'll try and remap to theme colors for sure and let you know.

image
image

packages/shared/src/components/checklist/NewSquadAndSiteMemberChecklistCard.tsx
capJavert: I'll refactor to show concrete example for the spec.

capJavert: @rebelchris see `SquadChecklistCard`

packages/shared/src/components/checklist/ChecklistCard.tsx
capJavert: @rebelchris so for this the right value should be `text-theme-color-salt`?

packages/shared/src/components/checklist/ChecklistCard.tsx
```diff
@@ -0,0 +1,91 @@
...
+        return undefined;
+      }
+
+      return action.type;
+    });
+  };
+
+  return (
+    <div className={className}>
+      <Card className="p-0 rounded-14 border-cabbage-40 hover:!border-cabbage-40 w-[340px]">
```
rebelchris, line 37: This line still have some wrong colors.

rebelchris: Also make sure you covered the mobile view I think that might be missing from this PR now.

packages/shared/src/components/checklist/ChecklistStep.tsx
```diff
@@ -0,0 +1,96 @@
...
+  checked = false,
+  active = false,
+  onToggle,
+  children,
+}: ChecklistStepProps): ReactElement => {
+  const isCompleted = !!step.action.completedAt;
+  const isOpen = !isCompleted && checked;
+
+  return (
+    <div className={className.root}>
```
rebelchris, line 21: Not urgent, but think normally we would call this container

packages/shared/src/components/checklist/ChecklistStep.tsx
```diff
@@ -0,0 +1,96 @@
...
+        disabled={isCompleted}
+        className="flex justify-between items-center w-full"
+        onClick={(event) => {
+          event.preventDefault();
+
+          onToggle(step.action);
+        }}
+      >
+        <div className="flex flex-1 gap-1 items-center flex-start">
+          <QuaternaryButton
```
rebelchris, line 33: Why do we need the button inside a button?
Can't the outside button carry all the styling directly?

packages/shared/src/components/checklist/InstallExtensionChecklistStep.tsx
```diff
@@ -0,0 +1,25 @@
...
+const InstallExtensionChecklistStep = (
+  props: ChecklistStepProps,
+): ReactElement => {
+  // TODO WT-1293-checklist-components add business logic
+
+  return (
+    <ChecklistStep {...props}>
+      <Button className="btn-primary">
+        <FlexCentered className="gap-2">
+          <BrowsersIcon className="text-theme-label-primary !w-12" /> Add to
```
rebelchris, line 17: This would actually be better as a plain SVG then, it seems weird to have a icon that we by default have to modify.

packages/shared/src/components/checklist/NotificationChecklistStep.tsx
```diff
@@ -0,0 +1,19 @@
+import React, { ReactElement } from 'react';
+import { ChecklistStepProps } from '../../lib/checklist';
+import { Button } from '../buttons/Button';
+import { ChecklistStep } from './ChecklistStep';
+import BellIcon from '../icons/Bell';
+
+const NotificationChecklistStep = (props: ChecklistStepProps): ReactElement => {
+  // TODO WT-1293-checklist-components add business logic
```
rebelchris, line 8: Just a question how can we maintain that this happens when we merge this to intermediate?
I want to avoid us forgetting about it.

packages/shared/src/components/icons/ChecklistA/index.tsx
```diff
@@ -0,0 +1,15 @@
+import React, { ReactElement } from 'react';
+import Icon, { IconProps } from '../../Icon';
+import FilledIcon from './filled.svg';
+import OutlinedIcon from './outlined.svg';
+
+const ChecklistAIcon = ({ size, ...rest }: IconProps): ReactElement => (
```
rebelchris, line 6: I'm still a bit confused what this means `ChecklistA` can't we give it a more explicit name?
Pretty sure we are not using `Checklist` anyway so maybe just use that?

packages/shared/src/graphql/actions.ts
```diff
@@ -2,6 +2,13 @@ import request, { gql } from 'graphql-request';
 import { graphqlUrl } from '../lib/config';
 
 export enum ActionType {
+  CreateSquad = 'createSquad',
```
rebelchris, line 5: Just make sure these match the backend as I think we use snake_case there.

packages/shared/src/lib/checklist.ts
```diff
@@ -0,0 +1,73 @@
...
+    description: string;
+  }>;
+  step: ChecklistStepType;
+  checked: boolean;
+  active: boolean;
+  onToggle: (action: Action) => void;
+  children?: ReactNode;
+};
+
+export const actionsPerRoleMap: Partial<
```
rebelchris, line 34: Not 100% sure this is valid as certain types can show based outside of role (for instance the create my feed.
Would this cover that?

packages/shared/src/components/checklist/ChecklistCard.tsx
```diff
@@ -0,0 +1,91 @@
...
+        <div className="p-4 rounded-t-14 bg-theme-bg-cabbage-opacity-24">
```
rebelchris, line 38: Is this not `cabbage-opacity-24`?

packages/shared/src/components/checklist/ChecklistStep.tsx
```diff
@@ -0,0 +1,96 @@
...
+            className={classNames(
+              '-ml-2 text-theme-label-tertiary',
+              isCompleted && 'text-theme-label-quaternary',
+            )}
+            id={step.action.type}
+            disabled={isCompleted}
+            icon={
+              <div
+                className={classNames(
+                  active && 'p-1 rounded-full bg-theme-bg-cabbage-opacity-24',
```
rebelchris, line 43: Same here is this one not `cabbage-opacity-24`?

packages/shared/src/components/checklist/ChecklistStep.tsx
```diff
@@ -0,0 +1,96 @@
...
+const ChecklistStep = ({
+  className = {},
+  step,
+  checked = false,
+  active = false,
+  onToggle,
+  children,
+}: ChecklistStepProps): ReactElement => {
+  const isCompleted = !!step.action.completedAt;
+  const isOpen = !isCompleted && checked;
```
rebelchris, line 18: Does this one mean every non completed step is open?
I think the idea is that only one step at the time will be active.

rebelchris (changes requested):
Wow really looking great already.
Have some suggestions and questions around certain aspects.

packages/shared/src/components/checklist/ChecklistStep.tsx
capJavert: I do not see this class existing

packages/shared/src/lib/checklist.ts
capJavert: But checklist is always firstly shown based on squad role, and then we would just add that action to both roles.

packages/shared/src/components/checklist/ChecklistStep.tsx
capJavert: I added this just so styles match what is in sketch (since it is that button there) outside one is a actually clickable handler. Let me know what would be better way to do it.

packages/shared/src/components/checklist/ChecklistStep.tsx
rebelchris: That the new one you introduced in this PR no?

packages/shared/src/components/checklist/NotificationChecklistStep.tsx
capJavert: Added tasks already to board.

packages/shared/src/components/icons/ChecklistA/index.tsx
capJavert: This is how it was in sketch, thought it would be best to just name them the same so down the line we know which icons we added and which ones we are missing.

packages/shared/src/graphql/actions.ts
capJavert: Done

packages/shared/src/components/icons/ChecklistA/index.tsx
rebelchris: Oh I see, would you mind checking with Design if they can rename it? 😅
If not it's fine, but would clean it a bit.

packages/shared/src/components/checklist/ChecklistStep.tsx
capJavert: No, just a clunky naming, refactored in 9409c10e30a4dd22875ead5299d74669e3276f03

packages/shared/src/components/checklist/ChecklistStep.tsx
capJavert: Yeah, it is called `bg-theme-bg-cabbage-opacity-24`, do you want me to rename or?

packages/shared/src/components/icons/ChecklistA/index.tsx
capJavert: I'll just rename it, don't want to waste time on.

packages/shared/src/components/checklist/ChecklistStep.tsx
capJavert: I did have styling issues with text wrapping and positioning because `children` of `QuaternaryButton` are wrapped in `label` which also does not have styling exposed. 

I removed the `QuaternaryButton` in fcc3ad5635c3ed36edc3c1d63fca3dc36bb524fa and just used plain styles, it looks the same.

capJavert:
@rebelchris I think I fixed everything from the comments and also refactored some stuff for easier reuse. Also added hide/show logic and mobile style (modal). Aside from some design stuff pending this is ready for review again.

packages/shared/src/components/checklist/ChecklistStep.tsx
```diff
@@ -0,0 +1,95 @@
...
+  step,
+  isOpen = false,
+  isActive = false,
+  onToggle,
+  children,
+}: ChecklistStepProps): ReactElement => {
+  const isCompleted = !!step.action.completedAt;
+
+  return (
+    <Container className={className.root}>
```
rebelchris, line 22: Ah sorry think you misinterpreted my previous comment, I mean instead of `root` we use `container` normally. Not specifically the div :)

packages/shared/src/components/checklist/ChecklistStep.tsx
```diff
@@ -0,0 +1,95 @@
...
+      <button
```
rebelchris, line 23: Way to go on. this! Looks way neater with just the one button.

packages/shared/src/components/checklist/SquadChecklistCard.tsx
```diff
@@ -0,0 +1,59 @@
...
+    SQUAD_CHECKLIST_VISIBLE_KEY,
+    false,
+  );
+  const { sidebarRendered } = useSidebarRendered();
+
+  if (!isChecklistVisible) {
+    return null;
+  }
+
+  const checklistElement = (
```
rebelchris, line 26: would it make any difference defining this after the modal return as it's not used if that's the case?

packages/shared/src/components/squads/SquadHeaderBar.tsx
```diff
@@ -47,6 +50,26 @@ export function SquadHeaderBar({
     key: TutorialKey.CopySquadLink,
   });
 
+  const { steps } = useSquadChecklistSteps({ squad });
+
+  const checklistTooltipText = useMemo(() => {
+    const completedStepsCount = steps.filter(
```
rebelchris, line 56: I have another suggestion here.
Maybe you can abstract the map inside the hook, so it's outside the memo.
Then keep this logic inside the hook (like `completedSteps`)
raising it because down the line you might want it somewhere else then all logic is handled in the hook.

You could also add `totalSteps` for instance

Let me know what you think.

packages/shared/src/components/squads/SquadHeaderMenu.tsx
```diff
@@ -65,12 +66,10 @@ export default function SquadHeaderMenu({
   const items = useMemo(() => {
     const list: ContextMenuItemProps[] = [
       {
-        Icon: TourIcon,
-        onClick: () =>
-          openModal({
-            type: LazyModal.SquadTour,
-          }),
-        label: 'Learn how Squads work',
+        Icon: FeedbackIcon,
```
rebelchris, line 69: Did you remove the tour in favor, not 100% sure that was intended (but could be wrong)
Please just make double sure they don't just want to add a option.

packages/shared/src/components/tooltips/ContextMenuItem.tsx
```diff
@@ -17,14 +18,17 @@ function ContextMenuItem({
...
   Icon,
   className,
+  target,
   ...props
 }: ContextMenuItemProps): ReactElement {
   return (
     <ConditionalWrapper
       condition={!!href}
       wrapper={(children) => (
         <Link href={href}>
```
rebelchris, line 28: Not sure don't you need target here as well for nextjs?

packages/shared/src/hooks/useChecklist.ts
```diff
@@ -0,0 +1,51 @@
...
+
+export type UseChecklist = {
+  openStep: string | undefined;
+  onToggleStep: (action: Action) => void;
+  isDone: boolean;
+  activeStep: ActionType | undefined;
+};
+
+const useChecklist = ({ steps }: UseChecklistProps): UseChecklist => {
+  const activeStep = useMemo(
```
rebelchris, line 17: Not 100% sure, but maybe it's also something for the hook?
In my opinion all calculation regarding to steps should happen in that isolated scoped, but let me know what you think.  (pros/cons)

packages/webapp/pages/squads/[handle]/index.tsx
```diff
@@ -62,6 +62,14 @@ const SquadEmptyScreen = dynamic(
     ),
 );
 
+const SquadChecklistCard = dynamic(async () => {
```
Comment (rebelchris) [inline packages/webapp/pages/squads/[handle]/index.tsx:65]: Just for my understanding why did you choose dynamic import when it's always rendered?
Seems to have no specific benefit that way.

rebelchris (changes requested):
Got some minor suggestions, but looking very good so far :)

packages/shared/src/components/checklist/SquadChecklistCard.tsx
capJavert: It is used in both cases. I just use Modal vs InteractivePopup as wrappers.

packages/webapp/pages/squads/[handle]/index.tsx
Comment (rebelchris) [reply packages/webapp/pages/squads/[handle]/index.tsx]: Alright cool makes sense though was just wondering the thoughts.
I'm not against it especially if it has benefits 🙏

packages/shared/src/components/tooltips/ContextMenuItem.tsx
capJavert: That should work on newer versions of Next.js. This one needs `target` on `a` to work. Tested it.

packages/shared/src/components/squads/SquadHeaderMenu.tsx
capJavert: Per design it should be removed but I'll check.

packages/shared/src/hooks/useChecklist.ts
capJavert: Actually my main point here is that I wish to separate base checklist logic vs specific steps logic. 

In the future we will have other checklist for new features that we don't even know about and this is where that kind of separation will hopefully pay out. 

For example active step for the checklist will be the same always, it is the next uncompleted step in steps list and it is defined by spec/design. Checklist does not know how many more steps are there or what they do, it just knows it needs to show active/next step to the user. 

This is also why each `ChecklistStep` can accept `component` where you can customize what each step does on per case by case basis.

So to conclude `useChecklist` and `Checklist*` components contains all the generic checklist logic that is tied to checklist UI and then steps creation is something that will be done separately for each new checklist eg. for squads we have `useSquadChecklistSteps` and `SquadChecklistCard` where that specific logic lives.

Hope it makes more sense.

packages/shared/src/components/squads/SquadHeaderBar.tsx
capJavert: We can move this to `useSquadChecklistSteps` sure 👍

packages/shared/src/components/checklist/ChecklistCard.tsx
```diff
@@ -0,0 +1,75 @@
+import React, { ReactElement } from 'react';
+import classNames from 'classnames';
+import { Card } from '../cards/Card';
+import { ChecklistStep } from './ChecklistStep';
+import { ChecklistCardProps } from '../../lib/checklist';
+import { useChecklist } from '../../hooks/useChecklist';
+import { ButtonSize } from '../buttons/Button';
+import CloseButton from '../CloseButton';
+import { RankConfetti } from '../../svg/RankConfetti';
```
capJavert, line 9: Used existing confetti.

packages/shared/src/components/squads/SquadHeaderBar.tsx
capJavert: Streamlined usage even more.

packages/shared/src/hooks/useChecklist.ts
rebelchris: Ah yes got it, so in that case we do need to ensure all calculations happen from the relevant hook, not the UI endpoints correct? (so useSquadChecklist)

capJavert: Yep, I streamlined everything even more in latest commit. Same as composing `Checklist*` components into `SquadChecklistCard` I did the same with `useChecklist` hook inside `useSquadChecklist`.

capJavert:
@sshanzel I would appreciate if you could also take a look just so we are all aligned.

packages/shared/src/components/squads/SquadHeaderBar.tsx
```diff
@@ -88,14 +95,23 @@ export function SquadHeaderBar({
           memberCount={memberCount}
         />
       )}
-      <SimpleTooltip placement="top" content="Feedback">
+      <SimpleTooltip
```
rebelchris, line 98: Totally fine for now, just noting down if this file is getting bloated maybe make a new component for this particular item. (non blocking/non urgent)

packages/shared/src/graphql/actions.ts
```diff
@@ -2,6 +2,14 @@ import request, { gql } from 'graphql-request';
...
 export enum ActionType {
+  CreateSquad = 'create_squad',
+  JoinSquad = 'join_squad',
+  EditWelcomePost = 'edit_welcome_post',
+  SquadFirstComment = 'squad_first_comment',
+  SquadFirstPost = 'squad_first_post',
+  MyFeed = 'my_feed',
+  SquadInvite = 'squad_invite',
+  BrowserExtension = 'browser_extension',
   Notification = 'notification',
```
rebelchris, line 13: This one is different though, make sure it's aligned with the DR

packages/shared/src/hooks/useChecklist.ts
```diff
@@ -0,0 +1,59 @@
...
+    () => steps.find((item) => !item.action.completedAt)?.action.type,
+    [steps],
+  );
+  const [openStep, setOpenStep] = useState<string>(activeStep);
+
+  useEffect(() => {
+    setOpenStep(activeStep);
+  }, [activeStep]);
```
rebelchris, line 27: Why do you need both?
Is it not enough to simply use `activeStep`?

packages/shared/src/hooks/useChecklist.ts
```diff
@@ -0,0 +1,59 @@
...
+    });
+  }, []);
+
+  const isDone = useMemo(() => {
+    return steps.every((item) => !!item.action.completedAt);
+  }, [steps]);
+
+  const completedSteps = useMemo(() => {
+    return steps.filter((step) => !!step.action.completedAt);
+  }, [steps]);
```
rebelchris, line 45: Do you need useMemo for each individual item if you already memoize the whole return object? (Not 100% sure)
Also tagging @idoshamun from his point of view.

rebelchris (approved):
Looks pretty neat!
Like the hooks you created, just have some minor non blocking comments.

Will approve from my side, keep in mind we need 2 approvals for this one as it's quite big.

packages/shared/src/hooks/useChecklist.ts
capJavert: We would not need it if we compute everything inside that single `useMemo`, if not `completedSteps` would for example have new reference on each render and then again useMemo at the bottom would recompute on each render making it invalid. The new rule of hooks would actually catch that in this case.

In regards to `useMemo` at the bottom in return, all of our hooks return stable reference so at this point just a consistency.

packages/shared/src/hooks/useChecklist.ts
capJavert: This is to sync up when `activeStep` changes due to next action being done, it will auto focus next action in the checklist. It actually provides good UX.

packages/shared/src/graphql/actions.ts
capJavert: It is already used as this in the PR that @sshanzel merged. @sshanzel maybe you can look if this should be `enable_notification`, I might have missed something.

sshanzel: Right. That is the placeholder value. We should have it converted to `enable_notification`. We haven't released this to prod (app.daily.dev) so it is safe for us to update here.

packages/shared/src/styles/base.css
```diff
@@ -166,6 +167,7 @@ body {
   --theme-background-bun: theme('colors.bun.20');
   --theme-background-onion: theme('colors.onion.20');
   --theme-background-pepper: theme('colors.pepper.20');
+  --theme-background-cabbage-opacity-24: theme('colors.cabbage.50')24;
```
sshanzel, line 170: Not sure if I am looking at the right Sketch part relevant to the usage of this.

The spec says 40 and opacity of 18%.

`18%` opacity I think should be `2B`.

Link for reference: https://gist.github.com/lopspower/03fb1cc0ac9f32ef38f4

Screenshot 2023-05-04 at 8 22 37 PM

packages/shared/src/hooks/useActions.ts
```diff
@@ -21,7 +21,7 @@ export const useActions = (): UseActions => {
   const { data: actions } = useQuery(
     generateQueryKey(RequestKey.Actions, user),
     getUserActions,
-    { enabled: !!user },
+    { enabled: !!user?.id },
```
sshanzel, line 24: This change shouldn't be needed as it is coming from the context, and if it is anonymous user, we return null for user.

packages/shared/src/components/tooltips/ContextMenuItem.tsx
```diff
@@ -1,14 +1,15 @@
...
-import React, { ReactElement } from 'react';
+import React, { HTMLAttributeAnchorTarget, ReactElement } from 'react';
 import ConditionalWrapper from '../ConditionalWrapper';
 import { IconProps, IconSize } from '../Icon';
 
 export interface ContextMenuItemProps extends Omit<ItemProps, 'children'> {
   href?: string;
   label: string;
   Icon: React.ComponentType<IconProps>;
+  target?: HTMLAttributeAnchorTarget;
```
sshanzel, line 12: As the props grow, I think it would be best to have the whole anchor attributes in a single property as we discussed in the past, then just pass along whatever is needed.

packages/shared/src/components/checklist/ChecklistCard.tsx
```diff
@@ -0,0 +1,75 @@
...
+  onRequestClose,
+}: ChecklistCardProps): ReactElement => {
+  const { isDone, openStep, onToggleStep, activeStep } = useChecklist({
+    steps,
+  });
+
+  return (
+    <div className={className}>
+      <Card className="p-0 bg-transparent rounded-14 !border-theme-color-cabbage hover:!border-theme-color-cabbage w-[340px]">
+        <div className="overflow-hidden relative p-4 bg-gradient-to-t from-cabbage-90 to-cabbage-50 rounded-t-12">
```
sshanzel, line 25: Not sure we should be using `from-cabbage-90 to-cabbage-50`. We need to use the one defined in our config, we might just have to find the correct one. I understand it is a pain for now, but hopefully we can create a documentation mapping this.

packages/shared/src/hooks/useChecklist.ts
rebelchris: Not sure they would get new reference if steps doesn't change?
I personally think just running useMemo on the return would be enough. *(this just feels like a overkill at this point)

sshanzel: I think the one creating new references is the `.filter`.

sshanzel (changes requested):
Wow! Massive work here 💪 

Really amazing with all the thoughts put into this, just some minor comments.

packages/shared/src/components/checklist/ChecklistCard.tsx
rebelchris: I think for this it's fine it's a exclusion that's the same render for both light/dark modes 😅
It's a bit tricky but we used it before like this. (Correct me if i'm wrong though)

sshanzel: Ah. On the banner. We might probably need to come up with some sort of system for this in the future. Just to be consistent and inline. In that case, kindly disregards this one 🚢

packages/shared/src/hooks/useActions.ts
capJavert: I just thought no point in doing a separate API request for anonymous since it will just return nothing. This way it should be only one once we know user is logged in and actually capable of completing actions.

packages/shared/src/components/checklist/ChecklistCard.tsx
capJavert: I see I posted comment late without seeing your comments. 

Ok, going with this, tnx for checking 🍻

rebelchris: Agree with Lee though, @capJavert can you add a backlog ticket just so we have it :)

capJavert: For sure!

capJavert: https://dailydotdev.atlassian.net/browse/WT-1316

packages/shared/src/hooks/useActions.ts
sshanzel: I mean on an anonymous user this won't make a request still because the `user` object is null if it is anonymous.

capJavert: In some cases I was getting `{ name: 'anonymous' }` or something like that, tried it now and I can not get it anymore. You might know?

packages/shared/src/components/tooltips/ContextMenuItem.tsx
capJavert: Added in 030a5a0ccb4cc68bd5fbdb828c440306e8fa5884

packages/shared/src/styles/base.css
capJavert: Checked with Tomer, it should be `24%` to match rest of the elements.

packages/shared/src/hooks/useChecklist.ts
capJavert: To be honest I would prefer removing return object `useMemo` from all of our hooks since we rarely need reference equality for the whole return value and most often use destructuring for specific properties, which is better pattern of course. Then it is also easier to selectively memoize specific values. 

I am open on removing return object `useMemo` for `useChecklist` and also we can add it as a rule of the thumb for all new hooks.

packages/shared/src/hooks/useActions.ts
sshanzel: Oh, interesting. I'll be sure to watch out if I encounter any. Just FYI, if ever you need the anonymous data (like tracking id) there should be an `anonymous` property from `AuthContext` 🚀

packages/shared/src/graphql/actions.ts
```diff
@@ -2,7 +2,15 @@ import request, { gql } from 'graphql-request';
...
-  Notification = 'notification',
+  CreateSquad = 'create_squad',
+  JoinSquad = 'join_squad',
+  EditWelcomePost = 'edit_welcome_post',
+  SquadFirstComment = 'squad_first_comment',
+  SquadFirstPost = 'squad_first_post',
+  MyFeed = 'my_feed',
+  SquadInvite = 'squad_invite',
+  BrowserExtension = 'browser_extension',
+  EnableNotification = 'enable_notification',
```
sshanzel, line 13: Thanks for fixing this!

packages/shared/src/styles/base.css
```diff
@@ -52,6 +52,7 @@ body {
   --theme-background-bun: theme('colors.bun.80');
   --theme-background-onion: theme('colors.onion.80');
   --theme-background-pepper: theme('colors.pepper.20');
+  --theme-background-cabbage-opacity-24: theme('colors.cabbage.40')24;
```
sshanzel, line 55: The hex value for 24% should be 3D, I think.

Reference: https://gist.github.com/lopspower/03fb1cc0ac9f32ef38f4

```suggestion
  --theme-background-cabbage-opacity-24: theme('colors.cabbage.40')3D;
```

packages/shared/src/styles/base.css
```diff
@@ -166,6 +167,7 @@ body {
   --theme-background-bun: theme('colors.bun.20');
   --theme-background-onion: theme('colors.onion.20');
   --theme-background-pepper: theme('colors.pepper.20');
+  --theme-background-cabbage-opacity-24: theme('colors.cabbage.40')24;
```
sshanzel, line 170: Just tagging for visibility.
```suggestion
  --theme-background-cabbage-opacity-24: theme('colors.cabbage.40')3D;
```

sshanzel (approved):
Got nothing else to say but great work! 🚢 🚢 🚢

packages/shared/src/styles/base.css
capJavert: Did not realize we are using hex values. Updated 👌

sshanzel: Me too, it was until @rebelchris gave me that link months ago 😂

packages/shared/src/hooks/useChecklist.ts
capJavert: `useMemo` does not control re-renders of the component that is using it. It only optimizes potentially heavy computations of the values it accepts. Having or not having `useMemo` component will re-render if:
1. its state changes (eg. useState)
2. any of the parent props that component (and/or hook) accepts changes

With `useMemo` we by default merely save on computations during those re-renders. Take the above `completedSteps` calculation. Without `useMemo` it would run on every render hence increasing render time. With `useMemo` it will run only if `steps` changes and will use memoized value on every other re-render (skipping actual iteration in `filter`).

Where `useMemo` helps potentially is with re-renders of children components. By default React function components will always re-render if parent component re-renders, which as we said above will happen due to points 1. or 2. Even if we memoize our props children components will re-render if parent component re-renders. This is how React works.

Where we can fruit benefits is if we were to use [memo](https://react.dev/reference/react/memo) on our components. This wrapper will skip rendering components that accept primitive values as props if neither of them change. `memo` also accepts `areEqual` function which can be used to handle non primitive values but much better way of handling this is to just always pass primitive values. How can we do that with arrays and objects who obviously are not primitives and they reference change? We `useMemo` to make them primitive. 

To go back to our `completedSteps` example which is `array` and not primitive value. Without `useMemo` it would have a new reference on each render and our `memo` wrapped component would re-render because it does not know it is actually visually same array. React under the hood uses [Object.is](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/is) to check that equality. 

With `useMemo` reference will stay the same and `memo` will see it as primitive value. It will only change reference if its dependencies change (in this case `steps`) and only then it will also trigger re-render of our `memo` wrapped children component. This is where we would directly save on re-renders with `useMemo`.

This is why removing `useMemo` from return object would no really do anything unless we were to pass whole return object to some component that uses `memo`. Most of the time we pass specific properties of our return object as props to our children components and this is why I mentioned below statement:

> To be honest I would prefer removing return object useMemo from all of our hooks since we rarely need reference equality for the whole return value and most often use destructuring for specific properties, which is better pattern of course. Then it is also easier to selectively memoize specific values.

In all other cases I like to think about `useMemo` as just a in render optimisation that can potentially (if needed) lead to easier re-render optimisations if we ever need to introduce `memo` on some of our components.

Some additional points:
1. `useMemo` of course helps with `useEffect` which also only knows how to check primitive values, and if `useEffect` calls a state setter then we could say `useMemo` here helped with re-renders eg. if we had `useEffect` that was dependant of our `completedSteps`
2. Should we add `memo` to all of our components, not really, even though it does a lot of work (vs some other frameworks like Svelte and SolidJS) React is plenty fast 99% of the time, only when we see some bottlenecks then we can utilise `memo`
3. We also use some 3party packages and components which do some complex stuff might be wrapped in `memo` by library author so by using `useMemo` we potentially save there
4. There is also a big discussion of `useMemo` overhead vs just relying on V8 runtime optimizations. My rule of the thumb that I sleep better if any kind of iteration or algorithmic function inside components is behind `useMemo`.. but that is a rabbit whole for another day 😆 

I hope I explained it here, would love to talk more about this if needed any time.

capJavert: @idoshamun @rebelchris @sshanzel since we are merging this to epic branch and we are onboard for all the features part of the PR I'll merge this now so we can continue on other squads v5 tasks and I can do another PR to resolve `useMemo` depending on the path we decide to take.

idoshamun: Happy to continue the discussion here even though you merged :)
So one thing that you missed is that whenever you define a new return object as json, even if the underlying props are the same it will be referred as a new object and will trigger re-render for those who use this hook. I encountered few years ago when I saw a suggestion to useMemo when returning a json object. Again my assumption might have changed but this is another use case for useMemo

capJavert: > So one thing that you missed is that whenever you define a new return object as json, even if the underlying props are the same it will be referred as a new object and will trigger re-render for those who use this hook.

Yes, this is true, but it would need to be used in specific ways, let's look at specific examples on when would this matter:
1. The whole hook value needs to be passed as dependency so:
```jsx
const checklist = useChecklist()

return <Child checklist={checklist} />
```

If `Child` is component wrapped in `memo` it will now re-render every time if `useChecklist` return object is not using `useMemo` because of course object is not primitive value. This is where return object `useMemo` matters as you said. ✅ 

That said, we mostly have this kind of usages:
```jsx
const checklist = useChecklist()

return <Child {...checklist} />
```

In above example it does not matter if return object is memoized or not it only matters if specific properties are memoized (and they are in the case of `useChecklist`) eg. `completedSteps`, `isDone`...

The above is of course shortcut for something like this:
```jsx
const { completedSteps, isDone } = useChecklist()

return <Child completedSteps={completedSteps} isDone={isDone}  />
```

The same principal applies here, `useMemo` is not needed for return object if we already memoize `completedSteps` and `isDone` separately, since those are actually passed as props separately.

Of course this all comes with a caveat that we have to use `memo` for `Child`, if it is just a function component it will render no matter no matter the props if its parent component(s) re-render.

Your point also stands for `useEffect` where of course we have this:
```jsx
const checklist = useChecklist()

useEffect(() => {
    // do something
}, [checklist])
```

it matters that return object is memoized here since it is directly used as dependency.

Again, we mostly use stuff like this.

```jsx
const { completedSteps } = useChecklist()

useEffect(() => {
    // do something
}, [completedSteps])
```

This is why I think according of how we use our special hooks we do not need to do this double memoization (for return object and separate properties), but leaving it we are not loosing anything drastic and re-renders wise we are not really saving anything since we do not utilise `React.memo` wrapper.

idoshamun: I'd love to see a minimal project where we can compare both and see the number of re-renders. You might be right and if so we need to have proper guidelines on that

capJavert: Here is a minimal implementation to see how different examples above re-render https://codesandbox.io/s/re-render-tests-with-usememo-and-memo-p7dg9r (see `index.js` for info on how to use different examples)

idoshamun: Thanks for setting it up so quickly. I think we need a bit different example, we need to check what happens when `steps` is updated by the deep value is the same. Whether it will cause re-renders in some cases or not. Anyway, I think it will be better to have a meeting all of us :)
Feel free to schedule.

PR #1773, opened 2023-04-28
feat: notification toggle on commentary and use actions hook

sshanzel:
- Notification toggle in commentary section.
- Refactor to clean up some prop management in Commentary of posting to Squad.
- Introduction of Actions that the user has completed already including completing ones.
- Complete notification action if the user has already subscribed.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1278 #done


packages/shared/src/contexts/AuthContext.tsx
```diff
@@ -121,6 +124,11 @@ export const AuthContextProvider = ({
   const [loginState, setLoginState] = useState<LoginState | null>(null);
   const endUser = user && 'providers' in user ? user : null;
   const referral = user?.referrer;
+  const { data: actions } = useQuery(
+    generateQueryKey(RequestKey.Actions, endUser),
+    getUserActions,
+    { enabled: !!endUser?.id },
+  );
```
capJavert, line 131: I think it is better to just move this into the `useActions` hook. I don't now if there are any benefits of tying this to `AuthContext` (we can always wait for user id there), we would also expose `isLoading` and other things for actions easier if query us just inside the hook. Also if for any reasons `actions` become heavy it is better to have it outside of AuthContext where we can then fetch only when needed.

capJavert (changes requested):
I like how this looks so far 👌 actions interface looking good. Comment about place of data fetching.

packages/shared/src/contexts/AuthContext.tsx
sshanzel: My concern is that whenever we initialize the `useActions` it will fetch the actions data. We could have instantiated on other places where we just need the completion function. So when a refetch triggers (let's say refocus) it could be triggered on 5 different areas which are unnecessary.

> Also if for any reasons actions become heavy it is better to have it outside of AuthContext where we can then fetch only when needed.

The call for the `actions` is async so it won't be affecting any other areas. Also that we discussed to fetch this only once on load after boot.

packages/shared/src/components/modals/PostToSquadModal.tsx
```diff
@@ -34,10 +38,18 @@ function PostToSquadModal({
   requestMethod = request,
   ...props
 }: PostToSquadModalProps): ReactElement {
+  const notificationState = useState(true);
```
rebelchris, line 41: why do you need state if it can't be set?

packages/shared/src/contexts/AuthContext.tsx
rebelchris: I agree with Ante, don't think this should be part of the auth context.
Even rather just remove it for now and tackle in separate PR maybe?

rebelchris (changes requested):
Please try to keep PR separate as it makes it harder to review if they introduce multiple tickets in one PR.
Added some minor comments.

packages/shared/src/components/modals/PostToSquadModal.tsx
sshanzel: It would be set from some components down the line (we pass this as a whole, didn't want to separate to two props).

packages/shared/src/contexts/AuthContext.tsx
sshanzel: > I agree with Ante, don't think this should be part of the auth context.

Alright, let's keep the query inside the hook.

> Even rather just remove it for now and tackle in separate PR maybe?

The hook itself is needed here, so decided to create it now than be blocked.

sshanzel:
> Please try to keep PR separate as it makes it harder to review if they introduce multiple tickets in one PR.

Didn't actually realize we had a separate ticket for the hook alone, only ended up assigning it to myself after looking for the next task to work on. Will try to separate things as much as possible.

rebelchris (approved):
Alright, like this approach 🙏

packages/shared/src/hooks/useActions.ts
```diff
@@ -0,0 +1,57 @@
...
+      );
+
+      return () =>
+        client.setQueryData<Action[]>(
+          generateQueryKey(RequestKey.Actions, user),
+          actions,
+        );
+    },
+    onError: (_, __, rollback: () => void) => {
+      rollback?.();
```
capJavert, line 44: Would prefer `if (rollback)` here, not really a fan of this inline optionals, error prone and harder to build on when adding additional stuff.

PR #1775, opened 2023-05-01
feat: prompt notification from a new notification item type

sshanzel:
- From the spec, it is mentioned that we should just open the prompt when clicked, so it may happen that the user received the in-app notification and when clicked we should just open the prompt.

Note: if we will have an email counterpart, we might need to support checking the URL for query parameters if it came from clicking some CTA inside the email. -still waiting from product's response.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1281 #done

### DIFF
--- a/packages/shared/src/components/notifications/InAppNotification.tsx
+++ b/packages/shared/src/components/notifications/InAppNotification.tsx
@@ -86,7 +86,7 @@ export function InAppNotificationElement(): ReactElement {
   }, [payload]);
 
   const onNotificationClick = (
-    e: React.MouseEvent<HTMLAnchorElement>,
+    event: React.MouseEvent<HTMLAnchorElement>,
     id: string,
     type: NotificationType,
   ) => {
@@ -97,7 +97,7 @@ export function InAppNotificationElement(): ReactElement {
     });
 
     if (type === NotificationType.SquadSubscribeNotification) {
-      e.preventDefault();
+      event.preventDefault();
       onTogglePermission(NotificationPromptSource.NotificationItem);
     }
   };
@@ -130,9 +130,9 @@ export function InAppNotificationElement(): ReactElement {
       />
       <InAppNotificationItem
         {...payload.notification}
-        onClick={(e) =>
+        onClick={(event) =>
           onNotificationClick(
-            e,
+            event,
             payload.notification.id,
             payload.notification.type,
           )
--- a/packages/shared/src/components/notifications/InAppNotification.tsx
+++ b/packages/shared/src/components/notifications/InAppNotification.tsx
@@ -13,11 +13,16 @@ import CloseButton from '../CloseButton';
 import { InAppNotificationItem } from './InAppNotificationItem';
 import styles from './InAppNotification.module.css';
 import { useAnalyticsContext } from '../../contexts/AnalyticsContext';
-import { AnalyticsEvent, Origin } from '../../lib/analytics';
+import {
+  AnalyticsEvent,
+  NotificationPromptSource,
+  Origin,
+} from '../../lib/analytics';
 import { NotificationType } from './utils';
 import FeaturesContext from '../../contexts/FeaturesContext';
 import { InAppNotificationPosition } from '../../lib/featureValues';
 import { ButtonSize } from '../buttons/Button';
+import { useNotificationContext } from '../../contexts/NotificationsContext';
 
 const Container = classed(
   'div',
@@ -35,6 +40,7 @@ export function InAppNotificationElement(): ReactElement {
   const client = useQueryClient();
   const { trackEvent } = useAnalyticsContext();
   const { clearNotifications, dismissNotification } = useInAppNotification();
+  const { onTogglePermission, isSubscribed } = useNotificationContext();
   const [isExit, setIsExit] = useState(false);
   const closeNotification = () => {
     setIsExit(true);
@@ -79,15 +85,26 @@ export function InAppNotificationElement(): ReactElement {
     };
   }, [payload]);
 
-  const onNotificationClick = (id: string, type: NotificationType) => {
+  const onNotificationClick = (
+    e: React.MouseEvent<HTMLAnchorElement>,
+    id: string,
+    type: NotificationType,
+  ) => {
     trackEvent({
       event_name: AnalyticsEvent.ClickNotification,
       target_id: id,
       extra: JSON.stringify({ origin: Origin.RealTime, type }),
     });
+
+    if (type === NotificationType.DemotedToMember) {
+      e.preventDefault();
+      onTogglePermission(NotificationPromptSource.NotificationItem);
+    }
   };
 
-  if (!payload?.notification) {
+  const isNotifTypeSubscribe =
+    payload?.notification?.type === NotificationType.SquadSubscribeNotification;
+  if (!payload?.notification || (isSubscribed && isNotifTypeSubscribe)) {
     return null;
   }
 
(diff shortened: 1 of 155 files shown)


packages/shared/src/components/notifications/InAppNotification.tsx
```diff
@@ -79,15 +85,26 @@ export function InAppNotificationElement(): ReactElement {
...
     trackEvent({
       event_name: AnalyticsEvent.ClickNotification,
       target_id: id,
       extra: JSON.stringify({ origin: Origin.RealTime, type }),
     });
+
+    if (type === NotificationType.SquadSubscribeNotification) {
+      event.preventDefault();
+      onTogglePermission(NotificationPromptSource.NotificationItem);
+    }
```
rebelchris, line 102: I would actually handle all of this on the `?notification` part as it has nothing to do with this.

rebelchris (changes requested):
I think the handling should rather happen on click which would solve all use-cases.
Including email which I don't think we'll have.

Either user clicks via notification center or the inapp -> targetURL = ?enable_notifications that will pick up the action.

LMWYT

packages/shared/src/components/notifications/InAppNotification.tsx
sshanzel: The issue is, the link is based on the BE. You could be at the notifications page or on the Squads page seeing the in app notification, and when clicking, the URL gives us a complete link, not just adding a parameter to the current query.

sshanzel:
> Either user clicks via notification center or the inapp -> targetURL = ?enable_notifications that will pick up the action.

The only issue is, the `targetUrl?notification=true` will never change, let's say we are pointing to the home page, when the user clicks the notification item, they would redirect first than just show the prompt.

One option I was actually thinking of is to open our dedicated page located at `/pages/popup/notification/enable` whether you are at extension or webapp.

packages/shared/src/components/notifications/InAppNotification.tsx
rebelchris: Yeah exactly so normally we use the main page for all of these.
As mentioned in the DR we decided on using the same approach for squad creation which is the homepage.
I would stick to what was decided there.

sshanzel: Yeah but the DR dictates a **similar** approach not necessarily the same. But if it is not really important to consider the UX then I'm all for redirection.

packages/webapp/pages/notifications.tsx
```diff
@@ -105,16 +119,28 @@ const Notifications = (): ReactElement => {
...
-                ({ node: { id, readAt, type, ...props } }) => (
-                  <NotificationItem
-                    key={id}
-                    {...props}
-                    type={type}
-                    isUnread={!readAt}
-                    onClick={() => onNotificationClick(id, type)}
-                  />
-                ),
+              page.notifications.edges.reduce(
```
sshanzel, line 122: Used `reduce` instead of `filter` then `map`.

packages/shared/src/contexts/NotificationsContext.tsx
```diff
@@ -195,6 +216,12 @@ export const NotificationsContextProvider = ({
     });
   }, [isInitialized, isInitializing, user]);
 
+  useEffect(() => {
+    if (isSubscribed || !router?.query.notify) return;
+
+    onTogglePermission(NotificationPromptSource.NotificationItem);
+  }, [onTogglePermission, isSubscribed, router]);
```
sshanzel, line 223: The `onTogglePermission` has to be a dependency which led to the conversion of the functions above to be used in a `useCallback`.

packages/webapp/pages/notifications.tsx
```diff
@@ -105,16 +113,28 @@ const Notifications = (): ReactElement => {
...
+                    return nodes;
+                  }
+
+                  nodes.push(
+                    <NotificationItem
+                      key={id}
+                      {...props}
+                      type={type}
+                      isUnread={!readAt}
+                      onClick={(e) => onNotificationClick(e, id, type)}
```
rebelchris, line 131: Why do we need the event now?

packages/shared/src/components/notifications/InAppNotification.tsx
```diff
@@ -79,15 +81,21 @@ export function InAppNotificationElement(): ReactElement {
...
+  ) => {
     trackEvent({
       event_name: AnalyticsEvent.ClickNotification,
       target_id: id,
       extra: JSON.stringify({ origin: Origin.RealTime, type }),
     });
   };
 
-  if (!payload?.notification) {
+  const isNotifTypeSubscribe =
```
rebelchris, line 96: But this would already not be rendered right? So why do we bother handling the click specificly?

packages/shared/src/contexts/NotificationsContext.tsx
rebelchris: Why do we need this anyway?
Not sure I follow the use for this.

rebelchris (changes requested):
I'm missing the hook we described in the DR that would handle the query param on the homepage correct?

packages/webapp/pages/notifications.tsx
sshanzel: Oh, this was just an unreverted change. Probably better to keep this as it is normal for event handlers to have this.

packages/shared/src/contexts/NotificationsContext.tsx
sshanzel: That's the handler to show the prompt of accepting permission for notifications. That's the core feature in this task.

packages/shared/src/components/notifications/InAppNotification.tsx
sshanzel: Just to catch any edge case and never show any form of subscribe to notification.

sshanzel:
> I'm missing the hook we described in the DR that would handle the query param on the homepage correct?

I have instead written it in the NotificationsContext because in the `_app.tsx` there is no idea of context yet where we need to check whether they have subscribed already.

sshanzel:
@rebelchris to keep this one and future implementations, I introduced a generically-named hook. Should be ready for review now 🥳

packages/shared/src/hooks/usePageParams.ts
```diff
@@ -0,0 +1,25 @@
+import { useEffect } from 'react';
+import { useRouter } from 'next/router';
+import { NotificationPromptSource } from '../lib/analytics';
+import { useNotificationContext } from '../contexts/NotificationsContext';
+import { stripLinkParameters } from '../lib/links';
+
+export const usePageParams = (): void => {
```
rebelchris, line 7: I do like the approach here, but now it's very focussed on a singular hook right?

I think we have two solutions:
1. Rename this to `useNotificationParams` (as it's only handling this for now)
2. Change the implementation so it can subscribe to multiple elements and handle them accordingly (example is squad creation and auth having these in place already)

Let me know what you think?
(My suggestion would be option 1 for now, we can abstract for another one?)

rebelchris (changes requested):
nice one small remark :)

packages/shared/src/hooks/usePageParams.ts
sshanzel: That makes sense, better to be clear than vague especially it does only one thing for now. Let's go with option 1.

rebelchris (approved):
Great good to go from my side.

PR #1776, opened 2023-05-02
feat: adding auto-population functionality to username on signup

TomDowling:
- On signup take the users name and generate a username
- Attempts 3 times before returning

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-804 #done


rebelchris (changes requested):

packages/shared/src/components/auth/RegistrationForm.tsx
```diff
@@ -58,14 +62,39 @@ export const RegistrationForm = ({
...
     });
   }, []);
 
+  useEffect(() => {
+    if (username.length) return;
+
+    if (generatedUsername?.generateUniqueUsername) {
+      setUsername(generatedUsername.generateUniqueUsername);
+    }
+  }, [generatedUsername]);
```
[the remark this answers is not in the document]
TomDowling: This isn't for the social login, this is only used to set the state/value of the username field once we get a result returned, hence I am unsure how this could be done on the API side?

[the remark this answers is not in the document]
rebelchris: Oh right, could we not simply call `setUsername` from the return value?
Seems like we can avoid the useEffect here.

packages/shared/src/components/auth/SocialRegistrationForm.tsx
```diff
@@ -174,6 +205,8 @@ export const SocialRegistrationForm = ({
           value={user?.name}
           valid={!nameHint && !hints?.name}
           hint={hints?.name || nameHint}
+          // value={name}
```
rebelchris, line 208: This seems to be a mistake right?

rebelchris (changes requested):
Looking good,

My main concern is the useEffects would love to see if we can either combine those 2 in a hook or simply use the callback from React query directly? (Maybe @sshanzel can help you)

And generic question, for the social signup is there no way we can already handle it when the user get's created? (This would avoid frontend calls)
Leave it up to @capJavert  and @sshanzel  to decide on that.

packages/shared/src/components/auth/RegistrationForm.tsx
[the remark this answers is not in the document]
TomDowling: Yes this wasn't something i was aware we could do, this has now be fixed in the next commit

packages/shared/src/components/auth/SocialRegistrationForm.tsx
TomDowling: This was mean't to be removed, I have since deleted it

packages/shared/src/components/auth/SocialRegistrationForm.tsx
```diff
@@ -60,11 +64,40 @@ export const SocialRegistrationForm = ({
...
   const { trackEvent } = useContext(AnalyticsContext);
-  const { user } = useContext(AuthContext);
+  const { user, updateUser } = useContext(AuthContext);
   const [nameHint, setNameHint] = useState<string>(null);
   const [usernameHint, setUsernameHint] = useState<string>(null);
   const [twitterHint, setTwitterHint] = useState<string>(null);
+  const [name, setName] = useState<string>('');
   const isAuthorOnboarding = trigger === AuthTriggers.Author;
+  const { requestMethod } = useRequestProtocol();
+  const usernameQueryKey = ['generateUsername', name];
```
rebelchris, line 74: This code is now pretty much duplicate, except the success handler being different. 
I'd suggest abstracting it to a custom hook so we can re-use it.

Maybe make it accept a callback function to handle both cases.

rebelchris (changes requested):
One more abstraction suggestion

packages/shared/src/components/auth/RegistrationForm.tsx
```diff
@@ -58,7 +62,33 @@ export const RegistrationForm = ({
...
+    {
+      enabled: !!name.length,
+      onSuccess: (data) => {
+        if (!data.generateUniqueUsername.length || !!username.length) return;
+
+        if (data?.generateUniqueUsername) {
+          setUsername(data.generateUniqueUsername);
+        }
+      },
+    },
```
capJavert, line 90: It is actually an anti pattern to use this callback since it is being removed from `useQuery` in future versions of `react-query`. https://tkdodo.eu/blog/breaking-react-querys-api-on-purpose

They actually propose `useEffect` as alternative, I realize it was something @rebelchris commented against but not sure what other thing we could use. Hopefully now that we merged https://github.com/dailydotdev/apps/pull/1745 we can at least use it more reliably.

capJavert (changes requested):
Also agree with @rebelchris about extracting the code the hook now that is finished. Additional comment about `onSuccess` usage.

packages/shared/src/components/auth/RegistrationForm.tsx
rebelchris: Ah sorry my bad, wasn't aware of this, but thanks for raising this.

So my suggestion when we switch to a hook we can leverage useEffect again.
@capJavert Maybe you can help tom with this solution if you agree with this?

capJavert: Talked with @rebelchris, `useEffect` should be fine here to sync up the generate username but for sure let's move whole logic to hook to keep it contained.

sshanzel: Oh, wow. That is a huge change. Thanks for letting us know about it.

packages/shared/src/components/auth/SocialRegistrationForm.tsx
```diff
@@ -60,11 +64,40 @@ export const SocialRegistrationForm = ({
   isLoading,
 }: SocialRegistrationFormProps): ReactElement => {
   const { trackEvent } = useContext(AnalyticsContext);
-  const { user } = useContext(AuthContext);
+  const { user, updateUser } = useContext(AuthContext);
   const [nameHint, setNameHint] = useState<string>(null);
   const [usernameHint, setUsernameHint] = useState<string>(null);
   const [twitterHint, setTwitterHint] = useState<string>(null);
+  const [name, setName] = useState<string>('');
```
sshanzel, line 71: Since we pass the string literal, I think the type will be inferred.
```suggestion
  const [name, setName] = useState('');
```

sshanzel (approved):
All good from my side 🥳 Just the open comments and I think we will be good to go 🚢

packages/shared/src/components/auth/SocialRegistrationForm.tsx
```diff
@@ -60,18 +64,45 @@ export const SocialRegistrationForm = ({
...
+  useEffect(() => {
+    if (!!user?.username || !!user?.username?.length) return;
+
+    if (generatedUsername?.generateUniqueUsername) {
+      updateUser({
+        ...user,
+        username: generatedUsername.generateUniqueUsername,
+      });
+    }
+  }, [generatedUsername]);
```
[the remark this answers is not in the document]
rebelchris: I personally think maybe just wait till the user submits?
No point in calling API here if they won't complete anyway.

Is that inline with your thoughts?

packages/shared/src/components/auth/RegistrationForm.tsx
```diff
@@ -58,7 +59,18 @@ export const RegistrationForm = ({
...
   const isAuthorOnboarding = trigger === AuthTriggers.Author;
+  const { data: usernameData, isLoading } = useGenerateUsername(name);
+
+  useEffect(() => {
+    if (username.length || isLoading) return;
+
+    if (usernameData?.generateUniqueUsername) {
+      setUsername(usernameData.generateUniqueUsername);
+    }
+  }, [usernameData, isLoading, username, setUsername]);
```
rebelchris, line 73: Was it not possible to add this inside the hook rather as it's basically the same code twice now?

rebelchris (commented):
One more question which was my idea for the hook.

rebelchris (approved):
Nice work, @sshanzel and @capJavert love to see your thoughts as well.

packages/shared/src/hooks/index.ts
```diff
@@ -0,0 +1 @@
+export * from './useGenerateUsername';
```
sshanzel, line 1: Thank you for the initiative 🙇

sshanzel (approved):
Neat!! ✨

packages/shared/src/hooks/useGenerateUsername.ts
```diff
@@ -0,0 +1,47 @@
...
+      setUsername(data.generateUniqueUsername);
+    }
+  }, [data, isLoading, username]);
+
+  return useMemo(
+    () => ({
+      username,
+      setUsername,
+    }),
+    [username, setUsername],
```
capJavert, line 45: `setUsername` as `SetStateAction` is stable by React so you don't need it here (if you remove it you will see that linter does not complain). 

Per new guidelines I would also remove `useMemo` from here completely.

capJavert (approved):
🚀 good job, single comment about the hook but approving it still 👌

PR #1781, opened 2023-05-04
feat: adjustments to useActions after some usage in other tickets

capJavert:
- make sure optimistic update works when no actions are completed, currently optimistic update would not go through (API call worked)
- skip API call if action is completed locally, optimization to avoid calling `checkHasCompleted` in all places we use `completeAction`

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done

### DIFF
--- a/packages/shared/src/hooks/useActions.ts
+++ b/packages/shared/src/hooks/useActions.ts
@@ -28,9 +28,17 @@ export const useActions = (): UseActions => {
       client.setQueryData<Action[]>(
         generateQueryKey(RequestKey.Actions, user),
         (data) => {
-          if (!Array.isArray(data)) return [];
+          const optimisticAction = {
+            userId: user.id,
+            type,
+            completedAt: new Date(),
+          };
 
-          return [...data, { userId: user.id, type, completedAt: new Date() }];
+          if (!Array.isArray(data)) {
+            return [optimisticAction];
+          }
+
+          return [...data, optimisticAction];
         },
       );
 
--- a/packages/shared/src/hooks/useActions.ts
+++ b/packages/shared/src/hooks/useActions.ts
@@ -45,13 +45,20 @@ export const useActions = (): UseActions => {
     },
   });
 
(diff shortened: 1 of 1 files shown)


packages/shared/src/hooks/useActions.ts
```diff
@@ -45,13 +45,20 @@ export const useActions = (): UseActions => {
     },
   });
 
-  return useMemo<UseActions>(
-    () => ({
+  return useMemo<UseActions>(() => {
+    const checkHasCompleted = (type: ActionType) =>
+      actions?.some((action) => action.type === type && !!action.completedAt);
```
rebelchris, line 50: This can be outside the return no? Just seems more clean (and then set dependency)

rebelchris (commented):
Minor suggesiton

sshanzel:
Oh, would it not reflect correctly even after setting `onMutate`?

capJavert:
> Oh, would it not reflect correctly even after setting `onMutate`? Weird stuff!

Yeah, `onMutate` is called before actual API request so if user completed no actions `length` would be `0` and optimistic insert would not happen because of that early `return`.

packages/shared/src/hooks/useActions.ts
```diff
@@ -28,7 +28,7 @@ export const useActions = (): UseActions => {
       client.setQueryData<Action[]>(
         generateQueryKey(RequestKey.Actions, user),
         (data) => {
-          if (!data?.length) return [];
+          if (!Array.isArray(data)) return [];
```
sshanzel, line 31: In that case, should the fix be as below? Yeah, I totally missed this one.
```suggestion
          if (!Array.isArray(data)) return [{ userId: user.id, type, completedAt: new Date() }];
```

sshanzel (approved):
Nice work! 🚢 🚢 🚢

PR #1784, opened 2023-05-05
feat(checklist): add share first post step

capJavert:
- Add step completion logic and components

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1261 #done

### DIFF
--- a/packages/shared/src/components/checklist/SharePostChecklistStep.tsx
+++ b/packages/shared/src/components/checklist/SharePostChecklistStep.tsx
@@ -6,8 +6,6 @@ import { FlexRow } from '../utilities';
 import { useLazyModal } from '../../hooks/useLazyModal';
 import { LazyModal } from '../modals/common/types';
 import { Squad } from '../../graphql/sources';
-import { useActions } from '../../hooks/useActions';
-import { ActionType } from '../../graphql/actions';
 import OnboardingContext from '../../contexts/OnboardingContext';
 
 const SharePostChecklistStep = ({
@@ -15,7 +13,6 @@ const SharePostChecklistStep = ({
   ...props
 }: ChecklistStepProps & { squad: Squad }): ReactElement => {
   const { openModal } = useLazyModal();
-  const { completeAction } = useActions();
   const { onInitializeOnboarding, showArticleOnboarding } =
     useContext(OnboardingContext);
 
@@ -31,9 +28,6 @@ const SharePostChecklistStep = ({
                 ...props,
                 squad,
                 preview: { url: '' },
-                onSharedSuccessfully: () => {
-                  completeAction(ActionType.SquadFirstPost);
-                },
               },
             });
           }}
--- a/packages/shared/src/components/checklist/SharePostChecklistStep.tsx
+++ b/packages/shared/src/components/checklist/SharePostChecklistStep.tsx
@@ -20,7 +20,7 @@ const SharePostChecklistStep = ({
     <ChecklistStep {...props}>
       <FlexRow className="gap-4">
         <Button
-          className="btn-secondary"
+          className={showArticleOnboarding ? 'btn-secondary' : 'btn-primary'}
           onClick={() => {
             openModal({
               type: LazyModal.PostToSquad,
--- a/packages/shared/src/components/checklist/SharePostChecklistStep.tsx
+++ b/packages/shared/src/components/checklist/SharePostChecklistStep.tsx
@@ -1,4 +1,4 @@
-import React, { ReactElement } from 'react';
+import React, { ReactElement, useContext } from 'react';
 import { ChecklistStepProps } from '../../lib/checklist';
 import { Button } from '../buttons/Button';
 import { ChecklistStep } from './ChecklistStep';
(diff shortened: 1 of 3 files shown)


packages/shared/src/components/modals/PostToSquadModal.tsx
```diff
@@ -57,9 +59,12 @@ function PostToSquadModal({
     buttonText: 'Done',
     preview,
   });
+  const { completeAction } = useActions();
 
   const shouldShowToggle = !isDismissed && !isSubscribed;
   const onPostSuccess = async (squadPost?: Post) => {
+    completeAction(ActionType.SquadFirstPost);
```
capJavert, line 66: @sshanzel @rebelchris what do you think of where should we actually finish this kind of action. 

Should we keep this kind of call from the frontend or we could potentially add logic to mutation on API side and then just call `invalidateQueries`( as we do for `sourceFeed` in this same callback) to refetch actions and show updated state in checklist?

For actions like extension download of course we have to do it from the frontend but for this one we could do it on the API since we already know action is completed there. I can see pros/cons for both ways, curious what you think.

sshanzel: This is already handled in the BE (postAdded worker) though we might need to update/invalidate the cache from the frontend.

capJavert: I'll add cache invalidation then as mentioned above.

packages/shared/src/components/checklist/SharePostChecklistStep.tsx
```diff
@@ -1,17 +1,49 @@
...
+        {showArticleOnboarding && (
+          <Button
+            className="btn-primary"
+            onClick={() => {
+              onInitializeOnboarding();
+            }}
+          >
+            Explore
+          </Button>
+        )}
```
capJavert, line 46: If the user completed onboarding then we do not show explore button right? This could be a good check do differentiate new user maybe.

Do we even need separate action for this (we now have `ActionType.MyFeed`)? I know in some PR there were mentions of replacing `alerts` with this, but can not find it.

capJavert, line 46: If the user completed onboarding then we do not show explore button right? This could be a good check do differentiate new user maybe.

Do we even need separate action for this (we now have `ActionType.MyFeed`)? I know in some PR there were mentions of replacing `alerts` with this, but can not find it.

packages/shared/src/components/modals/PostToSquadModal.tsx
```diff
@@ -70,7 +71,10 @@ function PostToSquadModal({
     }
 
     displayToast('This post has been shared to your Squad');
-    await client.invalidateQueries(['sourceFeed', user.id]);
+    await Promise.allSettled([
+      client.invalidateQueries(['sourceFeed', user.id]),
+      client.invalidateQueries(generateQueryKey(RequestKey.Actions, user)),
```
capJavert, line 76: Completion done server side, invalidating per comment here https://github.com/dailydotdev/apps/pull/1784#discussion_r1186670795

packages/shared/src/components/checklist/SharePostChecklistStep.tsx
rebelchris: Hmm, onboarding also finishes when user presses skip I think.
However we should have the action for this already as `my_feed` (Just not sure how it's set now, so maybe @sshanzel can mention this)

packages/shared/src/hooks/useSquadChecklist.tsx
```diff
@@ -69,16 +70,20 @@ const useSquadChecklist = ({
...
-        },
-        actions,
-      }),
+      [ActionType.SquadFirstPost]:
+        verifyPermission(squad, SourcePermissions.Post) &&
+        createChecklistStep({
+          type: ActionType.SquadFirstPost,
+          step: {
+            title: 'Share your first post',
+            description:
```
rebelchris, line 79: Just note that this description is different depending on whether someone is new as well.

rebelchris (changes requested):
Small feedback point.
We should have this alert in place already.

packages/shared/src/hooks/useSquadChecklist.tsx
```diff
@@ -69,16 +72,21 @@ const useSquadChecklist = ({
...
-        step: {
-          title: 'Share your first post',
-          description:
-            'Share your first post to help other squad members discover content you found interesting.',
-          component: SharePostChecklistStep,
-        },
-        actions,
-      }),
+      [ActionType.SquadFirstPost]:
+        verifyPermission(squad, SourcePermissions.Post) &&
```
rebelchris, line 76: Could you check with Sab on this please.

What happens when squad doesn't allow posting, but user is new, should we simply render different options? (or not maybe it's to edge-case)

rebelchris (approved):
One small question - non blocking

packages/shared/src/components/checklist/SharePostChecklistStep.tsx
capJavert: `ActionType.MyFeed` gets set inside `updateAlerts` on the API side so it is covered, [reference change here](https://github.com/dailydotdev/daily-api/commit/6b1962a77337aefb0315355ebd2f52fbe34d6f83#diff-7632be2baf678486983af835e6be44d8f3d75a9641b49c9cc38942379b3a171eR114-R116)

PR #1788, opened 2023-05-08
feat(checklist): sort steps by completedAt

capJavert:
- I evaluated to do this on the API side but since user actions on the API are generic interface and they just happen to be sorted in checklist steps based on completion I decided to do it inside `useChecklist` hook that contains other checklist specific logic

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1309 #done

### DIFF
--- a/packages/shared/src/hooks/useChecklist.ts
+++ b/packages/shared/src/hooks/useChecklist.ts
@@ -45,8 +45,20 @@ const useChecklist = ({ steps }: UseChecklistProps): UseChecklist => {
   }, [steps]);
 
   return useMemo(() => {
+    const sortedStepsByCompletion = [...steps].sort((a, b) => {
+      if (a.action.completedAt && !b.action.completedAt) {
+        return -1;
+      }
+
+      if (!a.action.completedAt && b.action.completedAt) {
+        return 1;
+      }
+
+      return 0;
+    });
+
     return {
-      steps,
+      steps: sortedStepsByCompletion,
       openStep,
       onToggleStep,
       isDone,


packages/shared/src/hooks/useChecklist.ts
```diff
@@ -45,8 +45,20 @@ const useChecklist = ({ steps }: UseChecklistProps): UseChecklist => {
...
+      if (a.action.completedAt && !b.action.completedAt) {
+        return -1;
+      }
+
+      if (!a.action.completedAt && b.action.completedAt) {
+        return 1;
+      }
+
+      return 0;
+    });
```
rebelchris, line 58: Wouldn't this already work?

```
steps.sort((a, b) => a.completedAt - b.completedAt);
```

rebelchris (changes requested):
Just a generic question.
Also is it easier to handle this on API rather? (just a question at this point)

packages/shared/src/hooks/useChecklist.ts
capJavert: I think it would but personally I like verbosity when not working with numbers.

As for API, I think doing it here just takes care of it from single point in code as said in description of this PR, we would also have to do sorting logic in optimistic mutation in completeAction.

> I evaluated to do this on the API side but since user actions on the API are generic interface and they just happen to be sorted in checklist steps based on completion I decided to do it inside useChecklist hook that contains other checklist specific logic

rebelchris (approved):

PR #1789, opened 2023-05-08
feat(checklist): add enable notification step

capJavert:
- Add step completion logic and components

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1311 #done

### DIFF
--- a/packages/shared/src/components/checklist/NotificationChecklistStep.tsx
+++ b/packages/shared/src/components/checklist/NotificationChecklistStep.tsx
@@ -15,7 +15,7 @@ const NotificationChecklistStep = (props: ChecklistStepProps): ReactElement => {
         icon={<BellIcon />}
         className="btn-primary"
         onClick={() => {
-          onTogglePermission(NotificationPromptSource.SquadPage);
+          onTogglePermission(NotificationPromptSource.SquadChecklist);
         }}
       >
         Subscribe
--- a/packages/shared/src/components/checklist/NotificationChecklistStep.tsx
+++ b/packages/shared/src/components/checklist/NotificationChecklistStep.tsx
@@ -1,15 +1,23 @@
-import React, { ReactElement } from 'react';
+import React, { ReactElement, useContext } from 'react';
 import { ChecklistStepProps } from '../../lib/checklist';
 import { Button } from '../buttons/Button';
 import { ChecklistStep } from './ChecklistStep';
 import BellIcon from '../icons/Bell';
+import NotificationsContext from '../../contexts/NotificationsContext';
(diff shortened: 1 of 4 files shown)


capJavert:
Tested locally and works, `NotificationsContextProvider` automatically completes the step from previous implementation.

packages/shared/src/components/checklist/NotificationChecklistStep.tsx
```diff
@@ -1,15 +1,23 @@
...
+  const { onTogglePermission } = useContext(NotificationsContext);
 
   return (
     <ChecklistStep {...props}>
-      <Button icon={<BellIcon />} className="btn-primary">
+      <Button
+        icon={<BellIcon />}
+        className="btn-primary"
+        onClick={() => {
+          onTogglePermission(NotificationPromptSource.SquadPage);
```
rebelchris, line 18: But this won't save it in the database right?
I think we need that part for other actions we might trigger later.

capJavert: It will get triggered from `useEffect` in `NotificationsContextProvider` if user actually accepts the prompt from browser to allow notifications. 

We can call `completeAction` here to complete the step no matter if user actually accepts notifications or not - this is a question from my point of view.

rebelchris: Ah no was just missing that context in this PR, if that's the case it's fine with me.
Maybe just ask @sshanzel if that was his intended behaviour.

sshanzel: If our goal is to just make user click the CTA, we can probably send the completeAction already, but if we want to force them to subscribe before completing, then we will have to rely on the context's capacity. Probably a question for product on their intent with this one.

capJavert: From [spec](https://dailydotdev.atlassian.net/wiki/spaces/HAN/pages/472743940/Squads+closed+beta+v5.0) it says:
image

Current implementation matches that so I would leave it at this.

rebelchris: I actually think the discussions mention click not completion, we had this discussion already.

rebelchris, line 18: Would just mind checking with Sab if we wants new tracking or this is ok?
Not sure they might want a new context from checklist.

rebelchris (approved):
One question, non breaking the approval though

PR #1792, opened 2023-05-09
feat(checklist): tests for component and the hook

capJavert:
- Add tests for checklist components and hook

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done

### DIFF
--- a/packages/shared/src/components/checklist/ChecklistStep.spec.tsx
+++ b/packages/shared/src/components/checklist/ChecklistStep.spec.tsx
@@ -1,0 +1,178 @@
+import React from 'react';
+import {
+  fireEvent,
+  render,
+  RenderResult,
+  screen,
+} from '@testing-library/react';
+import { ChecklistStep } from './ChecklistStep';
+import { ChecklistStepProps } from '../../lib/checklist';
+import { ActionType } from '../../graphql/actions';
+
+const step = {
+  title: 'Step 1',
+  description: 'Step 1 description',
+  action: {
+    type: ActionType.CreateSquad,
+    completedAt: null,
+  },
+};
+const noop = jest.fn();
+const customChildren = (
+  <div data-testid="checklist-step-custom-component">Custom component</div>
+);
+
+describe('ChecklistStep component', () => {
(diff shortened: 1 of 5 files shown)


rebelchris (changes requested):
Nice test cases are solid, would just try to refactor in avoiding test-id as they quickly become difficult to manage and update.
Relying on what's actually rendered should be a better approach in those cases.

packages/shared/src/components/checklist/ChecklistStep.spec.tsx
```diff
@@ -0,0 +1,178 @@
...
+      isActive: false,
+      onToggle,
+      children: customChildren,
+    });
+
+    const checklistElement = await screen.findByTestId('checklist-step');
+    expect(checklistElement).toBeInTheDocument();
+
+    const closed = await screen.findByTestId('checklist-step-closed');
+    expect(closed).toBeInTheDocument();
```
rebelchris, line 161: Same here

rebelchris (changes requested):
One small remark for uniforming test cases.
Once that is done consider it approved :)

packages/shared/src/components/checklist/ChecklistStep.spec.tsx
capJavert: @rebelchris what is the remark here, avoiding `data-testid` again or something else?

rebelchris: Yes correct, I see my other comment is gone, but you can use `findByTitle('Close')` to find these buttons.

capJavert: Yeah, I updated one in `ChecklistCard`, but this one does not have the title since `CloseButton` is not used.

rebelchris (approved):
Nice work, thanks a lot for this 🙏

PR #1794, opened 2023-05-11
feat: post modal/page for welcome post type

sshanzel:
- The UI of welcome post and share post looks very very similar, understandably since because they are both a Squad post.
- Introduce new type of post which requires different rendering.
- Fixed the SharePost author image size (should have been 64px all along).
- Refactored the SquadPostModal to be reused for welcome post type.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-WT-1329 #done


packages/shared/src/components/post/SharePostContent.tsx
```diff
@@ -0,0 +1,111 @@
+import React, { ReactElement, useContext, useMemo, useState } from 'react';
```
sshanzel, line 1: Copy-pasted the whole current elements after the title.

packages/shared/src/components/ProfilePicture.tsx
```diff
@@ -26,23 +27,25 @@ export interface ProfilePictureProps
...
-const sizeClasses = {
+const sizeClasses: Record<ProfileImageSize, string> = {
   xsmall: 'w-5 h-5',
   small: 'w-6 h-6',
   medium: 'w-8 h-8',
   large: 'w-10 h-10',
   xlarge: 'w-12 h-12',
   xxlarge: 'w-14 h-14',
-  xxxlarge: 'w-24 h-24',
+  xxxlarge: 'w-16 h-16',
```
sshanzel, line 37: Our guideline states `xxxlarge` to be `w-16 h-16`.

rebelchris: Does it have impact on any existing code?
Please tripple check 🙏

packages/shared/src/components/Feed.tsx
```diff
@@ -133,6 +133,7 @@ const getStyle = (useList: boolean, spaciness: Spaciness): CSSProperties => {
 const PostModalMap: Record<PostType, typeof ArticlePostModal> = {
   [PostType.Article]: ArticlePostModal,
   [PostType.Share]: SharePostModal,
+  [PostType.Welcome]: SharePostModal,
```
rebelchris, line 136: Why do you add this if it's the same modal?
Not quite sure I get this one.

packages/shared/src/components/ProfilePicture.tsx
```diff
@@ -11,7 +11,8 @@ export type ProfileImageSize =
   | 'large'
   | 'xlarge'
   | 'xxlarge'
-  | 'xxxlarge';
+  | 'xxxlarge'
+  | 'profile';
```
rebelchris, line 15: Rather check with Tomer, I'd like to keep things uniform

rebelchris (commented):
Looks good, just not sure why we have a new modal when it's all the same?

packages/shared/src/components/ProfilePicture.tsx
sshanzel: There was only one implementation which is the profile. Should be covered here 🚢

packages/shared/src/components/Feed.tsx
sshanzel: This is to identify which modal the feed will use for the post clicked.

rebelchris: But they all render the same modal?
Should we not rahter have a PostModal with prop type? (It seems a bit redundant to have 2 calls do the same modal.

rebelchris: Ah sorry was misreading this piece, it's just the render that's the same.
For some reason thought you where introducing a new lazy modal with the same modal rendered 😅

packages/shared/src/components/ProfilePicture.tsx
sshanzel: Agree with that. Awaiting response now 🚢

rebelchris (approved):
Nice work Lee 🙏

PR #1795, opened 2023-05-11
feat: add WelcomePostCard

capJavert:
- Add welcome post card, will be potentially refactored to match freeform post card in v6
- Used existing components, added custom header and footer to match other extensions (as SharedPostCard)

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1292 #done

### DIFF
--- a/packages/shared/src/components/BookmarkFeedLayout.tsx
+++ b/packages/shared/src/components/BookmarkFeedLayout.tsx
@@ -15,7 +15,7 @@ import BookmarkEmptyScreen from './BookmarkEmptyScreen';
 import { Button } from './buttons/Button';
 import ShareIcon from './icons/Share';
 import { generateQueryKey, RequestKey } from '../lib/query';
-import { PostType } from '../graphql/posts';
+import { supportedTypesForPrivateSources } from '../graphql/posts';
 
 export type BookmarkFeedLayoutProps = {
   searchQuery?: string;
@@ -48,7 +48,7 @@ export default function BookmarkFeedLayout({
         query: SEARCH_BOOKMARKS_QUERY,
         variables: {
           query: searchQuery,
-          supportedTypes: [PostType.Article, PostType.Share, PostType.Welcome],
+          supportedTypes: supportedTypesForPrivateSources,
         },
         emptyScreen: <SearchEmptyScreen />,
       };
@@ -58,7 +58,7 @@ export default function BookmarkFeedLayout({
       feedQueryKey: defaultKey,
       query: BOOKMARKS_FEED_QUERY,
       variables: {
-        supportedTypes: [PostType.Article, PostType.Share, PostType.Welcome],
+        supportedTypes: supportedTypesForPrivateSources,
       },
       onEmptyFeed: () => setShowEmptyScreen(true),
       options: { refetchOnMount: true },
--- a/packages/shared/src/components/BookmarkFeedLayout.tsx
+++ b/packages/shared/src/components/BookmarkFeedLayout.tsx
@@ -15,6 +15,7 @@ import BookmarkEmptyScreen from './BookmarkEmptyScreen';
 import { Button } from './buttons/Button';
 import ShareIcon from './icons/Share';
 import { generateQueryKey, RequestKey } from '../lib/query';
+import { PostType } from '../graphql/posts';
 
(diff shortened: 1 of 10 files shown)


packages/shared/src/components/cards/WelcomePostCard.tsx
```diff
@@ -0,0 +1,79 @@
...
+      </CardTitle>
+      <CardSpace />
+      <Container ref={containerRef}>
+        <WelcomePostCardFooter post={post} />
+        <ActionButtons
+          openNewTab={openNewTab}
+          post={post}
+          onUpvoteClick={onUpvoteClick}
+          onCommentClick={onCommentClick}
+          onBookmarkClick={onBookmarkClick}
```
rebelchris, line 68: Just make sure this supports in bookmarks as well, sounds like something that could break 😅

rebelchris (approved):
No comments besides Tomer potential change.

packages/shared/src/components/cards/WelcomePostCard.tsx
capJavert: @rebelchris good catch, fixed here e301f3e4d7a9126f0142390829aa3679d784889c

packages/shared/src/components/cards/WelcomePostCardHeader.tsx
```diff
@@ -0,0 +1,34 @@
...
+          source={source}
+          className="absolute -right-2 -bottom-2"
+          size="xsmall"
+        />
+      </div>
+      <div className="flex flex-col flex-1 flex-grow mr-6 ml-2 typo-footnote">
+        <span className="font-bold line-clamp-2">{author.name}</span>
+        <PostMetadata
+          className="break-words line-clamp-1"
+          createdAt={createdAt}
```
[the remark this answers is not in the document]
sshanzel: Probably just comment it out if ever or add some condition if it is a welcome post?

sshanzel (approved):
Looks good to me code-wise 🥳

packages/shared/src/components/BookmarkFeedLayout.tsx
```diff
@@ -45,14 +46,20 @@ export default function BookmarkFeedLayout({
         feedName: 'search-bookmarks',
         feedQueryKey: defaultKey.concat(searchQuery),
         query: SEARCH_BOOKMARKS_QUERY,
-        variables: { query: searchQuery },
+        variables: {
+          query: searchQuery,
+          supportedTypes: [PostType.Article, PostType.Share, PostType.Welcome],
```
rebelchris, line 51: Suggestion can we add one export somewhere for supportedTypes?
It makes it way easier to maintain and less error prone?

rebelchris (approved):
One more optimisation request to make supportedTypes uniform as we always use it the same now.

packages/shared/src/components/BookmarkFeedLayout.tsx
capJavert: @rebelchris added it here 44bcbc9ed0e1ac6dc167d432d79b7ba35c97d36d, not 100% on naming though but I think just `supportedTypes` would not work because we do not support `Welcome` on regular feeds.

rebelchris: Ah you are right, in either way we could have two maybe then?

capJavert: We already have `SUPPORTED_TYPES` that is just a string that we concat into queries as default.

capJavert: @rebelchris forgot to tag, if above is good I would merge because this PR is needed for pulse animation PR.

rebelchris: Yep, all good :)

packages/shared/src/components/cards/WelcomePostCardHeader.tsx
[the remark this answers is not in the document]
capJavert: Merging, it can be adjusted when we get update from design.

PR #1796, opened 2023-05-11
feat(checklist): add welcome post steps

capJavert:
- Edit welcome post step
- Comment on welcome post step

TODO:

- [ ] check if below caveats are ok with product
- [ ] check `openNewComment` usage if `squad checklist` origin is valid, [thread here](https://dailydotdev.slack.com/archives/C0506KU4RH9/p1683711706803829)
- [ ] remove TODO comments from the changes

Currently with caveats:
- when pressing customize on welcome post edit step I just open the edit post modal
  - opening article modal is tricky
  - navigation to welcome post page is not possible due to lazy modals being closed automatically on navigation so if I try to router.push before opening lazy modal it will just close it when navigation completes
- when pressing "Say hi" on squad first comment step I open article page and toggle comment box with `?comment` query param
  - opening article modal is tricky 

WT-1313 #done


capJavert:
@rebelchris @sshanzel This technically works and enables completion of all the steps but not ideal. Would appreciate ideas on how to tackle if any.. 🙏

packages/shared/src/components/checklist/SquadFirstCommentChecklistStep.tsx
```diff
@@ -0,0 +1,32 @@
...
+  const welcomePost = useFindSquadWelcomePost(squad);
+
+  return (
+    <ChecklistStep {...props}>
+      <Button
+        className="btn-primary"
+        disabled={!welcomePost}
+        onClick={() => {
+          // TODO WT-1313-squad-welcome-post-checklist-steps we might need to open article modal here
+          router.push(`/posts/${welcomePost.id}?comment`);
```
rebelchris, line 23: Lee actually has the same caveat now from BE perspective my suggestion was:

`posts/{id}?comment={plaintext}`

So maybe align on who will tackle which part :)
Great to have a generic solution like that.

packages/shared/src/components/post/PostEngagements.tsx
```diff
@@ -82,6 +84,31 @@ function PostEngagements({
     }
   }, [shouldOnboardAuthor]);
 
+  const hasCommentToggleQueryParam =
+    typeof router.query.comment !== 'undefined';
+
+  useEffect(() => {
```
rebelchris, line 90: Just this part I think we have a hook now for interacting with query params.
Would prefer to use that to keep everything inline.

And please check with @sshanzel on the actual implementation.
Really lke the suggestion here though.

rebelchris (changes requested):
Nice almost there, like the idea :)

packages/shared/src/components/post/PostEngagements.tsx
capJavert: Refactored inside `usePostComment` here bea2d8438dc0de12aad077cbb8f5863018d5f225

packages/shared/src/components/checklist/SquadEditWelcomePostChecklistStep.tsx
```diff
@@ -0,0 +1,43 @@
...
+  });
+
+  return (
+    <ChecklistStep {...props}>
+      <Button
+        className="btn-primary"
+        disabled={!welcomePost}
+        icon={<EditIcon />}
+        onClick={() => {
+          // TODO WT-1313-squad-welcome-post-checklist-steps we might need to open article modal here
```
rebelchris, line 29:
```suggestion
```

packages/shared/src/components/checklist/SquadFirstCommentChecklistStep.tsx
```diff
@@ -0,0 +1,32 @@
...
+  const router = useRouter();
+  const welcomePost = useFindSquadWelcomePost(squad);
+
+  return (
+    <ChecklistStep {...props}>
+      <Button
+        className="btn-primary"
+        disabled={!welcomePost}
+        onClick={() => {
+          // TODO WT-1313-squad-welcome-post-checklist-steps we might need to open article modal here
```
rebelchris, line 22:
```suggestion
```

packages/shared/src/hooks/useFindSquadWelcomePost.ts
```diff
@@ -0,0 +1,34 @@
...
+  const client = useQueryClient();
+  const { user } = useContext(AuthContext);
+
+  const squadFeed = client.getQueryData<InfiniteData<FeedData>>([
+    'sourceFeed',
+    user?.id ?? 'anonymous',
+    Object.values({
+      source: squadId,
+      ranking: 'TIME',
+      supportedTypes: [PostType.Article, PostType.Share, PostType.Welcome],
```
rebelchris, line 19: We have a global const for this now right?

packages/shared/src/hooks/usePostComment.ts
```diff
@@ -257,6 +266,30 @@ export const usePostComment = (
...
+  const hasCommentToggleQueryParam = typeof router.query.comment === 'string';
+
+  useEffect(() => {
+    if (!hasCommentToggleQueryParam || parentComment) {
+      return;
+    }
+
+    const { comment, ...queryWithoutComment } = router.query;
+
+    // TODO WT-1313-squad-welcome-post-checklist-steps correct origin
```
rebelchris, line 278: @sshanzel Would you mind having a look here

rebelchris (changes requested):
Some initial comments

packages/shared/src/components/checklist/SquadEditWelcomePostChecklistStep.tsx
rebelchris: Sab approved we can do as proposed so [not](https://dailydotdev.slack.com/archives/C0506KU4RH9/p1683888116686299) the article modal.

packages/shared/src/components/checklist/SquadFirstCommentChecklistStep.tsx
rebelchris: Sab approved we can do as proposed so [not](https://dailydotdev.slack.com/archives/C0506KU4RH9/p1683888116686299) the article modal.

packages/shared/src/hooks/usePostComment.ts
rebelchris: So what's the TODO here still?

rebelchris (changes requested):
Looks pretty good, just one more question about the TODO that was added?

packages/shared/src/hooks/usePostComment.ts
sshanzel: Oh, made some changes from the code way above and forgot to remove this. Will update quickly

sshanzel: Have tested locally and works as intended 🚀

rebelchris (approved):
Let's give it a test spin ✅

PR #1798, opened 2023-05-11
feat: edit welcome post

sshanzel:
- General note: everything here will most likely be removed due to the fact that we will create a dedicated page for updating a post and the revamp that we will be having on making new comments.
- Refactored some components that can be reused.
- Found an issue with `requestProtocol` when the request key string's length is greater than the usual amounts (sample here is 1k characters). We might want to monitor it or find a better way to communicate the result of requests to our query cache when in Companion.
- Edit post modal introduction.
- Refactored the comment box to accommodate edit post UI

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done

### DIFF
--- a/packages/shared/src/components/fields/MarkdownPreview.tsx
+++ b/packages/shared/src/components/fields/MarkdownPreview.tsx
@@ -1,0 +1,47 @@
+import React, { ReactElement } from 'react';
+import { useQuery } from 'react-query';
+import { useRequestProtocol } from '../../hooks/useRequestProtocol';
+import { graphqlUrl } from '../../lib/config';
+import { PREVIEW_COMMENT_MUTATION } from '../../graphql/comments';
+import Markdown from '../Markdown';
+
+export interface MarkdownPreviewProps {
+  input: string;
+  sourceId: string;
+  parentSelector: () => HTMLElement;
+  enabled?: boolean;
+}
+
+interface QueryResult {
+  preview: string;
+}
+
+function MarkdownPreview({
+  input,
+  sourceId,
+  parentSelector,
+  enabled = true,
+}: MarkdownPreviewProps): ReactElement {
+  const { requestMethod, isCompanion } = useRequestProtocol();
+  const query = ['markdown_preview', input.replace('\n', '')];
+  const { data: previewContent } = useQuery<QueryResult>(
+    query,
+    () =>
+      requestMethod(
+        graphqlUrl,
+        PREVIEW_COMMENT_MUTATION,
+        { content: input, sourceId },
+        { requestKey: isCompanion ? JSON.stringify(query) : undefined },
+      ),
+    { enabled: input?.length > 0 && enabled },
+  );
+
+  return (
+    <Markdown
+      content={previewContent?.preview}
+      appendTooltipTo={parentSelector}
+    />
+  );
+}
+
+export default MarkdownPreview;
--- a/packages/shared/src/components/fields/MarkdownPreview.tsx
+++ b/packages/shared/src/components/fields/MarkdownPreview.tsx
@@ -23,7 +23,7 @@ function MarkdownPreview({
   enabled = true,
 }: MarkdownPreviewProps): ReactElement {
   const { requestMethod, isCompanion } = useRequestProtocol();
-  const query = ['markdown_preview', input.replace('\n', '')];
+  const query = ['markdown_preview', input];
   const { data: previewContent } = useQuery<QueryResult>(
     query,
     () =>
--- a/packages/shared/src/components/modals/CommentBox.tsx
+++ b/packages/shared/src/components/modals/CommentBox.tsx
@@ -36,9 +36,10 @@ export interface CommentBoxProps {
   errorMessage?: string;
   sendingComment?: boolean;
   parentSelector?: () => HTMLElement;
-  sendComment: (event: MouseEvent | KeyboardEvent) => Promise<void>;
+  sendComment: (event: MouseEvent | KeyboardEvent) => Promise<unknown>;
   onInput?: (value: string) => unknown;
   useUserMentionOptions: UseUserMentionOptions;
+  isComment?: boolean;
 }
 
 function CommentBox({
@@ -50,9 +51,16 @@ function CommentBox({
   sendComment,
   parentSelector,
   useUserMentionOptions,
+  isComment = true,
 }: CommentBoxProps): ReactElement {
-  const { authorImage, authorName, publishDate, contentHtml, editContent } =
-    parentComment;
+  const {
+    authorImage,
+    authorName,
+    publishDate,
+    contentHtml,
+    editContent,
+    post,
+  } = parentComment;
   const { user } = useContext(AuthContext);
   const {
     onMentionClick,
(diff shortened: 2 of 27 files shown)


packages/shared/src/components/modals/EditWelcomePostModal.tsx
```diff
@@ -0,0 +1,126 @@
...
+  const [errorMessage, setErrorMessage] = useState<string>(null);
+  const { mutateAsync: updatePost, isLoading } = useMutation(editPost, {
+    onSuccess: (updatedPost) => {
+      if (!updatedPost) return;
+
+      const key = getPostByIdKey(post.id);
+      client.setQueryData<PostData>(key, (data) =>
+        !data
+          ? { post: updatedPost }
+          : { post: { ...data.post, ...updatedPost } },
```
capJavert, line 48: Does edit mutation not return complete entity after db update? So we do not need this spread and we can just use `updatedPost`?

packages/shared/src/components/modals/CommentBox.tsx
```diff
@@ -79,7 +87,8 @@ function CommentBox({
     }
 
     element.setAttribute('data-min-height', element.offsetHeight.toString());
-    // @NOTE see https://dailydotdev.atlassian.net/l/cp/dK9h1zoM
+    fixHeight(element);
+    // needs to happen on load
```
capJavert, line 91: What does the above `element.setSelectionRange` actually do? I did not see any difference when removing practically the whole part:
```
if (value) {
    element.value = value;
    element.setSelectionRange(element.value.length, element.value.length);
}
```

capJavert (commented):
Massive job, works well already, some comments.

packages/shared/src/components/modals/CommentBox.tsx
capJavert: Just looking for ways to split up and simplify amount of work we need to do in this `useEffect`.

packages/shared/src/components/modals/EditWelcomePostModal.tsx
sshanzel: They are two different queries, there could be a discrepancy with their properties defined on their own.

packages/shared/src/components/modals/CommentBox.tsx
sshanzel: > What does the above element.setSelectionRange actually do?

I think it is to focus on the very last part of the content.

sshanzel: > Just looking for ways to split up and simplify amount of work we need to do in this useEffect.

This is going to be deleted/refactored on v6/v7 anyway, but I'm open to suggestions 🚀

packages/shared/src/components/fields/MarkdownPreview.tsx
```diff
@@ -0,0 +1,47 @@
...
+}
+
+function MarkdownPreview({
+  input,
+  sourceId,
+  parentSelector,
+  enabled = true,
+}: MarkdownPreviewProps): ReactElement {
+  const { requestMethod, isCompanion } = useRequestProtocol();
+  const query = ['markdown_preview', input.replace('\n', '')];
```
rebelchris, line 26: You probably don't need the input here then?
My suggestion would be to name it markdown_preview, userID and just invalidate always? (or every x second)

@capJavert what would you say?

packages/shared/src/components/modals/ContentWriteTab.tsx
```diff
@@ -0,0 +1,58 @@
...
+import { useUserMention } from '../../hooks/useUserMention';
+import { Post } from '../../graphql/posts';
+
+interface ContentWriteTabProps
+  extends Omit<CommentBoxProps, 'useUserMentionOptions'> {
+  post: Post;
+  tabName: string;
+}
+
+function ContentWriteTab({
```
rebelchris, line 18: Fine for now, but in v6 we might consider doing sub folders, let me know what you guys think. Just more structure only.

packages/shared/src/components/modals/EditWelcomePostModal.tsx
```diff
@@ -0,0 +1,126 @@
...
+  okButton: {
+    className: 'btn-primary-ketchup',
+    title: 'Discard',
+  },
+  cancelButton: {
+    title: 'Stay',
+  },
+};
+
+function EditWelcomePostModal({
```
rebelchris, line 30: Eventually in v6 you probably suggest to make this generic right?
For now its fine like this.

rebelchris (approved):
Wow massive, no clear blockers from my side.
Think we have to carefully test out v5 together again, just general questions non blocking at this point.

capJavert:
> @capJavert For useMutation do they also suggest changing the handlers? Can't seem to find anything or was it just for useQuery?

@rebelchris just for `useQuery`, handlers are just fine and welcome for mutations.

packages/shared/src/components/modals/ContentWriteTab.tsx
sshanzel: Agree with that. We will need to be more organized, I will create a `tabs` folder now in this `modals` folder.

packages/shared/src/components/modals/CommentBox.tsx
capJavert: We don't have to delve on it too much just because it will be updated in v6/v7. My recommendation for this kind of cases would be to split up `useEffect` in multiple smaller effects that do specific things and then also have less or no dependencies in dependency array.

> It is to focus on the very last part of the content on load. QoL kind of thing.

For me it was focusing without that code also, so probably worth checking if actually needed down the line.

packages/shared/src/components/modals/EditWelcomePostModal.tsx
capJavert: @rebelchris just for useQuery, handlers are just fine and welcome for mutations.

packages/shared/src/components/modals/EditWelcomePostModal.tsx
sshanzel: In v6 we will have a whole new page, right?

packages/shared/src/components/fields/MarkdownPreview.tsx
capJavert: To answer directly, by leaving `input` in query key you get invalidate for free when input changes which is a great feature of react-query.

capJavert (approved):
🚀

packages/shared/src/components/fields/MarkdownPreview.tsx
sshanzel: Right. Also might be a good time to discuss how we handle queries in Companion, will explain in the future how it works but probably something to write down to be remembered.

packages/shared/src/components/modals/CommentBox.tsx
sshanzel: Oh, I see. We've had some instances in the past on some devices where it was not working. But happy to take a look at it again once v6/v7 comes.

With regards to smaller `useEffects`, the reason we didn't break it up into smaller pieces, is we had a technical session with tkDodo in the past where he doesn't advise having multiple `useEffects` when it can be clustered into just one due to its performance issues. cc @rebelchris

rebelchris (approved):
Looks great

packages/shared/src/components/modals/CommentBox.tsx
capJavert: I would be interested to see what specific performance issues would be added here, without context on that session I would probably say it had something to do with state setters from useEffect since those of course introduce additional re-renders hence some performance impact. 

We can leave it like this and tackle down the line 🤝

PR #1802, opened 2023-05-12
fix: add v5 tracking events

rebelchris:
- @capJavert as we don't have the new hook yet I still used the old useEffect hack to only fire once.
- The only one missing is `checklist completion` we need to find a way to unique identify it 

## Events

| New | impression (onboarding checklist) | extra: { squad } |
| New | checklist close | extra: { squad } |

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1328 #done

### DIFF
--- a/packages/shared/src/components/checklist/SquadChecklistCard.tsx
+++ b/packages/shared/src/components/checklist/SquadChecklistCard.tsx
@@ -1,4 +1,4 @@
-import React, { ReactElement } from 'react';
+import React, { ReactElement, useContext, useEffect } from 'react';
 import { ChecklistCard } from './ChecklistCard';
 import { useSquadChecklist } from '../../hooks/useSquadChecklist';
 import { Squad } from '../../graphql/sources';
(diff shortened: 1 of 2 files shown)


capJavert (commented):
Reposting my thoughts from slack to keep it in single place:
> From frontend perspective we could send event when useChecklist.isDone  becomes true .. but due to our checklist being dynamic we could possibly send multiple done events for user. Eg. squad posting is set to moderators only, if admin sets it to all members - members will get their checklist updated with new item and make them not done again (which is actually good UX wise since now they can learn to post) - but tracking wise.. double events.
On the backend side I do not know how much flexibility we have to merge that data on the graph/funnel side but we could create a fixed map like we have actionsPerRoleMap for in the frontend that contains actions that each role needs to complete and then query if user has all of the actions to see if their checklist is done.

idoshamun (approved):

PR #1807, opened 2023-05-15
fix: image on the edit welcome post is shouldn't be rounded

TomDowling:
## Changes
Removed the prop to make the image circular. I didn't find anywhere affected from removing this prop

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1361 #done


rebelchris (approved):

sshanzel (changes requested):
We actually have to keep it depending on the type of comment box. This is called `ProfilePicture` but unfortunately being used by `Source` entity on some cases.

packages/shared/src/components/modals/CommentBox.tsx
```diff
@@ -140,12 +140,20 @@ function CommentBox({
...
-          <ProfilePicture
-            size="large"
-            rounded="full"
-            user={{ image: authorImage, username: authorName }}
-            nativeLazyLoading
-          />
+          {isComment ? (
+            <ProfilePicture
+              size="large"
+              rounded="full"
```
rebelchris, line 146: You can just make this line optional based on isComment, but is it the right evaluation? Not sure what it fires on.

packages/shared/src/components/modals/CommentBox.tsx
```diff
@@ -140,12 +140,20 @@ function CommentBox({
...
+              user={{ image: authorImage, username: authorName }}
+              nativeLazyLoading
+            />
+          ) : (
+            <ProfilePicture
+              size="large"
+              user={{ image: authorImage, username: authorName }}
+              nativeLazyLoading
+            />
+          )}
```
sshanzel, line 156: I think their only difference is the property. I think we can just apply the condition from the property itself. WDYT?
```suggestion
          <ProfilePicture
            size="large"
            rounded={isComment && 'full'}
            user={{ image: authorImage, username: authorName }}
            nativeLazyLoading
          />
```

sshanzel (changes requested):
Just one last thing 🥳

sshanzel (approved):
Way better with `undefined`! 🚀 🚀 🚀

rebelchris (approved):
Love it 🤙

PR #1810, opened 2023-05-17
feat: create a reusable component to add a marker to a cards such as the hot marker

TomDowling:
Using the TrendingFlat component i created a new component that accepts a type so change the appearance and text of the label.

I have called the component RaisedLabel, but please if someone has a better suggestion let me know

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1340 #done


packages/shared/src/components/cards/RaisedLabel.tsx
```diff
@@ -0,0 +1,46 @@
...
+import { SimpleTooltip } from '../tooltips/SimpleTooltip';
+
+export function RaisedLabel({
+  trending,
+  listMode,
+  type,
+}: {
+  trending: number;
+  listMode?: boolean;
+  type: 'hot' | 'pinned';
```
rebelchris, line 13: I'd suggest making this a enum in this case just to be more explicit.
See for instance ModalFooter Justify function.

packages/shared/src/components/cards/RaisedLabel.tsx
```diff
@@ -0,0 +1,46 @@
...
+}): ReactElement {
+  const description = `${trending} devs read it last hour`;
```
rebelchris, line 15: This will make it very specific to trending though right?

packages/shared/src/components/cards/RaisedLabel.tsx
```diff
@@ -0,0 +1,46 @@
...
+        'absolute flex items-start',
+        listMode ? 'top-0 right-full mt-5' : 'left-0 bottom-full ml-5 h-5',
+      )}
+    >
+      <SimpleTooltip content={description}>
+        <div
+          className={classNames(
+            'flex items-center px-1',
+            styles.flag,
+            type === 'hot' ? 'bg-theme-status-error' : 'bg-theme-bg-bun',
```
rebelchris, line 28: Maybe we can have a typeToContainer function (as seen in icons for instance)

packages/shared/src/components/cards/RaisedLabel.tsx
```diff
@@ -0,0 +1,46 @@
...
+            listMode
+              ? 'h-5 w-full justify-center mouse:translate-x-9 rounded-l'
+              : 'h-full flex-col mouse:translate-y-4 rounded-t',
+          )}
+        >
+          <span className="font-bold text-white uppercase typo-caption2">
+            {type === 'hot' ? 'Hot' : 'Pinned'}
```
rebelchris, line 35: This I would also use something like typeToContent or typeToText

packages/shared/src/components/cards/RaisedLabel.tsx
```diff
@@ -0,0 +1,46 @@
...
+      className={classNames(
+        'absolute flex items-start',
+        listMode ? 'top-0 right-full mt-5' : 'left-0 bottom-full ml-5 h-5',
+      )}
+    >
+      <SimpleTooltip content={description}>
+        <div
+          className={classNames(
+            'flex items-center px-1',
+            styles.flag,
```
rebelchris, line 27: Just a question can we solve this with Tailwind possibly? 
If not it's fine, but prefer to eventually get rid of these.

rebelchris (changes requested):
Nice!

Think we can make it a bit more explicit and dynamic.
Also you should be able to remove `TrendingFlag` now right?

packages/shared/src/components/cards/RaisedLabel.tsx
sshanzel: I agree, I think we can make the description come from the prop.

sshanzel:
> Think we can make it a bit more explicit and dynamic.

Agree with this 🚀 

> Also you should be able to remove TrendingFlag now right?

I think we can keep it but be updated to implement the introduced reusable component. So we can keep the description details in one place as that component will be used by both PostCard andPostList, iirc.

packages/shared/src/components/cards/RaisedLabel.tsx
```diff
@@ -0,0 +1,46 @@
+import React, { ReactElement } from 'react';
+import classNames from 'classnames';
+import styles from './Card.module.css';
+import { SimpleTooltip } from '../tooltips/SimpleTooltip';
+
+export function RaisedLabel({
+  trending,
+  listMode,
+  type,
+}: {
```
sshanzel, line 10: Let's introduce an interface in this same file for these properties, just to tidy it up a bit.

sshanzel (commented):
Looking great! Got nothing more to add from Chris' comments 🚢 🚢 🚢

packages/shared/src/components/cards/RaisedLabel.tsx
TomDowling: After investigating we are using the css module as of the transform and transition properties we are using are not support by tailwind

packages/shared/src/components/cards/PostList.tsx
```diff
@@ -84,7 +84,7 @@ export const PostList = forwardRef(function PostList(
     return (
       <div className={`relative ${styles.cardContainer}`}>
         {card}
-        <TrendingFlag trending={trending} listMode />
+        <TrendingFlag trending={trending} />
```
sshanzel, line 87: I think this should be in list mode.

```suggestion
        <TrendingFlag trending={trending} listMode />
```

packages/shared/src/components/cards/RaisedLabel.tsx
```diff
@@ -0,0 +1,57 @@
+import React, { ReactElement } from 'react';
+import classNames from 'classnames';
+import styles from './Card.module.css';
+import { SimpleTooltip } from '../tooltips/SimpleTooltip';
+
+export enum RaisedLabelType {
+  hot = 'Hot',
+  pinned = 'Pinned',
```
sshanzel, line 8: For enums, we keep it using PascalCase, currently not part of the linter but hoping to be added soon. Then just probably update the use-cases.

```suggestion
  Hot = 'Hot',
  Pinned = 'Pinned',
```

packages/shared/src/components/cards/TrendingFlag.tsx
```diff
@@ -1,43 +1,11 @@
...
 
-export default function TrendingFlag({
-  trending,
-  listMode,
-}: {
-  trending: number;
-  listMode?: boolean;
-}): ReactElement {
-  const description = `${trending} devs read it last hour`;
+export function TrendingFlag({ trending }: { trending: number }): ReactElement {
```
sshanzel, line 4: I think we must keep the `listMode` to support both list and card layouts.
```suggestion
interface TrendingFlagProps {
  trending: number;
  listMode?: boolean;
}

export function TrendingFlag({ trending, listMode }: TrendingFlagProps): ReactElement {
```

packages/shared/src/components/cards/TrendingFlag.tsx
```diff
@@ -1,43 +1,11 @@
...
-      {!listMode && (
-        <span className="mouse:hidden ml-2 typo-footnote text-theme-label-tertiary">
-          {description}
-        </span>
-      )}
-    </div>
+    <RaisedLabel
+      type={RaisedLabelType.hot}
+      description={`${trending} devs read it last hour`}
+    />
```
sshanzel, line 9: Just including the listMode property 🥳 

```suggestion
    <RaisedLabel
      listMode={listMode}
      type={RaisedLabelType.hot}
      description={`${trending} devs read it last hour`}
    />
```

sshanzel (changes requested):
Just these last few things we should be good to go 🚀

packages/shared/src/components/cards/PostList.tsx
TomDowling: This TrendingFlag wasn't in listMode before i worked on it so just checking on this as seems this would have a change somewhere if i'm not trying to change any ui just add a new component? @sshanzel

sshanzel (approved):
Looks amazing to me! I already branched out from this branch to use the component you've created 🥳 

Updating the usages of the enum should fix the build 🚀

rebelchris (approved):
Looks great

PR #1812, opened 2023-05-18
feat: pin or unpin a post

sshanzel:
- Pin or unpin a post
- Refactor implementations to using `ConditionalWrapper` for flags.
- Created a reusable function to be utilize on the prop for ConditionalWrapper.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1340 #done


packages/shared/src/components/PostOptionsMenu.tsx
```diff
@@ -25,6 +25,8 @@ import { ShareBookmarkProps } from './post/PostActions';
 import BookmarkIcon from './icons/Bookmark';
 import { AnalyticsEvent, Origin } from '../lib/analytics';
 import { usePostMenuActions } from '../hooks/usePostMenuActions';
+import PinIcon from './icons/Pin';
```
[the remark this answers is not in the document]
sshanzel: I 100% agree with that. Let me create one for the icons folder.

TomDowling (approved):
Looks good for the icon stuff, but of course i'd wait for others approval fist still :)

packages/shared/src/components/PostOptionsMenu.tsx
```diff
@@ -99,9 +109,26 @@ export default function PostOptionsMenu({
...
+        }));
+      }
+
+      await client.invalidateQueries(feedQueryKey);
+      displayToast(
+        post.pinnedAt
+          ? 'Your post has been unpinned'
+          : '📌 Your post has been pinned',
+      );
+    },
```
rebelchris, line 131: Is this not something we can handle within the hook?
It seems like the wrong place to handle it per implementation.

packages/shared/src/graphql/feed.ts
```diff
@@ -170,7 +173,7 @@ export const SOURCE_FEED_QUERY = gql`
       ...FeedPostConnection
     }
   }
-  ${FEED_POST_CONNECTION_FRAGMENT}
+  ${getFeedPostFragment('pinnedAt')}
```
rebelchris, line 176: Why do you add them conditional? I think we should just clean them into a FeedPost no?

packages/shared/src/hooks/usePostMenuActions.ts
```diff
@@ -61,10 +64,20 @@ export const usePostMenuActions = ({
       SourcePermissions.PostDelete,
     );
 
+  const canPin = post?.source.currentMember?.permissions?.includes(
+    SourcePermissions.PostPin,
+  );
+
+  const { mutateAsync: onPinPost } = useMutation(
+    () => updatePinnedPost({ id: post.id, pinned: !post.pinnedAt }),
+    { onSuccess: onPinSuccessful },
```
rebelchris, line 73: I would handle the mutation success here actually, the hook seems like the perfect place to have this logic.

packages/shared/src/components/cards/PostList.tsx
```diff
@@ -39,54 +39,52 @@ export const PostList = forwardRef(function PostList(
...
-  const card = (
-    <ListCard
-      {...props}
-      className={getPostClassNames(post, className)}
-      ref={ref}
+  return (
+    <ConditionalWrapper
+      condition={!!trending}
+      wrapper={(component) => getTrendingFlag(component, trending, true)}
     >
```
rebelchris, line 46: I'm not really sure this is the right way.
Multiple places doing the same thing, is there not a way we can uniform a cards raisedLabel based on props?

if it has `pinnedAt` show pinned, if it has trending -> show that, else nothing?
Keep in mind any post can be pinned.

rebelchris (changes requested):
Looks great already.
Would love to see if we can make the card raised label a bit more generic for future proofing.

packages/shared/src/components/PostOptionsMenu.tsx
sshanzel: Have actually done that initially. But then I saw `PostOptionsMenu` handling all the toast and invalidations so I've thought of keeping everything in the same place. WDYT?

packages/shared/src/components/cards/PostList.tsx
sshanzel: Let me check how we can uniform it.

sshanzel:
@rebelchris have refactored them to use the same container wrapper. Feel free to test the pin/unpin function in our preview deployment.

https://preview.app.daily.dev/squads/dailydev2

rebelchris (changes requested):
Minor remarks, almost there.

packages/shared/src/components/cards/FeedItemContainer.tsx
```diff
@@ -0,0 +1,47 @@
...
+}
+
+function FeedItemContainer(
+  { flagProps, ...props }: FeedItemContainerProps,
+  ref?: Ref<HTMLElement>,
+): ReactElement {
+  const { listMode, pinnedAt, trending } = flagProps;
+  const Component = listMode ? ListCard : Card;
+  const type = pinnedAt ? RaisedLabelType.Pinned : RaisedLabelType.Hot;
+  const description =
```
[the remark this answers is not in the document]
sshanzel: Since we only need this in one place now, there's no need to create a component. Hence, applied here manually as this will be the container of all Feed items.

Will remove unused components.

packages/shared/src/components/cards/FlaggedContainer.tsx
```diff
@@ -0,0 +1,26 @@
...
+  children: ReactNode,
+  props: Partial<RaisedLabelProps> = {},
+): ReactElement => (
+  <div className={`relative ${styles.cardContainer}`}>
+    {children}
+    <RaisedLabel type={RaisedLabelType.Pinned} {...props} />
+  </div>
+);
+
+const FlaggedContainer = ({
```
rebelchris, line 20: Where do you use this?
I think you can just delete this whole file actually.

rebelchris (changes requested):

packages/shared/src/components/cards/FlaggedContainer.tsx
sshanzel: We are still using the `getFlaggedContainer`. Not anywhere atm but kept it there if anyone needs it. But if you want we can definitely remove this one.

packages/shared/src/components/cards/FlaggedContainer.tsx
```diff
@@ -0,0 +1,37 @@
+import React, { ReactElement, ReactNode } from 'react';
+import styles from './Card.module.css';
+import { RaisedLabel, RaisedLabelProps, RaisedLabelType } from './RaisedLabel';
+
+interface FlaggedContainerProps {
+  children: ReactNode;
+  labelProps: RaisedLabelProps;
+}
+
+export const getFlaggedContainer = (
```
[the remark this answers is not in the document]
sshanzel: This one is used.

rebelchris (approved):
Looks great!

PR #1818, opened 2023-05-22
feat: hiding the read post button on welcome post

TomDowling:
## Changes
On mobile the read article button was showing on the scrolling modal header, so I have removed this is the post.type === 'welcome

- Hiding the read button for welcome post on mobile modal

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1375 #done


packages/shared/src/components/post/PostModalActions.tsx
```diff
@@ -79,7 +79,7 @@ export function PostModalActions({
 
   return (
     <Container {...props} className={classNames('gap-2', className)}>
-      {onReadArticle && (
+      {onReadArticle && post?.type !== 'welcome' && (
```
rebelchris, line 82: Not sure this will be the most solid fix.
Is there anyway we can not add onReadArticle if the post is welcome?

Asking this because soon we'll introduce more types that don't have clicks, so we need to have a more generic way down the line.

rebelchris (changes requested):
A suggestion for another approach.

packages/shared/src/components/post/PostModalActions.tsx
TomDowling: I agree, this isn't very scalable, I have made changes to use the onReadArticle to try and make it more scalable now

packages/shared/src/graphql/posts.ts
```diff
@@ -28,6 +28,9 @@ export enum PostType {
   Welcome = 'welcome',
 }
 
+export type InternalReadType = PostType[];
```
rebelchris, line 31: Export not needed here, and can actually also just type it directly as it's just the one type

packages/shared/src/components/post/PostModalActions.tsx
```diff
@@ -23,7 +23,7 @@ import SettingsContext from '../../contexts/SettingsContext';
 import { Origin } from '../../lib/analytics';
 
 export interface PostModalActionsProps extends ShareBookmarkProps {
-  post: Post;
+  post?: Post;
```
rebelchris, line 26: I don't think these changes are needed anymore right?

rebelchris (changes requested):
Don't think you need the post changes anymore.

packages/shared/src/components/post/PostNavigation.tsx
```diff
@@ -29,6 +30,7 @@ export interface PostNavigationProps extends PostActions {
   onNextPost?: () => unknown;
   className?: PostNavigationClassName;
   children?: ReactNode;
+  post: Post;
```
rebelchris, line 33: Can you double check these changes in this file?
think it's already in the props, as far as i can tell its not needed

rebelchris (changes requested):

packages/shared/src/components/post/PostNavigation.tsx
TomDowling: Apologies, I was getting the two files mixed up. Cleaning it up now

rebelchris (approved):
Perfect thanks Tom 👍

PR #1820, opened 2023-05-22
feat: markdown input

sshanzel:
## Changes
The current one we have for the comment box is very specific only to mentions only and the processes won't be re-used to any of the new commands. As we can see now, the mention is just reusing the functions provided by what we have created in `textarea.ts`. Once we hit v7, we will **delete** the old hook `useUserMention`, and the parts along with it in the `elements.ts` file.

As we add more and more features with Markdown editing, we've created the building blocks for adding new commands. We were able to reuse the generic function for textarea with both Link and Mention features in the new implementation. It would easily allow the implementation of bold, italic, etc. by adding literally a simple function like shown below:

```
const onBoldCommand = async () => {
    const getUsernameReplacement = (_, { word }) => ({ replacement: `__${word}__` });
    await replaceWord(textarea, getUsernameReplacement, setInput);
  };

```

This is also in preparation for v7 where we will revamp the comment box where we will have a bit different layout than writing Freeform posts but the functionality will be the same. The `useMarkdownInput` hook should be easy to use as we did on `MarkdownInput`.

imho the new [`MarkdownInput`](https://github.com/dailydotdev/apps/blob/907b4ea5301c30afb34274a2967ee87492d89784/packages/shared/src/components/fields/MarkdownInput/index.tsx) is more maintainable than our legacy [`CommentBox`](https://github.com/dailydotdev/apps/blob/main/packages/shared/src/components/modals/CommentBox.tsx). New introduction of markdown input such as on the comments section would be as easy as how we did on the `MarkdownInput`

This is currently deployed in preview: https://preview.app.daily.dev

Note: upload is for a separate PR (though it would be similar to the code example).

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1337 #done
WT-1349 #done // unfortunately, I had to include this task here so I can test how it is easy to add new commands.


sshanzel:
@idoshamun would love to get your opinion here as this is a bit heavy one.

packages/shared/src/lib/textarea.ts
```diff
@@ -0,0 +1,154 @@
...
+  word?: string;
+  trailingChar?: string;
+}
+
+export type GetReplacementFn = (
+  type: CursorType,
+  word?: GetReplacementOptionalProps,
+) => GetReplacement;
+
+export const replaceWord = async (
```
sshanzel, line 105: This is where it all happens. There should not be too much change here in the future as its purpose is simple, to replace the word based on the parameters. The optional parameter is how to set the cursor position if the default behavior is not what you would expect.

packages/shared/src/components/post/PostEngagements.tsx
```diff
@@ -97,11 +97,12 @@ function PostEngagements({
         actionsClassName="hidden laptop:flex"
         origin={analyticsOrigin}
       />
-      <NewComment
-        user={user}
-        className="my-6"
-        isCommenting={!!parentComment}
-        onNewComment={() => openNewComment('start discussion button')}
+      {/* TODO:: Revert this after testing phase */}
```
idoshamun, line 100: Need to be reverted I guess?

idoshamun (changes requested):
Oh wow that looks like huge amount of work! 👏 
Still requires some tweaks

packages/shared/src/lib/textarea.ts
sshanzel: Agree with that. Let me see how we can break this down into pieces.

packages/shared/src/lib/markdown.ts
```diff
@@ -0,0 +1,50 @@
+import { CursorType, GetReplacementFn, isFalsyOrSpace } from './textarea';
+
+const urlText = 'url';
+const getUrlText = (content = '') => `[${content}](${urlText})`;
```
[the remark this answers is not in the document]
sshanzel: I was using it to have an easier reference of the length when we generate the offset but agree, this suggestion would make it simpler. I'll just create another variable with the exact expected offset with a proper name.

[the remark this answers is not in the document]
idoshamun: I didn't notice that. If this is the case you can keep it :)

sshanzel:
@idoshamun I have made some refactoring, the important part is, instead of doing different calculations on each type of Cursor placement on the core function, I made them just focus on the last character of the replacement made. This should be okay in most cases, then if the consuming end wants a customized placement of the cursor after the replacement, then they will compute it themselves and pass it onto the `getReplacement` function that is available rather than having a hard time on the generic function.

Also, I've made a refactor to separate the functions of each type as they calculate a bit differently on their own and I wanted them to be isolated and easily understandable.

With that, I have also tried to do a class-based approach, not sure if you would prefer it, but I believe there are some benefits, in some aspects. Such as passing the same parameters over and over again, etc., we can easily revert it though since I made it in a single commit. You can check the difference here https://github.com/dailydotdev/apps/pull/1820/commits/d627472c5db259d9981e1e8cd379ddf1e41ab643

packages/shared/src/hooks/input/useMarkdownInput.ts
```diff
@@ -0,0 +1,201 @@
...
+  offset: number[];
+  selected: number;
+  onLinkCommand: () => Promise<unknown>;
+  onMentionCommand: () => Promise<void>;
+  onApplyMention: (username: string) => Promise<void>;
+  checkMention: (position?: number[]) => void;
+  mentions: UserShortProfile[];
+}
+
+export const useMarkdownInput = ({
```
idoshamun, line 54: Please add `useMemo` according to the team guidelines

idoshamun (approved):
Please make sure to add `useMemo` before merging but we're all clear!

sshanzel:
> Please make sure to add useMemo before merging but we're all clear!

I have applied it now on the `mentions` property, but have kept the rest the same as they are already stable coming from a `useState` or `useQuery` 🥳 

The callbacks are left for now as they don't do heavy computations and are not yet referenced on any memoized values/components.

idoshamun:
@sshanzel I'm not sure about it but let's proceed for now. Once Chris/Ante are back let's discuss with them

idoshamun (approved):

sshanzel:
@capJavert @rebelchris Not urgent, just tagging for visibility regarding the useMemo when you get the time 🚢

capJavert:
Mentions `data` are already stable so memoizing them makes no difference. Analogy with `useState` is good since `useQuery` holds that data in state which changes only on refetch.

We do not need to memoize `useState` values because state itself only changes when updated which makes it stable. eg:

```js
  const [input, setInput] = useState(initialContent);
  
  const stableInput = useMemo(() => input, [input]); // this does nothing since `input` is already stable because it is state
```

If we were to add some computation for example if for some reason we used `filter` to include only some mentions then `useMemo` would be needed because `filter` would produce new array each time.

Same thing goes for `data ?? []` expression that was the last example here, `[]` would produce new array on each render as long as `data` is `undefined` so `useMemo` is valid option. ✅ 

We could however for this specific case just use `useQuery` `initialData: () => []` since we are only guarding from initial missing of data.

I would also memoize handlers like `onLinkCommand` with `useCallback` so that any other hook using those can memoize properly.

PR #1824, opened 2023-05-24
feat: create post

sshanzel:
## Changes
- Create a freeform post.
- Edit a freeform post.
- Created a hook to reuse the notification toggle functionality as it seems to be more frequent now.
- Updated the ImageInput to be "closeable".
- Fixed some typo
- Created a new hook useSquad which returns an `isForbidden` property out of the box and avoids the unnecessary state. (tested and still works as intended).
- Refactored the unnecessary usage of state for firing tracking events.

**Note:** the freeform post modal and card have their own dedicated tasks, so there are rough edges when viewing the post page of the newly created freeform post but it is for the sake of testing.

You might wonder if some parts are missing like the image uploads on the content, and saving on the cache. Those things can be worked on in parallel and have their own tasks too. Once this is merged, those areas should be unblocked to anyone who wants to work with v6.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done


packages/shared/src/components/fields/ImageInput.tsx
```diff
@@ -22,17 +24,21 @@ interface ClassName {
...
-  alwayShowHover?: boolean;
+  enableHover?: boolean;
+  closeable?: boolean;
+  alwaysShowHover?: boolean;
+  children?: ReactNode;
+  fileSizeLimitMB?: number;
 }
 
-const TWO_MEGABYTES = 2 * 1024 * 1024;
+const MEGABYTE = 2 * 1024 * 1024;
```
idoshamun, line 41:
```suggestion
const MEGABYTE = 1024 * 1024;
```

sshanzel:
Have added the edit page and should be working as intended now too.

packages/shared/src/components/post/SquadPostContent.tsx
```diff
@@ -23,6 +23,7 @@ import { SourcePermissions, Squad } from '../../graphql/sources';
 import EditIcon from '../icons/Edit';
 
 const ContentMap = {
+  [PostType.Freeform]: WelcomePostContent,
```
idoshamun, line 26: What does it mean? What is `WelcomePostContent`?

packages/shared/src/graphql/posts.ts
```diff
@@ -487,3 +493,43 @@ interface UpdatePinnedProps {
...
+  ) {
+    createFreeformPost(
+      sourceId: $sourceId
+      title: $title
+      content: $content
+      image: $image
+    ) {
+      ...SharedPostInfo
+      trending
+      views
```
idoshamun, line 512: These are not needed right? (trending and views)

packages/shared/src/graphql/posts.ts
```diff
@@ -487,3 +493,43 @@ interface UpdatePinnedProps {
...
+      contentHtml
+      source {
+        ...SourceBaseInfo
+      }
+      description
+      summary
+      toc {
+        text
+        id
+      }
```
idoshamun, line 523: Also these

idoshamun (changes requested):
Please see my comments

packages/shared/src/components/post/SquadPostContent.tsx
sshanzel: This is just a placeholder to test it, but basically, we have a dedicated ticket to introduce a `PostContent` dedicated to FreeformPost.

idoshamun (approved):

PR #1826, opened 2023-05-25
feat: squads v6

sshanzel:

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done


packages/shared/src/components/post/freeform/write/WritePage.tsx
```diff
@@ -0,0 +1,39 @@
...
+  squad,
+  isLoading,
+  isForbidden,
+  ...props
+}: WritePageProps): ReactElement {
+  const isVerified = verifyPermission(squad, SourcePermissions.Post);
+
+  if (isLoading) return <WriteFreeFormSkeleton isEdit={isEdit} />;
+
+  if (isForbidden || !isVerified || !squad) return <Unauthorized />;
```
capJavert, line 31: Looks to me that we could just return `Unauthorized` first without waiting for loading since all info should be there squad, post, user?

So just flipping two ifs.

capJavert (changes requested):
Wow, lot of changes, tried to go through it and comment if anything looked odd.

packages/shared/src/components/post/freeform/write/WritePage.tsx
sshanzel: We need to show the loading first, we don't fetch the static props for this page, so we won't have the proper information.

packages/shared/src/lib/form.ts
```diff
@@ -9,6 +9,12 @@ export function formToJson<T>(form: HTMLFormElement, initialValue?: T): T {
     if (val.type === 'radio' && !val.checked) {
       return acc;
     }
+    if (val.type === 'file') {
+      if (val.files.length === 0) return { ...acc, [val.name]: null };
+
+      const value = val.files.length === 1 ? val.files[0] : val.files;
+      return { ...acc, [val.name]: value };
```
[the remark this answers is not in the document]
sshanzel: To easily use the form. But probably it would be better off using an Array, so it will be consistent with the usage.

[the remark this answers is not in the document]
sshanzel: Yep, would probably be better to be an array in the long run.

packages/webapp/pages/squads/[handle]/create.tsx
```diff
@@ -67,6 +67,7 @@ function CreatePost(): ReactElement {
       <WritePage
         onSubmitForm={onClickSubmit}
         isLoading={!isReady || !isDraftReady}
+        isPosting={isPosting}
```
Comment (capJavert) [inline packages/webapp/pages/squads/[handle]/create.tsx:71]: it think this should be `isPosting || isSuccess` because for a brief second until page navigation happens "Post" button will become active again and if user clicks on it by mistake we will have double post. Same can be applied to `edit` page.

capJavert (changes requested):
Much better with loader :ok_hand: Edge case after testing.

capJavert (approved):
Let's ship it 🚀

PR #1832, opened 2023-05-26
Wt 1378 stop showing the checklist until information is loaded

TomDowling:
## Changes
Using the new isChecklistReady to render the Checklist card only once the data has been fetched.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done

### DIFF
--- a/packages/shared/src/components/checklist/SquadChecklistCard.tsx
+++ b/packages/shared/src/components/checklist/SquadChecklistCard.tsx
@@ -1,5 +1,4 @@
 import React, { ReactElement, useContext, useEffect } from 'react';
-import classNames from 'classnames';
 import { ChecklistCard } from './ChecklistCard';
 import { useSquadChecklist } from '../../hooks/useSquadChecklist';
 import { Squad } from '../../graphql/sources';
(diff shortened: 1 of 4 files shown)


sshanzel (changes requested):
Looks good with the usage of the new property, though I think it might be better to do an early return of `null` instead, if it is not yet ready. WDYT?

sshanzel (changes requested):
Nice! I checked the preview2 deployment and it seems to have fixed the issue! Just the comment I made earlier, then I think we should be good to go.

packages/shared/src/components/checklist/SquadChecklistCard.tsx
```diff
@@ -48,6 +50,7 @@ const SquadChecklistCard = ({ squad }: { squad: Squad }): ReactElement => {
       description="5 simple steps to squad greatness!"
       steps={steps}
       onRequestClose={onRequestClose}
+      className={classNames(isChecklistReady ? 'flex' : 'hidden')}
```
sshanzel, line 53: I think we can then revert this line.
```suggestion
```

sshanzel (approved):
Looks amazing to me! Just revert the ternary operation in the class as it should not be needed anymore 🥳

sshanzel (approved):
Let's do it! 🥳

PR #1833, opened 2023-05-29
refactor: share post bar to include link to write freeform

sshanzel:
## Changes
- There used to be a single element for displaying the button to choose from reading history, but after getting back to this, it felt difficult to make some changes since the orders are a bit confusing while considering the responsiveness.

Previews:

Screenshot 2023-05-29 at 6 39 27 PM

Screenshot 2023-05-29 at 6 38 37 PM

Screenshot 2023-05-29 at 6 38 58 PM

Screenshot 2023-05-29 at 6 39 15 PM

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1336 #done


packages/shared/src/components/squads/SharePostBar.tsx
```diff
@@ -59,48 +62,59 @@ function SharePostBar({
...
-          buttonSize={sidebarRendered ? ButtonSize.Small : ButtonSize.Medium}
+      <span className="flex relative flex-row items-center w-full">
+        <ProfilePicture
+          className="m-3"
+          user={user}
+          size="large"
+          nativeLazyLoading
+        />
+        <input
+          ref={urlRef}
```
capJavert, line 78: Why do we need `urlRef` now vs before?

packages/shared/src/components/squads/SquadPageHeader.tsx
```diff
@@ -95,16 +98,28 @@ export function SquadPageHeader({
...
-          'absolute bottom-0 w-full translate-y-1/2 px-6 laptopL:px-0 bg-theme-bg-primary',
+          'relative tablet:absolute flex flex-col tablet:flex-row justify-center items-center pt-8 tablet:p-0 bottom-0 w-full tablet:translate-y-1/2 laptopL:px-0 bg-theme-bg-primary laptop:max-w-[41.5rem]',
           shouldShowHighlightPulse && 'highlight-pulse',
-          MAX_WIDTH,
         )}
       >
+        <Divider />
         <SharePostBar
-          className="w-full"
+          className="w-full max-w-[30.25rem]"
```
capJavert, line 107: Can we standardize this size maybe to some rounded value? Non blocking.

capJavert (approved):
Some non blocking comments that caught my eye, looking good 🤘

capJavert (changes requested):
Oh, caught one edge case when posting is limited to moderators, "New post" button still shows. Should be hidden or we need design input on this case?

image

sshanzel:
> Oh, caught one edge case when posting is limited to moderators, "New post" button still shows. Should be hidden or we need design input on this case?

Oh, nice catch. Let me fix that up quickly 🚀

packages/shared/src/components/squads/SharePostBar.tsx
sshanzel: Ah, it seems to be not used anymore, I was working on something have probably decided not to purse, will remove this one 🚀

packages/shared/src/components/squads/SquadPageHeader.tsx
sshanzel: I think this should be acceptable as the equivalent would still be divisible by 4. Though we might want to store these values somewhere so we can standardize the usage. WDYT?z

sshanzel:
Should look like below now 🚢 

Screenshot 2023-05-30 at 11 09 01 AM

packages/shared/src/components/squads/SquadPageHeader.tsx
capJavert: I was thinking of introducing more sizes in our design system like `w-121` class would be equivalent, but not sure if can be standardized inside design.

capJavert (approved):
👍 nice, it does look little cramped to me, like the bar is missing some right padding for text to look good inside of it, won't block it but if you find it off as well you can check with design. 

image

sshanzel:
> 👍 nice, it does look little cramped to me, like the bar is missing some right padding for text to look good inside of it, won't block it but if you find it off as well you can check with design.
> 
> image

Oh, cool. Thanks for the review. I will verify first before merging.

packages/shared/src/components/squads/SquadPageHeader.tsx
sshanzel: Ah, right. That's actually better, I see these values also being used more often now. Let's probably discuss this in the web team meeting and coordinate with the design team what should we label these sizes.

sshanzel:
Should be back to normal now 🥳 

Screenshot 2023-05-30 at 3 59 30 PM

packages/shared/src/components/squads/SquadPageHeader.tsx
capJavert: 🤝 agreed

capJavert:
@sshanzel gg wp 🚀

PR #1837, opened 2023-05-29
feat: ask a prompt for unsaved changes before leaving

sshanzel:
## Changes
- Introduced a hook to manage when the user is trying to leave the page.
- Introduced a hook to re-use some validation logic.
- Make content not required from the graphql schema.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1352 #done


packages/shared/src/hooks/useExitConfirmation.ts
```diff
@@ -0,0 +1,52 @@
...
+  onValidateAction,
+}: ExitProps): UseExitConfirmation {
+  const confirmRef = useRef(true);
+  const router = useRouter();
+
+  useEffect(() => {
+    if (!router.isReady) return null;
+
+    const closeHandler = (e: BeforeUnloadEvent) => {
+      if (!confirmRef.current || onValidateAction()) return;
```
capJavert, line 24: would move ti to variable `shouldAskConfirmation` and then check `if(shouldAskConfirmation)`, just so we know what this does without looking at inner workings of `beforeunload` and `returnValue`.

packages/shared/src/hooks/useExitConfirmation.ts
```diff
@@ -0,0 +1,52 @@
...
+    };
+
+    window.addEventListener('beforeunload', closeHandler);
+    router.events.on('routeChangeStart', routeHandler);
+
+    return () => {
+      window.removeEventListener('beforeunload', closeHandler);
+      router.events.off('routeChangeStart', routeHandler);
+    };
+  }, [confirmRef, onValidateAction, message, router]);
```
capJavert, line 45: refs are not needed for `useEffect` when passed directly since they are not reactive, you should be able to remove `confirmRef` from dependency array.

capJavert (changes requested):
Some minor code style comments.

packages/shared/src/hooks/useExitConfirmation.ts
sshanzel: Makes sense. Let me add that.

capJavert (approved):
Nice, I like this approach 🚀

PR #1838, opened 2023-05-30
refactor: welcome post card to support freeform posts

sshanzel:
## Changes
- Configured line clamping to be up until 9 as required by our Card when there is no title.
- Fixed the Welcome Post modal to handle null cover image.

Reference: https://www.sketch.com/s/2867ec63-0966-4647-b1d7-b6fe696ba251/a/qbbzMDV#Inspect

Previews:

Screenshot 2023-05-30 at 11 41 04 AM

Screenshot 2023-05-30 at 11 36 29 AM

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1336 #done


packages/shared/src/graphql/feed.ts
```diff
@@ -178,7 +178,7 @@ export const SOURCE_FEED_QUERY = gql`
       ...FeedPostConnection
     }
   }
-  ${getFeedPostFragment('pinnedAt')}
+  ${getFeedPostFragment('pinnedAt content')}
```
capJavert, line 181: Not specific to this PR but I was thinking by returning `content` for feed listing which has `4000` char limit we might have considerably bigger responses if there is more freeform posts in the future. @idoshamun FYI interested what you think.

capJavert (approved):
Looking good 🤘 can be merged I think but minor comment for future proofing.

packages/shared/src/graphql/feed.ts
idoshamun: That's a good point! I also think that we don't need for post modal in 99% of the cases. We will need it only for edit mode. Let's log the post modal as something to fix and in this PR let's remove the content for the feed

sshanzel: For the feed, we will need the content (switch to using `contentHtml` now) on the feed to display when there is no cover image on the post. Do we maybe want to make it conditional as to only return it when there is no image?

PR #1841, opened 2023-05-30
fix: draft image

sshanzel:
## Changes
- When a draft image is saved locally, we are displaying it when the user comes back, but since we don't apply the value to the element on load, we should properly check if there is anything available in the draft.
- Also refactored the submit function to simplify and to just receive the params then change it if necessary.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done

### DIFF
--- a/packages/shared/src/components/post/freeform/write/WriteFreeformContent.tsx
+++ b/packages/shared/src/components/post/freeform/write/WriteFreeformContent.tsx
@@ -74,7 +74,7 @@ export function WriteFreeformContent({
     useNotificationToggle();
 
   const getDraftImage = () => {
-    if (!draft.image) return null;
+    if (!draft?.image) return null;
 
     return base64ToFile(draft.image, draft.filename ?? defaultFilename);
   };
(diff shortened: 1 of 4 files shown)


packages/shared/src/components/post/freeform/write/WriteFreeformContent.tsx
```diff
@@ -65,10 +70,18 @@ export function WriteFreeformContent({
   const { shouldShowCta, isEnabled, onToggle, onSubmitted } =
     useNotificationToggle();
 
+  const getImage = () => {
```
capJavert, line 73: `getDraftImage`?

capJavert (changes requested):
Oh, we missed that one with image 👍 

Something broke with submit function though, because now when I click "Post" it does not submit but updates URL with query params and nothing happens. Like the wrong form action is triggered. 

image

capJavert (approved):
Looks good now 🚀

PR #1846, opened 2023-05-31
feat: make sure correct draft key is created after post creation

capJavert:
- squad.id is default identifier in draft inside `useDiscardPost` so this makes sure that correct one is used

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done

### DIFF
--- a/packages/webapp/pages/squads/[handle]/create.tsx
+++ b/packages/webapp/pages/squads/[handle]/create.tsx
@@ -39,7 +39,7 @@ function CreatePost(): ReactElement {
       onAskConfirmation(false);
     },
     onSuccess: async (post) => {
-      const key = generateWritePostKey();
+      const key = generateWritePostKey(squad?.id);
       await deleteCache(key);
       await push(post.commentsPermalink);
     },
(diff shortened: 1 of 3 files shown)


packages/webapp/pages/squads/[handle]/create.tsx
```diff
@@ -39,7 +39,7 @@ function CreatePost(): ReactElement {
       onAskConfirmation(false);
     },
     onSuccess: async (post) => {
-      const key = generateWritePostKey();
+      const key = generateWritePostKey(squad?.id);
```
Comment (sshanzel) [inline packages/webapp/pages/squads/[handle]/create.tsx:42]: I suggest we actually return a `clearDraft` on the `useDiscardPost` and pass it down lang the `updateDraft`, this is to avoid ending up in this same situation in the future.

Comment (capJavert) [reply packages/webapp/pages/squads/[handle]/create.tsx]: Good idea

sshanzel (approved):
Amazing! Thank you for the fix! 🙏

PR #1847, opened 2023-05-31
fix: feed card rendering without tags

sshanzel:
## Changes
- Instead of rendering the raw content, we should use the html markdown content then strip all the tags.
- Would love to hear your thoughts on this.

Quick link: https://preview.app.daily.dev/squads/dailydev2

Before:
image

After:
Screenshot 2023-05-31 at 5 19 16 PM

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done

### DIFF
--- a/packages/shared/src/components/cards/WelcomePostCardFooter.tsx
+++ b/packages/shared/src/components/cards/WelcomePostCardFooter.tsx
@@ -12,7 +12,8 @@ export const WelcomePostCardFooter = ({
   post,
 }: WelcomePostCardFooterProps): ReactElement => {
   const content = useMemo(
-    () => sanitize(post.contentHtml, { ALLOWED_TAGS: [] }),
+    () =>
+      post?.contentHtml ? sanitize(post.contentHtml, { ALLOWED_TAGS: [] }) : '',
     [post?.contentHtml],
   );
 
--- a/packages/shared/src/components/cards/WelcomePostCardFooter.tsx
+++ b/packages/shared/src/components/cards/WelcomePostCardFooter.tsx
@@ -1,4 +1,4 @@
-import React, { ReactElement, useRef } from 'react';
+import React, { ReactElement, useMemo } from 'react';
 import { sanitize } from 'dompurify';
 import { CardImage } from './Card';
 import { Post } from '../../graphql/posts';
@@ -11,7 +11,10 @@ type WelcomePostCardFooterProps = {
 export const WelcomePostCardFooter = ({
   post,
 }: WelcomePostCardFooterProps): ReactElement => {
-  const contentRef = useRef<HTMLDivElement>();
+  const content = useMemo(
+    () => sanitize(post.contentHtml, { ALLOWED_TAGS: [] }),
+    [post?.contentHtml],
+  );
 
   if (post.image) {
     return (
@@ -26,18 +29,7 @@ export const WelcomePostCardFooter = ({
   }
   if (post.content) {
     return (
-      <>
-        <p className="px-2 break-words line-clamp-6 typo-callout">
-          {contentRef?.current?.textContent}
-        </p>
-        <div
-          ref={contentRef}
-          className="hidden"
-          dangerouslySetInnerHTML={{
-            __html: sanitize(post.contentHtml, { ADD_ATTR: ['target'] }),
-          }}
-        />
-      </>
+      <p className="px-2 break-words line-clamp-6 typo-callout">{content}</p>
     );
   }
 
--- a/packages/shared/src/components/cards/WelcomePostCardFooter.tsx
+++ b/packages/shared/src/components/cards/WelcomePostCardFooter.tsx
@@ -24,7 +24,6 @@ export const WelcomePostCardFooter = ({
       />
     );
   }
-
   if (post.content) {
     return (
       <>
--- a/packages/shared/src/components/cards/WelcomePostCardFooter.tsx
+++ b/packages/shared/src/components/cards/WelcomePostCardFooter.tsx
@@ -1,4 +1,5 @@
-import React, { ReactElement } from 'react';
+import React, { ReactElement, useRef } from 'react';
+import { sanitize } from 'dompurify';
 import { CardImage } from './Card';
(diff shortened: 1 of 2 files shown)


packages/shared/src/components/cards/WelcomePostCardFooter.tsx
```diff
@@ -21,12 +24,20 @@ export const WelcomePostCardFooter = ({
...
-      </p>
+      <>
+        <p className="px-2 break-words line-clamp-6 typo-callout">
+          {contentRef?.current?.textContent}
+        </p>
+        <div
+          ref={contentRef}
+          className="hidden"
+          dangerouslySetInnerHTML={{
+            __html: sanitize(post.contentHtml, { ADD_ATTR: ['target'] }),
```
capJavert, line 37: I would memoize this call, I think it does a lot of things so could be expensive when rendering a lot of cards.

packages/shared/src/components/cards/WelcomePostCardFooter.tsx
```diff
@@ -21,12 +24,20 @@ export const WelcomePostCardFooter = ({
...
   }
-
   if (post.content) {
     return (
-      <p className="px-2 break-words line-clamp-6 typo-callout">
-        {post.content}
-      </p>
+      <>
+        <p className="px-2 break-words line-clamp-6 typo-callout">
+          {contentRef?.current?.textContent}
```
capJavert, line 31: The main point is that this is not reactive so it feels clunky. Could you not just use result of `sanitize` directly?

packages/shared/src/graphql/feed.ts
```diff
@@ -178,7 +178,7 @@ export const SOURCE_FEED_QUERY = gql`
       ...FeedPostConnection
     }
   }
-  ${getFeedPostFragment('pinnedAt content')}
+  ${getFeedPostFragment('pinnedAt content contentHtml')}
```
capJavert, line 181: Same thing as https://github.com/dailydotdev/apps/pull/1838#discussion_r1209884396, might simnifically increase our response size.

capJavert (changes requested):
Fast turnaround 👏 added my thoughts below.

packages/shared/src/components/cards/WelcomePostCardFooter.tsx
sshanzel: Good point. Was also worried about the performance difference after this. Memoizing is the way to go.

packages/shared/src/components/cards/WelcomePostCardFooter.tsx
```diff
@@ -1,4 +1,5 @@
-import React, { ReactElement } from 'react';
+import React, { ReactElement, useRef } from 'react';
+import { sanitize } from 'dompurify';
```
capJavert, line 2: Just for a completeness, have you checked for possible alternatives with smaller bundle size? This one is 7kb gzipped which should be fine if it tree shakes.

sshanzel: I believe tree shaking works on this package. Though I was not present when this lib was introduced and had no history about the other options.

packages/shared/src/components/cards/WelcomePostCardFooter.tsx
capJavert: Hmm but would you not be able to use `FORBID_TAGS` from the lib to remove `img`? You are currently using `textContent` for that right?

packages/shared/src/components/cards/WelcomePostCardFooter.tsx
capJavert: Oh sorry, thought it was introduced in this PR. Ignore.

packages/shared/src/components/cards/WelcomePostCardFooter.tsx
sshanzel: Not just the images, but all the noise in the content for example when making a heading which requires a `#` character or a bold text from using `*`.

sshanzel: Haven't heard of `FORBID_TAGS` though. Care to elaborate?

capJavert: It is part of this DOMPurify lib we are using for `sanitize`, check in the examples here https://github.com/cure53/DOMPurify#can-i-configure-dompurify

capJavert: Basically you should be able to specific tags that will be removed from HTML string which is exactly what we do here.

sshanzel: Oh, nice. That is probably a better option to set `ALLOWED_TAGS` to an empty array so there will be no rendered tag. Will try this once I get back on it a bit later.

sshanzel: We basically want no tag since it will affect the card's layout. So `ALLOWED_TAGS` with empty content would probably our best bet.

sshanzel:
@capJavert wow! That turned out pretty well! Thanks for the suggestion! 🚀 

Screenshot 2023-05-31 at 6 27 22 PM

capJavert (approved):
Yep, nice, minor comment.

packages/shared/src/components/cards/WelcomePostCardFooter.tsx
```diff
@@ -10,6 +11,11 @@ type WelcomePostCardFooterProps = {
 export const WelcomePostCardFooter = ({
   post,
 }: WelcomePostCardFooterProps): ReactElement => {
+  const content = useMemo(
+    () => sanitize(post.contentHtml, { ALLOWED_TAGS: [] }),
```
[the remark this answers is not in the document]
sshanzel: Oh, right. Though I have used an empty string literal as we are expecting a string and it might render the word `undefined`, just to be safe.

PR #1853, opened 2023-06-02
feat: lego landing page

capJavert:
- Add lego campaign landing page
- Add generic join page for future referrals

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1412 #done

### DIFF
--- a/packages/webapp/pages/join.tsx
+++ b/packages/webapp/pages/join.tsx
@@ -1,0 +1,64 @@
+import { GET_USERNAME_BY_ID_QUERY } from '@dailydotdev/shared/src/graphql/users';
+import { ReferralCampaignKey } from '@dailydotdev/shared/src/hooks';
+import { graphqlUrl } from '@dailydotdev/shared/src/lib/config';
+import { isDevelopment } from '@dailydotdev/shared/src/lib/constants';
+import request from 'graphql-request';
+import { GetServerSideProps } from 'next';
+import React, { ReactElement } from 'react';
+
+type PageProps = {
+  referrerUsername: string;
+};
+
+const Page = ({ referrerUsername }: PageProps): ReactElement => {
+  return <div>Join {referrerUsername} at daily.dev</div>;
+};
+
+export const getServerSideProps: GetServerSideProps<PageProps> = async ({
+  query,
+  res,
+}) => {
+  const { userid: userId, cid } = query;
+  const campaign = Object.values(ReferralCampaignKey).find(
+    (item) => item === cid,
+  );
+
+  if (!userId || !campaign) {
+    return {
+      redirect: {
+        destination: '/',
+        permanent: false,
+      },
+    };
+  }
+
+  const result = await request<{ user: { username: string } }>(
+    graphqlUrl,
+    GET_USERNAME_BY_ID_QUERY,
+    {
+      id: userId,
+    },
+  );
+
+  if (result) {
+    res.setHeader(
+      'Set-Cookie',
+      [
+        `join_referral=${userId}:${campaign}`,
+        `max-age=${1 * 365 * 24 * 60 * 60}`,
+        isDevelopment ? undefined : 'Secure',
+        `domain=${process.env.NEXT_PUBLIC_DOMAIN}`,
+      ]
+        .filter(Boolean)
+        .join('; '),
+    );
+  }
+
+  return {
+    props: {
+      referrerUsername: result.user.username,
+    },
+  };
+};
+
+export default Page;
--- a/packages/eslint-rules/lib/rules/no-custom-color.js
+++ b/packages/eslint-rules/lib/rules/no-custom-color.js
@@ -14,6 +14,7 @@ const alwaysValidClasses = [
   'bg-gradient-to-l',
   'bg-gradient-to-t',
   'bg-theme-bg-inherit',
+  'bg-gradient-to-br',
 ];
 const findTargetClassNames = (classNames) => {
   if (!classNames) {
--- a/packages/shared/src/components/LegoReferralBadge.tsx
+++ b/packages/shared/src/components/LegoReferralBadge.tsx
@@ -1,0 +1,53 @@
+import React, { ReactElement } from 'react';
+import classNames from 'classnames';
+import { ReferralCampaignKey, useReferralCampaign } from '../hooks';
+import { Image } from './image/Image';
+import { FlexRow } from './utilities';
+import { useActions } from '../hooks/useActions';
+import { ActionType } from '../graphql/actions';
+import { cloudinary } from '../lib/image';
+
+type LegoReferralBadgeProps = {
+  className?: string;
+  campaignKey: ReferralCampaignKey;
+};
+
+const LegoReferralBadge = ({
+  className,
+  campaignKey,
+}: LegoReferralBadgeProps): ReactElement => {
+  const { checkHasCompleted } = useActions();
+  const { isReady, referralCurrentCount, referralTargetCount } =
+    useReferralCampaign({
+      campaignKey,
+    });
+
+  if (checkHasCompleted(ActionType.LegoMay2023Hide)) {
+    return null;
+  }
+
+  if (!isReady) {
+    return null;
+  }
+
+  return (
+    <button
+      type="button"
+      className={classNames(
+        className,
+        'rounded-12 p-[3px] bg-gradient-to-br from-[#6DE8BE] via-[#DD3DFC] to-[#F9DD53]',
+      )}
+    >
+      <FlexRow className="gap-1.5 justify-center items-center py-1 px-3 font-bold rounded-10 bg-theme-bg-primary typo-callout text-theme-label-tertiary">
+        {referralCurrentCount}/{referralTargetCount}
+        <Image
+          width={33}
+          height={29}
+          src={cloudinary.referralCampaign.lego.piece}
+        />
+      </FlexRow>
+    </button>
+  );
+};
+
+export default LegoReferralBadge;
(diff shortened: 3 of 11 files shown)


packages/shared/src/contexts/AuthContext.tsx
```diff
@@ -120,7 +122,29 @@ export const AuthContextProvider = ({
 }: AuthContextProviderProps): ReactElement => {
   const [loginState, setLoginState] = useState<LoginState | null>(null);
   const endUser = user && 'providers' in user ? user : null;
-  const referral = user?.referrer;
+  const referralCampaign = useMemo<{
+    referralId: string;
+    referralOrigin: string;
+  }>(() => {
+    const cookie = typeof window !== 'undefined' ? getCookieObject() : {};
```
idoshamun, line 129: Wouldn't the window will always be undefined at the beginning in the webapp?

packages/shared/src/contexts/AuthContext.tsx
```diff
@@ -120,7 +122,29 @@ export const AuthContextProvider = ({
...
+    if (!joinReferralCookie) {
+      return undefined;
+    }
+
+    const [referralId, referralOrigin] = joinReferralCookie.split(':');
+
+    return {
+      referralId,
+      referralOrigin,
+    };
```
idoshamun, line 143: Let's turn it into a standalone function. Will look much better

idoshamun: Another thought about this, shouldn't we fetch only when needed? Instead of eagerly loading we can load it only when needed. Hence, ensuring window is defined and everything is there already

packages/webapp/pages/join.tsx
```diff
@@ -0,0 +1,199 @@
...
+  const result = await request<{ user: PageProps['referringUser'] }>(
+    graphqlUrl,
+    GET_REFERRING_USER_QUERY,
+    {
+      id: userId,
+    },
+  );
+
+  if (result) {
+    res.setHeader(
```
idoshamun, line 179: I think we should set the cookie on the client side because of the gdpr and so on.
We already have a cookie prompt that is shown automatically on first visit

packages/webapp/pages/join.tsx
```diff
@@ -0,0 +1,199 @@
...
+          />
+        </Modal>
+      )}
+    </div>
+  );
+};
+
+const validateUserId = (value: string) => !!value && value !== '404';
+
+export const getServerSideProps: GetServerSideProps<PageProps> = async ({
```
idoshamun, line 152: We'll have to check the performance of SSR in production. I like this approach but let's see how it performs

idoshamun (changes requested):
Some more comments on top of what I mentioned:
* I think you miss some tracking events
* We have to remember to take into account the referral in this case only in the extension
* When you do manual QA please make sure to do an end to end test that covers webapp -> installing extension -> signing up from extension -> checking referral count. Also do the same in the webapp and make sure referral doesn't go up
Overall looks great!

packages/webapp/pages/join.tsx
capJavert: If we are setting cookie on client side then we do not need SSR. Regarding this comment https://github.com/dailydotdev/apps/pull/1853#discussion_r1215299987

packages/webapp/pages/join.tsx
capJavert: I am fine with this, really depending on the legal. Should I then wait for setting a cookie until user click "I like cookies"?

packages/shared/src/contexts/AuthContext.tsx
capJavert: It would be during SSR, but on first render of the app window would be so the context would initialise with correct value.

packages/webapp/pages/join.tsx
idoshamun: Let me open it for discussion on the channel

packages/shared/src/contexts/AuthContext.tsx
capJavert: Refactored to use boot.

packages/webapp/pages/join.tsx
capJavert: @idoshamun even though we are setting cookie through client side JS I still left this as SSG page. 

Reasons:
- we need dynamic page so we can read query params (static pages do not have `router.query` until hydration)
- it is much cleaner from error handling perspective to check userid and cid params on the server then to do it on client side
- also we can easily fetch referring user on the server too and then pass to page as props
- regarding performance I added `s-max-age=60` header which makes this page cached for 60 seconds on [vercel CDN](https://vercel.com/docs/concepts/edge-network/caching#server) (same as ISR on squad invite page)
  - I did this many times on Vercel with Next.js and can vouch performance is going to be very close or same as static

packages/shared/src/contexts/AuthContext.tsx
```diff
@@ -120,7 +122,8 @@ export const AuthContextProvider = ({
 }: AuthContextProviderProps): ReactElement => {
   const [loginState, setLoginState] = useState<LoginState | null>(null);
   const endUser = user && 'providers' in user ? user : null;
-  const referral = user?.referrer;
+  const referral = user?.referralId || user?.referrer;
```
capJavert, line 125: From what I can see `user?.referrer` is  no longer used, if we can simplify this it would be great. There is a comment on API side about this also.

idoshamun: If you confirmed that it is not used then you can simplify :)

packages/webapp/pages/join.tsx
```diff
@@ -0,0 +1,235 @@
...
+
+    router.replace('/');
+  }, [user?.id, router]);
+
+  useEffect(() => {
+    document.cookie = [
+      `join_referral=${referringUser.id}:${campaign}`,
+      `max-age=${1 * 365 * 24 * 60 * 60}`,
+      isDevelopment ? undefined : 'Secure',
+      `domain=${process.env.NEXT_PUBLIC_DOMAIN}`,
```
idoshamun, line 69: Will it be `app.daily.dev` or `daily.dev`?
We need another property called `SameSite` and set it to `Lax` so it will be accessible from all subdomains

idoshamun (changes requested):
One more feedback round I guess and we're there

packages/webapp/pages/join.tsx
capJavert: From what I can see it is `daily.dev`, confirmed it is on preview.

It is accessible for subdomains already since `Domain=.daily.dev`

I'all add Lax to to match the sharing of cookies for other requests since I see we do use it.

packages/webapp/pages/join.tsx
capJavert: Updated cache to 1 day, since data for referral does not change that oftenand we can profit even more from cached requests.

capJavert:
@sshanzel would love to hear what you think, also on the epic https://github.com/dailydotdev/apps/pull/1849

packages/webapp/pages/join.tsx
idoshamun: Oh maybe I'm wrong but if I recall correctly when Lax was not presented the extension couldn't reach it.
An interesting experiment for later (out of this scope) is to check whether the Lax is really needed for the extension for example when it comes to our authentication cookie

idoshamun (approved):

packages/shared/src/contexts/AuthContext.tsx
capJavert: Gonna add it to backlog to investigate and confirm properly https://dailydotdev.atlassian.net/browse/WT-1431

sshanzel (commented):
Wow, great work on taking this on with just you! 💪 

Just had a few questions, but nothing major.

packages/webapp/pages/join.tsx
```diff
@@ -0,0 +1,236 @@
...
+      'daily dev is a professional network for developers to learn, collaborate, and grow together. Developers come to daily.dev to discover a wide variety of professional knowledge, create groups where they can collaborate with other developers they appreciate, and discuss the latest trends in the developer ecosystem.',
+    openGraph: defaultOpenGraph,
+  };
+
+  return (
+    <>
+      <NextSeo {...Seo} />
+      <div
+        style={{
+          backgroundImage: `url(${cloudinary.referralCampaign.backgroundDark})`,
```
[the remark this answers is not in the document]
capJavert: Good catch, but we decided dark mode only for this campaign to simplify.

packages/webapp/pages/join.tsx
```diff
@@ -0,0 +1,236 @@
...
+    };
+  }
+
+  const result = await request<{ user: PageProps['referringUser'] }>(
+    graphqlUrl,
+    GET_REFERRING_USER_QUERY,
+    {
+      id: userId,
+    },
+  );
```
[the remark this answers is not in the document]
capJavert: Currently `request` will throw 500 and it will show our web app error page, I think it is valid because we want to see broken links or someone doing something fishy. We could redirect to homepage also as we do for not found campaigns eg. soft error when campaign expires. @idoshamun WDYT?

packages/webapp/pages/join.tsx
```diff
@@ -0,0 +1,236 @@
...
+  useEffect(() => {
+    document.cookie = [
+      `join_referral=${referringUser.id}:${campaign}`,
+      `Max-Age=${1 * 365 * 24 * 60 * 60}`,
+      isDevelopment ? undefined : 'Secure',
+      `Domain=${process.env.NEXT_PUBLIC_DOMAIN}`,
+      `SameSite=Lax`,
+    ]
+      .filter(Boolean)
+      .join('; ');
```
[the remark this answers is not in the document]
capJavert: Definitely, I would put some more time in this utility, we would have to make sure all options are supported. Added to backlog https://dailydotdev.atlassian.net/browse/WT-1432

packages/webapp/pages/join.tsx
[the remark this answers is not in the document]
idoshamun: Edge case. We can ignore for now

PR #1868, opened 2023-06-08
feat: Moving the submit button to the mobile header

TomDowling:
## Changes
For Mobile only, I have removed the button in the footer and added a close and submit button for squad article submit modal.
Below are screenshots showing before, after and desktop

before
after
desktop

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1389 #done


packages/shared/src/components/modals/PostToSquadModal.tsx
```diff
@@ -51,6 +55,11 @@ function PostToSquadModal({
     preview,
   });
   const { completeAction } = useActions();
+  const isMobile = !useMedia([tablet.replace('@media ', '')], [true], false);
+  let titleText = null;
+  if (!isMobile) {
+    titleText = shouldSkipHistory ? 'Post article' : 'Share post';
+  }
```
sshanzel, line 62: I think we can simplify it from here then add a condition in the `title` property of the `<Modal.Header`. WDYT?

```suggestion
  const titleText = shouldSkipHistory ? 'Post article' : 'Share post';
```

sshanzel (changes requested):
Looking great! Just one update to avoid using let if as much as possible.

packages/shared/src/components/modals/PostToSquadModal.tsx
TomDowling: The issue is on Mobile the text needs to be undefined or null so it doesn't show on mobile only, so instead of duplicating the header and conditionally rendering it I did this method, I tried to put all this logic in the title prop but got the linting error "Do not nest ternary expressions" so the above was the solution I came, if you have any other suggestions so that on mobile the title isn't defined please do suggest or if you think i should conditionally render the Modal.Header for mobile and desktop please do say.

sshanzel: That linting rule is alright, though you should not get that on two separate ternaries. After the changes above we can then update the part where it should be undefined.

The header should then look like below
```
<Modal.Header showCloseButton={!isMobile} title={isMobile ? null : titleText}>
```

TomDowling: I like it, will make the change now, thanks for the suggestion :)

packages/shared/src/components/squads/Comment.tsx
```diff
@@ -63,6 +65,7 @@ export function SquadComment({
   const [commentary, setCommentary] = useState(form.commentary);
   const [link, setLink] = useState(preview.url);
   const [linkHint, setLinkHint] = useState(preview.url);
+  const isMobile = !useMedia([tablet.replace('@media ', '')], [true], false);
```
rebelchris, line 68: Should we not use `const { sidebarRendered } = useSidebarRendered();`
It's a more condensed hook I think we preferred for mobile definition. (@sshanzel correct?)

rebelchris (changes requested):
Neat work, have some minor suggestions.

packages/shared/src/components/modals/PostToSquadModal.tsx
```diff
@@ -51,6 +55,12 @@ function PostToSquadModal({
     preview,
   });
   const { completeAction } = useActions();
+  const isMobile = !useMedia([tablet.replace('@media ', '')], [true], false);
```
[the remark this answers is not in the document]
sshanzel: I think the issue here is the design team wanted the UI for mobile and tablet to be different 😢 

We used to have them behave the same way, but I noticed recently they are trying to make tablet have the same UI as the laptop.

With that, this should be correct as we are only trying to apply it on mobile, not on tablet and above.

packages/shared/src/components/squads/Comment.tsx
sshanzel: Same as here https://github.com/dailydotdev/apps/pull/1868#discussion_r1231976391

sshanzel (changes requested):
Hey. Really sorry about this but I just realized, we have now deprecated the `PostToSquad` modal with the introduction of `CreateSharedPost` modal. Can you apply the changes for the modal header on that component?

sshanzel (approved):
Nice work! 🥳 🚢

rebelchris (approved):
Looks good 👍

PR #1869, opened 2023-06-08
feat: commentary mention support

sshanzel:
## Changes
- Allow mention in the commentary by using the MarkdownInput but having disabled all the features except mention.
- Refactored the Markdown
  - To allow to completely replace the footer.
  - To make preview feature an optional prop.
  - Also fixed some margin on the post button along the way.
- New modal for CreateSharedPost to be used by the Share button.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1391 #done


packages/shared/src/components/modals/post/CreateSharedPostModal.tsx
```diff
@@ -0,0 +1,125 @@
...
+    return onSubmitPost(e, squad.id, commentary);
+  };
+
+  const [checkUrl] = useDebounce((value: string) => {
+    const links = [link, updatedPreview?.url, updatedPreview?.permalink];
+    if (!isValidHttpUrl(value) || links.some((url) => url === value))
+      return null;
+
+    return getLinkPreview(value);
+  }, 1000);
```
rebelchris, line 57: We are using this 3 times now I see, perhaps we can simply it in a custom hook to keep it aligned if anything ever changes

packages/shared/src/components/modals/post/CreateSharedPostModal.tsx
```diff
@@ -0,0 +1,125 @@
...
+    setLink(value);
+    checkUrl(value);
+  };
+
+  return (
+    <Modal kind={Modal.Kind.FlexibleCenter} size={Modal.Size.Medium} {...props}>
+      <Modal.Header title="New post" />
+      <form
+        className="flex flex-col p-3 w-full"
+        action="#"
```
rebelchris, line 70: This feels a bit odd?

packages/shared/src/components/modals/post/CreateSharedPostModal.tsx
```diff
@@ -0,0 +1,125 @@
...
+                className="flex-col-reverse m-3 !w-auto"
+                preview={updatedPreview ?? preview}
+                link={link}
+                onLinkChange={onInput}
+              />
+            )
+          }
+        />
+      </form>
+      <Modal.Footer className="typo-caption1">
```
rebelchris, line 97: Why is this footer different from the guidelines?

rebelchris (changes requested):
Missing a little bit of context around this, so only reviewed it very high level.
Perhaps I can look closer after a quick catch up in our 1:1?

packages/shared/src/components/modals/common.tsx
```diff
@@ -64,6 +71,7 @@ export const modals = {
   [LazyModal.ReadingHistory]: ReadingHistoryModal,
   [LazyModal.SquadPromotion]: SquadPromotionModal,
   [LazyModal.LegoReferralCampaign]: LegoReferralCampaignModal,
+  [LazyModal.CreateSharedPost]: CreateSharedPostModal,
```
capJavert, line 74: Can we also remove `PostToSquadModal`? 

The new modal should replace it right, functionality wise? If yes we need to make sure `SharePostChecklistStep` and `CompanionMenu` are updated to use this new modal.

packages/shared/src/components/fields/MarkdownInput/index.tsx
```diff
@@ -89,28 +116,46 @@ function MarkdownInput({
...
             ref={textareaRef}
-            className="m-4 bg-transparent outline-none typo-body placeholder-theme-label-quaternary"
-            placeholder="Start a discussion, ask a question or write about anything that you believe would benefit the squad. (Optional)"
+            className={classNames(
+              'flex flex-1 bg-transparent outline-none typo-body placeholder-theme-label-quaternary',
+              showUserAvatar ? 'm-3' : 'm-4',
+            )}
             value={input}
-            onClick={() => checkMention()}
+            onClick={() => checkMention && checkMention()}
```
capJavert, line 154: Would prefer `if` here for verbosity and also return type of this expression.

packages/shared/src/components/fields/MarkdownInput/index.tsx
```diff
@@ -72,6 +97,8 @@ function MarkdownInput({
     enabledCommand,
   });
 
+  useImperativeHandle(ref, () => ({ textareaRef, onMentionCommand }));
```
capJavert, line 100: Just to check if we need this or we use this handle on ref anywhere in the code? Can not see it.

capJavert (changes requested):
Few thing that caught my eye, functionality wise I noticed the input does not get cleared after posting.

image

packages/shared/src/components/modals/common.tsx
sshanzel: That's right. I was waiting for some answer from the design team on empty state that `PostToSquadModal` provides. I believe they have gotten back to it already. I will see its new implementation.

packages/shared/src/components/fields/MarkdownInput/index.tsx
sshanzel: We use this on the Footer button which is located outside the editor. I will line the line of code once I stumble upon it.

sshanzel: Should be from this line:

https://github.com/dailydotdev/apps/pull/1869/files#diff-eba809a0ead9105aaba7c14726f5ed2bc7c5b59485163dd6b6d9f9de7d8fef9bR104

packages/shared/src/components/modals/post/CreateSharedPostModal.tsx
sshanzel: We always `preventDefault`, this is actually safer than having no `action` to ensure the page won't reload.

packages/shared/src/components/modals/post/CreateSharedPostModal.tsx
sshanzel: Not sure if there is any custom font size, but afair there is no font specified for `ModalFooter`.

The font used here though is for the screenshot below.

Screenshot 2023-06-15 at 6 05 09 PM

packages/shared/src/components/modals/post/CreateSharedPostModal.tsx
sshanzel: Good point. I have created a hook for this now.

sshanzel:
@capJavert nice catch! Though it is weird that the input is not getting updated with its controlled value. To fix this, I introduced a ref to manually update the value.

@rebelchris have now made the necessary refactors 🚢

packages/shared/src/components/modals/post/CreateSharedPostModal.tsx
rebelchris: Got it, must be a new addition indeed 🤙

rebelchris (approved):
Cool, nothing from my side 👍

sshanzel:
Have deployed this in preview2 to make it easier to test. Kindly try sharing some post to a Squad.

sshanzel:
Merging now. Ante mentioned to proceed without his final review as we already fixed his concerns.

PR #1878, opened 2023-06-14
feat: when sharing a post we show our modal and not the native popup by default

TomDowling:
## Changes
Making the Share modal always open and adding a button to create a new squad when you have no squads and a a Share to... button to open the native share menu when it is available (mostly mobile). Below are the screenshots for desktop when you don't have any squads and when you do

Screenshot 2023-06-14 at 12 30 59 pm
Screenshot 2023-06-14 at 12 31 09 pm

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1387 WT-1388 #done


packages/shared/src/components/widgets/SocialShare.tsx
```diff
@@ -25,15 +25,27 @@ import { Comment, getCommentHash } from '../../graphql/comments';
 import { useAuthContext } from '../../contexts/AuthContext';
 import { useLazyModal } from '../../hooks/useLazyModal';
 import { LazyModal } from '../modals/common/types';
-import { SourcePermissions, Squad } from '../../graphql/sources';
+import { SourcePermissions, Squad as SquadType } from '../../graphql/sources';
```
rebelchris, line 28: What was the conflict here, the inline SocialShareType?
If so it might be best to move that to a lib file instead.

packages/shared/src/components/widgets/SocialShare.tsx
```diff
@@ -78,77 +98,129 @@ export const SocialShare = ({
...
+                    type="button"
+                    className="flex flex-col items-center w-16 text-center"
+                    onClick={() =>
+                      openNewSquadModal({
+                        origin: Origin.Share,
+                        redirectAfterCreate: false,
+                      })
+                    }
+                  >
+                    <div className="flex justify-center items-center w-12 h-12 bg-cabbage-40 rounded-full">
```
rebelchris, line 144: The bg color should be `bg-theme-color-cabbage`

packages/shared/src/components/modals/ShareModal.tsx
```diff
@@ -103,8 +103,18 @@ export default function ShareModal({
           value={link}
           readOnly
         />
-        <p className="py-2.5 font-bold typo-callout">Share to</p>
         <SocialShare
```
rebelchris, line 106: Why do we load this dynamically?
Is it not simpler to have both inside the social share directly?

packages/shared/src/components/widgets/SocialShare.tsx
```diff
@@ -78,77 +98,129 @@ export const SocialShare = ({
...
+                ?.map((squad) => (
+                  <button
+                    type="button"
+                    className="flex flex-col items-center w-16 text-center"
+                    key={squad.id}
+                    onClick={() => onShareToSquad(squad)}
+                  >
+                    <SourceProfilePicture source={squad} />
+                    <ShareText className="mt-2 break-words">
+                      @{squad.handle}
```
rebelchris, line 126: Did you take the new guideline in sketch into account, it should have a max-width and dots when overflowing.

rebelchris (changes requested):
I think 4.0 and 4.1 are missing.
Looks good, would consider moving the logic inside the SocialShare as it has no benefit of rendering it conditionally now.

packages/shared/src/components/widgets/SocialShare.tsx
TomDowling: This was the conflict, i'll move it to lib

packages/shared/src/components/ShareBar.tsx
```diff
@@ -42,53 +53,155 @@ export default function ShareBar({ post }: { post: Post }): ReactElement {
...
+      <p className="mb-4 typo-callout text-theme-label-tertiary">
         Would you recommend this post?
       </p>
-      <div className="inline-flex relative flex-row items-center mt-3">
+
+      <div className="flex flex-wrap gap-2">
         <SimpleTooltip content="Copy link">
-          <ShareButton
+          <Button
+            className="group flex-col mb-2 font-normal text-center text-salt-90 hover:bg-transparent border-0 share btn-tertiary typo-caption1"
```
rebelchris, line 81:
```suggestion
            className="group flex-col mb-2 font-normal text-center hover:bg-transparent border-0 share btn-tertiary typo-caption1"
```

Although this share class feels a bit like we are abusing a button rather then using a styleguide at the moment.
I don't want to further block unless @sshanzel agrees with me we should rather change it.

packages/shared/src/components/ShareBar.tsx
```diff
@@ -42,53 +53,155 @@ export default function ShareBar({ post }: { post: Post }): ReactElement {
...
             onClick={trackAndCopyLink}
             pressed={copying}
-            icon={<CopyIcon />}
-            className="btn-tertiary-avocado"
-          />
+            textPosition="justify-between"
+            icon={
+              <div className="flex justify-center items-center mb-2 group-hover:bg-theme-hover rounded-xl">
+                <CopyIcon secondary className="w-7 h-7" />
```
rebelchris, line 87:
```suggestion
                <CopyIcon secondary size={IconSize.Medium} />
```
Avoid using direct implementations of width and height in favor of our guidelined set icon sizes.

packages/shared/src/components/ShareBar.tsx
```diff
@@ -42,53 +53,155 @@ export default function ShareBar({ post }: { post: Post }): ReactElement {
...
+            }
+          >
+            Copy link
+          </Button>
         </SimpleTooltip>
+
         <SimpleTooltip content="Share on WhatsApp">
-          <ColorfulShareButton
+          <Button
+            className="group flex-col mb-2 font-normal text-center text-salt-90 hover:bg-transparent border-0 share btn-tertiary typo-caption1"
```
rebelchris, line 97:
```suggestion
            className="group flex-col mb-2 font-normal text-center hover:bg-transparent border-0 share btn-tertiary typo-caption1"
```

The salt color is already the color of a tertiary button.
Also we should avoid using direct color names and prefer our named versions.

packages/shared/src/components/ShareBar.tsx
```diff
@@ -42,53 +53,155 @@ export default function ShareBar({ post }: { post: Post }): ReactElement {
...
             href={getWhatsappShareLink(href)}
             target="_blank"
             rel="noopener"
             onClick={() => onClick(ShareProvider.WhatsApp)}
-            icon={<WhatsappIcon secondary />}
-            className="btn-tertiary"
-          />
+            textPosition="justify-between"
+            icon={
+              <div className="flex justify-center items-center mb-2 group-hover:bg-theme-hover rounded-xl">
```
rebelchris, line 105: This is pretty much repeated a lot of times, suggest to create a functional component that can render it.

packages/shared/src/components/ShareBar.tsx
```diff
@@ -42,53 +53,155 @@ export default function ShareBar({ post }: { post: Post }): ReactElement {
...
+                <WhatsappIcon secondary className="w-7 h-7 text-white" />
```
rebelchris, line 106: For reference on sizing

packages/shared/src/components/ShareBar.tsx
```diff
@@ -42,53 +53,155 @@ export default function ShareBar({ post }: { post: Post }): ReactElement {
...
+          >
+            WhatsApp
+          </Button>
         </SimpleTooltip>
-        <SimpleTooltip content="Share on Twitter">
-          <ColorfulShareButton
+
+        <SimpleTooltip content="Share on Facebook">
+          <Button
+            className="group flex-col mb-2 font-normal text-center text-salt-90 hover:bg-transparent border-0 share btn-tertiary typo-caption1"
```
rebelchris, line 116: For reference on color

packages/shared/src/components/ShareBar.tsx
```diff
@@ -42,53 +53,155 @@ export default function ShareBar({ post }: { post: Post }): ReactElement {
...
-            onClick={() => onClick(ShareProvider.Twitter)}
-            icon={<TwitterIcon secondary />}
-            className="btn-tertiary"
-          />
+            onClick={() => onClick(ShareProvider.Facebook)}
+            textPosition="justify-start"
+            icon={
+              <div className="flex justify-center items-center mb-2 group-hover:bg-theme-hover rounded-xl">
+                <FacebookIcon secondary className="w-7 h-7 text-white" />
+              </div>
```
rebelchris, line 126: Reference on component + sizing

packages/shared/src/components/ShareBar.tsx
```diff
@@ -42,53 +53,155 @@ export default function ShareBar({ post }: { post: Post }): ReactElement {
...
+            }
+          >
+            Facebook
+          </Button>
         </SimpleTooltip>
-        <SimpleTooltip content="Share on Facebook">
-          <ColorfulShareButton
+
+        <SimpleTooltip content="Share on Twitter">
+          <Button
```
rebelchris, line 134: Reference on color, also suggestion maybe just create a generic classList of functional component?

packages/shared/src/components/ShareBar.tsx
```diff
@@ -42,53 +53,155 @@ export default function ShareBar({ post }: { post: Post }): ReactElement {
...
-            onClick={() => onClick(ShareProvider.Facebook)}
-            icon={<FacebookIcon secondary />}
-            className="btn-tertiary"
-          />
+            onClick={() => onClick(ShareProvider.Twitter)}
+            textPosition="justify-start"
+            icon={
+              <div className="flex justify-center items-center mb-2 group-hover:bg-theme-hover rounded-xl">
+                <TwitterIcon secondary className="w-7 h-7 text-white" />
+              </div>
```
rebelchris, line 145: Mention on component + sizing

packages/shared/src/components/ShareBar.tsx
```diff
@@ -42,53 +53,155 @@ export default function ShareBar({ post }: { post: Post }): ReactElement {
...
+        {squads
+          ?.filter(
+            (squadItem) =>
+              squadItem.active &&
+              verifyPermission(squadItem, SourcePermissions.Post),
+          )
+          ?.map((squad) => (
+            <SimpleTooltip content={`Share on ${squad.name}`} key={squad.id}>
+              <Button
+                className="group flex-col mb-2 font-normal text-center text-salt-90 hover:bg-transparent border-0 share btn-tertiary typo-caption1"
```
rebelchris, line 161: Color name

packages/shared/src/components/ShareBar.tsx
```diff
@@ -42,53 +53,155 @@ export default function ShareBar({ post }: { post: Post }): ReactElement {
...
+                key={squad.id}
+                onClick={() => onShareToSquad(squad)}
+                textPosition="justify-between"
+                icon={
+                  <div className="flex justify-center items-center mb-2">
+                    <SourceProfilePicture
+                      source={squad}
+                      className="w-10 h-10 squad"
```
rebelchris, line 169: I think for this we also support sizing, secondary, why is the squad class needed?

packages/shared/src/components/ShareBar.tsx
```diff
@@ -42,53 +53,155 @@ export default function ShareBar({ post }: { post: Post }): ReactElement {
...
+              >
+                <div className="truncate">@{squad.handle}</div>
+              </Button>
+            </SimpleTooltip>
+          ))}
+
+        {hasSquadAccess && squads?.length === 0 && (
+          <SimpleTooltip content="Create new squad">
+            <Button
+              className="group flex-col mb-2 font-normal text-center text-salt-90 hover:bg-transparent border-0 share btn-tertiary typo-caption1"
```
rebelchris, line 182: Text color

packages/shared/src/components/ShareBar.tsx
```diff
@@ -42,53 +53,155 @@ export default function ShareBar({ post }: { post: Post }): ReactElement {
...
+              onClick={() =>
+                openNewSquadModal({
+                  origin: Origin.Share,
+                  redirectAfterCreate: false,
+                })
+              }
+              textPosition="justify-between"
+              icon={
+                <div className="flex justify-center items-center mb-2">
+                  <div className="flex justify-center items-center w-10 h-10 rounded-full squad bg-theme-color-cabbage">
```
rebelchris, line 192: It shared the squad class from above?
Not sure this is then a generic component?

packages/shared/src/components/ShareBar.tsx
```diff
@@ -42,53 +53,155 @@ export default function ShareBar({ post }: { post: Post }): ReactElement {
...
+                    <PlusIcon
+                      size={IconSize.Small}
+                      className="text-pepper-90"
```
rebelchris, line 195: This is the primary label color then.
`text-theme-label-primary` (But it should come down from the above I think actually already?

packages/shared/src/components/modals/ShareModal.tsx
```diff
@@ -36,7 +42,9 @@ export default function ShareModal({
     ? `${post?.commentsPermalink}${getCommentHash(comment.id)}`
     : post?.commentsPermalink;
   const { trackEvent } = useContext(AnalyticsContext);
-  const [, copyUrl] = useCopyLink(() => link);
+  const [commentary, setCommentary] = useState<string>('');
```
rebelchris, line 45: You don't need to typecast <string> if you pass a empty string directly.
```suggestion
  const [commentary, setCommentary] = useState('');
```

packages/shared/src/components/modals/ShareModal.tsx
rebelchris: I still think these two <SocialShare> can just be inside the SocialShare component.

packages/shared/src/components/post/write/WriteLinkPreview.tsx
```diff
@@ -12,37 +12,57 @@ import {
...
+};
+
 export function WriteLinkPreview({
   link,
   preview,
   onLinkChange,
   className,
+  showOpenLink = true,
+  showPreviewLink = true,
+  sourceInfoFormat = 'long',
```
rebelchris, line 43: I'd prefer this to be a enum if we using this.

packages/shared/src/components/post/write/WriteLinkPreview.tsx
```diff
@@ -51,16 +71,31 @@ export function WriteLinkPreview({
...
-          <Button
-            icon={<OpenLinkIcon />}
-            className="btn-tertiary"
-            type="button"
-            tag="a"
-            target="_blank"
-            rel="noopener noreferrer"
-            href={link}
+          {preview.source && sourceInfoFormat === 'avatar' && (
+            <div className="flex absolute right-28 justify-center items-center w-8 h-8 rounded-full bg-theme-bg-primary">
```
rebelchris, line 75: Are these styles not already applied on the SourceAvatar?
What was the difference here

packages/shared/src/components/widgets/SocialShare.tsx
```diff
@@ -62,93 +84,154 @@ export const SocialShare = ({
...
+                      openNewSquadModal({
+                        origin: Origin.Share,
+                        redirectAfterCreate: false,
+                      })
+                    }
+                  >
+                    <div className="flex justify-center items-center w-12 h-12 rounded-full bg-theme-color-cabbage">
+                      <PlusIcon
+                        size={IconSize.Large}
+                        className="text-pepper-90"
```
rebelchris, line 150: Primary color

packages/shared/src/components/widgets/SocialShare.tsx
```diff
@@ -62,93 +84,154 @@ export const SocialShare = ({
...
+          </>
+        )}
+
+      {type === SocialShareType.External && (
+        <>
+          <p className="pb-4 font-bold typo-callout">Share externally</p>
+          <section className="grid overflow-x-auto tablet:overflow-x-visible tablet:grid-cols-5 grid-flow-col tablet:grid-flow-row auto-cols-auto gap-4">
+            <SocialShareIcon
+              icon={<CopyIcon size={IconSize.Large} />}
+              className="text-pepper-90"
```
rebelchris, line 169: Primary

packages/shared/src/components/widgets/SocialShareIcon.tsx
```diff
@@ -20,22 +22,33 @@ export const SocialShareIcon = ({
...
-        iconOnly
-        icon={icon}
-      />
+    <Button
+      tag={href ? 'a' : 'button'}
+      href={href}
+      onClick={onClick}
+      target="_blank"
+      rel="noopener"
+      className={classNames(className, 'reset flex-col w-[62px] border-0')}
```
rebelchris, line 37: What is this reset? also why does it feel like a custom button, not sure that's correct at this point.
Was it not our guidelined button?

packages/shared/src/components/widgets/SocialShareIcon.tsx
```diff
@@ -20,22 +22,33 @@ export const SocialShareIcon = ({
...
+      icon={
+        <div
+          className={classNames(
+            iconBg,
+            'flex justify-center items-center w-12 h-12 rounded-[14px] mb-2',
```
rebelchris, line 42: Did we not have this rounded on 14?
Better to check with design and avoid using non guidelined radius.

packages/shared/src/styles/components/buttons.css
```diff
@@ -51,6 +51,27 @@
     color: #FFFFFF;
   }
 
+  &.share,
+  &.share:not(.iconOnly) {
+    padding: 0;
+    width: 4rem;
+    height: fit-content;
+    /* min-height: 3.875rem; */
```
rebelchris, line 59: Not needed?

rebelchris (changes requested):
Mainly some code cleanup and guideline issues.
Functionality wise looks alright, one suggestion to make some functional components to share same snippets.

packages/shared/src/components/modals/ShareModal.tsx
```diff
@@ -53,11 +61,6 @@ export default function ShareModal({
       }),
     );
 
-  const trackAndCopyLink = () => {
-    baseTrackingEvent('share post', { provider: ShareProvider.CopyLink });
-    copyUrl();
-  };
-
```
sshanzel, line 60: Was it intended to remove the tracking here? @TomDowling @rebelchris

rebelchris: We should not remove any tracking for this PR as far as I'm concerned.

sshanzel: Thought so too. Let me bring this back.

packages/shared/src/components/ShareBar.tsx
sshanzel: Agreed. Changes to the styling of the button should be as minimal as possible if any at all.

packages/shared/src/components/modals/ShareModal.tsx
TomDowling: This was removed as the copy link is now within the SocialShare component instead of direct on the ShareModal

sshanzel: Ah, I see. Thought it was totally removed. Thanks for the note!

packages/shared/src/components/ShareBar.tsx
```diff
@@ -42,53 +44,70 @@ export default function ShareBar({ post }: { post: Post }): ReactElement {
...
+      },
+    });
+  };
+
   return (
     <WidgetContainer className="hidden laptop:flex flex-col p-3">
-      <p className="typo-callout text-theme-label-tertiary">
-        Would you recommend this post?
+      <p className="mb-4 typo-callout text-theme-label-tertiary">
+        One-click share
```
rebelchris, line 66: @sshanzel I know design says this, but we decided afterwords it should remain the same, sorry it's a bit messy in sketch.

packages/shared/src/components/modals/ShareModal.tsx
rebelchris: @sshanzel Just a note this shouldn't be needed right?

packages/shared/src/components/post/write/WriteLinkPreview.tsx
```diff
@@ -15,34 +15,48 @@ import { ExternalLinkPreview } from '../../../graphql/posts';
...
+        />
+      )}
       {preview.title && preview.image && (
-        <WritePreviewContent>
+        <WritePreviewContent className={isMinimized && '!py-2 !px-3'}>
           <div className="flex flex-col flex-1 typo-footnote">
             <span className="font-bold line-clamp-2">{preview.title}</span>
-            {preview.source && (
+            {preview.source && isMinimized ? (
+              <div className="flex absolute right-28 justify-center items-center w-8 h-8 rounded-full bg-theme-bg-primary">
```
rebelchris, line 52: Is this wrapper needed here?
I thought the SourceAvatar handled most of this? (but could be wrong)

packages/shared/src/components/squads/SquadsToShare.tsx
```diff
@@ -0,0 +1,72 @@
...
+  onClick: (e: MouseEvent, squad: Squad) => void;
+  size?: ButtonSize;
+  squadIconSize?: ProfileImageSize;
+}
+
+export function SquadsToShare({
+  isLoading,
+  onClick,
+  size = ButtonSize.Large,
+  squadIconSize = 'xlarge',
```
rebelchris, line 25: This could probably be the IconSize enum right?

packages/shared/src/components/widgets/SocialShare.tsx
```diff
@@ -62,93 +72,103 @@ export const SocialShare = ({
...
-        onSharedSuccessfully: () => {
-          trackEvent(postAnalyticsEvent(AnalyticsEvent.ShareToSquad, post));
-          return onSquadShare;
-        },
-      },
-    });
+  const { onSubmitPost, isPosting } = usePostToSquad({
+    initialPreview: post,
+    onPostSuccess: () => {
+      trackEvent(postAnalyticsEvent(AnalyticsEvent.ShareToSquad, post));
```
rebelchris, line 78: With this did we now loose the "StartShareToSquad" tracking or is it moved somewhere?
Maybe it's just not needed anymore, but good to know for analytics

rebelchris (changes requested):
Amazing!
Some really minor things, should be quick fixes, then ready to roll I think.

packages/shared/src/components/ShareBar.tsx
sshanzel: Oh, cool. I shall revert this.

packages/shared/src/components/modals/ShareModal.tsx
sshanzel: Ah, right. My bad, let me update.

packages/shared/src/components/squads/SquadsToShare.tsx
sshanzel: Unfortunately, they are not an `enum` yet. It would be a big change to start here. We should probably write a ticket to migrate. WDYT?

packages/shared/src/components/widgets/SocialShare.tsx
sshanzel: Let me verify quickly.

packages/shared/src/components/squads/SquadsToShare.tsx
rebelchris: Ah sorry it sounded like a `IconSize` but it's a `ProfileImageSize` perhaps we should rename this then?
It's not an icon but a squadImage rather? (Just to avoid confusion down the line)

packages/shared/src/components/post/write/WriteLinkPreview.tsx
sshanzel: You are right! Just needed a few classes but we should not need the wrapper.

packages/shared/src/components/squads/SquadsToShare.tsx
sshanzel: Ah, that is true. Will update quickly.

packages/shared/src/components/widgets/SocialShare.tsx
sshanzel: The "StartShareToSquad" is used when opening the share comment modal. The "SocialShare" component is already opened, so we have tracked it already somewhere else in this case.

rebelchris: Hmm ok, I'll double check with Ido later, but good for now, thanks for checking this.

packages/shared/src/components/ShareBar.tsx
```diff
@@ -42,53 +44,70 @@ export default function ShareBar({ post }: { post: Post }): ReactElement {
...
+    trackEvent(postAnalyticsEvent(AnalyticsEvent.StartShareToSquad, post));
+    openModal({
+      type: LazyModal.CreateSharedPost,
+      props: {
+        squad,
+        preview: post,
+        onSharedSuccessfully: () => {
+          return trackEvent(
+            postAnalyticsEvent(AnalyticsEvent.ShareToSquad, post),
+          );
```
rebelchris, line 57: This would do right?
```suggestion
        onSharedSuccessfully: () => trackEvent(
            postAnalyticsEvent(AnalyticsEvent.ShareToSquad, post),
          )
```

rebelchris (approved):
Amazing, good to go from my side.
One small non blocking comment :)

PR #1879, opened 2023-06-14
fix: add option to go back to squad from article

rebelchris:
- Added a option to go back to a squad on a article if the origin of flow was squad's
- Used query params as it should only flow if user moves via this flow
- Design is slightly adjusted to make things easier for now (will refactor in later stage, see discussion [here](https://dailydotdev.slack.com/archives/C051TNXSXR8/p1686750909746819))

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1395 #done

### DIFF
--- a/packages/shared/src/components/post/SharePostContent.tsx
+++ b/packages/shared/src/components/post/SharePostContent.tsx
@@ -39,7 +39,7 @@ function SharePostContent({
           href={
             isUnknownSource
               ? post.sharedPost.permalink
-              : `${post.sharedPost.commentsPermalink}?squad=${post.source.name}`
+              : `${post.sharedPost.commentsPermalink}?squad=${post.source.handle}&n=${post.source.name}`
           }
           as={isUnknownSource ? undefined : post.sharedPost.commentsPermalink}
         >
--- a/packages/shared/src/components/post/SharePostContent.tsx
+++ b/packages/shared/src/components/post/SharePostContent.tsx
@@ -36,7 +36,7 @@ function SharePostContent({
           href={
             post.sharedPost.source.id === 'unknown'
               ? post.sharedPost.permalink
-              : post.sharedPost.commentsPermalink
+              : `${post.sharedPost.commentsPermalink}?squad=${post.source.id}&n=${post.source.name}`
           }
           title="Go to post"
           target="_blank"
--- a/packages/shared/src/components/post/SharePostContent.tsx
+++ b/packages/shared/src/components/post/SharePostContent.tsx
@@ -29,25 +29,25 @@ function SharePostContent({
     return shouldShowSummary ? height : 0;
   }, [shouldShowSummary, height]);
 
+  const isUnknownSource = post.sharedPost.source.id === 'unknown';
+
   return (
     <>
       <p className="mt-6 whitespace-pre-line typo-title3">{post.title}</p>
       <div className="flex flex-col mt-8 rounded-16 border border-theme-divider-tertiary hover:border-theme-divider-secondary">
         <Link
           href={
-            post.sharedPost.source.id === 'unknown'
+            isUnknownSource
               ? post.sharedPost.permalink
               : `${post.sharedPost.commentsPermalink}?squad=${post.source.name}`
           }
-          as={
-            post.sharedPost.source.id === 'unknown'
-              ? undefined
-              : post.sharedPost.commentsPermalink
-          }
+          as={isUnknownSource ? undefined : post.sharedPost.commentsPermalink}
         >
           {/* eslint-disable-next-line react/jsx-no-comment-textnodes, jsx-a11y/click-events-have-key-events, jsx-a11y/no-static-element-interactions */}
           <a
             title="Go to post"
+            target={isUnknownSource ? '_blank' : undefined}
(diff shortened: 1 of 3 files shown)


packages/shared/src/components/post/SharePostContent.tsx
```diff
@@ -36,7 +36,7 @@ function SharePostContent({
           href={
             post.sharedPost.source.id === 'unknown'
               ? post.sharedPost.permalink
-              : post.sharedPost.commentsPermalink
+              : `${post.sharedPost.commentsPermalink}?squad=${post.source.id}&n=${post.source.name}`
```
capJavert, line 39: Can't we just use squad handle and also mask the query params with `as`? Masking also makes sure nobody will copy the link so no danger of broken links if squad handle changes. Don't know if we want users to actually copy this link anyway, it is more of a internal app thing. WDYT?

capJavert (commented):
Simple enough, single idea about masking the URL.

packages/shared/src/components/post/SharePostContent.tsx
rebelchris: Only internal for now, how do you mean masking with as?
Not sure i follow correctly

capJavert: Using `as` prop from nextjs `Link`, something like this:
```jsx
<Link
  href={`${post.sharedPost.commentsPermalink}?squad=${post.source.id}&n=${post.source.name}`}
  as={post.sharedPost.commentsPermalink}
>
  <a>...</a>
</Link>
```

rebelchris: Oh wasnt aware of this! Feel free to adjust in the PR if you want 👍
Is it a web thing or a specific react thing?

capJavert: Hmm looked at it now, we use `_blank` for opening those links, is that intentional? Because since `as` is nextjs router thing (the browser just sees the final URL) so the masking will not work with `_blank`.

capJavert: Sorry for not seeing it while reviewing initially.

rebelchris: I think we can make it conditional of its internal or not, could you please check with @idoshamun?
Else I'll check tomorrow 🙏

capJavert: Because that is the mechanism to show "Back to squad" link when you open a shared post that is from daily.dev. 

image

idoshamun: Oh don't mind me, I thought it's a share externally component. I suggest we follow Ante's advice. It doesn't makes sense anymore to open it in new tab because we have this nice utility button now

packages/shared/src/components/post/SharePostContent.tsx
```diff
@@ -32,48 +33,55 @@ function SharePostContent({
...
-            <LazyImage
-              imgSrc={post.sharedPost.image}
-              imgAlt="Post cover image"
-              ratio="52%"
-              eager
-              fallbackSrc={cloudinary.post.imageCoverPlaceholder}
-            />
-          </div>
-        </a>
+          {/* eslint-disable-next-line react/jsx-no-comment-textnodes, jsx-a11y/click-events-have-key-events, jsx-a11y/no-static-element-interactions */}
```
capJavert, line 48: Needed to add this since `a` without `href` was triggering eslint rules and our nextjs version does not support `onClick` on `Link`.

packages/webapp/pages/posts/[id]/index.tsx
```diff
@@ -142,15 +145,21 @@ const PostPage = ({ id, initialData }: Props): ReactElement => {
     <SquadPostPageNavigation squadLink={post.source.permalink} />
   );
   const navigation: Record<PostType, ReactNode> = {
-    article: null,
+    article: !!router?.query?.squad && (
+      <Link href={`/squads/${router.query.squad}`}>
```
Comment (rebelchris) [inline packages/webapp/pages/posts/[id]/index.tsx:149]: @capJavert this is not correct now, you made it the name, but squad handle could be different.
That's why i added both, one for the url and one for text in the link.
Or did i miss something in squad urls maybe?

packages/shared/src/components/post/SharePostContent.tsx
rebelchris: @idoshamun / @capJavert I think we should only redirect without blank to internal links right?

- Links to `post.sharedPost. commentsPermalink` then don't add _blank
- Is external only `post.sharedPost.permalink` then do add _blank

packages/webapp/pages/posts/[id]/index.tsx
Comment (capJavert) [reply packages/webapp/pages/posts/[id]/index.tsx:149]: Should be fixed now, completely missed the part of logic.

capJavert (approved):
🚀

sshanzel (approved):
Nice to see we were able to easily use the `navigation` object for the article. Great work! 🚀

PR #1881, opened 2023-06-16
refactor: comment box to markdown input

sshanzel:
# Initiative 15

Core changes: we used to open a modal for making or editing a comment but we will now instead show a comment box underneath the comment the user is trying to make a reply to. We have removed a lot of codes due to the fact that the whole process is entirely different, and we removed the legacy markdown input with the introduction of the reused `MarkdownInput`.

I would appreciate testing. Currently deployed at https://preview.app.daily.dev/squads/dailydev2

I have also moved some utility functions from the extension package to the shared package. We have abstracted some process flow to make it easier to maintain across the app. Another option is, instead of moving them to shared package, we will make it come from the `useRequestProtocol` and call it from the hook where it was used. WDYT?

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done


packages/shared/src/components/fields/MarkdownInput/CommentMarkdownInput.tsx
```diff
@@ -0,0 +1,209 @@
...
+  parentCommentId,
+  initialContent,
+  replyTo,
+  editCommentId,
+  className = {},
+  onCommented,
+}: CommentMarkdownInputProps): ReactElement {
+  const client = useQueryClient();
+  const markdownRef = useRef<MarkdownRef>();
+  const key = useMemo(
```
rebelchris, line 64: This seems pretty intense, should we not consider this a global key function?
 (which we have for some other parts)

packages/shared/src/components/fields/MarkdownInput/CommentMarkdownInput.tsx
```diff
@@ -0,0 +1,209 @@
...
+        editCommentId,
+        parentCommentId,
+      ].filter((value) => !!value),
+    [postId, sourceId, editCommentId, parentCommentId],
+  );
+  const { requestMethod, isCompanion } = useRequestProtocol();
+  const onSuccess = (comment: Comment) => {
+    if (!comment) return;
+
+    const comments = ['post_comments', postId];
```
rebelchris, line 79: This is the key right? could apply same comment about generic key

packages/shared/src/components/fields/MarkdownInput/CommentMarkdownInput.tsx
```diff
@@ -0,0 +1,209 @@
...
+    if (editCommentId) {
+      return editComment({ id: editCommentId, content });
+    }
+
+    return onComment({ content, id: parentCommentId ?? postId });
+  };
+
+  useEffect(() => {
+    markdownRef?.current?.textareaRef?.current?.focus();
+  }, []);
```
rebelchris, line 173: Does this need to be in a useEffect? Think you can even simply just call if in the render

packages/shared/src/hooks/companion/common.ts
```diff
@@ -0,0 +1,4 @@
+export const getCompanion = (): HTMLElement =>
```
rebelchris, line 1: I'm not really sure why this makes sense in shared if only extension can ever use it?

rebelchris (commented):
Wow, must have been a big challenge!
Testing it seems fine, do have some concerns about moving extension specific code, what's the reasoning for it?

packages/shared/src/hooks/companion/common.ts
sshanzel: Right. Let me see how I can just pass the requirement of this one.

sshanzel:
Commits for analytics:
https://github.com/dailydotdev/apps/pull/1881/commits/16da914f9f2022fd7b744291583fa7a2d3b3f901
https://github.com/dailydotdev/apps/pull/1881/commits/4e23eef1b6d5a0cedc9d80037307a99c000ab383

packages/shared/src/components/fields/MarkdownInput/CommentMarkdownInput.tsx
sshanzel: Right. Let's use that one 🚢

packages/shared/src/components/fields/MarkdownInput/CommentMarkdownInput.tsx
sshanzel: Do you mean being inside the `ref` callback on the form element?

packages/shared/src/hooks/companion/common.ts
sshanzel: There will be a lot of prop drilling to happen if we don't host it here. Another option is to another property in the `useRequestProtocol` on where to append tooltips. Though it feels out of context for the said hook by its name.

Probably a new hook that sets it across the app? Would also help a lot of other areas (mostly on modals) to not think about it anymore whenever working with tooltips/popups. WDYT?

sshanzel: @rebelchris would love to hear what you think on this one.

rebelchris: I personally think anything "extension" "companion" only should be in that repo, not in shared.
It shouldn't include prop drilling, but maybe I'm not getting the right picture here?

packages/shared/src/hooks/post/useComments.ts
```diff
@@ -0,0 +1,64 @@
...
+import { useAuthContext } from '../../contexts/AuthContext';
+import { isNullOrUndefined } from '../../lib/func';
+import { useAnalyticsContext } from '../../contexts/AnalyticsContext';
+import { postAnalyticsEvent } from '../../lib/feed';
+import { AnalyticsEvent, Origin } from '../../lib/analytics';
+import { Post } from '../../graphql/posts';
+
+type ReplyTo = [Comment, string, boolean?];
+
+const initialState: ReplyTo = [null, null, false];
```
rebelchris, line 13: What are these values, not super clear from a review point of view.

packages/shared/src/hooks/companion/common.ts
sshanzel: When the app is Companion, we need to append the tooltips to the ShadowRoot in order to display the right content. We have a preview tab in the comment box where the tooltip is used when hovering a mention preview, this same shared component is used for making a comment in the companion. That means we will have to drill down the companion wrapper from the app down to the comment box because the shared component itself will append it to the default value which is the `document`.

rebelchris (changes requested):
It's quite a massive PR with a lot going on, finding it a bit hard to review the actual changes.
I did test the frontend which seems alright.

Main question: Why do we place extension only code in shared, that part I don't really get.
As it's quite big and has tracking events changes, also asked @idoshamun to check if he has time.

packages/shared/src/hooks/companion/common.ts
rebelchris: Well yes, but what does it matter if this code sits in "extension" or "shared"?

packages/shared/src/hooks/post/useComments.ts
sshanzel: The initial state of the ReplyTo object. We have a state we use in the component and this is the initial value. We want to be able to return to this initial state at some point so instead of explicitly doing it twice, we are keeping it in a variable.

sshanzel:
> Why do we place extension only code in shared, that part I don't really get

We abstracted a lot of functionalities in the `CommentMarkdownInput` which consists of doing requests. Each and every markdown input will have its own watcher for the requests such as preview, etc., this is to avoid missing any request by defining each watcher on their own one by one (combination of input, post id, comment id, etc), we instead support it out of the box. This will ensure we are doing the necessary cache updates whenever any of those watchers receive a response.

Maintenance-wise, we would benefit too as there are a lot of things to consider just to make them tie up together. And another change similar to this would make us go back and re-write a lot of things. The changes here can avoid that.

packages/shared/src/hooks/companion/common.ts
sshanzel: So the shared can use it too since it would be then a dependency cycle if the `shared` depends on the `extension`.

packages/shared/src/hooks/post/useComments.ts
sshanzel: Ah, right. Let me type it up to make it explicit.

idoshamun:
@rebelchris @sshanzel I reviewed only the analytics part and it seems great!

sshanzel:
Finally found the (green) light at the end of the tunnel 😭

rebelchris (approved):
What a tremendous amount of work Lee.
Sorry took me a while to go over everything.

It looks good on the high level, have one question about memo, but non blocking for now, just check it please.

Also if you go live with this, take some time to test it in production please 🙏

packages/shared/src/hooks/post/useComments.ts
```diff
@@ -0,0 +1,66 @@
...
+        : replyComment?.author.username;
+    const replyToContent = replyToUsername ? `@${replyToUsername} ` : undefined;
+
+    return {
+      replyTo: replyToUsername,
+      initialContent: isEdit ? replyComment?.content : replyToContent,
+      editCommentId: isEdit ? replyComment?.id : null,
+      parentCommentId: parentId,
+    };
+  }, [user, replyTo]);
```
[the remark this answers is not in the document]
sshanzel: The `isEdit` was defined inside the function itself, so we don't have to make it a dependency 🚢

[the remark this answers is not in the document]
rebelchris: Ah yes sorry didn't see it in the array there 😅

rebelchris (approved):

sshanzel:
> Also if you go live with this, take some time to test it in production please 🙏

Yes. I will do an end-to-end tests from preview, then once merged, I will do it on production too since the flow was completely changed. Woul also send a message in our general channel to watch out for any quirky behavior.

PR #1891, opened 2023-06-20
fix: username generation on empty username

rebelchris:
- Fix for username generation to not fire on manual emptying username

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1464 #done

### DIFF
--- a/packages/shared/src/hooks/useGenerateUsername.ts
+++ b/packages/shared/src/hooks/useGenerateUsername.ts
@@ -1,5 +1,5 @@
 import { useQuery } from 'react-query';
-import { useState, useEffect } from 'react';
+import { useState, useEffect, useRef } from 'react';
 import { useRequestProtocol } from './useRequestProtocol';
 import { GET_USERNAME_SUGGESTION } from '../graphql/users';
 import { graphqlUrl } from '../lib/config';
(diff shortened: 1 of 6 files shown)


idoshamun (approved):

idoshamun (commented):
@rebelchris another idea is to remove the `{ name }` from the query key. This will make sure it doesn't get dispatched more then once right?

packages/shared/src/hooks/useGenerateUsername.ts
```diff
@@ -13,9 +13,10 @@ export const useGenerateUsername = (
   name: string | undefined,
 ): UseGenerateUsername => {
   const [username, setUsername] = useState('');
+  const usernameRef = useRef(false);
   const { requestMethod } = useRequestProtocol();
-  const usernameQueryKey = ['generateUsername', name];
-  const { data, isLoading } = useQuery<{
+  const usernameQueryKey = ['generateUsername'];
```
capJavert, line 18: If this hook was re-used in another part of our webapp it would then have the same value as the first call even though the actual hook would be called with the new `name`, which would be wrong.

capJavert (changes requested):

packages/shared/src/hooks/useGenerateUsername.ts
rebelchris: @capJavert Let me know what you think of the change

capJavert: I think it has to include it because we built this hook as a feature hook that should be re-usable across the webapp. Someone could use it down the line and encounter bugs without even knowing.

idoshamun (approved):
Let's do it!

PR #1894, opened 2023-06-21
fix: create cache key for pnpm lock file

rebelchris:
- Make cache dependant on pnpm lock file
## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -12,14 +12,14 @@ jobs:
           command: npm i --prefix=$HOME/.local -g pnpm
       - restore_cache:
           keys:
-            - deps-v5-{{ checksum "pnpm-lock.yaml" }}
-            - deps-v5-{{ .Branch }}
-            - deps-v5-
+            - deps-v6-{{ checksum "pnpm-lock.yaml" }}
+            - deps-v6-{{ .Branch }}
+            - deps-v6-
       - run:
           name: Install dependencies
           command: pnpm install
       - save_cache:
-          key: deps-v5-{{ checksum "pnpm-lock.yaml" }}
+          key: deps-v6-{{ checksum "pnpm-lock.yaml" }}
           paths:
             - ./node_modules
             - ./packages/eslint-config/node_modules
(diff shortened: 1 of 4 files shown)


idoshamun (approved):

.circleci/config.yml
```diff
@@ -12,13 +12,14 @@ jobs:
           command: npm i --prefix=$HOME/.local -g pnpm
       - restore_cache:
           keys:
+            - deps-v5-{{ checksum "pnpm-lock.yaml" }}
             - deps-v5-{{ .Branch }}
             - deps-v5-
```
capJavert, line 17: I think we don't need this anymore, it will save on some HTTP requests in ci/cd.

capJavert (approved):

.circleci/config.yml
idoshamun: Why? I think it can reduce some time

capJavert: We don't use those cache keys anywhere else, so my logic is why would we add them here for restoration?

rebelchris: I also think it's like a fallback mechanism for cache right?
To me it's like react query keys that work in a "array" like way

But could be totally wrong.

rebelchris: Yeah pretty sure that's [how](https://circleci.com/docs/caching/#restoring-cache) it works.

rebelchris: Although with that said it will now default to use "v5" cache so maybe it's time to bump so it will use v6 as we made massive changes 😅

capJavert: It works like that but we don't have any other caches aside from `deps-v6-{{ checksum "pnpm-lock.yaml" }}`.

So for sure we don't need `deps-v5-{{ .Branch }}` because it does not exist anymore.

We can leave `deps-v5-` as a wildcard sure but in that case each job would pull more caches then it needs to which would increase job running time since it is unpacking more cache objects.

idoshamun (approved):

PR #1896, opened 2023-06-23
feat: update post navigation per design changes

capJavert:
Before:
image
After:
image

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1447 #done


packages/shared/src/components/post/PostNavigation.tsx
```diff
@@ -50,7 +50,7 @@ function PostNavigation({
       {onPreviousPost && (
         <SimpleTooltip content="Previous">
           <Button
-            className="-rotate-90 btn-secondary"
+            className="border-none -rotate-90 btn-secondary"
```
rebelchris, line 53:
```suggestion
            className="-rotate-90 btn-tertiary"
```

I think this should do looking at the design spec.

packages/shared/src/components/post/PostNavigation.tsx
```diff
@@ -62,7 +62,7 @@ function PostNavigation({
       {onNextPost && (
         <SimpleTooltip content="Next">
           <Button
-            className="rotate-90 btn-secondary"
+            className="border-none rotate-90 btn-secondary"
```
rebelchris, line 65:
```suggestion
            className="rotate-90 btn-tertiary"
```

rebelchris (changes requested):

packages/shared/src/components/post/PostNavigation.tsx
capJavert: Ah sorry about that. Updated. I would expect `Ter` to be up here also, but it is not. 🤦 

image

rebelchris (approved):
Wohoo, you can probably even go directly to main for this one actually doesn't seem breaking changes 👍

PR #1897, opened 2023-06-23
feat: downvote

capJavert:
- [DR for the feature](https://dailydotdev.atlassian.net/l/cp/LCSbMFqW)
- Available on [preview](https://preview.app.daily.dev/)

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1121 #done


packages/shared/src/components/cards/ArticlePostCard.tsx
```diff
@@ -37,10 +39,21 @@ export const ArticlePostCard = forwardRef(function PostCard(
...
   const onPostCardClick = () => onPostClick(post);
   const { trending, pinnedAt } = post;
   const customStyle = !showImage ? { minHeight: '15.125rem' } : {};
 
+  if (data?.showTagsPanel && post.tags.length > 0) {
+    return (
+      <PostTagsPanel
+        className="overflow-hidden h-full max-h-[23.5rem]"
+        post={post}
+        toastOnSuccess
```
rebelchris, line 52: What makes this difference for a list or card?

packages/shared/src/components/post/block/PostTagsPanel.tsx
```diff
@@ -0,0 +1,122 @@
...
+
+export function PostTagsPanel({
+  post,
+  className,
+  toastOnSuccess = true,
+}: PostTagsPanelProps): ReactElement {
+  const { feedSettings } = useFeedSettings();
+  const hasBlockedSource = () =>
+    feedSettings?.excludeSources?.some(({ id }) => id === post.source.id);
+  const [initialPreference] = useState(hasBlockedSource);
```
rebelchris, line 27: Do you need a useState if there is no setter?
Seems like `hasBlockedSource` itself would do it right?

packages/shared/src/components/post/block/PostTagsPanel.tsx
```diff
@@ -0,0 +1,122 @@
...
+      >
+        <Button
+          className={shouldBlockSource ? 'btn-primary' : 'btn-tertiaryFloat'}
+          buttonSize={ButtonSize.Small}
+          icon={<SourceAvatar source={post.source} />}
+          onClick={() => setShouldBlockSource(!shouldBlockSource)}
+        >
+          {post.source.name}
+        </Button>
+        {post.tags.map((tag) => (
```
rebelchris, line 97: Not urgent, but wanted to say we use this often, potentially could be a `TagButton` component.

packages/shared/src/components/widgets/FurtherReading.tsx
```diff
@@ -116,9 +116,15 @@ export default function FurtherReading({
...
-      postAnalyticsEvent(postEventName({ bookmarked }), post, {
-        extra: { origin: 'recommendation' },
-      }),
+      postAnalyticsEvent(
+        bookmarked
+          ? AnalyticsEvent.BookmarkPost
+          : AnalyticsEvent.RemovePostBookmark,
+        post,
+        {
+          extra: { origin: 'recommendation' },
```
rebelchris, line 125: This should be a defined enum origin (probably legacy issue)

packages/shared/src/hooks/post/useBlockPost.ts
```diff
@@ -0,0 +1,239 @@
...
+      container.push(tag);
+    });
+  }
+
+  return { blocks, unblocks };
+};
+
+const ignoredCall = () => Promise.resolve({ successful: true });
+
+export const useBlockPost = (
```
rebelchris, line 63: From review point of view as someone looking at first time.
- Very big file, lots of things going on, not super clear what's happening really.

Ideally it would be either better named, or split. (non blocking comment)

rebelchris (changes requested):
Looks great guys,
got some minor things, and non blocking comments (maybe for future discussions?)

packages/shared/src/components/cards/ArticlePostCard.tsx
sshanzel: They should be the same, the only differences would probably the className.

packages/shared/src/components/post/block/PostTagsPanel.tsx
sshanzel: Yes. We need it here so it never changes, because once they have clicked the "Block" button, they would then have the updated value if it would come from the function.

rebelchris: Oh I see :) Ok

packages/shared/src/components/cards/ArticlePostCard.tsx
rebelchris: The one has `toastOnSuccess` the other doesn't please check if that's alright then 🙏

sshanzel: Oh, might have been missed. Yep, they should both have the same as the don't have the panel to Undo the changes.

packages/shared/src/components/post/block/PostTagsPanel.tsx
sshanzel: Ah, right. I've been looking for this. I ended up seeing the `TagLinks` which I saw was probably an outdated implementation.

We can definitely go with `TagButton`.

packages/shared/src/hooks/post/useBlockPostPanel.ts
```diff
@@ -0,0 +1,273 @@
...
+            return;
+          }
+
+          trackEvent(
+            postAnalyticsEvent('report post', reportedPost, {
+              extra: { origin: Origin.TagsFilter },
+            }),
+          );
+        },
+        postIndex: -1,
```
rebelchris, line 198: This feels a bit weird to me, can we not make it optional and not send it if it doesn't exist?
We don't use it for this case so it feels very weird.

Do we even need it for the old flow?

packages/shared/src/hooks/post/useBlockPostPanel.ts
```diff
@@ -0,0 +1,273 @@
...
+    () => setShowTagsPanel({ showTagsPanel: true }),
+    [setShowTagsPanel],
+  );
+
+  const onDismissPermanently = useCallback(() => {
+    completeAction(ActionType.HideBlockPanel);
+    setShowTagsPanel({ showTagsPanel: undefined });
+  }, [completeAction, setShowTagsPanel]);
+
+  const onReport = useCallback(() => {
```
rebelchris, line 168: Just a generic comment, we now use a mix of lazy modal and the direct modal, I'd suggest maybe making one generic usage if possible.
So converting the old one to use lazyModal as well.

And if that's the case we can probably move this onReport logic to a custom hook so it can be handled from one place? 

I'd prefer this in case we decide to add something (tags is good example) or tracking needs an addition change.

Love to hear what you think.

rebelchris (changes requested):
Looks amazing, got another suggestion for uniforming the codebase a little bit.

packages/shared/src/hooks/post/useBlockPostPanel.ts
sshanzel: Agreed. After checking, it seems we can actually make it simpler by executing the report post from the modal instead of having a callback. We do keep the callback though, but only for cases where you need it, and not should be a required prop from the modal itself.

rebelchris: Looks great, great modification 👍

rebelchris (approved):
Think we're good to go 🙏

PR #1898, opened 2023-06-23
feat(eslint): updated no restricted globals rule

capJavert:

Reminded myself on potential update to our eslint, I just had this piece of code (image below) for event tracking.

Even though I did not have origin prop or variable in context there was no JS or linter error and that is because window.origin and origin is a global var in JS :smile:

Updating our `no-restricted-globals` for some of those globals to avoid errors. We already had some from extended configs so I moved them all to our config so we don't overwrite.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done


packages/shared/src/contexts/OnboardingContext.tsx
```diff
@@ -85,7 +85,7 @@ export const OnboardingContextProvider = ({
     trackEvent({
       event_name: AnalyticsEvent.ClickArticleAnonymousCTA,
       target_id: ExperimentWinner.ArticleOnboarding,
-      extra: JSON.stringify({ origin }),
+      extra: JSON.stringify({ origin: window.origin }),
```
rebelchris, line 88: Just a note, origin is actually never the "origin" as in http origin.
It's more an internal measure.

In this case it should be one of these two according to analytics:
```
article page, article modal
```

I think maybe by accident this got lost?

rebelchris (changes requested):
Like the addition, however the tracking change is not according to analytics doc.

packages/shared/src/contexts/OnboardingContext.tsx
capJavert: In this case it was passed as `window.origin` for quite some time. So this is a bug?

capJavert: Because there is no `origin` prop or variable in `OnboardingContext`

rebelchris: Yep 100% a bug, should not be the case.
I'll also check with Ido how it impacted analytics.

idoshamun: Below are the correct values, but there's no significant impact anyway at this point.
Thanks for pointing that out. Let's fix that when possible (not urgent).
image

rebelchris: Thanks for checking ill add it as a seperate task then 👍

capJavert: @rebelchris we can merge this then?

rebelchris (approved):

PR #1900, opened 2023-06-23
fix: downvote option modal state

capJavert:
- when clicking downvote from modal option menu mutation handlers were missing and state would go out of sync with upvote/downvote buttons in the post modal actions

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done

### DIFF
--- a/packages/shared/src/hooks/useUpdatePost.ts
+++ b/packages/shared/src/hooks/useUpdatePost.ts
@@ -16,6 +16,13 @@ export default function useUpdatePost(): UseBookmarkPostRet {
       const postQueryKey = ['post', props.id];
       await client.cancelQueries(postQueryKey);
       const oldPost = client.getQueryData<PostData>(postQueryKey);
+
+      if (!oldPost) {
+        // nothing in cache so we skip any update
+
+        return () => undefined;
+      }
+
       client.setQueryData<PostData>(postQueryKey, {
         post: {
           ...oldPost.post,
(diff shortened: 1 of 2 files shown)


capJavert:
@rebelchris let me know if this also fixes your issue from discussion on slack 🙏

packages/shared/src/hooks/useUpdatePost.ts
```diff
@@ -16,6 +16,13 @@ export default function useUpdatePost(): UseBookmarkPostRet {
       const postQueryKey = ['post', props.id];
       await client.cancelQueries(postQueryKey);
       const oldPost = client.getQueryData<PostData>(postQueryKey);
+
+      if (!oldPost) {
+        // nothing in cache so we skip any update
+
+        return () => undefined;
```
rebelchris, line 23: Would a `return;` not be enough here?

rebelchris (approved):
Minor question, but otherwise good to go.

packages/shared/src/hooks/useUpdatePost.ts
capJavert: We use this return value for rollback in `mutation.onError` so want to avoid and `undefined()` calls which would break parts of the app.

rebelchris: Alright then

PR #1902, opened 2023-06-25
feat: fetch the first image from the content if no banner image set for a squad post

TomDowling:
Search the post content for the first image and render it if no banner image is set for the squad post
post
listing

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1424 #done


packages/shared/src/components/cards/WelcomePostCardFooter.tsx
```diff
@@ -11,17 +11,32 @@ type WelcomePostCardFooterProps = {
...
 
-  if (post.image) {
+  const imageCheck = useMemo(() => {
+    if (image) return true;
+
+    const parser = new DOMParser();
+    const doc = parser.parseFromString(post?.contentHtml, 'text/html');
+    const imgTag = doc.querySelector('img');
+    if (imgTag) {
+      setImage(imgTag.getAttribute('src') || null);
```
capJavert, line 28: You don't need state here, state setting from `useMemo` is anti-pattern because it is the same as setting state during rendering.

You can just return the `image` from `useMemo` and use it below for boolean check.

packages/shared/src/components/cards/WelcomePostCardFooter.tsx
```diff
@@ -11,17 +11,32 @@ type WelcomePostCardFooterProps = {
...
+      return true;
+    }
+
+    return false;
+  }, [post?.contentHtml, image]);
+
+  if (imageCheck) {
```
capJavert, line 35: Just use `!!image` which you return from `useMemo` per my first comment.

capJavert (changes requested):
Looks good feature wise, one optimization.

packages/shared/src/components/cards/WelcomePostCardFooter.tsx
TomDowling: Very good point, iv'e updated it to remove the useState and just use useMemo instead

packages/shared/src/components/cards/WelcomePostCardFooter.tsx
```diff
@@ -17,15 +17,31 @@ export const WelcomePostCardFooter = ({
...
+    if (post?.image) return post?.image;
+
+    const parser = new DOMParser();
+    const doc = parser.parseFromString(post?.contentHtml, 'text/html');
+    const imgTag = doc.querySelector('img');
+    if (imgTag) {
+      return imgTag.getAttribute('src') || null;
+    }
+
+    return null;
```
capJavert, line 30: No need to return `null`, `undefined` is just fine, also above with `|| null`.

packages/shared/src/components/cards/WelcomePostCardFooter.tsx
```diff
@@ -17,15 +17,31 @@ export const WelcomePostCardFooter = ({
     [post?.contentHtml],
   );
 
-  if (post.image) {
+  const image = useMemo(() => {
+    if (post?.image) return post?.image;
+
+    const parser = new DOMParser();
+    const doc = parser.parseFromString(post?.contentHtml, 'text/html');
```
capJavert, line 24: Also non blocking but just make sure if for ever reasons `post?.contentHtml` is `undefined` we want to make sure that `parseFromString` does not break.

capJavert (approved):
Looking good 💪 .. few more non blocking comments, feel free to merge after.

packages/shared/src/components/cards/WelcomePostCardFooter.tsx
TomDowling: I have made this change

packages/shared/src/components/cards/WelcomePostCardFooter.tsx
TomDowling: Have tested and it doesn't break if it's undefined

PR #1903, opened 2023-06-26
feat: Adding logic  to show shared items that don't have a title

TomDowling:
## Changes
Added logic to set the article page title if the article is shared and has no title.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1460 #done


packages/webapp/pages/posts/[id]/index.tsx
```diff
@@ -108,8 +108,16 @@ const PostPage = ({ id, initialData }: Props): ReactElement => {
       sidebarRendered &&
       modalSizeToClassName[ModalSize.Large],
   );
+  const seoTitle = () => {
+    if (
+      post?.type !== PostType.Share ||
+      (post?.type === PostType.Share && post?.title !== null)
+    )
+      return post?.title;
+    return `Shared post at ${post?.source?.name}`;
```
Comment (sshanzel) [inline packages/webapp/pages/posts/[id]/index.tsx:117]: Probably just a personal preference, but I think it is better to isolate the odd one out.

```suggestion
    if (post?.type === PostType.Share && post?.title === null) return `Shared post at ${post?.source?.name}`;
    
    return post?.title;
```

sshanzel (approved):
Looks amazing! 🚢 

Just had one non-blocking comment.

packages/webapp/pages/posts/[id]/index.tsx
Comment (TomDowling) [reply packages/webapp/pages/posts/[id]/index.tsx:117]: I did have it like this initially, but my concern was when there wasn't a title but not Shared, maybe there isn't a use case now but down the line "Shared post at" might not be relevant text to use in these use cases? @sshanzel, however, if you think thats okay i will make this change :)

Comment (rebelchris) [reply packages/webapp/pages/posts/[id]/index.tsx:117]: I think Lee's suggestion is ok, anything non squads should have mandatory title these days for it to become visible.
Let's go with this, if it becomes an issue we can always address it again 🙏

rebelchris (approved):

PR #1909, opened 2023-06-26
feat: block from modal

sshanzel:
## Changes
- Block tags panel when the user downvoted a post.
- Migrated the report modal to a lazymodal for ease of access.
- Once validated, I will raise a separate PR utilizing the components here on the PostCard.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1445 #done


packages/shared/src/hooks/post/useBlockPost.ts
```diff
@@ -0,0 +1,153 @@
...
+        shouldBlockSource,
+      );
+
+      if (!successful) return;
+
+      setShowTagsPanel({
+        showTagsPanel: toastOnSuccess ? undefined : false,
+        blocked: { tags, sourceIncluded: shouldBlockSource },
+      });
+    },
```
rebelchris, line 151: Why is this an inline function, would it not be better to abstract above as useCallback maybe?

rebelchris (commented):
One more generic question, but love to see Ante's feedback on this as well.

packages/shared/src/hooks/useVotePost.ts
```diff
@@ -110,6 +112,7 @@ const useVotePost = <T extends { id: string } = { id: string }>({
       mutationKey: downvotePostMutationKey,
       onMutate: onDownvotePostMutate,
       onError: (err, _, rollback) => rollback?.(),
+      onSuccess: onDownvoted,
```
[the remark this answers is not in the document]
capJavert: I understand why is it here, Lee wanted to make sure the server side call actually passed before offering user blocking options.

That said, since we do have optimistic UI here I think we might want to just show the blocking UI `onMutate` since there could be a delay until server call and user could already navigate elsewhere. Downvote and blocking are not connected so no harm in having one or the other. So I would say we do not need `onDownvoted` here currently.

packages/shared/src/hooks/post/useBlockPost.ts
capJavert: Agreed, same for other functions, this is crucial so they can be used efficiently in other components and hooks.

packages/shared/src/graphql/actions.ts
```diff
@@ -12,6 +12,7 @@ export enum ActionType {
   BrowserExtension = 'browser_extension',
   EnableNotification = 'enable_notification',
   WritePost = 'write_post',
+  HideBlockPanel = 'hide_block_panel',
```
capJavert, line 15: We need to add this to `UserActionType` on API side to keep enums in sync.

packages/shared/src/hooks/post/useBlockPost.ts
```diff
@@ -0,0 +1,153 @@
...
+    client.setQueryData<BlockData>(key, (current) => ({
+      ...current,
+      ...params,
+    }));
+
+  const value = useMemo<BlockData>(
+    () => ({
+      ...data,
+      showTagsPanel: checkHasCompleted(ActionType.HideBlockPanel)
+        ? undefined
```
capJavert, line 82: `false` would make more sense?

packages/shared/src/hooks/post/useBlockPost.ts
```diff
@@ -0,0 +1,153 @@
...
+  const client = useQueryClient();
+  const { user } = useAuthContext();
+  const { checkHasCompleted, completeAction } = useActions();
+  const key = generateStorageKey(
+    StorageTopic.Post,
+    `block:${post?.id}`,
+    user?.id,
+  );
+  const { data } = useQuery<BlockData>(key, () => client.getQueryData(key), {
+    initialData: {},
```
capJavert, line 70: Good to add `disabledRefetch` options just in case.

packages/shared/src/components/post/block/PostTagsPanel.tsx
```diff
@@ -0,0 +1,111 @@
...
+            onClick={() => setTags({ ...tags, [tag]: !tags[tag] })}
+          >
+            #{tag}
+          </Button>
+        ))}
+      </span>
+      <span className="flex flex-row gap-2 p-3 -mx-4 mt-4 border-t border-theme-divider-tertiary">
+        <Button className="ml-auto btn-tertiary" onClick={onReport}>
+          Report
+        </Button>
```
capJavert, line 101: While testing I noticed clicking this just dismisses the panel, should it open reporting modal?

packages/shared/src/components/post/block/PostTagsPanel.tsx
```diff
@@ -0,0 +1,111 @@
...
+}: PostTagsPanelProps): ReactElement {
+  const { feedSettings } = useFeedSettings();
+  const [shouldBlockSource, setShouldBlockSource] = useState(
+    feedSettings?.excludeSources?.some(({ id }) => id === post.source.id),
+  );
+  const [tags, setTags] = useState<BlockTagSelection>(
+    feedSettings?.blockedTags?.reduce(
+      (block, tag) => ({ ...block, [tag]: true }),
+      {},
+    ),
```
capJavert, line 32: Since this is iteration using function to create initial value for `useState` is better so it does not run this `reduce` on each re-render.

capJavert: Same for `shouldBlockSource` above.

capJavert (changes requested):
Looking solid 👌 , some comments for improvement on the rendering side and one issue I found during testing.

capJavert (changes requested):
Also, if we could add some tests for this in `PostPage` I think it would be great since it is going to be used a lot by our users. 🙏

packages/shared/src/hooks/post/useBlockPost.ts
sshanzel: `undefined` is more for actually not showing anything. False is for the after-message "5 topics blocked".

packages/shared/src/components/post/block/PostTagsPanel.tsx
sshanzel: Oh, nice catch. I've never know that.

packages/shared/src/components/post/block/PostTagsPanel.tsx
sshanzel: Right. I missed this in my commits. This should be migrated to a lazymodal.

packages/shared/src/hooks/useVotePost.ts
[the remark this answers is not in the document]
sshanzel: > This is kind of what the onMutate handles no?

Yes, just to make sure it has been completed.

Ante's comment makes sense. @rebelchris WDYT about it?

packages/shared/src/hooks/post/useBlockPost.ts
sshanzel: Yep. Was actually hesitant, but yeah, since these are functions from a hook, they better be stable ones.

packages/shared/src/hooks/useVotePost.ts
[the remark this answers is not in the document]
rebelchris: I think we always favor optimistic UI, so would put it all into the mutation handler.
We should aim to trust our calls will succeed basically.

packages/shared/src/graphql/actions.ts
sshanzel: Currently unused in the backend. Not sure if we have to. @rebelchris WDYT?

sshanzel:
@rebelchris @capJavert should be ready for another round of review, just working on to add more tests.

capJavert (approved):
🚀 Don't have any further comments except that we possibly miss some tests for new blocking UI for `PostPage`.

sshanzel:
@capJavert have added the tests now 🥳

capJavert (approved):
@sshanzel gg wp 🥳

PR #1913, opened 2023-06-27
feat: add setCookie utility

capJavert:
- add `setCookie`
  - since we have `getCookieObject` implementation it was logical to implement our own setter utility
- on the server we should use native req.cookies
- base for implementation https://developer.mozilla.org/en-US/docs/Web/API/Document/cookie
  - also looked at other implementations like https://www.npmjs.com/package/cookie which are mostly bloated but did include same edge cases I included in our utility
- refactor usages of `document.cookie` to use `setCookie`
  - tested cookies were set correctly when reading on client and server (in boot)

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1432 #done


packages/shared/src/hooks/useCookieBanner.ts
```diff
@@ -12,9 +13,12 @@ export function useCookieBanner(): [
...
     setShowCookie(false);
-    document.cookie = `${consentCookieName}=true;path=/;domain=${
-      process.env.NEXT_PUBLIC_DOMAIN
-    };samesite=lax;expires=${60 * 60 * 24 * 365 * 10}`;
+    setCookie(consentCookieName, true, {
+      path: '/',
+      domain: process.env.NEXT_PUBLIC_DOMAIN,
+      sameSite: 'lax',
+      maxAge: 60 * 60 * 24 * 365 * 10,
+    });
```
capJavert, line 21: I think we had a bug in our code where our cookie banner `ilikecookies` cookie was always a session cookie. 

This was due to `expires` being set as `number` of seconds and per spec it should be date formatted string. 

I replaced it with `max-age` which correctly sets expiration of cookie in seconds (in this case 10 years). 

We did not notice this because nowadays browsers do not clear session cookies until their process is closed in full vs closing a tab.

rebelchris (approved):
Thanks for picking this up, looks neat.
Just maybe wait for approval from Lee just in case 🙏

capJavert:
@sshanzel could you also take a look if you have time?

packages/shared/src/lib/cookie.ts
```diff
@@ -35,6 +35,53 @@ export const getCookieFeatureFlags = (): Record<string, FeatureValue> => {
...
+    partitioned: options.partitioned,
+    path: options.path,
+    samesite: options.sameSite,
+    secure: options.secure,
+  };
+
+  const cookieValue: string = Object.keys(parsedOptions).reduce((acc, key) => {
+    const option = parsedOptions[key];
+
+    if (typeof option === 'undefined') {
```
sshanzel, line 71: I think we can use the `isNullOrUndefined`.

packages/shared/src/hooks/useCookieBanner.ts
sshanzel: Amazing. Thanks for the retro fix. I think we can introduce a variable named `const TEN_YEARS` or something similar, to make easier to understand in one look.

packages/shared/src/lib/cookie.ts
```diff
@@ -35,6 +35,53 @@ export const getCookieFeatureFlags = (): Record<string, FeatureValue> => {
...
+  options: Partial<{
+    domain: string;
+    expires: Date | string;
+    maxAge: number;
+    partitioned: boolean;
+    path: string;
+    sameSite: 'lax' | 'strict' | 'none';
+    secure: boolean;
+  }> = {},
+): void => {
```
sshanzel, line 50: Just some non-blocking cleanups.
```suggestion
interface CookieOptions {
  domain: string;
  expires: Date | string;
  maxAge: number;
  partitioned: boolean;
  path: string;
  sameSite: 'lax' | 'strict' | 'none';
  secure: boolean;
}

export const setCookie = (
  name: string,
  value: string | number | boolean,
  options: Partial<CookieOptions> = {},
): void => {
```

packages/webapp/pages/join.tsx
```diff
@@ -64,15 +65,12 @@ const Page = ({ referringUser, campaign }: PageProps): ReactElement => {
...
-      `join_referral=${referringUser.id}:${campaign}`,
-      `Max-Age=${1 * 365 * 24 * 60 * 60}`,
-      isDevelopment ? undefined : 'Secure',
-      `Domain=${process.env.NEXT_PUBLIC_DOMAIN}`,
-      `SameSite=Lax`,
-    ]
-      .filter(Boolean)
-      .join('; ');
+    setCookie('join_referral', `${referringUser.id}:${campaign}`, {
+      maxAge: 1 * 365 * 24 * 60 * 60,
```
sshanzel, line 69: Also this, I think we can set a variable somewhere.

sshanzel (approved):
Wow, great work. Amazing to see you added the test cases too.

Just had non-blocking comments. Other than that, everything looks great.

packages/shared/src/hooks/useCookieBanner.ts
capJavert: @rebelchris @sshanzel one fun fact for Chrome https://developer.chrome.com/blog/cookie-max-age-expires/#:~:text=With%20this%20change%2C%20Chrome%20caps,set%20to%20400%20days%20instead

I think we can still leave 10 years for other browsers but something to keep in mind.

PR #1918, opened 2023-06-29
fix: avoid modal close on tags/source block

capJavert:

So the issue with article modal auto closing was only happening on "My feed" when blocking tag or source from the modal. All other actions like hide and report do not close modal (which is good ✅ ).

This PR does below changes to enable selective invalidation of feed queries when blocking tags/sources:
- move logic to invalidate feed queries on tags/settings change to `useTagAndSource` hook
  - this proved to be smallest footprint change which fixes the underlying issue
- remove `useFeedSettings` useEffect that was triggering too often (and unused options/state related to it)
  - invalidation is called from actions inside `useTagAndSource` directly
- add `shouldInvalidateQueries` to `useTagAndSource` since that hook is responsible for actual actions execution
- disable feed queries invalidation from `PostOptionsMenu` to prevent modal close on blocking

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done

### DIFF
--- a/packages/shared/src/components/PostOptionsMenu.tsx
+++ b/packages/shared/src/components/PostOptionsMenu.tsx
@@ -96,6 +96,7 @@ export default function PostOptionsMenu({
   } = useTagAndSource({
     origin: Origin.PostContextMenu,
     postId: post?.id,
+    shouldInvalidateQueries: false,
   });
   const [reportModal, setReportModal] = useState<{
     index?: number;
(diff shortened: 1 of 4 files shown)


rebelchris (approved):
Looks great 🙏

sshanzel (approved):
I'm a simple man. I see `useEffect` removed, I approve 🤣 

Kidding aside, I think we should also add the `false` value on the usage of `useTagAndSource` at the new hook named `useBlockPost`.

PR #1927, opened 2023-07-03
feat: adding enableSourceHeader to squad PostCards to render differently based on location

TomDowling:
## Changes
Changed the feedName for the squad page and then on Squad Post items render a different header based on if you are on the squad page or elsewhere. Screenshots below of the two types of render.

Screenshot 2023-07-03 at 10 43 23 am
Screenshot 2023-07-03 at 10 43 10 am

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1496 #done


rebelchris (changes requested):
Looks good, only from the preview I think on Sketch it states outside squad variant should show "user's name" not the handle with @ in front.

rebelchris (approved):
Looks good to me.
@capJavert or @sshanzel could you also quickly check if this PR is good to go 🙏

packages/shared/src/components/cards/WelcomePostCardHeader.tsx
```diff
@@ -1,32 +1,44 @@
...
+type WelcomePostCardHeaderProps = Pick<
+  Post,
+  'author' | 'source' | 'createdAt'
+> & { enableSourceHeader?: boolean };
 
 export const WelcomePostCardHeader = ({
   author,
   createdAt,
   source,
+  enableSourceHeader,
```
capJavert, line 17: I would default it to `enableSourceHeader = false` just to keep it boolean always.

packages/shared/src/components/cards/WelcomePostCard.tsx
```diff
@@ -26,6 +26,7 @@ export const WelcomePostCard = forwardRef(function SharePostCard(
     children,
     style,
     onReadArticleClick,
+    enableSourceHeader,
```
capJavert, line 29: I would default it to enableSourceHeader = false just to keep it boolean always.

capJavert (changes requested):
Looks good, minor comments.

We are missing this for share posts though. It should apply to all squad posts not just welcome and freeform as it is now (since freeform uses the same card). https://dailydotdev.atlassian.net/wiki/spaces/HAN/pages/554074122/WT-1454+Squads+MVP+open+squads#Frontend

packages/shared/src/components/cards/WelcomePostCardHeader.tsx
```diff
@@ -1,32 +1,44 @@
...
 }: WelcomePostCardHeaderProps): ReactElement => {
   return (
     <div className="flex relative flex-row gap-2 m-2 mb-3">
       <div className="relative">
-        <ProfilePicture user={author} />
+        <ProfilePicture
+          user={author}
+          size={enableSourceHeader ? undefined : 'xsmall'}
+          className={classNames(!enableSourceHeader && 'top-7 -right-2.5')}
```
sshanzel, line 25: Just updating to make it in line with the updates.
```suggestion
          className={!enableSourceHeader && 'top-7 -right-2.5'}
```

sshanzel (approved):
Nice work! Just had one non-blocking comment but I think this is good to go! 🚀

packages/shared/src/components/cards/SharedPostCardHeader.tsx
```diff
@@ -7,29 +7,37 @@ import PostMetadata from './PostMetadata';
...
   source,
+  enableSourceHeader = false,
 }: SharedPostCardHeaderProps): ReactElement => {
   return (
     <div className="flex relative flex-row gap-2 m-2 mb-3">
       <div className="relative">
-        <ProfilePicture user={author} />
+        <ProfilePicture
+          user={author}
+          size={enableSourceHeader ? undefined : 'xsmall'}
```
rebelchris, line 23: Not sure should we just say explicit `xlarge` here?

capJavert (commented):
Looks good to me, one thing I noticed and it is a typo in our DR (my bad)..

Currently in codebase `enableSourceHeader=true` means show the current squad card layout eg. small squad image and big profile image. And `false` means show the new layout eg. big squad image and small profile image.

It should be in reverse where current layout is default and it is shown when `enableSourceHeader=false` and new layout shows when `true`.

DR says:
> This new design only needs to show on non squad feeds.

But then in conditions it says `feedName === 'squad'`(which is typo)  and it should be `feedName !== 'squad'` since that would mark non squad feeds.

Everything works with current implementation it just seems inverted now. Tagging @rebelchris for visibility.

packages/shared/src/components/FeedItemComponent.tsx
```diff
@@ -147,6 +147,9 @@ export default function FeedItemComponent({
         : PostTypeToTag[item.post.type] ?? ArticlePostCard;
       return (
         <PostTag
+          enableSourceHeader={
+            feedName !== 'squad' && item.post.source.type === 'squad'
```
capJavert, line 151: `source?.` because it might be undefined per TS, not sure why it is not erroring.

packages/shared/src/components/cards/WelcomePostCardHeader.tsx
```diff
@@ -4,29 +4,40 @@ import { ProfilePicture } from '../ProfilePicture';
...
-          size="xsmall"
+          className={!enableSourceHeader && 'absolute -right-2 -bottom-2'}
+          size={enableSourceHeader ? 'large' : 'xsmall'}
         />
       </div>
       <div className="flex flex-col flex-1 flex-grow mr-6 ml-2 typo-footnote">
-        <span className="font-bold line-clamp-2">{author.name}</span>
+        <span className="font-bold line-clamp-2">
+          {enableSourceHeader ? source.name : author.name}
+        </span>
```
capJavert, line 36: Also `source` might be undefined, so guard with `!!source && ( ... )`

capJavert (approved):
Nice, tnx for refactoring, one comment but overall good to go 🚀 👏

packages/shared/src/components/cards/WelcomePostCardHeader.tsx
sshanzel: Is this a race-condition kind of thing? I mean, in which case where we get the `source` to be `undefined`?

capJavert: Do we not have `sourceId=unknown`? In that case `source` would be `undefined` right?

packages/shared/src/components/FeedItemComponent.tsx
capJavert: Looks like we have this disabled https://www.typescriptlang.org/tsconfig#strict not sure why because a lot of checks like `strictNullChecks` (which would show error in this case since `source` is optional) are not working. cc @rebelchris

packages/shared/src/components/cards/WelcomePostCardHeader.tsx
sshanzel: Yes, that's right. Though we never show post with `unknown` source in a modal or page, and I think for WelcomePost/FreeformPost the source is always the Squad (for the purpose of the component in the subject).

capJavert: Yeah sure, let's leave it like this here. Tnx for double checking.

sshanzel: Agreed. Also would love to touch more about the topic you mentioned relating to `strictNullChecks`. Would be amazing if we can have that enabled.

packages/shared/src/components/FeedItemComponent.tsx
TomDowling: Fixed the above and made note of this for the next web meeting

PR #1931, opened 2023-07-04
fix: hot reload

capJavert:
- looks like preact still needs some "magic" to work with next hot reload
- tried just adding https://github.com/preactjs/prefresh but it does not work with webpack integration
- reintroduced `next-plugin-preact` and patched it to not enable `preact/debug` because that was the core issue of why our webapp was slow in development
- long term we should really evaluate preact and how it integrates with Next 13 and above since it will probably block us down the road

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,5 +4,10 @@
   "dependencies": {
     "@babel/runtime": "^7.17.9",
     "postcss-rem-to-responsive-pixel": "^5.1.1"
+  },
+  "pnpm": {
+    "patchedDependencies": {
+      "next-plugin-preact@3.0.7": "patches/next-plugin-preact@3.0.7.patch"
+    }
   }
 }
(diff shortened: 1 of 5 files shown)


rebelchris (approved):
Nice patch, looks great.
Tested and still have the speed now.

Just something to maybe keep in mind moving forward.

sshanzel:
> long term we should really evaluate preact and how it integrates with Next 13 and above since it will probably block us down the road

Totally agree with this, I remember in the past we were not able to proceed with some upgrades relating to Next due to some open issues with Preact working with Next. It kinda holds us off to be the most up to date for some moments.

PR #1932, opened 2023-07-04
feat: add squad create/edit pages

capJavert:
- Add squad create and edit pages
- Remove modals that were previously used
- Also implements WT-1499 since closely related
- One related fix on API side https://github.com/dailydotdev/daily-api/pull/1338
- new `useSquadNavigation` hook to replace old specific `useCreateSquadModal` hook and provide generic functionality

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android


packages/shared/src/components/squads/SquadTypeCard.tsx
```diff
@@ -0,0 +1,63 @@
...
+      )}
+    >
+      <div>
+        <h4 className="mb-2 font-bold typo-headline">{title}</h4>
+        <p className="typo-footnote">{description}</p>
+      </div>
+      <Button
+        className="btn-secondary"
+        buttonSize={ButtonSize.XSmall}
+        // @ts-expect-error Property 'disabled' does not exist on type
```
capJavert, line 48: Something is off with our types for `Button` it was complaining that `disabled` does not exists, same for `href` in `Details` component. 🤔

packages/shared/src/components/squads/SquadsToShare.tsx
```diff
@@ -63,7 +63,6 @@ export function SquadsToShare({
       onClick={() =>
         openNewSquadModal({
           origin: Origin.Share,
-          redirectAfterCreate: false,
```
capJavert, line 66: We now have the separate page so we should always redirect, alternatively we would have to remember the post id to get user back to it for sharing. Edge case?

packages/shared/src/lib/constants.ts
```diff
@@ -32,6 +32,7 @@ export const updateFirefoxExtensionLink = 'https://r.daily.dev/firefoxupdate';
 export const downloadBrowserExtension = 'https://r.daily.dev/download';
 export const referralToC = 'https://r.daily.dev/referral-toc';
 export const twitter = 'https://r.daily.dev/twitter';
+export const publicSquadWaitlist = 'https://r.daily.dev/public-squad-waitlist';
```
capJavert, line 35: Not working now, gonna ping Nimrod to put up a placeholder.

packages/shared/src/styles/base.css
```diff
@@ -215,6 +218,9 @@ body {
   --theme-gradient-cabbage: theme('colors.cabbage.10')66;
   --theme-gradient-onion: theme('colors.onion.10')66;
 
+  --theme-overlay-cabbage-opacity: theme('colors.cabbage.90')A3;
+  --theme-overlay-onion-opacity: theme('colors.onion.40')A3;
```
sshanzel, line 222: Are these supposed to switch when in light mode? 90 to 40 then 40 to 90?

Also, I think we need a more specific name for the use case instead of `opacity`. Similar to `theme-hover`, etc.

packages/shared/src/svg/BetaBadge.tsx
```diff
@@ -5,18 +5,16 @@ export default function BetaBadge({
 }: HTMLAttributes<SVGElement>): ReactElement {
```
sshanzel, line 5: Not particular to this PR, but this looks to be something in the cloudinary.

sshanzel (changes requested):
Nice work! Thank you for also cleaning up unused files/modals 🚀 

Just had a few comments but nothing major.

packages/shared/src/svg/BetaBadge.tsx
capJavert: @sshanzel not sure what you mean here?

packages/shared/src/styles/base.css
capJavert: Yeah, talked about naming with Chris, we really need to find something better here but we did make this as a consensus for now. All these generic `opacity` and `theme` names are not really good since they do not say "for what is this style used for" but say this style represents these values".

packages/shared/src/components/squads/SquadsToShare.tsx
capJavert: No, but the hook is also really specifically called modal, I'll think of the way to update it.

packages/shared/src/styles/base.css
rebelchris: It's because mainly the "what" changes or has multi purpose, for now these are good enough as they once off and needed just for one case.
We can decide on better naming when we get a breather and maybe re-do the whole color naming?

packages/shared/src/components/modals/ShareModal.spec.tsx
```diff
@@ -100,8 +101,14 @@ describe('ShareModal Test Suite:', () => {
 
     expect(btn).toBeInTheDocument();
     btn.click();
+    const useRouterMock = useRouter as jest.Mock;
```
rebelchris, line 104: Maybe we should refactor the names of these tests as well, as they still mention "modal"
It could become confusing down the line.

packages/shared/src/components/squads/Details.tsx
```diff
@@ -155,85 +151,126 @@ export function SquadDetails({
...
+                  <div className="flex gap-2 items-center">
+                    Public squad <BetaBadge />
+                  </div>
+                }
+                description="Everyone can see the content and the posts may appear on the main feed"
+                isSelected={isPublic}
+                buttonProps={{
+                  text: 'Join waitlist',
+                  tag: 'a',
+                  // @ts-expect-error Property 'disabled' does not exist on type
```
rebelchris, line 226: We are not even referring disabled here, does it expect it?
Maybe we better off fixing the issues down the props.

packages/shared/src/hooks/useCreateSquadModal.ts
```diff
@@ -26,14 +23,14 @@ export const useCreateSquadModal = ({
   isFlagsFetched,
 }: UseCreateSquadModalProps): UseCreateSquadModal => {
   const router = useRouter();
-  const { openModal } = useLazyModal();
   const [hasTriedOnboarding, setHasTriedOnboarding, isLoaded] =
     usePersistentContext<boolean>(SQUAD_ONBOARDING, hasSquads);
 
   // @NOTE see https://dailydotdev.atlassian.net/l/cp/dK9h1zoM
```
rebelchris, line 29: Is this still needed here?

packages/shared/src/components/squads/SquadTypeCard.tsx
rebelchris: We should fix that indeed.

packages/shared/src/components/squads/Details.tsx
```diff
@@ -6,25 +6,27 @@ import { Button } from '../buttons/Button';
...
 import { cloudinary } from '../../lib/image';
 import CameraIcon from '../icons/Camera';
 import { formToJson } from '../../lib/form';
-import { Modal } from '../modals/common/Modal';
 import { blobToBase64 } from '../../lib/blob';
 import { checkExistingHandle, SquadForm } from '../../graphql/squads';
 import { capitalize } from '../../lib/strings';
 import { IconSize } from '../Icon';
-import { Justify } from '../utilities';
+import { FlexCol } from '../utilities';
```
rebelchris, line 17: I'm not sure we should rely on these for now.
It becomes dangerous when someone decides to add a class to this FlexCol for one bug and forgets to check other places.
It's more a generic comment for now until we can support a full blown component library maybe?

@sshanzel What do you think?

packages/webapp/pages/squads/[handle]/edit.tsx
```diff
@@ -0,0 +1,132 @@
...
+  ButtonSize,
+} from '@dailydotdev/shared/src/components/buttons/Button';
+import ArrowIcon from '@dailydotdev/shared/src/components/icons/Arrow';
+import { defaultOpenGraph, defaultSeo } from '../../../next-seo';
+import { getLayout as getMainLayout } from '../../../components/layouts/MainLayout';
+
+type EditSquadPageProps = { handle: string };
+
+const seo: NextSeoProps = {
+  title: 'Create your Squad',
```
Comment (rebelchris) [inline packages/webapp/pages/squads/[handle]/edit.tsx:35]: This should be different, maybe check quickly what the old modal title was.

packages/webapp/pages/squads/[handle]/edit.tsx
```diff
@@ -0,0 +1,132 @@
...
+  openGraph: { ...defaultOpenGraph },
+  ...defaultSeo,
+};
+
+const EditSquad = ({ handle }: EditSquadPageProps): ReactElement => {
```
Comment (rebelchris) [inline packages/webapp/pages/squads/[handle]/edit.tsx:40]: Also generic note, don't forget we need tracking here probably.

packages/webapp/pages/squads/new.tsx
```diff
@@ -0,0 +1,123 @@
...
+  const { displayToast } = useToastNotification();
+  const { addSquad } = useBoot();
+  const [form] = useState<Partial<SquadForm>>({
+    public: false,
+  });
+
+  useEffect(() => {
+    trackEvent({
+      event_name: AnalyticsEvent.Impression,
+      target_type: TargetType.CreateSquadPopup,
```
rebelchris, line 50: This should likely change, just flagging to block for now.

packages/webapp/pages/squads/new.tsx
```diff
@@ -0,0 +1,123 @@
...
+  const onCreate = async (e, newSquadForm: SquadForm) => {
+    e.preventDefault();
+    const data = { ...newSquadForm } as SquadForm;
+
+    try {
+      const newSquad = await createSquad({ ...data });
+
+      if (!newSquad) {
+        return;
+      }
```
rebelchris, line 66: What case is this, an error?
If so should we not show something to the user?

packages/webapp/pages/squads/new.tsx
```diff
@@ -0,0 +1,123 @@
...
+
+      await router.replace(newSquad.permalink);
+    } catch (err) {
+      displayToast(DEFAULT_ERROR);
+    }
+  };
+
+  const handleClose = async () => {
+    router.push('/squads');
+  };
```
rebelchris, line 82: Is this correct?
Not sure if we should maybe handle going back to where you came from, or simply not even showing a close button as it's a page?

packages/webapp/pages/squads/new.tsx
```diff
@@ -0,0 +1,123 @@
...
+    <ManageSquadPageContainer>
+      <NextSeo {...seo} titleTemplate="%s | daily.dev" />
+      <ManageSquadPageMain>
+        <div
+          style={{
+            backgroundImage: `url(${cloudinary.squads.createSquad})`,
+          }}
+          className="flex flex-col justify-center items-center mb-6 w-full h-48 bg-cover"
+        >
+          <SourceIcon className="!w-20 !h-20" />
```
rebelchris, line 102: Why the overwrites here?
We have default icon sizes we should use.

rebelchris (changes requested):
Looks like a great start, some questions and feedback.
Some are more discussion points

packages/webapp/pages/squads/new.tsx
capJavert: This is how it was agreed in spec and DR.

packages/webapp/pages/squads/[handle]/edit.tsx
Comment (capJavert) [reply packages/webapp/pages/squads/[handle]/edit.tsx:35]: There was no title
image

packages/shared/src/components/squads/Details.tsx
capJavert: I mean we did encourage using those in the past. If you ask me we don't even need those util components but happy with whatever you decide 👍

packages/shared/src/hooks/useCreateSquadModal.ts
capJavert: Refactored the whole hook in 91333732c13665fe09ef49056ce8d64255124d38

packages/webapp/pages/squads/[handle]/edit.tsx
Comment (capJavert) [reply packages/webapp/pages/squads/[handle]/edit.tsx:35]: Just `Create squad` should do?

packages/shared/src/components/squads/SquadsToShare.tsx
capJavert: Refactored here 91333732c13665fe09ef49056ce8d64255124d38 @sshanzel

packages/webapp/pages/squads/[handle]/edit.tsx
Comment (capJavert) [reply packages/webapp/pages/squads/[handle]/edit.tsx:40]: Discussion here https://dailydotdev.slack.com/archives/C0592NL7YCU/p1688983153989259

packages/webapp/pages/squads/new.tsx
capJavert: Discussion here https://dailydotdev.slack.com/archives/C0592NL7YCU/p1688983153989259

packages/webapp/pages/squads/new.tsx
capJavert: Just a leftover from old modal. The above `createSquad` should throw indeed. Removed it. 👌

packages/shared/src/components/squads/Details.tsx
rebelchris: To me they propose more danger then benefit anything from them 😅

packages/webapp/pages/squads/[handle]/edit.tsx
Comment (capJavert) [reply packages/webapp/pages/squads/[handle]/edit.tsx:35]: `Squads early access` seems weird since you are on squad create page.

Comment (rebelchris) [reply packages/webapp/pages/squads/[handle]/edit.tsx:35]: It's the edit page right?
So should be something in line of Edit your squad

packages/webapp/pages/squads/new.tsx
rebelchris: Alright probably to focus on the public directory forget about my comment then 🙏

packages/shared/src/components/squads/Details.tsx
capJavert: `target` would be an error here `disabled` is just a typo, `@ts-expect-error` disables the next error no matter what but requires some kind of note. 

image

packages/webapp/pages/squads/[handle]/edit.tsx
Comment (capJavert) [reply packages/webapp/pages/squads/[handle]/edit.tsx:35]: Oh sorry, I somehow seen it on create, it was `Squad settings` on modal. I'll update.

packages/webapp/pages/squads/new.tsx
capJavert: We are going to use `page view` and remove this impression.

packages/webapp/pages/squads/[handle]/edit.tsx
Comment (capJavert) [reply packages/webapp/pages/squads/[handle]/edit.tsx:42]: We are going to use `page view`, no additional events needed.

packages/shared/src/components/buttons/Button.tsx
```diff
@@ -46,6 +46,7 @@ export interface BaseButtonProps {
   displayClass?: string;
   textPosition?: string;
   position?: string;
+  disabled?: boolean;
```
capJavert, line 49: Since `HTMLAnchorElement` does not have `disabled` attribute I added `disabled` in `BaseButtonProps` to allow for any button to pass in disabled state as we do support it in `Tag`.

This does not show when calling `<Button tag="a" disabled />` for some reasons, probably due some kind of TS misconfiguration.

It did show if you did the following:
```tsx
const props = {
  tag: 'a',
  disabled: false
}

<Button {...props} />
```

All in all now it's fixed.

packages/shared/src/components/squads/SquadTypeCard.tsx
capJavert: Fixed in 49e908cf31d6706a8903c279f3079238255c6d8e

packages/shared/src/svg/BetaBadge.tsx
sshanzel: I mean this asset I think should have been in the cloudinary when it was introduced in the past. I could be wrong though.

capJavert: Since it was already present I'll leave it for now. 🙏

sshanzel: Yep. Agreed with that.

sshanzel (approved):
Massive work! Thanks for the additional cleanups! 🚀 🚀 🚀

packages/shared/src/hooks/useSquadNavigation.ts
```diff
@@ -1,39 +1,42 @@
...
-  hasAccess: boolean;
-  isFlagsFetched: boolean;
-};
-
-export const useCreateSquadModal = ({
-  hasSquads = false,
-  hasAccess = false,
-  isFlagsFetched,
-}: UseCreateSquadModalProps): UseCreateSquadModal => {
+export const useSquadNavigation = (): UseSquadNavigation => {
```
rebelchris, line 18: Is this not a bit of overkill.
Basically we just have hrefs to static pages now right, with no additional logic.

Feels like we could simply use `href={newSquad}` and editSquad on each respective place.
Do note we are removing the auto-open function here, so maybe you can incorporate that?
https://dailydotdev.atlassian.net/jira/software/projects/WT/boards/11?selectedIssue=WT-1527

rebelchris (approved):
Nice work!
Just one comment that we can probably remove the hook you created and go for simple approach of URLs.
But won't block this PR if you want to seperate it for another ticket.

PR #1934, opened 2023-07-05
fix: add promote/demote option for moderators

rebelchris:
- Added option to promote/demote on the frontend (see changes to expose flag in API [here](https://github.com/dailydotdev/daily-api/pull/1336/commits/089a6b71f6c78a4308e42db04d53bc264d10dac1))

@idoshamun Will this suffice or did you want a more explicit to only show this when it's a open squad?

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1503 #done

### DIFF
--- a/packages/shared/src/components/post/PostModalActions.tsx
+++ b/packages/shared/src/components/post/PostModalActions.tsx
@@ -11,7 +11,12 @@ import CloseIcon from '../icons/MiniClose';
 import OpenLinkIcon from '../icons/OpenLink';
 import { Roles } from '../../lib/user';
 import AuthContext from '../../contexts/AuthContext';
-import { banPost, internalReadTypes, Post } from '../../graphql/posts';
+import {
+  banPost,
+  internalReadTypes,
+  Post,
+  promotePost,
+} from '../../graphql/posts';
 import useReportPostMenu from '../../hooks/useReportPostMenu';
 import classed from '../../lib/classed';
 import { SimpleTooltip } from '../tooltips/SimpleTooltip';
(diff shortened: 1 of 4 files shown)


idoshamun:
@rebelchris fine by me :)
Because it's system mod only I wouldn't worry about edge cases and so on

packages/shared/src/graphql/posts.ts
```diff
@@ -39,6 +39,16 @@ export const supportedTypesForPrivateSources = [
...
 ];
 
+type PostFlags = {
+  sentAnalyticsReport: boolean;
+  banned: boolean;
+  deleted: boolean;
+  private: boolean;
+  visible: boolean;
+  showOnFeed: boolean;
+  promoteToPublic: number;
```
sshanzel, line 49: Was it intentional that the type for `promoteToPublic` is a number?

packages/shared/src/components/post/PostModalActions.tsx
```diff
@@ -78,6 +84,27 @@ export function PostModalActions({
...
+        title: promoteFlag ? 'Demote' : 'Promote',
+      },
+    };
+    if (await showPrompt(options)) {
+      if (promoteFlag) {
+        await demotePost(post.id);
+      } else {
+        await promotePost(post.id);
+      }
+    }
```
sshanzel, line 105: Should we not always try to utilize calling requests through `useMutation`?

sshanzel (approved):
Looks amazing to me. Just had a few thoughts around requests and the type for promoting to public.

packages/shared/src/graphql/posts.ts
rebelchris: Yeah its a unix timestamp

packages/shared/src/components/post/PostModalActions.tsx
rebelchris: I think for these ones its fine, only ido uses them, and no problem  if they have simple requests

PR #1936, opened 2023-07-07
feat: onboarding experiment v2

sshanzel:
## Changes
- Refactored some of the existing Onboarding flow to accommodate usage without the context of a Modal.
- Introduced a new property called `simplified` on `AuthOptions` component to know whether to display the header or not - and to make the footer have the same width as the content by utilizing the `AuthContainer`.

Currently deployed at: [preview](https://preview.app.daily.dev/) 

Just run the command below to enable it.
```
window.setFeature('onboarding_v2', 'v1');

// to set the title execute as below:
window.setFeature('onboarding_filtering_title', 'v1');
```

Previews:

image
Screenshot 2023-07-07 at 4 20 05 PM
Screenshot 2023-07-07 at 4 19 57 PM
Screenshot 2023-07-07 at 4 19 36 PM

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1506 #done


capJavert (commented):
Overall looks good, just two general comments, I would like to take a look when first visit part is connected 👍

packages/shared/src/components/MainLayout.tsx
```diff
@@ -137,6 +143,9 @@ export default function MainLayout({
       {customBanner || (
         <PromotionalBanner bannerData={bannerData} setLastSeen={setLastSeen} />
       )}
+      {!user && isOnboardingOpen && onboardingV2 !== OnboardingV2.Control && (
```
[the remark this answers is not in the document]
sshanzel: Makes sense. Let's do that.

capJavert:
@sshanzel did a little testing, few cases if you could check on your side:
1. Logging out freezes the browser
- Start logged in into preview
- Go to your feed
- Logout, at this point you will not be able to refresh page or close the tab because it feezes
2. Login CTA not working
image

Nothing hapens when I click on "Log in"

3. `AnonymousFeed` query still fires when user not logged in
I think we can probably just disable the query until user is logged in, as we don't need or show that data. Also probably blocking it on the API when experiment is done and accepted.

packages/shared/src/components/containers/FeedTopicCard.tsx
```diff
@@ -31,8 +32,8 @@ const BackgroundLayer = classed(
 const background: Record<OnboardingFiltersLayout, ReactNode> = {
   grid: (
     <>
-      <BackgroundLayer className="-z-1 opacity-64 group-hover:-rotate-12" />
-      <BackgroundLayer className="opacity-24 -z-2 group-hover:-rotate-[24deg]" />
+      <BackgroundLayer className="invisible group-hover:visible -z-1 opacity-64 group-hover:-rotate-12" />
+      <BackgroundLayer className="invisible group-hover:visible opacity-24 -z-2 group-hover:-rotate-[24deg]" />
```
sshanzel, line 36: Had to add this to make the resting position look normal. Not sure why it works as intended on previous implementations (AuthModal).
image

sshanzel:
@capJavert thank you very much for testing it thoroughly. We have changed a bit the mechanism on how we render it, some of the concerns you have listed should have also been fixed. Including the CTAs that are not working.

packages/webapp/pages/onboard.tsx
```diff
@@ -0,0 +1,176 @@
...
+      }),
+    });
+  }, [
+    trackEvent,
+    isPageReady,
+    onboardingV2,
+    onboardingMinimumTopics,
+    router,
+    user,
+  ]);
```
sshanzel, line 106: @idoshamun tracking for impression. This also makes sure it doesn't send an `impression` when the page loads and the parameters are not yet considered.

packages/webapp/pages/onboard.tsx
```diff
@@ -0,0 +1,176 @@
...
+    const screen = isFiltering ? OnboardingStep.Topics : OnboardingStep.Intro;
+
+    trackEvent({
+      event_name: AnalyticsEvent.ClickOnboardingNext,
+      extra: JSON.stringify({ screen_value: screen }),
+    });
+
+    if (!isFiltering) return setIsFiltering(true);
+
+    return setIsAuthenticating(true);
```
sshanzel, line 64: @idoshamun tracking of clicking the next buttons. There are only two "Next" buttons, in the introduction, and on the topics filtering state.

The tracking of clicking topics is still intact on how it used to work on the previous onboardings.

packages/webapp/pages/onboard.tsx
```diff
@@ -0,0 +1,176 @@
...
+
+    if (user || onboardingV2 === OnboardingV2.Control) {
+      router.push('/');
+      return;
+    }
+
+    trackEvent({
+      event_name: AnalyticsEvent.Impression,
+      target_type: TargetType.MyFeedModal,
+      target_id: onboardingV2,
```
idoshamun, line 92: What will be the value here for control and this variation?

packages/webapp/pages/onboard.tsx
idoshamun: Can you please make sure it doesn't get triggered twice? The rest looks great!

packages/webapp/pages/onboard.tsx
idoshamun: Looks great!

rebelchris (commented):
Logged some things on Slack.
Further nothing major I can see on the codebase, only thing I'd like to ask as this impact authentication as a generic.

Please do test the critical flows once you decide to deploy.
Feel free to continue without my approval as I'm off tomorrow.

packages/shared/src/components/containers/FeedTopicCard.tsx
rebelchris: Works now :)

packages/shared/src/components/MainLayout.tsx
```diff
@@ -138,15 +139,30 @@ export default function MainLayout({
...
+    onboardingV2 !== OnboardingV2.Control &&
+    (!page || feeds.includes(page)) &&
+    isPageReady &&
+    !existingAnonymousUser;
+
+  if (shouldShowOverlay) {
+    router.push('/onboard');
+  }
+
+  if (!isPageReady || shouldShowOverlay) return null;
```
sshanzel, line 158: Just wanted to note that on any page, we are not showing anything, anymore until the page is ready. cc @rebelchris @idoshamun @capJavert @TomDowling

capJavert: Has some implications of how our first paint looks currently..
Before - some top skeleton:
image


After - just a blank screen:
image

packages/webapp/pages/onboarding.tsx
```diff
@@ -0,0 +1,210 @@
...
+      }),
+    });
+  }, [
+    trackEvent,
+    isPageReady,
+    onboardingV2,
+    onboardingMinimumTopics,
+    router,
+    user,
+  ]);
```
capJavert, line 125: This should be split in two `useEffect`. One will call `trackEvent` and the second one will be in charge of redirect. That way you will have less dependencies to work with in both effects and they will also be easier to understand.

capJavert (changes requested):
Nice, I like the dedicated page much better 👌 

Some concerns, It is pretty big PR so just want to understand. From functionality perspective it works good.

packages/webapp/pages/onboarding.tsx
rebelchris: Added this idea to web team meeting, think we would benefit from a useEffect hook that runs once for tracking purposes.

packages/shared/src/components/MainLayout.tsx
```diff
@@ -132,6 +140,24 @@ export default function MainLayout({
     );
   };
 
+  const router = useRouter();
+  const feeds = Object.values(MainFeedPage);
+  const page = router?.route?.substring(1).trim() as MainFeedPage;
+  const isPageReady = (isFeaturesLoaded && router.isReady) || isTesting;
+  const shouldShowOverlay =
```
rebelchris, line 147: Naming is no longer correct here 🙏

packages/shared/src/components/MainLayout.tsx
```diff
@@ -132,6 +140,24 @@ export default function MainLayout({
...
+    onboardingV2 !== OnboardingV2.Control &&
+    (!page || feeds.includes(page));
+
+  useEffect(() => {
+    if (!shouldShowOverlay) return;
+
+    router.push(`${webappUrl}/onboarding`);
+  }, [shouldShowOverlay, router]);
+
+  if (!isFeaturesLoaded || shouldShowOverlay) return null;
```
rebelchris, line 159: This should never come to shouldShowOverlay right?
Because the useEffect above would already redirect.

packages/shared/src/components/auth/AuthModalFooter.tsx
```diff
@@ -15,7 +16,10 @@ function AuthModalFooter({
   onIsLogin,
 }: AuthModalFooterProps): ReactElement {
   return (
-    <Modal.Footer className={className} justify={Justify.Center}>
+    <Modal.Footer
+      className={classNames(className, 'gap-[unset]')}
```
rebelchris, line 20: I think it might be more efficient to add it to the tailwind config, these bracket once as I remember add more bloat stuff.

packages/shared/src/lib/featureManagement.ts
```diff
@@ -137,6 +139,24 @@ export class Features<T extends FeatureValue = string> {
     [InAppNotificationPositionEnum.Bottom, InAppNotificationPositionEnum.Top],
   );
 
+  static readonly OnboardingV2 = new Features(
+    'onboarding_v2',
+    OnboardingV2.Control,
+    [OnboardingV2.Control, OnboardingV2.V1],
+  );
```
rebelchris, line 146: Non breaking, but reading these broke my head a bit 😂
OnboardingV2.V1 seems very counter intuative, but for now let's keep it.
Just a headsup for future we might need to think a bit better on naming features.

rebelchris (changes requested):
Also small note did you test it on mobile?
It seems the onboard page is not fully responsive.

packages/shared/src/components/MainLayout.tsx
capJavert: First render would still happen before `useEffect` is run for the first time.

capJavert: Same for updates, update to `shouldShowOverlay` would render and then `useEffect` would run.

packages/shared/src/contexts/AuthContext.tsx
```diff
@@ -142,7 +142,7 @@ export const AuthContextProvider = ({
       user: endUser,
       referral: loginState?.referral ?? referral,
       referralOrigin: loginState?.referralOrigin ?? referralOrigin,
-      isFirstVisit: user?.isFirstVisit ?? false,
+      firstVisit: user?.firstVisit,
```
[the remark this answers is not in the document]
sshanzel: `isFirstVisit, is not used anymore so we turned this to the `firstVisit` string instead. It is also optional as anonymous users can only get it.

packages/shared/src/components/MainLayout.tsx
sshanzel: @capJavert had to. The placement of the logo from other pages can be not consistent, like in this onboarding page.

packages/webapp/pages/onboarding.tsx
sshanzel: We wouldn't want to have users with the same condition of redirecting to get the impression event. They are still going to have the same dependencies. I used to separate `useEffects` in the past too, but lesser `useEffects` are always better.

packages/shared/src/lib/featureManagement.ts
sshanzel: Yeah, we will definitely need to come up with something more understandable 😂

sshanzel:
@rebelchris it should be responsive. The image attachment seems to be broken. Can you specify which screen size it was?

packages/shared/src/components/auth/AuthModalFooter.tsx
sshanzel: Good point. At a later stage, we might want to take a look at the ModalFooter on why it has a `gap-3` built-in.

sshanzel:
Looks okay to me on XR. Can you specify which step it was getting messed up?

Also I noticed an issue with the tooltip on mobile, and this is the onboarding, we don't want to give first time users to have bad UX, so I fixed it right away. 

https://github.com/dailydotdev/apps/assets/13744167/139506f3-aba1-4092-bb0e-aea0586d2cb4

rebelchris:
@sshanzel Seems fixed now on mobile 😅
I do see a 1px scroll both ways, not sure how well you'll notice it on a actual device.

sshanzel:
> I do see a 1px scroll both ways, not sure how well you'll notice it on a actual device.

Oh, right. Let me check any overflow. Definitely better to have it removed.

packages/webapp/pages/onboarding.tsx
rebelchris: But this will fire multiple times, then it's better to have a empty dependency array for now and exclude it.
Or add a state to ensure it only tracks once.

rebelchris (changes requested):
Just one constraint otherwise good to go

packages/webapp/pages/onboarding.tsx
sshanzel: Right. Yeah, let's add a state to ensure sending a single event. And create a hook in the future for these cases.

packages/shared/src/components/onboarding/MemberAlready.tsx
```diff
@@ -0,0 +1,41 @@
...
+  className = {},
+  onLogin,
+}: MemberAlreadyProps): ReactElement {
+  const { showLogin } = useAuthContext();
+
+  const onClick = () => {
+    if (onLogin) return onLogin();
+
+    return showLogin(AuthTriggers.Onboarding, { isLogin: true });
+  };
```
capJavert, line 27: Is this supposed to return when `onLogin` is provided as prop?

If yes I would even go as far to make `onLogin` required and explicitly pass `onLogin={() => showLogin(AuthTriggers.Onboarding, { isLogin: true }))` on places where it is needed. 

Think it will be better then component having hidden logic that will get overriden if you provide a prop.

packages/webapp/pages/onboarding.tsx
```diff
@@ -0,0 +1,223 @@
...
+
+const seo: NextSeoProps = {
+  title: 'Get started',
+  openGraph: { ...defaultOpenGraph },
+  ...defaultSeo,
+};
+
+export function OnboardPage(): ReactElement {
+  const router = useRouter();
+  const [isTracked, setIsTracked] = useState(false);
```
capJavert, line 63: `useRef` should be better to avoid causing re-renders.

capJavert (approved):
Good to go I think 👏 Two more non blocking comments.

packages/webapp/pages/onboarding.tsx
sshanzel: Agreed.

packages/shared/src/components/onboarding/MemberAlready.tsx
sshanzel: Makes sense. Let's do that.

sshanzel:
I have also introduced a change in the test setup to make it easier to make future changes that affect all pages by creating a TestBootProvider.

sshanzel:
@rebelchris fixed the assets that can vary by introducing a hook. Can be found in these two commits:

https://github.com/dailydotdev/apps/pull/1936/commits/2b0be3f021bc13ca5217b23fdb394f9945b535fa
https://github.com/dailydotdev/apps/pull/1936/commits/6d0ec22395998066093cdb078b93b1424bfc633d

PR #1941, opened 2023-07-10
feat: adding the ability to report comments

TomDowling:
## Changes
Added the ability to report any comment from the context menu. As well as adding tests and tracking for these

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1492 #done


packages/shared/src/components/modals/report/ReportModal.tsx
```diff
@@ -0,0 +1,74 @@
...
+        {title && (
+          <p className="mb-6 text-theme-label-tertiary typo-callout">{title}</p>
+        )}
+        <Radio
+          name="report_reason"
+          options={typeof reasons === 'function' ? reasons(reason) : reasons}
+          value={reason}
+          onChange={setReason}
+        />
+        {reason === OTHER_KEY && (
```
capJavert, line 49: For post modal the textarea is always shown but optional to fill in. It is required when `reason === OTHER_KEY`. Not sure if the spec changed but this should remain as it is now.

capJavert: It should be same logic for comment report modal.

capJavert (changes requested):
Good work so far, some notes below.

packages/shared/src/graphql/comments.ts
```diff
@@ -1,10 +1,24 @@
 import request, { gql } from 'graphql-request';
+import { ReactElement } from 'react';
 import { Connection, Upvote } from './common';
 import { COMMENT_FRAGMENT, USER_SHORT_INFO_FRAGMENT } from './fragments';
 import { EmptyResponse } from './emptyResponse';
 import { UserShortProfile } from '../lib/user';
 import { graphqlUrl } from '../lib/config';
 
+export interface OptionMenuItem {
```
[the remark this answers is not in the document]
TomDowling: @capJavert, So i am using this in two places already PostOptionsMenu and CommentActionButtons, and looking at the codebase, there is only maybe one more place with this same interface, so would you suggest I just move this to somewhere generic (unsure the correct place) or change this one for Comment and add another for the post?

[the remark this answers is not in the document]
capJavert: I think `/shared/src/graphql/common` should be ok place to put it and we can leave it at that 👍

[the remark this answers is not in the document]
capJavert: @TomDowling I think maybe move it to `/shared/src/lib/element`, cc @rebelchris ?

packages/shared/src/components/modals/report/ReportModal.tsx
sshanzel: Oh, wow. Thank you for the link to the discussion. We should probably ask the design team to update the Sketch too in order to avoid confusion.

packages/shared/src/components/modals/report/ReportCommentModal.tsx
```diff
@@ -0,0 +1,67 @@
...
+    note: string,
+  ): Promise<void> => {
+    const { successful } = await reportComment({
+      commentId: comment.id,
+      reason,
+      note,
+    });
+    if (!successful) return;
+
+    trackEvent({
```
rebelchris, line 42: This should be a postAnalyticsEvent as we should have reference to the post as well actually.

packages/shared/src/components/modals/report/ReportModal.tsx
rebelchris: So this should always render correct?
Not sure why it's marked as resolved.

rebelchris (changes requested):
Two minor comments, otherwise looks great 🙏

packages/shared/src/graphql/comments.ts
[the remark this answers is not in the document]
rebelchris: Yeah agreed indeed on that :)

packages/shared/src/components/modals/report/ReportModal.tsx
TomDowling: @sshanzel @capJavert, Could you please confirm, the way i read the above discussion was the textarea only showed when the Other was selected. But could you please confirm

capJavert: It should always show.

TomDowling: Apologies i will then fix this then

packages/shared/src/components/comments/CommentActionButtons.spec.tsx
```diff
@@ -68,11 +68,6 @@ const renderComponent = (
   );
 };
 
-it('should not show options button when user is not the author', async () => {
-  renderComponent();
-  expect(screen.queryByLabelText('Options')).not.toBeInTheDocument();
-});
```
capJavert, line 74: Do we need a test to make sure options with "Report comment" is shown instead?

packages/shared/src/components/Feed.spec.tsx
```diff
@@ -675,9 +674,6 @@ it('should report broken link with comment', async () => {
   contextBtn.click();
   const brokenLinkBtn = await screen.findByText('Broken link');
   brokenLinkBtn.click();
-  const input = (await screen.findByRole('textbox')) as HTMLTextAreaElement;
-  fireEvent.change(input, { target: { value: 'comment' } });
-  input.dispatchEvent(new Event('input', { bubbles: true }));
```
capJavert, line 680: This needs to be reverted because of https://github.com/dailydotdev/apps/pull/1941#discussion_r1259787245 right?

packages/shared/src/hooks/useReportComment.ts
```diff
@@ -0,0 +1,57 @@
...
+    void,
+    unknown,
+    ReportCommentProps
+  >((variables) =>
+    requestMethod(graphqlUrl, REPORT_COMMENT_MUTATION, variables),
+  );
+
+  const onReportCallback = async (): Promise<void> => {
+    displayToast('🚨 Thanks for reporting!');
+  };
```
capJavert, line 43: Just wondering if we need this extracted here? I think it would be better to just call it where needed? Would love to hear what @rebelchris and @sshanzel think as well.

capJavert (approved):
Works good ✅ Few minor comments but feel free to merge after resolving.

packages/shared/src/hooks/useReportComment.ts
rebelchris: Ideally actually we abstract the actual text in a common lib file so we only need to change it in one place.
Once that's done we can simply inline the toast yes.

packages/shared/src/lib/element.ts
```diff
@@ -62,3 +63,9 @@ export const isCompanionActivated = (): boolean =>
   !!document
     .querySelector('daily-companion-app')
     ?.shadowRoot?.querySelector?.('#daily-companion-wrapper')?.firstChild;
+
+export interface OptionMenuItem {
+  icon: ReactElement;
+  text: string;
+  action: () => unknown;
+}
```
sshanzel, line 71: I think we have an existing OptionItem menu interface in the `PortalMenu.tsx`.

sshanzel (changes requested):
Nice work! Looks great overall! Also, thank you for adding a test suite for the reporting of comment modal. Really goes a long way.

Just had a few minor comments, but we're almost there! 🚀

packages/shared/src/components/modals/report/ReportCommentModal.tsx
```diff
@@ -0,0 +1,78 @@
...
+      extra: JSON.stringify({
+        commentId: comment.id,
+      }),
+    });
+
+    trackEvent(
+      postAnalyticsEvent('report comment', post, {
+        extra: { commentId: comment.id },
+      }),
+    );
```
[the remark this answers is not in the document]
TomDowling: I was following the examples I found from the app? Unless this is wrong this seems to be the way its always done?
Screenshot 2023-07-14 at 5 06 21 pm

packages/shared/src/lib/element.ts
TomDowling: The interface in the `PortalMenu.tsx` returns `Promise<void>` which is different to the above. Possibly the action should return void instead of unknown. I can make the change however, will need to update 5 or 6 functions to match that return type so just want to make sure thats the right thing to do?

packages/shared/src/components/modals/report/ReportCommentModal.tsx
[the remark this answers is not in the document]
sshanzel: I mean there is nothing wrong with the event itself but the line exactly above this is also sending an event.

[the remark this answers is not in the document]
TomDowling: Ah good spot, yes you are correct, I'll remove the top one

packages/shared/src/lib/element.ts
sshanzel: Yes, I agree. I think it would be better to update the type to `() => unknown  | Promise<unknown>`.

Or even better, make it generic to accept what type it would return with a default of unknown. WDYT?

packages/shared/src/components/comments/CommentActionButtons.tsx
```diff
@@ -144,6 +151,43 @@ export default function CommentActionButtons({
...
+      label: 'Delete comment',
+      action: () => onDelete(comment, parentId),
+      Icon: <TrashIcon />,
+    });
+  }
+
+  if (!isAuthor) {
+    commentOptions.push({
+      label: 'Report comment',
+      action: () => openReportCommentModal(),
```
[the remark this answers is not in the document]
rebelchris: The above should do

packages/shared/src/components/modals/report/ReportCommentModal.tsx
```diff
@@ -0,0 +1,71 @@
...
+  ): Promise<void> => {
+    const { successful } = await reportComment({
+      commentId: comment.id,
+      reason,
+      note,
+    });
+    if (!successful) return;
+
+    trackEvent(
+      postAnalyticsEvent('report comment', post, {
```
rebelchris, line 48: I'd suggest using the `AnalyticsEvent` enum for this `report comment` instead

rebelchris (changes requested):
Almost there 🥳

packages/shared/src/components/squads/SquadHeaderMenu.tsx
```diff
@@ -60,29 +61,37 @@ export default function SquadHeaderMenu({
   const items = useMemo(() => {
     const list: MenuItemProps[] = [
       {
-        Icon: TourIcon,
+        icon: <TourIcon size={IconSize.Small} className="mr-2" />,
```
rebelchris, line 64: Just to check did we need the size and classes here?

packages/shared/src/components/squads/SquadHeaderMenu.tsx
```diff
@@ -60,29 +61,37 @@ export default function SquadHeaderMenu({
...
+            icon: <ExitIcon size={IconSize.Small} className="mr-2" />,
+            action: onLeaveSquad,
+            label: 'Leave Squad',
+          },
     ];
 
     if (canEditSquad) {
       list.unshift({
-        Icon: SettingsIcon,
+        icon: <SettingsIcon size={IconSize.Small} className="mr-2" />,
```
rebelchris, line 94: Since this is quite repetitive, can we maybe abstract to inline icon or see if it's needed all togehter?
Think sidebar does something like that.

rebelchris (approved):
Non blocking comments only

sshanzel (approved):
Thank you for the extra effort of unifying the types 🙏

packages/shared/src/components/squads/SquadHeaderMenu.tsx
TomDowling: Abstracted it to a component called `ContextMenuIcon` to work like the sidebar sharing the classes and size

rebelchris (approved):
Looks good, think we are ready.
Just be sure to test the report comment as well as post report and generic context menus when deploying.

PR #1943, opened 2023-07-12
feat: adding a user check for postmenuaction downvote

TomDowling:
## Changes
Currently we are not doing a user check for the downvote button in the postmenuactions, this gives us an error as you are trying to downvote when not logged in. Iv'e added a user check for this function that shows the login popup if you are not logged in

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1510 #done


packages/shared/src/hooks/usePostMenuActions.ts
```diff
@@ -113,23 +114,28 @@ export const usePostMenuActions = ({
...
 
-        return cancelPostDownvote({ id: post.id });
-      }
-
-      trackEvent(
-        postAnalyticsEvent(AnalyticsEvent.DownvotePost, post, {
-          extra: { origin },
-        }),
-      );
+        cancelPostDownvote({ id: post.id });
```
rebelchris, line 129: Why did we change this part?

rebelchris (approved):
Nothing urgent, more of a generic question.
CC: @capJavert

packages/shared/src/hooks/usePostMenuActions.ts
capJavert: Looks like a refactor to use if else instead of early return. It should be fine either way but additional changes in PR often make reviews harder because you gotta wrap your hand around what is going on instead of just looking for additions and actual fix.

capJavert (approved):
🚀 Nice catch completely missed it in initial implementation.

packages/shared/src/hooks/usePostMenuActions.ts
TomDowling: I preferred to not change so much as i agree makes it look like more has changed. But, I had to change the markup as i was getting tsc compile errors with different types of returns basically

capJavert: Oh got it, I am fine with it 👍 

Only issue is that per `UsePostMenuActions` this function should be `async` and return promise so to keep it in line we should await mutations inside if else.

capJavert: Another example where our tsc does not catch that type inconsistency. 😢

PR #1945, opened 2023-07-13
revert: remove preact plugin patch

capJavert:
- Unfortunately while https://github.com/dailydotdev/apps/pull/1931 fixed hot reload and lag during local development it introduced the issues with error handling and debugging
- This will introduce the lag during local development when working on a main feed

As said in first PR:
> long term we should really evaluate preact and how it integrates with Next 13 and above since it will probably block us down the road

Looking at it briefly I don't think we actually benefit much from bundle size differences by using preact. We could always use it only for extension where it does make a difference. Maybe also worth evaluating as a long term solution. 🤞 

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done


rebelchris (approved):
Looks good, yes we can evaluate

sshanzel (approved):
Looks good to me! Would love to see how we will compare the pros and cons of moving away from preact, on webapp, at least.

TomDowling (approved):
Looks good

PR #1948, opened 2023-07-14
feat: open squads page adjustment

capJavert:
- Adjustments for open squad page 
- This also makes open squad page visible to anonymous users
- added `useJoinReferral` hook to be able to handle setting referral cookie in different places on the client
- added `useJoinSquad` hook and refactored `useLeaveSquad` to only contain logic without side effects
  - with that both hook functions can be used with `useMutation` to support loading and all other cases where needed
- needs https://github.com/dailydotdev/daily-api/pull/1341 for API referralUrl to work

## Events

| Type | event_name | target_type | value |
|--------|-------|------|-------|
|New | impression | squad join button | extra: { squad, origin, squad_type } |

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done


packages/shared/src/components/squads/SharePostBar.tsx
```diff
@@ -73,7 +73,11 @@ function SharePostBar({
         )}
       >
         <LockIcon size={IconSize.Small} />
-        <p className="typo-callout">Only admins and moderators can post</p>
+        <p className="typo-callout">
```
rebelchris, line 76: This is the disabled render only right, not sure why this change was needed then?

packages/shared/src/components/squads/SquadHeaderMenu.tsx
```diff
@@ -76,11 +82,29 @@ export default function SquadHeaderMenu({
...
+
+    if (
+      squad.currentMember &&
+      squad.currentMember.role !== SourceMemberRole.Admin
+    ) {
+      list.push({
+        Icon: ExitIcon,
+        onClick: () => {
+          onLeaveSquad();
+        },
```
rebelchris, line 103:
```suggestion
        onClick: onLeaveSquad,
```

This should do right

packages/shared/src/components/squads/SquadJoinButton.tsx
```diff
@@ -0,0 +1,70 @@
...
+  const isCurrentMember = useMemo(
+    () => squads?.some(({ id }) => id === squad.id) || false,
+    [squad, squads],
+  );
+
+  const { mutateAsync: joinSquad, isLoading: isJoiningSquad } = useMutation(
+    useJoinSquad({ squad }),
+    {
+      onError: () => {
+        displayToast('🚫 Something went wrong, please try again.');
```
rebelchris, line 31: Please check Tom's latest PR it introduces labels for re-usable texts I see some duplicates here as well, so loved to start using it.

packages/shared/src/hooks/useJoinReferral.ts
```diff
@@ -0,0 +1,60 @@
...
+
+    if (campaign && referringUserId) {
+      setCookie('join_referral', `${referringUserId}:${campaign}`, {
+        maxAge: ONE_YEAR,
+        secure: !isDevelopment,
+        domain: process.env.NEXT_PUBLIC_DOMAIN,
+        sameSite: 'lax',
+      });
+
+      const removeCookie = () => {
```
rebelchris, line 33: Should we not consider adding a removeCookie function in the cookie lib?
Could take a name for instance and do everything else inside?

packages/webapp/pages/squads/[handle]/[token].tsx
```diff
@@ -128,21 +126,14 @@ const SquadReferral = ({
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [member, trackedImpression]);
 
-  const sourceId = member?.source?.id;
   const { mutateAsync: onJoinSquad } = useMutation(
-    () => joinSquadInvitation({ sourceId, token }),
+    useJoinSquad({
+      squad: { handle, id: member?.source?.id },
+      referralToken: token,
```
Comment (rebelchris) [inline packages/webapp/pages/squads/[handle]/[token].tsx:132]: The token here is either a user token or the squad token?
Just for my understanding.

rebelchris (changes requested):
Nice work, some minor suggestions and a question.

packages/webapp/pages/squads/[handle]/[token].tsx
Comment (capJavert) [reply packages/webapp/pages/squads/[handle]/[token].tsx:132]: Here it is always user token, with open squad we don't send token at all.

Referral is attributed through new referral cookie.

packages/shared/src/components/squads/SquadHeaderMenu.tsx
capJavert: Typescript complaining.

packages/shared/src/components/squads/SharePostBar.tsx
capJavert: This copy is from open squad page design.

rebelchris: But do they not intertwine?
Just making sure the evaluation is ok because the setting to even render this might be different.

packages/shared/src/components/squads/SquadHeaderBar.tsx
```diff
@@ -58,8 +58,7 @@ export function SquadHeaderBar({
       )}
     >
       <div className="relative">
-        {(squad.public ||
-          verifyPermission(squad, SourcePermissions.Invite)) && (
+        {verifyPermission(squad, SourcePermissions.Invite) && (
```
capJavert, line 61: https://dailydotdev.slack.com/archives/C0592NL7YCU/p1689595125220649

packages/shared/src/hooks/useJoinReferral.ts
capJavert: Added as `expireCookie` bb675b5c824b20bae4f2a26fd63c1ef9884b8d07

packages/shared/src/components/squads/SquadJoinButton.tsx
capJavert: Waiting for https://github.com/dailydotdev/apps/pull/1941 to merge and I'll add

packages/shared/src/components/squads/SquadJoinButton.tsx
```diff
@@ -0,0 +1,86 @@
...
+  joinText = 'Join squad',
+  leaveText = 'Leave squad',
+}: SquadJoinProps): ReactElement => {
+  const queryClient = useQueryClient();
+  const { displayToast } = useToastNotification();
+  const { squads, user, showLogin } = useAuthContext();
+  const isCurrentMember = useMemo(
+    () => squads?.some(({ id }) => id === squad.id) || false,
+    [squad, squads],
+  );
```
rebelchris, line 30: Does squad not have a `currentMember` property already?

packages/shared/src/components/squads/SquadJoinButton.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+  const { mutateAsync: joinSquad, isLoading: isJoiningSquad } = useMutation(
```
rebelchris, line 32: Is the useJoinSquad not already a mutation?
I think ideally that logic all sits in the hook normally

packages/shared/src/components/squads/SquadJoinButton.tsx
```diff
@@ -0,0 +1,86 @@
...
+    useJoinSquad({ squad }),
+    {
+      onError: () => {
+        displayToast(labels.error.generic);
+      },
+    },
+  );
+
+  const { mutateAsync: leaveSquad, isLoading: isLeavingSquad } = useMutation(
```
rebelchris, line 41: Basically same comment for all the hooks

packages/shared/src/components/squads/SquadPageHeader.tsx
```diff
@@ -122,6 +123,11 @@ export function SquadPageHeader({
               allowedToPost && 'max-w-[30.25rem]',
             )}
             disabled={!allowedToPost}
+            disabledText={
+              isSquadMember
+                ? 'Only admins and moderators can post'
+                : 'Join the squad to create new posts'
```
rebelchris, line 129: Would still not be valid if the public squad would be set to only admins right?
Just checking for future reference, I know it's an edge-case for now

packages/shared/src/hooks/useJoinReferral.ts
```diff
@@ -0,0 +1,60 @@
...
+export const useJoinReferral = (): void => {
+  const { user, refetchBoot, isAuthReady } = useAuthContext();
+  const router = useRouter();
+  const { cid, userid } = router.query;
+  const shouldSetReferralCookie = isAuthReady && !user;
+
+  useQuery(
+    ['join_referral', { cid, userid }],
+    async () => {
+      const ONE_YEAR = 1 * 365 * 24 * 60 * 60;
```
rebelchris, line 21: This can be top level outside component

packages/shared/src/hooks/useJoinReferral.ts
```diff
@@ -0,0 +1,60 @@
...
+import { useAuthContext } from '../contexts/AuthContext';
+import { disabledRefetch } from '../lib/func';
+
+export const useJoinReferral = (): void => {
+  const { user, refetchBoot, isAuthReady } = useAuthContext();
+  const router = useRouter();
+  const { cid, userid } = router.query;
+  const shouldSetReferralCookie = isAuthReady && !user;
+
+  useQuery(
```
rebelchris, line 18: Optional, but maybe open for discussion...

I personally find it more readable to have a function defined that handles the logic and the useQuery to simply invoke that function.
Would love to hear if we can generalise one way moving forward.

packages/shared/src/hooks/useSquadOnboarding.ts
```diff
@@ -23,9 +24,10 @@ export const useSquadOnboarding = (
   const { openModal } = useLazyModal<LazyModal.SquadTour>();
   const [hasTriedOnboarding, setHasTriedOnboarding, isCacheFetched] =
     usePersistentContext(SQUAD_ONBOARDING_KEY, false);
+  const { user } = useAuthContext();
 
   useEffect(() => {
-    const conditions = [isPageReady, isFetched, isCacheFetched];
+    const conditions = [isPageReady, isFetched, isCacheFetched, user?.id];
```
rebelchris, line 30: Just checking do we not have something like isLoggedUser?
Maybe more explicit moving forward.

packages/webapp/pages/join.tsx
```diff
@@ -68,6 +68,7 @@ const Page = ({ referringUser, campaign }: PageProps): ReactElement => {
     const ONE_YEAR = 1 * 365 * 24 * 60 * 60;
```
rebelchris, line 68: I see multiple introductions of this, maybe we can util it :)

packages/webapp/pages/squads/[handle]/[token].tsx
```diff
@@ -128,39 +127,30 @@ const SquadReferral = ({
...
+    }),
     {
       onSuccess: (data) => {
-        trackEvent({
-          event_name: AnalyticsEvent.CompleteJoiningSquad,
-          extra: joinSquadAnalyticsExtra(),
-        });
```
Comment (rebelchris) [inline packages/webapp/pages/squads/[handle]/[token].tsx:139]: Did we intentionally remove the tracking here?

packages/webapp/pages/squads/[handle]/index.tsx
```diff
@@ -127,7 +149,12 @@ const SquadPage = ({ handle }: SourcePageProps): ReactElement => {
   );
 
   return (
-    <ProtectedPage seo={seo} fallback={<></>} shouldFallback={!user}>
+    <PageComponent
```
Comment (rebelchris) [inline packages/webapp/pages/squads/[handle]/index.tsx:152]: We also have `ConditionalWrapper` that can handle these kind of things, leaving it up to you what to use though.

rebelchris (changes requested):
Some small comments, nothing major looks pretty good!

packages/webapp/pages/squads/[handle]/[token].tsx
Comment (capJavert) [reply packages/webapp/pages/squads/[handle]/[token].tsx:139]: No it is moved to `useJoinSquad`

packages/shared/src/hooks/useSquadOnboarding.ts
capJavert: No, I think actually it might be good to add it to auth context.

packages/shared/src/components/squads/SquadJoinButton.tsx
capJavert: Not really it is just a function that is exported that does "bunch of things that are related". Kinda like your comment https://github.com/dailydotdev/apps/pull/1948#discussion_r1266810470 with which I agree.

If we start just re-exporting useMutation wrapped in custom hook we then often return from it something like:
`{ onJoinSquad, isLoading }`

Then we need for example another property `mutation.isDone`, we add that.

Then we need `onSuccess`, we add that to hook props.. and so on..

Pretty fast hook has X props and Y return values.

I think it is much cleaner and we should think of `useMutation` as a wrapper for loading state, success/error callbacks in specific implementation vs hardcoding them inside custom hooks which then become really specific and hard to expand.

packages/shared/src/components/squads/SquadJoinButton.tsx
capJavert: It was because squad query would get stale due to only depending on handle and `currentMember` would be left false. Fixed here 06aa0932e015cb18fbc12d3ed37ce8f7625fe01a

packages/shared/src/hooks/squads/useSquad.ts
```diff
@@ -19,8 +19,8 @@ interface UseSquad {
 }
 
 export const useSquad = ({ handle }: UseSquadProps): UseSquad => {
-  const { isFetched: isBootFetched } = useContext(AuthContext);
-  const queryKey = ['squad', handle];
+  const { isFetched: isBootFetched, user } = useContext(AuthContext);
+  const queryKey = ['squad', handle, user?.id];
```
capJavert, line 23: invalidateQueries with `['squad', handle]` still work because we use array query keys.

packages/webapp/pages/squads/[handle]/index.tsx
Comment (capJavert) [reply packages/webapp/pages/squads/[handle]/index.tsx:152]: Hmm yeah, but would not work since `ProtectedPage` requires `seo`, and I would have to change the interface to support it and then again move more things around. Would prefer keeping it here as current.

packages/shared/src/components/squads/SquadPageHeader.tsx
capJavert: It would not be valid only if user was not a member, since it would always say `Join the squad to create new posts`. But I think that is copy wise because:
1. we might wanna always invite people to join (no matter of squad preferences)
2. we don't expose actual posting permission to anon users (so we could not calculate it)

We would need API adjustments to allow `memberPostingRank` for anon users and check with product about the messaging. I personally think it would just take time but I can pursue if you want.

packages/shared/src/hooks/useSquadOnboarding.ts
capJavert: I can add in `AuthContext`?

packages/shared/src/components/squads/SquadJoinButton.tsx
rebelchris: Alright valid point, but at which you would almost argue why the custom hook then.
For me it's a wrapping place where all related logic happens so we don't have to do it in multiple places.

Tracking, rollbacks, toast etc

packages/shared/src/components/squads/SquadPageHeader.tsx
rebelchris: Alright let's leave it as is 👍

packages/shared/src/hooks/useSquadOnboarding.ts
rebelchris: Would be a cool addition I think 👍
Seeing more and more of those cases

packages/webapp/pages/squads/[handle]/index.tsx
Comment (rebelchris) [reply packages/webapp/pages/squads/[handle]/index.tsx:152]: No problem 🙏

packages/shared/src/components/squads/SquadJoinButton.tsx
```diff
@@ -0,0 +1,84 @@
...
+    useLeaveSquad({ squad }),
+    {
+      onSuccess: () => {
+        displayToast('👋 You have left the squad.');
+
+        queryClient.invalidateQueries(['squad', squad.handle]);
+        queryClient.invalidateQueries(['squadMembersInitial', squad.handle]);
```
rebelchris, line 45: Do we actually need to invalidate, what's the catch of not doing it?

rebelchris (changes requested):
Also seems we are missing the impression on the join button.
Would suggest to introduce it here in case we forget.
And see you already have the completion introduced here.

Screenshot 2023-07-19 at 13 14 36

packages/shared/src/components/squads/SquadJoinButton.tsx
capJavert: So that this is updated, it is optional though yes. Don't have the especial preference here if its worth it, WDYT?

image

packages/shared/src/components/squads/SquadJoinButton.tsx
capJavert: > Alright valid point, but at which you would almost argue why the custom hook then.

Custom hook here is because we need context and React context to actually get to our tracking function and updating boot. If not we would not need it and it could just be exported.

> Tracking, rollbacks, toast etc

I would argue side effects should not be there because more times those differ from component to component. 

Or we would have to be very specific for what hook is used. Basically each new argument we add to the hook we make it more complex and with more variations on how it works. 

We could thought copy what we are doing `useVotePost` where `mutationHandlers` are exported and reused across different usages of the hooks. 

That way we could still have the function combine all the related logic but side effects can be injected again where needed from predefined functions (or hooks).

packages/shared/src/components/squads/SquadJoinButton.tsx
rebelchris: Ah I see, well let's do it.
Doesn't cost too much, was just wondering what actually would change 😅

capJavert:
@rebelchris good catch, missed that new impression that was added to the table.

packages/shared/src/components/squads/SquadJoinButton.tsx
```diff
@@ -0,0 +1,112 @@
...
+      target_type: TargetType.SquadJoinButton,
+      extra: JSON.stringify({
+        squad: squad.id,
+        origin,
+        squad_type: squad.public ? 'public' : 'private',
+      }),
+    });
+    // @NOTE see https://dailydotdev.atlassian.net/l/cp/dK9h1zoM
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [origin, squad.id, squad.public]);
```
rebelchris, line 46: Not a big fan of passing depedency for the sake of impression calls.
It's important they don't trigger more then once.

I'd either suggest simply removing dependencies here or adding a state to ensure it only fires once per mount.

I know very unlikely it would change, but keep in mind you invalidate queries, so we could accidentally trigger again.

packages/webapp/pages/squads/[handle]/[token].tsx
```diff
@@ -123,44 +126,46 @@ const SquadReferral = ({
...
+    trackEvent({
+      event_name: AnalyticsEvent.Impression,
+      target_type: TargetType.SquadJoinButton,
+      extra: JSON.stringify({
+        squad: member.source.id,
+        origin: Origin.SquadInvitation,
+        squad_type: member.source.public ? 'public' : 'private',
+      }),
+    });
+
```
Comment (rebelchris) [inline packages/webapp/pages/squads/[handle]/[token].tsx:139]: This should not live here right?
It's already handled inside the button which makes more sense.

Comment (rebelchris) [reply packages/webapp/pages/squads/[handle]/[token].tsx:139]: I see I think this is what I meant with duplicate code issue with splitting these buttons.
Can we find a way to uniform the behaviour?

It's pretty dangerous to have duplicated tracking events that do the exact same.
I'm not sure how much other stuff of this is duplicated?

rebelchris (changes requested):
Just some concerns about the tracking.

packages/webapp/pages/squads/[handle]/[token].tsx
Comment (idoshamun) [reply packages/webapp/pages/squads/[handle]/[token].tsx:139]: I agree this code duplication is highly error-prone. Ideally, we want to have one function/hook that triggers events. Duplicating is so risky, especially when it comes to tracking. I would suggest creating a hook to make sure that or a function. Happy with either of them

packages/shared/src/components/squads/SquadJoinButton.tsx
capJavert: But we need those dependencies, if squad id changes or button get's rendered with different squad ID or place (with different origin) we want to trigger impression again right? @rebelchris

rebelchris: Which would be a new mount, afaik there is only a danger of retriggering impressions.
CC: @idoshamun Maybe your thoughts?

idoshamun: If there's a chance that it will get rendered then it's better to have the deps. Otherwise let's stick to no deps. I think when we discussed with tkdodo he also mentioned tracking as one exception use case for not using the proper deps

packages/webapp/pages/squads/[handle]/[token].tsx
Comment (capJavert) [reply packages/webapp/pages/squads/[handle]/[token].tsx]: Yeah agreed, made adjustments here a798c15530f4168858f031567b739efea55d54d7, decided to move tracking logic to button which is now reused on squad invitation page and open squad page.

Changes:
- added `SimpleSquadJoinButton` which contains base logic and tracking for button
- this one is used for `SquadJoinButton` which is the new one with join/leave capabilities needed for open squad
- `SimpleSquadJoinButton` is used on squad invite token page
- refactored layout of squad token page to not render button twice on mobile/desktop (would trigger events twice) - visually layout is the same

idoshamun:
@capJavert don't wait for my approval :)
I'm just jumping randomly on some comments and then disappear 😂

packages/shared/src/components/squads/SquadJoinButton.tsx
rebelchris: @capJavert for me this one is still open 🫣

rebelchris (changes requested):
Just the one thing, from my perspective just ensure it can only fire once, not for instance once on loading when one param is null, and then again once its set.

I really think empty props is ok for tracking for now and then we good to go.

packages/shared/src/components/squads/SquadJoinButton.tsx
capJavert: I'll just concede for now, let's discuss when we get on implementing special tracking hooks. 🤝

rebelchris (approved):
Looks good lets go 🙏

PR #1950, opened 2023-07-17
fix: submitting comment from preview tab

sshanzel:
## Changes
- In the initial implementation, we had `shouldMountInactive={false}` so that it won't send a mutation request for the preview on each type since the parameter of the query changes by not rendering the other tab when it is not opened.
- So when you click the submit button when the preview tab is opened, we tried to fetch the data from the Form for which the textarea is not rendered anymore due to the above condition.
- To fix it, we now render both ends anytime, but we will enable the query only when the `Preview` tab is active, and in that case, we have to track it through the state. Tested it locally, and works as normal now.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1539 #done

### DIFF
--- a/packages/shared/src/components/fields/MarkdownInput/index.tsx
+++ b/packages/shared/src/components/fields/MarkdownInput/index.tsx
@@ -8,6 +8,7 @@ import React, {
   TextareaHTMLAttributes,
   useImperativeHandle,
   useRef,
+  useState,
 } from 'react';
 import classNames from 'classnames';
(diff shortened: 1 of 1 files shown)


TomDowling (approved):
Looks good 🎉

rebelchris (approved):
Thanks, please also check on deployment and maybe update the slack items so Francesco can update the user.

PR #1952, opened 2023-07-18
fix: onboarding intro image

sshanzel:
## Changes
- When it is light mode, the image is not the same size or ratio of the dark mode image.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done

### DIFF
--- a/packages/shared/src/hooks/utils/useThemedAsset.ts
+++ b/packages/shared/src/hooks/utils/useThemedAsset.ts
@@ -5,6 +5,7 @@ import { cloudinary } from '../../lib/image';
 interface UseCloudinaryAsset {
   onboardingIntroduction: string;
   scrollBlock: string;
+  isLight: boolean;
 }
 
 export const useThemedAsset = (): UseCloudinaryAsset => {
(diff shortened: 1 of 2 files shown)


packages/shared/src/hooks/utils/useThemedAsset.ts
```diff
@@ -19,6 +20,7 @@ export const useThemedAsset = (): UseCloudinaryAsset => {
   }, [themeMode]);
 
   return {
+    isLight,
```
capJavert, line 23: Hmm I get why you need it but not sure `isLight` is something that should be returned from this hook since I would expect all keys are specific asset keys. It would more fit useColorMode or useTheme that would just return `light | dark`.

capJavert (commented):
Nice catch 🤘 , one suggestion I had.

packages/shared/src/hooks/utils/useThemedAsset.ts
sshanzel: Agree with that. Would be better off in a separate hook or rename this hook completely. Though for the moment, we will close this PR as we have decided that the fix should actually be from the asset to be having the same ratio and size since the variation is technically of the "same" image.

sshanzel: It seems we will have to keep this PR, but instead update the links and the usage of the assets 😅

rebelchris (approved):
Let's go 🎉

sshanzel:
@rebelchris I've also fixed the missing cookie banner for the onboarding page.

PR #1956, opened 2023-07-19
fix: missing comment tracking

sshanzel:
## Changes
- After further investigation, it seems what we were missing is the code block below:

https://github.com/dailydotdev/apps/blob/10b9383df0e53219f19d21929afdfe8e46541b61/packages/shared/src/hooks/usePostComment.ts#L251-L255

This is what tracks new comments in the past coming from the hook that was reused when making comments.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done

### DIFF
--- a/packages/shared/src/components/post/PostEngagements.tsx
+++ b/packages/shared/src/components/post/PostEngagements.tsx
@@ -12,6 +12,7 @@ import { PostComments } from './PostComments';
 import { PostUpvotesCommentsCount } from './PostUpvotesCommentsCount';
 import { Comment } from '../../graphql/comments';
 import { Origin } from '../../lib/analytics';
+import { postAnalyticsEvent } from '../../lib/feed';
 
 const AuthorOnboarding = dynamic(
   () => import(/* webpackChunkName: "authorOnboarding" */ './AuthorOnboarding'),
@@ -61,6 +62,9 @@ function PostEngagements({
 
   const onCommented = (comment: Comment, isNew?: boolean) => {
     if (isNew) {
+      postAnalyticsEvent('comment post', post, {
+        extra: { commentId: comment.id, origin: 'comment modal' },
+      });
       setPermissionNotificationCommentId(comment.id);
       onShowShareNewComment(comment.id);
     }
--- a/packages/shared/src/components/post/PostEngagements.tsx
+++ b/packages/shared/src/components/post/PostEngagements.tsx
@@ -11,7 +11,7 @@ import { PostActions, ShareBookmarkProps } from './PostActions';
 import { PostComments } from './PostComments';
 import { PostUpvotesCommentsCount } from './PostUpvotesCommentsCount';
 import { Comment } from '../../graphql/comments';
-import { Origin } from '../../lib/analytics';
+import { AnalyticsEvent, Origin } from '../../lib/analytics';
 import { postAnalyticsEvent } from '../../lib/feed';
 
 const AuthorOnboarding = dynamic(
@@ -62,7 +62,7 @@ function PostEngagements({
 
   const onCommented = (comment: Comment, isNew?: boolean) => {
     if (isNew) {
-      postAnalyticsEvent('comment post', post, {
+      postAnalyticsEvent(AnalyticsEvent.CommentPost, post, {
         extra: { commentId: comment.id, origin: 'comment modal' },
       });
       setPermissionNotificationCommentId(comment.id);
--- a/packages/shared/src/lib/analytics.ts
+++ b/packages/shared/src/lib/analytics.ts
@@ -30,6 +30,7 @@ export enum LoginTrigger {
 }
 
 export enum AnalyticsEvent {
+  CommentPost = 'comment post',
   Impression = 'impression',
   ManageTags = 'click manage tags',
   SearchTags = 'search tags',


packages/shared/src/components/post/PostEngagements.tsx
```diff
@@ -61,6 +62,9 @@ function PostEngagements({
 
   const onCommented = (comment: Comment, isNew?: boolean) => {
     if (isNew) {
+      postAnalyticsEvent(AnalyticsEvent.CommentPost, post, {
+        extra: { commentId: comment.id, origin: 'comment modal' },
```
sshanzel, line 66: Similar to the `open comment modal` event, we kept the origin here to make the data consistent.

idoshamun (approved):

packages/shared/src/components/post/PostEngagements.tsx
rebelchris, line 66: I see you import origin, why don't you use the origin enum here 🙏

rebelchris (approved):
small request when fixed good to go

packages/shared/src/components/post/PostEngagements.tsx
rebelchris: Just to check this triggers and comments and reply to comments right?

idoshamun: replies should have different event if I remember correctly

rebelchris: @idoshamun I don't seem to find a different one in the analytics taxonomy.
Did we maybe miss it initially?

idoshamun: Oh you're right! We use the same event `comment post` but the `commentId` field in extra should indicate the parent comment if it's a reply. @sshanzel please adjust

rebelchris: Actually also the origin is not actually correct anymore, it can't be the modal as it no longer exists.
Should probably check with @idoshamun or Dave what's more appropriate. (And maybe ask to update the taxonomy sheet for this one)

idoshamun: We don't use `origin` for the `comment post` in our analysis right now. Since it's kinda streamlines already as you mentioned it's safe to ignore

sshanzel: Oh, right. Missed that one. Will utilize the enum before merging.

sshanzel: Amazing. I will update the `commentId` to be coming from the parent comment's id.

packages/shared/src/components/post/PostEngagements.tsx
```diff
@@ -59,8 +60,11 @@ function PostEngagements({
     onShowShareNewComment,
   } = useShareComment(analyticsOrigin, enableShowShareNewComment);
 
-  const onCommented = (comment: Comment, isNew?: boolean) => {
+  const onCommented = (comment: Comment, isNew: boolean, parentId?: string) => {
     if (isNew) {
+      postAnalyticsEvent(AnalyticsEvent.CommentPost, post, {
+        extra: { commentId: parentId, origin: Origin.CommentModal },
```
rebelchris, line 66: Please remove the origin here as discussed above.

Also commentId should then only be added if it's available

packages/shared/src/lib/analytics.ts
```diff
@@ -1,4 +1,5 @@
 export enum Origin {
+  CommentModal = 'comment modal',
```
rebelchris, line 2: This can be removed again, sorry for back and forth

packages/shared/src/components/comments/SubComment.tsx
```diff
@@ -58,9 +61,9 @@ function SubComment({
           {...inputProps}
           className={className}
           post={props.post}
-          onCommented={(data, isNew) => {
+          onCommented={(...params) => {
             onReplyTo(null);
-            onCommented(data, isNew);
+            onCommented(...params);
```
rebelchris, line 66: What's the difference between these 2?

rebelchris (changes requested):
Small changes, and one question

packages/shared/src/components/comments/SubComment.tsx
rebelchris: I mean `onReplyTo` and `onCommented`

sshanzel: `onReplyTo` is some state holder for which comment they are replying to. Since it was successful, the `replyTo` state is getting null in this one.

rebelchris: Ah good, wasn't sure if it was the success callback maybe :) 
In that case I have my answer

packages/shared/src/components/post/PostEngagements.tsx
sshanzel: Yes. That parameter is optional. So that value can become undefined when applicable.

rebelchris: Ah yes but wouldn't this be cleaner?

```
...(parentId && { commentId: parentId })
```

That way we don't even add the prop when it's empty

sshanzel: I mean, if it is undefined it is not really passed on the tracking so it would be as if there was no `commentId` prop which is the result of the suggestion.

rebelchris: Are you sure, does it get send without the commentId if parentId is empty?
Or do we get commentId: undefined/null?

If the first is the case and you tested it's fine.

rebelchris: Ok all good then

rebelchris (approved):

PR #1957, opened 2023-07-19
feat: Updating the error messages on the login form

TomDowling:
## Changes
Changed the modal header text size for the auth modals as well as fixing the validation / error states for the inputs in login. As well as adding a toast when a user try's to register with an existing email
login-form

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1164 #done

### DIFF
--- a/packages/shared/src/components/auth/AuthDefault.tsx
+++ b/packages/shared/src/components/auth/AuthDefault.tsx
@@ -21,6 +21,7 @@ import AuthContainer from './AuthContainer';
 import AuthModalHeader from './AuthModalHeader';
 import { ExperimentWinner } from '../../lib/featureValues';
 import ConditionalWrapper from '../ConditionalWrapper';
+import { useToastNotification } from '../../hooks/useToastNotification';
 
 interface AuthDefaultProps extends AuthFormProps {
   children?: ReactNode;
@@ -82,6 +83,8 @@ const AuthDefault = ({
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [shouldLogin]);
 
+  const { displayToast } = useToastNotification();
+
   const onEmailSignup = async (e: React.FormEvent) => {
     e.preventDefault();
 
@@ -106,6 +109,9 @@ const AuthDefault = ({
 
     if (res?.result) {
       setRegisterEmail(email);
+      displayToast(
+        "There's already an account for the same credentials. Can you please try logging in instead?",
+      );
       return setShouldLogin(true);
     }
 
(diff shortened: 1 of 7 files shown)


packages/shared/src/components/auth/LoginForm.tsx
```diff
@@ -43,6 +43,7 @@ function LoginForm({
     const form = formToJson<LoginFormParams>(e.currentTarget);
     onPasswordLogin(form);
   };
+  const [hasFocused, setHasFocused] = useState(autoFocus);
```
rebelchris, line 46: Technically the naming is in reverse here, might cause confusion down the line.
You set hasfocus to false, but actually you focussed it.

Maybe shouldFocus is better in that case?
Or negate the start to be reversed !autoFocus and set it to true

packages/shared/src/components/fields/PasswordField.tsx
```diff
@@ -64,8 +64,8 @@ export function PasswordField({
       valueChanged={onChange}
       leftIcon={<LockIcon size={IconSize.Small} />}
       hint={!!value && showStrength ? hint : props.hint}
-      validityChanged={setIsValid}
-      valid={isValid || !props.hint}
+      validityChanged={() => !props.hint && setIsValid}
```
rebelchris, line 67: It still looks like we miss the passing of the actual value to setIsValid as the () is empty.
Not sure it makes a difference for areas where we use an internal hint actually.
As for instance the hint space password field (unused atm)

rebelchris (changes requested):
Neat, small comments but with those should be fine

packages/shared/src/components/auth/AuthDefault.tsx
```diff
@@ -82,6 +83,8 @@ const AuthDefault = ({
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [shouldLogin]);
 
+  const { displayToast } = useToastNotification();
```
rebelchris, line 86: Can you just move this to the top where all the other states are.
Should not be placed below render functions

rebelchris (changes requested):
Small comment, otherwise we can go

rebelchris (approved):
Good from my side

sshanzel (approved):
Looks good to me. Just that one line I mentioned to be removed.

packages/shared/src/components/auth/LoginForm.tsx
```diff
@@ -70,7 +72,6 @@ function LoginForm({
         value={email}
         data-testid="login_email"
         saveHintSpace
```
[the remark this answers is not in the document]
rebelchris: Should remain to keep the same space we use for all fields.
it was the easier fix to simply leave it although naming now is a bit funky.

PR #1958, opened 2023-07-19
feat: growthbook support

idoshamun:

Add support in our new experimentation platform. This is done using their react package with a bit of sugar coating on top to optimize performance using our boot endpoint. Feature rules are encrypted which to prevent from users to understand our experimentation rules. This is also a concern as I'm not sure how decrypting will impact our performance.

Another thing to consider, we have to check the additional bundle size. Perhaps we should take only the parts we use from the growthbook package because there's so many features there but we barely use them.

For more info: https://docs.growthbook.io/lib/react

EDIT: I took out the relevant part from the SDK and embedded them in our repo. It had a significant impact on the bundle size. Let me know if you'd like me to do it any other way. I think we can squeeze it even more.

NOTE: You don't have to review the `growthbook` folder. The code is taken from their repo (see my comment above)

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done


packages/shared/src/components/crypto.ts
```diff
@@ -0,0 +1,29 @@
...
+    ['encrypt', 'decrypt'],
+  );
+  const [iv, cipherText] = input.split(':');
+  const plainTextBuffer = await subtle.decrypt(
+    { name: algorithmName, iv: base64ToBuf(iv) },
+    keyObject,
+    base64ToBuf(cipherText),
+  );
+
+  return new TextDecoder().decode(plainTextBuffer);
```
capJavert, line 28: I think this should be okay, if needed one idea is to potentially decrypt on the SSR side if needed.

packages/shared/src/components/GrowthBookProvider.tsx
```diff
@@ -0,0 +1,127 @@
...
+export const GrowthBookProvider = ({
+  app,
+  user,
+  deviceId,
+  version,
+  experimentation,
+  fetchMethod = fetch,
+  children,
+}: GrowthBookProviderProps): ReactElement => {
+  const { mutateAsync: sendAllocation } = useMutation(
```
capJavert, line 41: Allocation is basically adding user to experiment?

packages/shared/src/components/GrowthBookProvider.tsx
```diff
@@ -0,0 +1,127 @@
...
+      });
+      await res?.text();
+    },
+    {
+      retry: 3,
+    },
+  );
+
+  const callback = useRef<Context['trackingCallback']>();
+  const [gb] = useState<GrowthBook>(
```
capJavert, line 65: Could this not be initiated outside of the provider since we don't set it anywhere?

packages/shared/src/lib/growthbook/index.ts
```diff
@@ -0,0 +1,3 @@
+export type { Context } from './types/growthbook';
```
capJavert, line 1: Maybe we can use [pnpm patch](https://pnpm.io/cli/patch) to remove all the garbage we don't need from their SDK. This way we will not need to add that code to our codebase and it would still load through node modules.

capJavert (commented):
Simple implementation actually 👍 

I am a little out of loop maybe but would like to understand few points on why we were looking for alternative platform, just in short terms, it would help to evaluate how this new integration would sit in the long term.

packages/shared/src/components/GrowthBookProvider.tsx
idoshamun: I do use the tracingCallback. I need to have the ref available

packages/shared/src/lib/growthbook/index.ts
idoshamun: Good idea. I'll see what can be improved :)

packages/shared/src/components/GrowthBookProvider.tsx
```diff
@@ -0,0 +1,127 @@
...
+export type GrowthBookProviderProps = {
+  app: BootApp;
+  user: BootCacheData['user'];
+  deviceId: string;
+  version?: string;
+  experimentation?: {
+    f: string;
+    e: string[];
+  };
+  fetchMethod?: typeof fetch;
```
rebelchris, line 28: Ugh, really time we context this 😅

packages/shared/src/components/GrowthBookProvider.tsx
rebelchris: I'm also curious about this one, what does it entail?

packages/shared/src/components/GrowthBookProvider.tsx
```diff
@@ -0,0 +1,127 @@
...
+      atts = {
+        ...atts,
+        loggedIn: false,
+        firstVisit: user.firstVisit,
+      };
+    }
+    gb.setAttributes(atts);
+  }, [app, user, deviceId, gb, version]);
+
+  return <Provider growthbook={gb}>{children}</Provider>;
```
rebelchris, line 126: Does our boot not handle most of the experiments in this case?
The context just wraps then right, but why does it need it's own calls?

rebelchris (commented):
Looks neat, some basic questions for my understanding, but overall nothing pops out

packages/shared/src/components/GrowthBookProvider.tsx
idoshamun: Yes exactly

packages/shared/src/components/GrowthBookProvider.tsx
idoshamun: This is how they built their SDK. That's how you can use their hooks everywhere to get the latest feature value and it all plays together

packages/shared/src/components/GrowthBookProvider.tsx
sshanzel: We have a hook that handles this. Should be named `useRequestProtocol`. It can return the `fetchMethod` which is set automatically by the companion when needed.

rebelchris: Ah yes, totally forgot about that, that should do neatly for this purpose.

idoshamun: I'll check that out!

packages/shared/src/lib/growthbook/index.ts
idoshamun: At the end I forked the project, removed whatever needed and used pnpm to force using my fork. @capJavert wdyt?

capJavert: Sounds good also, we just need to be aware of maintaining that when they make updates 👍

packages/shared/src/components/layout/MainLayoutHeader.tsx
```diff
@@ -114,6 +116,7 @@ function MainLayoutHeader({
               />
             )}
           </div>
+          <div className="typo-body">{dummyFeature}</div>
```
sshanzel, line 119: Just tagging to be removed after debugging.

sshanzel (commented):
Looks amazing to me, end to end. Thank you for setting it up. Just tagged one part for reverting before the merge phase.

idoshamun:
@rebelchris merge whenever you want :)

rebelchris (approved):
Let's see what it can do

PR #1944, opened 2023-07-13
feat: open squads

capJavert:
### Describe what this PR does
- Squads MVP aka Open Squads
- DR [here](https://dailydotdev.atlassian.net/l/cp/2ke51f1m)

## Events

Did you introduce any new tracking events?
Don't forget to update the [Analytics Taxonomy sheet](https://docs.google.com/spreadsheets/d/18Lv7zXges9QfVX5VYL1a-Hyl0e1sQ3sLr0OK8YZWKXI/edit#gid=0)

Log the new/changed events below:

| Type   | event_name  | value |
|--------|-------------|-------|
| Change/New | event name  | extra: { ... } |

### On those affected packages:
- [ ] Have you done sanity checks in the webapp?
- [ ] Have you done sanity checks in the extension?
- [ ] Does this not break anything in companion?

### Did you test the modified components media queries?
- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

#### Did you test on actual mobile devices?
- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done

PR #1965, opened 2023-07-25
fix: remove squad creation constraint

rebelchris:
- Remove the squad creation constraint

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1527 #done


packages/shared/src/components/ShareOptionsMenu.tsx
```diff
@@ -47,10 +47,10 @@ export default function ShareOptionsMenu({
   onHidden,
   contextId = 'share-context',
 }: ShareOptionsMenuProps): ReactElement {
+  const { openNewSquadModal } = useCreateSquadModal();
```
capJavert, line 50: I think we should merge this PR into https://github.com/dailydotdev/apps/pull/1944 instead, since there is additional cleanup there that was started and pending. It will raise more conflicts like this. We also have `useSquadNavigation` hook there that can also potentially be removed with `useCreateSquadModal` all together.

capJavert (commented):
Generally looks good but we might wanna merge it in open squads PR.

packages/shared/src/components/ShareOptionsMenu.tsx
rebelchris: We want to merge this to main today, that's kind of the constraint for now.
I have another one open to merge it into squads-mvp, so will adjust once that's done.

capJavert (approved):
🚀

PR #1970, opened 2023-07-26
feat: notification preference

sshanzel:
- Squad notifications modal to adjust user preferences.
- Notifications page item to show options whether to mute or unmute a thread.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done


packages/shared/src/graphql/source/common.ts
```diff
@@ -0,0 +1,23 @@
...
+  client.setQueryData<Squad>(queryKey, (oldData) => ({
+    ...oldData,
+    currentMember: {
+      ...oldData.currentMember,
+      flags: {
+        ...(oldData.currentMember.flags ?? {}),
+        ...update,
+      },
+    },
+  }));
```
capJavert, line 22: We might wanna check if `oldData` is undefined and then not set anything since query will fetch fresh data for the first time when mounted. Since undefined means it did not mount before.

packages/shared/src/graphql/sources.ts
```diff
@@ -26,12 +26,17 @@ export enum SourcePermissions {
   WelcomePostEdit = 'welcome_post_edit',
 }
 
+export interface SourceMemberFlag {
+  hideFeedPosts: boolean;
```
capJavert, line 30: This one can be `null` on the API so we should make it optional.

packages/webapp/pages/notifications.tsx
```diff
@@ -91,6 +134,73 @@ const Notifications = (): ReactElement => {
...
+
+  const getIcon = () => {
+    if (isFetching) return <Loader />;
+
+    return preferences[0]?.status === NotificationPreferenceStatus.Muted ? (
+      <BellIcon />
+    ) : (
+      <BellDisabledIcon />
+    );
+  };
```
capJavert, line 171: More React way is I thin calling it `Icon` and then just rendering below like `<Icon />`, same for `getCopy` below.

packages/webapp/pages/notifications.tsx
```diff
@@ -34,6 +43,40 @@ const hasUnread = (data: InfiniteData<NotificationsData>) =>
...
+  },
+  [NotificationType.SquadPostAdded]: {
+    mute: 'Mute new posts',
+    unmute: 'Unmute new posts',
+  },
+  [NotificationType.SquadMemberJoined]: {
+    mute: 'Mute new members joining',
+    unmute: 'Unmute new members joining',
+  },
+};
```
capJavert: Maybe extract this to common, just think it might be reused in other components/pages down the line.

packages/shared/src/hooks/notifications/useNotificationPreference.ts
```diff
@@ -0,0 +1,113 @@
...
+        updateFlagsCache(client, squad, user, { hideFeedPosts: false });
+      },
+    },
+  );
+
+  return {
+    hideSourceFeedPosts: () => hideSourceFeedPostsAsync(squad?.id),
+    showSourceFeedPosts: () => showSourceFeedPostsAsync(squad?.id),
+    isFetching: isLoading,
+    preferences: data ?? [],
```
capJavert, line 108: You can move `[]` to `initialData: () => []` in `useQuery`.

capJavert (changes requested):
Smooth work as always, I really like how `useNotificationPreference` hook implementation ended up simple after all the work you did. Few comments mainly due to refactoring but other then that 💪

packages/shared/src/hooks/squads/useSquad.ts
```diff
@@ -20,7 +21,7 @@ interface UseSquad {
 
 export const useSquad = ({ handle }: UseSquadProps): UseSquad => {
   const { isFetched: isBootFetched, user } = useContext(AuthContext);
-  const queryKey = ['squad', handle, user?.id];
+  const queryKey = generateQueryKey(RequestKey.Squad, user, handle);
```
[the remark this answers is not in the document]
sshanzel: Right. This was my concern too. Let me check whether there was invalidation involved in any of the existing keys.

packages/shared/src/graphql/source/common.ts
sshanzel: Right. Better to check than break the prod 😅

packages/shared/src/hooks/notifications/useNotificationPreference.ts
sshanzel: Oh, right. That is much better.

packages/shared/src/hooks/notifications/useNotificationPreference.ts
```diff
@@ -0,0 +1,113 @@
...
+    {
+      onSuccess: () => {
+        updateFlagsCache(client, squad, user, { hideFeedPosts: false });
+      },
+    },
+  );
+
+  return {
+    hideSourceFeedPosts: () => hideSourceFeedPostsAsync(squad?.id),
+    showSourceFeedPosts: () => showSourceFeedPostsAsync(squad?.id),
```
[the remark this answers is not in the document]
rebelchris: Just checking why is this resolved but not actioned on? 👀

packages/webapp/pages/notifications.tsx
```diff
@@ -91,6 +126,73 @@ const Notifications = (): ReactElement => {
 
   usePromotionModal();
 
+  const { show } = useContextMenu({ id: contextId });
```
rebelchris, line 108: I personally would move this to a custom component, as we currently do for context menu's
Just in case it becomes general.

packages/webapp/pages/notifications.tsx
```diff
@@ -91,6 +126,73 @@ const Notifications = (): ReactElement => {
...
+      ? [
+          {
+            type: notificationPreferenceMap[notification.type],
+            referenceId: notification.referenceId,
+          },
+        ]
+      : [],
+  });
+
+  const onOptionsClick = (e: React.MouseEvent, item: Notification) => {
```
rebelchris, line 126: This logic could also go in that hook

packages/shared/src/hooks/squads/useSquad.ts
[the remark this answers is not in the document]
sshanzel: Fixed here: https://github.com/dailydotdev/apps/pull/1970/commits/f36b85d668d16d3be4272fc95c1b8888f7371f1b

rebelchris (changes requested):
Must say very clean implementation Lee!
Really nice work here 👏

Just some minor comments

packages/shared/src/components/tooltips/Portal.tsx
```diff
@@ -6,7 +6,7 @@ interface PortalProps {
 }
 
 function Portal({ children }: PortalProps): ReturnType<typeof createPortal> {
-  return createPortal(children, document.body);
+  return createPortal(children, globalThis?.document?.body);
```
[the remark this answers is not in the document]
sshanzel: This actually fixes the issue of directly using `document`. SSR would fail without this as they don't have yet a context of the document.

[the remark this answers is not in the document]
rebelchris: Makes sense, my point still stands please 🙏

packages/webapp/pages/notifications.tsx
sshanzel: Right. Let me move this and the `onOptionsClick` you mentioned.

packages/webapp/pages/notifications.tsx
rebelchris: I see you adjusted regardless thanks, looks a bit more split out 🙏

rebelchris (approved):
Thanks for the cleanup, looks solid to me.
@capJavert if you can just give your last run-through 👍

packages/shared/src/hooks/notifications/useNotificationPreference.ts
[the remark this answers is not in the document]
sshanzel: Oh, I just manually clicked it once I commit but before the push. Bad habit, I guess 😓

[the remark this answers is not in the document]
rebelchris: But was the idea not to put them in useCallback?
Same with the `preferences` below not sure I read the feedback the same as you adjusted?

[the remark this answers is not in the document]
sshanzel: Oh, the idea is the opposite, they should be in a callback since they are not a stable reference when we do arrow function.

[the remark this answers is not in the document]
rebelchris: You say "They _should_" but they are not correct?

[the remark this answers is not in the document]
sshanzel: They are in a callback on the latest code now. This was an outdated comment before the fix.

packages/shared/src/hooks/notifications/useNotificationPreference.ts
sshanzel: This is an outdated comment, kindly check the files tab for the most updated ones.

rebelchris: Oh dang, for some reason it kept showing the old one, I see it now!
Sorry github must be tripping on me

packages/shared/src/hooks/notifications/useNotificationPreference.ts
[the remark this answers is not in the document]
rebelchris: Oh dang, for some reason it kept showing the old one, I see it now!
Sorry github must be tripping on me

packages/shared/src/graphql/sources.ts
capJavert: This one missing `optional` 🙏

packages/shared/src/graphql/source/common.ts
```diff
@@ -0,0 +1,27 @@
...
+          currentMember: {
+            ...oldData.currentMember,
+            flags: {
+              ...(oldData.currentMember.flags ?? {}),
+              ...update,
+            },
+          },
+        }
+      : null,
+  );
```
capJavert, line 26: I would make it early if and return to not call `setQueryData` at all if no oldData, that is because like this it will actually write data for that query to be `null` vs not existing at all which can mess with initial data checks on the query.

capJavert (approved):
Looks ready to go 🚀 Two minor comments.

packages/shared/src/graphql/sources.ts
sshanzel: I don't know what's up with GitHub, but I have actually changed this already.

https://github.com/dailydotdev/apps/pull/1970/files#diff-01cb5b194208871f2275f21d46482164aa75f246128e30dd67acc408efa1f735R29-R30

packages/shared/src/graphql/source/common.ts
sshanzel: Makes sense. Let's do that.

sshanzel: But the old data is coming from the `setQueryData`. Do you mean we should fetch first through `getQueryData`?

capJavert: Yeah, that should work.

packages/shared/src/hooks/useJoinSquad.ts
```diff
@@ -41,7 +44,8 @@ export const useJoinSquad = ({
     });
 
     addSquad(result);
-    queryClient.invalidateQueries(['squad', squad.handle]);
+    const queryKey = generateQueryKey(RequestKey.Squad, user, squad.handle);
+    queryClient.invalidateQueries(queryKey);
```
capJavert, line 48: Ah found one edge case. @sshanzel 

In `SquadJoinButton` we pass `joinSquad` to `showLogin`, but since in that function reference user is still for anonymous it actually clears the cache for `[squad, anonymous, handle]` and the new key (because user is logged) is `[squad, user.id, handle]`. Which then shows stale page like user is not part of squad. If you refresh the page he is of course. 

We can fix in two ways I think:
- by just invalidating all the squad data for now invalidateQueries([RequestKey.Squad])
- using useRef to always have a fresh instance of `joinSquad` when calling in login success callbacks

capJavert: Don't know if you have any other ideas, we can also go through the case on a quick call.

sshanzel: If you now have a new query key by the `anonymous` getting replaced with the user's id, then it should automatically refetch since the query params have changed.

sshanzel: Do you have any steps where I can reproduce this?

capJavert: Yeah, but the refetch happens at the moment user logs in (in useSquad) `joinSquad` mutation is done so it gets that user is still not part of squad.

PR #1973, opened 2023-07-27
fix: companion commenting and tracking

sshanzel:
## Changes
- Fixed the query key as the comparison is failing due to the last item of the key being an array. Fixed by spreading the array.
- Moved the tracking to instead happen right away instead of waiting in the callback since all new comments must be tracked.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1516 #done

### DIFF
--- a/packages/shared/src/components/fields/MarkdownInput/CommentMarkdownInput.tsx
+++ b/packages/shared/src/components/fields/MarkdownInput/CommentMarkdownInput.tsx
@@ -137,7 +137,11 @@ export function CommentMarkdownInput({
   const mutation = parentCommentId
     ? COMMENT_ON_COMMENT_MUTATION
     : COMMENT_ON_POST_MUTATION;
-  const { mutateAsync: onComment, isLoading: isCommenting } = useMutation(
+  const {
+    mutateAsync: onComment,
+    isLoading: isCommenting,
+    isSuccess,
+  } = useMutation(
     (variables: SubmitComment) =>
       requestMethod(graphqlUrl, mutation, variables, {
         requestKey: JSON.stringify(key),
@@ -199,6 +203,7 @@ export function CommentMarkdownInput({
         sourceId={sourceId}
         showUserAvatar
         isLoading={isCommenting || isEditing}
+        disabledSubmit={isSuccess}
         initialContent={initialContent}
         textareaProps={{
           name: 'content',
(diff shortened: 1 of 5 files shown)


packages/shared/src/components/fields/MarkdownInput/CommentMarkdownInput.tsx
```diff
@@ -129,6 +133,12 @@ export function CommentMarkdownInput({
 
     if (!editCommentId) {
       updatePostCache(client, postId, { numComments: post.numComments + 1 });
+
+      trackEvent(
```
rebelchris, line 137: Does this mean we missed some tracking events before from companion?

rebelchris (approved):
Just some secondary concerns.
@idoshamun Just to note I think we might have miss some tracking events from companion.

packages/shared/src/components/fields/MarkdownInput/CommentMarkdownInput.tsx
sshanzel: Only after the migration we did from Modal to CommentBox.

idoshamun: Why don't we send this event from the component itself? The fact we have to add everywhere is a bit messy

idoshamun (commented):
@sshanzel have you validated that it works locally?

sshanzel:
> @sshanzel have you validated that it works locally?

Yes, I installed companion locally too. Then tested the normal commenting and both ends have worked well 🚀

packages/shared/src/components/fields/MarkdownInput/CommentMarkdownInput.tsx
sshanzel: Yes, this is the fix now, we are sending it from the component itself than the callback.

rebelchris: @idoshamun Just mentioned you as I'm not sure data team needs to be aware of these things?

idoshamun: Thanks. Since we don't have so many companion users it's fine :)

PR #1974, opened 2023-07-27
fix: missing query params on redirect

sshanzel:
## Changes
- Keep query params when redirecting for onboarding.
- Tested and tried to include the parameters. Also checked on extension and verified to be working.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done

### DIFF
--- a/packages/shared/src/components/MainLayout.tsx
+++ b/packages/shared/src/components/MainLayout.tsx
@@ -158,7 +158,7 @@ export default function MainLayout({
   useEffect(() => {
     if (!shouldRedirectOnboarding) return;
 
-    router.push(`${webappUrl}/onboarding`);
+    router.push({ pathname: `${webappUrl}/onboarding`, query: router.query });
   }, [shouldRedirectOnboarding, router]);
 
   if (
--- a/packages/shared/src/components/MainLayout.tsx
+++ b/packages/shared/src/components/MainLayout.tsx
@@ -158,13 +158,21 @@ export default function MainLayout({
   useEffect(() => {
     if (!shouldRedirectOnboarding) return;
 
-    const url = new URLSearchParams();
+    const onboarding = `${webappUrl}/onboarding`;
+    const entries = Object.entries(router.query);
 
-    Object.entries(router.query).forEach(([key, value]) => {
-      url.append(key, value as string);
+    if (entries.length === 0) {
+      router.push(onboarding);
+      return;
+    }
+
+    const params = new URLSearchParams();
+
+    entries.forEach(([key, value]) => {
+      params.append(key, value as string);
     });
 
-    router.push(`${webappUrl}/onboarding?${url.toString()}`);
+    router.push(`${webappUrl}/onboarding?${params.toString()}`);
   }, [shouldRedirectOnboarding, router]);
 
   if (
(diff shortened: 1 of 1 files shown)


rebelchris (approved):
@idoshamun Just for your eyes, looks good to me 🙏

packages/shared/src/components/MainLayout.tsx
```diff
@@ -158,7 +158,13 @@ export default function MainLayout({
...
     if (!shouldRedirectOnboarding) return;
 
-    router.push(`${webappUrl}/onboarding`);
+    const url = new URLSearchParams();
+
+    Object.entries(router.query).forEach(([key, value]) => {
+      url.append(key, value as string);
+    });
+
+    router.push(`${webappUrl}/onboarding?${url.toString()}`);
```
idoshamun, line 167: Let's only validate that there are query params before we add them. I don't want the url to look weird without a reason :)

capJavert, line 167: I think more efficient would be to use Object argument for router.push, it will avoid forEach and manual concatenation of url. Currently the `?` would be always hanging with or without query params.

```js
router.push({
  pathname: `${webappUrl}/onboarding`,
  query: router.query,
});
```

capJavert (changes requested):
Fix works but just some edge cases.

packages/shared/src/components/MainLayout.tsx
sshanzel: Initially did this but it wasn't working with the extension.

sshanzel: The browser automatically removes the question mark after the redirection. Though to ensure all browsers, let's do that.

idoshamun: Yes, extension is not really a nextjs app which makes it harder. Good catch @sshanzel!

capJavert: Yeah, but you are using `router.push` still, it would not work even for that right?

sshanzel: If it is a simple redirection (passing just a string param), it seems to work with `router.push`.

capJavert: Weird are we polyfilling `router` in extension maybe and then it supports only string, because `useRouter` should not even work since there is not nextjs in extension context right?

sshanzel: No, we do not polyfill but it works in simple cases such as redirecting.

sshanzel: Would not work for the webapp in that case. URL constructor requires a valid http/s URL. And the `webappUrl` is just `/` in the webapp environment. I'd say we move forward with this.

capJavert: Oh man, didn't now that, seem clunky to call it URL then. Yeah, agreed let's merge, sorry for drilling it just looked as something we can make better. 🚀

sshanzel: No, it is cool. Would have chosen the URL path too if only the webapp url works on both cases. It ensures everything would go as intended. I trust native classes than my own code 😂

capJavert (approved):

idoshamun (approved):

PR #1979, opened 2023-07-27
fix: redirection from extension to webapp

sshanzel:
## Changes
- Missing location - tested and worked on both ends.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1573 #done

### DIFF
--- a/packages/shared/src/hooks/useSquadNavigation.ts
+++ b/packages/shared/src/hooks/useSquadNavigation.ts
@@ -3,6 +3,7 @@ import { useCallback, useContext, useEffect } from 'react';
 import { Origin } from '../lib/analytics';
 import { AuthTriggers } from '../lib/auth';
 import AuthContext from '../contexts/AuthContext';
+import { webappUrl } from '../lib/constants';
 
 type OpenNewSquadProps = { origin: Origin };
 type EditSquadProps = { handle: string };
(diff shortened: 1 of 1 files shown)


packages/shared/src/hooks/useSquadNavigation.ts
```diff
@@ -21,14 +22,14 @@ export const useSquadNavigation = (): UseSquadNavigation => {
         showLogin(AuthTriggers.CreateSquad);
         return;
       }
-      router.push(`/squads/new?origin=${props.origin}`);
+      router.push(`${webappUrl}/squads/new?origin=${props.origin}`);
```
capJavert, line 25: Would this not introduce double `//` in webapp since webappUrl is `/`?

capJavert (approved):
Just one concern but ready to go.

packages/shared/src/hooks/useSquadNavigation.ts
sshanzel: Not sure how `useRouter` handles it, but it works 😂 
Probably fixed under the hood.

https://github.com/dailydotdev/apps/assets/13744167/7a6aba0d-bb5b-4dad-af20-60c1e133ae79

capJavert: Yeah cool it works, but I would rather check those cases because it could just be a router bug and when they stop supporting that edge case our app breaks. 

Because now the string we are sending up there is `//squads/new?origin=sidebar` which is invalid.

PR #1981, opened 2023-07-27
feat: post page anonymous user fixes

capJavert:
General fixes for post page
1. fixed case where view post mutation was triggering for anonymous user
2. found an issue where post.source and squad source (from useSquad) hook could get mixed up
  - fixed by only trusting squad data from `useSquad` because this one is used across the app
3. fixed the case where squad could be added to sidebar two times after social login if user was already part of squad but went through login after clicking "Join squad"

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done


sshanzel (approved):
Nice fix! 🚀

capJavert:
Hmm actually I might have found something, it's and edge case with login state, investigating.

rebelchris (approved):

packages/shared/src/components/post/SquadPostWidgets.tsx
```diff
@@ -23,14 +23,18 @@ interface PostWidgetsProps
 const SquadCard = ({ squadSource }: { squadSource: Squad }) => {
   const { isFetched } = useAuthContext();
   const { id: squadId, handle } = squadSource;
-  const { squad = squadSource } = useSquad({ handle });
+  const { squad } = useSquad({ handle });
```
sshanzel, line 26: Oh, I just saw this. Why did we end up doing another squad request in this widget?

capJavert: It's just to get the data but if you open post modal from squad page no request will be made since it will all ready be fetched. 

Addition is for the single post page where source request will have to be made but I think that is ok since we need that data. And then again if you go to squad page from that page there will be no request since it will be cached from post page. Does that make sense?

packages/shared/src/components/post/SquadPostContent.tsx
```diff
@@ -62,12 +64,12 @@ function SquadPostContent({
...
 
   useEffect(() => {
-    if (!post?.id) {
+    if (!post?.id || !user?.id) {
       return;
     }
 
     onSendViewPost(post.id);
-  }, [post?.id, onSendViewPost]);
+  }, [post.id, onSendViewPost, user?.id]);
```
rebelchris, line 72: @idoshamun I just wanted to double check with you if this had any potential impact on the metrics/feed for anonymous user?

idoshamun: This could have had minor impact in case of a race condition. But I think it's an edge case, so I wouldn't worry about that for now. Thanks for pointing this out!

PR #1982, opened 2023-07-27
fix: squad page cache after login

capJavert:
In `SquadJoinButton` we pass `joinSquad` to `showLogin`, but since in that function reference user is still for anonymous it actually clears the cache for `[squad, anonymous, handle]` and the new key (because user is logged) is `[squad, user.id, handle]`. Which then shows stale page like user is not part of squad. If you refresh the page he is of course. 

In this PR I clear all squad cache by `RequestKey.Squad` to make sure caches are cleared and user see proper join state after login.

Downside is that all the squad source query data (not the feed) will be invalidated in react query but since we do refetch that data on each visit to squad page it does not make any difference performance wise.

Alternative would be to provide [predicate function to invalidateQueries](https://tanstack.com/query/v4/docs/react/guides/query-invalidation) to check each query key manually but I think complication in this case. 

- Original discussion https://github.com/dailydotdev/apps/pull/1970#discussion_r1276018108

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done

### DIFF
--- a/packages/shared/src/hooks/useJoinSquad.ts
+++ b/packages/shared/src/hooks/useJoinSquad.ts
@@ -5,8 +5,7 @@ import { useAnalyticsContext } from '../contexts/AnalyticsContext';
 import { Squad } from '../graphql/sources';
 import { AnalyticsEvent } from '../lib/analytics';
 import { useBoot } from './useBoot';
-import { generateQueryKey, RequestKey } from '../lib/query';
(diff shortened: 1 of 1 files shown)


sshanzel (approved):
LGTM! 🚀

PR #1987, opened 2023-07-28
fix: display the read post link on the article page on mobile

DragosIorgulescu:
- uses the `PostModalActions` component which already includes the link and menu actions
- only shown when not rendering in a modal
- adds the `Read Post` and menu actions to the post page header

## Events

- [x] MobileL (420px)
- [x] Tablet (656px)
- [x] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1512 #done


DragosIorgulescu:
the ticket associated with this mentions the `Post modal`, however, this already has the link and actions in the header. The post page however did not, so this is where the changes were made. Let me know if this needs to be applied to the modal as well. See screenshots below:

left: Article view inside modal
right: Article view as standalone page

Screenshot 2023-07-28 at 12 37 52
Screenshot 2023-07-28 at 12 38 11

packages/shared/src/components/post/PostContent.tsx
```diff
@@ -108,7 +108,23 @@ export function PostContent({
...
         />
       )}
-      <PostContainer className={classNames('relative', className?.content)}>
+
+      <PostContainer
+        className={classNames('relative', className?.content)}
+        data-testid="postContainer"
+      >
+        {!hasNavigation && (
+          <PostModalActions
```
rebelchris, line 117: More generic comment, maybe we should consider renaming it, as it's not particular related to modal anymore.

rebelchris: I also see we already load this component in PostNavigation is it maybe not as simple as allowing it to be enabled from there? (I'm not quite sure they share the same placement though)

As technically the modal and page share the same content it should be available already.

rebelchris (commented):
Noticed a small issue if this bar is active (it's a conditional render)
Screenshot 2023-07-28 at 14 53 16

Also see my remarks if we can somehow uniform it in one place?

DragosIorgulescu:
> More generic comment, maybe we should consider renaming it, as it's not particular related to modal anymore.
@rebelchris I completely agree. However, it looks like there is a `PostActions` component in this module already defined. 

Overall it looks like this area could use a little bit of cleanup in naming conventions and component usage. Would it make more sense for me to create a separate PR that takes care of this? Or do you prefer that I just update the `PostModalActions` component to something else here for now?

rebelchris:
@DragosIorgulescu How much cleanup would you propose?
If it's limited to component renaming I would introduce it here.
If it requires more then that, separate indeed.

packages/shared/src/components/post/PostContent.tsx
DragosIorgulescu: @sshanzel 

> Also, kindly check if this shows on post page for squad post where there is another navigation header.

Thanks for pointing this out. This looks good on the squad post page -- as far as I can tell the squad posts actually use a different component entirely to render (the new header is only added to the `PostContent` component
 https://github.com/dailydotdev/apps/blob/06978d27eada1f40d3d826ecfa4cfdb51e40f39e/packages/webapp/pages/posts/%5Bid%5D/index.tsx#L71-L74

Below is a preview as well:

Screenshot 2023-08-01 at 13 49 13

---

> One area we can probably look at as well is the navigation variable in the post page to ensure it would only show to specific post types, in this case, the article only. Would also allow us to avoid using the hasNavigation as it would only be applied on the post page.

There is just one caveats here:
- related to the props `PostModalActions` expects to be passed in. We already have all the variables/functions to pass defined within the `PostContent` component (they are passed to the `PostWidget` component as well). So if we decide to move this further up the tree, we would need to redefine those which would create some duplication as well as mixing up responsibilities a bit. 

If you feel that is an acceptable tradeoff, I can go ahead and change it to use the navigation variable.

DragosIorgulescu:
> @DragosIorgulescu How much cleanup would you propose?
If it's limited to component renaming I would introduce it here.
If it requires more then that, separate indeed.

@rebelchris I'm mostly thinking about the `PostActions` and `PostModalActions` components which seem very similar in their purpose. But I think for now I will just rename `PostModalActions` to `PostHeaderActions`. 

We could consolidate these into one component but they look to be pretty different visually, so I am not convinced that would be a good idea

---

BTW the spacing issue when the banner is enabled should be fixed as well:

Screenshot 2023-08-01 at 13 59 40

rebelchris (approved):
Alright, looks solid to me, let's give it a go 🙏

packages/shared/src/components/post/PostContent.tsx
sshanzel: Good point. We will be pulling a lot of hooks and callbacks just to satisfy having it in the `navigation` variable. Either of the two is okay to me. Would love to hear what others think of it too. cc @rebelchris @capJavert @TomDowling 

Context: we have a variable in the post page file named `navigation` and it holds the value for which kind of navigation to render that is processed within the components inside the `PostContent`. The key difference here is that using the `navigation` would keep all of our navigations in the same place, easier to manage, but it would require pulling a lot of hooks to satisfy the requirement of the actions we can perform on the component to render (PostModalActions) - which on the current implementation does not require any additional hook instances since the callbacks are readily available to where it is currently rendered. There are pros and cons on both ends but would love to hear more thoughts.

rebelchris: Agree with Ante, great point for down the line improvement, but then again the whole render could do with a revamp 👍

sshanzel (approved):
Nice work! Let's go shipping! 🚀

PR #1992, opened 2023-07-28
fix: updated say hi comment default

rebelchris:
- We needed to update the say hi default comment
- As we don't have all the data on currentMember I decided to locally use the authContext

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1581 #done

### DIFF
--- a/packages/shared/src/components/checklist/SquadFirstCommentChecklistStep.tsx
+++ b/packages/shared/src/components/checklist/SquadFirstCommentChecklistStep.tsx
@@ -22,9 +22,11 @@ const SquadFirstCommentChecklistStep = ({
         disabled={!welcomePost}
         onClick={() => {
           router.push(
-            `/posts/${welcomePost.id}?comment=Hi my name is ${
-              user?.name || user.username
-            } happy to be here! I joined ${squad.name} because…`,
+            `/posts/${welcomePost.id}?comment=${encodeURIComponent(
+              `Hi my name is ${
+                user?.name || user.username
+              } happy to be here! I joined ${squad.name} because…`,
+            )}`,
           );
         }}
       >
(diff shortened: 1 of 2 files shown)


packages/shared/src/components/checklist/SquadFirstCommentChecklistStep.tsx
```diff
@@ -19,7 +21,11 @@ const SquadFirstCommentChecklistStep = ({
         className="btn-primary"
         disabled={!welcomePost}
         onClick={() => {
-          router.push(`/posts/${welcomePost.id}?comment=Say hi 👋`);
+          router.push(
+            `/posts/${welcomePost.id}?comment=Hi my name is ${
+              user?.name || user.username
```
capJavert, line 26: Since user can be `undefined` We either need to do optional on both `user` getters or just remove them to expect user is always there. Like this its gonna break either way and its not clear what we want.

capJavert (approved):
Just one consistency issue.

packages/shared/src/components/checklist/SquadFirstCommentChecklistStep.tsx
rebelchris: Anonymous won't have checklist rendered right?

capJavert: Yeah, but if for ever reason if we render checklist for anonymous. Our typescript should be screaming that user could be undefined. Won't block but I think it's right from general logic perspective.

rebelchris: Oh sorry so name is optional it can be empty on a user, but username is not.
Hope that actually answers it better.

We do expect user to be there always.

sshanzel: Since this is going to be passed as part of the query param and I am unsure yet of the allowed special characters for the name, I think we should run it with `encodeURIcomponent`.

sshanzel (approved):
Just one thing to consider but I have nothing blocking :shipit: 🥳

packages/shared/src/components/checklist/SquadFirstCommentChecklistStep.tsx
rebelchris: Ah good point, we can do that for the name alone indeed 🙏
Username shouldn't allow it anyway

sshanzel (approved):
Looks amazing to me! 🚀

PR #1993, opened 2023-07-28
fix: stale squad response after join/leave

capJavert:
- we had an edge case where sometimes after user joining or leaving squad the next Source API request would return stale data with `currentMember` as if user was still part of the squad.
Following flow would fail:
- join squad
- join mutation passes
- we invalidate source data locally (react query)
- source query fires API request for Source to refetch
  - API request returns the stale data (like user is not a member)

I checked that order of requests on the frontend is valid and that join mutation request finishes before the Source request is made. So I don't think it is a frontend race condition.

This PR fixes the above where we actually reuse the response from join squad mutation. As bonus we make one less network request which is even better.

This fixes the issue but we are still not sure why API sometimes returns the stale response.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done

### DIFF
--- a/packages/shared/src/components/squads/SquadJoinButton.tsx
+++ b/packages/shared/src/components/squads/SquadJoinButton.tsx
@@ -109,7 +109,15 @@ export const SquadJoinButton = ({
         displayToast('👋 You have left the Squad.');
 
         const queryKey = generateQueryKey(RequestKey.Squad, user, squad.handle);
-        queryClient.invalidateQueries(queryKey);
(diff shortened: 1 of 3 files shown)


capJavert:
@idoshamun not urgent or blocker but if you can have a look next week if you know why would the issue on API happen. Do we have some kind of short cache maybe?

sshanzel (approved):
Looks good to me! :shipit:

PR #1999, opened 2023-08-01
refactor: merging of finished actions

sshanzel:
## Changes
- UserActions are immutable, we don't delete any data on this entity in any scenario, but since there can be some latency issues with background workers updating the finished actions, we need to keep the manually stored cache to avoid discrepancies with some checklists.
- We have two options, either merge it or disable the `refetchOnMount`. The latter is a good option too, though we will have to manually add to FE cache every actions that are registered in an async manner in the BE.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done

### DIFF
--- a/packages/shared/src/hooks/useActions.ts
+++ b/packages/shared/src/hooks/useActions.ts
@@ -19,35 +19,40 @@ interface UseActions {
 export const useActions = (): UseActions => {
   const client = useQueryClient();
   const { user } = useAuthContext();
+  const actionsKey = generateQueryKey(RequestKey.Actions, user);
   const { data: actions, isFetched: isActionsFetched } = useQuery(
-    generateQueryKey(RequestKey.Actions, user),
-    getUserActions,
-    { enabled: !!user, refetchOnMount: false },
+    actionsKey,
+    async () => {
+      const current = client.getQueryData<Action[]>(actionsKey);
+      const data = await getUserActions();
+
+      if (!current || !Array.isArray(current)) return data;
+
+      const filtered = data.filter(({ type }) =>
+        current.every((action) => action.type !== type),
+      );
+
+      return [...current, ...filtered];
+    },
+    { enabled: !!user },
   );
   const { mutateAsync: completeAction } = useMutation(completeUserAction, {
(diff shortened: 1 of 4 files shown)


packages/shared/src/hooks/useActions.ts
```diff
@@ -19,35 +19,40 @@ interface UseActions {
...
-      );
+        return [...data, optimisticAction];
+      });
 
-      return () =>
-        client.setQueryData<Action[]>(
-          generateQueryKey(RequestKey.Actions, user),
-          actions,
-        );
+      return () => client.setQueryData<Action[]>(actionsKey, actions);
```
sshanzel, line 55: Key change here is just the query key used to be defined as a const now.

packages/shared/src/hooks/useActions.ts
```diff
@@ -19,35 +19,40 @@ interface UseActions {
...
+
+      if (!current || !Array.isArray(current)) return data;
+
+      const filtered = data.filter(({ type }) =>
+        current.every((action) => action.type !== type),
+      );
+
+      return [...current, ...filtered];
+    },
+    { enabled: !!user },
```
sshanzel, line 37: Core difference is bringing back the `refetchOnMount`

capJavert (approved):
Looking good, only thing is I would update `useJoinSquad` to just use `completeAction(ActionType.JoinSquad)`, no need to manually `setQueryData` since `completeAction` does the same logic.

sshanzel:
> Looking good, only thing is I would update useJoinSquad to just use completeAction(ActionType.JoinSquad), no need to manually setQueryData since completeAction does the same logic.

Agreed. It is much cleaner to do that too.

capJavert:
@sshanzel Nice, let's merge 🚀

sshanzel:
> @sshanzel Nice, let's merge 🚀

The test run is failing weirdly. Once it turns green, I will merge asap.

PR #2005, opened 2023-08-01
feat: creating the search bar as well as other components the bar uses

TomDowling:
## Changes
This has the creation of the SearchBar, SearchBarSuggestion and SearchProgressBar components as well as some tests. Note that this has fake functions, tests not complete etc as it has dependencies on other tickets that will be merged to WT-1535-search.

Screenshot 2023-08-01 at 2 30 56 pm

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1552 #done


packages/webapp/components/layouts/MainFeedPage.tsx
```diff
@@ -72,7 +72,8 @@ export default function MainFeedPage({
       isSearchOn={isSearchOn}
       onFeedPageChanged={(page) => router.replace(`/${page}`)}
       searchQuery={router.query?.q?.toString()}
-      searchChildren={<PostsSearch placeholder="Search posts" />}
+      searchChildren={<SearchBar inputId="search" completedTime="12:12" />}
+      className={{search: 'overflow-x-visible'}}
```
rebelchris, line 76: This can be hardcoded right as we don't have overwrites anymore?

packages/webapp/pages/discussed.tsx
```diff
@@ -22,6 +22,6 @@ const Discussed = (): ReactElement => {
 };
 
 Discussed.getLayout = getMainFeedLayout;
-Discussed.layoutProps = mainFeedLayoutProps;
+Discussed.layoutProps = { ...mainFeedLayoutProps, isSearchOn: true };
```
rebelchris, line 25: I think we should just add this to the mainFeedLayoutProps

packages/shared/src/components/search/SearchProgressBar.tsx
```diff
@@ -0,0 +1,44 @@
...
+export interface SearchProgressBarProps {
+  progress: number;
+}
+
+export const SearchProgressBar = ({
+  progress,
+}: SearchProgressBarProps): ReactElement => {
+  return (
+    <div className="flex relative gap-6 h-2" data-testId="SearchProgressBar">
+      <div className="relative flex-1 h-full bg-white rounded-full">
```
rebelchris, line 12: The background is actually different.
(Also the animation, but let's leave that out for another ticket)
Screenshot 2023-08-02 at 12 11 50

packages/shared/src/components/search/SearchBar.tsx
```diff
@@ -0,0 +1,327 @@
...
+            <SearchBarSuggestion
+              key={index}
+              suggestion={suggestion.suggestion}
+              onClick={suggestion.onClick}
+            />
+          ))}
+        </div>
+      )}
+
+      {reportModal && (
```
rebelchris, line 313: Think removing this is neater

packages/shared/src/components/search/SearchBar.tsx
```diff
@@ -0,0 +1,327 @@
...
+                  buttonSize={ButtonSize.Small}
+                  title="Search history"
+                  onClick={seeSearchHistory}
+                  icon={<TimerIcon size={IconSize.Small} />}
+                  disabled={searchHistory.length === 0}
+                />
+              </div>
+            </SimpleTooltip>
+            {!isMobile && (
+              <Button
```
rebelchris, line 267: This should be disabled with tooltip if no query is active.
Screenshot 2023-08-02 at 11 59 31

packages/shared/src/components/search/SearchBar.tsx
```diff
@@ -0,0 +1,327 @@
...
+                {...rightButtonProps}
+                className="btn-primary"
+                buttonSize={ButtonSize.Medium}
```
rebelchris, line 270: Medium is default, no need to explicitly set

packages/shared/src/components/search/SearchBar.tsx
```diff
@@ -0,0 +1,327 @@
...
+    event.stopPropagation();
+    alert('see search history');
+  };
+
+  // TODO: Temp for testing
+  const progressEmulator = () => {
+    setInterval(() => {
+      setProgress((oldProgress) => {
+        if (oldProgress === 100) {
+          // setCompletedTime('12:12');
```
rebelchris, line 166: Comment should be removed/opened

packages/shared/src/components/search/SearchBar.tsx
```diff
@@ -0,0 +1,327 @@
...
+}
+
+export interface SearchBarProgressBarProps {
+  progress: number;
+}
+
+// const ButtonIcon = ({ isPrimary }: { isPrimary: boolean }) =>
+//   isPrimary ? <CloseIcon /> : <ArrowIcon className="rotate-90" />;
+
+export const SearchBar = forwardRef(function SearchBar(
```
rebelchris, line 73: You got quite a few props here that are commented out, can they be removed?

packages/shared/src/components/search/SearchBar.spec.tsx
```diff
@@ -0,0 +1,120 @@
+/*
```
rebelchris, line 1: Test have lots of comments/commented out code, it's not really neat to merge until they are cleaned/removed rather.
If you can't resolve them, add them to a ticket rather.

rebelchris (changes requested):
Amazing, tested, and looks pretty awesome I must say!
Have some cleanup remarks, shouldn't be too crazy.

packages/shared/src/components/search/SearchBar.tsx
```diff
@@ -0,0 +1,327 @@
...
+        <div className="mt-6">
+          <SearchProgressBar progress={progress} />
+
+          {progress > 0 && progress < 100 && (
+            <div className="mt-2 typo-callout text-theme-label-tertiary">
+              🚀 Generating answer
+            </div>
+          )}
+
+          {progress === 100 && completedTime && (
```
sshanzel, line 293: I think this one alone can tell us it is done.
```suggestion
          {completedTime && (
```

sshanzel (changes requested):
Massive work! From mocking alone to writing actual tests, things are looking amazing.

I just had a few comments, and to add to Chris' comments, we should probably remove as much mocked areas as possible to avoid missing it once get close production.

packages/shared/src/components/search/SearchBarSuggestion.tsx
```diff
@@ -0,0 +1,25 @@
...
+  onClick: () => void;
+}
+
+export const SearchBarSuggestion = ({
+  suggestion,
+  onClick,
+}: SearchBarSuggestionProps): ReactElement => {
+  return (
+    <Button
+      className="btn-secondary border-theme-divider-tertiary typo-subhead text-theme-label-tertiary"
```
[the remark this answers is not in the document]
rebelchris: It's not actually a border from design perspective, it's a tag, and for that matter a new one.
So for now this implementation is ok 🙏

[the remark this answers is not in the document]
sshanzel: Oh, cool. Thanks for the clarification, maybe we will have to coordinate it with the design to probably create the right object in Sketch to avoid confusion as the current object tells a `Button`. Let's keep this then for the moment 🚢

packages/shared/src/components/search/SearchBar.tsx
```diff
@@ -0,0 +1,327 @@
...
+    if (!user) {
+      setSuggestions([
+        {
+          suggestion:
+            'Sign up and read your first post to get search recommendations',
+          onClick: () => showLogin('search bar suggestion'),
+        },
+      ]);
+    }
+  }, [user]);
```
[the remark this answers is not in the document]
capJavert: Also if not, I think we can get due without useEffect and separate state for it it could just be conditional return or useMemo depending on `user`.

packages/shared/src/components/search/SearchBar.spec.tsx
```diff
@@ -0,0 +1,120 @@
...
+  });
+
+  it('should render with clear button when there is a value and clear on click', async () => {
+    renderComponent(true, { value: 'search' });
+    const input = screen.queryByRole('textbox') as HTMLInputElement;
+    const clear = screen.queryByTitle('Clear query');
+
+    await waitFor(() => expect(clear).toBeInTheDocument());
+    expect(input).toHaveValue('search');
+    clear.click();
```
capJavert, line 78: More of a general note, prefer `fireEvent` instead of `click` because it triggers events more like it happened in the browser. More [info](https://github.com/dailydotdev/apps/pull/2005/files). We could even use `userEvent` but those are better suited for e2e testing.

packages/shared/src/components/cards/RaisedLabel.module.css
```diff
@@ -0,0 +1,12 @@
...
+  z-index: -1;
+  transition: transform 0.1s linear;
+}
+
+.raiseLabelContainer:hover .flag,
+.flag:hover {
+  transform: none;
+  z-index: 1;
+  transition: transform 0.1s linear, z-index 0.1s step-end;
+}
```
capJavert, line 12: Why can't we do this in tailwind?

capJavert (commented):
The guys already pointed out most of the things, got just minor observations, usage wise it looks great in our web app 🤘

packages/shared/src/components/search/SearchBar.tsx
```diff
@@ -0,0 +1,327 @@
...
+    }, 100);
+  };
+
+  return (
+    <div className="mb-2 w-full max-w-2xl">
+      <div className={classNames('relative', moreStyles.raiseLabelContainer)}>
+        <BaseField
+          {...props}
+          className={classNames(
+            'relative items-center px-3 h-16 rounded-14 border !border-theme-divider-tertiary !bg-theme-bg-primary',
```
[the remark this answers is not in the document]
TomDowling: they are needed as the styles set from the BaseField override these otherwise

TomDowling:
@dailydotdev/web-team, This is ready for testing again. Note that if you view the branch it won't work anymore as i removed demo code

packages/shared/src/components/search/SearchBar.tsx
```diff
@@ -0,0 +1,269 @@
...
+                </div>
+              </SimpleTooltip>
+            )}
+          </div>
+        </BaseField>
+
+        <RaisedLabel type={RaisedLabelType.Beta} />
+      </RaisedLabelContainer>
+
+      {sidebarRendered && showProgress && (
```
rebelchris, line 238: Not sure all these sidebarRendered evaluations are correct actually
Don't we also show this on mobile? (Same question for multiple items here)

packages/shared/src/components/search/SearchProgressBar.tsx
rebelchris: @TomDowling I still see bg-white here, not sure it's fixed?

packages/shared/src/components/search/SearchBar.tsx
```diff
@@ -0,0 +1,269 @@
...
+                  className="btn-tertiary"
+                  buttonSize={ButtonSize.Small}
+                  title="Search history"
+                  onClick={seeSearchHistory}
+                  icon={<TimerIcon />}
+                  disabled={searchHistory.length === 0}
+                />
+              </div>
+            </SimpleTooltip>
+            {sidebarRendered && (
```
rebelchris, line 216: This should not be conditional right?

rebelchris (changes requested):
Looks great, just some minor comments on mobile layout

packages/shared/src/components/search/SearchBar.spec.tsx
```diff
@@ -0,0 +1,126 @@
...
+import { generateTestSquad } from '../../../__tests__/fixture/squads';
+import { SearchBar, SearchBarProps } from './SearchBar';
+
+beforeEach(async () => {
+  nock.cleanAll();
+  jest.clearAllMocks();
+});
+
+const squads = [generateTestSquad()];
+// TODO: How do i render sidebar in tests?
```
rebelchris, line 22: Should be removed :)

rebelchris (approved):
One small remark, once removed it's good for me 🙏

packages/shared/src/components/cards/RaisedLabel.tsx
```diff
@@ -1,41 +1,51 @@
...
 export interface RaisedLabelProps {
   listMode?: boolean;
   type: RaisedLabelType;
   description?: string;
+  className?: string | undefined;
 }
 
+export const RaisedLabelContainer = classed(
+  'div',
+  'temp relative group group-hover:transform-none group-hover:z-10 group-hover:transition-transform group-hover:ease-linear group-hover:transition-z-index group-hover:duration-100 group-hover:step-end',
```
sshanzel, line 27: Is it possible to remove the `temp` now?
```suggestion
  'relative group group-hover:transform-none group-hover:z-10 group-hover:transition-transform group-hover:ease-linear group-hover:transition-z-index group-hover:duration-100 group-hover:step-end',
```

packages/shared/src/components/cards/RaisedLabel.tsx
```diff
@@ -1,41 +1,51 @@
...
         listMode ? 'top-0 right-full mt-5' : 'left-0 bottom-full ml-5 h-5',
+        className,
       )}
     >
       <SimpleTooltip content={description}>
         <div
           className={classNames(
             'flex items-center px-1',
-            styles.flag,
+            'will-change-auto -z-1 transition-transform duration-100 ease-linear group-hover:transform-none group-hover:z-10 group-hover:transition-transform group-hover:ease-linear group-hover:transition-z-index group-hover:duration-100 group-hover:step-end',
```
sshanzel, line 48: [non-actionable] Wow, we were able to translate it but it did become a huge chunk of classes. I wonder how we can name this better. Though, still the goal is to lessen the bundle by reducing CSS modules that bloat the app. Well done 🚢

sshanzel (approved):
Love the search cleanup! One less legacy flow 🚀

packages/shared/src/components/cards/RaisedLabel.tsx
TomDowling: Great spot, fixing now

PR #2011, opened 2023-08-02
feat: adding border to google buttons on light mode

TomDowling:
## Changes
Adding a border so the google button is more visible on light mode. Screenshots added below

dark-mode
light-mode

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1560 #done

### DIFF
--- a/packages/shared/src/components/auth/AuthDefault.tsx
+++ b/packages/shared/src/components/auth/AuthDefault.tsx
@@ -155,17 +155,18 @@ const AuthDefault = ({
     <>
       {!simplified && <AuthModalHeader title={title} />}
       <AuthContainer className={disableRegistration && 'mb-6'}>
-        {providers.map(({ provider, ...props }) => (
-          <ProviderButton
-            key={provider}
-            provider={provider}
-            label={shouldLogin ? 'Log in with' : 'Sign up with'}
-            className="mb-1 border-0"
-            onClick={() => onSocialClick(provider.toLowerCase())}
-            loading={!isReady}
-            {...props}
-          />
-        ))}
+        <div className="flex flex-col gap-4">
+          {providers.map(({ provider, ...props }) => (
+            <ProviderButton
+              key={provider}
+              provider={provider}
+              label={shouldLogin ? 'Log in with' : 'Sign up with'}
+              onClick={() => onSocialClick(provider.toLowerCase())}
+              loading={!isReady}
+              {...props}
+            />
+          ))}
+        </div>
         {getOrDivider()}
         {getForm()}
       </AuthContainer>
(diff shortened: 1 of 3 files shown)


packages/shared/src/components/auth/AuthDefault.tsx
```diff
@@ -160,7 +160,7 @@ const AuthDefault = ({
             key={provider}
             provider={provider}
             label={shouldLogin ? 'Log in with' : 'Sign up with'}
-            className="mb-1"
+            className="mb-1 border-0"
```
rebelchris, line 163: Does this not overwrite the className in the providers map though? 👀

rebelchris (commented):
Small concern, just not sure on how it renders all together now.

packages/shared/src/components/auth/AuthDefault.tsx
TomDowling: Yes it does i was leaving this pages setup as it was, but i could remove this and deal with the spacing (margin) with gap or space-y which would be better?

rebelchris: I mean more for this it won't take the border color then right? 
As we never merge the -1 border now.

packages/shared/src/components/auth/ProviderButton.tsx
```diff
@@ -24,7 +24,6 @@ function ProviderButton({
       style={{
         justifyContent: label ? 'flex-start' : 'center',
         color: '#FFFFFF',
-        border: 0,
```
sshanzel, line 27: I think the reason we had this as an inline style was to override the button's base class for border. Though if the `className` can do that no issues, then we can move forward with this.

packages/shared/src/components/auth/common.tsx
```diff
@@ -41,6 +41,7 @@ export const providerMap: ProviderMap = {
     icon: <GoogleIcon className="socialIcon" secondary />,
     provider: 'Google',
     style: { backgroundColor: '#FFFFFF', color: '#0E1217' },
+    className: 'border border-theme-divider-tertiary',
```
sshanzel, line 44: Is this expected for both light and dark?

TomDowling: From the slack thread i read it as they wanted this color (divider-teritary) for both

rebelchris: Yep, should be fine indeed, it sorts light/dark itself

packages/shared/src/components/auth/AuthDefault.tsx
TomDowling: This doesn't get used on the google button as we are passing classNames in the providers map, so overrides this classNames to the one supplied in the map as its passed in { ...props } at the bottom of the props. I have confirmed it works currently as can be seen in the screenshots above :)

TomDowling: I have refactored this now so will resubmit for review in a few

rebelchris (approved):
Great!

Let's go 🎉

PR #2014, opened 2023-08-03
feat: change to the use notifications logic to show subscribe banner

TomDowling:
## Changes
Changed the logic to show the subscribe banner in squads as the banner was showing even if you were subscribed

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1513 #done

### DIFF
--- a/packages/shared/src/hooks/useEnableNotification.ts
+++ b/packages/shared/src/hooks/useEnableNotification.ts
@@ -72,7 +72,6 @@ export const useEnableNotification = ({
     !isInitialized,
     !isNotificationSupported,
     (isSubscribed || hasPermissionCache) && !isEnabled,
-    hasEnabled,
   ];
   const shouldShowCta = !conditions.some((passed) => passed);
 
--- a/packages/shared/src/hooks/useEnableNotification.ts
+++ b/packages/shared/src/hooks/useEnableNotification.ts
@@ -72,7 +72,7 @@ export const useEnableNotification = ({
     !isInitialized,
     !isNotificationSupported,
     (isSubscribed || hasPermissionCache) && !isEnabled,
-    hasEnabled && source === NotificationPromptSource.SquadPostModal,
+    hasEnabled,
   ];
   const shouldShowCta = !conditions.some((passed) => passed);
 


packages/shared/src/hooks/useEnableNotification.ts
```diff
@@ -72,7 +72,7 @@ export const useEnableNotification = ({
     !isInitialized,
     !isNotificationSupported,
     (isSubscribed || hasPermissionCache) && !isEnabled,
-    hasEnabled && source === NotificationPromptSource.SquadPostModal,
```
sshanzel, line 75: Not sure how this affects the Squads page. It should have a different `source` value.

rebelchris: Oh wait did we introduce this so it only shows on squad post modal, not normal post modals?
If so, the solution is to split the two!

Basically now we had the exclusion where it could ALWAYS show on squad postr modal, even if you enabled already.

sshanzel: I think this is the other way around, the logic is negated down below from this code. So it would always show except for that source.

rebelchris: Maybe you can have a check with Tom in that case.
The issue is that notifications sometimes show even if you already accepted.

I think it has mainly to do with this logic, but you might be able to debug it better as you know it a bit better in detail.

TomDowling: @sshanzel After investigating, do you suggest we leave this as it is currently or do this change?

rebelchris: @sshanzel i think it was also a team member who checked and 100% already gave permissions.
To me this particular code did look like it would be able to fire even when you already have them enabled wouldnt you think?

sshanzel: Hmm. After carefully evaluating each variable, it seems it should only display as necessary. Can you specify which ones that raise doubt so we can investigate it further?

rebelchris (approved):
Let's give it a try, but also retest on production 🙏

sshanzel:
Wait. I remember now.

sshanzel:
The reason we have that condition is that, normally, when you enable it through the banner, we have a post-message that tells you some bunch of stuff (like go to notifications page to disable it, etc.) but on `SquadPostModal`, there is no post-message since it is just a strip on top of the modal itself. So we will remove it immediately once accepted.

sshanzel:
The reason we have that condition is that, normally, when you enable it through the banner, we have a post-message that tells you some bunch of stuff (like go to notifications page to disable it, etc.) but on `SquadPostModal`, there is no post-message since it is just a strip on top of the modal itself. So we will remove it immediately once accepted.

sshanzel:
The reason we have that condition is that, normally, when you enable it through the banner, we have a post-message that tells you some bunch of stuff (like go to notifications page to disable it, etc.) but on `SquadPostModal`, there is no post-message since it is just a strip on top of the modal itself. So we will remove it immediately once accepted.

sshanzel:
The reason we have that condition is that, normally, when you enable it through the banner, we have a post-message that tells you some bunch of stuff (like go to notifications page to disable it, etc.) but on `SquadPostModal`, there is no post-message since it is just a strip on top of the modal itself. So we will remove it immediately once accepted.

sshanzel:
> @sshanzel, How do you suggest we progress with this ticket?

We might have to further investigate any other symptoms. I think we can start with the issue we encountered on your end. Can you check if there is any log that leads to the concern? If there is no useful message, kindly check if you will be able to reproduce it locally.

To make the local env work similar to prod, enable your experimental features to manually add insecure origin as secure. On Chrome:
1. Visit chrome://flags/
2. Find "Insecure origins treated as secure".
3. Switch to Enabled.
4. Set the values to `http://webapp.local.com:5002,http://local.com:5002`

TomDowling:
Closing as we have been unable to diagnose this enough to release a fix

sshanzel:
> Closing as we have been unable to diagnose this enough to release a fix

Agreed. Sab is one of the key users we're trying to get more information from. So we can probably watch out for his messages on the issues channel.

PR #2016, opened 2023-08-03
refactor: show no content while loading actions

sshanzel:
## Changes
- Wait until the actions data is loaded before fetching.
- Might introduce some time difference since we now wait for the actions to be loaded first.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1614 #done

### DIFF
--- a/packages/webapp/pages/squads/index.tsx
+++ b/packages/webapp/pages/squads/index.tsx
@@ -24,6 +24,8 @@ import { ApiError } from 'next/dist/server/api-utils';
 import { oneHour } from '@dailydotdev/shared/src/lib/dateFormat';
 import { Connection } from '@dailydotdev/shared/src/graphql/common';
 import { Squad } from '@dailydotdev/shared/src/graphql/sources';
+import { useActions } from '@dailydotdev/shared/src/hooks/useActions';
+import { ActionType } from '@dailydotdev/shared/src/graphql/actions';
 import { mainFeedLayoutProps } from '../../components/layouts/MainFeedPage';
 import FeedLayout, { getLayout } from '../../components/layouts/FeedLayout';
 import { defaultOpenGraph, defaultSeo } from '../../next-seo';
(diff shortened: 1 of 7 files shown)


packages/webapp/pages/squads/index.tsx
```diff
@@ -60,6 +64,8 @@ const SquadsPage = ({ initialData }: Props): ReactElement => {
     },
   );
 
+  if (!isAuthReady || (user && !isActionsFetched)) return null;
```
sshanzel, line 67: The `isActionsFetched` is always `false` when it is for anonymous user as the query fails.

rebelchris (approved):
Looks ok to me, but maybe ask @idoshamun what he thinks of the preview link you put above here.

packages/webapp/pages/squads/index.tsx
idoshamun: I think it's a bad practice to block the rendering altogether as it will interrupt the seo aspects of the page. I suggest conditionally rendering the relevant parts

rebelchris: Makes sense 😊

sshanzel:
@rebelchris I have instead used the `visibility` CSS property so there won't be any shift with the height of the header since the actual element is considered even when `invisible`. Let me know if the new behavior works.

https://daily-webapp-3vkn3gqn8-dailydotdev.vercel.app/squads

rebelchris (approved):
Looks good to me 👍

sshanzel:
@idoshamun we now render everything except the button but use the invisible property to avoid the header shifting its height once finished. Let us know what you think with the new behavior 🚢 

https://daily-webapp-3vkn3gqn8-dailydotdev.vercel.app/squads

idoshamun (approved):
Perfect

PR #2019, opened 2023-08-03
feat: search response streaming

capJavert:
- Render markdown from streamed search response (`RenderMarkdown` component)
- add typing styles and animation
- add streamed response handling with `useChat` (currently mock until we have API/EventSource ready)
  - `useChat` is modeled by https://sdk.vercel.ai/docs/api-reference/use-chat as much as we could so we can potentially in the future adapt more general purpose LLMs
  - all TODOs are inside this hook and can be merged with this PR and then replaced with implementation when backend is ready
- connected to search with existing components
  - you can try dummy search with mock to see text streaming
- based on my [POC](https://codesandbox.io/s/stream-chat-gpt-demo-wcrw3m)

https://github.com/dailydotdev/apps/assets/9803078/176d779a-886a-4928-8291-a2cfacf49166

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done


packages/shared/jest.config.js
```diff
@@ -12,5 +12,6 @@ module.exports = {
   moduleNameMapper: {
     '\\.svg$': '<rootDir>/__mocks__/svgrMock.ts',
     '\\.css$': 'identity-obj-proxy',
+    'react-markdown': '<rootDir>/__mocks__/reactMarkdownMock.tsx',
```
capJavert, line 15: Needed because `react-markdown` is ESM module

packages/shared/src/components/RenderMarkdown.tsx
```diff
@@ -0,0 +1,167 @@
...
+          className: codeClassName,
+          children,
+          style,
+          ...props
+        }) {
+          const match = /language-(\w+)/.exec(codeClassName || '');
+          const language = match?.[1];
+
+          if (language) {
+            loadLanguages();
```
capJavert, line 137: We are using async syntax highlighter to save on bundle size. When first code block is encountered we trigger languages chunk load. Those are then added to highlighter as they are loaded. Currently loading all languages but if we wanna support only subset we can hardcode specific ones.

packages/shared/src/components/RenderMarkdown.tsx
```diff
@@ -0,0 +1,167 @@
...
+    loadPlugins().then((loadedPlugins) => {
+      if (mounted) {
+        setPlugins(loadedPlugins);
+      }
+    });
+
+    return () => {
+      mounted = false;
+    };
+  }, []);
```
capJavert, line 117: `react-markdown` can use plugins so these are also loaded async to save on bundle size.

packages/shared/src/components/RenderMarkdown.tsx
```diff
@@ -0,0 +1,167 @@
...
+  const loadPlugin = async (loadFn: () => Promise<unknown>) => {
+    return loadFn();
+  };
+
+  try {
+    if (!pluginsLoad) {
+      pluginsLoad = Promise.all([
+        loadPlugin(async () => {
+          const importedModule = await import(
+            /* webpackChunkName: "remarkGfm" */ 'remark-gfm'
```
capJavert, line 77: used to enable raw links and few other markdown styles our comments support.

packages/shared/src/components/search/SearchResult.tsx
```diff
@@ -5,15 +5,24 @@ import LogoIcon from '../../svg/LogoIcon';
...
+  searchMessageProps,
+}: SearchResultProps): ReactElement => (
+  <main className="order-3 laptop:order-3 col-span-2 px-4 laptop:px-8 mb-5">
     <WidgetContainer className="flex p-4">
       <div className="flex p-2 mr-4 w-10 h-10 rounded-10 bg-theme-color-cabbage">
         <LogoIcon className="max-w-full" />
       </div>
-      <div>
-        <p>Result here</p>
+      <div className="w-[calc(100%-3.5rem)]">
```
capJavert, line 24: @rebelchris this is the fix for width I mentioned on slack.

packages/shared/src/components/RenderMarkdown.tsx
```diff
@@ -0,0 +1,167 @@
...
+          const language = match?.[1];
+
+          if (language) {
+            loadLanguages();
+          }
+
+          return !inline && language ? (
+            <SyntaxHighlighterAsync
+              {...props}
+              customStyle={containerReset}
```
rebelchris, line 143: Was wondering can't we do this with tailwind classes?

packages/shared/src/hooks/useChat.ts
```diff
@@ -0,0 +1,353 @@
...
+    // eslint-disable-next-line @typescript-eslint/no-use-before-define
+    const endStream = mockChatStream({ onMessage });
+
+    return () => {
+      endStream();
+    };
+  }, [prompt]);
+
+  return {
+    ...state,
```
rebelchris, line 141: Nice we can use the state to indicate which progress level to show 🙏

packages/shared/src/hooks/useChat.ts
```diff
@@ -0,0 +1,353 @@
...
+        break;
+      }
+
+      onMessage({
+        data: JSON.stringify(stream.pop()),
+      });
+    }
+  };
+
+  mockStream();
```
rebelchris, line 348: We just need to ensure we maybe add a ticket for cleanup once we have implementation details

packages/webapp/__mocks__/reactMarkdownMock.tsx
```diff
@@ -0,0 +1,11 @@
+import React, { ReactElement, ReactNode } from 'react';
```
rebelchris, line 1: Why do we have 2 of these exact ones?

packages/shared/tsconfig.eslint.json
```diff
@@ -10,6 +10,7 @@
     "**/.eslintrc.js",
     "__tests__/**/*.ts",
     "__tests__/**/*.tsx",
-    "__mocks__/**/*.ts"
+    "__mocks__/**/*.ts",
+    "__mocks__/**/*.tsx"
```
rebelchris, line 14: I've seen this in shared and extension, don't they share the shared ones?

rebelchris (approved):
Looks great, nothing breaking.
Just some minor questions.

packages/shared/tsconfig.eslint.json
capJavert: From what I can see no, they extend the root one, and even if they did tsconfig is not deeply extendable it will always overwrite top level properties of extended config.

packages/webapp/__mocks__/reactMarkdownMock.tsx
capJavert: Just a mock for each app. It is also done for svgr for example.

packages/shared/src/hooks/useChat.ts
capJavert: It should be the `EventSource` implementation as mentioned in the DR so we can add it to that task.

packages/shared/src/components/RenderMarkdown.tsx
capJavert: @rebelchris I can use `PreTag` prop of syntax highlighter but I would have to use `!important` on tailwind classes because some default styles are applied through `style` by default.

capJavert: In my opinion current is cleaner.

rebelchris: Agreed :)

rebelchris (approved):
Looks good!

packages/shared/src/hooks/useChat.ts
capJavert: Linked in this task https://dailydotdev.atlassian.net/browse/WT-1626

PR #2027, opened 2023-08-08
feat: add ssr seo to squad page

capJavert:

- [x] - Add seo data fetch for public squads
- [x] - Private squads query will just fail since private squads need user auth and will behave as they do currently
- [x] - Add custom seo if squad page is accessed from open squad invitation

Page is not `getServerSideProps` page. Caching is the same as `getStaticProps` only through cache header. This was needed to take into account query params `userid` and `cid` when squad page is shared as invitation. 

regular:
image

invite:
image

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done

### DIFF
--- a/packages/webapp/next-seo.ts
+++ b/packages/webapp/next-seo.ts
@@ -1,3 +1,5 @@
+import { Source } from '@dailydotdev/shared/src/graphql/sources';
+import { cloudinary } from '@dailydotdev/shared/src/lib/image';
 import { DefaultSeoProps, NextSeoProps, OpenGraph } from 'next-seo/lib/types';
 
 const config: DefaultSeoProps = {
@@ -25,3 +27,14 @@ export const defaultOpenGraph: Partial<OpenGraph> = {
     },
   ],
 };
+
+export const getSquadOpenGraph = ({
+  squad,
+}: {
+  squad?: Pick<Source, 'image'>;
+}): Partial<OpenGraph> => ({
+  images:
+    squad?.image && squad.image !== cloudinary.squads.imageFallback
+      ? [{ url: squad.image }]
+      : defaultOpenGraph.images,
+});
--- a/packages/webapp/pages/squads/[handle]/index.tsx
+++ b/packages/webapp/pages/squads/[handle]/index.tsx
@@ -18,7 +18,10 @@ import { SOURCE_FEED_QUERY } from '@dailydotdev/shared/src/graphql/feed';
 import AuthContext from '@dailydotdev/shared/src/contexts/AuthContext';
 import { SquadPageHeader } from '@dailydotdev/shared/src/components/squads/SquadPageHeader';
 import { BaseFeedPage } from '@dailydotdev/shared/src/components/utilities';
-import { getSquadMembers } from '@dailydotdev/shared/src/graphql/squads';
+import {
+  SQUAD_QUERY,
+  getSquadMembers,
+} from '@dailydotdev/shared/src/graphql/squads';
 import { SourceMember, Squad } from '@dailydotdev/shared/src/graphql/sources';
 import Unauthorized from '@dailydotdev/shared/src/components/errors/Unauthorized';
 import SquadLoading from '@dailydotdev/shared/src/components/errors/SquadLoading';
@@ -30,6 +33,10 @@ import useSidebarRendered from '@dailydotdev/shared/src/hooks/useSidebarRendered
 import classNames from 'classnames';
 import { supportedTypesForPrivateSources } from '@dailydotdev/shared/src/graphql/posts';
 import { useJoinReferral, useSquad } from '@dailydotdev/shared/src/hooks';
+import { oneHour } from '@dailydotdev/shared/src/lib/dateFormat';
+import request, { ClientError } from 'graphql-request';
+import { graphqlUrl } from '@dailydotdev/shared/src/lib/config';
+import { ApiError } from '@dailydotdev/shared/src/graphql/common';
 import { mainFeedLayoutProps } from '../../../components/layouts/MainFeedPage';
 import { getLayout } from '../../../components/layouts/FeedLayout';
 import ProtectedPage, {
@@ -54,7 +61,7 @@ const SquadChecklistCard = dynamic(
     ),
 );
 
-type SourcePageProps = { handle: string };
+type SourcePageProps = { handle: string; initialData?: Squad };
 
 const PageComponent = (props: ProtectedPageProps & { squad: Squad }) => {
   const { squad, seo, children, ...restProtectedPageProps } = props;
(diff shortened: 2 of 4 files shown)


packages/webapp/pages/squads/[handle]/index.tsx
```diff
@@ -176,20 +209,73 @@ SquadPage.layoutProps = mainFeedLayoutProps;
...
+}: GetServerSidePropsContext<SquadPageParams>): Promise<
+  GetServerSidePropsResult<SourcePageProps>
+> {
+  const { handle } = params;
+  const { userid: userId, cid: campaign } = query;
+
+  const setCacheHeader = () => {
+    res.setHeader(
+      'Cache-Control',
+      `public, max-age=0, must-revalidate, s-maxage=${oneHour}`,
```
Comment (capJavert) [inline packages/webapp/pages/squads/[handle]/index.tsx:229]: I would even put this to one day, since squad metadata does not change as often.

Comment (sshanzel) [reply packages/webapp/pages/squads/[handle]/index.tsx:229]: Makes sense. Longer but reasonable the better. Is there a way though to manually invalidate the cache?

packages/webapp/pages/squads/[handle]/index.tsx
```diff
@@ -123,18 +134,40 @@ const SquadPage = ({ handle }: SourcePageProps): ReactElement => {
...
+  const seo = !!seoData && (
+    <NextSeo
+      title={
+        referringUser
+          ? `${referringUser.name} invited you to ${seoData.name}`
+          : `${seoData.name} posts on daily.dev`
+      }
+      description={seoData.description}
+      openGraph={{
+        images: seoData.image ? [{ url: seoData.image }] : undefined,
```
Comment (sshanzel) [inline packages/webapp/pages/squads/[handle]/index.tsx:147]: Should we maybe have a fallback image for this?

packages/webapp/pages/squads/[handle]/index.tsx
```diff
@@ -123,18 +134,40 @@ const SquadPage = ({ handle }: SourcePageProps): ReactElement => {
...
 
-  if ((isFallback || isLoading) && !isFetched) return <SquadLoading />;
+  const seoData = squad || initialData;
+  const seo = !!seoData && (
+    <NextSeo
+      title={
+        referringUser
+          ? `${referringUser.name} invited you to ${seoData.name}`
+          : `${seoData.name} posts on daily.dev`
+      }
```
Comment (sshanzel) [inline packages/webapp/pages/squads/[handle]/index.tsx:144]: We follow a specific template for titles. Have we confirmed this with the product team?

sshanzel (changes requested):
Just had a few comments but great work adding the proper SEO details 🚢

packages/webapp/pages/squads/[handle]/index.tsx
Comment (rebelchris) [reply packages/webapp/pages/squads/[handle]/index.tsx:144]: Good point, probably good to keep it inline with the token page perhaps?
We also have something like `templatedTitle` that does the end part for you.

packages/webapp/pages/squads/[handle]/index.tsx
Comment (rebelchris) [reply packages/webapp/pages/squads/[handle]/index.tsx:147]: Also perhaps in place for token page (not sure)

packages/webapp/pages/squads/[handle]/index.tsx
Comment (rebelchris) [reply packages/webapp/pages/squads/[handle]/index.tsx:229]: I personally think one hour is fine, in case we do decide to make drastic changes it can be easier.
An hour is also pretty good considering what the page does.

packages/webapp/pages/squads/[handle]/index.tsx
```diff
@@ -176,20 +209,73 @@ SquadPage.layoutProps = mainFeedLayoutProps;
...
+    res.setHeader(
+      'Cache-Control',
+      `public, max-age=0, must-revalidate, s-maxage=${oneHour}`,
+    );
   };
+
+  try {
+    const promises = [];
+
+    promises.push(
```
Comment (rebelchris) [inline packages/webapp/pages/squads/[handle]/index.tsx:236]: Why don't you set this one as the default already? (Seeing it's non conditional)

rebelchris (changes requested):
Looks good, some minor additions to Lee's comment, but overal great.

packages/webapp/pages/squads/[handle]/index.tsx
Comment (capJavert) [reply packages/webapp/pages/squads/[handle]/index.tsx:229]: Ok, let's leave it.

> Is there a way though to manually invalidate the cache?

Same as for static, redeploy or clearing through data cache interface https://vercel.com/dailydotdev/daily-webapp/settings/data-cache

packages/webapp/pages/squads/[handle]/index.tsx
Comment (capJavert) [reply packages/webapp/pages/squads/[handle]/index.tsx:144]: @rebelchris @sshanzel It is in line with token page.

packages/webapp/pages/squads/[handle]/index.tsx
Comment (capJavert) [reply packages/webapp/pages/squads/[handle]/index.tsx:236]: Just our typescript being funny. It then infers the type to always be Promises of result of squad query. 😆

packages/webapp/pages/squads/[handle]/index.tsx
Comment (capJavert) [reply packages/webapp/pages/squads/[handle]/index.tsx:144]: Btw. looking at `getTemplatedTitle` why do we add `| daily.dev` to `titleTemplate` in `DefaultSeo` in `_app`? That would apply to all in one go.

packages/webapp/pages/squads/[handle]/index.tsx
Comment (rebelchris) [reply packages/webapp/pages/squads/[handle]/index.tsx:236]: Ah sigh haha

packages/webapp/pages/squads/[handle]/index.tsx
Comment (rebelchris) [reply packages/webapp/pages/squads/[handle]/index.tsx:144]: Hmm, not sure that's correct, maybe present the current ones to Sab and ask they still good.

As for templated, there are exclusions where we want it differently :)

Comment (capJavert) [reply packages/webapp/pages/squads/[handle]/index.tsx:144]: Confirmed with Daphna and Product that current titles are good. 👌

packages/webapp/pages/squads/[handle]/index.tsx
Comment (capJavert) [reply packages/webapp/pages/squads/[handle]/index.tsx]: Product decided we should use default image as fallback for now. Added it here for squad page and token page 1091a6b05797524dc6b49245d072bdad06fcfbdf

packages/webapp/next-seo.ts
```diff
@@ -25,3 +27,14 @@ export const defaultOpenGraph: Partial<OpenGraph> = {
...
   ],
 };
+
+export const getSquadOpenGraph = ({
+  squad,
+}: {
+  squad?: Pick<Source, 'image'>;
+}): Partial<OpenGraph> => ({
+  images:
+    squad?.image && squad.image !== cloudinary.squads.imageFallback
```
capJavert, line 37: Since squad image is automatically set to placeholder during creation we actually did not need fallback for squad image.

This check is to make sure if placeholder is set that we show our default SEO image as requested by Product.

packages/webapp/pages/squads/[handle]/index.tsx
Comment (sshanzel) [reply packages/webapp/pages/squads/[handle]/index.tsx:145]: I think we have it in DefaultSeo, can't remember exactly, but it is sometimes not working as intended.

sshanzel (approved):
Nice work! 🚀

rebelchris (approved):
Alright that should do! 🙏

idoshamun:
@capJavert feel free to merge without my approval even, not sure I'll get the time today and don't want to block

PR #2029, opened 2023-08-09
feat: checking if toast isAnimating before starting again

TomDowling:
## Changes
This stops the toasts timer resetting when scrolling or changing page. However, it does seem to have a performance issue as you scroll it almost freezes the timer as you scroll then continues. I have tried adapting the setInterval, using a useEffect and setTimeout instead all none of them improve this. If anyone has an ideas please do say otherwise agreed with Chris this was okay to progress as is

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1521 #done

### DIFF
--- a/packages/shared/src/components/notifications/Toast.tsx
+++ b/packages/shared/src/components/notifications/Toast.tsx
@@ -42,7 +42,7 @@ const Toast = ({
     {
       enabled: false,
       onSuccess: async (data) => {
-        if (!data) {
+        if (!data || isAnimating) {
           return;
         }
 
--- a/packages/shared/src/components/notifications/Toast.tsx
+++ b/packages/shared/src/components/notifications/Toast.tsx
@@ -32,10 +32,11 @@ const Toast = ({
   const router = useRouter();
   const client = useQueryClient();
   const testRef = useRef(null);
-  const { isAnimating, endAnimation, startAnimation } = useTimedAnimation({
-    autoEndAnimation: autoDismissNotifications,
-    onAnimationEnd: () => client.setQueryData(TOAST_NOTIF_KEY, null),
-  });
+  const { timer, isAnimating, endAnimation, startAnimation } =
+    useTimedAnimation({
+      autoEndAnimation: autoDismissNotifications,
+      onAnimationEnd: () => client.setQueryData(TOAST_NOTIF_KEY, null),
+    });
   const { data: toast } = useQuery<ToastNotification>(
     TOAST_NOTIF_KEY,
     () => client.getQueryData(TOAST_NOTIF_KEY),
--- a/packages/shared/src/components/notifications/Toast.tsx
+++ b/packages/shared/src/components/notifications/Toast.tsx
@@ -1,6 +1,7 @@
 import React, { ReactElement, useEffect } from 'react';
 import { useRouter } from 'next/router';
 import { useQuery, useQueryClient } from 'react-query';
+import classNames from 'classnames';
 import {
   ToastNotification,
   TOAST_NOTIF_KEY,
@@ -10,12 +11,7 @@ import { Button, ButtonSize } from '../buttons/Button';
 import styles from './Toast.module.css';
 import XIcon from '../icons/MiniClose';
 import { isTouchDevice } from '../../lib/tooltip';
-import {
-  NotifContainer,
-  NotifContent,
-  NotifMessage,
-  NotifProgress,
-} from './utils';
+import { NotifContainer, NotifContent, NotifMessage } from './utils';
 import { useTimedAnimation } from '../../hooks/useTimedAnimation';
 import { nextTick } from '../../lib/func';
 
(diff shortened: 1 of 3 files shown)


packages/shared/src/components/notifications/Toast.tsx
```diff
@@ -115,7 +107,17 @@ const Toast = ({
           aria-label="Dismiss toast notification"
         />
         {autoDismissNotifications && (
-          <Progress style={{ width: `${progress}%` }} />
+          <div className="flex absolute -bottom-2 left-0 justify-center w-full ease-in-out">
+            <div
+              className={classNames(
+                'h-1 bg-theme-status-cabbage rounded-full transition-all ease-linear',
```
rebelchris, line 113: I would limit to transition only the width, it can have some negative impacts down the line if we change other variables otherwise.

packages/shared/src/components/notifications/Toast.tsx
```diff
@@ -115,7 +107,17 @@ const Toast = ({
...
+                isAnimating ? 'w-0' : 'w-full',
+              )}
+              style={{
+                transitionDuration: `${toast.timer}ms`,
+              }}
```
rebelchris, line 118: Just wondering if this is not available as a tailwind class, where we could limit the timing maybe?

rebelchris (approved):
Looks neat, small feedback on top of the item that was mentioned on slack about restarting

TomDowling:
@dailydotdev/web-team, this has been updated note that it is now using the JS instead of CSS. A ticket has been added to the backlog to pick up the conversion to CSS, to optimise the Toast implementation

packages/shared/src/components/notifications/Toast.tsx
```diff
@@ -31,6 +31,7 @@ const Toast = ({
 }: ToastProps): ReactElement => {
   const router = useRouter();
   const client = useQueryClient();
+  const testRef = useRef(null);
```
rebelchris, line 34: Can we please name it better 🙏

packages/shared/src/components/notifications/Toast.tsx
```diff
@@ -47,11 +48,19 @@ const Toast = ({
...
   );
 
+  if (!testRef.current && toast?.message) {
+    testRef.current = toast;
+    startAnimation(toast.timer);
+  } else if (testRef.current && testRef.current !== toast && toast?.message) {
+    endAnimation();
+    testRef.current = toast;
+    startAnimation(toast.timer);
+  }
```
rebelchris, line 62:
```suggestion
  if(toast?.message) {
   if(testRef.current && testRef.current !== toast && toast?.message) {
   endAnimation();
   }
   
    testRef.current = toast;
    startAnimation(toast.timer);
  }
```

packages/shared/src/components/notifications/Toast.tsx
```diff
@@ -47,11 +48,19 @@ const Toast = ({
         }
 
         await nextTick(); // wait (1ms) for the component to render so animation can be seen
```
rebelchris, line 50: We don't need this whole onSuccess anymore right?

packages/shared/tailwind.config.js
```diff
@@ -217,6 +217,12 @@ module.exports = {
         logo: '1.125rem',
         'logo-big': '2.0625rem',
       },
+      transitionProperty: {
+        width: 'width',
+      },
+      transitionDuration: {
+        5000: '5000ms',
+      },
```
rebelchris, line 225: This won't be needed until we do CSS one?

rebelchris (changes requested):
Small cleanup task 🙏

packages/shared/src/components/notifications/Toast.tsx
TomDowling: From testing I can't see a way to refactor this as without the !testRef.current being top level it create a loop that keeps re doing the startAnimation

packages/shared/src/components/notifications/Toast.tsx
```diff
@@ -41,17 +41,17 @@ const Toast = ({
...
     },
   );
 
+  if (!toastRef.current && toast?.message) {
+    toastRef.current = toast;
+    startAnimation(toast.timer);
+  } else if (toastRef.current && toastRef.current !== toast && toast?.message) {
+    toastRef.current = toast;
+    startAnimation(toast.timer);
+  }
```
rebelchris, line 53: Wait we missing the cleanup here now right?

rebelchris (changes requested):

packages/shared/src/components/notifications/Toast.tsx
```diff
@@ -41,17 +41,18 @@ const Toast = ({
...
-
-        await nextTick(); // wait (1ms) for the component to render so animation can be seen
-        startAnimation(data.timer);
-      },
     },
   );
 
+  if (!toastRef.current && toast?.message) {
+    toastRef.current = toast;
+    startAnimation(toast.timer);
```
capJavert, line 49: Since `startAnimation` is calling `setTimer` this effectively sets state while rendering which is bad an can lead to infinite loops. I get it that you guard with refs but there must be a better solution.

Won't block if you tested it and you decided with others to move CSS solution to backlog but we should be very careful with this.

capJavert (commented):
My only concern is with state setting during rendering.

packages/shared/src/components/notifications/Toast.tsx
TomDowling: Having spoke to Chris, have confirmed we are okay to continue especially with the other ticket now on the backlog for the refactor. but I will keep an eye on this incase if does have a negative effect on anything

rebelchris (approved):

PR #2032, opened 2023-08-10
refactor: search input as a standalone component

sshanzel:
Note: The component in itself was written amazingly, **this is purely a refactor** to break down the components and to ease the review down the line.

## Changes
- To revamp the Feed Header, there are certain differences between the placements of the elements from the Search page but using the same `SearchInputBar`.
- Broke down the component to allow other implementations to have freedom of what to include.
- @TomDowling would love to hear your thoughts and if I missed anything as you are very familiar with the component 🙏 

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1639

### DIFF
--- a/packages/shared/src/components/search/SearchBar.tsx
+++ b/packages/shared/src/components/search/SearchBar.tsx
@@ -8,9 +8,15 @@ import AuthContext from '../../contexts/AuthContext';
 import useSidebarRendered from '../../hooks/useSidebarRendered';
 import { SearchBarInput, SearchBarInputProps } from './SearchBarInput';
 
-export type SearchBarProps = Pick<SearchBarInputProps, 'className'>;
+export type SearchBarProps = Pick<
+  SearchBarInputProps,
+  'className' | 'valueChanged' | 'onSubmit'
+>;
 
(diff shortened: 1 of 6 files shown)


packages/shared/src/components/search/SearchBar.tsx
```diff
@@ -1,264 +1,46 @@
...
+        suggestion: 'Sign up and read your ',
+        onClick: () => showLogin('search bar suggestion'),
+      },
+      {
+        suggestion:
+          'Sign up and read your first post to get search recommendations',
+        onClick: () => showLogin('search bar suggestion'),
+      },
+    );
   }
```
rebelchris, line 35: This should only have one right?

rebelchris (approved):
Just one comment, you might want to address before merging.

packages/shared/src/components/search/SearchBar.tsx
sshanzel: Oh, right. Was testing some stuff. Let me fix it up.

PR #940, opened 2022-03-29
refactor: add tldr summary to seo description plus test

ayadotdev:
### Describe what this PR does

This PR adds a TLDR summary to the seo description.

### On those affected packages:
- [x] Have you done sanity checks in the webapp?
- [x] Have you done sanity checks in the extension?

### Did you test the modified components media queries?
- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

DD-520 #done

PR #2033, opened 2023-08-10
fix: context menu state management

ashrafchowdury:
## Changes

Added a `useState` to control the flow of the `onMenuClick` function. This will help us to toggle the menu button very easily.

- **Fixed the toggle menu issue**, now we can toggle the **Post menu and also Squade menu** very easily.

## Events

- [x] MobileL (420px)
- [x] Tablet (656px)
- [x] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

Closes #940


packages/shared/src/components/post/PostHeaderActions.tsx
```diff
@@ -130,7 +123,7 @@ export function PostHeaderActions({
         <Button
           className={classNames('btn-tertiary', !inlineActions && 'ml-auto')}
           icon={<MenuIcon />}
-          onClick={(event) => showPostOptionsContext(event)}
+          onClick={(event) => onMenuClick(event)}
```
sshanzel, line 126: I think we can simplify this with the suggestion below.
```suggestion
          onClick={onMenuClick}
```

packages/shared/src/hooks/useContextMenu.ts
```diff
@@ -1,15 +1,18 @@
...
 import { useContextMenu as useContexifyContextMenu } from '@dailydotdev/react-contexify';
 
 export default function useContextMenu({ id }: { id: string }): {
   onMenuClick: (e: React.MouseEvent) => void;
 } {
+  const [menu, setMenu] = useState(true);
   const { show } = useContexifyContextMenu({ id });
 
   const onMenuClick = (e: React.MouseEvent) => {
+    !menu ? setMenu(true) : setMenu(false);
```
sshanzel, line 11: This should do the setting of the value itself.
```suggestion
    setMenu(!menu);
```

packages/shared/src/hooks/useContextMenu.ts
```diff
@@ -1,15 +1,18 @@
+import { useState } from 'react';
 import { useContextMenu as useContexifyContextMenu } from '@dailydotdev/react-contexify';
 
 export default function useContextMenu({ id }: { id: string }): {
   onMenuClick: (e: React.MouseEvent) => void;
 } {
+  const [menu, setMenu] = useState(true);
```
sshanzel, line 7: The default value should be false.
```suggestion
  const [menu, setMenu] = useState(false);
```

packages/shared/src/hooks/useContextMenu.ts
```diff
@@ -1,15 +1,18 @@
...
 
   const onMenuClick = (e: React.MouseEvent) => {
+    !menu ? setMenu(true) : setMenu(false);
     const { right, bottom } = e.currentTarget.getBoundingClientRect();
-    show(e, {
-      position: { x: right, y: bottom + 4 },
-    });
+    menu && show(e, {
+        position: { x: right, y: bottom + 4 },
+      });
```
sshanzel, line 15: I think the condition is the opposite. Also, let's also move my suggestion about `setMenu` to the last line of the this function.
```suggestion
    if (!menu) {
      show(e, {
        position: { x: right, y: bottom + 4 },
      });
    }
```

sshanzel (changes requested):
Great work! There are still a few things to change, but you are well on your way! 🚀

sshanzel (commented):
Looks great! Just had some minor non-blocking comments but we are almost there!

I just found out one edge-case, when you open the menu and click anywhere else, it would close, but then the state was not updated with the closing. Hence, state still says `true` and if you click the button to open it again, it won't since the state thinks it is still open when it is not.

There are numerous ways to go about this so I will discuss it with the team on what would be the preferred method for everyone else.

sshanzel:
Hi @ashrafchowdury thank you for your patience. The team has decided on how we should go about this now. Instead of managing the state through `useState`, we will do it through `useQuery` to make it easier to manage the values anywhere else. It would look exactly like the `useLazyModal` hook which manages the state of our modals through `useQuery`. The only difference would then be the context menu id is going to be part of the query key (first parameter of `useQuery`).

After that, it would allow us to fix the edge case I have mentioned in [this comment](https://github.com/dailydotdev/apps/pull/2033#pullrequestreview-1582416672) easily. There is a `PortalMenu` file, and on line 29 you will find the component we are currently using for all the menus. You will then use the context menu hook here and extract the function to hide the menus to set the state properly through `onHidden` property of the `PortalMenu`. This way we will only do it once instead of updating existing implementations.

sshanzel:
I understand we have personal obligations and we really appreciate the time and effort you have put into this 🙏 

This has become unexpectedly complicated to work with now, let me know if you need any help or if you would prefer me to take over the last missing pieces 🙌

sshanzel:
@rebelchris I have now updated the implementation to utilize query cache over state to avoid the edge case mentioned in this PR.

Now, this should automatically fix existing implementations utilizing the same hook and component. If we ever come across any context menu that is not using the same, we should then update accordingly or would you want to create another ticket to run through the whole app and check everything?

Preview:

https://github.com/dailydotdev/apps/assets/13744167/79c465b1-fbad-473f-a10f-ebd45e3092af

packages/shared/src/hooks/useContextMenu.ts
```diff
@@ -13,20 +15,38 @@ interface UseContextMenu {
 export default function useContextMenu({
   id,
 }: UseContextMenuProps): UseContextMenu {
+  const key = useMemo(
+    () => generateQueryKey(RequestKey.ContextMenu, null, id),
```
rebelchris, line 19: Probably tag user here

rebelchris (approved):
To me it looks good just the user key part.
@capJavert Maybe you can quickly look for the dependency arrays?

packages/shared/src/hooks/useContextMenu.ts
capJavert: Since we don't actually set any user related state, its just boolean.. I would say its ok for this not to be user tagged.

sshanzel:
@ashrafchowdury we have now merged the fix. Thank you for the effort! 🥳

ashrafchowdury:
> @ashrafchowdury we have now merged the fix. Thank you for the effort! 🥳

Thank you for your immersive support @sshanzel 🥰

You are one of the great OS maintainers I have ever seen 💯🤝

PR #2036, opened 2023-08-10
feat: waiting to redirect from onboarding till feed is ready

TomDowling:
## Changes
Currently after onboarding the flicker happens as the homepage loads before the user my-feed is ready. after a few seconds it is created and the page re-renders taking the user from popular to my-feed. This fix leaves the flow the same if you haven't selected any topics in the onboarding but wait for your feed if you have selected some

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1546 #done

### DIFF
--- a/packages/shared/src/components/onboarding/FilterOnboarding.tsx
+++ b/packages/shared/src/components/onboarding/FilterOnboarding.tsx
@@ -1,4 +1,4 @@
-import React, { ReactElement, useEffect, useState } from 'react';
+import React, { ReactElement, useState } from 'react';
 import classNames from 'classnames';
 import FeedTopicCard, { ButtonEvent } from '../containers/FeedTopicCard';
 import useTagAndSource from '../../hooks/useTagAndSource';
@@ -8,13 +8,11 @@ import { Origin } from '../../lib/analytics';
 interface FilterOnboardingProps {
   onSelectedTopics?(tags: Record<string, boolean>): void;
   className?: string;
-  setHasCategories?: (hasCategories: boolean) => void;
 }
 
 export function FilterOnboarding({
   onSelectedTopics,
   className,
-  setHasCategories,
 }: FilterOnboardingProps): ReactElement {
   const [invalidMessage, setInvalidMessage] = useState<string>(null);
   const [selectedTopics, setSelectedTopics] = useState({});
@@ -36,13 +34,6 @@ export function FilterOnboarding({
       setInvalidMessage(null);
     }
   };
-
-  useEffect(() => {
-    const hasSelectedTopics = Object.values(selectedTopics).some(
-      (value) => value === true,
-    );
-    if (setHasCategories) setHasCategories(hasSelectedTopics);
-  }, [selectedTopics, setHasCategories]);
 
   return (
     <div
--- a/packages/shared/src/components/onboarding/FilterOnboarding.tsx
+++ b/packages/shared/src/components/onboarding/FilterOnboarding.tsx
@@ -1,4 +1,4 @@
-import React, { ReactElement, useState } from 'react';
+import React, { ReactElement, useEffect, useState } from 'react';
 import classNames from 'classnames';
 import FeedTopicCard, { ButtonEvent } from '../containers/FeedTopicCard';
 import useTagAndSource from '../../hooks/useTagAndSource';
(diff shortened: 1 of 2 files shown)


packages/webapp/pages/onboarding.tsx
```diff
@@ -101,9 +110,21 @@ export function OnboardPage(): ReactElement {
   const onSuccessfulTransaction = () => {
     onShouldUpdateFilters(true);
     setFinishedOnboarding(true);
+    if (hasCategories) {
+      return;
+    }
```
sshanzel, line 115: When does this happen? Why are we stopping the redirection from here?

TomDowling: We are doing this to ensure if the user has set categories we wait for their feed to be created before redirecting them so the page doesn't just going to popular then my-feed

sshanzel: Right. Good catch. Thanks for the clarification.

sshanzel (approved):
Looks good to me! Thanks for the fix!

Would appreciate to have another pair of eyes before merging as this is part of our critical flows.

@dailydotdev/web-team

packages/shared/src/components/onboarding/FilterOnboarding.tsx
```diff
@@ -35,6 +37,13 @@ export function FilterOnboarding({
     }
   };
 
+  useEffect(() => {
+    const hasSelectedTopics = Object.values(selectedTopics).some(
+      (value) => value === true,
+    );
+    if (setHasCategories) setHasCategories(hasSelectedTopics);
+  }, [selectedTopics, setHasCategories]);
+
```
capJavert, line 46: Since `setHasCategories` is optional we can early return before ever calculating `hasSelectedTopics`.

packages/shared/src/components/onboarding/FilterOnboarding.tsx
```diff
@@ -8,11 +8,13 @@ import { Origin } from '../../lib/analytics';
 interface FilterOnboardingProps {
   onSelectedTopics?(tags: Record<string, boolean>): void;
   className?: string;
+  setHasCategories?: (hasCategories: boolean) => void;
 }
 
 export function FilterOnboarding({
   onSelectedTopics,
   className,
+  setHasCategories,
```
capJavert, line 17: I would make this something more generic.. eg. `onSelectTopic` and then parent component can calculate `hasCategories` from topics it receives. It would be more reusable down the line if we want to hook up something else related to topics selection. WDYT?

capJavert (commented):
Few concerns over the flow for your evaluation 🙏

packages/shared/src/components/onboarding/FilterOnboarding.tsx
sshanzel: Also, after looking at the code of `onChangeSelectedTopic` I think we don't have to add another callback and just base it on the `onSelectedTopics` callback. This would also reduce the usage of `useEffect`.

packages/webapp/pages/onboarding.tsx
```diff
@@ -92,6 +100,7 @@ export function OnboardPage(): ReactElement {
   };
 
   const formRef = useRef<HTMLFormElement>();
+  const [onSelectTopic, setOnSelectTopic] = useState(false);
```
sshanzel, line 103: Let's reserve the `on` prefix on names for functions/handlers. I think we can swap the `hasSelectedTopics` to this one.

packages/webapp/pages/onboarding.tsx
```diff
@@ -101,9 +110,21 @@ export function OnboardPage(): ReactElement {
...
   };
 
+  useEffect(() => {
+    if (!hasSelectTopics || !alerts?.myFeed) return;
+
+    if (alerts.myFeed === 'created') {
+      router.push('/');
+    }
+  }, [alerts, hasSelectTopics, router]);
+
```
capJavert, line 127: Could we not even move this to `hasSelectedTopics` callback and just redirect from there? Sorry for multiple changes, but just realised it now when you adjusted.

TomDowling: The issue i found it wasn't ready at that point the alerts took a while to update to created / till it was ready. Doing it in hasSelectedTopics would be too early

capJavert: Ah, of course because of update query.

capJavert (approved):

PR #2039, opened 2023-08-14
feat: support for companion call to action modal

DragosIorgulescu:
## Changes

- [x] add a new `CompanionModal` component as per spec
- [x] add a custom hook handling the companion modal trigger -- wraps the existing read article handlers
- [x] update feed and post page read article actions to trigger the modal accordingly
- [x] refactor `useExtensionPermission` hook to work outside the extension package
- [x] add feature flag support via GrowthBook
- [x] add a new action type to make sure the modal is displayed just once
  - triggered when closing the modal, or when clicking on any of the CTA 

**Preview**
--- 

companion-modal-feed
companion-modal-post-modal

## Events

A new event has been added: https://docs.google.com/spreadsheets/d/18Lv7zXges9QfVX5VYL1a-Hyl0e1sQ3sLr0OK8YZWKXI/edit#gid=0&range=169:169

| New | decline content scripts  | extra: { origin: 'companion modal permission button' } |

- [x] MobileL (420px)
- [x] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1601 #done


packages/shared/src/components/GrowthBookProvider.tsx
```diff
@@ -80,7 +80,8 @@ export const GrowthBookProvider = ({
           'AES-CBC',
           128,
         ).then((features) => {
-          gb.setFeatures(JSON.parse(features));
+          console.log(features);
+          // gb.setFeatures(JSON.parse(features));
```
DragosIorgulescu, line 84: This was returning an invalid JSON string for me locally, will need some support debugging this as I don't have direct access to Grouth Book

packages/shared/src/components/modals/CompanionModal.tsx
```diff
@@ -0,0 +1,128 @@
...
+        className="hidden tablet:block rounded-16 w-fit"
+        poster="https://daily-now-res.cloudinary.com/image/upload/s--7w2Deabi--/f_auto/v1689497816/web-image_yomdsz"
+        loop
+        autoPlay
+      >
+        <source
+          src="https://daily-now-res.cloudinary.com/video/upload/s--2N7_JImI--/v1689497504/final_web_rkupko.mp4"
+          type="video/mp4"
+        />
+      </video>
```
DragosIorgulescu, line 56: this is bad, but we need a function which dictates if we are on desktop or mobile outside of css media queries so we can handle this. Is there anything already in the codebase for this? 

If not, I can add something, unless someone has a different idea.

DragosIorgulescu:
I tried to be as unintrusive as possible with adding this, which posed some challenges due to how the `read article` links are being implemented, since there are multiple triggers for this and each one works slightly differently.

@rebelchris @capJavert I have a couple of notes/questions before wrapping this up
- is there a way to determine whether we are on mobile/tablet/etc via js in the app currently -- the modal size and the video tag depend on this
- the `useExtensionPermission` looks to be deeply tied to the extension package. However, since the components where we need to hook it up are shared, simply importing the hook where needed breaks. It seems to be related to `import 'content-scripts-register-polyfill';` but I am not entirely sure yet.
- I'd like to test if the events are firing as expected -- can someone point me to how you usually do this?
- I will be updating specs once I get a green light on the overall approach

packages/shared/src/components/Feed.tsx
```diff
@@ -67,6 +68,11 @@ const SharePostModal = dynamic(
     import(/* webpackChunkName: "sharePostModal" */ './modals/SharePostModal'),
 );
 
+const ActivateCompanionModal = dynamic(
+  () =>
+    import(/* webpackChunkName: "CompanionModal" */ './modals/CompanionModal'),
+);
+
```
DragosIorgulescu, line 75: Ideally, we would not have this duplicated usage of the `CompanionModal` in both `Feed` and `PostContent`, we could place it higher in the tree, but then we would need to add a `Context` to manage its state -- I am open to suggestions if the team prefer this

sshanzel: Good point. Though the state updates might cause re-renders. We currently have not been using `memo`, so we are not yet optimized for unexpected re-renders.

packages/shared/src/components/GrowthBookProvider.tsx
rebelchris: We can maybe look into the SDK to see if there is a local overwrite available

packages/shared/src/components/modals/CompanionModal.tsx
rebelchris: We do have a `useMedia` hook, that's likely the one you want to use here.

packages/shared/src/hooks/useCompanionTrigger.tsx
```diff
@@ -0,0 +1,122 @@
...
+  ) => Promise<void>,
+): CompanionTrigger {
+  const isExtension = process.env.TARGET_BROWSER;
+  const { trackEvent } = useContext(AnalyticsContext);
+  const [data, setData] = useState<CompanionTriggerProps>();
+  const [isOpen, toggleOpen] = useState(false);
+  const featureEnabled = useFeatureIsOn(
+    Features.EngagementLoopJuly2023Companion.toString(), // this feels very smelly, but otherwise complains because of the Features<string> type
+  );
+  // const { requestContentScripts, contentScriptGranted, isFetched } =
```
rebelchris, line 42: This is commented

packages/shared/src/hooks/useCompanionTrigger.tsx
```diff
@@ -0,0 +1,122 @@
+import React, { useState, useCallback, useContext } from 'react';
+import { useFeatureIsOn } from '@growthbook/growthbook-react';
+import { FeedPostClick } from './feed/useFeedOnPostClick';
+import { Post } from '../graphql/posts';
+import { Features } from '../lib/featureManagement';
+import AnalyticsContext from '../contexts/AnalyticsContext';
+
+// import { useExtensionPermission } from '../../../extension/src/companion/useExtensionPermission';
```
rebelchris, line 8: This is commented

packages/shared/src/hooks/useCompanionTrigger.tsx
```diff
@@ -0,0 +1,122 @@
...
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
+
+  const articleClickHandler = useCallback(
+    async (
+      e: React.MouseEvent,
+      { post, index, row, column }: Partial<CompanionTriggerProps> = {},
+    ) => {
+      if (!isExtension) {
+        // TODO: add contentScriptGranted || isFetched from useExtensionPermission
```
rebelchris, line 71: Todo still needed?

DragosIorgulescu: Yep, this and the 3 above you mentioned need to be enabled once we figure out how to go about using the `useExtensionPermission` hook -- see second note here: https://github.com/dailydotdev/apps/pull/2039#issuecomment-1676987569

TomDowling:
@dailydotdev/web-team, What are peoples thoughts on the desktop modal the gap to the left where the video goes behind the content looks odd imo, especially when there is text, it looks to flicker and becomes distracting, would love to get other peoples thoughts though.

https://github.com/dailydotdev/apps/assets/42202149/291a8070-adfa-4180-89d5-0462f3349405

DragosIorgulescu:
@TomDowling I totally agree, it is not pleasant on the eyes. This is how the design spec requires this to look, but maybe we can change it if we get a green light from @tsahimatsliah

packages/shared/src/components/Feed.tsx
```diff
@@ -99,6 +105,7 @@ export default function Feed<T>({
   options,
   allowPin,
 }: FeedProps<T>): ReactElement {
+  React.useState(false);
```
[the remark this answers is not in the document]
DragosIorgulescu: removed, just some leftover code from my first iteration on this task

packages/extension/src/companion/useExtensionPermission.ts
```diff
@@ -77,16 +72,53 @@ export const useExtensionPermission = ({
...
+      });
+
+      if (onPermission) onPermission(false);
     }
 
     return granted;
   };
+};
+
+export const useContentScriptStatus = (): {
```
rebelchris, line 94: I'm not really sure why we need another hook for this?
Don't we have this knowledge already somewhere (the one we base the companion button off?)

packages/shared/src/components/Feed.tsx
```diff
@@ -436,6 +454,13 @@ export default function Feed<T>({
           onRequestClose={closeSharePost}
         />
       )}
+      <ActivateCompanionModal
```
rebelchris, line 457: We probably should convert this to a lazyModal as it has multiple invocation.
I think we can then even handle some of the shared login inside the companion activation hook or a new simple hook.

packages/shared/src/components/modals/CompanionModal.tsx
```diff
@@ -0,0 +1,128 @@
...
+        </p>
+
+        <Button
+          className="mb-4 w-full btn-primary"
+          onClick={onActivateCompanion}
+        >
+          Read and activate widget
+        </Button>
+
+        <Button
```
rebelchris, line 75: We'll have to add the relevant `rel` attribute here (same we use for the normal button should do)

packages/shared/src/components/post/PostContent.tsx
```diff
@@ -96,6 +103,17 @@ export function PostContent({
...
+  const {
+    onPostArticleClick: onReadArticleClick,
+    activateCompanion,
+    openArticle,
+    isOpen: companionModalOpen,
+    toggleOpen: companionModalToggle,
+
+    // @NOTE see https://dailydotdev.atlassian.net/l/cp/dK9h1zoM
+    // eslint-disable-next-line react-hooks/rules-of-hooks
+  } = useCompanionTrigger(onReadArticle);
```
rebelchris, line 115: Yeah seeing this, by converting to lazyModal we should be able to move all logic to one place :)

packages/shared/src/hooks/useExtensionPermission.ts
```diff
@@ -0,0 +1,38 @@
...
+}: UseExtensionPermissionProps): ExtensionPermission => {
+  const client = useQueryClient();
+  const { requestContentScripts, contentScriptGranted, isFetched } =
+    client.getQueryData<{
+      requestContentScripts: RequestContentScripts;
+      contentScriptGranted: boolean;
+      isFetched: boolean;
+    }>(EXTENSION_PERMISSION_KEY) || {};
+
+  return {
```
rebelchris, line 33: Not sure we need this one or the other one.

packages/shared/src/lib/featureManagement.ts
```diff
@@ -8,6 +8,10 @@ export class Features<T extends FeatureValue = string> {
 
   static readonly SubmitArticle = new Features('submit_article');
 
+  static readonly EngagementLoopJuly2023Companion = new Features(
+    'engagement_loop_july2023_companion',
+  );
```
rebelchris, line 13: Features was still based on flagsmith as our experiment platform, I'm not sure you still need it for Growthbook values

packages/webapp/pages/_app.tsx
```diff
@@ -81,6 +82,9 @@ function InternalApp({ Component, pageProps, router }: AppProps): ReactElement {
   useInAppNotification();
   const { modal, closeModal } = useLazyModal();
   usePrompt();
+
+  useQuery(EXTENSION_PERMISSION_KEY, () => ({}));
```
rebelchris, line 86: Wait we reset it to empty?
Can't we make it cleaner, this seems very viable to someone removing it.

packages/shared/src/components/Feed.tsx
DragosIorgulescu: done with https://github.com/dailydotdev/apps/pull/2039/commits/7fa0a89f8e9e8b041e43e7a0da91a27277b5db0c

packages/shared/src/components/Feed.tsx
```diff
@@ -67,6 +69,11 @@ const SharePostModal = dynamic(
     import(/* webpackChunkName: "sharePostModal" */ './modals/SharePostModal'),
 );
 
+const ActivateCompanionModal = dynamic(
+  () =>
+    import(/* webpackChunkName: "CompanionModal" */ './modals/CompanionModal'),
+);
```
rebelchris, line 75: This can be removed right?

packages/shared/src/components/Feed.tsx
```diff
@@ -436,6 +452,14 @@ export default function Feed<T>({
...
         />
       )}
+
+      {lazyModal?.type === LazyModal.CompanionModal && (
+        <ActivateCompanionModal
+          isOpen
+          onRequestClose={() => companionModalToggle(false)}
+          {...lazyModal.props}
+        />
+      )}
```
rebelchris, line 462: Ah you can actually simple call openModal from lazyModal to open it.
No need to have it explicitly in the files anymore.
This way you can potentially also move the logic to useCompanionTrigger?

packages/extension/src/companion/useExtensionPermission.ts
DragosIorgulescu: it's not exactly another hook. It is the same function which previously was implemented inside the `useExtensionPermission` hook -- I extracted it out so we can dynamically attach it.

The reason you see duplication here is because the legacy `useExtensionPermission` hook is still used in a few places, so replacing it entirely would be out of scope for this task

packages/shared/src/components/Feed.tsx
DragosIorgulescu: This should be done now, let me know what you think

packages/shared/src/hooks/useExtensionPermission.ts
DragosIorgulescu: This one is used when we need to use this hook inside a component that is shared between the webapp and extension packages. Long term, we would remove both and have a dedicated context handling this.

packages/shared/src/lib/featureManagement.ts
DragosIorgulescu: As per [DR requirements](https://dailydotdev.atlassian.net/wiki/spaces/HAN/pages/578617368/WT-1528+Engagement+loops#Experiment-setup), this was specified to be added to the `Features` class -- however, I agree that we can safely move away from it, at the same time, having these constants here does not hurt either?

packages/webapp/pages/_app.tsx
DragosIorgulescu: I will add a note here explaining what this does, however, keep in mind that this is not the long term solution, I am completely open to working on a cleaner way to manage these things in a separate PR if you want.

packages/shared/src/lib/featureManagement.ts
rebelchris: Since Growthbook is something new, rather we find a new way of using them.
So maybe use the Features class, but a new way of defining them.

Think Growthbook should be ok with simply enum type somewhere maybe?

sshanzel: Have found a way to still keep the Features. The hook is in the link below.

Sample usage:
```
const searchValue = useFeature(Features.Search);
```

The hook would utilize the default value we set from the original class. 

https://github.com/dailydotdev/apps/pull/2046/files#diff-8e32a1b02c9379ceef35c485660f1fbde721268a14e781f7dde68d706fafa48fR145-R148

packages/extension/src/companion/CompanionPopupButton.tsx
```diff
@@ -42,6 +44,8 @@ export const CompanionPopupButton = (): ReactElement => {
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [contentScriptGranted, isFetched]);
 
+  console.log(contentScriptGranted, isFetched);
```
rebelchris, line 47: console log still here :)

packages/extension/src/lib/extensionScripts.ts
```diff
@@ -77,21 +60,39 @@ export const useExtensionPermission = ({
...
+      if (onPermission) {
+        await onPermission(true);
+      } else {
+        window.open(companionPermissionGrantedLink, '_blank');
+      }
+    } else {
+      trackEvent({
+        event_name: 'decline content scripts',
+        extra: JSON.stringify({ origin }),
+      });
```
rebelchris, line 73: Where was this before?

packages/shared/src/components/Feed.tsx
```diff
@@ -225,6 +227,15 @@ export default function Feed<T>({
     'go to link',
   );
 
+  const {
+    onFeedArticleClick: onReadArticleClick,
+    toggleOpen: companionModalToggle,
+    lazyModal,
```
rebelchris, line 233: These would not be needed here right?

packages/shared/src/components/post/PostContent.tsx
```diff
@@ -96,6 +98,15 @@ export function PostContent({
     onRemovePost,
   };
 
+  const {
+    onPostArticleClick: onReadArticleClick,
+    toggleOpen: companionModalToggle,
+    lazyModal,
```
rebelchris, line 104: Same here :)

packages/shared/src/hooks/useCompanionTrigger.tsx
```diff
@@ -0,0 +1,159 @@
...
+
+  const articleClickHandler = useCallback(
+    async (
+      e: React.MouseEvent,
+      { post, index, row, column }: Partial<CompanionTriggerProps> = {},
+    ) => {
+      // the check whether the user is logged in is done on the GrowthBook side
+      // the feature flag is enabled only for logged-in users
+      // -- check that this is correct
+      console.log({ isExtension, contentScriptGranted });
```
rebelchris, line 63: Log here

packages/shared/src/hooks/useCompanionTrigger.tsx
```diff
@@ -0,0 +1,159 @@
...
+      if (!isExtension || (isExtension && contentScriptGranted)) {
+        await customPostClickHandler(post, index, row, column);
+      } else {
+        e.preventDefault();
+        // setData({ post, index, row, column });
```
rebelchris, line 68: Not sure what this is for?

packages/webapp/pages/_app.tsx
```diff
@@ -81,6 +82,14 @@ function InternalApp({ Component, pageProps, router }: AppProps): ReactElement {
...
   const { modal, closeModal } = useLazyModal();
   usePrompt();
+
+  // This is here to ensure compatibility between the extension and the webapp
+  // Do not remove, otherwise things will fail
+  // TODO: clean up the way this is handled so we don't need to do this
+  useQuery(EXTENSION_PERMISSION_KEY, () => ({
+    useContentScriptStatus: () => ({}),
+  }));
+
```
rebelchris, line 92: I think with this new flow we can actually omit this and no longer use this.

packages/extension/src/lib/extensionScripts.ts
DragosIorgulescu: this did not exist in the past, it is a new event being tracked as per spec: https://dailydotdev.atlassian.net/wiki/spaces/HAN/pages/578617368/WT-1528+Engagement+loops#Tracking-changes

Updated the analytics taxonomy as well: https://docs.google.com/spreadsheets/d/18Lv7zXges9QfVX5VYL1a-Hyl0e1sQ3sLr0OK8YZWKXI/edit#gid=0&range=169:169

packages/shared/src/hooks/useCompanionTrigger.tsx
DragosIorgulescu: leftover code from a different approach I took where I was storing the relevant data in a local state variable.

packages/shared/src/lib/featureManagement.ts
DragosIorgulescu: 👍  thanks @sshanzel , I've added the equivalent `useFeatureIsOn` to this PR: https://github.com/dailydotdev/apps/pull/2039/files#diff-8e32a1b02c9379ceef35c485660f1fbde721268a14e781f7dde68d706fafa48fR146-R147

packages/extension/src/lib/extensionScripts.ts
```diff
@@ -77,21 +60,39 @@ export const useExtensionPermission = ({
...
-    () => ({
-      contentScriptGranted,
-      requestContentScripts,
-      isFetched,
-      registerBrowserContentScripts,
-    }),
-    // @NOTE see https://dailydotdev.atlassian.net/l/cp/dK9h1zoM
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-    [contentScriptGranted, isFetched],
+export const useContentScriptStatus = (): {
```
rebelchris, line 82: Maybe just me missing a explanation, but why is this one still needed as duplicate?

packages/shared/src/components/modals/CompanionModal.tsx
```diff
@@ -0,0 +1,143 @@
...
+      {...props}
+      className="border-none"
+      kind={Modal.Kind.FlexibleCenter}
+      size={size}
+    >
+      {isTablet && (
+        // eslint-disable-next-line jsx-a11y/media-has-caption
+        <video
+          className="tablet:hidden rounded-16 w-fit"
+          poster={cloudinary.companion.videoThumbnailMobile}
```
rebelchris, line 46: I would like to rather see the poster and src change based on the isTablet, think it will cleanup the duplicate video a bit.

packages/webapp/pages/_app.tsx
```diff
@@ -39,6 +39,7 @@ import { useLazyModal } from '@dailydotdev/shared/src/hooks/useLazyModal';
 import { usePrompt } from '@dailydotdev/shared/src/hooks/usePrompt';
 import { ReactQueryDevtools } from 'react-query/devtools';
 import { defaultQueryClientConfig } from '@dailydotdev/shared/src/lib/query';
+import { EXTENSION_PERMISSION_KEY } from '@dailydotdev/shared/src/hooks/useExtensionPermission';
```
rebelchris, line 42: Assume we can remove these right, as it doesn't see use (might have fixed it with build lints already)

packages/shared/src/hooks/useCompanionTrigger.tsx
```diff
@@ -0,0 +1,176 @@
...
+  ) => Promise<void>,
+): CompanionTrigger {
+  const isExtension = process.env.TARGET_BROWSER;
+  const { trackEvent } = useContext(AnalyticsContext);
+  const { closeModal, openModal } = useLazyModal();
+  const { isActionsFetched, checkHasCompleted, completeAction } = useActions();
+
+  const featureEnabled = useFeatureIsOn(
+    Features.EngagementLoopJuly2023Companion,
+  );
```
rebelchris, line 48: This is a essential part, we have to ensure this will only trigger when you are on the extension + on a main feed page.
Any other case should not even trigger it, it has to do with when the experiment needs to start for users.

Personally think we can basically have a early exit if you are not on extension + mainFeed *(there is some example of the later bit for onboarding redirection for example)

packages/shared/src/hooks/useCompanionTrigger.tsx
```diff
@@ -0,0 +1,176 @@
...
+  }, [isActionsFetched]);
+
+  const articleClickHandler = useCallback(
+    async (
+      e: React.MouseEvent,
+      { post, index, row, column }: Partial<CompanionTriggerProps> = {},
+    ) => {
+      // the check whether the user is logged in is done on the GrowthBook side
+      // the feature flag is enabled only for logged-in users
+      // -- check that this is correct
```
rebelchris, line 72: Not sure this is the case with growthbook (flagsmith didn't have this, but we need to check with @idoshamun )

packages/shared/src/hooks/useCompanionTrigger.tsx
```diff
@@ -0,0 +1,176 @@
...
+      if (
+        !isExtension ||
+        !featureEnabled ||
+        alreadyCompleted ||
+        contentScriptGranted
+      ) {
+        await customPostClickHandler(post, index, row, column);
+      } else {
+        e.preventDefault();
```
rebelchris, line 81: Suggestion is alternative maybe we can only define the useFeatureIsOn here?
I know it's a bit anti-pattern, but that would make sure everything keeps working.

packages/shared/src/hooks/useCompanionTrigger.tsx
```diff
@@ -0,0 +1,176 @@
...
+        // eslint-disable-next-line @typescript-eslint/no-use-before-define
```
rebelchris, line 82: As for this, can't we simply move the modal down?

packages/shared/src/hooks/useCompanionTrigger.tsx
```diff
@@ -0,0 +1,176 @@
...
+        toggleModal(true, { post, index, row, column });
+      }
+    },
+
+    // @NOTE see https://dailydotdev.atlassian.net/l/cp/dK9h1zoM
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
rebelchris, line 88: This can also take all deps as it's not a once-off correct?

packages/shared/src/hooks/useCompanionTrigger.tsx
```diff
@@ -0,0 +1,176 @@
...
+  const onPostArticleClick = useCallback<PostClick>(
+    async (e, { post }) => {
+      await articleClickHandler(e, { post });
+    },
+    [articleClickHandler],
+  );
+
+  const activateCompanion = useCallback(async () => {
+    await requestContentScripts();
+    // eslint-disable-next-line @typescript-eslint/no-use-before-define
```
rebelchris, line 114: I would move things so we adhear to the flow of order.

packages/shared/src/hooks/useCompanionTrigger.tsx
```diff
@@ -0,0 +1,176 @@
...
+    toggleModal(false);
+    handleCompleteAction();
+
+    // @NOTE see https://dailydotdev.atlassian.net/l/cp/dK9h1zoM
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
rebelchris, line 119: Callback, so we should be able to insert all deps right?

packages/shared/src/hooks/useCompanionTrigger.tsx
```diff
@@ -0,0 +1,176 @@
+import React, { useCallback, useContext } from 'react';
```
rebelchris, line 1: Not a real action point, but would love to hear what @capJavert and @sshanzel think.
To me this file is very hard to follow because of the complexity of hijacking.
I'm not even sure what the alternative is (expect our discussion of uniforming the click, but later on)

Just wondering if it's clear enough to proceed or we can do some simple renaming/refactoring for clarity sake.

rebelchris (changes requested):
Thanks for the changes, overall looks very good.
One important remark with growthbook is WHEN we invoke the hook. This should basically never happen when it's not viable for the experiment position.

That is pretty much the only blocking comment at this point.
Well done 👌

packages/shared/src/components/modals/CompanionModal.tsx
DragosIorgulescu: I actually did try that, but it turns out that if you update the `src` on a video while it's in a loop, it actually keeps playing the initial video instead of switching to the new one. 

For this reason, I kept it as is, even though not ideal. I'm open to other suggestions though

packages/shared/src/hooks/useCompanionTrigger.tsx
DragosIorgulescu: Gotcha, that makes sense, nice catch! Since this is not just checking the flag, but also enrolling the user in the experiment. I'll update 👍

packages/shared/src/hooks/useCompanionTrigger.tsx
idoshamun: This is actually right, we can set this parameter on growthbook's side. Just have to remember

packages/shared/src/components/modals/CompanionModal.tsx
```diff
@@ -0,0 +1,143 @@
...
+  onActivateCompanion,
+  url,
+  ...props
+}: Props): ReactElement {
+  const isTablet = !useMedia([tablet.replace('@media ', '')], [true], false);
+  const isMobile = !useMedia([mobileL.replace('@media ', '')], [true], false);
+
+  const size = useMemo(() => {
+    if (isMobile) return Modal.Size.XSmall;
+    if (isTablet) return Modal.Size.Small;
```
sshanzel, line 30: Shouldn't our modals be responsive enough that we don't have to do this through JS? I could be wrong though.

packages/shared/src/components/modals/CompanionModal.tsx
```diff
@@ -0,0 +1,143 @@
...
+
+  return (
+    <Modal
+      {...props}
+      className="border-none"
+      kind={Modal.Kind.FlexibleCenter}
+      size={size}
+    >
+      {isTablet && (
+        // eslint-disable-next-line jsx-a11y/media-has-caption
```
sshanzel, line 43: I think it would be better to add a caption than disabling the lint about the `a11y` concern.

packages/shared/src/components/modals/CompanionModal.tsx
```diff
@@ -0,0 +1,143 @@
...
+        >
+          <source
+            src={cloudinary.companion.backgroundVideoMobile}
+            type="video/mp4"
+          />
+        </video>
+      )}
+
+      {!isTablet && (
+        // eslint-disable-next-line jsx-a11y/media-has-caption
```
sshanzel, line 58: Same comment with regards to the caption.

packages/shared/src/hooks/useCompanionTrigger.tsx
sshanzel: Understandable though I agree we can probably do some refactoring, such as putting together the `toggleModal` and `activateCompanion` functions in a hook, probably called `useCompanionModal` or something similar. That alone would probably reduce the lines from this file and isolate the actions.

sshanzel (changes requested):
Looks great overall! Can only imagine the amount of work that has put into this 💪 

I've had a few comments but I wanted to mention that generally, we would want to sway away from disabling lint whenever possible. Especially with `a11y` even though we haven't focused more on it yet, we want to at least be compliant on cases we easily can.

packages/extension/src/lib/extensionScripts.ts
```diff
@@ -41,31 +35,20 @@ export const getContentScriptPermissionAndRegister =
...
-    contentScriptKey,
-    getContentScriptPermission,
-    {
-      refetchOnMount: false,
-      refetchOnReconnect: false,
-      refetchOnWindowFocus: false,
-      refetchIntervalInBackground: false,
-    },
-  );
+  const client = useQueryClient();
```
capJavert, line 44: Watch out this is a hook call inside a regular function. I am not sure how this even works because there is no context available. Probably preact magic but in React this would never work.

packages/shared/src/components/Feed.tsx
```diff
@@ -225,6 +226,12 @@ export default function Feed<T>({
     'go to link',
   );
 
+  const {
+    onFeedArticleClick: onReadArticleClick,
+    // @NOTE see https://dailydotdev.atlassian.net/l/cp/dK9h1zoM
+    // eslint-disable-next-line react-hooks/rules-of-hooks
+  } = useCompanionTrigger(triggerReadArticleClick);
+
```
capJavert, line 234: We need to add this to the potential Feed refactor to avoid calling hooks conditionally.  cc @rebelchris

packages/webapp/pages/_app.tsx
```diff
@@ -81,6 +81,7 @@ function InternalApp({ Component, pageProps, router }: AppProps): ReactElement {
   useInAppNotification();
   const { modal, closeModal } = useLazyModal();
   usePrompt();
+
```
capJavert, line 84: Just a cleanup to avoid unnecessary formatting changes like this.

packages/shared/src/hooks/useExtensionPermission.ts
```diff
@@ -0,0 +1,55 @@
...
+  requestContentScripts: () => Promise<boolean>;
+  registerBrowserContentScripts: () => Promise<never>;
+}
+
+interface UseExtensionPermissionProps {
+  origin: string;
+  onPermission?: (granted: boolean) => void;
+}
+
+export const useExtensionPermission = ({
```
capJavert, line 26: We should re-export his from `hooks/index.ts` as we do with all new hooks.

packages/shared/src/hooks/useCompanionTrigger.tsx
capJavert: I find it ok, more worried over a lot of `eslint-disable` for hook rules, all of those should be resolved before merging.

packages/extension/src/lib/extensionScripts.ts
capJavert: I think those should be passed as arguments to `requestContentScripts` or similar.

capJavert (changes requested):
Good job overall, I like how `useExtensionPermission` is now in shared 👌 

Generally most of the issues for me are with hooks usage and linter overrides. 

Also just to double check all the events actually fire per [DR](https://dailydotdev.atlassian.net/wiki/spaces/HAN/pages/578617368/WT-1528+Engagement+loops#Tracking-changes)? 🙏 This is important to track our [guardrails](https://dailydotdev.atlassian.net/wiki/spaces/HAN/pages/568492046/Companion+Install+Experiment+Design#Guardrails) for experiment.

packages/extension/src/lib/extensionScripts.ts
DragosIorgulescu: great catch, thanks 🙏

packages/shared/src/hooks/useCompanionTrigger.tsx
DragosIorgulescu: I've updated this to check for the feature flag as the last check -- had to switch to using the growthbook object instead of the hook -- let me know what you think

packages/shared/src/components/modals/CompanionModal.tsx
DragosIorgulescu: They are responsive but it does not seem like they match the design requirements by default -- they scale with the screen but are either too wide/don't leave any margins.

packages/extension/src/lib/extensionScripts.ts
DragosIorgulescu: It's not actually a duplicate, we just export it as a separate function so we can attach it to the context the same way we do with `requestContentScripts` and the others

sshanzel (approved):
Got nothing else to say but great work! 🚀 

Just make sure to run end to end tests before merging (after receiving approval from others too) as publishing to stores can take a while sometimes.

packages/shared/src/hooks/useCompanionTrigger.tsx
```diff
@@ -0,0 +1,160 @@
...
+      completeAction(ActionType.EngagementLoopJuly2023CompanionModal);
+  }, [isActionsFetched, completeAction]);
+
+  const closeModal = useCallback(() => {
+    closeLazyModal();
+    handleCompleteAction();
+
+    trackEvent({
+      event_name: `close companion permission popup`,
+      extra: JSON.stringify({ origin: 'automatic' }),
```
rebelchris, line 66: Prefer these in the OriginEnum please 🙏

packages/shared/src/hooks/useCompanionTrigger.tsx
```diff
@@ -0,0 +1,160 @@
...
+    if (isActionsFetched)
+      completeAction(ActionType.EngagementLoopJuly2023CompanionModal);
+  }, [isActionsFetched, completeAction]);
+
+  const closeModal = useCallback(() => {
+    closeLazyModal();
+    handleCompleteAction();
+
+    trackEvent({
+      event_name: `close companion permission popup`,
```
rebelchris, line 65: i do think we also have a eventNames enum if i'm not mistaken

packages/shared/src/hooks/useCompanionTrigger.tsx
```diff
@@ -0,0 +1,160 @@
...
+  );
+
+  const articleClickHandler = useCallback(
+    async (
+      e: React.MouseEvent,
+      { post, index, row, column }: Partial<CompanionTriggerProps> = {},
+    ) => {
+      // the check whether the user is logged in is done on the GrowthBook side
+      // the feature flag is enabled only for logged-in users
+      // -- check that this is correct
```
rebelchris, line 117: We can remove this comment now :)

packages/shared/src/hooks/useCompanionTrigger.tsx
```diff
@@ -0,0 +1,160 @@
...
+      if (!isExtension || alreadyCompleted || contentScriptGranted) {
+        await customPostClickHandler(post, index, row, column);
+      } else if (
+        // checking this also enrolls the user in the experiment
+        // so this must be the last check after all others
+        gb.isOn(Features.EngagementLoopJuly2023Companion.id)
```
rebelchris, line 123: @idoshamun FYI, this method is acceptable?

packages/shared/src/hooks/useCompanionTrigger.tsx
```diff
@@ -0,0 +1,160 @@
...
+        props: {
+          url: data?.post?.permalink,
+          onReadArticleClick: openArticle(data),
+          onActivateCompanion: activateCompanion,
+        },
+      });
+
+      trackEvent({
+        event_name: `open companion permission popup`,
+        extra: JSON.stringify({ origin: 'automatic' }),
```
rebelchris, line 104: Same comment as above on the type, please take careful not of the taxonomy sheet

packages/shared/src/hooks/useCompanionTrigger.tsx
rebelchris: Also according to taxonomy there can be three values:
`manual, auto, read post` please use as strict as defined

packages/shared/src/hooks/useCompanionTrigger.tsx
idoshamun: Yes

idoshamun: From what I understand this will be called upon click only. According to this document https://dailydotdev.atlassian.net/wiki/spaces/HAN/pages/568492046/Companion+Install+Experiment+Design#Enrolment-Event
The enrollment event should already be dispatched on the feed visit, even if the user hasn't clicked on anything. @DragosIorgulescu is my understanding of the current implementation correct?
I'm double checking with Dave and I'll get back to you.

rebelchris: Only extension feed (so excluding squad eg) version where it's applicable correct

DragosIorgulescu: 👍  rolled back to the initial implementation which triggers enrollment when the hook first loads

rebelchris: Yep we just discussed, Dragos will propse a new flow where it's only limited to extension for enabling users

DragosIorgulescu: apologies for the back and forth here, the current implementation should be correct: enrollment is triggered when landing on the feed only on the extension

idoshamun (approved):
Enrollment looks great!

rebelchris (approved):
Nice!
1 non blocking thing :)

packages/extension/src/lib/extensionScripts.ts
```diff
@@ -41,31 +33,17 @@ export const getContentScriptPermissionAndRegister =
...
-    },
-  );
-
-  const requestContentScripts = async () => {
+  client,
+  trackEvent,
+) => {
+  return async () => {
     trackEvent({
       event_name: 'request content scripts',
```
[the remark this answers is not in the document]
DragosIorgulescu: nice catch 🙏

packages/extension/src/lib/extensionScripts.ts
```diff
@@ -41,57 +34,55 @@ export const getContentScriptPermissionAndRegister =
...
+} => {
+  const { data: contentScriptGranted, isFetched } = useQuery(
+    contentScriptKey,
+    getContentScriptPermission,
+    {
+      refetchOnMount: false,
+      refetchOnReconnect: false,
+      refetchOnWindowFocus: false,
+      refetchIntervalInBackground: false,
+    },
```
capJavert, line 84: We have a `disableRefetch` export in `src/lib/func` that contains these options.

packages/shared/src/hooks/useCompanionTrigger.tsx
```diff
@@ -0,0 +1,167 @@
...
+    });
+
+  const { contentScriptGranted } = useContentScriptStatus();
+  const alreadyCompleted = checkHasCompleted(
+    ActionType.EngagementLoopJuly2023CompanionModal,
+  );
+
+  const handleCompleteAction = useCallback(() => {
+    if (isActionsFetched)
+      completeAction(ActionType.EngagementLoopJuly2023CompanionModal);
```
capJavert, line 67: You don't have to wait for this because we merge any locally completed actions once they are fetch so you can just call `completeAction`

packages/shared/src/hooks/useCompanionTrigger.tsx
```diff
@@ -0,0 +1,167 @@
...
+  const { closeModal: closeLazyModal, openModal: openLazyModal } =
+    useLazyModal();
+  const { isActionsFetched, checkHasCompleted, completeAction } = useActions();
+  const gb = useGrowthBook();
+
+  const featureEnabled = useMemo(() => {
+    // gb.isOn triggers enrollment in the experiment
+    // to avoid noise in the data, only call this if we are in the extension
+    if (isExtension)
+      return gb.isOn(Features.EngagementLoopJuly2023Companion.id);
```
capJavert, line 50: I find these one line `if` without brackets error prone, especially when function has more then one line because condition can easily be missed if new lines are added. cc @dailydotdev/web-team WDYT? We can maybe even add a lint rule for this so its auto formatted.

packages/shared/src/hooks/useCompanionTrigger.tsx
```diff
@@ -0,0 +1,167 @@
...
+  const articleClickHandler = useCallback(
+    async (
+      e: React.MouseEvent,
+      { post, index, row, column }: Partial<CompanionTriggerProps> = {},
+    ) => {
+      if (
+        !isExtension ||
+        !featureEnabled ||
+        alreadyCompleted ||
+        contentScriptGranted
```
capJavert, line 130: You can pull this condition outside of `useCallback` and call it something meaningful. Eg. `shouldOverwritePostClickHandler`. It will make it easier to someone to understand what it checks and also lower the amount dependencies for `useCallback`.

capJavert (commented):
Few ideas to make the code more readable but nothing blocking. Nice work. 

@DragosIorgulescu just to check, events are now all sent, you managed to fix the read article issue?

packages/shared/src/hooks/useCompanionTrigger.tsx
capJavert: This one https://eslint.org/docs/latest/rules/curly

capJavert (commented):
I do have additional comment about how loading of the modal looks.

It looks a little bit janky/broken while assets load and it would be even more noticeable for people with slower connection and since modal is once off I think we need to make sure it looks good?

Some kind of progressive loading with predefined width, cover image etc comes to mind? Thoughts?

https://github.com/dailydotdev/apps/assets/9803078/1db587cc-9fea-423b-957a-f1fd20aedbf2

packages/shared/src/hooks/useCompanionTrigger.tsx
rebelchris: Yeah agreed, it actually has happened before where someone would accidentally add another element and it breaks the statement.
Good point on the rule :) (I'll add a ticket)

packages/shared/src/hooks/useCompanionTrigger.tsx
DragosIorgulescu: 👍  thank you for pointing that out. But if it does not matter, isn't it better to be more explicit?

capJavert: When I think about it, with condition like this there could be a case where actions fetch request does not resolve and user closes the modal which would mean action will not be completed and user will see the modal once again, which is not per spec. 

To your point thought `isActionsFetched` is more important when reading since you probably want to wait server state before deciding action is completed. So if that was the case we would use it 👍

packages/shared/src/components/modals/CompanionModal.tsx
```diff
@@ -39,105 +41,108 @@ export default function CompanionModal({
...
-          loop
-          autoPlay
-        >
-          <source
-            src={cloudinary.companion.backgroundVideoMobile}
-            type="video/mp4"
-          />
-        </video>
-      )}
+      <Modal.Body className={classNames('p-0', { 'min-h-[640px]': !loaded })}>
```
rebelchris, line 44: Can't we rather mock the video size, the body should auto adjust.
Alternative we can wait before render?

In anycase we don't use px values ever, always use REM when you need.
(Expection is a 1px line)

packages/shared/src/hooks/useCompanionTrigger.tsx
DragosIorgulescu: I think I tested this and if `isActionFetched` is `undefined`, `completeAction` fails as well

capJavert: Please double check 🙏 if it does not work thats a bug.

DragosIorgulescu:
@capJavert 
>  @DragosIorgulescu just to check, events are now all sent, you managed to fix the read article issue?

yes, these are being emitted now

> It looks a little bit janky/broken while assets load and it would be even more noticeable for people with slower connection and since modal is once off I think we need to make sure it looks good?

here is how the loading state looks like now on a slow connection:

companion-loading-state

rebelchris:
@DragosIorgulescu That loading state looks great 🙏

packages/shared/src/hooks/useCompanionTrigger.tsx
DragosIorgulescu: Looks like this works fine, I've updated and cleaned it up a little 👍

capJavert (approved):

PR #2044, opened 2023-08-15
fix: minor tailwind issues

rebelchris:
- Found a couple more tailwind issues.

I ran through all screens and modals I could test locally, and didn't see anything other weirdly out of place.
However do notify me as soon as you see anything odd *(or unsure off)

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done

### DIFF
--- a/packages/shared/src/components/fields/MarkdownInput/index.tsx
+++ b/packages/shared/src/components/fields/MarkdownInput/index.tsx
@@ -251,8 +251,10 @@ function MarkdownInput(
               type="button"
               buttonSize={actionButtonSizes}
               className={classNames(
-                'btn-tertiary',
-                uploadingCount && 'text-theme-color-cabbage',
+                'btn-tertiary font-normal',
+                uploadingCount
+                  ? 'text-theme-color-cabbage'
+                  : 'text-theme-label-quaternary',
               )}
               icon={icon}
               onClick={() => uploadRef?.current?.click()}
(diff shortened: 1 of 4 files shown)


omBratteng (approved):

packages/shared/src/components/fields/MarkdownInput/index.tsx
```diff
@@ -278,7 +280,7 @@ function MarkdownInput(
             wrapper={(children) => (
               <span
                 className={classNames(
-                  'grid grid-cols-3 gap-3',
+                  'flex gap-3',
```
sshanzel, line 283: Just a question, why did we end up moving to `flex gap`, is the `grid grid-cols-n` doesn't work as how it used to anymore?

sshanzel (approved):
Looks good to me! 🚀

packages/shared/src/components/fields/MarkdownInput/index.tsx
rebelchris: I think this was a fault to be honest.
Don't know why that was chosen, but on one screen we only render 1 item, which made it align super weird (in the first of 3 columns)

Flex does exactly this, but automatic 😅
Love to hear if it has a particular use-case I'm unaware off

sshanzel: Nothing really. Was just curious about the change if it was tailwind update related. Flex does the job better anyways, and won't limit us to the number of elements. Way to go! 🚀

PR #2046, opened 2023-08-15
feat: search experiment feed header

sshanzel:
## Changes

Previews:
- There are a lot of changes on this PR though it was only revolving around the feed header change.
- Due to the large differences between the header for control and v1 values for the experiment, we have decided to introduce separate components to render based on version rather than having numerous conditions on the classes based on the said experiment.
- The changes related to `isSearchOn` is just reverting some of the changes in the intermediate branch, but they are basically the legacy code.
- Moved the `FeedFilters` modal to be under the `LazyModal` usage.
- The dedicated search page would still render based on the flag. Both variants will be under the same URL.

Screenshot 2023-08-15 at 6 16 39 PM
Screenshot 2023-08-15 at 5 46 42 PM
Screenshot 2023-08-15 at 5 46 32 PM
Screenshot 2023-08-15 at 5 46 13 PM

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1639 #done


packages/shared/src/components/feeds/FeedContainer.tsx
```diff
@@ -122,10 +139,33 @@ export const FeedContainer = ({
...
+                  field: 'w-full',
+                }}
+                showProgress={false}
+                onSubmit={onSearch}
+              />
+              {besideSearch}
+            </FlexRow>
+          )}
+          {isV1Search && (
+            <FlexRow className="flex flex-1 mt-4">
```
rebelchris, line 156: Still not a fan of these 😂
Why don't we simply pass the row, these are super prone to changes down the line and just don't really make sense at this point.

packages/shared/src/components/filters/MyFeedHeading.tsx
```diff
@@ -34,21 +42,42 @@ function MyFeedHeading({
...
+      wrapper: 'mr-auto z-modal',
+    },
+    message: filterAlertMessage,
+    placement:
+      sidebarRendered && searchVersion === SearchExperiment.Control
+        ? AlertPlacement.Right
+        : AlertPlacement.Bottom,
+  };
+
+  if (searchVersion === SearchExperiment.V1) {
```
rebelchris, line 64: We do render for both right?

packages/shared/src/components/sidebar/DiscoverSection.tsx
```diff
@@ -57,16 +61,19 @@ export function DiscoverSection({
         </span>
       ),
     },
-    {
+  ];
+
+  if (searchValue !== SearchExperiment.V1) {
```
rebelchris, line 66: Should we not say if it's explicitly control?
What happens if we introduce V2 for instance

packages/shared/src/hooks/feed/useFeedName.ts
```diff
@@ -0,0 +1,34 @@
...
+export const useFeedName = ({
+  feedName,
+  isSearchOn,
+}: UseFeedNameProps): UseFeedName => {
+  const searchVersion = useFeature(Features.Search);
+
+  return {
+    isUpvoted:
+      searchVersion === SearchExperiment.Control
+        ? !isSearchOn && feedName === 'upvoted'
```
rebelchris, line 27: Can't we rather abstract the logic above?
To say isSearchOn is false//null when it's not control?

rebelchris (changes requested):
Loads of logic 🤯
Have some generic feedback, can we somehow maybe abstract logic so we don't have so much conditional stuff?

Else I'm already praying for a removal task 😂

packages/shared/src/components/filters/MyFeedHeading.tsx
sshanzel: Yes, we have the default that starts at line 77.

packages/shared/src/components/sidebar/DiscoverSection.tsx
sshanzel: Makes sense. Let's do that.

packages/shared/src/hooks/feed/useFeedName.ts
sshanzel: That `isSearchOn` is coming from way up the tree based on a state. Can you provide a simple example of how you envision it to be?

packages/shared/src/components/feeds/FeedContainer.tsx
rebelchris: Yes correct :) Think we should just remove them to be honest.

sshanzel: Amazing. Should be removed by now.

sshanzel:
@rebelchris I've now added the Cloudinary links. Should be ready for final review 🚢

packages/shared/src/lib/image.ts
```diff
@@ -1,4 +1,14 @@
 export const cloudinary = {
+  feed: {
+    bg: {
+      laptop:
+        'https://daily-now-res.cloudinary.com/image/upload/v1665224467/public/Desktop-ask-bg.png',
```
rebelchris, line 5: All of these seem to be wrong, missing the f_auto can you ensure with design please?
Or I can help get them if not

rebelchris (changes requested):
One small request, otherwise looks great to me

packages/shared/src/lib/image.ts
sshanzel: They shared one with it, but it wasn't working, not sure if the files are not yet public. Should be found here: https://dailydotdev.slack.com/archives/C05EM7ZFQMQ/p1692271488033119?thread_ts=1692265064.294219&cid=C05EM7ZFQMQ

sshanzel:
Merged so we can now test on preview. Will create a ticket for updating cloudinary links.

PR #2055, opened 2023-08-18
fix: types and managing of stream data

sshanzel:
## Changes
- Updated the types to follow what was explained in the DR.
- Simplified state management.
- Moved away from using a `useEffect` for triggering the prompt instead of executing right there and then.
- Introduced a SearchContainer to be used for mobile app's modal version.
- Updated the stream types handlers.
- Should be ready for what is expected in the DR.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1653 #done

### DIFF
--- a/packages/shared/src/components/search/SearchContainer.tsx
+++ b/packages/shared/src/components/search/SearchContainer.tsx
@@ -13,7 +13,7 @@ export function SearchContainer({
 }: SearchContainerProps): React.ReactElement {
   return (
     <main className="m-auto w-full">
-      <div className="grid grid-cols-1 laptop:grid-cols-3 gap-y-6 pt-8 m-auto max-w-screen-laptop">
+      <div className="flex flex-wrap gap-y-6 pt-8 m-auto max-w-screen-laptop">
         <main className="flex flex-col flex-1 col-span-2 px-4 laptop:px-8">
           <SearchBar onSubmit={onSubmit} />
         </main>
--- a/packages/shared/src/components/search/SearchContainer.tsx
+++ b/packages/shared/src/components/search/SearchContainer.tsx
@@ -1,0 +1,25 @@
+import React, { ReactNode } from 'react';
+import { SearchBar } from './SearchBar';
+import { SearchFeedback } from './SearchFeedback';
+
+interface SearchContainerProps {
+  children: ReactNode;
+  onSubmit(event: React.MouseEvent, value: string): void;
+}
+
+export function SearchContainer({
+  children,
+  onSubmit,
+}: SearchContainerProps): React.ReactElement {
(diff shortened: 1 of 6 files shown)


packages/shared/src/hooks/useChat.ts
```diff
@@ -21,130 +27,136 @@ enum UseChatMessageType {
...
+        return;
+      }
 
-    return () => {
-      endStream();
-    };
-  }, [prompt]);
+      if (endStreamRef.current) {
+        endStreamRef.current();
+      }
```
capJavert, line 140: My only concern with this approach If hook unmounts the stream will be left open.

capJavert: We can probably clean it in empty useEffect with only cleanup function.

sshanzel: Oh, right. Yes, let's open a `useEffect` to clean it up.

packages/shared/src/components/search/SearchContainer.tsx
```diff
@@ -0,0 +1,25 @@
...
+  onSubmit(event: React.MouseEvent, value: string): void;
+}
+
+export function SearchContainer({
+  children,
+  onSubmit,
+}: SearchContainerProps): React.ReactElement {
+  return (
+    <main className="m-auto w-full">
+      <div className="grid grid-cols-1 laptop:grid-cols-3 gap-y-6 pt-8 m-auto max-w-screen-laptop">
```
rebelchris, line 16: Is the grid necessary here?
If not prefer flex in case things dynamic render.

rebelchris (approved):
Nothing that seems off to me.
But let's wait for @capJavert response before merging.

Just wondering does this already cover the loader changes based on state or will it be seperate?

capJavert (approved):
Looks good, I guess any changes needed can be done when we actually have real implementation 🚀

packages/shared/src/components/search/SearchContainer.tsx
sshanzel: Oh, this was the existing classes when I moved this container. Let me check if it is necessary.

sshanzel:
> Looks good, I guess any changes needed can be done when we actually have real implementation 🚀

I agree. Things might still change once the actual integration becomes available. Let's see how it goes 🚢

packages/shared/src/components/search/SearchContainer.tsx
sshanzel: Seems okay with flex 🥳

PR #2058, opened 2023-08-18
fix: upload toast if verify fails early

rebelchris:
- If our JS based verification failed we didn't show anything.
- So if user tries to upload wrong type of file he would never know.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1531 #done


rebelchris:
@sshanzel Would like a review from you especially 🙏

TomDowling (approved):

sshanzel (approved):
Thanks for utilizing the same message from a single source 🚢

sshanzel:
One thing though, the `verifyFile` function is used in a loop, so if the user tried to upload two files that are wrong, the toast will trigger twice. Nothing blocking, but just wanted it to be known.

rebelchris:
@sshanzel Did see that indeed, not quite sure how you can upload more, I think from what I've seen we limit the picker to one?
But regardless the two toast is acceptable for now.

sshanzel:
Dragging multiple files at the same time allows multiple uploads. Similar to GH upload. Probably the picker must be updated to allow multiple files.

> But regardless the two toast is acceptable for now.

I agree, shouldn't be too noticeable making it reset due to numerous toast.

PR #2067, opened 2023-08-22
Fix loading layout issues on public squad posts

DragosIorgulescu:
## Changes

Fixes the post loading skeleton layout
- make use of the post source to determine whether this is a public squad post
- pass down the post source from the feed rather than waiting for the post record to load

## Preview

modal-skeleton-layout

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1619 #done


packages/shared/src/components/Feed.tsx
```diff
@@ -424,6 +424,7 @@ export default function Feed<T>({
           onPreviousPost={onPrevious}
           onNextPost={onNext}
           postPosition={postPosition}
+          postSource={selectedPost.source}
```
DragosIorgulescu, line 427: passing down the source from the feed so we can use it to determine the modal loading layout

packages/shared/src/components/modals/SharePostModal.tsx
```diff
@@ -34,14 +37,15 @@ export default function PostModal({
     offset: showArticleOnboarding ? ONBOARDING_OFFSET : 0,
   });
   const containerClass = 'post-content';
-  const isPublicSquad = isSourcePublicSquad(post?.source);
+  const isPublicSquad = isSourcePublicSquad(postSource);
```
DragosIorgulescu, line 40: using `post.source` causes jittering UI since the modal size depends on this. `postSource` is present in props before the `post` record is loaded for this reason

packages/shared/src/components/Feed.tsx
sshanzel: I think we should pass the whole post instead. From there we will use the `source` property. We might even be able to leverage the `post.type` on some cases.

rebelchris (approved):
Overall looks good, do agree with Lee, since we always know the post already for the modal, can't we simply pass it down?

@sshanzel Do you remember why we didn't do that in the first place?

rebelchris (commented):
Also noticed in the video that there is still a layout shift on the right where the top buttons are.
Can you include that space as well for the modal loader?

DragosIorgulescu:
@rebelchris @sshanzel 

> I think it is probably due to small iterations where we didn't need it at first and now, looking back at it, seems we can leverage the existing data.

It may also be that the `post` record coming from the feed does not include all the data it needs to be rendered in the post page/modal. (Since there is a query for the post record [in the modal components as well](https://github.com/dailydotdev/apps/blob/ba1def774960be7421baf7157263923f9baece94/packages/shared/src/components/modals/ArticlePostModal.tsx#L33))

I can pass the whole post object down, but this will create some duplication in the modal components since we will have a `post` coming from props, and a `post` variable defined for the query.

sshanzel:
> I can pass the whole post object down, but this will create some duplication in the modal components (SharePostModal & ArticlePostModal) since we will have a post coming from props, and a post variable defined for the query.

I think it is possible to set it as the `initialData` from the `useQuery`, that way we would still be using the same object.

DragosIorgulescu:
After making the change to use the `post` object passed from the feed, here is how this now looks:

modal-no-skeleton

sshanzel:
> After making the change to use the post object passed from the feed, here is how this now looks:

That looks alright. cc @rebelchris wdyt?

rebelchris:
Looks good to me on the loaders indeed

PR #2070, opened 2023-08-23
feat: stream integration

sshanzel:
## Changes
- Usage of `EventSource` for receiving the stream of data relating to the search query.
- Updated the Cloudinary links.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1653 #done


packages/shared/src/graphql/search.ts
```diff
@@ -127,3 +127,16 @@ export const updateSearchData = (
 
   return updated;
 };
+
+export const searchQueryUrl = 'https://api.daily.dev/search/query';
```
rebelchris, line 131: We should make this a dynamic URL based on the API you are at probably

packages/shared/src/hooks/useChat.ts
```diff
@@ -130,26 +133,28 @@ export const useChat = ({ id: idFromProps }: UseChatProps): UseChat => {
...
+      if (sourceRef.current?.OPEN) {
+        sourceRef.current.close();
       }
 
       setSearchQuery(undefined);
-      // eslint-disable-next-line @typescript-eslint/no-use-before-define
-      endStreamRef.current = mockChatStream({ onMessage });
+      const source = await sendSearchQuery(value, accessToken.token);
+      source.onmessage = onMessage;
+      source.onerror = () => source.close();
```
rebelchris, line 148: Just keep in mind we need to show a error as well.
You can log it seperate if you want.

packages/shared/src/hooks/useChat.ts
```diff
@@ -130,26 +133,28 @@ export const useChat = ({ id: idFromProps }: UseChatProps): UseChat => {
...
-      if (endStreamRef.current) {
-        endStreamRef.current();
+      if (sourceRef.current?.OPEN) {
+        sourceRef.current.close();
       }
 
       setSearchQuery(undefined);
-      // eslint-disable-next-line @typescript-eslint/no-use-before-define
-      endStreamRef.current = mockChatStream({ onMessage });
+      const source = await sendSearchQuery(value, accessToken.token);
```
rebelchris, line 146: Why do you pass it, was wondering if it's not easier to abstract it into the sendSearchQuery?

packages/shared/src/lib/featureManagement.ts
```diff
@@ -30,7 +30,7 @@ export class Features<T extends FeatureValue = string> {
     ],
   );
 
-  static readonly Search = new Features('search', SearchExperiment.Control);
+  static readonly Search = new Features('search', SearchExperiment.V1);
```
rebelchris, line 33: Do write a TODO or something, just so we 100% don't forget this 😅

rebelchris (changes requested):
Small changes only

packages/shared/src/lib/featureManagement.ts
sshanzel: Oh, right. We will have to keep this back to Control.

packages/shared/src/hooks/useChat.ts
sshanzel: Yep, we should have the [ticket](https://dailydotdev.atlassian.net/jira/software/projects/WT/boards/11?selectedIssue=WT-1678) dedicated for it.

packages/shared/src/graphql/search.ts
sshanzel: 100% agree. Let's pull the `apiUrl` we have.

packages/shared/src/hooks/useChat.ts
sshanzel: Hmm. The only concern is, we are only trying to add callbacks for now. Though if there will be anything specific to Search, then I think we can have it inside the `sendSearchQuery`. Wdyt?

rebelchris: Not sure I understand, this function is the only one that needs specific token right?
it's dedicated to search already.

sshanzel: Oh, my bad. Misunderstood. You mean the token? I thought it would be based on the user's JWT token. Which is coming from the AuthContext. So the function won't be able to use the hook as it is a plain function.

rebelchris: ah yep, you are right 👍

rebelchris (approved):

PR #2077, opened 2023-08-24
fix: available actions on mobile only

sshanzel:
## Changes
- The mobile Search page will display the suggestions and search history.
- Keep the value of input based on the design (I think this is because we don't display anywhere the user's query).
- The Search component should display the Search mobile popup on click.
- Utilize search suggestion query.
- Add skeleton loaders for history and search suggestion.
- Support running query from redirection through query params.

Preview:

https://github.com/dailydotdev/apps/assets/13744167/6dc5c4f1-12c1-465a-a01b-60eeab49e3da

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1654 #done


packages/shared/src/components/feeds/FeedContainer.tsx
```diff
@@ -153,7 +160,9 @@ export const FeedContainer = ({
           )}
           {isV1Search && (
             <span className="flex flex-row flex-1 mt-4">
-              <SearchBarSuggestionList className="hidden tablet:flex overflow-hidden flex-1 mr-3" />
+              {isTabletAbove && (
+                <SearchBarSuggestionList className="hidden tablet:flex overflow-hidden flex-1 mr-3" />
```
rebelchris, line 164: Why was the CSS only solution not good enough?

packages/shared/src/components/search/SearchBarInput.tsx
```diff
@@ -108,40 +145,37 @@ function SearchBarInputComponent(
...
+            externalOnFocus?.(event);
+          }}
+          onBlur={(event) => {
+            onBlur();
+            externalOnBlur?.(event);
+          }}
+          onClick={(event) => {
+            onInputClick();
+            externalOnClick?.(event);
+          }}
```
rebelchris, line 167: Are these actually needed?
I think it's something or field handles no?

rebelchris (changes requested):
Mainly wondering why we need the JS evaluation for these items?

packages/shared/src/components/search/SearchBarInput.tsx
sshanzel: I think this component was based on the previous Search component. With that, this is to support external and internal handlers.

packages/shared/src/components/feeds/FeedContainer.tsx
sshanzel: There are some that require JS evaluation. However keep note that this PR is still WIP. A lot of things here are subject to change in my upcoming commits.

packages/shared/src/components/search/MobileSearch.tsx
```diff
@@ -0,0 +1,104 @@
...
+  onClose,
+  onSubmit,
+}: MobileSearchProps): ReactElement {
+  const { user } = useAuthContext();
+  const [input, setInput] = useState(initialInput);
+  const inputRef = useRef<HTMLInputElement>();
+  const { data, isLoading } = useQuery(
+    generateQueryKey(RequestKey.SearchHistory, user),
+    getSearchSuggestions,
+  );
```
rebelchris, line 41: Should we not add this more top level as it's reused by two?

packages/shared/src/components/search/MobileSearch.tsx
```diff
@@ -0,0 +1,104 @@
...
+    event.preventDefault();
+    onSubmit(event, input);
+  };
+
+  useEffect(() => {
+    inputRef.current?.focus();
+  }, []);
+
+  return (
+    <InteractivePopup
```
rebelchris, line 53: Why did we not go for a modal, or is this the way we always do the overlapping ones?

packages/shared/src/components/search/MobileSearch.tsx
```diff
@@ -0,0 +1,104 @@
...
+          type="text"
+          value={input}
+          ref={inputRef}
+          placeholder="Ask anything..."
+          onInput={(e) => setInput(e.currentTarget.value)}
+        />
+      </form>
+      <div className="flex flex-col p-4">
+        <SearchBarSuggestionList
+          suggestions={data?.map(({ question }) => ({ prompt: question }))}
```
rebelchris, line 93: Maybe alternative we can just load the suggestion inside this component?

packages/shared/src/components/search/SearchBarSuggestionList.tsx
```diff
@@ -3,30 +3,46 @@ import classNames from 'classnames';
...
+import { ElementPlaceholder } from '../ElementPlaceholder';
 
 interface SearchBarSuggestionListProps {
   className?: string;
-  suggestions?: SearchSession[];
+  isLoading?: boolean;
+  suggestions?: Partial<SearchSession>[];
 }
 
+const Pill = classed(ElementPlaceholder, 'h-8 rounded-12 w-auto');
```
rebelchris, line 16: You defined it below as utils, can't you use those?

packages/shared/src/components/tooltips/InteractivePopup.tsx
```diff
@@ -50,7 +50,7 @@ function InteractivePopup({
     <Portal>
       <div
         className={classNames(
-          'fixed z-3 bg-theme-bg-primary rounded-16 overflow-hidden',
+          'fixed z-modal bg-theme-bg-primary rounded-16 overflow-hidden',
```
rebelchris, line 53: We need to tripple check this doesn't affect any other use-cases please 🙏

rebelchris (changes requested):
Massive amount of work!
ONly couple little things, overall looks good.

packages/shared/src/components/search/MobileSearch.tsx
sshanzel: Good point! Let's reuse this part.

packages/shared/src/components/search/MobileSearch.tsx
sshanzel: We currently don't have a modal configuration that has the same header. So thought of using this as it is an experiment and might change soon enough.

packages/shared/src/components/search/MobileSearch.tsx
sshanzel: This component is used for both history and suggestions. Supplying the `suggestions` prop would allow it to work on both.

packages/shared/src/components/search/SearchBarSuggestionList.tsx
sshanzel: Oh, copy-paste issue. Let me move this one.

packages/shared/src/components/tooltips/InteractivePopup.tsx
sshanzel: Agreed. Will further check this one.

packages/shared/src/components/search/MobileSearch.tsx
rebelchris: Ah right they have different suggestions :)

packages/shared/src/components/tooltips/InteractivePopup.tsx
sshanzel: Introduced a new one to be less than the modal to ensure no collision will happen.

packages/shared/src/components/search/SearchBarInput.tsx
```diff
@@ -81,43 +92,67 @@ function SearchBarInputComponent(
...
   };
 
   const seeSearchHistory = (event: MouseEvent): void => {
     event.stopPropagation();
   };
 
+  const onInputClick = () => {
+    if (isTabletAbove || isMobileOpen || !shouldShowPopup) return null;
+
+    if (!user) return showLogin('search input');
```
rebelchris, line 116: This should be a value of AuthTriggers, I don't know who keeps adding the string expection, but it's not expected 😅

rebelchris (changes requested):
One small fix :)

packages/shared/src/components/search/SearchBarInput.tsx
sshanzel: Oh, right. That is correct.  Let me update this one. Probably due to some forgotten TODOs.

rebelchris (approved):
Let's give it a go 🙏

PR #2084, opened 2023-08-25
fix: border bottom on history page

sshanzel:
## Changes
- It won't be possible to have the border from the inside out to the main layout.
- Feels like a hack though to stamp a div on top of everything.
- If you have any suggestions, feel free to shoot.

Previews:

Screenshot 2023-08-25 at 2 10 39 PM
Screenshot 2023-08-25 at 2 10 30 PM
Screenshot 2023-08-25 at 2 10 21 PM

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1696 #done

### DIFF
--- a/packages/shared/src/components/tabs/TabContainer.tsx
+++ b/packages/shared/src/components/tabs/TabContainer.tsx
@@ -44,6 +44,7 @@ export interface TabContainerProps<T extends string> {
   showHeader?: boolean;
   controlledActive?: string;
   tabListProps?: Pick<TabListProps, 'className'>;
+  showBorder?: boolean;
 }
 
 export function TabContainer<T extends string = string>({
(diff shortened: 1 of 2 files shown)


rebelchris (commented):
Another idea is to simply show the left-right border in tablet mode?
That would also look neat and maybe less hacky.

Although this proposed solution is not as bad.

sshanzel:
That can work too. For now, we can probably stick with this one since the design team mentioned they want it to look like the profile page.

rebelchris:
Sure that's fine for now, how do we handle it on profile page then?

sshanzel:
> Sure that's fine for now, how do we handle it on profile page then?

Seems like a hack more than what we did here 😂 

Screenshot 2023-08-25 at 2 54 58 PM

rebelchris (approved):
Let's go, maybe we can add backlog ticket to support this better down the line.
Especially if there is more then one implementation

sshanzel:
> Let's go, maybe we can add backlog ticket to support this better down the line.
Especially if there is more then one implementation

Agreed. I can add a ticket after merging 🚀

PR #2095, opened 2023-08-25
feat: add language & copy button to code search results

DragosIorgulescu:
## Changes

- add heading section to each code block with the language & copy button

## Events

_no new/updated events_

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1643 #done


DragosIorgulescu:
**Preview**

search-results-copy

I still need to clean up the UI a little bit to match the design specs

packages/shared/src/components/RenderMarkdown.tsx
```diff
@@ -137,23 +141,46 @@ const RenderMarkdown = ({
...
-            >
-              {String(children).replace(replaceNewLineRegex, '')}
-            </SyntaxHighlighterAsync>
-          ) : (
-            <code {...props} className={codeClassName}>
-              {children}
-            </code>
+          return (
+            <>
+              {language && (
```
rebelchris, line 146: I think we should still render the option to copy even if we don't know the language

packages/shared/src/hooks/useCopyLink.ts
DragosIorgulescu, line 1: This file could probably be renamed to `useCopy` or `useClipboard`, thoughts?

packages/shared/src/components/RenderMarkdown.tsx
```diff
@@ -137,23 +141,46 @@ const RenderMarkdown = ({
...
+                <div className="flex justify-between py-3 px-6 bg-theme-active align-center">
+                  <span>{language}</span>
+                  <Button
```
rebelchris, line 149: Probably need: `btnSize: Xsmall` here

packages/shared/src/components/RenderMarkdown.tsx
```diff
@@ -137,23 +141,46 @@ const RenderMarkdown = ({
...
-              {String(children).replace(replaceNewLineRegex, '')}
-            </SyntaxHighlighterAsync>
-          ) : (
-            <code {...props} className={codeClassName}>
-              {children}
-            </code>
+          return (
+            <>
+              {language && (
+                <div className="flex justify-between py-3 px-6 bg-theme-active align-center">
```
rebelchris, line 147: Probably best to add text-teriary here as well to ensure color of the language
And not sure if it's a different size

rebelchris: Think also the padding is off (seems to big)

packages/shared/src/hooks/useCopyLink.ts
```diff
@@ -39,3 +40,20 @@ export function useCopyLink(
...
   return [copying, copy];
 }
+
+export function useCopyText(text?: string): [boolean, CopyNotifyFunction] {
+  const [copying, setCopying] = useState(false);
+  const { displayToast } = useToastNotification();
+
+  const copy: CopyNotifyFunction = async (props = {}) => {
+    await navigator.clipboard.writeText(props.textToCopy || text);
+    displayToast(props.message || defaultMessage, props);
```
rebelchris, line 50: This seems off, as the toast would say "Copied link"

I think we need to abstract the useCopyLink a bit to handle generic toast rather if that makes sense

rebelchris (changes requested):
Nice work, overall looks good.
Some minor cleanup we can do.

packages/shared/src/hooks/useCopyLink.ts
sshanzel: Agree with that. A bit vague currently. Just might require a lot of change on existing implementations.

sshanzel (commented):
Great work! Have nothing to add. Chris have covered everything already 🔭

packages/shared/src/hooks/useCopyLink.ts
capJavert: We can always introduce `useCopy` and make `useCopyLink` use it under the hood and then phase out the old one as time goes.

DragosIorgulescu: I don't personally mind having 2 different hooks, since it looks like `useCopyLink` does something more with the URL. For now I renamed the file to `useCopy` and kept both `useCopyLink` and `useCopyText` in it.

packages/shared/src/components/RenderMarkdown.tsx
DragosIorgulescu: good point 👍

packages/shared/src/hooks/useCopyLink.ts
DragosIorgulescu: I updated the default message and kept both hooks for now as it feels cleaner. I can consolidate these if you prefer though

packages/shared/src/components/RenderMarkdown.tsx
DragosIorgulescu: 👍

packages/shared/src/components/RenderMarkdown.tsx
```diff
@@ -137,23 +141,46 @@ const RenderMarkdown = ({
...
+            <>
+              <div className="flex justify-between py-2 px-5 bg-theme-active align-center">
+                <span className="inline leading-8 text-theme-label-tertiary">
+                  {language}
+                </span>
+                <Button
+                  className="btn-tertiary"
+                  icon={<CopyIcon />}
+                  disabled={copying}
+                  buttonSize={ButtonSize.Small}
```
rebelchris, line 154: Was small ok?
Think the design states `XSmall` it should be a 20x20 button size.

rebelchris (approved):
That works for now.
You verified the styling on the box matches design?

packages/shared/src/components/RenderMarkdown.tsx
DragosIorgulescu: Felt too small, but you're right, updated as per spec

DragosIorgulescu:
> You verified the styling on the box matches design?

Double-checking with design on some of the colors, but other than that this should match the design spec

rebelchris (approved):
Good to go

PR #2100, opened 2023-08-28
fix: companion modal redirect on activation + other small fixes

DragosIorgulescu:
## Changes

- redirect to the post link when accepting/denying the extension permissions (was redirecting to the companion blog post page)
- fix layout issues
- add close button as per spec
- update icons as per spec
- update secondary button color as per spec

## Preview

Screenshot 2023-08-28 at 12 26 17

## Events

_no events changes_

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1729 #done

### DIFF
--- a/packages/extension/src/companion/CompanionPermission.tsx
+++ b/packages/extension/src/companion/CompanionPermission.tsx
@@ -38,7 +38,7 @@ const CompanionPermissionComponent = (
         </p>
         <Button
           className="mt-1 w-[12.5rem] btn btn-primary"
-          onClick={requestContentScripts}
+          onClick={() => requestContentScripts()}
           buttonSize={ButtonSize.Small}
         >
           {button}
--- a/packages/shared/src/components/modals/CompanionModal.tsx
+++ b/packages/shared/src/components/modals/CompanionModal.tsx
@@ -79,7 +79,7 @@ export default function CompanionModal({
         <Button
           icon={<MiniCloseIcon />}
           className="btn-secondary"
-          buttonSize={ButtonSize.XSmall}
+          buttonSize={ButtonSize.Small}
           onClick={props.onRequestClose}
         />
       </div>
@@ -89,7 +89,7 @@ export default function CompanionModal({
           Level up your reading experience
         </h2>
 
-        <p className="pb-6 text-xl text-center text-theme-label-tertiary">
+        <p className="pb-6 text-center typo-title3 text-theme-label-tertiary">
           AI generated summaries, upvoting, commenting, sharing, and more from
           anywhere on the web
         </p>
@@ -115,7 +115,7 @@ export default function CompanionModal({
         <div className="flex flex-col flex-1 justify-between">
           <div className="hidden tablet:flex flex-col items-center">
             <div className="flex items-center pb-4">
-              <div className="inline-block p-2 mr-2 bg-theme-active rounded-16">
+              <div className="inline-block p-2 mr-4 rounded-16 bg-theme-overlay-cabbage">
                 <UnreadIcon
                   size={IconSize.Medium}
                   secondary
@@ -123,31 +123,31 @@ export default function CompanionModal({
                 />
               </div>
 
-              <span>Get a TLDR on the original post</span>
+              <span className="typo-body">Get a TLDR on the original post</span>
             </div>
             <div className="hidden tablet:flex items-center pb-4">
-              <div className="inline-block p-2 mr-2 bg-theme-active rounded-16">
+              <div className="inline-block p-2 mr-4 rounded-16 bg-theme-overlay-blueCheese">
                 <CommentIcon
                   size={IconSize.Medium}
                   secondary
                   className="text-theme-color-blueCheese"
                 />
               </div>
-              <span>Discuss from everywhere</span>
+              <span className="typo-body">Discuss from everywhere</span>
             </div>
             <div className="hidden tablet:flex items-center pb-4">
-              <div className="inline-block p-2 mr-2 bg-theme-active rounded-16">
+              <div className="inline-block p-2 mr-4 rounded-16 bg-theme-overlay-bun">
                 <BookmarkIcon
                   secondary
                   size={IconSize.Medium}
                   className="text-theme-color-bun"
                 />
               </div>
-              <span>Bookmark with a single click</span>
+              <span className="typo-body">Bookmark with a single click</span>
             </div>
           </div>
 
-          <span className="text-center text-theme-label-tertiary typo-footnote">
+          <span className="text-center text-theme-label-tertiary typo-subhead">
             Activating the widget requires extra permissions <br />
             (Give it a try, you can always disable it later)
           </span>
--- a/packages/shared/src/components/modals/CompanionModal.tsx
+++ b/packages/shared/src/components/modals/CompanionModal.tsx
@@ -14,6 +14,7 @@ import { cloudinary } from '../../lib/image';
 
 type Props = LazyModalCommonProps & {
   url: string;
+  onClose: () => void;
   onReadArticleClick: (any) => void;
   onActivateCompanion: (redirectUrl?: string) => void;
 };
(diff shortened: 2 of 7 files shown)


packages/extension/src/companion/CompanionPermission.tsx
```diff
@@ -38,7 +38,7 @@ const CompanionPermissionComponent = (
         </p>
         <Button
           className="mt-1 w-[12.5rem] btn btn-primary"
-          onClick={requestContentScripts}
+          onClick={() => requestContentScripts()}
```
rebelchris, line 41: This change shouldn't be needed

packages/shared/src/components/modals/CompanionModal.tsx
```diff
@@ -41,110 +42,117 @@ export default function CompanionModal({
       kind={Modal.Kind.FlexibleCenter}
       size={size}
     >
-      <Modal.Body className="p-0">
```
DragosIorgulescu, line 44: For some reason when running this locally the `p-0` is respected, but when running in production it is ignored? I removed it entirely for now as it is not needed, hopefully this should fix the scroll/padding issue.

packages/shared/src/components/modals/CompanionModal.tsx
```diff
@@ -41,110 +42,117 @@ export default function CompanionModal({
...
-                </div>
-                <span>Bookmark with a single click</span>
+              <span>Discuss from everywhere</span>
+            </div>
+            <div className="hidden tablet:flex items-center pb-4">
+              <div className="inline-block p-2 mr-2 bg-theme-active rounded-16">
+                <BookmarkIcon
+                  secondary
+                  size={IconSize.Medium}
+                  className="text-theme-color-avocado"
```
rebelchris, line 143: This should be orange color

rebelchris (changes requested):
Two minor things, Will also take it for a test locally :)

packages/shared/src/components/modals/CompanionModal.tsx
rebelchris: It could be the tailwind order on production, but ideally should match locally.
Have you run pnpm install latelty? (After the tailwind upgrade)

packages/extension/src/companion/CompanionPermission.tsx
DragosIorgulescu: the `requestContentScripts` function now accepts a `skipRedirect` parameter as an argument, so passing it directly causes a TS error since `onClick` expects an event instance to be the argument of the function being passed in

packages/shared/src/hooks/useCompanionTrigger.tsx
```diff
@@ -71,33 +71,53 @@ export default function useCompanionTrigger(
...
   const openModal = useCallback(
     (data?: CompanionTriggerProps) => {
       openLazyModal({
         type: LazyModal.CompanionModal,
         props: {
           url: data?.post?.permalink,
           onReadArticleClick: openArticle(data),
-          onActivateCompanion: activateCompanion,
+          onActivateCompanion: activateCompanion(data),
+          onRequestClose: closeModal,
```
DragosIorgulescu, line 120: I am a bit at a loss regarding why this line causes a TS error -- `onRequestClose` should be an accepted attribute here based on how `LazyModalProps` is defined: https://github.com/dailydotdev/apps/blob/ecdf285060500aef2ea8b981ca12d48f6f03e975/packages/shared/src/components/modals/common/Modal.tsx#L32-L35

rebelchris: You shouldn't need to manually push that.
the idea behind the lazymodal is that it can trigger that itself

packages/shared/src/hooks/useExtensionPermission.ts
```diff
@@ -18,7 +18,7 @@ export type ContentScriptStatus = () => {
 
 interface ExtensionPermission {
   useContentScriptStatus: ContentScriptStatus;
-  requestContentScripts: () => Promise<boolean>;
+  requestContentScripts: (skipRedirect?: boolean) => Promise<boolean>;
```
capJavert, line 21: Better maybe to add it as object param

```ts
interface ExtensionPermission {
  useContentScriptStatus: ContentScriptStatus;
  requestContentScripts: ({ skipRedirect }?: { skipRedirect: boolean }) => Promise<boolean>;
```

this will allow us more customisability in the future when adding more options.

capJavert, line 21: Better maybe to add it as object param

```ts
interface ExtensionPermission {
  useContentScriptStatus: ContentScriptStatus;
  requestContentScripts: ({ skipRedirect }?: { skipRedirect: boolean }) => Promise<boolean>;
```

this will allow us more customisability in the future when adding more options.

packages/shared/src/hooks/useCompanionTrigger.tsx
DragosIorgulescu: Things get a bit tricky when you need a custom close button like the situation we have here. What are your thoughts on handling this?

rebelchris: You don't really need a custom one though, the action already exists.
You can just use it on the specific modal right

packages/shared/src/components/modals/CompanionModal.tsx
DragosIorgulescu: I can't find this color in the style guide -- double checking with design

packages/shared/src/components/modals/CompanionModal.tsx
DragosIorgulescu: I'm pretty sure I was up to date with `main`, installed the latest as well

packages/shared/src/components/modals/CompanionModal.tsx
rebelchris: It's the bun color, just not sure if we added it as a theme color already.

packages/shared/src/hooks/useCompanionTrigger.tsx
DragosIorgulescu: you're right, of course, I think I had a brain fart here

packages/shared/src/components/modals/CompanionModal.tsx
```diff
@@ -41,110 +42,117 @@ export default function CompanionModal({
...
-                    secondary
-                    size={IconSize.Medium}
-                    className="text-theme-color-avocado"
-                  />
-                </div>
-                <span>Bookmark with a single click</span>
+              <span>Discuss from everywhere</span>
+            </div>
+            <div className="hidden tablet:flex items-center pb-4">
+              <div className="inline-block p-2 mr-2 bg-theme-active rounded-16">
```
rebelchris, line 139: These should have the respective color as background.
For avocado for instance: `bg-theme-overlay-float-avocado`

packages/shared/src/components/modals/CompanionModal.tsx
```diff
@@ -41,110 +42,117 @@ export default function CompanionModal({
...
+              <span>Bookmark with a single click</span>
             </div>
-
-            <span className="text-center text-theme-label-tertiary typo-footnote">
-              Activating the widget requires extra permissions <br />
-              (Give it a try, you can always disable it later)
-            </span>
           </div>
+
+          <span className="text-center text-theme-label-tertiary typo-footnote">
```
rebelchris, line 150:
```suggestion
          <span className="text-center text-theme-label-tertiary typo-subhead">
```

rebelchris (changes requested):
Final feedback, other then this it looks great

packages/shared/src/components/modals/CompanionModal.tsx
DragosIorgulescu: @rebelchris looks like `bg-theme-overlay-blueCheese` & `-bun` are not defined, I tried adding these, but I think I am missing something as it does not seem to register the new color. What are all the places which need to be updated to add a new color?

rebelchris: Hmm those should be ok, not sure if the float was needed.
I'll check now :)

rebelchris (approved):
Let's go

PR #2101, opened 2023-08-28
fix: do not set cache if post was not fetched

capJavert:

Issue was that when commenting from modal we were trying to update single post cache. Since that one does not exist there was a runtime error in comment `onSuccess`. Comment would get posted but from UI it would look like it did not. User would then click comment again thus adding more then one comment.

We now skip cache update if there is not current single post cache.

This prevents cache update error when commenting from modal and also prevents filling partial post data to cache.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1728 #done

### DIFF
--- a/packages/shared/src/hooks/usePostById.ts
+++ b/packages/shared/src/hooks/usePostById.ts
@@ -31,13 +31,20 @@ export const updatePostCache = (
   client: QueryClient,
   id: string,
   { id: _, ...postUpdate }: Partial<Post>,
-): PostData =>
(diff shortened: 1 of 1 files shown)


rebelchris (approved):
Good for now.
Let's note down separate issue for the backlog

capJavert:
It was discovered that we do not update post cache in feed when commenting on modal. Ticket added here https://dailydotdev.atlassian.net/browse/WT-1730

cc: @rebelchris

PR #2107, opened 2023-08-28
fix: progress bar shows only while searching

capJavert:

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done

### DIFF
--- a/packages/shared/src/components/search/SearchContainer.tsx
+++ b/packages/shared/src/components/search/SearchContainer.tsx
@@ -17,7 +17,7 @@ export function SearchContainer({
     <main className="m-auto w-full">
       <div className="grid grid-cols-1 laptop:grid-cols-3 gap-y-6 py-8 m-auto w-full max-w-screen-laptopL">
         <main className="flex flex-col flex-1 col-span-2 px-4 laptop:px-8">
-          <SearchBar onSubmit={onSubmit} chunk={chunk} showProgress />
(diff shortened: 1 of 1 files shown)


packages/shared/src/components/search/SearchContainer.tsx
```diff
@@ -17,7 +17,7 @@ export function SearchContainer({
     <main className="m-auto w-full">
       <div className="grid grid-cols-1 laptop:grid-cols-3 gap-y-6 py-8 m-auto w-full max-w-screen-laptopL">
         <main className="flex flex-col flex-1 col-span-2 px-4 laptop:px-8">
-          <SearchBar onSubmit={onSubmit} chunk={chunk} showProgress />
+          <SearchBar onSubmit={onSubmit} chunk={chunk} showProgress={!!chunk} />
```
capJavert, line 20: It fills clunky that `!!chunk` is a condition here for `showProgress`, I expected `isLoading` here but it becomes `false`when search ends which then does not show final `Done` status label.

rebelchris: Hmm maybe a non urgent improvement we can backlog?
Technically it can probably get the data rather from a wrapping hook, instead of passing it down anyway

rebelchris (approved):
Works for now, see tip for future

sshanzel (approved):
Looks good to me 🚀

PR #2110, opened 2023-08-29
fix: mobile suggestion wrap

sshanzel:
## Changes
- Fix the wrapping of recommendation.

Preview:
image
Screenshot 2023-08-29 at 6 31 26 PM

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1708 #done

### DIFF
--- a/packages/shared/src/components/tooltips/InteractivePopup.tsx
+++ b/packages/shared/src/components/tooltips/InteractivePopup.tsx
@@ -12,6 +12,7 @@ export enum InteractivePopupPosition {
   LeftStart = 'leftStart',
   LeftCenter = 'leftCenter',
   LeftEnd = 'leftEnd',
+  Screen = 'screen',
 }
 
 interface InteractivePopupProps {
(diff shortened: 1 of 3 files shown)


rebelchris (changes requested):
I think this task was mainly about missing the top buttons on the actual device.
Does this solve that issue as well?

sshanzel:
> I think this task was mainly about missing the top buttons on the actual device.
Does this solve that issue as well?

Oh, there was no mention of what to fix and that overflow was the one that popped out. Let me check that as well.

packages/shared/src/components/tooltips/InteractivePopup.tsx
```diff
@@ -37,6 +38,7 @@ const positionClass: Record<InteractivePopupPosition, string> = {
   leftStart: classNames(leftClass, startClass),
   leftCenter: classNames(leftClass, centerClassY),
   leftEnd: classNames(leftClass, endClass),
+  screen: 'inset-0 w-screen h-screen',
```
sshanzel, line 41: At line 32, there are certain classes for `center` position that we don't need. So we introduce a specific one for the case we have now.

sshanzel:
@rebelchris have updated it now 🥳

packages/shared/src/components/tooltips/InteractivePopup.tsx
rebelchris: Ah right, something modal ofcourse does, good to know for future I guess :)

rebelchris (approved):

PR #2112, opened 2023-08-29
fix: disabling links while response is still streaming

sshanzel:
## Changes
- While streaming the data, we update the link to `#` so even while the link is still incomplete, it won't be redirected to any wrong url.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1734 #done


rebelchris (commented):
Is there no way we can make links clickable and working during streaming?
Not sure it's supported
Or show them as "disabled"

It's the same for the copy code button, they just don't work,. ideally they do.

rebelchris:
Hmm that's annoying, also for the copy codeblocks?
I tested it where it already streamed beyond the first block so expected it to work

capJavert (commented):
From what I see the links are clickable as soon as their markdown block is rendered.. only way when they are wrong its when markdown is not complete eg. `[This is link](https://something` (so the closing bracket is missing) but I would expect users understanding that this is not clickable. So links are clickable always from what I see.

As for code I think it might be due to parent div being updated constantly and events not bubbling because react is busy rendering vs propagating events. Just main thread being busy.

sshanzel:
The only concern is then, we currently don't have a way for us to identify whether the link is now rendered complete. If there is any, feel free to shoot.

DragosIorgulescu (approved):
Looks good 👌 as a quick fix solution. Regarding the points raised above, it might be worth exploring using web workers for this

rebelchris (approved):

PR #2119, opened 2023-08-30
fix: border issue on history page

sshanzel:
## Changes
- Removed the unnecessary max height.

Preview:
Screenshot 2023-08-30 at 12 02 52 PM

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1724 #done


sshanzel:
Also, I noticed something odd with our pages. There seems to be an extra element that makes the scroll show every time even when not needed as shown below:

Screenshot 2023-08-30 at 11 59 28 AM

The issue is that the Search bar for the Control Search page is always shown even when not needed.

Fix:

Screenshot 2023-08-30 at 11 59 40 AM

rebelchris (approved):
Oh interesting, was it not added to always show the border till bottom even if theres only 1 item?
Not sure we missing that part now?

If so maybe a min-height or just the h-full will work?

If it works already approving 🙏

sshanzel:
> Oh interesting, was it not added to always show the border till bottom even if theres only 1 item?

Also got me confused why it was not consuming the whole screen if there was only 1 item.

For the min-height, we have it already from the main container so we should be good with that.

PR #2121, opened 2023-08-30
Wt 1739 sources fixes

rebelchris:
- Make full search bar clickable
- Going to revamp once design is applicable

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done


packages/shared/src/components/search/SearchSourceList.tsx
```diff
@@ -28,14 +28,24 @@ export const SearchSourceList = ({
...
+    <PageWidgets
+      tablet={false}
+      className="relative order-3 laptop:order-last col-span-2 laptop:col-span-1"
+    >
       <i className="hidden laptop:block absolute top-8 -left-8 w-12 h-px bg-theme-divider-tertiary" />
       <div className={classNames('flex flex-col', widgetClasses)}>
         <div
           className={classNames(
-            'flex justify-between items-center py-1.5 laptop:py-4 px-4 laptop:!pb-0 laptop:bg-transparent rounded-t-16 bg-theme-bg-secondary',
+            'cursor-pointer laptop:cursor-auto flex justify-between items-center py-1.5 laptop:py-4 px-4 laptop:!pb-0 laptop:bg-transparent rounded-t-16 bg-theme-bg-secondary',
```
sshanzel, line 39: Should we maybe turn this into a `button` element instead, so we won't have to manually add some props and classes.

rebelchris: I thought about that, but since on desktop it's not a button I felt super conflicted 😐
Would you say just render a different wrapper on mobile rather?

sshanzel: Ah, right. This is only for mobile. Then I think it would be better to keep it this way.

sshanzel: Yep. Accordion component would be the best here and having a disabled state to make the header unclickable, probably.

packages/shared/src/components/search/SearchSourceList.tsx
```diff
@@ -28,14 +28,24 @@ export const SearchSourceList = ({
...
             isSourcesOpen ? 'rounded-t-16' : 'rounded-16',
           )}
+          role="button"
+          tabIndex={0}
+          aria-label="Sources"
+          onKeyDown={() =>
+            !sidebarRendered && setIsSourcesOpen((prev) => !prev)
+          }
+          onClick={() => !sidebarRendered && setIsSourcesOpen((prev) => !prev)}
```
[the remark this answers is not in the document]
rebelchris: Ah good point, will do that 🙏

packages/shared/src/components/search/SearchSourceList.tsx
capJavert: Fine by me, as long as the proper role and attributes are set, and from what I see they are 👍

PR #2126, opened 2023-08-30
feat: search analytics events tracking

DragosIorgulescu:
## Changes

Adding events tracking for the following:
- search focus & submit events
- search history -- open / impressions
- click events for recommendations & sources
- upvote/downvote/copy events
- search error event

## Events

| Type | event_name  | value |
|------|-------------|-------|
| new  | focus search |
| new  | submit search       | 
| new  | open search history | 
| new  | impression |   `extra: { origin: home page / history tooltip / history page }` |   
| new  | click |  `extra: { origin: home page / history tooltip / history page }` |   
| new  | click         |   
| new  | upvote search   | 
| new  | downvote search  |
| new  | copy search  | 
| new  | error search |  

The full (and detailed) list can be found here: https://docs.google.com/spreadsheets/d/18Lv7zXges9QfVX5VYL1a-Hyl0e1sQ3sLr0OK8YZWKXI/edit#gid=1166511776

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1666 #done


packages/shared/src/components/search/SearchBarSuggestion.tsx
```diff
@@ -7,11 +7,51 @@ import {
...
-}: ButtonProps<TagName>): ReactElement => {
+}: SearchBarSuggestionProps): ReactElement => {
+  const { trackEvent } = useContext(AnalyticsContext);
+
+  useEffect(() => {
+    if (props.suggestion?.id) {
+      trackEvent({
+        event_name: AnalyticsEvent.Impression,
+        target_type: TargetType.SearchRecommendation,
+        target_id: props.suggestion.id,
```
DragosIorgulescu, line 35: using `id` here, but I wanted to point to where suggestions are loaded in `useSearchSuggestions` as it looks to me like the `id` is being dropped here? https://github.com/dailydotdev/apps/blob/f63091bfe9423bf3153f4ba16b6f6d45d0fd946d/packages/shared/src/hooks/search/useSearchSuggestions.ts#L26-L32

packages/shared/src/components/search/SearchHistory.tsx
```diff
@@ -19,12 +21,19 @@ export function SearchHistory({
...
     result: { isLoading },
     nodes,
     infiniteScrollRef,
   } = useSearchHistory();
 
+  useEffect(() => {
+    trackEvent({ event_name: AnalyticsEvent.OpenSearchHistory });
+
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
```
rebelchris, line 35: I think we need to change this one to trigger once clicking the search history tab.
It might trigger multiple times that way.

packages/shared/src/components/search/SearchBarSuggestion.tsx
```diff
@@ -7,11 +7,51 @@ import {
...
+        feed_item_title: props.suggestion.prompt,
+        extra: JSON.stringify({ origin }),
+      });
+    }
+
+    // @NOTE see https://dailydotdev.atlassian.net/l/cp/dK9h1zoM
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [props.suggestion?.id]);
```
rebelchris, line 43: This should probably be empty to avoid it retriggering.
I'm also not sure if there is a case where it's not set.

rebelchris: Ah is it to distinguish between history and recommendations?

packages/shared/src/components/search/SearchBarSuggestion.tsx
```diff
@@ -7,11 +7,51 @@ import {
...
 import { AiIcon } from '../icons';
+import { AnalyticsEvent, Origin, TargetType } from '../../lib/analytics';
+import { SearchSession } from '../../graphql/search';
+import AnalyticsContext from '../../contexts/AnalyticsContext';
 
-export const SearchBarSuggestion = <TagName extends AllowedTags>({
+export type SuggestionOrigin =
+  | Origin.HomePage
+  | Origin.HistoryTooltip
+  | Origin.HistoryPage;
```
rebelchris, line 17: This can only be search bar now

packages/shared/src/components/search/SearchBarSuggestion.tsx
```diff
@@ -7,11 +7,51 @@ import {
...
+        feed_item_title: props.suggestion.prompt,
+        extra: JSON.stringify({ origin }),
+      });
+    }
+
+    // @NOTE see https://dailydotdev.atlassian.net/l/cp/dK9h1zoM
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [props.suggestion?.id]);
+
+  const handleSuggestionsClick = useCallback(() => {
```
rebelchris, line 45: Same comment as above

packages/shared/src/lib/analytics.ts
```diff
@@ -93,6 +108,8 @@ export enum TargetType {
   LoginButton = 'login button',
   SignupButton = 'signup button',
   SquadJoinButton = 'squad join button',
+  SearchRecommendation = 'search rec',
+  SearchSource = 'search source',
```
rebelchris, line 117: I think this ones implementation is missing

packages/shared/src/components/search/SearchResult.tsx
```diff
@@ -68,11 +71,31 @@ export function SearchResult({
...
+
+        const eventNames = {
+          1: AnalyticsEvent.UpvoteSearch,
+          [-1]: AnalyticsEvent.DownvoteSearch,
+        };
+
+        trackEvent({
+          event_name: eventNames[chunk.feedback],
+          target_id: chunk.id,
+        });
```
rebelchris, line 83: I think we need to move this to the onMutate, it's important to know regardless of it suceeding or not.

Assumption for now.
Tagging @idoshamun to verify that was the case for tracking.

packages/shared/src/hooks/useChat.ts
```diff
@@ -164,7 +168,12 @@ export const useChat = ({ id: idFromProps, query }: UseChatProps): UseChat => {
       },
       progress: -1,
     });
-  }, [setSearchQuery]);
+
+    trackEvent({
+      event_name: AnalyticsEvent.ErrorSearch,
+      target_id: id,
+    });
```
rebelchris, line 175: I think this one only trigger on stream error, we should also trigger it on error we receive inside the stream.

rebelchris (changes requested):
Nice work!
Overall looks great, have some minor adjustments

packages/shared/src/lib/analytics.ts
DragosIorgulescu: good catch 👍

packages/shared/src/components/search/SearchResult.tsx
DragosIorgulescu: moved to `onMutate` for now, let me know if I need to rollback

packages/shared/src/components/search/SearchBarSuggestion.tsx
DragosIorgulescu: `suggestion` is not guaranteed to exist here, an example can be found here https://github.com/dailydotdev/apps/blob/e839ca1f453ee3b860e06c31748b68576eb399b9/packages/shared/src/components/search/SearchBarSuggestionList.tsx#L28-L36

packages/shared/src/components/search/SearchBarSuggestion.tsx
rebelchris: In that case we'll need to add it right, to make this work?

packages/shared/src/components/search/SearchBarSuggestion.tsx
rebelchris: ah ok, that's true indeed
Good spot

packages/shared/src/components/search/SearchSourceItem.tsx
```diff
@@ -9,9 +11,22 @@ interface SearchSourceItemProps {
...
 }: SearchSourceItemProps): ReactElement {
-  if (!item) return null;
+  const { trackEvent } = useContext(AnalyticsContext);
+  const { id: itemId, name, snippet, url } = item || {};
 
-  const { name, snippet, url } = item;
+  const handleSourceClick = useCallback(() => {
+    if (itemId) {
+      trackEvent({
+        event_name: AnalyticsEvent.Impression,
```
rebelchris, line 20: This should be a click handle

rebelchris (changes requested):
Almost there!

packages/shared/src/components/search/SearchBar.tsx
```diff
@@ -19,7 +20,7 @@ export function SearchBar({
   isLoading,
   ...props
 }: SearchBarProps): ReactElement {
-  const suggestionsProps = useSearchSuggestions();
+  const suggestionsProps = useSearchSuggestions({ origin: Origin.HomePage });
```
idoshamun, line 23: Search bar can show up on the search page as well, isn't it?

packages/shared/src/components/search/SearchBarSuggestion.tsx
DragosIorgulescu: yep, I added it in this PR

packages/shared/src/components/search/SearchBar.tsx
rebelchris: Ah good spot, yes it can!

packages/shared/src/components/search/SearchSourceItem.tsx
idoshamun: Good catch!

packages/shared/src/components/search/SearchBar.tsx
DragosIorgulescu: yes it does, however this distinction was not made in the taxonomy sheet so I ignored it -- do you want to keep track of this? -- IMO not sure how valuable is the information of knowing a suggestion showed up on the home page or search page

idoshamun (commented):
Overall looks great some comments. I also think we miss `open search history` when clicking on the tooltip right?

packages/shared/src/components/search/SearchSourceItem.tsx
DragosIorgulescu: good catch 🙏

DragosIorgulescu:
> Overall looks great some comments. I also think we miss `open search history` when clicking on the tooltip right?

@idoshamun this event should be covered here: https://github.com/dailydotdev/apps/pull/2126/files#diff-84677ba741cf1a29eb1431ef02c8348cef571a9c2d8a0555b6960406f3b551c3R53-R55

packages/shared/src/components/search/SearchResult.tsx
idoshamun: It's ok :)

idoshamun:
@DragosIorgulescu gotcha! I missed that

rebelchris:
@idoshamun with open search history firing in two spots, don't you want to maybe add origin there?

idoshamun:
@rebelchris for now we can skip it, we can use the event_page property if needed

rebelchris (approved):
Think we there :)

PR #2129, opened 2023-08-30
fix: single line recommendations

sshanzel:
## Changes
- Make the search recommendation in a single scrollable line.
- Making it have ellipsis won't be possible due to the nature of the component.

Screenshot 2023-08-30 at 8 31 50 PM

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1701 #done

### DIFF
--- a/packages/shared/src/components/feeds/FeedContainer.tsx
+++ b/packages/shared/src/components/feeds/FeedContainer.tsx
@@ -162,7 +162,7 @@ export const FeedContainer = ({
             <span className="flex flex-row flex-1 mt-4">
               <SearchBarSuggestionList
                 {...suggestionsProps}
-                className="hidden tablet:flex overflow-hidden flex-1 mr-3"
+                className="hidden tablet:flex overflow-hidden flex-1 tablet:flex-none mr-3"
               />
               {actionButtons && (
                 <span className="flex flex-row gap-3 pl-3 ml-auto border-l border-theme-divider-tertiary">
--- a/packages/shared/src/components/search/SearchBarSuggestionList.tsx
+++ b/packages/shared/src/components/search/SearchBarSuggestionList.tsx
@@ -50,7 +50,12 @@ export function SearchBarSuggestionList({
   }
 
   return (
-    <div className={classNames('flex flex-wrap gap-4', className)}>
+    <div
+      className={classNames(
+        'flex max-w-full overflow-scroll gap-4 p-2 pl-0',
+        className,
+      )}
+    >
       {suggestions.map(({ id, prompt }) => (
         <SearchBarSuggestion
           tag="a"
--- a/packages/shared/src/hooks/search/useSearchHistory.ts
+++ b/packages/shared/src/hooks/search/useSearchHistory.ts
@@ -23,6 +23,7 @@ export const useSearchHistory = ({
     generateQueryKey(RequestKey.SearchHistory, user, `limit:${limit}`),
     ({ pageParam }) => getSearchHistory({ after: pageParam, first: limit }),
     {
+      enabled: !!user,
       getNextPageParam: (lastPage) =>
         lastPage?.history?.pageInfo.hasNextPage &&
         lastPage?.history?.pageInfo.endCursor,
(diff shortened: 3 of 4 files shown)


rebelchris:
@sshanzel It should not have the scrollbar please 🙏
Let's maybe check together then if we can find a better solution

sshanzel:
> @sshanzel It should not have the scrollbar please 🙏 Let's maybe check together then if we can find a better solution

Sure. What should be our alternatives?

rebelchris:
@sshanzel 
Showing less of the options on desktop when space is limited.
Flex options maybe to shrink items with ellipsis?
JS to add ellipsis?

Not really sure what's right, but we need to rethink it :)

sshanzel:
Would this be a viable option for laptop?
image

sshanzel:
@rebelchris have updated it now.

Also, disabled search history and recommendations for anonymous users as there is none to return in that case.

Also fixed the issue with the pill taking the whole space all the time.

packages/shared/src/components/feeds/FeedContainer.tsx
```diff
@@ -162,7 +162,7 @@ export const FeedContainer = ({
             <span className="flex flex-row flex-1 mt-4">
               <SearchBarSuggestionList
                 {...suggestionsProps}
-                className="hidden tablet:flex overflow-hidden flex-1 mr-3"
+                className="hidden tablet:flex overflow-hidden flex-1 tablet:flex-none mr-3"
```
rebelchris, line 165: You have `tablet:flex` and `tablet:flex-none`

Not sure which one get's rendered 👀
Maybe it's russian roulette

packages/shared/src/components/search/SearchBarSuggestion.tsx
```diff
@@ -14,7 +14,7 @@ export const SearchBarSuggestion = <TagName extends AllowedTags>({
 }: ButtonProps<TagName>): ReactElement => {
   return (
     <Button
-      spanClassName="w-fit py-2 flex-shrink"
+      spanClassName="w-fit my-2 flex-shrink laptop:line-clamp-2"
```
rebelchris, line 17: What actually happens here if you did line-clamp-1?
Just wondering

rebelchris (changes requested):
One breaking issue

packages/shared/src/components/feeds/FeedContainer.tsx
sshanzel: The `flex-none` is not conflicting with the `flex` as the former only unsets the `flex-1` but keeps the `display: flex` intact.

packages/shared/src/components/search/SearchBarSuggestion.tsx
sshanzel: It would only keep 1 line.

rebelchris: Hmm but the one-line with ellipsis wouldn't work?
It's more for a future improvement. I might take a look at it later, fun CSS Challenge

packages/shared/src/components/feeds/FeedContainer.tsx
rebelchris: Ah right, :) tailwind naming brain fart

rebelchris (approved):

packages/shared/src/components/search/SearchBarSuggestion.tsx
sshanzel: It should work. It would look like below. Not sure if that is what you are envisioning, but I agree, it would be a cool challenge.

image

packages/shared/src/components/feeds/FeedContainer.tsx
sshanzel: Just like ellipsis classes. Not sure which one to use, `text-ellipsis` or `overflow-ellipsis` 😂

packages/shared/src/components/search/SearchBarSuggestion.tsx
rebelchris: Yeah did you share this as a alternative with design?
I'd be cool to find out how you could make them dynamically flex-grow available space and just hide the ones that don't fit 😅

sshanzel: Yes, the single line is what I initially shared. It was mostly through JavaScript, calculating how many pills would exceed the container's width and then deciding how much would fit.

PR #2131, opened 2023-08-30
refactor: chat logic to support navigation and different sessions

capJavert:
- Fixing issues related to search navigation and mixup of streaming and existing session results 
- More info here https://dailydotdev.slack.com/archives/C05EM7ZFQMQ/p1693402562918429
- split up useChat hook internally into two parts
- refactor query keys to avoid wildcard `new` and always load correct ID from `session_created` message.
- move all router logic to `SearchPage` since it is specific to that page

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done

### DIFF
--- a/packages/shared/src/hooks/chat/useChatStream.ts
+++ b/packages/shared/src/hooks/chat/useChatStream.ts
@@ -1,71 +1,25 @@
+import { useRef, useState, useCallback, useEffect } from 'react';
+import { useQueryClient } from 'react-query';
+import { useAuthContext } from '../../contexts/AuthContext';
 import {
-  MouseEvent,
-  useCallback,
-  useEffect,
-  useMemo,
-  useRef,
-  useState,
-} from 'react';
-import { QueryKey, useQuery, useQueryClient } from 'react-query';
-import {
-  getSearchSession,
-  initializeSearchSession,
-  Search,
   SearchChunk,
-  SearchChunkError,
-  SearchChunkSource,
-  sendSearchQuery,
   updateSearchData,
-} from '../graphql/search';
-import { generateQueryKey, RequestKey } from '../lib/query';
-import { useAuthContext } from '../contexts/AuthContext';
+  initializeSearchSession,
+  SearchChunkError,
+  sendSearchQuery,
+  Search,
+} from '../../graphql/search';
+import { generateQueryKey, RequestKey } from '../../lib/query';
+import {
+  UseChatMessage,
+  UseChatMessageType,
+  CreatePayload,
+  SourcesMessage,
+  TokenPayload,
+  UseChatStream,
+} from './types';
 
-interface UseChatProps {
-  id?: string;
-}
-
-enum UseChatMessageType {
-  SessionCreated = 'session_created',
-  WebSearchFinished = 'web_search_finished',
-  WebResultsFiltered = 'web_results_filtered',
-  StatusUpdated = 'status_updated',
-  NewTokenReceived = 'new_token_received',
-  Completed = 'completed',
-  Error = 'error',
-  SessionFound = 'session_found',
-}
-
-interface SourcesMessage {
-  sources: SearchChunkSource[];
-}
-
-interface UseChatMessage<Payload = unknown> {
-  type: UseChatMessageType;
-  status?: string;
-  timestamp: number;
-  payload: Payload;
-}
-
-interface UseChat {
-  queryKey: QueryKey;
-  data: Search;
-  isLoading: boolean;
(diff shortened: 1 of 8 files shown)


packages/shared/src/hooks/useChat.ts
```diff
@@ -176,13 +81,96 @@ export const useChat = ({ id: idFromProps, query }: UseChatProps): UseChat => {
         sourceRef.current.close();
       }
 
-      setSearchQuery(undefined);
+      setSessionId(undefined);
+
+      let queryKey: ReturnType<typeof generateQueryKey> = null;
```
rebelchris, line 86: Just to note according to TKdodo this is very error prone and queryKeys should ideally always be consts as close to the query as possible.
I do see why you chose this and just leaving this as a mention, not sure if there even is a action point

packages/shared/src/hooks/useChat.ts
```diff
@@ -176,13 +81,96 @@ export const useChat = ({ id: idFromProps, query }: UseChatProps): UseChat => {
...
+              client.setQueryData(queryKey, sessionData);
+              sourceRef.current?.close();
+              break;
+            }
+            default:
+              break;
+          }
+        } catch (error) {
+          // eslint-disable-next-line no-console
+          console.error('[EventSource][message] error', error);
```
rebelchris, line 153: Not sure why we even need this if it's just a console log, is it handled somewhere actually?

packages/shared/src/hooks/useChat.ts
```diff
@@ -65,106 +63,13 @@ interface TokenPayload {
   token: string;
 }
 
-export const useChat = ({ id: idFromProps, query }: UseChatProps): UseChat => {
+export const useChatStream = (): Pick<UseChat, 'handleSubmit'> & {
```
rebelchris, line 66: Small request can we maybe make a `chat` or `useChat` directory inside hooks which exports a index and split these three in seperate files?
I'm more a fan of keeping files smaller and easier to distinguish.

rebelchris (changes requested):
Massive work and find.
Have some minor comments overall looks good.

Would like to see it split up personally, but let me know what you think.
Also will be testing it out a bit this morning.

packages/webapp/components/search/SearchV1Page.tsx
```diff
@@ -7,19 +7,53 @@ import {
...
+    if (!searchId) {
+      return;
+    }
+
+    const newRoute = `${searchPageUrl}?id=${searchId}`;
+
+    router.replace(newRoute, undefined, {
+      shallow: true,
+    });
+    // router reference is not stable https://github.com/vercel/next.js/issues/18127
```
sshanzel, line 34: Oh, cool to know. Should we keep this on moving forward for other dependencies with `router` object?

packages/shared/src/hooks/useChat.ts
capJavert: > Just to note according to TKdodo this is very error prone and queryKeys should ideally always be consts as close to the query as possible.

I know what you are hinting at but the fact we have a generator with `generateQueryKey` makes it ok since we don't generate query keys by hand.

packages/shared/src/hooks/useChat.ts
capJavert: For sure 👍

packages/webapp/components/search/SearchV1Page.tsx
capJavert: If we are using just the methods then yes, if we use some value like `router.query.something`, then we should include. It is fixed in Next 13 so hopefully down the line we upgrade and can use the new hook 🤞

packages/shared/src/hooks/chat/useChatStream.ts
```diff
@@ -0,0 +1,168 @@
...
+    async (value: string) => {
+      if (!value) {
+        return;
+      }
+
+      if (sourceRef.current?.OPEN) {
+        sourceRef.current.close();
+      }
+
+      setSessionId(undefined);
```
rebelchris, line 41: undefined or null?

rebelchris (approved):
Amazing, super cleanup 🙏

packages/shared/src/hooks/chat/useChatStream.ts
capJavert: yeah, `null` is probably better since we are using it in queryKey.

PR #2136, opened 2023-08-31
fix: list markdown line break

DragosIorgulescu:
## Preview

https://github.com/dailydotdev/apps/assets/1225100/902f214d-4cc4-44ec-bfbc-c0899934918e

## Events

_no events changed_

- [x] MobileL (420px)
- [x] Tablet (656px)
- [x] Laptop (1020px)

- [x] iOS (Chrome and Safari)
- [ ] Android

WT-1726 #done


packages/shared/src/components/RenderMarkdown.tsx
```diff
@@ -128,6 +128,9 @@ const RenderMarkdown = ({
       linkTarget="_blank"
       remarkPlugins={plugins}
       components={{
+        ol({ children }) {
+          return <ol>{children}</ol>;
+        },
```
rebelchris, line 133: Why is this needed, does it not already return it like this?

rebelchris (approved):
Looks ok, just not sure why it happens still.

packages/shared/src/components/RenderMarkdown.tsx
DragosIorgulescu: it does return the exact same thing without this, but for some reason explicitly setting this fixes the issue -- my hunch is that it has to do with re-renders of these components during streaming, but I could not tell you exactly why this solves it...

capJavert: No idea why it would work but if it works let's role 🚀

capJavert (approved):
Let's role 🚀

PR #2142, opened 2023-08-31
feat: rate limit error handling

capJavert:
### Describe what this PR does
- handle rate limit errors and show custom message

## Events

Did you introduce any new tracking events?
Don't forget to update the [Analytics Taxonomy sheet](https://docs.google.com/spreadsheets/d/18Lv7zXges9QfVX5VYL1a-Hyl0e1sQ3sLr0OK8YZWKXI/edit#gid=0)

Log the new/changed events below:

| Type   | event_name  | value |
|--------|-------------|-------|
| Change/New | event name  | extra: { ... } |

### On those affected packages:
- [ ] Have you done sanity checks in the webapp?
- [ ] Have you done sanity checks in the extension?
- [ ] Does this not break anything in companion?

### Did you test the modified components media queries?
- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

#### Did you test on actual mobile devices?
- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done

PR #2149, opened 2023-08-31
fix: initial error not showing due to wrong queryKey

capJavert:
- initial error (like rate limit) did not show because I introduced the new query key based on current timestamp for each new query https://github.com/dailydotdev/apps/pull/2142#discussion_r1311516282
- this however made `useChatSession` to not know about it 🤦 
- in this PR I fallback to empty query key with `null` and clean it before each new query
  - also makes less mess inside our react query devtools

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done

### DIFF
--- a/packages/shared/src/hooks/chat/useChatStream.ts
+++ b/packages/shared/src/hooks/chat/useChatStream.ts
@@ -45,9 +45,10 @@ export const useChatStream = (): UseChatStream => {
       let queryKey: ReturnType<typeof generateQueryKey> = generateQueryKey(
         RequestKey.Search,
         user,
-        Date.now().toString(),
+        null,
       );
       let streamId: string = null;
+      client.removeQueries(queryKey);
 
       const initSession = (payload: Pick<CreatePayload, 'id'>) => {
         queryKey = generateQueryKey(RequestKey.Search, user, payload.id);


rebelchris (approved):
Instead of null you can just not pass anything there right?
Let's add a ticket for some test cases before we forget lol

packages/shared/src/hooks/chat/useChatStream.ts
```diff
@@ -45,9 +45,10 @@ export const useChatStream = (): UseChatStream => {
       let queryKey: ReturnType<typeof generateQueryKey> = generateQueryKey(
         RequestKey.Search,
         user,
-        Date.now().toString(),
+        null,
```
sshanzel, line 48: The last parameter is optional, right?

capJavert: So if we don't set anything then the query key would be `['search', userId]`, and when clearing it up with `removeQueries` it would remove all other queries related to search. That is why I added `null` to have I guess shared state for those initial errors (rate limit is the only one currently).

I see the bad side in this where if we used two streams in the same time they would share this error state. I do think its something we can sort when adding support for multiple streams in the same time in the future.

cc: @rebelchris since you also noted it.

rebelchris: Hmm let's leave it as is for now then :)

capJavert: Merging this for now so it can be tested, if any further comments I also added a task https://dailydotdev.atlassian.net/browse/WT-1777

sshanzel: Agreed let's see how it goes. We can probably use the session id for the last parameter to make it linked to that id only.

capJavert: We are using it, but on initial errors like rate limit there is no session id. That the only case where it is null for now.

PR #2150, opened 2023-09-01
fix: top position hack

sshanzel:
## Changes
- The motivation behind wrapping the component with `isOpen` was to avoid the unnecessary scroll that happens when it is always rendered.
- We reverted it back, but we now removed the hack of hiding it from the screen from doing a `top-full` that places it to the bottom-most part which results in scrollbar showing even when not needed.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1773 #done

### DIFF
--- a/packages/shared/src/components/PostsSearch.tsx
+++ b/packages/shared/src/components/PostsSearch.tsx
@@ -151,15 +151,13 @@ export default function PostsSearch({
         aria-haspopup="true"
         aria-expanded={isOpen}
       />
-      {isOpen && (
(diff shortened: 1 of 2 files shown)


rebelchris:
Ah i did see this and weird enough typing a second letter worked so thought it was just a lag on my side ☠️

Thanks for the fix 🙏

rebelchris (approved):

PR #2156, opened 2023-09-04
feat: introduce curly eslint configuration

sshanzel:
## Changes
- Introduce an ESLint rule for curly braces.
- Used the configuration of `multi` based from this [discussion](https://dailydotdev.atlassian.net/browse/WT-1700?focusedCommentId=11070).
- Ran `npm run lint:fix` on all 3 packages.

Kindly prepare yourself as this can be an exhausting review, if even worth checking all the files since I just ran the command on all 3 packages.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1700 #done

### DIFF
--- a/packages/eslint-config/index.js
+++ b/packages/eslint-config/index.js
@@ -28,6 +28,7 @@ module.exports = {
     node: true,
   },
   rules: {
+    curly: ['error', 'all'],
     'react/prop-types': 0,
     'react/require-default-props': 0,
     'react/no-unused-prop-types': 0,
--- a/packages/extension/src/background/index.ts
+++ b/packages/extension/src/background/index.ts
@@ -28,7 +28,9 @@ const excludedCompanionOrigins = [
 ];
 
 const isExcluded = (origin: string) => {
-  if (!origin) return false;
+  if (!origin) {
+    return false;
+  }
   return excludedCompanionOrigins.some((e) => origin.includes(e));
 };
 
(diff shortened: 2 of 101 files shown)


packages/eslint-config/index.js
```diff
@@ -28,6 +28,7 @@ module.exports = {
     node: true,
   },
   rules: {
+    curly: ['error', 'multi'],
```
capJavert, line 31: Why not use `all`?

One of the main points we use and fail with is `if` where we do a one liner and then add another line and then we have to add the brackets. I think just add brackets always and that's it. 

Also minimizes changes made since brackets will be there from the beginning.

sshanzel:
> Why not use all?
> 
> One of the main points we use and fail with is if where we do a one liner and then add another line and then we have to add the brackets. I think just add brackets always and that's it.
> 
> Also minimizes changes made since brackets will be there from the beginning.

Good point. Stricter sometimes is better. We should maybe open a poll in the web team channel. Wdyt?

packages/eslint-config/index.js
capJavert: > Good point. Stricter sometimes is better. We should maybe open a poll in the web team channel. Wdyt?

@sshanzel yes! 👌 , we can also have a quick discussion on the next web team meet to quickly go through pros/cons I think it should be an easy choice.

sshanzel:
I've separated the commits per package to make it at least a bit easier to review.

rebelchris (approved):
Skimmed through it, didn't really see any odd ones

PR #2157, opened 2023-09-04
fix: ensure recommendation impression events are triggered on update

DragosIorgulescu:
## Events
_No events changes_

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1779 #done

### DIFF
--- a/packages/shared/src/components/search/SearchBarSuggestion.tsx
+++ b/packages/shared/src/components/search/SearchBarSuggestion.tsx
@@ -39,7 +39,7 @@ export const SearchBarSuggestion = ({
 
     // @NOTE see https://dailydotdev.atlassian.net/l/cp/dK9h1zoM
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, []);
+  }, [props.suggestion?.id]);
 
   const handleSuggestionsClick = useCallback(() => {
     if (props.suggestion?.id) {
--- a/packages/shared/src/graphql/search.ts
+++ b/packages/shared/src/graphql/search.ts
@@ -72,6 +72,7 @@ export const SEARCH_POST_SUGGESTIONS = gql`
 export const SEARCH_POST_RECOMMENDATION = gql`
   query SearchQuestionRecommendations {
     searchQuestionRecommendations {
+      id
       question
     }
   }
--- a/packages/shared/src/components/search/SearchBarSuggestionList.tsx
+++ b/packages/shared/src/components/search/SearchBarSuggestionList.tsx
@@ -65,7 +65,6 @@ export function SearchBarSuggestionList({
           suggestion={suggestion}
           key={suggestion.prompt}
           href={getSearchUrl({
-            id: suggestion.id,
             question: suggestion.prompt,
           })}
         >
--- a/packages/shared/src/components/search/SearchHistory.tsx
+++ b/packages/shared/src/components/search/SearchHistory.tsx
@@ -43,7 +43,6 @@ export function SearchHistory({
           suggestion={suggestion}
           href={getSearchUrl({
             id: suggestion.id,
-            question: suggestion.prompt,
           })}
         >
           {suggestion.prompt}


packages/shared/src/components/search/SearchBarSuggestion.tsx
```diff
@@ -39,7 +39,7 @@ export const SearchBarSuggestion = ({
 
     // @NOTE see https://dailydotdev.atlassian.net/l/cp/dK9h1zoM
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, []);
+  }, [props.isHistory, props.suggestion?.id]);
```
rebelchris, line 42: With this in place we need a local state to ensure it's only called once 🙏

rebelchris (changes requested):

packages/shared/src/components/search/SearchBarSuggestion.tsx
DragosIorgulescu: I could add a variable using `useMemo` for this, but isn't this check taking care of making sure the event is sent only once per suggestion? https://github.com/dailydotdev/apps/blob/bddb648fb3e059798b3b8e4f2750f08fb30e6096/packages/shared/src/components/search/SearchBarSuggestion.tsx#L30

DragosIorgulescu: actually, nevermind, I see what you mean

DragosIorgulescu:
@rebelchris I changed it to only execute when `suggestion?.id` changes, as `isHistory` is actually not relevant here. This should ensure the event is triggered only once per suggestion

packages/shared/src/components/search/SearchBarSuggestion.tsx
capJavert: Isn't it that search suggestions don even have `id`?

DragosIorgulescu: @capJavert hmm, in that case what would the `target_id` be here? https://docs.google.com/spreadsheets/d/18Lv7zXges9QfVX5VYL1a-Hyl0e1sQ3sLr0OK8YZWKXI/edit#gid=1166511776

capJavert: Not sure about event setup but if I am looking at `SEARCH_POST_RECOMMENDATION` query, it does not have an id.

DragosIorgulescu: I thought you meant the resource actually has no ID 😅 , I added the ID to the query, thanks for pointing it out

packages/shared/src/graphql/search.ts
```diff
@@ -72,6 +72,7 @@ export const SEARCH_POST_SUGGESTIONS = gql`
 export const SEARCH_POST_RECOMMENDATION = gql`
   query SearchQuestionRecommendations {
     searchQuestionRecommendations {
+      id
```
rebelchris, line 75: With this addition we actually have to be very careful.
It will push these as we already have a session and it won't exist yet. 😅

So maybe we have to escape it somewhere. to prevent urls from being wrong.

rebelchris (changes requested):

packages/shared/src/components/search/SearchBarSuggestion.tsx
rebelchris: It's a DB ID so it will fail if we redirect with it

packages/shared/src/graphql/search.ts
DragosIorgulescu: @rebelchris good point, I updated so the URL always includes just `q` query param for suggestions -- is there any reason we would ever want to pass the ID here?

packages/shared/src/components/search/SearchBarSuggestion.tsx
capJavert: But looks like we do need it for analytics confirmed by Ido here https://docs.google.com/spreadsheets/d/18Lv7zXges9QfVX5VYL1a-Hyl0e1sQ3sLr0OK8YZWKXI/edit?disco=AAAA47IXf14, ok, from my point of view this one is good but we do have to make sure we don't use it for recommendations. Maybe update the types to differentiate from search history and `id` there.

capJavert: Like make it `sessionId` and `id`.

capJavert: Currently `SearchSession` is used for both but `id` means different thing in regards to:
- history => session ID from previous search
- recommendation => internal API id for suggestion

capJavert (approved):

idoshamun:
@DragosIorgulescu I see this fix addresses only the impression side of things, but I haven't seen clicks as well

DragosIorgulescu:
> @DragosIorgulescu I see this fix addresses only the impression side of things, but I haven't seen clicks as well

@idoshamun good point, the click events should be fixed by the addition of the `id` column though, since these were dependent on it (which is why they were not firing before) https://github.com/dailydotdev/apps/blob/bddb648fb3e059798b3b8e4f2750f08fb30e6096/packages/shared/src/components/search/SearchBarSuggestion.tsx#L45-L55

PR #2158, opened 2023-09-04
refactor: rename `id` to `sessionId` on search history query results

DragosIorgulescu:
## Changes

To ensure the distinction between `SearchSession` and `SearchQuestion` objects is clear, we are renaming the `SearchSession.id` attribute to `SearcSession.sessionId` -- discussion started here: https://github.com/dailydotdev/apps/pull/2157#discussion_r1314869136

## Events

_no event changes_

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

### DIFF
--- a/packages/shared/src/components/search/SearchBarSuggestion.tsx
+++ b/packages/shared/src/components/search/SearchBarSuggestion.tsx
@@ -40,7 +40,7 @@ export const SearchBarSuggestion = ({
 
     // @NOTE see https://dailydotdev.atlassian.net/l/cp/dK9h1zoM
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [props.suggestion?.id]);
+  }, [props.id]);
 
   const handleSuggestionsClick = useCallback(() => {
     if (props?.id) {
--- a/packages/shared/src/components/search/SearchBarSuggestion.tsx
+++ b/packages/shared/src/components/search/SearchBarSuggestion.tsx
@@ -55,7 +55,7 @@ export const SearchBarSuggestion = ({
 
     // @NOTE see https://dailydotdev.atlassian.net/l/cp/dK9h1zoM
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [suggestionId, isHistory, impressionEmitted]);
+  }, [suggestionId, isHistory]);
 
   const handleSuggestionsClick = useCallback(() => {
     if (suggestionId) {
--- a/packages/shared/src/components/search/SearchBarSuggestion.tsx
+++ b/packages/shared/src/components/search/SearchBarSuggestion.tsx
@@ -1,4 +1,10 @@
-import React, { ReactElement, useCallback, useContext, useEffect } from 'react';
+import React, {
+  ReactElement,
+  useCallback,
+  useContext,
+  useEffect,
+  useState,
+} from 'react';
 import classNames from 'classnames';
 import {
   AllowedTags,
@@ -23,38 +29,44 @@ type SearchBarSuggestionProps = ButtonProps<AllowedTags> & {
 export const SearchBarSuggestion = ({
   className,
   origin,
+  id: suggestionId,
+  isHistory,
+  prompt,
   ...props
 }: SearchBarSuggestionProps): ReactElement => {
   const { trackEvent } = useContext(AnalyticsContext);
+  const [impressionEmitted, toggleImpressionEmitted] = useState(false);
 
   useEffect(() => {
-    if (!props.isHistory && props?.id) {
+    if (!isHistory && suggestionId && !impressionEmitted) {
       trackEvent({
         event_name: AnalyticsEvent.Impression,
         target_type: TargetType.SearchRecommendation,
-        target_id: props.id,
-        feed_item_title: props.prompt,
+        target_id: suggestionId,
+        feed_item_title: prompt,
         extra: JSON.stringify({ origin }),
       });
+
+      toggleImpressionEmitted(true);
     }
 
     // @NOTE see https://dailydotdev.atlassian.net/l/cp/dK9h1zoM
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [props.id]);
+  }, [suggestionId, isHistory, impressionEmitted]);
 
   const handleSuggestionsClick = useCallback(() => {
-    if (props?.id) {
+    if (suggestionId) {
       trackEvent({
         event_name: AnalyticsEvent.Click,
-        target_type: props.isHistory
+        target_type: isHistory
           ? TargetType.SearchHistory
           : TargetType.SearchRecommendation,
-        target_id: props.id,
-        feed_item_title: props.prompt,
+        target_id: suggestionId,
+        feed_item_title: prompt,
         extra: JSON.stringify({ origin }),
       });
     }
-  }, [origin, props.id, props.prompt, trackEvent, props.isHistory]);
+  }, [origin, suggestionId, prompt, trackEvent, isHistory]);
 
   return (
     <Button
--- a/packages/shared/src/components/search/SearchBarSuggestion.tsx
+++ b/packages/shared/src/components/search/SearchBarSuggestion.tsx
@@ -3,7 +3,7 @@ import React, {
   useCallback,
   useContext,
   useEffect,
-  useState,
+  useRef,
 } from 'react';
 import classNames from 'classnames';
 import {
@@ -35,10 +35,10 @@ export const SearchBarSuggestion = ({
   ...props
 }: SearchBarSuggestionProps): ReactElement => {
   const { trackEvent } = useContext(AnalyticsContext);
-  const [impressionEmitted, toggleImpressionEmitted] = useState(false);
+  const impressionEmitted = useRef(false);
 
   useEffect(() => {
-    if (!isHistory && suggestionId && !impressionEmitted) {
+    if (!isHistory && suggestionId && !impressionEmitted.current) {
       trackEvent({
         event_name: AnalyticsEvent.Impression,
         target_type: TargetType.SearchRecommendation,
@@ -47,7 +47,7 @@ export const SearchBarSuggestion = ({
         extra: JSON.stringify({ origin }),
       });
 
-      toggleImpressionEmitted(true);
+      impressionEmitted.current = true;
     }
 
     // @NOTE see https://dailydotdev.atlassian.net/l/cp/dK9h1zoM
--- a/packages/shared/src/components/search/SearchBarSuggestion.tsx
+++ b/packages/shared/src/components/search/SearchBarSuggestion.tsx
@@ -17,12 +17,15 @@ import { AnalyticsEvent, Origin, TargetType } from '../../lib/analytics';
 import { SearchQuestion } from '../../graphql/search';
 import AnalyticsContext from '../../contexts/AnalyticsContext';
 
-export type SuggestionOrigin = Origin.HomePage | Origin.SearchPage;
+export type SuggestionOrigin =
+  | Origin.HomePage
+  | Origin.SearchPage
+  | Origin.HistoryPage;
 
 type SearchBarSuggestionProps = ButtonProps<AllowedTags> & {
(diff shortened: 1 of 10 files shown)


packages/shared/src/components/search/SearchBarSuggestionList.tsx
```diff
@@ -63,13 +63,13 @@ export function SearchBarSuggestionList({
           tag="a"
           origin={origin}
           suggestion={suggestion}
-          key={suggestion.prompt}
+          key={suggestion.question}
           href={getSearchUrl({
             id: suggestion.id,
-            question: suggestion.prompt,
```
rebelchris, line 69: Somehow we need to ensure we distinguish between. prompt and question else it will become very unclear.
Not really sure what's right here, but maybe up for discussion.

packages/shared/src/components/search/SearchBarSuggestion.tsx
```diff
@@ -32,7 +32,7 @@ export const SearchBarSuggestion = ({
         event_name: AnalyticsEvent.Impression,
         target_type: TargetType.SearchRecommendation,
         target_id: props.suggestion.id,
```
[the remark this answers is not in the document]
DragosIorgulescu: yep, I missed this one

packages/shared/src/components/search/SearchBarSuggestionList.tsx
DragosIorgulescu: We could consolidate these so they use the same attribute name?

DragosIorgulescu:
> We really have to make sure tracking is not affected by this change.
@rebelchris I agree, I'd like to do some thorough testing before merging this one

packages/shared/src/components/search/SearchBarSuggestion.tsx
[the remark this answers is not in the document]
capJavert: Maybe think about solving this with composition. We leave the base component `SearchBarSuggestion` with base props and `onClick`.

We then create `SearchBarRecommendation` and `SearchBarHistory `:
- we add `useEffect` to `SearchBarRecommendation`
- we use `onClick` to send correct `target_type` in both components.
- one should accept `suggestion` as `SearchQuestion` and `SearchSession` for other one.

This will eliminate the rest of issues with double logic and component doing two things. 

Wrote above before you made your comment but I agree with your point completely though that component does many things. So up to you to see what is better composition or props refactor.

packages/shared/src/components/search/SearchBarSuggestion.tsx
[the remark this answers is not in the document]
capJavert: I think not, recommendation will never have `sessionId` right?

packages/shared/src/components/search/SearchBarSuggestionList.tsx
capJavert: `question` is column from our API and `prompt` coming from Magni's session. Quickest just to rename it on our client or long term make it consistent across our services.

packages/shared/src/components/search/SearchBarSuggestion.tsx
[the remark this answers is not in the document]
rebelchris: Yeah this is probably needed to avoid issues down the line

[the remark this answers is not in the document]
DragosIorgulescu: Thanks for the suggestions 🙏  I went with the option of updating the component props to make it less opinionated, while also not adding too many new components for this. 

-- We still have a slight issue I think with the naming of this component, as it's not really just a search bar suggestion component, but I don't think that's too bad

packages/shared/src/components/search/SearchBarSuggestion.tsx
```diff
@@ -27,33 +28,33 @@ export const SearchBarSuggestion = ({
...
+        target_id: props.id,
+        feed_item_title: props.prompt,
         extra: JSON.stringify({ origin }),
       });
     }
 
     // @NOTE see https://dailydotdev.atlassian.net/l/cp/dK9h1zoM
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [props.suggestion?.id]);
+  }, [props.id]);
```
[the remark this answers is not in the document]
rebelchris: It should never refire the tracking though, so with this in mind we need a simple state to keep track of whether it's been fired on this render or not.
It can be as simple as a useState boolean that get's set inside the evalaution.

I know it's probably overkill and never triggers, but let's be safe if something changes down the line.

packages/shared/src/components/search/SearchBarSuggestion.tsx
```diff
@@ -8,13 +8,14 @@ import {
...
 import { AnalyticsEvent, Origin, TargetType } from '../../lib/analytics';
-import { SearchSession } from '../../graphql/search';
+import { SearchQuestion } from '../../graphql/search';
 import AnalyticsContext from '../../contexts/AnalyticsContext';
 
 export type SuggestionOrigin = Origin.HomePage | Origin.SearchPage;
 
 type SearchBarSuggestionProps = ButtonProps<AllowedTags> & {
-  suggestion?: Partial<SearchSession>;
+  id?: SearchQuestion['id'];
```
rebelchris, line 17: With this new addition id and prompt actually become mantadory right?
Same actually with origin, I think we can make them mantadory

rebelchris (changes requested):
Agreed with Ante to be more explicit on the deps, but would then also like to see this simple state for keeping track of unique trackEvents

packages/shared/src/components/search/SearchBarSuggestion.tsx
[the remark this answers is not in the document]
capJavert: ref is better to not trigger re-renders, but yeah, if we want to make sure this only fires on component mount then this is legit way.

packages/shared/src/components/search/SearchBarSuggestion.tsx
```diff
@@ -8,13 +8,14 @@ import {
...
+  prompt?: SearchQuestion['question'];
   origin?: SuggestionOrigin;
```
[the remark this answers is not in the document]
DragosIorgulescu: Unless we want to use a different component for the recommendation shown when the user is not logged in, we can't make `id` mandatory here: https://github.com/dailydotdev/apps/blob/8cae5d365838b438d13e206e40e1b7077fa9ccb8/packages/shared/src/components/search/SearchBarSuggestionList.tsx#L29-L34

packages/shared/src/components/search/SearchBarSuggestion.tsx
DragosIorgulescu: Regarding `id` and `prompt`, as per the comment earlier, this component is being used to show the login CTA when as a search suggestion when the user is not logged in: https://github.com/dailydotdev/apps/blob/8cae5d365838b438d13e206e40e1b7077fa9ccb8/packages/shared/src/components/search/SearchBarSuggestionList.tsx#L29-L34

`origin` can't be mandatory for the same reason as above, but also because when this component is used in `Search History`, we are not passing any `origin`: https://github.com/dailydotdev/apps/blob/8e61ae29f061c60fb69e7774db8804efc0b3ffcb/packages/shared/src/components/search/SearchHistory.tsx#L38-L50

packages/shared/src/components/search/SearchBarSuggestion.tsx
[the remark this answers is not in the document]
DragosIorgulescu: Updated based on your feedback, we have a bool internal state which should make extra sure these are not emitted multiple times

rebelchris (approved):
Alright that would work I think

packages/shared/src/components/search/SearchBarSuggestion.tsx
capJavert: From my point of view it's ok if components need props to be optional but your tracking checks do not take it into account so someone could send partial data and then we would get partial data in events that would not be according to spec. 

For example `origin` is always required in events and in this component setup it could be undefined.

packages/shared/src/components/search/SearchBarSuggestion.tsx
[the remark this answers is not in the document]
capJavert: See my comment here https://github.com/dailydotdev/apps/pull/2158#discussion_r1317087102

packages/shared/src/components/search/SearchBarSuggestion.tsx
```diff
@@ -8,52 +14,59 @@ import {
...
 export const SearchBarSuggestion = ({
   className,
   origin,
+  id: suggestionId,
+  isHistory,
+  prompt,
   ...props
 }: SearchBarSuggestionProps): ReactElement => {
   const { trackEvent } = useContext(AnalyticsContext);
+  const [impressionEmitted, toggleImpressionEmitted] = useState(false);
```
capJavert, line 38: `useRef` here

DragosIorgulescu: right, missed that, thx 🙏

packages/shared/src/components/search/SearchBarSuggestion.tsx
DragosIorgulescu: @capJavert fair point, what would be your suggestion in this case? Should we add a new component used for the login CTA suggestion? 

-- I was also not aware `origin` is always required, based on the [taxonomy sheet](https://docs.google.com/spreadsheets/d/18Lv7zXges9QfVX5VYL1a-Hyl0e1sQ3sLr0OK8YZWKXI/edit?pli=1#gid=1166511776) it looks like most events don't require it? or maybe there is some default value being set somewhere in the frontend that I missed, if not passed explicitly?

capJavert: To make it specific to this implementation so we can close a PR

> Regarding id and prompt, as per the comment earlier, this component is being used to show the login CTA when as a search suggestion when the user is not logged in

I would create another component for login message

> origin can't be mandatory for the same reason as above, but also because when this component is used in Search History, we are not passing any origin

Per taxonomy sheet there should be `history page` origin that you can add and pass to `SearchHistory` component

Then we can make all props required right and avoid partial event edge cases?

DragosIorgulescu: @capJavert thank you for the suggestion, I updated based on this discussion, and in the process found a few missing events as well. Take a look at the last commit & let me know your thoughts

packages/shared/src/components/search/SearchBarSuggestion.tsx
```diff
@@ -8,52 +14,62 @@ import {
...
         extra: JSON.stringify({ origin }),
       });
+
+      impressionEmitted.current = true;
     }
 
     // @NOTE see https://dailydotdev.atlassian.net/l/cp/dK9h1zoM
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [props.suggestion?.id]);
+  }, [suggestionId, isHistory, impressionEmitted]);
```
capJavert, line 58: You don't need `impressionEmitted` since it's a `ref` object and not reactive.

capJavert (approved):
Looks awesome, thank you for all the cleanup 💪 

One minor comment but feel free to merge after that.

packages/shared/src/components/search/SearchHistoryButton.tsx
```diff
@@ -91,6 +101,9 @@ export function SearchHistoryButton(): ReactElement {
         id={contextMenuId}
         options={options}
         onHidden={() => setIsMenuOpen(false)}
+        onShown={() => {
+          trackEvent({ event_name: AnalyticsEvent.OpenSearchHistoryTooltip });
+        }}
```
rebelchris, line 106: This is not validated or requested by analytics team, we can't make them up for them. 🙏

rebelchris (changes requested):
One blocker from my side 🙏

rebelchris (approved):
Good to go 🙏

PR #2160, opened 2023-09-06
refactor: migration of experiments from flagsmith to growthbook

sshanzel:
# Blocked until we define all the flags in GrowthBook.
- Important to note that we should not remove the values in Flagsmith platform as there can still be extension users using the old version.
- With GrowthBook, we can easily define the date requirement for onboarding v2 directly.
- Once we have defined the GrowthBook values, we can then move to testing phase where the experimental values are sent properly.

## Changes
- Complete removal of flagsmith. Removed the package as well.
- Cleaned up unused experiments.
- Kept the `OnboardingV2` as there are still anonymous users that have access to feed.
- The `SubmitArticle` experiment was just a condition whether to display the popup. The requirement of reputation is handled in the backend fetched through a query.
- Hiring item in the Sidebar was asked to be completely removed.
- Refactored the `featureManagement` file due to static properties of a class is not allowed to use the generic parameter defined by the class. Resulting in not properly inferring the type of the feature and becoming a union type instead. Simplified it to define an object we export that contains all the values of the experiments.

Analytics does still send the `feature_flags` property:
Screenshot 2023-09-06 at 11 59 29 PM

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1766 #done


packages/shared/src/hooks/useReferralCampaign.ts
```diff
@@ -36,17 +36,14 @@ export const campaignToReferralTargetCountMap: Record<
...
 > = {};
 
 const useReferralCampaign = ({
   campaignKey,
 }: UseReferralCampaignProps): UseReferralCampaign => {
-  const { flags } = useContext(FeaturesContext);
   const featureFlag = campaignFeatureFlagMap[campaignKey];
-  const isCampaignEnabled = featureFlag
-    ? !!getFeatureValue(featureFlag, flags)
-    : true;
```
sshanzel, line 49: I am a bit confused here, does this mean the default value is true?

packages/shared/src/lib/cookie.ts
```diff
@@ -97,47 +61,3 @@ export const expireCookie = (
 
   setCookie(name, 'expired', { path, domain, maxAge: 0 });
 };
-
-export const updateFeatureFlags = (
```
sshanzel, line 101: We will have to come up another way of overriding feature flags for preview deployment but have decided to take it into a different PR as this is a large one already.

packages/shared/src/components/sidebar/ContributeSection.tsx
```diff
@@ -34,7 +28,7 @@ export function ContributeSection({
     trackEvent({
       event_name: 'start submit article',
       feed_item_title: 'Submit article',
-      extra: JSON.stringify({ has_access: canSubmitArticle }),
+      extra: JSON.stringify({ has_access: true }),
```
DragosIorgulescu, line 31: is this correct as always true? and if yes, why don't we remove it entirely?

packages/shared/src/contexts/__snapshots__/AnalyticsContext.spec.tsx.snap
DragosIorgulescu, line 1: do we neet to include growthbook flags here?

DragosIorgulescu (commented):
nice work 🤩  I have a couple of questions, but I really don't see any issues here

packages/shared/src/contexts/BootProvider.tsx
```diff
@@ -89,14 +87,8 @@ export const BootDataProvider = ({
...
   const loadedFromCache = !!cachedBootData;
-  const {
-    user,
-    settings,
-    flags = {},
-    alerts,
-    notifications,
-    squads,
-  } = cachedBootData || {};
+  const { user, settings, alerts, notifications, squads } =
```
rebelchris, line 90: With this we need to ensure we also remove it on API side.
But might have to consider rewriting the can_submit_article in that case.

packages/shared/src/contexts/__snapshots__/AnalyticsContext.spec.tsx.snap
rebelchris: They need to be removed all together from tracking events actually, good point! 👌

packages/shared/src/hooks/analytics/useAnalyticsSharedProps.ts
```diff
@@ -19,10 +19,11 @@ export default function useAnalyticsSharedProps(
   // Use ref instead of state to reduce renders
   const sharedPropsRef = useRef<Partial<AnalyticsEvent>>();
   const { query } = useRouter();
-  const { flags } = useContext(FeaturesContext);
   const { themeMode, spaciness, insaneMode } = useContext(SettingsContext);
   const { visit, anonymous, tokenRefreshed, user } = useContext(AuthContext);
   const [sharedPropsSet, setSharedPropsSet] = useState(false);
+  const { growthbook } = useGrowthBookContext();
+  const flags = growthbook.getFeatures();
```
rebelchris, line 26: We actually decided to remove these, as it will enroll people by default.
But maybe @idoshamun this way could potentially work? (not sure how you feel about it)

rebelchris (changes requested):
Wow massive amounts of work!
Overall good, one open question around the tracking, but need @idoshamun feedback on that as well.

packages/shared/src/components/sidebar/ContributeSection.tsx
sshanzel: This won't get triggered until the user will have the proper access. So it will always be true, yes, once the user is eligible.

packages/shared/src/contexts/__snapshots__/AnalyticsContext.spec.tsx.snap
sshanzel: Nice. Let's remove it then 🚀

packages/shared/src/components/sidebar/ContributeSection.tsx
sshanzel: In the past, this was always based on the flagsmith value which always `true` since we released it to everyone. But maybe what we should have done is verify whether the user is eligible.

This seems like we had the wrong value all along as it is always true.

packages/shared/src/contexts/BootProvider.tsx
sshanzel: We can't remove the flags on the API side yet to support extension users until we get certain amount of adoption rates. Let me create a task for clean up.

packages/shared/src/hooks/analytics/useAnalyticsSharedProps.ts
idoshamun: Yap you're right

capJavert (commented):
Did not find anything alarming, nice work on cleanup 👌

packages/shared/src/hooks/useReferralCampaign.ts
sshanzel: Amazing. Thanks for the confirmation.

packages/shared/src/components/sidebar/ContributeSection.tsx
sshanzel: I have now moved the tracking on the place where we identify whether the user is eligible.
https://github.com/dailydotdev/apps/pull/2160/commits/2ba26b35d112c224854ec2cdfe8efe1a2d17f548

sshanzel: After checking the API, this seems to be adjusted on the go. So we were getting the right values for this flag.

Since we will be removing flagsmith, and calls to feature flags are done directly to GB, we won't be able to do overrides anymore.

We have two options:
- The first is to go with the commit mentioned above.
- The second one is we will utilize the targeting conditions offered by GB since the value is based on the reputation which is found on the user's table which we can access from there.

Would love to hear what you guys think.

sshanzel: Good point, I totally missed that part. Yep, we can do the user state. Then for outdated extension users, we will just keep the API logic as it is where it checks whether the user should have it.

packages/shared/src/components/modals/SubmitArticleModal.tsx
```diff
@@ -173,6 +181,23 @@ export default function SubmitArticleModal({
     );
   };
 
+  useEffect(() => {
```
idoshamun, line 184: Why you have to introduce it here?

packages/shared/src/components/sidebar/SidebarBottomSection.tsx
```diff
@@ -54,15 +51,6 @@ export function SidebarBottomSectionSection({
     },
   ];
 
-  if (showHiring) {
```
idoshamun, line 57: Why do we move it from here?

packages/shared/src/hooks/useReferralCampaign.ts
```diff
@@ -36,17 +36,14 @@ export const campaignToReferralTargetCountMap: Record<
...
+  const isCampaignEnabled = useFeatureIsOn(featureFlag);
```
idoshamun, line 46: Haven't we concluded this already? Do we need it?

idoshamun (commented):
I had some comments but overall seems great

packages/shared/src/components/modals/SubmitArticleModal.tsx
sshanzel: We have previously thought it was wrong as it was coming from a concluded experiment. After checking BE, it was then getting updated. Now that we have introduced the user state relevant to the eligibility, we will check from there. We should revert this change and have the same process but utilize the said property.

packages/shared/src/components/sidebar/SidebarBottomSection.tsx
sshanzel: It was asked to be removed completely.

Relevant discussion: https://dailydotdev.slack.com/archives/C02JAUF8HJL/p1693840704759149?thread_ts=1693840563.128999&cid=C02JAUF8HJL

packages/shared/src/hooks/useReferralCampaign.ts
sshanzel: Yes, though I believe this was built to be dynamic, hence, the featureFlag is coming from a map. cc @capJavert

packages/shared/src/components/sidebar/ContributeSection.tsx
```diff
@@ -46,39 +31,38 @@ export function ContributeSection({
...
 
+  useEffect(() => {
+    const search = new URLSearchParams(window.location.search);
+    const query = Object.fromEntries(search);
+    if (!query?.scout) {
+      return;
+    }
+
+    const { origin, pathname } = window.location;
+    trackAndShowSubmitArticle();
```
rebelchris, line 54: Yeah I think we can track this inside the modal rather?

packages/shared/src/components/sidebar/ContributeSection.tsx
```diff
@@ -46,39 +31,38 @@ export function ContributeSection({
...
+      active: modal.type === LazyModal.NewSource,
     },
   ];
 
+  useEffect(() => {
+    const search = new URLSearchParams(window.location.search);
+    const query = Object.fromEntries(search);
+    if (!query?.scout) {
+      return;
+    }
```
rebelchris, line 51: Where was this handled before, is there still a case for it?

packages/shared/src/components/sidebar/ContributeSection.tsx
```diff
@@ -46,39 +31,38 @@ export function ContributeSection({
...
       ),
       title: 'Suggest new source',
-      action: () => setShowNewSourceModal(true),
-      active: showNewSourceModal,
+      action: () => openModal({ type: LazyModal.NewSource }),
+      active: modal.type === LazyModal.NewSource,
     },
   ];
 
+  useEffect(() => {
```
sshanzel, line 46: Since we are now using the lazy modal for the modals here, the hook that uses this side effect became so much simpler now and have decided to just put it here.

Referring to the `useSubmitArticle` hook.

packages/shared/src/components/sidebar/ContributeSection.tsx
sshanzel: Relevant comment: https://github.com/dailydotdev/apps/pull/2160#discussion_r1321037239

packages/shared/src/components/sidebar/ContributeSection.tsx
rebelchris: I would still put the tracking inside the modal actually.
Maybe the state that evaluates if we show the modal, but I'm unsure if it's actually used anymore.

packages/shared/src/components/sidebar/ContributeSection.tsx
sshanzel: Relevant comment: https://github.com/dailydotdev/apps/pull/2160#discussion_r1321037239

packages/shared/src/components/sidebar/ContributeSection.tsx
sshanzel: From inside the modal it would be under a `useEffect`. From here it would be right when the callback was executed. I think it would be better to avoid additional use effect. Wdyt?

rebelchris: But this is a useEffect?
And what about other states that open the modal?
Are they non tracked

packages/shared/src/components/sidebar/ContributeSection.tsx
sshanzel: I have now introduced the user state for identifying whether the user is eligible in the boot endpoint. So from that, it would result the same value.

packages/shared/src/components/sidebar/ContributeSection.tsx
sshanzel: This is a different use-case. This was coming from the deleted hook. This happens when the user received a notification that they are now eligible. They get redirected and we watch the query params to see if we should display.

Previously, we were actually not tracking it.

packages/shared/src/components/sidebar/ContributeSection.tsx
rebelchris: I'm not 100% sure I follow on this part, maybe it's best we sync on this quickly.

sshanzel: Cool. Let me know once you are free, I should be too.

sshanzel: This is basically the useEffect inside the deleted hook. This was never inside the modal.

packages/shared/src/components/modals/SubmitArticleModal.tsx
sshanzel: We've updated to keep this here to avoid missing the tracking event as we open this modal from different actions.

packages/shared/src/components/modals/SubmitArticleModal.tsx
```diff
@@ -173,6 +174,17 @@ export default function SubmitArticleModal({
...
+    trackEvent({
+      event_name: 'start submit article',
```
rebelchris, line 179: Oh just one thing I noted now, we should move this to our enum names :)
It's a legacy one, but good opportunity to fix.

Possible also the feed_item_title.

sshanzel: Good point. Let's update this.

packages/shared/src/hooks/useReferralCampaign.ts
capJavert: Yes, if campaign is behind feature flag it will only enable hook when campaign is active.

sshanzel:
@rebelchris would love to hear what you think 🚀

rebelchris (approved):
I don't see any blockers anymore.
The one thing I would like to "test" if possible is the onboarding wall, we need a rock solid test case where existing users don't get it now if they are on the growthbook version.

Would this be possible for you to recreate?

sshanzel:
Should be possible yeah. I can connect on our redis cache to update the firstVisit.

PR #2161, opened 2023-09-07
fix: auth copy changes

rebelchris:
- Cleaned some components as they got a bit mixed up
- Still not happy with how we render modals in normal pages, but for now it was too big of a effort to clean
- Mainly copy changes, and label usages

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1767 #done


packages/shared/src/components/auth/AuthHeader.tsx
```diff
@@ -5,17 +5,23 @@ import { CloseModalFunc } from '../modals/common';
...
+function AuthHeader({
+  simplified = false,
   title,
   className,
   onBack,
-}: AuthModalHeaderProps): ReactElement {
+}: AuthHeaderProps): ReactElement {
+  if (simplified) {
+    return <h2 className="font-bold text-center typo-title2">{title}</h2>;
+  }
```
rebelchris, line 23: Basically simplified means outside of the modal, so header needs to be inlined to show per screen :)

packages/shared/src/components/auth/LoginForm.tsx
```diff
@@ -108,6 +111,21 @@ function LoginForm({
           {loginButton}
         </Button>
       </span>
+      {hint && hint === labels.auth.error.invalidEmailOrPassword && (
```
rebelchris, line 114: This is a bit of a mission, but the request was to make it more clear so a custom hint exception for this specific hint only

packages/webapp/pages/onboarding.tsx
```diff
@@ -163,6 +161,65 @@ export function OnboardPage(): ReactElement {
     setHasSelectTopics(hasTopics);
   };
 
+  const getContent = (): ReactElement => {
```
rebelchris, line 164: Wasn't sure why this was "conditionalWrapped" it should have been conditionally rendered as adjusted.

sshanzel: I think the conditional wrap came from the requirement of auth options having the custom `title` and `content` as opposed to having just the `AuthOptions` now.

omBratteng (approved):
Looks good to me, also simplifies some of the code I'm currently working on.

omBratteng:
Should this copy also be updated?
https://github.com/dailydotdev/apps/blob/9bd26c30cf9246327f3e99d7bed1c2999dad38f0/packages/shared/src/components/onboarding/MemberAlready.tsx#L21

rebelchris:
@omBratteng The already member one is for a seperate task for now, @sshanzel  is working on that one.
It's basically a complete revamp of how it looks and works

packages/shared/src/components/auth/AuthModalFooter.tsx
```diff
@@ -1,33 +1,35 @@
...
-        {isLogin ? 'Don’t have an account?' : 'Already have an account?'}
-      </Modal.Text>
+      {text.body && <Modal.Text>{text.body}</Modal.Text>}
       <ClickableText
         className="ml-1 underline text-theme-label-primary"
-        onClick={() => onIsLogin(!isLogin)}
+        onClick={onClick}
       >
-        {isLogin ? 'Sign up' : 'Log in'}
+        {text.button}
```
DragosIorgulescu, line 32: this looks funky, what is the difference between `text.body` and `text.button` here?

packages/shared/src/components/auth/LoginForm.tsx
DragosIorgulescu: assuming `labels.auth.error.invalidEmailOrPass` is never null or undefined, I think there would be no need to check if `hint` exists as well
```suggestion
      {hint === labels.auth.error.invalidEmailOrPassword && (
```

DragosIorgulescu (approved):
solid work 👌  some very minor questions from me -- looks like there is a failing spec as well but not sure if it's just a flaky one

packages/shared/src/components/auth/AuthModalFooter.tsx
rebelchris: We render a button which has some text in it, and some descriptive optional text on top.

PR #2162, opened 2023-09-07
fix: correctly set the `sessionId` alias on search history query

DragosIorgulescu:
## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done

### DIFF
--- a/packages/shared/src/graphql/search.ts
+++ b/packages/shared/src/graphql/search.ts
@@ -125,7 +125,7 @@ export const SEARCH_HISTORY_QUERY = gql`
       }
       edges {
         node {
-          id: sessionId
+          sessionId: id
           prompt
           createdAt
         }


packages/shared/src/graphql/search.ts
```diff
@@ -125,7 +125,7 @@ export const SEARCH_HISTORY_QUERY = gql`
       }
       edges {
         node {
-          id: sessionId
+          sessionId: id
```
rebelchris, line 128: But with this your tracking on the history button will break right?
Used as `target_id: node.sessionId`

rebelchris (commented):
Afraid this might break tracking maybe?

packages/shared/src/graphql/search.ts
DragosIorgulescu: this is actually the correct way to alias an attribute in a query -- what I had before was actually telling the client to query for the `sessionId` attribute and alias it to `id`, instead of the other way around, which is why the graphql query is failing.

I did not catch this earlier because I could not find an easy way of testing search history locally.

rebelchris: Oh I see alias in graphQL are reversed 🤦

rebelchris (approved):

PR #2059, opened 2023-08-18
feat: adding the engagement loops for upvoting / downvoting

TomDowling:
### Describe what this PR does
Setting up the engagement loops for upvoting / downvoting. Adding the changes to the feeds to show the new Loop as well on the Reading history. This included updating all usages of upvoting / downvoting across the app.

For any discussion regarding voting logic see merge request https://github.com/dailydotdev/apps/pull/2164 which was merged into this one.

Also implements:
- WT-1657
- WT-1656

## Events

Did you introduce any new tracking events?
Don't forget to update the [Analytics Taxonomy sheet](https://docs.google.com/spreadsheets/d/18Lv7zXges9QfVX5VYL1a-Hyl0e1sQ3sLr0OK8YZWKXI/edit#gid=0)

Log the new/changed events below:

| Type   | event_name  | value |
|--------|-------------|-------|
| New| upvote post  | extra: { origin: 'feedback card' } |
| New| upvote post  | extra: { origin: 'history'} |
| New| downvote post  | extra: { origin: 'feedback card' } |
| New| downvote post  | extra: { origin: 'history'} |
| New | click dismiss feedback | |

Also adds `extra.feedback: true` to `click`, `impression` and `go to link` post events.

### On those affected packages:
- [ ] Have you done sanity checks in the webapp?
- [ ] Have you done sanity checks in the extension?
- [ ] Does this not break anything in companion?

### Did you test the modified components media queries?
- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

#### Did you test on actual mobile devices?
- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done

PR #2164, opened 2023-09-08
feat: new vote api in scope of engagement loops

capJavert:
- this is a intermediary branch to make reviewing specific changes to vote interfaces more easy
  - note this is gonna be merged into engagement loops branch since it requires the changes
- it is mostly made from comments  from target PR https://github.com/dailydotdev/apps/pull/2059
- changes here try and simplify how our useVotePost hook works to use new `userState.vote` across webapp and extension

TODO
- [x] working on fixing tests while opening for review

Some specific changes
- `useVotePost` now has tow additional functions returned `toggleUpvote` and `toggleDownvote`, then encapsulate usual checks we do when upvoting like cancel, login and base event tracking
  - hook still exposes base methods like `upvotePost` which are used in companion since it has a little different structure still in how it handles it's code 
- each type of upvote still has a separate hook eg `useFeedVotePost`, this is to be able to customize parameters needed for the hook while keeping return interface the same
- `useVotePost` now accepts single `onMutate` function which is optional for base hook
  - each custom hook can overwrite this function to create its own mutation and rollback for optimistic UI while using base mutation handlers from `useVotePost`
  - `useReadHistoryVotePost` is a example of completely new implementation
- vote related hooks were moved to `hooks/vote`, so we can keep them in one place and expand in the future with comments maybe
- `useFeedVotePost` used mutation subscription to sync feed and modal upvote states, it still uses it but change is that now the subscription and base hook reuse the same `mutateVoteFeedPost` function passed to `onMutate`
- the above made is simpler for mutation subscription is moved to separate hook to allow easier usage in the future

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1658 #done


packages/shared/src/components/FeedItemComponent.tsx
```diff
@@ -42,20 +44,8 @@ export type FeedItemComponentProps = {
...
-    upvoted: boolean,
-  ) => Promise<void>;
-  onDownvote: (
-    post: Post,
-    index: number,
-    row: number,
-    column: number,
-    upvoted: boolean,
-  ) => Promise<void>;
+  onUpvote: UseVotePost['toggleUpvote'];
```
rebelchris, line 47: Just wondering would this not be better:
`Pick<UseVotePost, 'toggleUpvote'>;`

packages/shared/src/components/FeedItemComponent.tsx
```diff
@@ -163,8 +153,28 @@ export default function FeedItemComponent({
...
+            onUpvote({
+              post,
+              origin: Origin.Feed,
+              opts: {
+                columns,
+                column,
+                row,
+              },
+            });
+          }}
```
rebelchris, line 166: Why is this one using the `useVotePost` if it's feedItemComponent.
Could we not use the `useFeedVotePost` here?

packages/shared/src/components/cards/FeedbackCard.tsx
```diff
@@ -55,7 +32,7 @@ export const FeedbackCard = ({ post }: FeedbackCardProps): ReactElement => {
         <Button
           id="upvote-post-btn"
           pressed={post?.userState?.vote === UserPostVote.Up}
-          onClick={() => toggleUpvote(post)}
+          onClick={() => toggleUpvote({ post, origin: Origin.Feed })}
```
rebelchris, line 35: This can get rendered on feed and modal right?
Not sure the origin matches in that case? (But now wondering how it was before then)

packages/shared/src/components/modals/report/ReportPostModal.tsx
```diff
@@ -26,14 +26,14 @@ interface OptionalProps {
 }
 
 export type ReportedCallback = (
-  post: Post | PostBootData,
+  post: Post | PostBootData | ReadHistoryPost,
```
rebelchris, line 29: No real action point but this is getting messy, I wonder if we can uniform them to one type that way potentially you could also have generic useVote hook and other interactions.

packages/shared/src/components/post/PostActions.tsx
```diff
@@ -42,95 +37,29 @@ export function PostActions({
...
-          extra: { origin },
-        }),
-      );
+      onClose(true);
+    }
 
-      downvotePost({ id: post.id });
+    if (post?.userState?.vote === UserPostVote.None) {
+      onShowPanel();
     }
```
rebelchris, line 60: Is this the new experiment panel?
If so it should also be under a flag.

packages/shared/src/components/post/PostItemCard.tsx
```diff
@@ -24,7 +28,8 @@ export interface PostItemCardProps {
   clickable?: boolean;
   onHide?: (params: HidePostItemCardProps) => Promise<unknown>;
   onContextMenu?: (event: React.MouseEvent, post: PostItem) => void;
-  isFeedbackEnabled?: boolean;
+  showVoteActions?: boolean;
+  analyticsOrigin?: Origin;
```
rebelchris, line 32: I think we generally just call if origin globally.

packages/shared/src/components/post/PostItemCard.tsx
```diff
@@ -39,43 +44,29 @@ export default function PostItemCard({
...
-    update: mutationHandlers.downvote(post),
-  });
-  const onCancelPostDownvoteMutate = updatePost({
-    id: post.id,
-    update: mutationHandlers.cancelDownvote(post),
-  });
+  const client = useQueryClient();
+  const readingHistory = client.getQueryData<ReadHistoryInfiniteData>(
+    generateQueryKey(RequestKey.ReadingHistory, user),
+  );
```
rebelchris, line 61: Why do we interact with reading history on the postItemCard?

packages/shared/src/hooks/vote/types.ts
```diff
@@ -0,0 +1,78 @@
...
+  origin: Origin;
+  post: Post | ReadHistoryPost;
+  opts?: PostAnalyticsEventFnOptions;
+};
+
+export type VoteProps = {
+  id: string;
+};
+
+export type UseVotePost = {
```
rebelchris, line 16: More of a generic question, why do we make a distinguisment between `upvote` and `toggleUpvote` why not one uniform way?

packages/shared/src/hooks/vote/useVotePost.ts
```diff
@@ -0,0 +1,177 @@
...
+      if (!user) {
+        showLogin(AuthTriggers.Upvote);
+
+        return;
+      }
+
+      if (post?.userState?.vote === UserPostVote.Up) {
+        trackEvent(
+          postAnalyticsEvent(AnalyticsEvent.RemovePostUpvote, post, {
+            ...opts,
```
rebelchris, line 109: Would this not mean the `extra` is also inside op these opts already?
Just need to ensure they don't get added twice basically.

packages/shared/src/hooks/vote/useVotePost.ts
```diff
@@ -0,0 +1,177 @@
...
+      );
+
+      await upvotePost({ id: post.id });
+    },
+    [cancelPostVote, showLogin, trackEvent, upvotePost, user, isLoading],
+  );
+
+  const toggleDownvote = useCallback(
+    async ({ post, origin, opts }: ToggleVoteProps) => {
+      if (!post || isLoading) {
```
rebelchris, line 133: did we have the `isLoading` eval on the old system?
Just wondering if it can give bad UX?

packages/shared/src/hooks/vote/utils.ts
```diff
@@ -0,0 +1,129 @@
...
+  }
+
+  const mutationHandler = voteMutationHandlers[vote];
+
+  if (!mutationHandler) {
+    return undefined;
+  }
+
+  const postIndexToUpdate = items.findIndex(
+    (item) => item.type === 'post' && item.post.id === id,
```
rebelchris, line 92: Not sure, but don't we have a type for 'post'

rebelchris (changes requested):
Really amazing refactor.
I do think overall It's still quite hard to follow what happens when you press upvote, but this is more a global tech debt I think.

Would be cool if we could eventually abstract even more so it's one uniform action.
Have some minor questions that you might be able to answer.

packages/shared/src/components/FeedItemComponent.tsx
capJavert: By picking you would get `{ toggleUpvote }` right?

We could do `FeedItemComponentProps & <Pick<UseVotePost, 'toggleUpvote'>`. Did you think that?

packages/shared/src/components/cards/FeedbackCard.tsx
capJavert: Feedback card is only for cards on feed. This is the new component that was added for engagement loops.

packages/shared/src/components/modals/report/ReportPostModal.tsx
capJavert: Agreed, looked at it but currently a lot of tracking hooks have a wide requirements for post type so this is where the problem lies. We need to use `Pick` as much as we can to only require what is needed. Something for a backlog or future refactor for sure. 👍

packages/shared/src/components/post/PostActions.tsx
capJavert: No, this is the block report panel after downvote.

packages/shared/src/components/post/PostItemCard.tsx
capJavert: `origin` is tricky due to `window.origin`

packages/shared/src/hooks/vote/types.ts
capJavert: Because in some places we still need base methods. Companion has a little different implementation and I did not want to break it up in this go. We also might want to just trigger mutation without toggle logic. I think it should be useful still.

packages/shared/src/hooks/vote/utils.ts
capJavert: Not sure what you mean?

packages/shared/src/components/FeedItemComponent.tsx
capJavert: Yes, but you would have to prop drill much more. Using base one allows us to just use without prop requirements and then mutation subscription catches it and invalidates feed cache from there.

I get the point it might be hard to grasp with subscriptions and basically event embitter but I think it works the best with our current cache setup.

For some kind of cache refactor I would advice keeping each post <-> user state in separate queryKey which would then allow invalidation without having access to whole feed or post index in feed array. But that is for some refactor down the line.

packages/shared/src/hooks/vote/useVotePost.ts
capJavert: No, but it is basically the case if someone spams the button which you can do now. Don't have strong preference we can revert.

packages/shared/src/hooks/vote/useVotePost.ts
capJavert: Line below overwrites and merges `opts.extra`

packages/shared/src/components/post/PostItemCard.tsx
capJavert: The component is used for history list and we need the data to manipulate the reading history cache.

Now that you said it I can probably move this inside the `useReadHistoryVotePost` hook.

capJavert: Updated here aab68d6842f9846bb3e1f75330c19d742b143a30

capJavert:
> I do think overall It's still quite hard to follow what happens when you press upvote, but this is more a global tech debt I think.

> Would be cool if we could eventually abstract even more so it's one uniform action.

I agree!

For some kind of cache refactor I would advice keeping each post <-> user state in separate queryKey which would then allow invalidation without having access to whole feed or post index in feed array. But something to note for the future since it would be major work. I'll add some backlog task from open discussions before merge.

packages/shared/src/components/post/PostActions.tsx
```diff
@@ -42,95 +37,27 @@ export function PostActions({
...
-    }
-
-    if (post?.userState?.vote === UserPostVote.Down) {
-      trackEvent(
-        postAnalyticsEvent(AnalyticsEvent.RemovePostDownvote, post, {
-          extra: { origin },
-        }),
-      );
+    await toggleUpvote({ post, origin });
+  };
```
DragosIorgulescu, line 51: Not something to address here, but a more general question on the usage of `useCallback` and `useMemo`. I'm noticing these are usually not used In the project, is this intentional or not considered necessary?

packages/shared/src/hooks/usePostById.ts
```diff
@@ -30,20 +30,28 @@ export const getPostByIdKey = (id: string): QueryKey => [POST_KEY, id];
 export const updatePostCache = (
   client: QueryClient,
   id: string,
-  { id: _, ...postUpdate }: Partial<Post>,
+  postUpdate:
+    | Partial<Omit<Post, 'id'>>
+    | ((current: Post) => Partial<Omit<Post, 'id'>>),
```
DragosIorgulescu, line 35: why is this needed?

DragosIorgulescu (commented):
Nice work 🤩 I really like the cleanup here 

-- not seeing any obvious issues, the only major comment I would have for the future is to look into why there are different query caches being used for the same post resource depending on where it's being loaded

packages/shared/src/components/post/PostActions.tsx
capJavert: We have some guidelines for usage of those in our hooks https://dailydotdev.atlassian.net/wiki/spaces/HAN/pages/248020993/Guidelines#Creating-custom-hooks, this is because hooks are used in multiple places as parts of core logic and we want them to be defined and as optimized as possible. 

While above guidelines can be applied to all components we mostly go by "optimize when needed" to avoid any premature optimizations.

packages/shared/src/hooks/usePostById.ts
capJavert: `setQueryData` can accept a new value or callback which will get the current value and needs to produce a new one. This change adds support for latter while still keeping compatibility with existing usages.

packages/shared/src/hooks/vote/utils.ts
sshanzel: I think this is for FeedItem which has a type: `ad`, `placeholder`, and a `post`.

packages/shared/src/components/FeedItemComponent.tsx
rebelchris: Yes that inded would work right?

packages/shared/src/components/FeedItemComponent.tsx
rebelchris: I see :)

packages/shared/src/components/post/PostItemCard.tsx
rebelchris: But it's not used as invocation right, prefer if we can keep it uniform for now.

packages/shared/src/hooks/vote/types.ts
rebelchris: Ah ok, let's rather refactor then down the line.
Don't want to increate scope, was just curious where we differentiate.

packages/shared/src/hooks/vote/useVotePost.ts
rebelchris: Yes but the above would still contain `extra` as well no?

packages/shared/src/hooks/vote/useVotePost.ts
rebelchris: Hmm let's see it's tricky as this doesn't have any callback to the user they might just be more frustrated lol

packages/shared/src/hooks/vote/utils.ts
rebelchris: I mean a enum typed version of `post` for this context

rebelchris (commented):
the only one is the origin naming, could we perhaps deploy this to preview2 to give it a test run?

packages/shared/src/components/FeedItemComponent.tsx
capJavert: Sure, I'll update.

packages/shared/src/hooks/vote/utils.ts
capJavert: Nope, it's typed as string, see `PostItem` in `useFeed`

packages/shared/src/components/FeedItemComponent.tsx
capJavert: 8d638aaa796be50d41b6a60aebd3c3a5511a1e61

packages/shared/src/components/post/PostItemCard.tsx
capJavert: We already have `analyticsOrigin` in bunch of places, that's why I also used it. Should be good I think.

packages/shared/src/hooks/vote/utils.ts
rebelchris: Ah ok, don't mind my comment then it's not urgent :)

packages/shared/src/hooks/vote/useVotePost.ts
capJavert: To avoid not needed changes I reverted it here 8cdc429a44fdfc03a889e9e89a61646c34bfd92a

packages/shared/src/hooks/vote/useVotePost.ts
capJavert: Refactored here to make it more clear d67d91385a973ddc014012d46f15415419cb544f

rebelchris (approved):
Nice that's amazing and actually can be used as a generic function for analytics opts :)

capJavert:
Logged cache handling refactor potential here https://dailydotdev.atlassian.net/browse/WT-1828 cc: @rebelchris

PR #2166, opened 2023-09-09
refactor: sign back flow

sshanzel:
## Changes
- UI changes relating to the sign back flow.
- The buttons are now all white.
- Display user avatar of the previously logged user.

Preview:

https://github.com/dailydotdev/apps/assets/13744167/33cb9b1f-a3b1-42f6-b9d3-d4ae9f4ad127

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1791 #done


packages/shared/src/styles/components/buttons.css
```diff
@@ -261,3 +261,21 @@
     --button-icon: var(--button-hover-icon);
   }
 }
+
+.btn-signback {
```
sshanzel, line 265: Tried to do it using the Button component, but there were **a lot** of overriding classes and wrapping of components. Which is way more complicated than creating one for the sign back in button.

packages/shared/src/components/auth/AuthContainer.tsx
```diff
@@ -2,10 +2,7 @@ import classNames from 'classnames';
 import React, { forwardRef, HTMLAttributes, MutableRefObject } from 'react';
 import classed from '../../lib/classed';
 
-const Container = classed(
-  'div',
-  'grid grid-cols-1 gap-3 self-center mt-6 w-full',
-);
+const Container = classed('div', 'flex flex-col gap-3 self-center mt-6 w-full');
```
rebelchris, line 5: This one is used quite often right, just something to be aware of :)

packages/shared/src/components/auth/AuthDefault.tsx
```diff
@@ -156,15 +153,16 @@ const AuthDefault = ({
       {!simplified && <AuthModalHeader title={title} />}
```
rebelchris, line 153: This will be a bit refactored with my changes I think :)

packages/shared/src/components/auth/AuthModalFooter.tsx
```diff
@@ -1,34 +1,48 @@
...
       className={classNames(className, 'gap-unset')}
       justify={Justify.Center}
     >
-      <Modal.Text>
-        {isLogin ? 'Don’t have an account?' : 'Already have an account?'}
-      </Modal.Text>
-      <ClickableText
-        className="ml-1 underline text-theme-label-primary"
-        onClick={() => onIsLogin(!isLogin)}
+      <ConditionalWrapper
```
rebelchris, line 26: I think this one I also modified completly

packages/shared/src/components/auth/AuthOptions.tsx
```diff
@@ -175,9 +192,18 @@ function AuthOptions({
     onSuccessfulLogin: onLoginCheck,
     ...(!isTesting && { queryEnabled: !user && isRegistrationReady }),
     trigger,
+    provider: chosenProvider,
   });
   const onProfileSuccess = async () => {
-    await refetchBoot();
+    const { data: boot } = await refetchBoot();
```
rebelchris, line 198: Why do we need both this one as the onValidRegistration?
Seems like we could use one right?

packages/shared/src/components/auth/AuthOptions.tsx
```diff
@@ -258,6 +284,15 @@ function AuthOptions({
       }
 
       if (!e.data?.social_registration) {
+        const { data: boot } = bootResponse;
```
rebelchris, line 287: Oh and here as well.

packages/shared/src/components/buttons/Button.tsx
```diff
@@ -115,7 +133,14 @@ function ButtonComponent<TagName extends AllowedTags>(
     >
       {icon && getIconWithSize(icon)}
       {children && <span className={spanClassName}>{children}</span>}
-      {rightIcon && getIconWithSize(rightIcon)}
+      {rightIcon && rightIcon && (
```
rebelchris, line 136: This one feels a bit weird, is it really needed?

packages/shared/src/styles/components/buttons.css
rebelchris: Hmm that's sad :( 
Maybe once we refactor the button

rebelchris (commented):
Overall there might be some changes with my branch, I'll try and merge ASAP.
But let's see might make things easier for you actually.

packages/shared/src/styles/components/buttons.css
sshanzel: Yeah, hopefully soon, then we can also do the input elements. Lots of things are going on in those components.

packages/shared/src/components/buttons/Button.tsx
sshanzel: Oh, this was some refactored down code. Forgot to remove the other condition.

packages/shared/src/components/auth/AuthOptions.tsx
sshanzel: Yep, we need these calls as our flow sometimes goes through registration flow even if it should be a login flow. However since it was identified to be a registered user, the end result is proving the same. So to catch all ends, we have to cover the callbacks.

packages/shared/src/components/auth/AuthOptions.tsx
sshanzel: Will verify soon which ones we can let go and the ones we should keep.

packages/shared/src/components/auth/AuthOptions.tsx
```diff
@@ -63,6 +69,7 @@ export enum AuthDisplay {
 
 export interface AuthOptionsProps {
   onClose?: CloseAuthModalFunc;
+  onAuthStateUpdate?: (isLoginFlow: boolean) => void;
```
capJavert, line 72: `{ isLoginFlow }` to allow futher customization?

packages/webapp/pages/onboarding.tsx
```diff
@@ -215,6 +215,9 @@ export function OnboardPage(): ReactElement {
               onSuccessfulRegistration={onSuccessfulTransaction}
               isLoginFlow={isLoginFlow}
               className={classNames('w-full', maxAuthWidth)}
+              onAuthStateUpdate={(isLogin) =>
+                setAuth({ isLoginFlow: isLogin, isAuthenticating: true })
```
capJavert, line 219: If `isLogin` is false do we then still set `isAuthenticating` to `true`?

capJavert (commented):
Few comments for sanity checking, overall looks good. 👍

packages/webapp/pages/onboarding.tsx
sshanzel: It can be registration which is still under authenticating, so both are always true for that property.

packages/shared/src/components/auth/AuthOptions.tsx
sshanzel: Good point. Let's do that.

packages/shared/src/components/auth/AuthDefault.tsx
```diff
@@ -156,15 +153,16 @@ const AuthDefault = ({
...
-          {providers.map(({ provider, ...props }) => (
-            <ProviderButton
+          {providers.map(({ provider, icon }) => (
+            <Button
               key={provider}
-              provider={provider}
-              label={shouldLogin ? 'Log in with' : 'Sign up with'}
-              onClick={() => onSocialClick(provider.toLowerCase())}
-              loading={!isReady}
-              {...props}
```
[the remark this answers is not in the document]
sshanzel: The loading is required indeed. The props should not anymore. Good catch.

packages/shared/src/components/modals/common/ModalHeader.tsx
```diff
@@ -42,6 +42,7 @@ export function ModalHeader({
   return (
     <ModalHeaderOuter
       className={classNames(
+        'relative',
```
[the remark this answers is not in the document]
sshanzel: All other places are broken actually, this fixes them 😅 

The close button should be relative to the header when you scroll, without this it overlaps the elements.

Will also keep an eye out to see any weird behaviors after this.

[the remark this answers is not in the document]
capJavert: Oh 😆 nothing to add then 👏

packages/shared/src/components/auth/AuthContainer.tsx
sshanzel: Yep. The goal was to have a column, hence, flex-col should be alright.

packages/shared/src/components/auth/AuthSignBack.tsx
```diff
@@ -1,72 +1,79 @@
...
+      <AuthContainer className="items-center">
+        <p className="mb-2 text-center typo-callout text-theme-label-secondary">
+          Log in to access your account
         </p>
-        <ProviderButton
-          provider={signback.provider}
-          label="Login with"
-          onClick={() => onProviderClick?.(signback.provider)}
-          {...signback}
+        <ConditionalWrapper
```
rebelchris, line 41: Wait this shouldn't be conditionally wrapped right, rather conditionally rendered?
We should not show this part if you are on email login flow?

packages/shared/src/components/buttons/Button.tsx
```diff
@@ -115,7 +127,14 @@ function ButtonComponent<TagName extends AllowedTags>(
...
       {children && <span className={spanClassName}>{children}</span>}
-      {rightIcon && getIconWithSize(rightIcon)}
+      {rightIcon && (
+        <ConditionalWrapper
+          condition={rightIconWrapper?.condition}
+          wrapper={rightIconWrapper?.wrapper}
+        >
+          {getIconWithSize(rightIcon)}
+        </ConditionalWrapper>
+      )}
```
rebelchris, line 137: I still think this is a little bit messy to be honest.
It feels like it abuses the conditional wrapper, what part is this actually needed for?
Maybe we can find a cleaner implementation or revert to rendering it in the component rather.

packages/shared/src/hooks/auth/useSignBack.ts
```diff
@@ -0,0 +1,44 @@
...
+import { SocialProvider } from '../../components/auth/common';
+import { LoggedUser } from '../../lib/user';
+
+export const SIGNIN_METHOD_KEY = 'signin_method';
+
+export const signBackKey = generateQueryKey(
+  RequestKey.Auth,
+  null,
+  SIGNIN_METHOD_KEY,
+).join(':');
```
rebelchris, line 13: Why do we use the join here? 👀

rebelchris: It's a react query key, used for something else, I would suggest either making a new function or finding another solution.
Try to keep things seperate

packages/shared/src/components/ConditionalWrapper.tsx
```diff
@@ -1,6 +1,6 @@
 import { ReactElement, ReactNode } from 'react';
 
-interface ConditionalWrapperProps {
+export interface ConditionalWrapperProps {
```
rebelchris, line 3: I think ideally this wouldn't be needed to be exported.

rebelchris (changes requested):
Wow amazing refactor, almost there I think!

packages/shared/src/components/auth/AuthSignBack.tsx
sshanzel: We need the profile picture for both ends. It will need a wrapper conditionally to display the social provider icon.

packages/shared/src/components/buttons/Button.tsx
sshanzel: This was some unreverted changes. This was part of using the button before the custom css. Shall revert this.

packages/shared/src/components/auth/AuthSignBack.tsx
rebelchris: Oh I see, ok that makes sense, but does it not render completely different?
Like one is in a button, and one is just the big version of profile image?

sshanzel: No, just really needs the wrapper with a relative position for the absolute placement of the icon.

packages/shared/src/hooks/auth/useSignBack.ts
sshanzel: This is also a key to be used for storage. What it accepts is a string and to keep the keys uniform, reused the function.

rebelchris: Maybe we can duplicate the function and name it explicitly and do the join in the function?
Will be more uniform moving forward for storage keys.

WDYT?

packages/shared/src/components/auth/AuthSignBack.tsx
rebelchris: Ah ok, closing this as resolved then.

packages/shared/src/hooks/auth/useSignBack.ts
sshanzel: Good point! Let's do that.

rebelchris (approved):
Alright looks good.
As it's a critical flow as well, please have one more review and perform critical flow tests once merged to main with updated test line.

PR #2177, opened 2023-09-12
fix: use the email the user filled in on the forgot password form

DragosIorgulescu:
## Changes

prepopulate-pass-forgot-form

## Events

_N/A_

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1789 #done


packages/shared/src/components/auth/AuthDefault.tsx
```diff
@@ -83,6 +85,15 @@ const AuthDefault = ({
...
   }, [shouldLogin]);
 
+  const getFormEmail = useCallback((e: React.FormEvent) => {
+    const form = e.currentTarget as HTMLFormElement;
+    const input = Array.from(form.elements).find(
+      (el) => el.getAttribute('name') === 'email',
+    ) as HTMLInputElement;
+
+    return input?.value?.trim();
+  }, []);
```
DragosIorgulescu, line 95: I extracted this because I initially wanted to use the same pattern, but turned out I couldn't.

DragosIorgulescu:
@dailydotdev/web-team I feel like there are some antipatterns used here, as well as that the auth components seem a bit more complex than they need to be. 

For example, I was trying to use the existing `email` inner state from `AuthOptions` to pass to the forgot password component, but its value seems to only be set upon form submission -- and on form submission, the value is taken directly from the input. 

I am most likely missing some reasoning behind this, but it feels to me like the auth email input should be controlled input, with a value and the `onChange` handler modifying that value

rebelchris:
@DragosIorgulescu Auth is very complex and definitely can use with refactor.
That's actually part of the sprint for papercuts (tech debt)
However it's a complex and very sensitive system (+ critical flow). so we have been very careful with it.

I think most complexity happened because of the many different versions and states (modal / onboarding page/ old onboarding flow / etc)
Now it's a bit more smaller, and personally I think we can abstract the wrapping layer a bit to make that contain some state, which would probably help make this clearer as well.

Hope this gives a little bit of background, happy to hear if you have specific tech debt for auth you'd like to see refactored?

packages/shared/src/components/auth/AuthDefault.tsx
rebelchris: We do have `registerEmail` could we not use that state to sync it?

packages/shared/src/components/auth/LoginForm.tsx
```diff
@@ -73,10 +74,15 @@ function LoginForm({
         name="identifier"
         label="Email"
         type="email"
-        value={email}
+        defaultValue={email}
```
rebelchris, line 77: I think you still need value as well right to sync the change?

packages/shared/src/components/auth/LoginForm.tsx
```diff
@@ -94,7 +100,7 @@ function LoginForm({
           <ClickableText
             type="button"
             className="flex-1 underline grow-[2] btn-primary"
-            onClick={onForgotPassword}
+            onClick={() => onForgotPassword(innerEmail)}
```
rebelchris, line 103: I think if you use value still, we can leave this was it was?

rebelchris (commented):
Would it be possible to uniform the existing state we have?

packages/shared/src/components/auth/AuthDefault.tsx
DragosIorgulescu: `registerEmail` is being set on submission the same way as `setEmail`, so we can't really use it as it will be undefined when clicking on forgot password link

packages/shared/src/components/auth/LoginForm.tsx
DragosIorgulescu: good catch, I was testing something when I changed this, rolled back now

packages/shared/src/components/auth/LoginForm.tsx
DragosIorgulescu: The `email` coming from props here will be undefined until the form is submitted, so we need to keep track of it separately for the forgot password handler

packages/shared/src/components/auth/AuthDefault.tsx
rebelchris: Ah yes, but I mean reuse the setter from the place where you do it now?
And simply pass that with prop drilling? (just so it's one less abstraction)

DragosIorgulescu:
> @DragosIorgulescu Auth is very complex and definitely can use with refactor. That's actually part of the sprint for papercuts (tech debt) However it's a complex and very sensitive system (+ critical flow). so we have been very careful with it.
> 
> I think most complexity happened because of the many different versions and states (modal / onboarding page/ old onboarding flow / etc) Now it's a bit more smaller, and personally I think we can abstract the wrapping layer a bit to make that contain some state, which would probably help make this clearer as well.
> 
> Hope this gives a little bit of background, happy to hear if you have specific tech debt for auth you'd like to see refactored?

@rebelchris thank you, this makes total sense. I don't think I fully understand all the flows here yet to propose a refactoring path with confidence, but one area I think could use improvement is how we keep track of the input values and how we are passing them around between these components

packages/shared/src/components/auth/AuthDefault.tsx
DragosIorgulescu: my bad, misunderstood your point -- so you would like to pass the `setRegisterEmail` or `setEmail` as a prop to `LoginForm`

rebelchris: Or basically trigger the `onForgotPassword` inside `LoginForm` to pass that internal email?
Then you only need to add `email` to the callback?

packages/shared/src/components/auth/AuthDefault.tsx
```diff
@@ -39,13 +40,14 @@ interface AuthDefaultProps extends AuthFormProps {
   isLoading?: boolean;
   isReady: boolean;
   loginButton?: string;
+  initialEmail?: string;
```
capJavert, line 43: This is not used currently right?

packages/shared/src/components/auth/AuthDefault.tsx
DragosIorgulescu: Ah, I see, I'm already doing that here (if I understand you correctly): https://github.com/dailydotdev/apps/pull/2177/files#diff-89ed52a0b28850e8d234eccd3a777f6218a989b38f297dd8e2c46387fc8d5702R103

This callback `getFormEmail` has nothing to do with the forgot password stuff, I just extracted it out of the `onEmailSignup` handler because I was planning on reusing it, but ended up not needing it.

rebelchris: Oh sorry yeah misunderstood your changes then :)

packages/shared/src/components/auth/AuthDefault.tsx
```diff
@@ -39,13 +40,14 @@ interface AuthDefaultProps extends AuthFormProps {
...
 }
 
 const AuthDefault = ({
   loginHint,
   onSignup,
   onProviderClick,
-  onForgotPassword,
+  onForgotPassword: handleForgotPassword,
```
rebelchris, line 50: Why do we need to rename it here? can we not use as is?

rebelchris (commented):
Alright one last change

packages/shared/src/components/auth/LoginForm.tsx
```diff
@@ -48,6 +48,7 @@ function LoginForm({
     onPasswordLogin(form);
   };
   const [shouldFocus, setShouldFocus] = useState(autoFocus);
+  const [innerEmail, setInnerEmail] = useState(email || '');
```
capJavert, line 51: I was thinking, we should be able to just use `ref` on `TextField` below and then just use `textFieldRef.current.value` when calling `onForgotPassword` below. It's clunky but at least we don't introduce additional state just for the sake of it and since field is uncontrolled.

packages/shared/src/components/auth/AuthDefault.tsx
capJavert: This can also go outside of component since it does not require any state.

capJavert (commented):
I think you summed it up as smallest footprint as you could in this situation 🤘 

Got few ideas but nothing blocking. I also noticed we do lose email state when we go back to login. I know it's not part of this task but maybe also log that @rebelchris ?

packages/shared/src/components/auth/AuthDefault.tsx
DragosIorgulescu: Good catch, removed 👍

DragosIorgulescu:
> I think you summed it up as smallest footprint as you could in this situation 🤘
> 
> Got few ideas but nothing blocking. I also noticed we do lose email state when we go back to login. I know it's not part of this task but maybe also log that @rebelchris ?

@capJavert I tried adding that too 😅 but my attempts so far have not been successful, no matter how I try to pass it around, the email input gets cleared when going back to login -- this is what prompted me to suggest looking into how these input values are tracked in this area

packages/shared/src/components/auth/AuthDefault.tsx
DragosIorgulescu: good catch 👍  leftover code from previous attempts, removed

packages/shared/src/components/auth/LoginForm.tsx
capJavert: I know `TextField` has some ghosts so if you already tried it let's leave it as it is.

rebelchris (approved):
Good from my side for now 
Thanks for exposing some pain points, we can definitely improve auth a lot!
(And forms)

capJavert (approved):
@DragosIorgulescu no need to include login email state in this PR but its appreciated if you can log it. 🙏

DragosIorgulescu:
> @DragosIorgulescu no need to include login email state in this PR but its appreciated if you can log it. 🙏

Added ticket here: https://dailydotdev.atlassian.net/browse/WT-1831

PR #2179, opened 2023-09-12
fix: verify session modal

sshanzel:
## Changes
- Fix verify session modal.
- Use lazy modal (so much cleaner without manually managing the state).

Deployed at preview.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done

### DIFF
--- a/packages/shared/src/components/auth/AuthDefault.tsx
+++ b/packages/shared/src/components/auth/AuthDefault.tsx
@@ -165,16 +165,16 @@ const AuthDefault = ({
       )}
       <AuthContainer className={disableRegistration && 'mb-6'}>
         <div className="flex flex-col gap-4">
-          {providers.map(({ provider, icon }) => (
+          {providers.map(({ label, value, icon }) => (
             <Button
-              key={provider}
+              key={label}
               icon={icon}
               className="btn-primary"
               buttonSize={ButtonSize.Large}
-              onClick={() => onSocialClick(provider.toLowerCase())}
+              onClick={() => onSocialClick(value)}
               loading={!isReady}
             >
-              {provider}
+              {label}
             </Button>
           ))}
         </div>
--- a/packages/shared/src/components/auth/AuthDefault.tsx
+++ b/packages/shared/src/components/auth/AuthDefault.tsx
@@ -165,7 +165,7 @@ const AuthDefault = ({
               icon={icon}
               className="btn-primary"
               buttonSize={ButtonSize.Large}
-              onClick={() => onSocialClick(provider)}
+              onClick={() => onSocialClick(provider.toLowerCase())}
               loading={!isReady}
             >
               {provider}
--- a/packages/shared/src/components/auth/AuthHeader.tsx
+++ b/packages/shared/src/components/auth/AuthHeader.tsx
@@ -1,4 +1,9 @@
-import React, { ReactElement } from 'react';
+import React, {
+  FormEvent,
+  MouseEvent,
+  KeyboardEvent,
+  ReactElement,
+} from 'react';
 import { Button } from '../buttons/Button';
 import ArrowIcon from '../icons/Arrow';
 import { Modal } from '../modals/common/Modal';
@@ -8,7 +13,7 @@ export interface AuthHeaderProps {
   simplified: boolean;
   title: string;
   className?: string;
-  onBack?: () => void;
+  onBack?: (e: MouseEvent | KeyboardEvent | FormEvent) => void;
 }
 
 function AuthHeader({
--- a/packages/shared/src/components/auth/AuthHeader.tsx
+++ b/packages/shared/src/components/auth/AuthHeader.tsx
@@ -1,7 +1,6 @@
 import React, { ReactElement } from 'react';
 import { Button } from '../buttons/Button';
 import ArrowIcon from '../icons/Arrow';
-import { CloseModalFunc } from '../modals/common';
 import { Modal } from '../modals/common/Modal';
 import { ModalHeaderKind } from '../modals/common/types';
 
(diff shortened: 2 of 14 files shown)


packages/shared/src/components/auth/AuthHeader.tsx
```diff
@@ -1,15 +1,19 @@
...
 import ArrowIcon from '../icons/Arrow';
-import { CloseModalFunc } from '../modals/common';
 import { Modal } from '../modals/common/Modal';
 import { ModalHeaderKind } from '../modals/common/types';
 
 export interface AuthHeaderProps {
   simplified: boolean;
   title: string;
   className?: string;
-  onBack?: CloseModalFunc;
```
rebelchris, line 12: Do we still use this somewhere?

rebelchris (approved):
Overall super clean!
Cleans up quite a lot with this modal type 🙌🤘

Lets do two reviews on this 🙏

packages/shared/src/components/auth/AuthHeader.tsx
sshanzel: I just manually typed the prop to avoid the deps cycle being thrown, but I believe yes, we still use this property and type for simplicity of declaring the callback.

sshanzel:
Hold still for a bit, still doing some refactor.

packages/shared/src/components/auth/AuthHeader.tsx
capJavert: I noticed in my vote refactor few deps cycles creeping up, we might wanna adopt `types.ts` as separate file for types to avoid it and also use `import type` vs just `import`. We can discuss on team meet?

sshanzel: Yep. A good discussion to tackle in the Gathering/web team meeting. Have had this issue a few times already too.

packages/shared/src/hooks/useLogin.ts
```diff
@@ -55,18 +54,13 @@ const useLogin = ({
   onSuccessfulLogin,
   queryEnabled = true,
   queryParams = {},
-  enableSessionVerification = false,
```
sshanzel, line 58: As we can see, the session is optional only from this point, hence, I think it would be better to have it just a prop since it is not necessary on all login flows.

Instead, the security page needs the session all the time, so it would be better to instantiate it there.

packages/shared/src/hooks/useLogin.ts
```diff
@@ -55,18 +51,13 @@ const useLogin = ({
...
+  session,
```
sshanzel, line 54: Also, I am actually in a dilemma with how we use react query. Do we pass this value as a prop or do we fetch this through `client.getQueryData`? Thoughts?

sshanzel:
Changes:
- Unified query keys.
- Reduced responsibility of some hooks.
- Removed unused returned properties from hooks.

Feel free to drop your comments! 🚀

packages/shared/src/components/auth/AuthDefault.tsx
```diff
@@ -171,7 +171,7 @@ const AuthDefault = ({
               icon={icon}
               className="btn-primary"
               buttonSize={ButtonSize.Large}
-              onClick={() => onSocialClick(provider)}
+              onClick={() => onSocialClick(provider.toLowerCase())}
```
capJavert, line 174: Can we maybe lowercase providers in `getKratosProviders` once so we don't have to do it all over the place?

packages/shared/src/hooks/useLogin.ts
```diff
@@ -209,7 +197,7 @@ const useLogin = ({
     }),
     // @NOTE see https://dailydotdev.atlassian.net/l/cp/dK9h1zoM
     // eslint-disable-next-line react-hooks/exhaustive-deps
-    [session, queryEnabled, queryParams, hint, login, isLoading],
+    [queryEnabled, queryParams, hint, login, isLoading],
```
capJavert, line 200: Not blocking but since you did refactor things here might be good idea to also remove `useMemo` and add `useCallback` for `onSubmitSocialLogin` and `onSubmitPasswordLogin`

capJavert (commented):
Few comments, other then that looks good to go.

packages/webapp/pages/account/security.tsx
```diff
@@ -110,11 +110,15 @@ const AccountSecurityPage = (): ReactElement => {
...
 
         setActiveDisplay(Display.Default);
-        await refetchSession();
+        await client.invalidateQueries(sessionKey);
       },
     },
   );
 
+  const { data: session } = useQuery(
+    generateQueryKey(RequestKey.CurrentSession, null),
```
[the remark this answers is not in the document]
sshanzel: Good point. It was not part of the original query, but I agree, better to have it now.

packages/shared/src/hooks/useLogin.ts
sshanzel: Oh, yep, I agree. I have cleared this part. Seems my commit did not push through. Let me push it properly 😅

packages/shared/src/components/auth/AuthDefault.tsx
sshanzel: They are in lowercase from the requests, the issue is, on some other parts, the label being displayed is used (for example, Google on the button).

rebelchris: I agree best to have them lowercase as values and upper them in usecases where needed maybe.
Or have both as value and label?

sshanzel: Yeah, makes sense to have both properties so we will never be wrong on the actual value.

sshanzel: I've also refactored the property `provider` to `label` for easier understanding.

sshanzel:
Tested all flows on preview and everything seems to work.

PR #2182, opened 2023-09-13
feat: onboarding v3

omBratteng:

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1673 #done


packages/webapp/pages/onboarding.tsx
```diff
@@ -48,13 +50,20 @@ const versionToTitle: Record<OnboardingFilteringTitle, string> = {
   [OnboardingFilteringTitle.V4]: 'Choose the topics you’re passionate about',
 };
 
-const Title = classed('h2', 'font-bold typo-title2');
+const Title = classed('h2', 'font-bold typo-title1');
```
rebelchris, line 53: One important nuance I haven't seen here is that it's an experiment so it should be able to serve a "normal" flow and this new and improved one based on a feature flag.

packages/webapp/pages/onboarding.tsx
```diff
@@ -176,44 +184,58 @@ export function OnboardPage(): ReactElement {
...
+                />
+                <source
+                  src={cloudinary.onboarding.video.webm}
+                  type="video/webm"
+                />
+              </video>
             }
-          />
-        </div>
+            <p className="relative z-3 mb-6 text-center opacity-64 typo-callout text-theme-label-quaternary">
```
rebelchris, line 236: We normally don't use opacity for text styling.
I do think design theme avoided our guidelines here

packages/webapp/pages/onboarding.tsx
```diff
@@ -48,13 +50,20 @@ const versionToTitle: Record<OnboardingFilteringTitle, string> = {
...
-  'flex flex-col overflow-x-hidden items-center min-h-[100vh] w-full h-full max-h-[100vh] flex-1 z-max bg-theme-bg-primary',
+  'flex flex-col overflow-x-hidden items-center min-h-[100vh] w-full h-full max-h-[100vh] flex-1 z-max bg-[#0A1119]',
+);
+
+const OnboardingTitle = classed(
+  'h1',
+  'mb-4 font-bold text-transparent bg-clip-text bg-gradient-to-r from-theme-color-bacon to-theme-color-cabbage',
 );
 
 const seo: NextSeoProps = {
```
rebelchris, line 69: In general this page is getting very bloated, maybe we should consider moving some stuff to smaller components to break it down a bit better?

packages/webapp/pages/onboarding.tsx
```diff
@@ -233,19 +255,103 @@ export function OnboardPage(): ReactElement {
...
+              <SignupDisclaimer className="mb-auto tablet:mb-0" />
+
+              <div className="block tablet:hidden">
+                <TrustedCompanies iconSize={IconSize.Small} />
+
+                <p className="relative z-3 mb-6 text-center opacity-64 typo-callout text-theme-label-quaternary">
+                  Trusted by 300K+ developers from the world&apos;s leading
+                  companies
+                </p>
+              </div>
```
rebelchris, line 341: Why do we duplicate this section?

packages/eslint-rules/lib/rules/no-custom-color.js
```diff
@@ -9,6 +9,7 @@ const alwaysValidClasses = [
   'text-lg',
   'text-xl',
   'text-ellipsis',
+  'bg-clip-text',
```
rebelchris, line 12: Where do we use this one?

rebelchris (changes requested):
Amazing work, have some minor comments on some nuances from our particular code base.
I'll test it more on UX level once you give the go 🙏

packages/eslint-rules/lib/rules/no-custom-color.js
omBratteng: On the title, to get the gradient effect.
https://github.com/dailydotdev/apps/blob/fae17e358f191ec7360cfe205ea1185bd8c884d4/packages/webapp/pages/onboarding.tsx#L64-L67

packages/webapp/pages/onboarding.tsx
omBratteng: Just refactored it into the `TrustedCompanies` component.

packages/webapp/pages/onboarding.tsx
omBratteng: Did not intend to use opacity, but to get the correct text colour

packages/shared/src/components/TrustedCompanies.tsx
```diff
@@ -0,0 +1,78 @@
...
+  className = '',
+  iconSize = IconSize.Large,
+}: TrustedCompaniesProps): ReactElement => {
+  return (
+    <div className={classNames('flex relative z-3 justify-center', className)}>
+      {Object.values(trustedCompaniesMap).map(({ Icon, label }, index) => (
+        <Icon
+          key={`trusted-company-${label}`}
+          className={classNames(
+            'text-salt-90 opacity-32',
```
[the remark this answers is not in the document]
omBratteng: Fixed this one

packages/shared/src/components/auth/EmailSignupForm.tsx
```diff
@@ -7,14 +8,52 @@ import AuthForm from './AuthForm';
 import { privacyPolicy, termsOfService } from '../../lib/constants';
 import { IconSize } from '../Icon';
 
+interface SignupDisclaimerProps {
+  className?: string;
+}
+export function SignupDisclaimer({
```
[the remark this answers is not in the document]
omBratteng: Moved it out

packages/shared/src/components/auth/OnboardingRegistrationForm.tsx
```diff
@@ -0,0 +1,156 @@
...
+}: OnboardingRegistrationFormProps): ReactElement => {
+  const { trackEvent } = useContext(AnalyticsContext);
+  const [shouldLogin, setShouldLogin] = useState(false);
+  // const { displayToast } = useToastNotification();
+  const [registerEmail, setRegisterEmail] = useState<string>(null);
+  const { mutateAsync: checkEmail } = useMutation((emailParam: string) =>
+    checkKratosEmail(emailParam),
+  );
+
+  useEffect(() => {
```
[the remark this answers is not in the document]
omBratteng: I think that sounds like a good idea. I do feel like the entire auth could get a second look at see what is needed and not needed anymore too.

packages/shared/src/components/auth/OnboardingRegistrationForm.tsx
```diff
@@ -0,0 +1,156 @@
...
+        ? AuthEventNames.OpenLogin
+        : AuthEventNames.OpenSignup,
+      extra: JSON.stringify({ trigger }),
+      target_id: targetId || ExperimentWinner.AuthVersion,
+    });
+    // @NOTE see https://dailydotdev.atlassian.net/l/cp/dK9h1zoM
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [shouldLogin]);
+
+  const getFormEmail = useCallback((e: React.FormEvent) => {
```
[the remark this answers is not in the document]
omBratteng: Yeah, I based it on the authDefault component

packages/shared/src/components/auth/OnboardingRegistrationForm.tsx
```diff
@@ -0,0 +1,156 @@
...
+        </Button>
+      </AuthForm>
+
+      <OrDivider className="mb-8" label="Or sign up with" />
+
+      <div className="flex gap-8">
+        {signupProviders.map(({ provider, icon }) => (
+          <Button
+            key={provider}
+            className="flex flex-1 text-white bg-theme-active border-none btn-primary !justify-center"
```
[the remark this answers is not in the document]
omBratteng: I removed the `btn-primary` and some other that are redundant. Based off of theme it is `text-white` and `bg-theme-active`. Will update if design changes.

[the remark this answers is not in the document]
rebelchris: That shouldn't be right, ill check with design because we have only limited buttons options 🫣

packages/shared/src/components/auth/OnboardingRegistrationForm.tsx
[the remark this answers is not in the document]
rebelchris: Yes 100% it was very big because of all the different experiment implementations.
I think in the current scope we can downsize the logic quite a bit.

packages/webapp/pages/onboarding.tsx
rebelchris: It should simply be quarternary without the extra opacity according to Tsahi.
Think the design just got overwritten.

https://dailydotdev.slack.com/archives/C05P8U95VFT/p1694669630642729?thread_ts=1694611265.506569&cid=C05P8U95VFT

omBratteng: Already changed it to it in 6e6d527

packages/webapp/pages/onboarding.tsx
omBratteng: I am working on adding it as an experiment

packages/webapp/pages/onboarding.tsx
```diff
@@ -233,18 +259,85 @@ export function OnboardPage(): ReactElement {
...
+            </OnboardingTitle>
+
+            <h2 className="mb-8 typo-body tablet:typo-title2">
+              Get one personalized feed for all the knowledge you need.
+            </h2>
+
+            {getAuthOptions(AuthDisplay.OnboardingSignup)}
+
+            <div className="flex relative flex-col tablet:flex-row flex-1 justify-end tablet:items-center pb-6 tablet:pb-0 mt-8 tablet:mt-auto w-full tablet:max-h-[10rem]">
+              <SignupDisclaimer className="mb-auto tablet:mb-0" />
```
rebelchris, line 324: In this section the signup disclaimer is not clickable.

omBratteng: Fixed it, just z-index on the image

capJavert (commented):
Nothing to add at this point, seems solid, gonna do another round when experimentation is added.

packages/shared/src/components/auth/OnboardingRegistrationForm.tsx
```diff
@@ -0,0 +1,155 @@
...
+    ) as HTMLInputElement;
+
+    return input?.value?.trim();
+  }, []);
+
+  const onEmailSignup = async (e: React.FormEvent) => {
+    e.preventDefault();
+
+    trackEvent({
+      event_name: 'click',
```
rebelchris, line 74: This should be from AuthEventNames

packages/shared/src/components/auth/SignupDisclaimer.tsx
```diff
@@ -0,0 +1,42 @@
+import React, { ReactElement } from 'react';
+import classNames from 'classnames';
+import { privacyPolicy, termsOfService } from '../../lib/constants';
+
+interface SignupDisclaimerProps {
+  className?: string;
+}
+
+function SignupDisclaimer({
+  className = '',
```
rebelchris, line 10:
```suggestion
  className,
```

packages/webapp/pages/onboarding.tsx
```diff
@@ -134,23 +210,31 @@ export function OnboardPage(): ReactElement {
...
         mandating_categories: 0,
       }),
     });
     isTracked.current = true;
-  }, [trackEvent, isPageReady, onboardingV2, router, user]);
+  }, [
+    trackEvent,
+    isPageReady,
+    onboardingV2,
+    router,
```
capJavert, line 233: I know `router` was here before but it has an unstable reference in Next 12.x so might be better to just remove it for now until we upgrade.

capJavert: Ref: https://github.com/vercel/next.js/issues/18127

packages/shared/src/components/auth/OnboardingRegistrationForm.tsx
```diff
@@ -0,0 +1,155 @@
...
+    }
+
+    return onSignup(email);
+  };
+
+  const onSocialClick = (provider: string) => {
+    onProviderClick?.(provider, shouldLogin);
+  };
+
+  const signupProviders = [providerMap.google, providerMap.github];
```
capJavert, line 99: We can pull this out of the component since it's basically a const.

packages/shared/src/components/auth/OnboardingRegistrationForm.tsx
```diff
@@ -0,0 +1,155 @@
...
+      target_type: AuthEventNames.SignUpProvider,
+      target_id: 'email',
+      extra: JSON.stringify({ trigger }),
+    });
+
+    const email = getFormEmail(e);
+
+    if (!email) {
+      return null;
+    }
```
capJavert, line 84: Do we (need to) validate this case on form?

packages/shared/src/components/auth/OnboardingRegistrationForm.tsx
```diff
@@ -0,0 +1,155 @@
...
+    const form = e.currentTarget as HTMLFormElement;
+    const input = Array.from(form.elements).find(
+      (el) => el.getAttribute('name') === 'email',
+    ) as HTMLInputElement;
+
+    return input?.value?.trim();
+  }, []);
+
+  const onEmailSignup = async (e: React.FormEvent) => {
+    e.preventDefault();
```
capJavert, line 71: Maybe also do a early return if `checkEmail` mutation is already loading?

packages/shared/src/components/auth/OnboardingRegistrationForm.tsx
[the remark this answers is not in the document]
capJavert: This can easily go to util/helper 👍

capJavert (changes requested):
Few comments/ideas.

packages/shared/src/components/auth/SignupDisclaimer.tsx
omBratteng: Removed

packages/shared/src/components/auth/OnboardingRegistrationForm.tsx
omBratteng: Moved out

packages/shared/src/components/auth/OnboardingRegistrationForm.tsx
omBratteng: Which one? I'll also update it in AuthDefault as well?

packages/shared/src/components/auth/OnboardingRegistrationForm.tsx
omBratteng: Now with early return

packages/webapp/pages/onboarding.tsx
omBratteng: Removed and added eslint ignore

packages/webapp/pages/onboarding.tsx
rebelchris: @omBratteng I think this still has a impact on the existing v2 right?

packages/webapp/pages/onboarding.tsx
```diff
@@ -67,26 +80,82 @@ const seo: NextSeoProps = {
   ...defaultSeo,
 };
 
+interface HeaderProps {
+  showOnboardingPage: boolean;
+  isOnboardingV3: boolean;
+  setAuth: React.Dispatch<React.SetStateAction<AuthProps>>;
```
rebelchris, line 86:
```suggestion
  setAuth: Dispatch<SetStateAction<AuthProps>>;
```

rebelchris (changes requested):

packages/webapp/pages/onboarding.tsx
omBratteng: It is set to `typo-title2` in `getContent` (V2), so it still had the same size as before. But refactored it a bit so V2 doesn't get both classnames and removed duplicate `font-bold` classname.

packages/webapp/pages/onboarding.tsx
omBratteng: Replaced

capJavert (commented):
Don't see anything else, probably just need to test it out now 👍

packages/shared/src/components/auth/OnboardingRegistrationForm.tsx
```diff
@@ -0,0 +1,144 @@
...
+  useEffect(() => {
+    trackEvent({
+      event_name: shouldLogin
+        ? AuthEventNames.OpenLogin
+        : AuthEventNames.OpenSignup,
+      extra: JSON.stringify({ trigger }),
+      target_id: targetId || ExperimentWinner.AuthVersion,
+    });
+    // @NOTE see https://dailydotdev.atlassian.net/l/cp/dK9h1zoM
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
DragosIorgulescu, line 54: should we just remove this rule entirely? I've rarely found it to be useful, since usually you know exactly what dependencies you need for a side effect or callback, and you rarely need to specify the full list

packages/shared/src/components/icons/Amazon/index.tsx
DragosIorgulescu, line 1: It feels like we could create a generic component for these that accepts the icon name as a prop and loads the appropriate SVG -- with dynamic loading of the SVG to avoid bundling things we don't need.

packages/webapp/pages/onboarding.tsx
```diff
@@ -117,15 +196,23 @@ export function OnboardPage(): ReactElement {
     router.push('/');
   };
 
+  const onSuccessfulLogin = () => {
+    setFinishedOnboarding(true);
+    router.push('/');
```
DragosIorgulescu, line 201: should this be a replace? do we want the user to be able to navigate back to the onboarding screen after logging in?
```suggestion
    router.replace('/');
```

DragosIorgulescu (commented):
great work overall 🙌  -- I posted some questions, mostly for my curiosity, not blockers

idoshamun (changes requested):
Don't merge while drinking

idoshamun (approved):
Danger is over 🤣

packages/webapp/pages/onboarding.tsx
```diff
@@ -67,27 +82,90 @@ const seo: NextSeoProps = {
...
+  const formRef = useRef<HTMLFormElement>();
+  const title = versionToTitle[filteringTitle];
+
+  useEffect(() => {
+    if (isOnboardingV3) {
+      setAuth({ ...auth, defaultDisplay: AuthDisplay.OnboardingSignup });
+    }
+    // Ignore the auth prop to avoid infinite loop
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [isOnboardingV3]);
```
capJavert, line 168: You can use `setAuth(currentAuth => ({ ...currentAuth }))` to avoid breaking rules of hooks.

capJavert (changes requested):
Few visual issues on slack, but other then that I think logic is solid. Nearly there.

packages/webapp/pages/onboarding.tsx
omBratteng: I've changed this on those created by V3, but should I also do it on the ones for V2 @rebelchris?

rebelchris: You should not be able to navigate back technically.
For v2 it will hard force you back to `/` anyway. *(it's existing logic)

So we can simply add this in v2 as well yes.

omBratteng: Done.

packages/webapp/pages/onboarding.tsx
omBratteng: Doh! Of course. I've been working too much with k8s, so I had forgotten that method 😬 
Fixed now.

packages/shared/src/components/icons/Amazon/index.tsx
omBratteng: I'd say out of scope for this, maybe re-work how we do all of our icons, because they are more or less all the same code just different names.
cc @rebelchris

rebelchris: Yeah think we actually already have a SVG optimisation task derived from the react migration.

idoshamun:
@omBratteng I suggest a different approach for experimentation here. Right now, you assume onboarding v2 and v3 are independent. A better approach will be to assume that v3 is available only to v2 users, which is true. This way, we don't have to check for v3 on a layout level only on the onboarding page, allowing us to roll out v3 without an extension version. How does it sound?

rebelchris:
@idoshamun Just one nuance there is that users might experience a flicker.
Meaning:

- Evaluate on feed: onboarding v2 is recognised
- Redirect to /onboarding (initially showing v2)
- Evaluating onboarding v3
- Onboarding v3 is recognised (re-render trigger)
- Flash to new design

It's something the current implementation would omit.
I think @capJavert has an idea of how to improve the load of the feature, but it would still happen I think, unless we decide to block onboarding till we load all features.

idoshamun:
@rebelchris i think we can block rendering as we anyway wait for boot and boot has this information already. How does that sound? Long term, we will not need to wait once we conclude the experiment

packages/shared/src/components/MainLayout.tsx
```diff
@@ -152,7 +153,8 @@ export default function MainLayout({
   const shouldRedirectOnboarding =
     !user &&
     isPageReady &&
-    onboardingV2 !== OnboardingV2.Control &&
+    (onboardingV2 !== OnboardingV2.Control ||
+      onboardingV3 !== OnboardingV3.Control) &&
```
omBratteng, line 157: @idoshamun you're talking about here? So remove the V3 check? Because to have V3 you also have V2?

idoshamun: Yes, exactly

packages/shared/src/contexts/OnboardingContext.tsx
```diff
@@ -82,7 +87,10 @@ export const OnboardingContextProvider = ({
   };
   const { sidebarRendered } = useSidebarRendered();
   const showArticleOnboarding =
-    sidebarRendered && alerts?.filter && onboardingV2 === OnboardingV2.Control;
+    sidebarRendered &&
+    alerts?.filter &&
+    (onboardingV2 === OnboardingV2.Control ||
+      onboardingV3 === OnboardingV3.Control);
```
omBratteng, line 93: What about here? @idoshamun

idoshamun: I think it's not needed here as well. V3 should just indicate if we should show the new version (v3) or control (v2). Should not impact the rest

packages/shared/src/components/MainLayout.tsx
omBratteng: Removed.

packages/shared/src/contexts/OnboardingContext.tsx
omBratteng: Removed.

packages/shared/src/components/auth/OnboardingRegistrationForm.tsx
```diff
@@ -0,0 +1,144 @@
...
+                  Log in.
+                </button>
+              </AlertParagraph>
+            </Alert>
+          </>
+        )}
+
+        <Button className="w-full btn-primary" type="submit" loading={!isReady}>
+          Sign up - it&rsquo;s free ➔
+        </Button>
```
capJavert, line 122: I would add `isLoading` from `checkEmail` mutation also here since it can take few seconds and the page looks stuck for the user.

packages/shared/tailwind.config.js
```diff
@@ -183,6 +183,7 @@ module.exports = {
       desktopL: '2156px',
       mouse: { raw: '(pointer: fine)' },
       responsiveModalBreakpoint: '420px',
+      laptopH: { raw: '(min-height: 800px)' },
```
capJavert, line 186: Just wanna make sure this is part of our design system? cc: @rebelchris

packages/webapp/components/layouts/Onboarding/index.module.css
```diff
@@ -0,0 +1,14 @@
...
+  @media (min-height: 800px) and (min-width: 1020px) {
+    --video-scale: 1.5;
+  }
+  @media (min-height: 800px) and (min-width: 1240px) {
+    --video-scale: 1.8;
+  }
+  @media (min-height: 1000px) and (min-width: 1668px) {
+    --video-scale: 2;
+  }
+}
```
capJavert, line 14: Since we introduced `laptopH` can't we do this with tailwind?

capJavert (changes requested):
Final things that caught the eye from latest changes, CSS/tailwind stuff is not blocking just wanna make sure we don't introduce something out of the ordinary.

packages/webapp/components/layouts/Onboarding/index.module.css
omBratteng: I tried, but it would clash with widths.

packages/shared/tailwind.config.js
omBratteng: Since I moved to a css module instead, could probably remove this one.

capJavert: I think one or the other is better then doing it in two separate places. 👍

omBratteng: Think the other place is better. The video is kinda difficult to tweak with just tailwind

packages/webapp/components/layouts/Onboarding/index.module.css
rebelchris: I think for now use this approach, it's a bit messy because non-designers are requesting changes.
I would say maybe rather use `@apply` to import the correct scaling, this is very susceptible to tailwind changing the variants one day.

So keep the inline media queries, but use `@apply` I think it should work.

packages/shared/tailwind.config.js
rebelchris: Yeah this one we can remove, it was the better approach when all we needed was one media query
I think this whole video scaling got a bit out of scope, but it's done now.

packages/shared/src/components/auth/OnboardingRegistrationForm.tsx
omBratteng: Added

packages/shared/tailwind.config.js
omBratteng: Removed this one.

omBratteng: Re-added. Forgot I had already used it two other places to tweak the styles on short screens

capJavert (approved):
Good to go from my side! 🚀

packages/shared/tailwind.config.js
rebelchris: I would prefer if we can stick to not using custom measurements for elements that don't need it.
Is it really only on height this one needs to add the padding?

I hoped we could simply use the existing `tablet:` or something for this part.

omBratteng: I could also just throw it into the css file since they are both connected

capJavert (approved):

rebelchris (approved):
We made it ✨

PR #2187, opened 2023-09-18
fix: move to conditional wrapper

rebelchris:
- We used a dynamic render, but should have use our conditional render instead since it would simply render a empty fragment.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1837 #done


packages/shared/src/components/widgets/PostUsersHighlights.tsx
```diff
@@ -78,15 +79,22 @@ const Image = (props: SourceAuthorProps) => {
 const UserHighlight = (props: SourceAuthorProps) => {
   const { id, handle, name, username, permalink, userType = 'source' } = props;
   const Icon = getUserIcon(userType);
-  const LinkWrapper = userType === 'source' ? React.Fragment : ProfileTooltip;
+  const userTypeNotSource = userType === 'source';
```
sshanzel, line 82: I think there was a typo.
```suggestion
  const userTypeNotSource = userType !== 'source';
```

sshanzel (changes requested):
Looks great! Definitely one to use the ConditionalWrapper. Just needs to have the right condition.

packages/shared/src/components/widgets/PostUsersHighlights.tsx
rebelchris: Ah yes my bad was testing it and accidently pushed the wrong one 🙈

sshanzel (approved):
Let's go 🚀

capJavert (approved):

PR #2195, opened 2023-09-19
fix: use the correct query key to update reading history cache

DragosIorgulescu:
## Changes

_self explanatory_

## Events

_N / A_

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1841 #done

### DIFF
--- a/packages/shared/src/components/PostOptionsReadingHistoryMenu.tsx
+++ b/packages/shared/src/components/PostOptionsReadingHistoryMenu.tsx
@@ -16,7 +16,11 @@ import {
 import { MenuIcon } from './MenuIcon';
 import { QueryIndexes } from '../hooks/useReadingHistory';
 import { postAnalyticsEvent } from '../lib/feed';
-import { updateInfiniteCache } from '../lib/query';
+import {
+  generateQueryKey,
+  RequestKey,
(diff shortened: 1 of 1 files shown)


packages/shared/src/components/PostOptionsReadingHistoryMenu.tsx
```diff
@@ -45,8 +49,8 @@ const updateReadingHistoryPost =
   ): ((args: { id: string }) => Promise<() => void>) =>
   async () => {
     const history = client.getQueryData<ReadHistoryInfiniteData>(queryKey);
-    const { node } = history.pages[page].readHistory.edges[edge];
-    const updated = { ...node, post: { ...node.post, ...post } };
+    const { node } = history?.pages[page].readHistory.edges[edge] || {};
+    const updated = { ...node, post: { ...node?.post, ...post } };
```
DragosIorgulescu, line 53: The main fix is using the correct query key as seen below, but still making sure this does not break if for some reason history ends up `undefined` is a good idea IMO

rebelchris (approved):

PR #2203, opened 2023-09-19
fix: passing non dom props to elements

sshanzel:
## Changes
- As part of our guideline when passing down props from entirely different components/elements, we should introduce a single prop rather than extending the interface.
- This also fixes the same issue with passing down the property `insaneMode` which we receive console errors.

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1843 #done


packages/shared/src/components/cards/WelcomePostCard.tsx
```diff
@@ -23,16 +23,10 @@ export const WelcomePostCard = forwardRef(function SharePostCard(
...
-    className,
     children,
-    style,
     onReadArticleClick,
     enableSourceHeader = false,
-    enableMenu: _e,
-    showImage: _s,
-    insaneMode: _i,
-    menuOpened: _m,
-    ...props
```
sshanzel, line 35: Won't have to introduce this now, should be safe to pass down the props.

rebelchris (approved):
Nice find!

capJavert (approved):

PR #2217, opened 2023-09-21
fix: cleanup unused components

rebelchris:
- Ran a evaluation on which components and functions where no longer user.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done


packages/shared/src/graphql/squads.ts
```diff
@@ -287,18 +287,6 @@ export interface SquadEdgesData {
...
-  sourceId: string,
-  memberId: string,
-): Promise<SourceMember> => {
-  const res = await request(graphqlUrl, CHECK_USER_MEMBERSHIP, {
-    sourceId,
-    memberId,
-  });
-
-  return res.member;
-};
```
capJavert, line 300: Seen this one is still there on the API, can we remove it there too or leave it in webapp for future? It's just an export so it's not adding to the bundle size..

packages/shared/src/graphql/posts.ts
```diff
@@ -446,12 +417,6 @@ export const POST_BY_URL_QUERY = gql`
   ${SHARED_POST_INFO_FRAGMENT}
 `;
 
-export const getPostByUrl = async (url: string): Promise<Post> => {
```
capJavert, line 449: Also still available in API as my previous comment.

packages/shared/src/graphql/posts.ts
```diff
@@ -362,16 +343,6 @@ export const dismissPostFeedback = (id: string): Promise<EmptyResponse> => {
...
-export const votePost = (
-  id: string,
-  vote: UserPostVote,
-): Promise<EmptyResponse> => {
-  return request(graphqlUrl, VOTE_POST_MUTATION, {
-    id,
-    vote,
-  });
-};
-
```
capJavert, line 374: Hmm this is not used but should we use it in `useVotePost` since it's the same thing?

sshanzel (approved):
Always love to see code getting deleted.

This made me think, we can probably organize the files as well since we have now introduced proper folders to contain relevant files.

capJavert (approved):
Slim it!

packages/shared/src/graphql/posts.ts
rebelchris: Based on requestMethod :)

PR #2223, opened 2023-09-25
fix: reposition copy link

rebelchris:
- Reposition of the invite link and minor alignment issues

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1800 #done

### DIFF
--- a/packages/shared/src/components/squads/SquadHeaderMenu.tsx
+++ b/packages/shared/src/components/squads/SquadHeaderMenu.tsx
@@ -66,24 +66,32 @@ export default function SquadHeaderMenu({
     const canEditSquad = verifyPermission(squad, SourcePermissions.Edit);
     const canDeleteSquad = verifyPermission(squad, SourcePermissions.Delete);
 
-    const list: MenuItemProps[] = [
-      {
-        icon: <ContextMenuIcon Icon={TourIcon} />,
-        action: () =>
-          openModal({
-            type: LazyModal.SquadTour,
-          }),
-        label: 'Learn how Squads work',
-      },
-    ];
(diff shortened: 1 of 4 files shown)


packages/shared/src/components/squads/SquadHeaderMenu.tsx
```diff
@@ -70,6 +77,14 @@ export default function SquadHeaderMenu({
       },
     ];
 
+    if (!squad.currentMember && squad.public) {
+      list.unshift({
+        icon: <ContextMenuIcon Icon={LinkIcon} />,
+        action: () => trackAndCopyLink(),
+        label: 'Invitation link',
+      });
+    }
```
capJavert, line 86: I would rather have this be the first condition then using `unshift`, it will be vulnerable to order changes or accidental additions.

So I would propose:
- make a `list` empty array
- check the above condition for Invitation link and push if true
- push "Learn how squads work" after it without any condition.

capJavert (changes requested):
👍 I have one logic issue

packages/shared/src/components/squads/SquadHeaderMenu.tsx
capJavert: Like that the flow is always the same and new items added at the end.

rebelchris: Good point, will do that :)

capJavert (approved):
🚀 nice!

PR #2225, opened 2023-09-26
fix: add tracking error event

rebelchris:
- Add tracking error event for registration fails (CORS 99% of time)

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1863 #done

### DIFF
--- a/packages/shared/src/hooks/useRegistration.ts
+++ b/packages/shared/src/hooks/useRegistration.ts
@@ -58,22 +58,22 @@ const useRegistration = ({
   const {
     data: registration,
     isLoading: isQueryLoading,
-    error,
+    error: registrationError,
   } = useQuery(key, () => initializeKratosFlow(AuthFlow.Registration), {
     refetchOnWindowFocus: false,
   });
 
   useEffect(() => {
-    if (error) {
+    if (registrationError) {
       trackEvent({
         event_name: AuthEventNames.RegistrationInitialisationError,
         extra: JSON.stringify({
-          error,
+          error: registrationError,
           origin: Origin.InitializeRegistrationFlow,
         }),
       });
     }
-  }, [error, trackEvent]);
+  }, [registrationError, trackEvent]);
 
   const referralTraits = useMemo(() => {
     return {
--- a/packages/shared/src/hooks/useRegistration.ts
+++ b/packages/shared/src/hooks/useRegistration.ts
@@ -54,11 +54,23 @@ const useRegistration = ({
   const { displayToast } = useToastNotification();
   const { trackingId, referral, referralOrigin } = useContext(AuthContext);
   const timezone = getUserDefaultTimezone();
-  const { data: registration, isLoading: isQueryLoading } = useQuery(
-    key,
-    () => initializeKratosFlow(AuthFlow.Registration),
-    { refetchOnWindowFocus: false },
-  );
+  const {
+    data: registration,
+    isLoading: isQueryLoading,
+    error,
+  } = useQuery(key, () => initializeKratosFlow(AuthFlow.Registration), {
+    refetchOnWindowFocus: false,
+  });
+
+  if (error) {
+    trackEvent({
+      event_name: AuthEventNames.RegistrationInitialisationError,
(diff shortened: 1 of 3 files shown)


packages/shared/src/hooks/useRegistration.ts
```diff
@@ -54,11 +54,23 @@ const useRegistration = ({
...
+
+  if (error) {
+    trackEvent({
+      event_name: AuthEventNames.RegistrationInitialisationError,
+      extra: JSON.stringify({
+        error,
+        origin: 'initialize registration flow',
+      }),
+    });
+  }
```
capJavert, line 73: You need `useEffect` for this or try catch inside `queryFn` above right? Like this it will fire multiple times if parent hook/component/context re-renders.

packages/shared/src/hooks/useRegistration.ts
```diff
@@ -54,11 +54,23 @@ const useRegistration = ({
...
+  } = useQuery(key, () => initializeKratosFlow(AuthFlow.Registration), {
+    refetchOnWindowFocus: false,
+  });
+
+  if (error) {
+    trackEvent({
+      event_name: AuthEventNames.RegistrationInitialisationError,
+      extra: JSON.stringify({
+        error,
+        origin: 'initialize registration flow',
```
capJavert, line 70: Move to `Origin`?

capJavert (changes requested):
Just some sanity checks

packages/shared/src/hooks/useRegistration.ts
rebelchris: Oh 🤦 I even added it there 😂

packages/shared/src/hooks/useRegistration.ts
rebelchris: Yeah good point, might be an edge case where it re-renders after throwing the error.
Let me move it.

rebelchris: @capJavert Please check of modification suits the need : )

packages/shared/src/hooks/useRegistration.ts
```diff
@@ -54,11 +55,25 @@ const useRegistration = ({
...
+    if (error) {
+      trackEvent({
+        event_name: AuthEventNames.RegistrationInitialisationError,
+        extra: JSON.stringify({
+          error,
+          origin: Origin.InitializeRegistrationFlow,
+        }),
+      });
+    }
+  }, [error, trackEvent]);
```
capJavert: We should remove `trackEvent` here right, since it is not stable and as we do until we sort out tracking hook. Disable lint is ok here but if you wanna avoid even that we can move it to `queryFn` and catch there, something like this:

```jsx
const { data: registration, isLoading: isQueryLoading } = useQuery(
  key,
  () => {
    try {
      const result = await initializeKratosFlow(AuthFlow.Registration);

      return result;
    } catch (error) {
      trackEvent({
        event_name: AuthEventNames.RegistrationInitialisationError,
        extra: JSON.stringify({
          error,
          origin: Origin.InitializeRegistrationFlow,
        }),
      });

      throw error;
    }
  },
  { refetchOnWindowFocus: false },
);
```

rebelchris: Cool let me disable it for now until we sort the hook

capJavert (approved):
Let's get them errors!

packages/shared/src/lib/auth.ts
```diff
@@ -22,6 +22,7 @@ export enum AuthEventNames {
   ForgotPassword = 'forgot password',
   SubmitForgotPassword = 'submit forgot password',
   RegistrationError = 'registration error',
+  RegistrationInitialisationError = 'registration initialisation error',
```
idoshamun, line 25: Let's use American form to be consistent. `Initialization` in both the key and value

idoshamun (approved):
Just fix my comment before merging please

PR #2233, opened 2023-09-27
fix: prevent search bar hijacking on new tab

omBratteng:
- refactors onboarding title to a shared component
- creates onboarding component for extension
- prevents search bar hijacking on new tab

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1865 #done

### DIFF
--- a/packages/extension/src/newtab/App.tsx
+++ b/packages/extension/src/newtab/App.tsx
@@ -11,7 +11,9 @@ import Modal from 'react-modal';
 import 'focus-visible';
 import { ProgressiveEnhancementContextProvider } from '@dailydotdev/shared/src/contexts/ProgressiveEnhancementContext';
 import { OnboardingContextProvider } from '@dailydotdev/shared/src/contexts/OnboardingContext';
-import AuthContext from '@dailydotdev/shared/src/contexts/AuthContext';
+import AuthContext, {
+  useAuthContext,
+} from '@dailydotdev/shared/src/contexts/AuthContext';
 import { SubscriptionContextProvider } from '@dailydotdev/shared/src/contexts/SubscriptionContext';
 import { browser } from 'webextension-polyfill-ts';
 import { useInAppNotification } from '@dailydotdev/shared/src/hooks/useInAppNotification';
@@ -28,6 +30,14 @@ import { usePrompt } from '@dailydotdev/shared/src/hooks/usePrompt';
 import { defaultQueryClientConfig } from '@dailydotdev/shared/src/lib/query';
 import { ReactQueryDevtools } from 'react-query/devtools';
 import { useWebVitals } from '@dailydotdev/shared/src/hooks/useWebVitals';
+import {
+  useFeature,
+  useGrowthBookContext,
+} from '@dailydotdev/shared/src/components/GrowthBookProvider';
+import { feature } from '@dailydotdev/shared/src/lib/featureManagement';
+import { isTesting } from '@dailydotdev/shared/src/lib/constants';
+import { OnboardingV2 } from '@dailydotdev/shared/src/lib/featureValues';
+import ExtensionOnboarding from '@dailydotdev/shared/src/components/ExtensionOnboarding';
 import CustomRouter from '../lib/CustomRouter';
 import { version } from '../../package.json';
 import MainFeedPage from './MainFeedPage';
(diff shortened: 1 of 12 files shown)


packages/shared/src/components/ExtensionOnboarding.tsx
```diff
@@ -0,0 +1,43 @@
...
+      </OnboardingTitleGradient>
+
+      <p className="mb-9 tablet:mb-16 max-w-[25rem] typo-title3 tablet:typo-title2">
+        Please resume onboarding to unlock the entire feature suite of
+        daily.dev.
+        <br />
+        The magic awaits inside! ✨
+      </p>
+
+      <Link href={webappUrl}>
```
rebelchris, line 25: I think we can go directly to `/onboarding` here.

omBratteng: Forgot that one. Created a global constant and replaced the other instance of linking to onboarding page as well.

omBratteng:
Refactored it, now it only sends one impression
image

idoshamun:
As for the enrollment, there's nothing to worry about. Onboarding v2 is no longer in experimentation, so there's no danger of messing the metrics. It has only forced values now. The important enrollment is v3 which is not touched

packages/shared/src/components/ExtensionOnboarding.tsx
```diff
@@ -0,0 +1,43 @@
...
+import Link from 'next/link';
+import Logo, { LogoPosition } from './Logo';
+import { OnboardingTitleGradient } from './onboarding/common';
+import { onboardingUrl } from '../lib/constants';
+import { Button, ButtonSize } from './buttons/Button';
+import { cloudinary } from '../lib/image';
+
+const ExtensionOnboarding = (): ReactElement => {
+  return (
+    <div className="flex overflow-hidden flex-col justify-center items-center px-7 antialiased text-center min-h-[100vh] max-h-[100vh]">
```
rebelchris, line 11: Don't think you need the antialiased here right, it's set globally for the app.
Only exception before was modals as they are outside the render tree.

rebelchris (approved):
Nice nothing further from my side.
Let's just await @capJavert or @sshanzel review 🙏

packages/extension/src/newtab/App.tsx
```diff
@@ -71,17 +81,26 @@ function InternalApp({
...
     getContentScriptPermission,
   }));
 
   useEffect(() => {
-    if (routeChangedCallbackRef.current) {
+    if (routeChangedCallbackRef.current && isPageReady) {
       routeChangedCallbackRef.current();
     }
-  }, [routeChangedCallbackRef]);
+  }, [isPageReady, routeChangedCallbackRef]);
```
capJavert, line 103: This will slow down our new tab experience because growthbook does not initialize inline? Not sure what we can do with existing setup but might be critical since we want newtab to be fast?

packages/extension/src/newtab/App.tsx
```diff
@@ -104,6 +123,18 @@ function InternalApp({
...
 
+  useEffect(() => {
+    if (shouldRedirectOnboarding) {
+      routeChangedCallbackRef.current();
+    }
+  }, [routeChangedCallbackRef, shouldRedirectOnboarding]);
+
+  if (shouldRedirectOnboarding) {
+    // eslint-disable-next-line no-param-reassign
+    pageRef.current = '/hijacking';
```
capJavert, line 134: `/onboarding`? Or does `hijacking` has some other meaning for analytics?

capJavert (commented):
Don't see any issues apart from `newtab` potentially being slower to allow growthbook to load.

packages/extension/src/newtab/App.tsx
capJavert: Ah actually, it does not really delay, seen it wrong where `isPageReady` is used, it's just for tracking.

omBratteng: Yeah, this is just for tracking. And if it turns out it slows it down, I think I can remove it and then just accept that we get sent an extra analytics event for `/` page.

packages/extension/src/newtab/App.tsx
rebelchris: Its a analytic split between onboarding page and this extension only one.
Didnt have a better name 😂

packages/extension/src/newtab/App.tsx
capJavert: Yeah, I was more worried for actual display/rendering. Should be ok.

capJavert (approved):
🚀

packages/shared/src/components/ExtensionOnboarding.tsx
omBratteng: Should I remove it? Don't think it matters that much?

omBratteng: Because it is rendered before the already antialiased `MainLayout`, component, I am leaving it as is.
https://github.com/dailydotdev/apps/blob/8ba80c467719c23e4be9b881286ded8821eac161/packages/shared/src/components/MainLayout.tsx#L188

PR #2240, opened 2023-09-29
refactor: share modal ui to open separate form for squad

sshanzel:
## Changes
- Remove the Markdown input from the ShareModal.
- When clicking a Squad icon, we will then open the modal to share a post to Squad.
- Since the ShareModal was not used in the LazyModal enum, we won't have to make any changes with the LazyModal itself since only one instance is opened.

Preview:

https://github.com/dailydotdev/apps/assets/13744167/0b2bc640-3775-489b-8203-01e12181df82

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1794 #done

### DIFF
--- a/packages/shared/src/components/modals/ShareModal.spec.tsx
+++ b/packages/shared/src/components/modals/ShareModal.spec.tsx
@@ -1,5 +1,5 @@
 import { render, RenderResult, screen, waitFor } from '@testing-library/react';
-import { QueryClient, QueryClientProvider } from 'react-query';
+import { QueryClient } from 'react-query';
 import React from 'react';
 import nock from 'nock';
 import { useRouter } from 'next/router';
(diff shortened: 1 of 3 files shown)


rebelchris (commented):
The only thing I might be overlooking how does the first modal get closed when the second opens?

sshanzel:
> The only thing I might be overlooking how does the first modal get closed when the second opens?

Do we have to close the first one when we just opened the second one? My concern is if the user cancels sharing with the Squad.

rebelchris:
There is no way to go back anyway right?
I think the cleanest is to simply close it while opening the new one.

What would be a flow of having it still? I can try it locally to see what happens.

Error will look into on Monday

sshanzel:
> There is no way to go back anyway right?

Closing the modal for input, then you should be back to the original modal.

> What would be a flow of having it still? I can try it locally to see what happens.

Like you wanted it to be shared on some Squad but doubted and didn't proceed. Instead, I want to copy the link to be shared through some other platforms.

rebelchris:
> > There is no way to go back anyway right?
> 
> Closing the modal for input, then you should be back to the original modal.
> 
> > What would be a flow of having it still? I can try it locally to see what happens.
> 
> Like you wanted it to be shared on some Squad but doubted and didn't proceed. Instead, I want to copy the link to be shared through some other platforms.

Alright, checked and let's leave like this it works

sshanzel:
The tests have passed now 🚀

packages/shared/src/components/modals/ShareModal.spec.tsx
```diff
@@ -40,43 +41,33 @@ const renderComponent = (
...
+        isOpen
+        onRequestClose={onRequestClose}
+        ariaHideApp={false}
+      />
+    </TestBootProvider>,
   );
 };
 
 describe('ShareModal Test Suite:', () => {
-  it('should render the article preview', async () => {
```
sshanzel, line 71: Not used here anymore. Different modal, different tests.

rebelchris (approved):
Let's go

PR #2241, opened 2023-09-29
fix: wait for boot to be completed before we set features

omBratteng:
- wait for boot to be completed before we set features

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done

### DIFF
--- a/packages/shared/src/contexts/BootProvider.tsx
+++ b/packages/shared/src/contexts/BootProvider.tsx
@@ -181,6 +181,7 @@ export const BootDataProvider = ({
       deviceId={deviceId}
       experimentation={cachedBootData?.exp}
       updateExperimentation={updateExperimentation}
+      isFetched={isFetched}
     >
       <AuthContextProvider
         user={user}
--- a/packages/shared/src/contexts/BootProvider.tsx
+++ b/packages/shared/src/contexts/BootProvider.tsx
@@ -181,7 +181,7 @@ export const BootDataProvider = ({
       deviceId={deviceId}
       experimentation={cachedBootData?.exp}
       updateExperimentation={updateExperimentation}
-      isFetched={isFetched}
+      firstLoad={initialLoad}
     >
       <AuthContextProvider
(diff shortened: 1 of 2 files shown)


packages/shared/src/contexts/BootProvider.tsx
```diff
@@ -181,6 +181,7 @@ export const BootDataProvider = ({
       deviceId={deviceId}
       experimentation={cachedBootData?.exp}
       updateExperimentation={updateExperimentation}
+      isFetched={isFetched}
```
sshanzel, line 184: We also have another state in this component named `initialLoad`, it might be better to combine checking if it is not null or undefined with `isFetched` since we watch for the changes of boot remote data and then set the result to state (hopefully we can do the boot refactoring soon) to understand we have now the value from the server. Just worried about a race condition.

rebelchris: You might indeed be able to use this initialLoad for it, should result in same effect from what I can see.

sshanzel: There could be a difference of 1 render. Once the query is done fetching, the `isFetched` is true now, but the `useEffect` to set the cachedBootData (since we set it to a new value in the `useEffect` has not been run yet and since the property value is now true, it would proceed with the previous cache data.

omBratteng: I'll take a look at `initialLoad`.

It already runs with the cached data on the experiment, hence why you see the hijack page on the extension after logging in. You first see the feed, then get thrown back to the hijack page once `updateExperimentation` has been run and set it to the old cache. After I page reload it is all good and dandy.

sshanzel (approved):
Code-wise, looks good to me 🚀

rebelchris (approved):

PR #2242, opened 2023-10-02
fix: remove background color on `post-content` class

omBratteng:
## Changes

<table>
<tr>
  <td>Before
  <td>After
<tr>
 <td>
 <td>
</table>

- removes the background colour from the `post-content` class
 
## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1867 #done

### DIFF
--- a/packages/shared/src/components/modals/BasePostModal.module.css
+++ b/packages/shared/src/components/modals/BasePostModal.module.css
@@ -10,21 +10,4 @@
     overflow-y: scroll;
     z-index: 10;
   }
-
-  & :global(.post-content) {
-    display: flex;
-    border: none;
-    border-radius: 0;
-    max-width: 100%;
-    padding-bottom: 4rem;
-
-    @screen tablet {
-      padding-bottom: 0;
-    }
-
-    @screen laptop {
-      height: unset;
-      border-radius: 1rem;
-    }
-  }
 }
--- a/packages/shared/src/components/modals/BasePostModal.module.css
+++ b/packages/shared/src/components/modals/BasePostModal.module.css
@@ -12,8 +12,6 @@
   }
 
   & :global(.post-content) {
-    @apply bg-theme-bg-primary;
-
     display: flex;
     border: none;
     border-radius: 0;
--- a/packages/shared/src/components/modals/ArticlePostModal.tsx
+++ b/packages/shared/src/components/modals/ArticlePostModal.tsx
@@ -40,7 +40,7 @@ export default function ArticlePostModal({
       onRequestClose={onRequestClose}
       postType={PostType.Article}
       source={post.source}
-      loadingClassName="post-content"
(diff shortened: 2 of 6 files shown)


omBratteng:
Could probably remove the entire `post-content` class, it's only the `SharePostModal` that uses it. And most of the properties doesn't have any effect on the content, only the padding-bottom is different.

sshanzel:
Aren't they supposed to be on primary like all of our modals? What did the design team say to this concern?

sshanzel:
> Could probably remove the entire post-content class, it's only the SharePostModal that uses it. And most of the properties doesn't have any effect on the content, only the padding-bottom is different.

This is a legacy styling so if we move forward, we should probably remove it if not used anymore and translate the other parts.

omBratteng:
Right now the modal has tertiary background, but also primary is set...
image

omBratteng:
I've sent a message to design to ask about the correct colour. I think it should be primary, and that tertiary gets inherited from somewhere that it shouldn't 🤔

omBratteng:
Replacing `bg-theme-bg-primary` with `!bg-theme-bg-primary` will probably fix the background color. Think this was error was introduced in the last tailwind update
https://github.com/dailydotdev/apps/blob/eb791b63c112de987c40c5b813c90d75cc70b5fa/packages/shared/src/components/modals/BasePostModal.tsx#L35

packages/shared/src/components/modals/BasePostModal.module.css
```diff
@@ -12,8 +12,6 @@
   }
 
   & :global(.post-content) {
-    @apply bg-theme-bg-primary;
```
rebelchris, line 15: This is still not needed?
Not sure what it was for tbh

rebelchris (approved):

packages/shared/src/components/modals/BasePostModal.module.css
omBratteng: The class `post-content` is used on `SharePostModal` and `SquadPostContent`.
But from what I could see, the only property that had any effect, outside of background colour) is the padding-bottom. I could transfer those to tailwind where applicable instead, and then remove this old class.

https://github.com/dailydotdev/apps/blob/eb791b63c112de987c40c5b813c90d75cc70b5fa/packages/shared/src/components/modals/SharePostModal.tsx#L36

https://github.com/dailydotdev/apps/blob/eb791b63c112de987c40c5b813c90d75cc70b5fa/packages/shared/src/components/post/SquadPostContent.tsx#L87

omBratteng: Also removed the background from a flew more places I found that it would crash (modal post skeleton loader).

rebelchris (approved):

PR #2253, opened 2023-10-04
feat: existing anonymous user onboarding footer

sshanzel:
## Changes
- Modified the login button component to allow the display of just one button.
- Added props to. the login button to allow different labels on the button itself.
- Introduce a banner in the footer for existing anonymous users only that is still under OnboardingV2.
- Rather than include changes in the API and increase the footprint of the changes, we have decided to put the config on whether to redirect to onboarding existing anonymous users, and be handled in the FE.
- Removed the condition to redirect users in the onboarding page to the main page if they are in control value for OnboardingV2.

Previews:
Screenshot 2023-10-04 at 8 42 37 PM
Screenshot 2023-10-04 at 8 42 29 PM
Screenshot 2023-10-04 at 8 24 46 PM

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done

### DIFF
--- a/packages/shared/src/components/banners/FixedBottomBanner.tsx
+++ b/packages/shared/src/components/banners/FixedBottomBanner.tsx
@@ -1,5 +1,6 @@
-import React, { ReactElement } from 'react';
+import React, { ReactElement, useEffect } from 'react';
 import { useRouter } from 'next/router';
+import { differenceInDays } from 'date-fns';
 import CloseButton from '../CloseButton';
 import LoginButton from '../LoginButton';
 import { Button } from '../buttons/Button';
@@ -11,6 +12,10 @@ import { useAuthContext } from '../../contexts/AuthContext';
 import usePersistentContext from '../../hooks/usePersistentContext';
 import { generateStorageKey, StorageTopic } from '../../lib/storage';
 
+const anonymousMigrationDate = new Date(2023, 10, 4);
+const now = new Date();
+const daysLeft = differenceInDays(anonymousMigrationDate, now);
+
 export function FixedBottomBanner(): ReactElement {
   const { user } = useAuthContext();
   const router = useRouter();
@@ -19,6 +24,15 @@ export function FixedBottomBanner(): ReactElement {
     generateStorageKey(StorageTopic.Onboarding, 'wall_dismissed'),
     false,
   );
+
+  useEffect(() => {
+    if (
+      router.pathname !== '/onboarding' &&
+      anonymousMigrationDate < new Date()
+    ) {
+      router.push('/onboarding');
+    }
+  }, [router]);
 
   if (
     user ||
@@ -40,7 +54,8 @@ export function FixedBottomBanner(): ReactElement {
       <div className="flex flex-col laptop:flex-row gap-4 justify-center laptop:items-center w-full laptop:max-w-[52.25rem]">
         <div className="flex flex-col flex-1 gap-2 w-full">
           <h1 className="font-bold typo-large-title">
-            Registration is going to be required in 30 days
+            Registration is going to be required in {daysLeft} day
+            {daysLeft > 1 ? 's' : ''}
           </h1>
           <p className="flex flex-1 w-full typo-title3">
             We’d love to see you join our community ❤︎. Signing up gets you free
--- a/packages/shared/src/components/banners/FixedBottomBanner.tsx
+++ b/packages/shared/src/components/banners/FixedBottomBanner.tsx
@@ -1,0 +1,59 @@
+import React, { ReactElement } from 'react';
+import { useRouter } from 'next/router';
+import CloseButton from '../CloseButton';
+import LoginButton from '../LoginButton';
+import { Button } from '../buttons/Button';
+import { tellMeWhy } from '../../lib/constants';
+import { useActions } from '../../hooks/useActions';
+import { ActionType } from '../../graphql/actions';
+import { useFeature } from '../GrowthBookProvider';
+import { feature } from '../../lib/featureManagement';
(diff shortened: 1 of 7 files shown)


packages/shared/src/components/banners/FixedBottomBanner.tsx
```diff
@@ -0,0 +1,52 @@
...
+  onDismiss(): void;
+  daysLeft: number;
+}
+
+export function FixedBottomBanner({
+  onDismiss,
+  daysLeft,
+}: FixedBottomBannerProps): ReactElement {
+  return (
+    <div className="flex fixed inset-0 z-max justify-center items-center p-6 w-full bg-gradient-to-l top-[unset] from-[#EF43FD] to-[#6451F3]">
```
rebelchris, line 17: Ah no not again custom gradients 😅
This is really getting out of control

packages/shared/src/components/banners/FixedBottomBanner.tsx
```diff
@@ -0,0 +1,52 @@
...
+            We’d love to see you join our community ❤︎. Signing up gets you free
+            access to the personalized feed, discussions, squads, AI search, and
+            more!
+          </p>
+        </div>
+        <div className="flex flex-row laptop:flex-col gap-2">
+          <LoginButton
+            showLoginButton={false}
+            copy={{ signup: 'Sign up now' }}
+          />
```
rebelchris, line 39: I thought the idea was it simply linked to `/onboarding` no?

packages/shared/src/components/banners/FixedBottomBanner.tsx
```diff
@@ -0,0 +1,52 @@
...
+          <Button
+            tag="a"
```
rebelchris, line 41: Missing rel here :)

rebelchris (commented):
I thought we could simplify the cta on the button to be honest.

packages/shared/src/components/banners/FixedBottomBanner.tsx
rebelchris: Yeah described in specs as well:


“Sign up” | If the user clicks on sign up, we open the onboarding v3 or v2 depending on the experimentation bucket
-- | --


So you can just use a normal button and link it hardcoded to webapp onboarding.
Then maybe revert changes on loginButton.
https://dailydotdev.atlassian.net/wiki/spaces/HAN/pages/614432779/Anon+users+sign+up+changes+Login+page+for+extension

Also good side point would be:
This makes a perfect example of verify before building as we discussed about.

packages/shared/src/components/banners/FixedBottomBanner.tsx
sshanzel: Just gonna be a one-off thing, I promise 😂

packages/shared/src/components/banners/FixedBottomBanner.tsx
sshanzel: Oh, there was no link to the spec in the ticket. I only based on the JIRA ticket which did not mention the behavior of the button itself.

sshanzel: Let me read the spec now and see if there are any differences with the current implementation.

packages/shared/src/components/banners/FixedBottomBanner.tsx
sshanzel: What should we put value to the rel attribute?

omBratteng: Should we create a global const so we can easily refer to what value to use (in most places)?

packages/shared/src/lib/featureValues.ts
```diff
@@ -29,3 +31,7 @@ export enum SearchExperiment {
   Control = 'control',
   V1 = 'v1',
 }
+
+const anonymousMigrationDate = new Date(2023, 10, 4);
```
rebelchris, line 35: I'm just asking product for the exact date, unless you got it already from them?

rebelchris (changes requested):
Just the rel, and correct date and then we can merge.
I'll request changes just to ensure we don't push it live without the right date 🙏

packages/shared/src/lib/featureValues.ts
rebelchris: Asked Nimrod and last day available should be:
So October 31st, 23:59 would be a good deadline I guess

So you would set 1st of November basically :) 
Let's do that and the rel and ship it 🚀

packages/shared/src/components/banners/FixedBottomBanner.tsx
sshanzel: Have added a const in a common file now to make it easier in the future 🚀

packages/shared/src/lib/featureValues.ts
sshanzel: Nice. Thanks for checking with them. Should be updated now 🚢

rebelchris (changes requested):
One small remark 👀

packages/shared/src/lib/featureValues.ts
```diff
@@ -29,3 +31,7 @@ export enum SearchExperiment {
   Control = 'control',
   V1 = 'v1',
 }
+
+const anonymousMigrationDate = new Date(2023, 10, 1);
```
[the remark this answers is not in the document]
sshanzel: In JS, the month starts at 0.

rebelchris (approved):

PR #2254, opened 2023-10-05
fix: convert webapp requests to api directly

rebelchris:
- We had some legacy code that for the webapp routed traffic internally first
- We now remove this to directly call API and avoid re-routing latency (potentially, probably low impact)

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1886 #done

### DIFF
--- a/packages/shared/src/lib/config.ts
+++ b/packages/shared/src/lib/config.ts
@@ -1,9 +1,4 @@
-export const apiUrl =
-  typeof window === 'undefined' ||
-  process.env.NODE_ENV === 'test' ||
-  process.env.TARGET_BROWSER
-    ? process.env.NEXT_PUBLIC_API_URL
-    : '/api';
+export const apiUrl = process.env.NEXT_PUBLIC_API_URL;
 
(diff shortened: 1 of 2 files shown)


idoshamun (approved):

idoshamun:
@rebelchris I actually see that it breaks live previews. Maybe we should enable it only for previews?

rebelchris:
@idoshamun We actually where wondering if we could simple allow these generic vercel sub domains as a origin?
Maybe the easier solution in the long term.

capJavert (commented):
Having API on separate domain will add CORS requests to each request. Most of the time that is worst then just reverse proxy routing through webapp `/api` as we had now.

idoshamun:
@capJavert are you sure it'll add cors even on `app.daily.dev` and `api.daily.dev`. I don't remember the exact rules of cors but they share the same primary domain

capJavert:
@idoshamun yes, its a matter of `origin` and subdomain is different `origin`.

PR #2257, opened 2023-10-05
fix: fallback value of bottom banner

sshanzel:
## Changes
- The fallback value is false which means it would be displayed.
- Deployed at preview2 to see the difference between prod. Verified to be working.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done

### DIFF
--- a/packages/shared/src/components/MainLayout.tsx
+++ b/packages/shared/src/components/MainLayout.tsx
@@ -193,13 +193,6 @@ export default function MainLayout({
   const shouldShowBanner =
     !user && onboardingV2 === OnboardingV2.Control && isFetched && !isDismissed;
 
-  console.log(
-    !user,
-    onboardingV2 === OnboardingV2.Control,
-    isFetched,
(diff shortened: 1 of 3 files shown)


packages/shared/src/components/MainLayout.tsx
```diff
@@ -99,7 +99,7 @@ export default function MainLayout({
   });
   const [isDismissed, setIsDismissed, isFetched] = usePersistentContext(
     generateStorageKey(StorageTopic.Onboarding, 'wall_dismissed'),
-    false,
+    true,
```
rebelchris, line 102: This sets the default to be dismissed?
How does it ever show to people then?

sshanzel: Now I understand why it doesn't show anymore. Brain fart moments 😂

PR #2259, opened 2023-10-05
fix: allow editing share type posts (squad only)

DragosIorgulescu:
## Dependencies

- https://github.com/dailydotdev/daily-api/pull/1461

## Changes

- updates the `sharePost` mutation usage to allow passing a `postID` and behave like an upsert operation
- use the existing `ShareLink` component to render the correct form when editing a `share` post type

## Preview

edit-share-post

## Events

_N / A_

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1801 #done


packages/shared/src/components/post/write/WriteLinkPreview.tsx
```diff
@@ -48,7 +48,7 @@ export function WriteLinkPreview({
         <WritePreviewContent className={isMinimized && '!py-2 !px-3'}>
           <div className="flex flex-col flex-1 typo-footnote">
             <span className="font-bold line-clamp-2">{preview.title}</span>
-            {preview.source &&
+            {preview.source?.id !== 'unknown' &&
```
DragosIorgulescu, line 51: This does not feel right, but based on my understanding of how these are stored in the db, it's the only way to determine if this is a URL preview of an external page, or a preview of an internal post

DragosIorgulescu:
I still have some UI cleanup to do, but overall I tried to be as uninstrusive as possible with implementing this.

packages/shared/src/components/post/write/WriteLinkPreview.tsx
rebelchris: Yep it's a set constraint at the moment :)

packages/webapp/pages/posts/[id]/edit.tsx
```diff
@@ -104,7 +105,19 @@ function EditPost(): ReactElement {
         isForbidden={!isVerified || !squad || !canEdit}
       >
         <WritePostHeader isEdit />
-        <WriteFreeformContent className="py-6 px-4" />
+
+        {post?.type === PostType.Share ? (
```
Comment (rebelchris) [inline packages/webapp/pages/posts/[id]/edit.tsx:109]: I think ideally we use a completely different mutation for editing actually.
Like we do for EditPost it's a better failsafe way.

Not 100% sure how it's tackled in the freeform one now, but maybe we can look at that logic and re-apply?

rebelchris (commented):
Think we can make it safer by applying seperate calls.

packages/shared/src/hooks/squads/usePostToSquad.ts
```diff
@@ -149,10 +171,33 @@ export const usePostToSquad = ({
     [preview, displayToast, onSubmitLink, onPost, isPosting],
   );
 
+  const onUpdatePost = useCallback(
+    (
+      e: BaseSyntheticEvent,
+      postId: Post['id'],
+      sourceId: Squad['id'],
+      commentary: string,
```
rebelchris, line 179: This is basically double-defined in the interface and here, can't we make it 1 type to share across both instances?

rebelchris (commented):
One small suggestion.
After API is resolved, I'd like to test it out locally before we merge this, so holding my approval for that 🙏

packages/shared/src/hooks/squads/usePostToSquad.ts
DragosIorgulescu: good point, cleaned it up 🙏

packages/shared/src/hooks/squads/usePostToSquad.ts
```diff
@@ -149,10 +171,28 @@ export const usePostToSquad = ({
...
 
+  const onUpdatePost = useCallback<UsePostToSquad['onUpdatePost']>(
+    (e, postId, sourceId, commentary) => {
+      e.preventDefault();
+
+      if (isUpdating) {
+        return null;
+      }
+
+      return updatePost({
```
rebelchris, line 182: For some reason I get the generic "your post has been shared to squad" maybe we need a different callback here?
(also not sure it fires the same tracking perhaps?)

rebelchris (changes requested):
When placing a new shared post (no edit) it fails now:
Screenshot 2023-10-13 at 16 57 35

packages/shared/src/hooks/squads/usePostToSquad.ts
```diff
@@ -77,8 +84,12 @@ export const usePostToSquad = ({
       },
     });
 
-  const onSharedPostSuccessfully = async () => {
-    displayToast('This post has been shared to your Squad');
+  const onSharedPostSuccessfully = async (update = false) => {
```
rebelchris, line 87: For now no action needed, but think moving forward let's prefer object ways?

rebelchris (approved):
Nice tested and validated 🤘

packages/shared/src/graphql/squads.ts
```diff
@@ -187,6 +187,14 @@ export const ADD_POST_TO_SQUAD_MUTATION = gql`
   }
 `;
 
+export const UPDATE_SQUAD_POST_MUTATION = gql`
+  mutation UpdateSquadPost($id: ID!, $sourceId: ID!, $commentary: String) {
```
sshanzel, line 191: Why do we need the `sourceId` as part of the parameters? Should the ID alone make it unique enough?

sshanzel (approved):
Looks great! Just had some non blocking comments 🚀

packages/shared/src/graphql/squads.ts
DragosIorgulescu: not really needed, but does not hurt to have it either, my reasoning is to query based on both source & id to ensure we are not accidentally accessing the wrong record.

rebelchris: I would also just stick with ID it's uniform with what we already use for other pieces and reduces overhead I think.

PR #2261, opened 2023-10-06
fix: set title length and validation

rebelchris:
- Set the max length frontend wise for freeform title
- I noticed we didn't have it working when accessing the edit page, so decided to set validity once it's known.

Related API:
https://github.com/dailydotdev/daily-api/pull/1465

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done

### DIFF
--- a/packages/shared/src/hooks/useInputFieldFunctions.ts
+++ b/packages/shared/src/hooks/useInputFieldFunctions.ts
@@ -43,6 +43,16 @@ function useInputFieldFunctions<
     setValidInput(inputRef.current.checkValidity());
   }, 1500);
 
+  useEffect(() => {
+    if (inputRef.current?.value) {
+      setInputLength(inputRef.current.value.length);
+      const inputValidity = inputRef.current.checkValidity();
+      if (inputValidity) {
+        setValidInput(true);
+      }
+    }
+  }, [inputRef]);
+
   const onInput = (event: SyntheticEvent<T, InputEvent>): void => {
     clearIdleTimeout();
     baseOnInput(event);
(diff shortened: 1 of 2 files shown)


omBratteng (approved):

packages/shared/src/hooks/useInputFieldFunctions.ts
```diff
@@ -43,6 +43,16 @@ function useInputFieldFunctions<
...
 
+  useEffect(() => {
+    if (inputRef.current?.value) {
+      setInputLength(inputRef.current.value.length);
+      const inputValidity = inputRef.current.checkValidity();
+      if (inputValidity) {
+        setValidInput(true);
+      }
+    }
+  }, [inputRef]);
```
capJavert, line 54: Hmm how does this work, because `inputRef` is not reactive or state so reference is always the same? We should instead check this in `setInput` I think.

rebelchris: I want it to be stale, don't want it to trigger all the time.
It basically waits till it exists and then sets the starting point once, otherwise we already handle on input change.

capJavert: So this should trigger once when input/hook renders? Wonder why it was not needed before?

rebelchris: Screenshot 2023-10-06 at 13 38 11
 
To give more context, the above (right 0) would show 250 before this fix.
I think we just never noticed it.
But basically it would never set anything if the value would be set already on load.

> Note: it already worked if you removed a char and re-typed though

capJavert: Makes sense, it just looked weird and harder to understand. Hoping down the line we can move to some input library like `react-hook-form`. 🤞

capJavert (approved):

packages/shared/src/hooks/useInputFieldFunctions.ts
rebelchris: Yeah, more and more edge-cases and duct-tape.
Lol really be nice to see a uniform library

PR #2262, opened 2023-10-06
feat: boot and features loading optimizations

capJavert:
## Changes

We want to make sure features data is available as soon as boot. To do this couple of changes were made:
- `decrypt` call was moved to `getBootData` (see analysis [here](https://dailydotdev.slack.com/archives/C02E2C3C13R/p1696510214606839?thread_ts=1696499665.469449&cid=C02E2C3C13R))
  - this allowed us to have features available sync during boot state setter in `BootProvider`
- `GrowtbookProvider` now initializes `gb` object and sets features earlier and has `ready` state (see HOC below)
  - useEffect is still here if we want to dynamically update features on boot refetch
- added early return in `BootProvider` (this revision can be seen here 82be0775bfacd4c812e9920c9e2591e4f8cda1b7)
  - `initialLoad` is `null` as long as boot data is not available, different providers use it differently, for example `MainLayout` always waits for `isAuthReady` which is just check if `initialLoad` of boot is not null
  - because we can not render our app before boot is fetched early return in this context prevents growthbook of loading and also rendering any `useFeature` hooks until boot is fetched and `features` are available
  - which then makes it that from first render `useFeature` hooks would have the value for current user
  - from my tests this does not hurt performance because either way we did not render pages before boot is fetched, boot is kinda like our init boundary
 - that said we might want to render some pages or parts of our app without waiting for boot/features (`/join` is one current example), that is why I also added `withFeaturesBoundary` HOC (higher order component) with which we can wrap any component to prevent rendering until boot is fetched
  - this change again removed early return from Boot (from previous point) in favor of selectively preventing rendering
  - rule of the thumb for the future pages/changes would be that the lower in component tree this HOC is used the better since more parts of the page will render sooner

With above HOC applied to `MainLayout` (available on [preview3](https://preview3.app.daily.dev/))

Some debug logging for My feed page and `MainLayout`:

Before:
image

After (features available on first render):
image

Before merging this we also need:
- [ ] check current experiments and adjust them if needed
- [x] test adjustments

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1881 #done


rebelchris (approved):
Quite a lean approach actually.
It's neat we can use it as is, but decide to add the boundary to something.

Don't see any blocking comments from my side.
(besides testing / test cases)

omBratteng (approved):
Code wise it looks great. Just have to test with the extension that we don't get the hijack flash after login.

packages/shared/src/lib/boot.ts
```diff
@@ -78,5 +85,20 @@ export async function getBootData(app: string, url?: string): Promise<Boot> {
     credentials: 'include',
     headers: { app, 'Content-Type': 'application/json' },
   });
-  return res.json();
+  const result = await res.json();
+
+  try {
+    const features = await decrypt(
```
idoshamun, line 91: This mean we will decrypt it every time we fetch. Before we decrypted and initialized GB only once, what behavior do we prefer?

packages/shared/src/components/withFeaturesBoundary.tsx
```diff
@@ -0,0 +1,22 @@
+import React, { ComponentType, ReactElement, useContext } from 'react';
+import { FeaturesReadyContext } from './GrowthBookProvider';
+
+const withFeaturesBoundary = <Props,>(
```
idoshamun, line 4: Love this approach!

packages/shared/src/components/MainLayout.tsx
```diff
@@ -238,3 +239,5 @@ export default function MainLayout({
     </div>
   );
 }
+
+export default withFeaturesBoundary(MainLayout);
```
idoshamun, line 243: This is pretty strict, means every MainLayout element will wait for features. Is this what we want?

packages/shared/src/lib/boot.ts
capJavert: Yes we do decrypt each time, thats why we did performance tests to see the impact.

GB is still initalized only once.

packages/shared/src/components/MainLayout.tsx
capJavert: Yeah it is, that is why I added a note in description to we should strive using this as "lowest" in component tree as possible. 

That said, `MainLayout` waits for `isPageReady` which has `isAuthReady` which waits for first boot load, so this does not add any additional time for the user in our current setup but it does make sure features are loaded from the start.

DragosIorgulescu (approved):
I like this, although having it at the top of the tree has both advantages and disadvantages,  the main one being, as you mentioned, regarding pages where we don't want to wait for features to load before rendering.

packages/shared/src/components/withFeaturesBoundary.tsx
```diff
@@ -0,0 +1,22 @@
...
+  WrappedComponent: ComponentType<Props>,
+): ComponentType<Props> => {
+  const WithFeaturesBoundary = (props: Props): ReactElement => {
+    const { ready: areFeaturesReady } = useContext(FeaturesReadyContext);
+
+    if (!areFeaturesReady) {
+      return null;
```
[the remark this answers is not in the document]
capJavert: Good idea, yes, I'll add.

packages/shared/src/components/MainLayout.tsx
capJavert: That would complicate it I think, while HOC can be used dynamically it would introduce again cases where based on props changes you could render first without HOC and then with HOC which would again possibly let app render without features -> which would beat the purpose of using HOC in the first place.

rebelchris (approved):
Alright no blockers from my side then.
I suggest we can give it a go on webapp and try impact for 1/2 days, test on prod with enrolments of experiments.
IF nothing pops up we can launch extension.

packages/shared/src/lib/boot.ts
```diff
@@ -78,5 +85,20 @@ export async function getBootData(app: string, url?: string): Promise<Boot> {
...
+      result.exp.f,
+      process.env.NEXT_PUBLIC_EXPERIMENTATION_KEY,
+      'AES-CBC',
+      128,
+    );
+
+    result.exp.features = JSON.parse(features);
+  } catch (error) {
+    // TODO features-loading-optimization log error
```
capJavert, line 100: @rebelchris can I maybe track decrypt error event here? We didn't track it in previous implementation? You mentioned there is a track error event somewhere already?

capJavert:
We validated that current onboarding works with feature loading, plan is to merge the webapp and then later next week do an extension.

packages/shared/src/lib/boot.ts
rebelchris: Yes we can add it :)
You can look at `useError` for a global one.
Or something like `RegistrationInitializationError` 

Probably the later is better so we can have a unique key basically.

capJavert: Since we didn't catch previously we will leave as is, we can add catch in WT-1882

PR #2266, opened 2023-10-10
fix: add some constraints to our debounce and refresh token

rebelchris:
- Trying to solve the infinite loop issue
- Main constraints:
- Add debounce to minimum of 200ms to avoid race conditions
- Add minimum call time to refresh call to avoid recalling within a minute (should be fine as our token is valid 15 minutes 😅)

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done

### DIFF
--- a/packages/shared/src/hooks/useRefreshToken.ts
+++ b/packages/shared/src/hooks/useRefreshToken.ts
@@ -15,13 +15,11 @@ export function useRefreshToken(
   const saveDelay =
     differencePlusTwoMinutes <= 200 ? 200 : differencePlusTwoMinutes;
 
-  const [useRefresh] = useDebounce(refresh, saveDelay, 1000 * 60);
+  const [callRefresh] = useDebounce(refresh, saveDelay, 1000 * 60);
 
   useEffect(() => {
     if (accessToken?.token) {
-      // @NOTE see https://dailydotdev.atlassian.net/l/cp/dK9h1zoM
-      // eslint-disable-next-line react-hooks/rules-of-hooks
-      useRefresh();
+      callRefresh();
     }
-  }, [accessToken?.token, useRefresh]);
+  }, [accessToken?.token, callRefresh]);
 }


rebelchris:
Seems to be validated and working:
https://dailydotdev.slack.com/archives/C03C3JZPRU4/p1696938514735759

packages/shared/src/hooks/useRefreshToken.ts
```diff
@@ -11,15 +11,17 @@ export function useRefreshToken(
...
+  const saveDelay =
+    differencePlusTwoMinutes <= 200 ? 200 : differencePlusTwoMinutes;
+
+  const [useRefresh] = useDebounce(refresh, saveDelay, 1000 * 60);
 
   useEffect(() => {
-    if (accessToken) {
+    if (accessToken?.token) {
       // @NOTE see https://dailydotdev.atlassian.net/l/cp/dK9h1zoM
       // eslint-disable-next-line react-hooks/rules-of-hooks
```
idoshamun, line 23: We don't need this anymore right?

idoshamun (approved):
Let's hope it'll take RPS to the floor

omBratteng (approved):
Tested it locally, now only does one request for new access token 🦾

packages/shared/src/hooks/useRefreshToken.ts
rebelchris: We do, it's a hook call inside a hook call (technically not according to React rules 😅)

idoshamun: You gotta love these rules

capJavert: if you just rename it to something not starting with `use` we can remove it. `useRefresh` is not really a hook.

PR #2270, opened 2023-10-11
fix: replace mouseUp with auxClick

rebelchris:
- Replace mouseUp with auxClick
- Verified locally on events, will do extra check once deployed as well

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1883 #done

### DIFF
--- a/packages/shared/src/components/cards/AdLink.tsx
+++ b/packages/shared/src/components/cards/AdLink.tsx
@@ -15,7 +15,7 @@ export default function AdLink({ ad, onLinkClick }: AdLinkProps): ReactElement {
       rel="noopener"
       title={ad.description}
       onClick={() => onLinkClick?.(ad)}
-      onMouseUp={(event) => event.button === 1 && onLinkClick?.(ad)}
+      onAuxClick={(event) => event.button === 1 && onLinkClick?.(ad)}
     />
   );
 }
--- a/packages/shared/src/components/cards/AdLink.tsx
+++ b/packages/shared/src/components/cards/AdLink.tsx
@@ -1,6 +1,7 @@
 import React, { ReactElement } from 'react';
 import { Ad } from '../../graphql/posts';
 import { CardLink } from './Card';
+import { combinedClicks } from '../../lib/click';
 
 export type AdLinkProps = {
   ad: Ad;
@@ -14,8 +15,7 @@ export default function AdLink({ ad, onLinkClick }: AdLinkProps): ReactElement {
       target="_blank"
       rel="noopener"
       title={ad.description}
-      onClick={() => onLinkClick?.(ad)}
-      onAuxClick={(event) => event.button === 1 && onLinkClick?.(ad)}
+      {...combinedClicks(() => onLinkClick?.(ad))}
     />
   );
 }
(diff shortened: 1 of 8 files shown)


packages/shared/src/components/cards/AdLink.tsx
```diff
@@ -15,7 +15,7 @@ export default function AdLink({ ad, onLinkClick }: AdLinkProps): ReactElement {
       rel="noopener"
       title={ad.description}
       onClick={() => onLinkClick?.(ad)}
-      onMouseUp={(event) => event.button === 1 && onLinkClick?.(ad)}
+      onAuxClick={(event) => event.button === 1 && onLinkClick?.(ad)}
```
omBratteng, line 18: FYI auxclick is not [supported on Safari](https://caniuse.com/auxclick) 😬

rebelchris: Weird, I tested it now and it fires correctly on Safari 🤯
I'll try mobile to check as well

omBratteng: What is the intended action to happen?

rebelchris: Best example:
Card -> Read article button
Middle click it
Actions to happen:
- Open in new tab
- Track event for go to link

rebelchris: Never mind my mobile comment lol 😂

omBratteng: Well, with the new iPhone 17 with USB-C you can connect a mouse I think 🤓

omBratteng: My local environment has crashed hard right now, so I can't test locally.
Though I think the default action for middle click on macOS is to open links in new tab, probably why it works.

You could try to remove `onAuxClick` and test it in Safari and see if it still works

rebelchris: Yep you are totally right, removing onAuxClick will still work on Safari indeed.
It opens new tab + tracks the link

Chrome with it disabled:
Opens new tab + does NOT track the link

With auxClick they both work as intended

omBratteng (approved):

packages/shared/src/components/cards/AdLink.tsx
omBratteng: Perfect 😄

capJavert (approved):
Awesome, can we also replace all separate `onAuxClick` instances with `combinedClicks`?

packages/shared/src/components/cards/AdLink.tsx
capJavert: > My local environment has crashed hard right now, so I can't test locally.
Though I think the default action for middle click on macOS is to open links in new tab, probably why it works.

> You could try to remove onAuxClick and test it in Safari and see if it still works

Good to know, I know we tested `onAuxClick` on firefox and safari and it worked, but did not know about default behaviour 🤘

capJavert (approved):

PR #2273, opened 2023-10-13
revert: boot and features loading optimizations

capJavert:
Reverts dailydotdev/apps#2262

One thing that hit me after all the reviews/checking is that boundary will block `isFallback` state that basically just renders SEO for bots.

Reverting.


rebelchris (approved):
Yep good you spotted it early :)

PR #2274, opened 2023-10-13
feat: changelog popup redesign

nensidosari:
[WT-1796](https://dailydotdev.atlassian.net/browse/WT-1796)

## Changes
- Separated Changelog Tooltip from changelog on Sidebar
- Changed ChangelogTooltip Component from using BaseTooltip to using InteractivePopup
- Added functionality to upvote button 
- Added functionality to comment button (functionality same as the release notes button)
- Added notification icon on changelog sidebar 

Fixes changelog tooltip UI/UX when there are a lot of items in the sidebar, adds functionality to upvote and comment buttons

## Events

I did introduce one, not 100% sure where to add on the doc, please help 🙏 

| New | upvote post  | extra: { origin: 'changelog popup' } |

- [x] MobileL (420px)
- [x] Tablet (656px)
- [x] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1796 #done

[WT-1796]: https://dailydotdev.atlassian.net/browse/WT-1796?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


rebelchris (changes requested):
When trying to do a normal post (non edit) it fails:
Screenshot 2023-10-13 at 16 57 35

packages/shared/src/components/sidebar/SidebarBottomSection.tsx
```diff
@@ -41,7 +41,13 @@ export function SidebarBottomSection({
       icon: () => <ListIcon Icon={() => <TerminalIcon />} />,
       title: 'Changelog',
       path: `${process.env.NEXT_PUBLIC_WEBAPP_URL}sources/daily_updates`,
-      navItemRef,
+      alert: changelog.isAvailable && (
+        <AlertDot
+          className="right-5 mt-1.5"
+          color={AlertColor.Cabbage}
+          position={AlertDotPosition.Fixed}
```
rebelchris, line 48: Should probably still work with absolute right, as it's absolute based on the sidebar element?

packages/shared/src/components/tooltips/ChangelogTooltip.tsx
```diff
@@ -100,137 +86,114 @@ function ChangelogTooltip<TRef extends HTMLElement>({
...
+            className="right-4"
+            onClick={onModalCloseClick}
+            data-testid="changelogModalClose"
+            buttonSize={ButtonSize.XSmall}
+          />
+        </header>
+        <section className="flex flex-col flex-1 p-5 h-full shrink max-h-full">
+          <Image
+            className="object-cover rounded-lg w-[207px] h-[108px]"
+            alt="Post Cover image"
```
rebelchris, line 113:
```suggestion
            alt="Post cover image"
```
I see it was already a issue :)

rebelchris (commented):
Some early feedback :)

packages/shared/src/components/tooltips/ChangelogTooltip.tsx
nensidosari: No worries, can add it 🙌

packages/shared/src/components/sidebar/SidebarBottomSection.tsx
nensidosari: It is absolute based on the Icon element within the sidebar element, that's why keeping it absolute wouldn't work and had to add this.

rebelchris: Ah it should be fixed to the changelog element alone, I see now it works in relation to the whole sidebar.

rebelchris: I think something like this might work:
```
      rightIcon: (_) =>
        changelog.isAvailable && (
          <AlertDot className="-top-1 right-0" color={AlertColor.Cabbage} />
        ),
```

Would look like this:
Screenshot 2023-10-14 at 09 46 07

nensidosari: Ah yes, I hadn't explored the right icon prop. Thank you! 🙌 Will keep this and remove the additions to AlertDot.

packages/shared/src/components/tooltips/ChangelogTooltip.spec.tsx
```diff
@@ -85,8 +80,8 @@ describe('ChangelogTooltip component', () => {
               loadedAlerts
             >
               <AlertDot color={AlertColor.Cabbage} />
-              <div ref={anchorRef} />
-              <ChangelogTooltip elementRef={anchorRef} />
+              <div />
```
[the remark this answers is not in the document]
rebelchris: Also perhaps you want to consider adding a test for upvote since it's new functionality now. (optional if you up for the challenge)

packages/shared/src/components/tooltips/ChangelogTooltip.tsx
```diff
@@ -20,12 +13,12 @@ import { updateFirefoxExtensionLink } from '../../lib/constants';
...
-import SettingsContext from '../../contexts/SettingsContext';
-import { AlertColor, AlertDot } from '../AlertDot';
 import { checkIsExtension } from '../../lib/func';
+import { UserPostVote } from '../../graphql/posts';
+import InteractivePopup, { InteractivePopupPosition } from './InteractivePopup';
+import { Origin } from '../../lib/analytics';
 
-interface ChangelogTooltipProps<TRef> extends BaseTooltipProps {
-  elementRef: MutableRefObject<TRef>;
+interface ChangelogTooltipProps extends BaseTooltipProps {
```
rebelchris, line 21: I don't think the BaseTooltipProps are extended anymore right?

packages/shared/src/components/tooltips/ChangelogTooltip.tsx
```diff
@@ -99,138 +90,122 @@ function ChangelogTooltip<TRef extends HTMLElement>({
...
-      {...props}
-      content={
-        !!post && (
-          <div
-            className="flex relative flex-col whitespace-normal break-words rounded-16 border shadow-2 focus:outline-none w-[24rem] changelog bg-theme-bg-tertiary border-theme-color-cabbage"
-            data-testid="changelog"
+    !!post && (
+      <InteractivePopup
+        {...props}
+        position={InteractivePopupPosition.Unset}
```
rebelchris, line 101: More of a question, why did you choose unset and then manually add them?
My suggestion would be to add something like bottomLeft as a position potentially? (easier if we ever need to re-use the logic)

packages/shared/src/components/tooltips/ChangelogTooltip.tsx
```diff
@@ -99,138 +90,122 @@ function ChangelogTooltip<TRef extends HTMLElement>({
...
+        className="bottom-0 left-60 mb-6 ml-6 border shadow-2 focus:outline-none w-[24rem] max-w-[360px] max-h-[466px] changelog bg-theme-bg-tertiary border-theme-color-cabbage"
```
rebelchris, line 102: The `changelog` class probably does nothing

packages/shared/src/components/tooltips/ChangelogTooltip.tsx
```diff
@@ -99,138 +90,122 @@ function ChangelogTooltip<TRef extends HTMLElement>({
...
+          </time>
+          {!!post.summary && (
+            <div
+              className="mt-2 w text-theme-label-tertiary typo-callout"
+              data-testid="changelogSummary"
+            >
+              {post.summary}
+            </div>
+          )}
+          <div className="flex mt-4 font-bold w text-theme-color-salt typo-footnote">
```
rebelchris, line 149: The following should not be needed here.
font-bold w text-theme-color-salt typo-footnote

packages/shared/src/components/tooltips/ChangelogTooltip.tsx
```diff
@@ -99,138 +90,122 @@ function ChangelogTooltip<TRef extends HTMLElement>({
...
+            <Button
+              id={`post-${post.id}-upvote-btn`}
```
rebelchris, line 151: Don't think this is needed here.

packages/shared/src/components/tooltips/ChangelogTooltip.tsx
```diff
@@ -99,138 +90,122 @@ function ChangelogTooltip<TRef extends HTMLElement>({
...
+              icon={
+                <UpvoteIcon
+                  secondary={post?.userState?.vote === UserPostVote.Up}
+                />
+              }
+              pressed={post?.userState?.vote === UserPostVote.Up}
+              onClick={onToggleUpvote}
+              buttonSize={ButtonSize.Small}
+              className="btn-tertiary-avocado bg-theme-bg-tertiary"
```
rebelchris, line 160: I think it's actually `quarternary` from design perspective, the bg should not be needed.

packages/shared/src/components/tooltips/ChangelogTooltip.tsx
```diff
@@ -99,138 +90,122 @@ function ChangelogTooltip<TRef extends HTMLElement>({
...
+              data-testid="changelogUpvotesCounter"
```
rebelchris, line 161: Are you using this one?

packages/shared/src/components/tooltips/ChangelogTooltip.tsx
```diff
@@ -99,138 +90,122 @@ function ChangelogTooltip<TRef extends HTMLElement>({
...
-              <Image
-                className="object-cover rounded-lg w-[207px] h-[108px]"
-                alt="Post Cover image"
-                src={post.image}
-                fallbackSrc={cloudinary.post.imageCoverPlaceholder}
-                loading="lazy"
-                data-testid="changelogImage"
+            </Button>
+            <Button
+              id={`post-${post.id}-comment-btn`}
```
rebelchris, line 168: Also not needed

packages/shared/src/components/tooltips/ChangelogTooltip.tsx
```diff
@@ -99,138 +90,122 @@ function ChangelogTooltip<TRef extends HTMLElement>({
...
+              icon={<CommentIcon secondary={post.commented} />}
+              pressed={post.commented}
+              tag="a"
+              href={post.commentsPermalink}
+              onClick={dismissChangelog}
+              buttonSize={ButtonSize.Small}
+              className="btn-tertiary-blueCheese bg-theme-bg-tertiary"
```
rebelchris, line 175: Here as well the bg is not needed, and think it should be quarternary button.

packages/shared/src/components/tooltips/ChangelogTooltip.tsx
```diff
@@ -99,138 +90,122 @@ function ChangelogTooltip<TRef extends HTMLElement>({
...
+              data-testid="changelogCommentsCounter"
```
rebelchris, line 176: Please remove if not needed

packages/shared/src/components/tooltips/ChangelogTooltip.tsx
```diff
@@ -99,138 +90,122 @@ function ChangelogTooltip<TRef extends HTMLElement>({
...
+            href={post.commentsPermalink}
+            data-testid="changelogReleaseNotesBtn"
+          >
+            Release notes
+          </Button>
+          {isExtension && (
+            <Button
+              tag={isFirefoxExtension ? 'a' : undefined}
+              href={isFirefoxExtension ? updateFirefoxExtensionLink : undefined}
+              className="bg-cabbage-40 btn-primary"
```
rebelchris, line 198: I see this was already there, but it should actually be 
```suggestion
              className="btn-primary-cabbage"
```

packages/shared/src/hooks/useChangelog.ts
```diff
@@ -1,27 +1,65 @@
...
   dismiss: () => Promise<void>;
+  toggleUpvote: (props: ToggleVoteProps) => Promise<void>;
 };
 
 export function useChangelog(): UseChangelog {
   const { alerts, updateAlerts } = useContext(AlertContext);
   const { sidebarRendered } = useSidebarRendered();
+  const client = useQueryClient();
+
+  const changelogQueryKey = ['changelog', 'latest-post'];
```
rebelchris, line 20: For query keys we should always aim to use `generateQueryKey` function it will ensure us the order is always matching expectations.

rebelchris (changes requested):
Wow amazing work and really lean on the hook implementation.

Have some minor styling/guidelined concerns.
Almost there it seems.

packages/shared/src/components/tooltips/ChangelogTooltip.spec.tsx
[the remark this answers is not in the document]
nensidosari: Yeah for sure, will do that now 😄

packages/shared/src/components/tooltips/ChangelogTooltip.tsx
nensidosari: True, will remove, thank u!

packages/shared/src/components/tooltips/ChangelogTooltip.tsx
nensidosari: Well, `BottomLeft` would be something similar to what is already there `LeftEnd` (bottom-8, left-8).  What we would want in this case is something set LeftEnd but only on the main feed container, the one already there takes the popup on top of the sidebar. The already specified bottom is also more then we need based on design (we need a bottom-6 rather then bottom-8)
We could create something like MainFeedLeftEnd (with `left-60` which is the sidebar size, and either bottom-6 or bottom-0 and mb-6), we would just not be picking from the already defined props there but would need to create new ones... my thoughts, at the moment this seems to be something quite specific to this use case, not reusable, thus the unset prop solution came to mind.

But maybe by not knowing the props that much I might be missing some other way how we might do this that makes sense and is more reusable, suggestions?

packages/shared/src/lib/analytics.ts
```diff
@@ -11,6 +11,7 @@ export enum Origin {
   TagsSearch = 'tags search',
   RealTime = 'realtime',
   NonRealTime = 'nonrealtime',
+  ChangelogPopup = 'changelog popup',
```
capJavert, line 14: This should be added to analytics changes table in PR description. 🙏

packages/shared/src/hooks/useChangelog.ts
capJavert: Also, you can add `RequestKey.Changelog` to be used instead of `changelog` string.

capJavert (commented):
Nice work, aside from missing tests in changelog hook and component (i see you are already on it) looks solid. 🚀

packages/shared/src/components/tooltips/ChangelogTooltip.tsx
nensidosari: Yes, it was already in the existing tests.

packages/shared/src/components/tooltips/ChangelogTooltip.tsx
nensidosari: This one already in the existing tests as well. It just checks whether the number shows up there or not.

packages/shared/src/lib/analytics.ts
nensidosari: Thank you! Just did that, if you can take another look just to make sure I haven't messed up something there and have set the right values. 🙏

packages/shared/src/hooks/useChangelog.ts
nensidosari: Where would 'latest-post' fit there? Something like this:
`generateQueryKey(RequestKey.Changelog, null, RequestKey.LatestPost)`

capJavert: `latestPost` can be just a string, since it's a sub resource if that makes sense. We use `RequestKey` only for top level resource so we can more easily invalidate.

capJavert: Other then that yeah, exactly like you did it 👌

packages/shared/src/components/tooltips/ChangelogTooltip.tsx
rebelchris: I see LeftEnd is actually never used, probably was a proposal. so for all I care you can overwrite that one.
Just double check I didn't miss anything.

packages/shared/src/components/tooltips/ChangelogTooltip.tsx
rebelchris: Ah wasn't aware we already checked that :) 
Then never mind this comment.

packages/shared/src/components/tooltips/ChangelogTooltip.tsx
nensidosari: Only needed if I convert the button to QuarternaryButton

packages/shared/src/components/tooltips/ChangelogTooltip.tsx
nensidosari: So the quarternary button gives the green background on hover, while if I use button and add the bg theme, it removes that background. Thought that was needed from the design.
Screenshot 2023-10-20 at 12 15 59 PM
Screenshot 2023-10-20 at 12 17 09 PM

Is there any other way to make the quarternary button to not have that green background when hovered?

packages/shared/src/components/tooltips/ChangelogTooltip.tsx
nensidosari: Same comment as on the upvote button -> https://github.com/dailydotdev/apps/pull/2274#discussion_r1366779234

packages/shared/src/components/tooltips/ChangelogTooltip.tsx
nensidosari: LeftEnd is composed of `classNames(leftClass, endClass)`, both of those classes are used in other positions, so in this case that LeftEnd would need to be composed of these 2 new classes I need (left-60, bottom-6), which can make it confusing... If instead I do name it `MainFeedLeftEnd` I can create 2 classes, mainFeedLeft = left-60 and mainFeedEnd = bottom-6, which at least would make it a bit more clear.

packages/shared/src/hooks/useChangelog.ts
nensidosari: Cool, thank you!!

packages/shared/src/components/tooltips/ChangelogTooltip.tsx
rebelchris: Sure let's do that just for re-usability :)

packages/shared/src/components/InteractionCounter.tsx
```diff
@@ -24,7 +25,7 @@ export default function InteractionCounter({
   }, [value]);
 
   if (shownValue === value) {
-    return <>{shownValue}</>;
+    return <span {...props}>{shownValue}</span>;
```
rebelchris, line 28: Just make sure this doesn't break any other implementations of this component.

rebelchris (commented):
Nice refactor, just think the background on the button is the only one left?

packages/shared/src/components/tooltips/ChangelogTooltip.tsx
nensidosari: Agreed to keep QuaternaryButton component to alight with the post cards buttons on the feed.

packages/shared/src/components/InteractionCounter.tsx
nensidosari: Yeah looking at tests now, usually span shouldn't break anything design wise, but 🤷‍♀️ ...

nensidosari:
> Nice refactor, just think the background on the button is the only one left?

@rebelchris that is done as well, but remaining the test on the upvote functionality. Struggling a bit to test that as it's through the hook.

packages/shared/src/components/tooltips/ChangelogTooltip.tsx
rebelchris: Yep, just remove the background please here :)

PR #2277, opened 2023-10-16
feat: search referral campaign

sshanzel:
Note: we have not yet performed an internal sync relevant to the integration. With that, we have left TODOs to remember areas we need updating.

## Changes
- Dynamic join page based on key value map.
  - Do not redirect automatically when the user is logged in as there can be campaigns that are not only targetting anonymous users.
  - Redirection based on campaign should be case-to-case basis.
- Introduce Search referral page.
- Search referral campaign popup.
- New hook to identify whether the user has the capacity to invite
  - By checking wether there is a token found in the URL returned by the query.
  - In BE you can only have a token once it is found in the invites table.
  - At the end of the day, it will be verified by the mutation whether you can actually allow giving out access.
- New settings property to identify the placement of the CTA.
- Display a pulse on the search input after accepting the feature.

TODO:
- ~~Display 404 page when the feature/campaign is not found in the key value map.~~
- End-to-end tests with integration working.
- ~~Search page banner's copy changes when the available keys left is 0.~~
- Validate the placement of the banner when there is a result.
- ~~Fix tests.~~
- ~~Events tracking~~ ~~waiting for the click/popup event discussion.~~

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1894 #done


packages/shared/src/components/ProfileMenu.tsx
```diff
@@ -4,10 +4,16 @@ import { Item } from '@dailydotdev/react-contexify';
...
 import AuthContext from '../contexts/AuthContext';
 import PowerIcon from './icons/Power';
-import { UserIcon } from './icons';
+import { KeyReferralIcon, UserIcon } from './icons';
 import DevCardIcon from './icons/DevCard';
 import SettingsIcon from './icons/Settings';
 import { IconSize } from './Icon';
+import { useLazyModal } from '../hooks/useLazyModal';
+import { LazyModal } from './modals/common/types';
+import TimerIcon from './icons/Timer';
```
rebelchris, line 13: Should also be from the main index export

packages/shared/src/components/ProfileMenu.tsx
```diff
@@ -58,6 +72,21 @@ export default function ProfileMenu(): ReactElement {
           </a>
         </Link>
       </Item>
+      {showSearchReferral && (
+        <Item>
+          <button
+            type="button"
+            className="flex items-center min-w-[12.5rem]"
```
rebelchris, line 79: The min-width feels odd, why is it needed?

rebelchris: Also why is it a normal html button?
Our own one would allow left and right icons right?

packages/shared/src/components/ProfileMenu.tsx
```diff
@@ -4,10 +4,16 @@ import { Item } from '@dailydotdev/react-contexify';
 import Link from 'next/link';
 import AuthContext from '../contexts/AuthContext';
 import PowerIcon from './icons/Power';
-import { UserIcon } from './icons';
+import { KeyReferralIcon, UserIcon } from './icons';
 import DevCardIcon from './icons/DevCard';
 import SettingsIcon from './icons/Settings';
 import { IconSize } from './Icon';
+import { useLazyModal } from '../hooks/useLazyModal';
+import { LazyModal } from './modals/common/types';
```
rebelchris, line 12: Don't think you ever need this here.

packages/shared/src/components/feeds/FeedContainer.tsx
```diff
@@ -118,10 +121,21 @@ export const FeedContainer = ({
   const isFinder = router.pathname === '/posts/finder';
   const isV1Search =
     searchValue === SearchExperiment.V1 && showSearch && !isFinder;
+  const shouldShowPulse =
+    checkHasCompleted(ActionType.AcceptedSearch) &&
+    !checkHasCompleted(ActionType.UsedSearch);
```
rebelchris, line 126: Can't this just be one, usedSearch? if not we can always pulse no?

packages/shared/src/components/feeds/FeedContainer.tsx
```diff
@@ -118,10 +121,21 @@ export const FeedContainer = ({
...
   const onSearch = (event: FormEvent, input: string) => {
     event.preventDefault();
     router.push(`${webappUrl}search?q=${encodeURIComponent(input)}`);
   };
+  const handleSearchFocus = () => {
+    if (!shouldShowPulse) {
+      return;
+    }
+
+    completeAction(ActionType.UsedSearch);
```
rebelchris, line 137: Maybe we should complete this on submit rather, just heavy to use focus for it.

packages/shared/src/components/icons/Key/KeyReferral/index.tsx
```diff
@@ -0,0 +1,10 @@
+import React, { ReactElement } from 'react';
+import Icon, { IconProps } from '../../../Icon';
+import OutlinedIcon from './outlined.svg';
+import FilledIcon from './filled.svg';
+
+export const KeyReferralIcon = (props: IconProps): ReactElement => (
+  <Icon {...props} IconPrimary={FilledIcon} IconSecondary={OutlinedIcon} />
```
rebelchris, line 7: Is the outlined correct? looks like a big blur?

packages/shared/src/components/modals/referral/SearchReferralModal.tsx
```diff
@@ -0,0 +1,79 @@
...
+import { useSettingsContext } from '../../../contexts/SettingsContext';
+import { CampaignCtaPlacement } from '../../../graphql/settings';
+import useMedia from '../../../hooks/useMedia';
+import { laptop } from '../../../styles/media';
+import { cloudinary } from '../../../lib/image';
+import { KeysRow } from './KeysRow';
+import { ReferralCampaignKey, useReferralCampaign } from '../../../hooks';
+
+function SearchReferralModal(modalProps: ModalProps): ReactElement {
+  const isLaptop = useMedia([laptop.replace('@media ', '')], [true], false);
```
rebelchris, line 16: Cries in pain, we really need to fix this hook

packages/shared/src/components/modals/referral/SearchReferralModal.tsx
```diff
@@ -0,0 +1,79 @@
...
+            have ${availableCount} invite keys left, use them wisely.`}
+          </p>
+          <KeysRow count={5} />
+          <Button
+            className={classNames(
+              'mt-5',
+              noKeysAvailable ? 'btn-secondary' : 'btn-primary',
+            )}
+            // TODO: action when no keys available
+            icon={noKeysAvailable ? null : <CopyIcon />}
```
rebelchris, line 55: We know the action here now right?

packages/shared/src/components/modals/referral/SearchReferralModal.tsx
```diff
@@ -0,0 +1,79 @@
...
+            alt="tablet popup background"
+          />
+        </span>
+        <div className="flex flex-col laptop:p-2 w-full max-w-[26rem]">
+          <h1 className="laptop:mt-4 font-bold text-center laptop:text-left typo-title3 tablet:typo-title1 laptop:typo-mega2">
+            {noKeysAvailable
+              ? `Need more keys?`
+              : `Give your friends early access to daily.dev's search!`}
+          </h1>
+          <p className="mt-6 text-center laptop:text-left !font-normal typo-body tablet:typo-headline laptop:typo-title3 text-theme-label-secondary">
```
rebelchris, line 42: Why do you need the override on normal, which one adds bold by default?

packages/shared/src/components/search/SearchFeedback.tsx
```diff
@@ -3,39 +3,33 @@ import { WidgetContainer } from '../widgets/common';
...
-        href={searchDocs}
-        icon={<DocsIcon />}
-        className="btn-tertiary"
-        buttonSize={ButtonSize.Small}
-      >
-        User guide
-      </Button>
-    </WidgetContainer>
-  </PageWidgets>
+  <WidgetContainer className="flex justify-around p-4">
```
rebelchris, line 10: Are you sure this didn't introduce edge-cases the order was quite important from what I recall.

packages/shared/src/components/search/SearchReferralBanner.tsx
```diff
@@ -0,0 +1,34 @@
...
+  });
+
+  if (!canInvite) {
+    return null;
+  }
+
+  const { availableCount, noKeysAvailable } = referralCampaign;
+
+  return (
+    <div className="flex flex-col gap-4 p-4 w-full rounded-14 bg-theme-overlay-from">
```
rebelchris, line 19: The background feels incomplete here?

packages/shared/src/components/search/SearchReferralBanner.tsx
```diff
@@ -0,0 +1,34 @@
...
+      <KeyReferralIcon size={IconSize.Large} className="mr-2" />
+      <h2 className="font-bold typo-title2">
+        {noKeysAvailable
```
rebelchris, line 22: Where did this copy come from?
Can't see if in design so just missing the context here.
I mainly think it's incomplete as for the button for instance

packages/shared/src/components/search/SearchReferralBanner.tsx
```diff
@@ -0,0 +1,34 @@
...
+          ? `Give more friends access to daily.dev's search!`
+          : `Give your friends early access to daily.dev's search!`}
+      </h2>
+      <p className="typo-title3 text-theme-label-secondary">
+        {noKeysAvailable
+          ? `We noticed you have used all your invites wisely. Here's a way to get access to some more invites!`
+          : `Be that cool friend who got access to yet another AI feature! You have ${availableCount} invites, use them wisely.`}
+      </p>
+      <Button className="btn-primary">Give early access</Button>
```
rebelchris, line 31: What's the action here?

packages/shared/src/hooks/referral/useReferralCampaign.ts
```diff
@@ -0,0 +1,115 @@
...
+}
+
+export type UseReferralCampaignProps = {
+  campaignKey: ReferralCampaignKey;
+};
+
+export const campaignToReferralTargetCountMap: Record<
+  ReferralCampaignKey,
+  number
+> = { search: 5 };
```
rebelchris, line 35: Why is this on the frontend?

packages/shared/src/hooks/referral/useReferralCampaign.ts
```diff
@@ -0,0 +1,115 @@
...
+
+const campaignFeatureFlagMap: Partial<
+  Record<ReferralCampaignKey, Feature<string>>
+> = {
+  search: feature.search,
+};
```
rebelchris, line 41: This should also come from the backend I think.

packages/shared/src/hooks/referral/useReferralCampaign.ts
```diff
@@ -0,0 +1,115 @@
...
+        { requestKey: JSON.stringify(queryKey) },
+      );
+
+      return result.referralCampaign;
+    },
+    {
+      enabled: !!user?.id && !!isCampaignEnabled,
+    },
+  );
+  const referralCountLimit =
```
rebelchris, line 71: What will happen when someone has 20/50?

packages/shared/src/hooks/referral/useReferralCampaign.ts
```diff
@@ -0,0 +1,115 @@
...
+    data?.referralCountLimit ?? campaignToReferralTargetCountMap[campaignKey];
+  const referredUsersCount = data?.referredUsersCount ?? 0;
+  const referralCurrentCount =
+    referredUsersCount > referralCountLimit
+      ? referralCountLimit
+      : referredUsersCount;
```
rebelchris, line 77: In general a lot of calculation happening here.
I don't think it's needed actually.

packages/shared/src/hooks/referral/useReferralCampaign.ts
```diff
@@ -0,0 +1,115 @@
...
+    referralCampaign,
+    get canInvite() {
+      try {
+        const link = new URL(url);
+
+        return link.searchParams.has('token');
+      } catch (err) {
+        return false;
+      }
+    },
```
rebelchris, line 111: Not sure I'm following this one correctly?

packages/shared/src/lib/featureManagement.ts
```diff
@@ -21,7 +21,7 @@ const feature = {
   feedVersion: new Feature('feed_version', 15),
   onboardingV2: new Feature('onboarding_v2', OnboardingV2.Control),
   onboardingV3: new Feature('onboarding_v3', OnboardingV3.Control),
-  search: new Feature('search', SearchExperiment.Control),
+  search: new Feature('search', SearchExperiment.V1),
```
rebelchris, line 24: Be aware :)

packages/webapp/components/invite/AISearchInvite.tsx
```diff
@@ -0,0 +1,100 @@
...
+}: JoinPageProps): ReactElement {
+  const router = useRouter();
+  const search = useFeature(feature.search);
+  const { completeAction } = useActions();
+  const { user, showLogin } = useAuthContext();
+  const { mutateAsync: onAcceptMutation } = useMutation(
+    acceptFeatureInvitation,
+    {
+      onSuccess: () => {
+        completeAction(ActionType.AcceptedSearch);
```
rebelchris, line 31: Can't we simply do the action in BE?
Or you want it to persist quickly in FE for checks?

(Both fine though, but should only be after accepting is applied right?

packages/webapp/components/invite/AISearchInvite.tsx
```diff
@@ -0,0 +1,100 @@
...
+    return onAcceptMutation({ token, referrerId: referringUser.id });
+  };
+
+  useEffect(() => {
+    if (search === SearchExperiment.Control) {
+      return;
+    }
+
+    router.push(campaignConfig.search.redirectTo);
+  }, [router, search]);
```
rebelchris, line 55: I'm not sure this is correct.
The user got a invite for search correct, so surely they will be on control then?

packages/webapp/components/invite/AISearchInvite.tsx
```diff
@@ -0,0 +1,100 @@
...
+
+  return (
+    <div className="flex relative flex-col flex-1 justify-center p-6 h-full min-h-[100vh]">
```
rebelchris, line 58: Not really urgent, but wonder if we should make a seperate layout for these things?

packages/webapp/pages/join.tsx
```diff
@@ -1,71 +1,32 @@
...
-  const isDesktop = useMedia([laptop.replace('@media ', '')], [true], false);
+const componentsMap: ReferralRecord<FunctionComponent<JoinPageProps>> = {
+  search: AISearchInvite,
+};
 
-  const onVideoModalClose = () => {
-    setVideoOpen(false);
-  };
+const Page = ({ referringUser, campaign }: JoinPageProps): ReactElement => {
+  const Component = componentsMap[campaign];
```
rebelchris, line 23: Maybe we should fallback to the "default" join page if no campaign set?

rebelchris (changes requested):
Wow amazing work.
Just leaving my code wise comments, will look into testing it now :)

packages/shared/src/hooks/referral/useReferralCampaign.ts
DragosIorgulescu: maybe something like:
```suggestion
    data?.referralCountLimit ?? campaignToReferralTargetCountMap[campaignKey];
  const referredUsersCount = data?.referredUsersCount ?? 0;
  const referralCurrentCount = Math.min(referredUsersCount, referralCountLimit);
```

packages/shared/src/components/ProfileMenu.tsx
sshanzel: Cool. Let me add it. Think this was a legacy one and was not included in the index import.

packages/shared/src/components/ProfileMenu.tsx
sshanzel: This is a button in the profile menu item. It would be very enlarged for the context menu.

packages/shared/src/components/feeds/FeedContainer.tsx
sshanzel: If we make it just one, everyone without going through the flow would receive the pulse. Not sure if that was the intended behavior here in the first place.

packages/shared/src/components/ProfileMenu.tsx
sshanzel: Can you elaborate on why?

packages/shared/src/components/feeds/FeedContainer.tsx
sshanzel: Cool. It is just a bit annoying to type something while the pulse was showing, in my own experience.

packages/shared/src/components/modals/referral/SearchReferralModal.tsx
sshanzel: Yep. Updated locally now.

packages/shared/src/components/modals/referral/SearchReferralModal.tsx
sshanzel: The `typo-headline` automatically applies bold.

packages/shared/src/components/search/SearchFeedback.tsx
sshanzel: I will be testing more of the layout on preview deployment. One of the TODOs in this PR.

packages/shared/src/components/search/SearchReferralBanner.tsx
sshanzel: Do you want me to check with design?

Screenshot 2023-10-18 at 6 30 26 PM

Screenshot 2023-10-18 at 6 30 38 PM

packages/shared/src/components/search/SearchReferralBanner.tsx
sshanzel: Should be from this message: https://dailydotdev.slack.com/archives/C05NRH3U31C/p1697443536775929?thread_ts=1697387678.267679&cid=C05NRH3U31C

packages/shared/src/components/icons/Key/KeyReferral/index.tsx
sshanzel: Let me check the asset.

sshanzel: After checking, yes, that is expected. That's the disabled state.

packages/shared/src/components/search/SearchReferralBanner.tsx
sshanzel: Should open the modal either way. Updated now.

packages/shared/src/hooks/referral/useReferralCampaign.ts
sshanzel: This used to be in FE. The TS was screaming when I introduced this. Should be applicable to campaigns only I think.

packages/shared/src/hooks/referral/useReferralCampaign.ts
sshanzel: It's been there since then. I think it was to put a ceiling on the number to display.

packages/shared/src/hooks/referral/useReferralCampaign.ts
sshanzel: Replaced by the property returned by API now.

packages/webapp/pages/join.tsx
sshanzel: When the component is not found or developed, we return 404.

packages/webapp/components/invite/AISearchInvite.tsx
sshanzel: Just wanted it to be quick since we will redirect where they will see the search bar.

packages/shared/src/hooks/referral/useReferralCampaign.ts
sshanzel: Let's check with product.

packages/shared/src/hooks/referral/useReferralCampaign.ts
sshanzel: We need it here to identify which feature is for what campaign. This is also to understand whether they have access to it already, and if so we redirect.

packages/shared/src/hooks/referral/useReferralCampaign.ts
sshanzel: After some discussion, we will cap both at 99.

packages/webapp/components/invite/AISearchInvite.tsx
sshanzel: We might need to collaborate with the design team so we can have a system we can rely on. Wdyt?

sshanzel:
@rebelchris should be ready for review. Once we get the API merged, let's sync with Dragos to test it further.

packages/webapp/pages/join.tsx
```diff
@@ -1,71 +1,41 @@
...
+}: JoinPageProps): ReactElement => {
+  const Component = componentsMap[campaign];
+  const { title, description, images, redirectTo } = useReferralConfig({
+    campaign,
+    referringUser,
+  });
 
   useEffect(() => {
-    if (!user?.id) {
+    if (!Object.hasOwn(componentsMap, campaign)) {
```
capJavert, line 35: Why not just `componentsMap[campaign]`? ..its more intuitive and has to check less thing internally.

packages/webapp/components/invite/AISearchInvite.tsx
```diff
@@ -0,0 +1,132 @@
...
+    return handleAccept();
+  };
+
+  useEffect(() => {
+    if (search === SearchExperiment.Control) {
+      return;
+    }
+
+    router.push(redirectTo);
+  }, [redirectTo, router, search]);
```
capJavert, line 84: Router is not stable in our version of next https://github.com/vercel/next.js/issues/18127, probably best to remove it for now, look for `router reference is not stable` in codebase.

packages/shared/src/hooks/referral/useReferralConfig.ts
```diff
@@ -0,0 +1,37 @@
...
+  campaign,
+  referringUser,
+}: UseReferralConfigProps): CampaignConfig => {
+  const getDefaultTitle = () =>
+    `${referringUser.name} invites you to use daily.dev`;
+  const getDefaultDescription = () =>
+    `daily dev is a professional network for developers to learn, collaborate, and grow together. Developers come to daily.dev to discover a wide variety of professional knowledge, create groups where they can collaborate with other developers they appreciate, and discuss the latest trends in the developer ecosystem.`;
+  const getDefaultImages = () => [
+    { url: `https://og.daily.dev/api/refs/${referringUser.id}` },
+  ];
```
capJavert, line 20: Why not just raw values instead of functions?

capJavert (commented):
Pretty neat implementation from the size of the DR, nice work 👍

packages/webapp/components/invite/AISearchInvite.tsx
```diff
@@ -0,0 +1,132 @@
...
+        onLoginSuccess: handleAccept,
+        onRegistrationSuccess: handleAccept,
+      });
+    }
+
+    // since in the page view, query params are tracked automatically,
+    // we don't need to send the params here explicitly
+    trackEvent({ event_name: AnalyticsEvent.AcceptInvitation });
+
+    return handleAccept();
```
DragosIorgulescu, line 75: I think this should be 
```suggestion
    return handleAccept;
```

This is most likely the reason we did not see this working in our e2e tests?

sshanzel: Right. Let's commit this one.

packages/webapp/pages/join.tsx
sshanzel: Don't wanna add more dependency in the array 😅

packages/shared/src/hooks/referral/useReferralConfig.ts
sshanzel: Ah, I moved this from a generic function to a hook. Yep, probably we can go with just raw values now since we have all the information already.

packages/webapp/components/invite/AISearchInvite.tsx
sshanzel: Nice. Thanks for the info. Updated now 🚢

packages/webapp/pages/join.tsx
capJavert: Since component map is defined outside of component and it is not reactive you will not have to add it to array either way.

packages/shared/src/components/ProfileMenu.tsx
rebelchris: Why do we need the LazyModal types here, the useLazy will handle everything right?

packages/shared/src/components/feeds/FeedContainer.tsx
rebelchris: Ah good point, let's leave it :)

packages/webapp/components/invite/AISearchInvite.tsx
rebelchris: Yeah 100%, let's take it for a next one :)

packages/shared/src/components/feeds/FeedContainer.tsx
```diff
@@ -110,6 +117,19 @@ export const FeedContainer = ({
   } as CSSProperties;
   const cardContainerStyle = { ...getStyle(isList, spaciness) };
   const suggestionsProps = useSearchSuggestions({ origin: Origin.HomePage });
+  const isTracked = useRef(false);
```
rebelchris, line 120: Why is this useRef and not useState?

packages/shared/src/components/modals/referral/SearchReferralModal.tsx
```diff
@@ -0,0 +1,122 @@
...
+      extra: JSON.stringify({ token: referralToken }),
+    });
+    copyLink();
+  };
+
+  const handleToggle = () => {
+    trackEvent({
+      event_name: AnalyticsEvent.HideFromHeader,
+      target_type: TargetType.HideInviteCheckbox,
+      target_id: !campaignCtaPlacement,
```
rebelchris, line 45: Is this correct?
value states true/false, but is campaignCtaPlacement not a string in our case?

So the evaluation would not be correct I think

packages/shared/src/lib/analytics.ts
```diff
@@ -101,7 +101,12 @@ export enum AnalyticsEvent {
   UpvoteSearch = 'upvote search',
   DownvoteSearch = 'downvote search',
   CopySearch = 'copy search',
+  CopyKeyLink = 'copy key link',
+  HideFromHeader = 'hide from header',
```
rebelchris, line 105: Taxonomy states `hide invite mechanism`

packages/webapp/pages/join.tsx
sshanzel: Right. Thought you wanted to use the already defined variable `const Component = componentsMap[campaign]`. In that case, we can definitely switch to that.

packages/shared/src/lib/analytics.ts
```diff
@@ -101,7 +101,12 @@ export enum AnalyticsEvent {
...
+  CloseInvitationPopup = 'close invitation popup',
   ErrorSearch = 'error search',
+  AcceptInvitation = 'accept invitation',
+  SearchHighlightAnimation = 'search highlight animation',
```
rebelchris, line 109: Taxonomy states `highlight search`

packages/shared/src/lib/links.ts
```diff
@@ -23,4 +23,8 @@ export const link = {
   post: {
     create: `${webappUrl}squads/create`,
   },
+  search: {
+    requestKeys:
+      'mailto:hi@daily.dev?subject=I want more invites for daily.dev search',
```
rebelchris, line 28: Nice!

packages/shared/src/components/ProfileMenu.tsx
sshanzel: To set the `type` of modal we want to open. For example: `openModal({ type: LazyModal.SearchRefferal })`.

rebelchris (changes requested):
Cool only have some minor tracking issues.

packages/shared/src/components/feeds/FeedContainer.tsx
sshanzel: We don't need the re-render for when it is tracked using a state. We just want to make sure it is tracked

packages/shared/src/lib/analytics.ts
sshanzel: Ah, must be recent change. Let me update this too.

capJavert (approved):

DragosIorgulescu (approved):
great work 🙌

packages/shared/src/components/modals/referral/SearchReferralModal.tsx
```diff
@@ -0,0 +1,123 @@
...
+    });
+    copyLink();
+  };
+
+  const handleToggle = () => {
+    trackEvent({
+      event_name: AnalyticsEvent.HideFromHeader,
+      target_type: TargetType.HideInviteCheckbox,
+      target_id: campaignCtaPlacement === CampaignCtaPlacement.Header,
+      // we need expected value after clicking the checkbox, that means, the previous value was header.
```
rebelchris, line 46: The taxonomy states True/False only.
Did you and Dave agree on anything different?

rebelchris (changes requested):
Just one more issue that's not aligned now.

sshanzel:
Previews:

Screenshot 2023-10-22 at 6 12 41 PM
Screenshot 2023-10-22 at 6 11 49 PM
Screenshot 2023-10-22 at 6 11 41 PM
Screenshot 2023-10-22 at 6 06 22 PM
Screenshot 2023-10-22 at 6 06 14 PM
Screenshot 2023-10-22 at 6 05 56 PM
Screenshot 2023-10-22 at 6 05 48 PM
Screenshot 2023-10-22 at 6 05 37 PM
Screenshot 2023-10-22 at 6 05 21 PM
Screenshot 2023-10-22 at 6 05 09 PM
Screenshot 2023-10-22 at 6 04 55 PM

packages/shared/src/components/search/SearchContainer.tsx
```diff
@@ -16,27 +20,40 @@ export function SearchContainer({
...
-      </div>
-    </main>
+      </main>
+      <PageWidgets tablet={false} className="gap-6 items-center !px-0">
+        <SearchFeedback className="order-2 laptop:order-1 max-w-widget" />
+        {!!content && (
+          <SearchSourceList
+            className={classNames(
+              'order-1 laptop:order-1',
+              chunk?.status && 'laptop:mt-6 laptop:pt-0.5',
```
sshanzel, line 49: The only place where I feel it is a bit hacky is the spacing we apply here on laptop. Though imo, it is better than rendering three same elements and placing conditions on each instance.

packages/shared/src/components/search/SearchResult.tsx
```diff
@@ -37,7 +38,7 @@ interface ContainerProps extends WithClassNameProps {
 const ContentPill = classed(Pill, 'w-full !h-3');
 
 const Container = ({ children, className }: ContainerProps) => (
-  <main className="order-2 laptop:order-3 col-span-2 px-4 laptop:px-8 mb-5 w-full">
+  <main className="order-2 laptop:order-3 col-span-2 w-full b-5">
```
rebelchris, line 41: b-5 seems odd?

rebelchris (approved):
Just one semantic issue, otherwise looks and feels ok :)

packages/shared/src/components/search/SearchResult.tsx
sshanzel: Oh, that should be removed 🚢

PR #2279, opened 2023-10-18
feat: squad other section

rebelchris:
## NOTE:

WIP: lots of TODO's and TS errors, but love to hear your overall thoughts.

Idea:
- lean feed is less dependant on props and can work independant
- lean actions should eventually be used by everything so there is one sole hook for actions per post (it needs a good cleanup/split maybe even)
- modals should work via lazyModal and able to sit on top (still todo)

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done


packages/shared/src/hooks/post/useLeanPostActions.ts
```diff
@@ -0,0 +1,464 @@
...
+  posts.map((post) =>
+    post.id === id
+      ? {
+          ...post,
+          ...update(post),
+        }
+      : post,
+  );
+
+const updateMap = {
```
rebelchris, line 50: The final idea is to have one cache for post states, regardless of who needs it. 
As the "intermediate" solution I thought we could have this mapping per query to say what transformer to run on the provided cache.  (one issue still is how to update multiple places)

packages/shared/src/components/widgets/FurtherReadingSquad.tsx
```diff
@@ -0,0 +1,121 @@
...
+
+  return (
+    <div className="flex flex-col flex-1 w-full bg-theme-overlay-float-avocado">
+      <div className="w-full border-b border-b-theme-divider-tertiary">
+        <p className="p-4 tablet:px-8 font-bold typo-callout text-theme-label-tertiary">
+          More posts from squad_name
+        </p>
+      </div>
+      <div className={classNames(className, 'flex flex-col gap-6')}>
+        <ActiveFeedContextProvider items={similarPosts} queryKey={queryKey}>
```
rebelchris, line 106: Items here can actually be removed, but had to leave it for legacy stuff for now.

packages/shared/src/contexts/ActiveFeedContext.tsx
```diff
@@ -1,11 +1,118 @@
...
+  } = useLeanPostActions({
+    queryKey,
+  });
+
+  // console.log(virtualGrid);
+  const data: ActiveFeedContextValue = useMemo(
+    () => ({
+      items,
+      queryKey,
+      onClick: onClick || onClickDefault,
```
rebelchris, line 77: We either can overwrite or keep the default click action.

packages/shared/src/components/feed/LeanFeed.tsx
```diff
@@ -0,0 +1,18 @@
+import React, { ReactElement } from 'react';
+import LeanFeedItemComponent from './LeanFeedItemComponent';
+import { useActiveFeedContext } from '../../contexts';
+import { FeedItem } from '../../hooks/useFeed';
+
+function LeanFeed({ items }: { items: FeedItem[] }): ReactElement {
+  const { feedRef } = useActiveFeedContext();
+
+  return (
+    <div className="grid grid-cols-2 gap-4" ref={feedRef}>
```
idoshamun, line 10: If this is to be a generic component, I suggest accepting classes from the outside and set it with default value

packages/shared/src/hooks/post/useLeanPostActions.ts
idoshamun: I suggest looking for normalized cache solutions or ignoring it, and in any way I wouldn't recommend doing it in the same PR. Developing one ourselves is too much effort and bug-prone

packages/shared/src/components/feed/LeanFeed.tsx
rebelchris: Good point indeed, most likely will take default "feed settings" and can get overwritten from outside.

packages/shared/src/components/feed/LeanFeed.tsx
```diff
@@ -0,0 +1,18 @@
...
+      {items.map((item, i) => {
+        return <LeanFeedItemComponent key={i} item={item} />;
+      })}
```
capJavert, line 13: It could accept `children` as a function.  So we can call:

Render cards:
```jsx
<LeanFeed>
    {({ items }) => {
        return items.map((item, i) => {
            return <LeanFeedItemComponent key={i} item={item} />;
        }) 
    }}
</LeanFeed>
```

Add ad to the first place:
```jsx
<LeanFeed>
    {({ items }) => {
        return items.map((item, i) => {
            if (i === 0) {
                return (
                    <>
                        <Ad />
                        <LeanFeedItemComponent key={i} item={item} />;
                    </>
                )
            }
    
            return <LeanFeedItemComponent key={i} item={item} />;
        }) 
    }}
</LeanFeed>
```

Of course internal render prop function is in fact just a component:
```jsx
const FeedWithAds =  {({ items }) => {
    return items.map((item, i) => {
        return <LeanFeedItemComponent key={i} item={item} />;
    }) 
}}

<LeanFeed>
    {FeedWithAds}
</LeanFeed>
```

I think conceptually it would allow us to split view and business logic.

packages/shared/src/hooks/post/useLeanPostActions.ts
```diff
@@ -0,0 +1,464 @@
...
+
+const updatePost =
+  (
+    queryClient: QueryClient,
+    queryKey: string[],
+    transformKey: string,
+    update: (oldPost: Post) => Partial<Post>,
+  ): ((args: { id: string }) => Promise<() => void>) =>
+  async ({ id }) => {
+    await queryClient.cancelQueries(queryKey);
```
capJavert, line 75: Why cancel queries?

packages/shared/src/contexts/ActiveFeedContext.tsx
```diff
@@ -1,11 +1,118 @@
...
+    onBlockTag,
+    onClick: onClickDefault,
+    onBookmark,
+    onUpvote,
+    onDownvote,
+    onHidePost,
+  } = useLeanPostActions({
+    queryKey,
+  });
+
```
capJavert, line 71: Why do we need this top level, it should be called just in nested components?

packages/shared/src/hooks/post/useLeanPostActions.ts
```diff
@@ -0,0 +1,464 @@
...
+    );
+    return () => {
+      queryClient.setQueryData(queryKey, previousData);
+    };
+  };
+
+export default function useLeanPostActions({
+  queryKey,
+  transformKey = 'further-reading',
+}) {
```
capJavert, line 89: I like the idea but I think component would need to be:
- split up into smaller parts/hooks
- based on your comment [here](https://dailydotdev.slack.com/archives/C02E2C3C13R/p1697703893665599?thread_ts=1697634724.796189&cid=C02E2C3C13R) it should support upvoting from single post as well
 - we should also see if it can utilize mutation subscriptions to sync state between different caches (as current hooks do)
 - should it solve props drilling in cards? feels like it should
 - don't get me wrong it looks great just with current functionality I think it's lacking any benefits vs current set of hooks 🙏 🚀

packages/shared/src/contexts/ActiveFeedContext.tsx
```diff
@@ -1,11 +1,118 @@
...
+      onPromotePost,
+      onBanPost,
+      canDeletePost,
+      onDeletePost,
+      canPinPost,
+      onPinPost,
+      onBlockSource,
+      onBlockTag,
+      queryKey,
+    ],
```
capJavert, line 108: We need to be careful, this currently is going to make feed rendering as slow as it is now or worse. 

Basically it should function the same as `LeanOptionsButton` where hooks down the render tree call `useLeanPostActions` and context only provides bare minimum like items and queryKey.

packages/shared/src/hooks/post/useLeanPostActions.ts
```diff
@@ -0,0 +1,464 @@
...
+      queryKey,
+      updateMap[transformKey].transformer({ previousData, id, update }),
+    );
+    return () => {
+      queryClient.setQueryData(queryKey, previousData);
+    };
+  };
+
+export default function useLeanPostActions({
+  queryKey,
```
capJavert, line 87: Technically, if we give single post query key it should invalidate single post cache?

capJavert (commented):
Great work on the POC, appreciate the effort 🚀 

Excited to plan and move it forward.

packages/shared/src/contexts/ActiveFeedContext.tsx
rebelchris: This can actually move into the nested components indeed.
The only exposure here I think was in case you can't to overwrite handlers?

Not sure if it's needed for that.

packages/shared/src/hooks/post/useLeanPostActions.ts
rebelchris: Yeah 100% inline with the idea, Indeed the mutation subscription could potentially just check all caches for that postId and do something with it :)

And fully agree it should be split up again in singular logic, this was easier from a start point only.

packages/shared/src/hooks/post/useLeanPostActions.ts
rebelchris: Yep!

packages/shared/src/components/cards/ActionButtons.tsx
```diff
@@ -132,7 +102,14 @@ export default function ActionButtons({
             />
           </QuaternaryButton>
         </SimpleTooltip>
-        {insaneMode && lastActionButton}
```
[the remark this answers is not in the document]
rebelchris: It's basically list mode at this point 😅

packages/shared/src/contexts/ActiveFeedContext.tsx
capJavert: You do the custom logic in that specialized component?

rebelchris: But let's say it's `canPinPost` the page itself defines this.
The specialized component would have no way of knowing about it, unless it's contexted/prop drilled correct? (as it's implementation sits quite deep)

capJavert: `canPinPost` would be something that is in the context then yes and component can decide what to do with it.

PR #2282, opened 2023-10-19
fix: comment box input focus on the end

sshanzel:
## Changes
- This fixes the issue when replying to someone and mention is included, focus will be at the end of the content.
- This also fixes the issue when you edit a comment, the focus would automatically be then at the end.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1896 #done

### DIFF
--- a/packages/shared/src/components/fields/MarkdownInput/index.tsx
+++ b/packages/shared/src/components/fields/MarkdownInput/index.tsx
@@ -6,6 +6,7 @@ import React, {
   ReactElement,
   ReactNode,
   TextareaHTMLAttributes,
+  useEffect,
   useImperativeHandle,
   useRef,
   useState,
@@ -35,6 +36,7 @@ import { useAuthContext } from '../../../contexts/AuthContext';
 import { Loader } from '../../Loader';
 import { Divider } from '../../utilities';
 import { usePopupSelector } from '../../../hooks/usePopupSelector';
+import { focusInput } from '../../../lib/textarea';
 
 interface ClassName {
   container?: string;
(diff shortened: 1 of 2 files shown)


rebelchris (approved):

capJavert (commented):
I see more and more hacks with inputs like this due to not having reactive state instance in recent times. I get it we don't have other option now but it looks like a lot of tech debt.

packages/shared/src/components/fields/MarkdownInput/index.tsx
```diff
@@ -150,6 +152,16 @@ function MarkdownInput(
...
 
+  useEffect(() => {
+    const content = textareaRef?.current?.value;
+
+    if (!content) {
+      return;
+    }
+
+    focusInput(textareaRef.current, [content.length, content.length]);
+  }, []);
```
capJavert, line 163: I see more and more hacks with inputs like this due to not having reactive state instance in recent times. I get it we don't have other option now but it looks like a lot of tech debt.

rebelchris: Very valid point, for now lets proceed, ill add it as discussion topic for web. Around moving to form lib. (See ido also mentioned one)

Is that ok?

capJavert: Yes of course, don't think of it as blocking.

omBratteng (approved):

packages/shared/src/components/fields/MarkdownInput/index.tsx
rebelchris: Most likely a form library.
There are some suggestions floating for instance https://tanstack.com/form/latest

sshanzel: I see. Yep. I would be glad if we could use any available packages like that, or `react-form-hooks` can be a good option as well. We can totally refactor our input components by then.

PR #2284, opened 2023-10-19
fix: regex handle uniformity

sshanzel:
## Changes
- Make the APIs recent changes be the basis of the regex for handles and socials.

Note: since this regex won't allow the character dash `-`, it would affect the comment box as this is also the basis of the mentions triggers. Do we maybe want to allow it in the mention in the meantime?

Screenshot 2023-10-19 at 11 02 22 AM

Screenshot 2023-10-19 at 11 02 31 AM

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1897 #done

### DIFF
--- a/packages/webapp/pages/account/profile.tsx
+++ b/packages/webapp/pages/account/profile.tsx
@@ -16,20 +16,12 @@ import CameraIcon from '@dailydotdev/shared/src/components/icons/Camera';
 import Textarea from '@dailydotdev/shared/src/components/fields/Textarea';
 import { useToastNotification } from '@dailydotdev/shared/src/hooks/useToastNotification';
 import { IconSize } from '@dailydotdev/shared/src/components/Icon';
-import {
-  handleRegex,
-  socialHandleRegex,
-} from '@dailydotdev/shared/src/graphql/users';
-import { UserProfile } from '@dailydotdev/shared/src/lib/user';
 import { getAccountLayout } from '../../components/layouts/AccountLayout';
 import AccountContentSection from '../../components/layouts/AccountLayout/AccountContentSection';
 import { AccountPageContainer } from '../../components/layouts/AccountLayout/AccountPageContainer';
 import { AccountTextField } from '../../components/layouts/AccountLayout/common';
 
 const id = 'avatar_file';
-
-type Socials = Pick<UserProfile, 'github' | 'hashnode' | 'twitter'>;
-const socials: Array<keyof Socials> = ['github', 'hashnode', 'twitter'];
 
 const AccountProfilePage = (): ReactElement => {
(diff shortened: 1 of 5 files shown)


rebelchris:
Let's keep the frontend uniform to what backend accepts.
Ido will proceed with retro fixing the handles for existing users soon, so should have no more of the broken ones moving forward. *([See discussion here](https://dailydotdev.slack.com/archives/C02AA18RD1D/p1697547604920889?thread_ts=1697547443.746619&cid=C02AA18RD1D))

packages/webapp/pages/account/profile.tsx
```diff
@@ -34,6 +42,28 @@ const AccountProfilePage = (): ReactElement => {
     e.preventDefault();
     const values = formToJson<UpdateProfileParameters>(formRef.current);
 
+    if (values.username !== user.username) {
```
rebelchris, line 45: Nice this was the only place it wasn't validating?
(Pretty sure we have it in onboarding/signup flows, but maybe double check the toast)

packages/webapp/pages/account/profile.tsx
```diff
@@ -34,6 +42,28 @@ const AccountProfilePage = (): ReactElement => {
...
+      const isValid = handleRegex.test(values.username);
+
+      if (!isValid) {
+        displayToast('Invalid characters found in username!');
```
rebelchris, line 49: Also maybe we can make this uniform somehow? (as there are multiple flows of setting/changing username)

rebelchris (approved):
Nice, just some validations needed, otherwise good to go.
(This would cover tagging as well?)

packages/webapp/pages/account/profile.tsx
sshanzel: Right. Let me verify that end. If so, it would probably be better to have this inside the calling function to cover all areas.

sshanzel:
> (This would cover tagging as well?)

Yes, the regex used is based on the [same one](https://github.com/dailydotdev/apps/blob/WT-1897-handles-regex-fix/packages/shared/src/hooks/input/useMarkdownInput.ts#L210) we edited.

Also tested the registration form now.

Screenshot 2023-10-20 at 1 38 05 PM

Screenshot 2023-10-20 at 1 43 25 PM

packages/webapp/pages/account/profile.tsx
sshanzel: Made it a generic function now 🚢

rebelchris (approved):
Nice!

PR #2285, opened 2023-10-19
chore: don't pass `sourceId` on update post

DragosIorgulescu:
## Changes

leftover item from the original PR for editing posts: https://github.com/dailydotdev/apps/pull/2259#discussion_r1360551010
API changes: https://github.com/dailydotdev/daily-api/pull/1487

## Events

n / a

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1801 #done

### DIFF
--- a/packages/shared/src/graphql/squads.ts
+++ b/packages/shared/src/graphql/squads.ts
@@ -53,7 +53,7 @@ type EditSquadOutput = {
 
 type PostToSquadProps = {
   id: string;
-  sourceId: string;
+  sourceId?: string;
   commentary: string;
 };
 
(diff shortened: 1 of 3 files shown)


packages/shared/src/graphql/squads.ts
```diff
@@ -53,7 +53,7 @@ type EditSquadOutput = {
 
 type PostToSquadProps = {
   id: string;
-  sourceId: string;
+  sourceId?: string;
```
rebelchris, line 56: Is this one used anywhere still?

rebelchris (approved):
Just one more question.

packages/shared/src/graphql/squads.ts
DragosIorgulescu: this interface is shared between the create & update mutation. Left it as optional since it's needed on create

PR #2288, opened 2023-10-23
fix: search spacing between the bar and the result

sshanzel:
## Changes
- Fix the distance between the search bar and the result.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done

### DIFF
--- a/packages/shared/src/components/search/SearchContainer.tsx
+++ b/packages/shared/src/components/search/SearchContainer.tsx
@@ -46,7 +46,7 @@ export function SearchContainer({
           <SearchSourceList
             className={classNames(
               'order-1 laptop:order-1',
-              chunk?.status && 'laptop:mt-6 laptop:pt-0.5',
+              chunk?.status && 'laptop:mt-12 laptop:pt-0.5',
             )}
             sources={chunk?.sources}
             isLoading={isLoading}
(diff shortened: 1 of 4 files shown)


rebelchris (approved):

packages/shared/src/components/search/SearchContainer.tsx
```diff
@@ -42,14 +42,16 @@ export function SearchContainer({
       </main>
       <PageWidgets tablet={false} className="gap-6 items-center !px-0">
         <SearchFeedback className="order-2 laptop:order-1 max-w-widget" />
-        {!!content && (
+        {!!chunk && (
```
rebelchris, line 45: Should we not evaluate if the chunk is the sources?

sshanzel: The chunk is the whole object, this would just let us know if something is processing, so we can either display the loading state, or the list itself once the value is there. The component itself checks the sources list.

DragosIorgulescu (approved):
looks good code-wise, but I wanted to ask if we should add previews as requirements for PRs like these, it always helps see UI the changes as well in a few screenshots

rebelchris (approved):
Let's give it a try, we might need to test it on the prod version.

sshanzel:
> looks good code-wise, but I wanted to ask if we should add previews as requirements for PRs like these, it always helps see UI the changes as well in a few screenshots

Yeah, my bad. The screenshot was in the thread of this issue, I'll make sure to include them on the next ones.

> Let's give it a try, we might need to test it on the prod version.

Let's do it 🚢

PR #2289, opened 2023-10-23
feat: search squad members

sshanzel:
## Changes
- In the Squad members modal, we should allow search.
- Fixed the generate query key function to not accept any kind of string but a rather more strict typing.

Preview:
Screenshot 2023-10-23 at 6 52 50 PM
Screenshot 2023-10-23 at 6 36 22 PM
Screenshot 2023-10-23 at 6 36 05 PM
Screenshot 2023-10-23 at 6 53 25 PM

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1815 #done

### DIFF
--- a/packages/shared/src/lib/query.ts
+++ b/packages/shared/src/lib/query.ts
@@ -10,11 +10,18 @@ import { EmptyObjectLiteral } from './kratos';
 import { LoggedUser } from './user';
 import { FeedData, Post, ReadHistoryPost } from '../graphql/posts';
 import { ReadHistoryInfiniteData } from '../hooks/useInfiniteReadingHistory';
+import { MainFeedPage } from '../components/utilities';
+
+export enum OtherFeedPage {
+  Squad = 'squad',
+}
+
+export type AllFeedPages = MainFeedPage | OtherFeedPage;
 
 export type MutateFunc<T> = (variables: T) => Promise<(() => void) | undefined>;
 
 export const generateQueryKey = (
-  name: RequestKey,
+  name: RequestKey | AllFeedPages,
   user: Pick<LoggedUser, 'id'> | null,
   ...additional: unknown[]
 ): unknown[] => {
@@ -47,7 +54,6 @@ export enum RequestKey {
   CurrentSession = 'current_session',
   PersonalizedDigest = 'personalizedDigest',
   Changelog = 'changelog',
-  MyFeed = 'my-feed',
 }
 
 export type HasConnection<
--- a/packages/shared/src/lib/query.ts
+++ b/packages/shared/src/lib/query.ts
@@ -10,7 +10,7 @@ import { EmptyObjectLiteral } from './kratos';
 import { LoggedUser } from './user';
 import { FeedData, Post, ReadHistoryPost } from '../graphql/posts';
 import { ReadHistoryInfiniteData } from '../hooks/useInfiniteReadingHistory';
-import { MainFeedPage } from '../components/utilities';
+import { SharedFeedPage } from '../components/utilities';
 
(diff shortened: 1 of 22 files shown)


packages/shared/src/lib/query.ts
```diff
@@ -10,11 +10,22 @@ import { EmptyObjectLiteral } from './kratos';
 import { LoggedUser } from './user';
 import { FeedData, Post, ReadHistoryPost } from '../graphql/posts';
 import { ReadHistoryInfiniteData } from '../hooks/useInfiniteReadingHistory';
+import { MainFeedPage } from '../components/utilities';
```
rebelchris, line 13: I'm not a big fan of the naming, it's a bit unclear for me even what is a main feed page and what is not, but it has nothing to do with this PR so will not mark it as fix.

packages/shared/src/lib/query.ts
```diff
@@ -10,11 +10,22 @@ import { EmptyObjectLiteral } from './kratos';
...
+  SearchBookmarks = 'search-bookmarks',
+}
+
+export type AllFeedPages = MainFeedPage | OtherFeedPage;
 
 export type MutateFunc<T> = (variables: T) => Promise<(() => void) | undefined>;
 
 export const generateQueryKey = (
-  name: string | RequestKey,
+  name: RequestKey | AllFeedPages,
```
rebelchris, line 28: Can't we make the RequestKey extend AllFeedPages or Include them in some way there?
I don't want to start adding exceptions only in certain places.

rebelchris (changes requested):
Thanks for the cleanup as well on the enums.
Only have some minor remarks.

packages/shared/src/lib/query.ts
sshanzel: Right. I agree with that. Probably SharedFeedPage since those are the shared feeds with the extension.

packages/shared/src/lib/query.ts
sshanzel: I wanted to do that too, but I found out enum is not extend-able 😭

However, I think this is the "only exception" for this implementation. So I kinda made my peace with it 😅

packages/shared/src/lib/query.ts
sshanzel: Should be refactored here https://github.com/dailydotdev/apps/pull/2289/commits/c92c2df1f7af7d5f1144fb7afb1d7b0b10abf7f5

packages/shared/src/lib/query.ts
rebelchris: I do think there are [ways](https://dev.to/afrazkhan/merge-typescript-enums-4mh7) to do it.
But it's not urgent so up to you :)

rebelchris (approved):
Approved from my side.
Optional if you want to explore the enum thing, not urgent.

packages/shared/src/lib/query.ts
sshanzel: Oh, nice. Thank you for the reference. I learned something new today. I think we can take this on after refactoring our feed flow. The querying part might be affected by the upcoming refactoring, so for the moment, I would try to keep the changes lean for now. Wdyt?

rebelchris: Yeah no problem at all :)

sshanzel:
@rebelchris seems like the issue with the request was the page param. This will mitigate sending false values, but I think what we should fix is the API end on its pagination part.

image

PR #2291, opened 2023-10-23
feat: collapse/expand squad pinned posts

DragosIorgulescu:
- added support for pinned posts toggle mutations
- use the `collapsePinnedPosts` flag on the source flag to determine if pinned posts should be displayed for the current user
 
## Preview

collapse-pinned-posts

## Events

N / A

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done


packages/shared/src/components/squads/SquadFeedHeading.tsx
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    return countPinned;
+  }, [items]);
+
+  return (
+    <div className="flex flex-row flex-wrap gap-4 justify-end items-center pb-2 w-full">
+      <span className="flex flex-row gap-3 pl-3 ml-auto border-l border-theme-divider-tertiary">
+        <Button
+          className="btn-tertiaryFloat"
```
DragosIorgulescu, line 47: The button does not match the styling exactly, the alternative is `btn-secondary` but that does not match the design spec either:

Screenshot 2023-10-23 at 17 36 18
Screenshot 2023-10-23 at 17 36 27

rebelchris: Non existing button I think, I'm [asking design](https://dailydotdev.slack.com/archives/C05LMGUD8HZ/p1698127221954679) if we can use the tertiaryFloat for it.

rebelchris: So [end](https://dailydotdev.slack.com/archives/C05LMGUD8HZ/p1698129789642549?thread_ts=1698127221.954679&cid=C05LMGUD8HZ) decision is to use `btn-tertiaryFloat` as you proposed.

packages/shared/src/components/Feed.tsx
```diff
@@ -113,6 +115,7 @@ export default function Feed<T>({
   showSearch = true,
   besideSearch,
   actionButtons,
+  squad,
```
rebelchris, line 118: I don't think we should introduce squads on feeds, it doesn't feel right.

packages/shared/src/components/FeedItemComponent.tsx
```diff
@@ -138,6 +141,13 @@ export default function FeedItemComponent({
 
   switch (item.type) {
     case 'post': {
+      if (
+        !!item.post.pinnedAt &&
+        squad?.currentMember?.flags?.collapsePinnedPosts
```
rebelchris, line 146: Technically the items source should have this knowledge I think.

packages/shared/src/components/squads/SquadFeedHeading.tsx
```diff
@@ -0,0 +1,65 @@
...
+      if (item.type === 'post' && !!item.post.pinnedAt) {
+        countPinned += 1;
+      } else {
+        break;
+      }
+    }
+
+    return countPinned;
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [itemsCount]);
```
rebelchris, line 46: Maybe we should use reduce here, I recall it being a little bit faster, and also a bit cleaner.

rebelchris (changes requested):
My only concern is that we probably should avoid prop drilling a squad to a feed.

packages/shared/src/components/squads/SquadFeedHeading.tsx
capJavert: `break` here will terminate the loop early if there is a post that is not pinned in array, so the count will be wrong

agree `reduce` will be cleaner here

also, why do we need `eslint-disable` here?

DragosIorgulescu: the goal here was to have the counter break early from the iteration since we know pinned posts will always be at the beginning of the list, so it would be inefficient to iterate through the whole array when we will know the pinned posts count after the first few iterations.

I agree that it does not look clean (although it's a little bit funny how vanilla code seems `dirty` now to us 😅 ), but if we are fine with the performance hit here, we can change it to a `reduce` or something else.

> also, why do we need eslint-disable here?

@capJavert I'm using the `items.length` as the dependency which causes linter to complain because it is expecting the `items` object to be passed in

capJavert: > the goal here was to have the counter break early from the iteration since we know pinned posts will always be at the beginning of the list, so it would be inefficient to iterate through the whole array when we will know the pinned posts count after the first few iterations.

Oh I see, I think problem here with this optimization is that it takes for granted that pinned posts will be first, if that was to change the code would not work. I think there is no need to do a premature optimization here.

> @capJavert I'm using the items.length as the dependency which causes linter to complain because it is expecting the items object to be passed in

If squad feed refetches and we get same number of items but different posts are pinned then the count would again be wrong. This is the exact case where that lint rule helps to avoid edge cases and unintentional bugs down the line.

packages/shared/src/components/FeedItemComponent.tsx
DragosIorgulescu: sorry, I missed this piece of feedback -- .the flags info is indeed also available via `item.post.source.currentMember.flags...`, but then if I use this, calling the mutation does not update the UI, since we are updating the squad flags cache on mutation success: https://github.com/dailydotdev/apps/blob/ce183c6fe9855d1795509af1a16660b97d55f7ee/packages/shared/src/hooks/squads/useSquadActions.ts#L70

We could get around this by invalidating the feed query cache, but then we are making an additional API call which we don't really need: https://github.com/dailydotdev/apps/blob/bfb276714e1aa6eeefe3a262bd092457c461baf7/packages/shared/src/hooks/squads/useSquadActions.ts#L56-L72

packages/shared/src/components/squads/SquadFeedHeading.tsx
DragosIorgulescu: updated 🙏  I still feel like we are sacrificing performance for readability here, but I guess since the squads are paginated we will only ever go through a few 10s of records when doing this, so that's fine

packages/shared/src/components/squads/SquadFeedHeading.tsx
```diff
@@ -0,0 +1,55 @@
...
+import { QueryKey } from 'react-query';
+import { Button } from '../buttons/Button';
+import { PinIcon } from '../icons';
+import { Squad } from '../../graphql/sources';
+import { ActiveFeedContext } from '../../contexts';
+import { useSquadActions } from '../../hooks';
+
+interface SquadFeedHeadingProps {
+  squad: Squad;
+  feedQueryKey: QueryKey;
```
rebelchris, line 11: What do you need this one for here?

packages/shared/src/hooks/squads/useSquadActions.ts
```diff
@@ -55,6 +62,23 @@ export const useSquadActions = ({
     onSuccess: () => client.invalidateQueries(membersQueryKey),
   });
 
+  const { mutateAsync: collapseSquadPinnedPosts } = useMutation(
+    collapsePinnedPosts,
+    {
+      onSuccess: () => {
+        client.invalidateQueries(feedQueryKey);
```
rebelchris, line 69: Generally it's best practice to perform optimistic updates if we know what the mutation will be.
In this case we could manually update the cache instead of simply invalidating it.

rebelchris (commented):
Nice, some minor feedback :)

packages/shared/src/hooks/squads/useSquadActions.ts
capJavert: I agree with @DragosIorgulescu in this case invalidation is better then optimistic update since it is not something crucial to the user.  As long as we validate actual pin is applied (no api caching is involved).

packages/shared/src/components/squads/SquadFeedHeading.tsx
rebelchris: @DragosIorgulescu Just this question remains I think.

DragosIorgulescu: this is to pass it down when invalidating the feed cache, I pushed one more commit to fix an issue with the cache update which was not causing the UI to reflect the changes properly. Let me know if this is good to merge

rebelchris: Ah but you can get it from the context no?
No need to pass it I think?

DragosIorgulescu: ah, my bad, loading it from context is much better. Updated 🙏

packages/webapp/pages/squads/[handle]/index.tsx
```diff
@@ -195,17 +202,15 @@ const SquadPage = ({
         <Feed
           className="px-6 pt-14 laptop:pt-10"
           feedName={OtherFeedPage.Squad}
-          feedQueryKey={[
-            'sourceFeed',
-            user?.id ?? 'anonymous',
-            Object.values(queryVariables),
-          ]}
+          feedQueryKey={feedQueryKey}
```
Comment (capJavert) [inline packages/webapp/pages/squads/[handle]/index.tsx:205]: This change is not needed anymore, so if you can adjust as it was to avoid unnecessary diff

capJavert (approved):
Good work!

packages/webapp/pages/squads/[handle]/index.tsx
Comment (DragosIorgulescu) [reply packages/webapp/pages/squads/[handle]/index.tsx]: > This change is not needed anymore, so if you can revert as to avoid unnecessary diff

but then how will I be able to gloat about how many lines of changes my PRs have? 😄

Comment (capJavert) [reply packages/webapp/pages/squads/[handle]/index.tsx]: 😆 touché!

rebelchris (approved):
Nice work

PR #2295, opened 2023-10-24
fix: notification item render as link

nensidosari:
## Changes
- Changed button on NotificationItem component to a Link with a tag
- Removed router usage as the behaviour is now managed by the link
- Added tests

- Replaces button functionality in notification items for link functionality (the notification can now be opened in a new tab)

https://github.com/dailydotdev/apps/assets/36197304/caaa4873-c33d-4e64-8d96-ce1d761c6416

## Events

### **Please make sure existing components are not breaking/affected by this PR**
Have tested manually the notifications page and all looks as it was. There is a change in behaviour where since the notification was a button, it would be activated with Space key, now since it is a link, it's default behaviour is to be activated by Enter Key.

- [x] MobileL (420px)
- [x] Tablet (656px)
- [x] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1891 #done


packages/shared/src/components/notifications/NotificationItem.tsx
```diff
@@ -53,59 +56,70 @@ function NotificationItem({
...
-        />
-      )}
-      <NotificationItemIcon
-        icon={icon}
-        iconTheme={notificationTypeTheme[type]}
-      />
-      <div className="flex flex-col flex-1 ml-4 w-full text-left typo-callout">
-        {hasAvatar && (
-          <span className="flex flex-row gap-2 mb-4">{avatarComponents}</span>
+      <div
```
sshanzel, line 77: I agree that we should turn this into a link/anchor tag, though the reason we made the recent change to div, there was another anchor tag nested here. What we need to do here is just change the `button` element to an anchor tag.

Also, every item in the notifications table, there would be an `onClick` value passed down for tracking purposes.

sshanzel (changes requested):
Looks great! Also, thank you for adding a test.

As mentioned, the only expected change here is the `button` element to be changed to an `anchor` tag. By nature, all of our notification items are a link.

packages/shared/src/components/notifications/NotificationItem.tsx
rebelchris: I think Lee is right, we should leverage the existing button to be converted to a `a tag` to maintain the order of having multiple a tags spanned here.

nensidosari: Yeah makes sense, was checking the implementation in history to make it consistent, but probably the issue you mentioned Lee on having nested a tags was not there. 
Cool, I'll just convert from using the button there to using an `a tag`. 

@sshanzel I'm not sure I understood correctly what you ment by:

> Also, for every item on the notifications page, there would be an onClick value passed down for tracking purposes. Meaning, they would always turn out to be not a link.

I've kept the onClick functionality there, so clicking the `a tag` would fire onClick as well in order to send the analytics event. Did you mean something else I've missed?

packages/shared/src/components/notifications/NotificationItem.tsx
```diff
@@ -60,12 +62,16 @@ function NotificationItem({
...
-          onClick={onClick}
-        />
+        <Link href={targetUrl} passHref>
+          {/* eslint-disable-next-line jsx-a11y/click-events-have-key-events, jsx-a11y/anchor-has-content, jsx-a11y/no-static-element-interactions */}
+          <a
+            aria-label="Open notification"
+            className="absolute inset-0 z-0"
+            onClick={onClick}
+            title="Open notification"
+            data-testid="openNotification"
```
rebelchris, line 72: It seems you are not using this right? (the testId)

packages/shared/src/components/notifications/NotificationItem.tsx
```diff
@@ -60,12 +62,16 @@ function NotificationItem({
...
     >
       {onClick && (
-        <button
-          type="button"
-          aria-label="Open notification"
-          className="absolute inset-0 z-0"
-          onClick={onClick}
-        />
+        <Link href={targetUrl} passHref>
+          {/* eslint-disable-next-line jsx-a11y/click-events-have-key-events, jsx-a11y/anchor-has-content, jsx-a11y/no-static-element-interactions */}
```
rebelchris, line 66: Ideally we wouldn't have these, but I see some we can't fix.
Also was wondering if we would prefer the `Button` with `tag=a` or not.
Not really sure when we make the switch now.

rebelchris (approved):
Nothing blocking, just minor cleanup

packages/shared/src/components/notifications/NotificationItem.tsx
nensidosari: I am using it in tests

packages/shared/src/components/notifications/NotificationItem.tsx
nensidosari: I tried that, but the our button component would need more modifications to it to make it work out of the box as this does work. So that's why I didn't use it. Though not sure either when we are supposed to use that 🤔

packages/shared/src/components/notifications/NotificationItem.tsx
sshanzel: Oh, I just went back and checked the condition, thought it was the other way around about making it an anchor tag or a button. Kindly disregard that part 🚢

packages/shared/src/components/notifications/NotificationItem.tsx
sshanzel: Agree on using a plain element since there won't be any content/UI to this. Just the disabling of eslint if we can trim it down. I remember this was the exact reason why we used the button but didn't realize yet the impact on the right click.

sshanzel (approved):
Amazing work! 🚢 

Non-blocking comment with regards to eslint but I believe this brings back everything to normal 🥳

packages/shared/src/components/notifications/NotificationItem.tsx
sshanzel: I think you forgot to clean up the disabled eslints before merging.

nensidosari: Oh sorry, didn't get that right. So we want to fix the eslints so we won't disable them? All are required at the moment with the current implementation. 
We could fix the key events one by adding a key event handler.

rebelchris: I think we might need the doubleClick handler which the button already handled.
It might cut down that ESlint.

The others I think we can't fix and have multiple issues with in different places.

PR #2299, opened 2023-10-24
feat: add squad pinned posts reordering

DragosIorgulescu:
API: https://github.com/dailydotdev/daily-api/pull/1496

## Changes

- adding support for & using the `swappedPinnedPosts` mutation

TODO:
- [x] use a separate mutation for swapping
- [x] agree on a long term solution

## Preview

pinned-post-reordering

## Events

No new events to track here, for now

- [x] MobileL (420px)
- [x] Tablet (656px)
- [x] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1812 #done


packages/shared/src/components/Feed.tsx
```diff
@@ -371,6 +371,8 @@ export default function Feed<T>({
       );
     },
     post,
+    prevPost: (items[postMenuIndex - 1] as PostItem)?.post,
+    nextPost: (items[postMenuIndex + 1] as PostItem)?.post,
```
rebelchris, line 375: Oef not against the implementation, but thinking about a leanFeed, this is yet another constraint in moving it.

packages/shared/src/graphql/posts.ts
```diff
@@ -519,8 +519,18 @@ export const editPost = async (
 };
 
 export const PIN_POST_MUTATION = gql`
-  mutation UpdatePinPost($id: ID!, $pinned: Boolean!) {
-    updatePinPost(id: $id, pinned: $pinned) {
+  mutation UpdatePinPost(
+    $id: ID!
+    $pinned: Boolean!
+    $nextPostId: ID
+    $prevPostId: ID
```
rebelchris, line 526: Do we need this change?

rebelchris (changes requested):
Nice work, very minor questions otherwise ok.

sshanzel (commented):
Great work. Nothing blocking but we should maybe include the new icons on the index file of the folder itself.

packages/shared/src/graphql/posts.ts
DragosIorgulescu: good catch, removed ✂️

DragosIorgulescu:
> Great work. Nothing blocking but we should maybe include the new icons on the index file of the folder itself.

@sshanzel good point, however, does tree shaking drop any unused imports if exported the way these are in [index](https://github.com/dailydotdev/apps/blob/f5191754485e855f295539d2e1797672c88e19ca/packages/shared/src/components/icons/index.ts)?

DragosIorgulescu:
closing this here as I manually squash merged into main

PR #2302, opened 2023-10-24
fix: prevent empty source list render

rebelchris:
- Prevent empty source render when done loading but sources still empty

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done

### DIFF
--- a/packages/shared/src/components/search/SearchSourceList.tsx
+++ b/packages/shared/src/components/search/SearchSourceList.tsx
@@ -29,6 +29,10 @@ export const SearchSourceList = ({
     limit: 3,
   });
 
+  if (!isLoading && !sources) {
+    return null;
+  }
+
   const handleSourceClick = (): void =>
     !sidebarRendered && setIsSourcesOpen((prev) => !prev);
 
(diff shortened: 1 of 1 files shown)


packages/shared/src/components/search/SearchSourceList.tsx
```diff
@@ -29,6 +29,10 @@ export const SearchSourceList = ({
     limit: 3,
   });
 
+  if (!isLoading && !sources) {
+    return null;
+  }
```
capJavert, line 34: Should we check `sources?.length > 0`?

rebelchris: Good point, changed now.

capJavert (approved):

packages/shared/src/components/search/SearchSourceList.tsx
```diff
@@ -29,6 +29,10 @@ export const SearchSourceList = ({
     limit: 3,
   });
 
+  if (!isLoading && sources?.length === 0) {
```
sshanzel, line 32: I think the condition should have been `!sources?.length`. Undefined or null would be false on this condition.

sshanzel: Nvm. I see the other PR now with the right condition.

PR #2305, opened 2023-10-25
fix: workaround to avoid eslint-disable

nensidosari:
## Changes
- added tabIndex={0} and role="link" in order to avoid disabling eslint for `jsx-a11y/no-static-element-interactions`
- added empty <span/> tag in order to avoid disabling eslint for `jsx-a11y/anchor-has-content`
- added onKeyDown event to make the link be clicked and work with Spacebar as well (as it was working as a button) in order to avoid disabling eslint for `jsx-a11y/click-events-have-key-events`

- Cleans up code from eslint-disbales

## Events

- [ ] Have you done sanity checks in the extension? No
- [ ] Does this not break anything in companion? No

- [x] MobileL (420px)
- [x] Tablet (656px)
- [x] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1891 #done

### DIFF
--- a/packages/shared/src/components/notifications/NotificationItem.tsx
+++ b/packages/shared/src/components/notifications/NotificationItem.tsx
@@ -9,6 +9,7 @@ import NotificationItemAttachment from './NotificationItemAttachment';
 import NotificationItemAvatar from './NotificationItemAvatar';
 import { notificationTypeTheme } from './utils';
 import OptionsButton from '../buttons/OptionsButton';
+import { KeyboardCommand } from '../../lib/element';
 
 export interface NotificationItemProps
   extends Pick<
@@ -78,7 +79,7 @@ function NotificationItem({
             data-testid="openNotification"
             tabIndex={0}
             onKeyDown={(e) => {
-              if (e.code === 'Space') {
+              if (e.code === KeyboardCommand.Space) {
                 onClick(e);
                 router.push(targetUrl);
               }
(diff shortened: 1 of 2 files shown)


packages/shared/src/components/notifications/NotificationItem.tsx
```diff
@@ -63,14 +69,24 @@ function NotificationItem({
...
+            role="link"
             aria-label="Open notification"
             className="absolute inset-0 z-0"
             onClick={onClick}
             title="Open notification"
             data-testid="openNotification"
-          />
+            tabIndex={0}
+            onKeyDown={(e) => {
+              if (e.code === 'Space') {
```
DragosIorgulescu, line 81: should we add this to the elements enum? https://github.com/dailydotdev/apps/blob/7c546fbf1006a430b9c54586ffd1094a30a2d486/packages/shared/src/lib/element.ts#L3-L5

DragosIorgulescu (commented):
just 1 small question, otherwise looks good 🙏

packages/shared/src/components/notifications/NotificationItem.tsx
nensidosari: Yeah sure, done :D

sshanzel (approved):
Looks so great without the disables 😍 

Thank you for the fix and the references on how to work it around! 🙇 🙏

DragosIorgulescu (approved):
💯

PR #2306, opened 2023-10-25
fix: Changelog popup additional tests

nensidosari:
## Changes
- added test that confirms call of toggleUpdate in ChangelogTooltip from useChangelog
- added test that confirms call of toggleUpdate in useChangelog from useVotePost

- Few more tests to confirm the right functions are called for upvote functionality on changelog popup

## Events

### **Please make sure existing components are not breaking/affected by this PR** ✅ 

WT-1796 #done


packages/shared/src/components/tooltips/ChangelogTooltip.spec.tsx
```diff
@@ -364,4 +365,31 @@ describe('ChangelogTooltip component', () => {
...
+  it('calls toggleUpvote on upvote button click', async () => {
+    const client = new QueryClient();
+    const toggleUpvote = jest.fn();
+
+    jest.spyOn(hooks, 'useChangelog').mockImplementation(() => ({
+      toggleUpvote,
+      isAvailable: true,
+      latestPost: defaultPost,
+      dismiss: jest.fn(),
+    }));
```
rebelchris, line 378: Do you need this part, I'm not 100% sure, but since you mock the function above it feels you could omit it.

nensidosari: The function above is just a mock not connected to the useChangelog toggleUpvote, line 374 mocks the real toggleUpvote to be our mock toggleUpvote. As for the rest of the props TS complains if those are not added.

rebelchris (approved):
Nice work

PR #2307, opened 2023-10-25
WT-955: remove AuthTriggersOrString, replaced with AuthTriggers

nensidosari:
## Changes
- Created AuthTriggersType composed of AuthTriggers and some Origins used as AuthTriggers
- Removed usage of AuthTriggersOrString and replaced with AuthTriggersType
- Removed any usage of plain strings in triggers and added those to AuthTriggers Enum
- Removed any usage of plain strings in origins and added those to Origins Enum
- Removed LoginTrigger as it's quite similar to AuthTrigger and added to AuthTriggers - please if this should not be done and there are reasons for keeping these separate let me know 🙏 
- Changed showLogin function signature, it now accepts an object as a param rather then 2 separate parameters
- changed all usages of showLogin function based on the above refactor

- triggers in showLogin no longer accept plain strings, only the ones defined in AuthTriggersType
- showLogin function now only accepts an object as a parameter with triggers and options. 

## Events

### Did you test the modified components media queries? No
- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-955 #done

### DIFF
--- a/packages/shared/src/components/Feed.spec.tsx
+++ b/packages/shared/src/components/Feed.spec.tsx
@@ -336,7 +336,7 @@ it('should open login modal on anonymous upvote', async () => {
   );
   const [el] = await screen.findAllByLabelText('Upvote');
   el.click();
-  expect(showLogin).toBeCalledWith('upvote');
+  expect(showLogin).toBeCalledWith({ trigger: 'upvote' });
 });
 
 it('should send add bookmark mutation', async () => {
@@ -420,7 +420,9 @@ it('should open login modal on anonymous bookmark', async () => {
   menuBtn.click();
   const el = await screen.findByText('Save to bookmarks');
   el.click();
-  await waitFor(() => expect(showLogin).toBeCalledWith('bookmark'));
+  await waitFor(() =>
+    expect(showLogin).toBeCalledWith({ trigger: 'bookmark' }),
+  );
 });
 
 it('should increase reading rank progress', async () => {
--- a/packages/shared/src/components/Feed.spec.tsx
+++ b/packages/shared/src/components/Feed.spec.tsx
@@ -57,6 +57,7 @@ import OnboardingContext from '../contexts/OnboardingContext';
 import { LazyModalElement } from './modals/LazyModalElement';
 import { feature } from '../lib/featureManagement';
 import { SearchExperiment } from '../lib/featureValues';
+import { AuthTriggers } from '../lib/auth';
 
 const showLogin = jest.fn();
 let nextCallback: (value: PostsEngaged) => unknown = null;
@@ -336,7 +337,7 @@ it('should open login modal on anonymous upvote', async () => {
   );
   const [el] = await screen.findAllByLabelText('Upvote');
   el.click();
-  expect(showLogin).toBeCalledWith({ trigger: 'upvote' });
+  expect(showLogin).toBeCalledWith({ trigger: AuthTriggers.Upvote });
 });
 
 it('should send add bookmark mutation', async () => {
@@ -421,7 +422,7 @@ it('should open login modal on anonymous bookmark', async () => {
   const el = await screen.findByText('Save to bookmarks');
   el.click();
   await waitFor(() =>
-    expect(showLogin).toBeCalledWith({ trigger: 'bookmark' }),
+    expect(showLogin).toBeCalledWith({ trigger: AuthTriggers.Bookmark }),
   );
 });
 
--- a/packages/shared/src/components/PostOptionsMenu.tsx
+++ b/packages/shared/src/components/PostOptionsMenu.tsx
@@ -36,7 +36,6 @@ import { useLazyModal } from '../hooks/useLazyModal';
 import { LazyModal } from './modals/common/types';
 import { labels } from '../lib';
 import { MenuItemProps } from './fields/PortalMenu';
-import { AuthTriggers } from '../lib/auth';
 
 const PortalMenu = dynamic(
   () => import(/* webpackChunkName: "portalMenu" */ './fields/PortalMenu'),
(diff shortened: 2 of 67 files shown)


packages/shared/src/components/PostOptionsMenu.tsx
```diff
@@ -87,7 +88,7 @@ export default function PostOptionsMenu({
     onBlockTags,
     onUnblockTags,
   } = useTagAndSource({
-    origin: Origin.PostContextMenu,
+    origin: AuthTriggers.PostContextMenu,
```
rebelchris, line 91: I would avoid changing these actually if we can, we should opt for keeping those seperate.

packages/shared/src/components/filters/BlockedFilter.tsx
```diff
@@ -5,13 +5,14 @@ import TagItemList from './TagItemList';
...
 import BlockIcon from '../icons/Block';
+import { AuthTriggers } from '../../lib/auth';
 
 export default function BlockedFilter({
   onUnblockItem,
 }: FilterMenuProps): ReactElement {
   const { feedSettings, isLoading } = useFeedSettings();
   const { onUnblockTags, onFollowSource } = useTagAndSource({
-    origin: 'blocked filter',
+    origin: AuthTriggers.BlockedFilter,
```
rebelchris, line 15: Same here, I'll come back to a better use

packages/shared/src/components/filters/TagsFilter.tsx
```diff
@@ -55,7 +56,7 @@ export default function TagsFilter({
     useTagContext();
   const { onFollowTags, onUnfollowTags, onBlockTags, onUnblockTags } =
     useTagAndSource({
-      origin: query?.length ? Origin.TagsSearch : Origin.TagsFilter,
+      origin: query?.length ? AuthTriggers.TagsSearch : AuthTriggers.TagsFilter,
```
rebelchris, line 59: Same here

packages/shared/src/components/onboarding/FilterOnboarding.tsx
```diff
@@ -18,7 +18,7 @@ export function FilterOnboarding({
   const [selectedTopics, setSelectedTopics] = useState({});
   const { tagsCategories } = useFeedSettings();
   const { onFollowTags, onUnfollowTags } = useTagAndSource({
-    origin: Origin.TagsFilter,
+    origin: AuthTriggers.TagsFilter,
```
rebelchris, line 21: And here

packages/shared/src/hooks/post/useBlockPostPanel.ts
```diff
@@ -68,7 +69,7 @@ export const useBlockPostPanel = (
   const { displayToast } = useToastNotification();
   const { onBlockTags, onUnfollowSource, onUnblockTags, onFollowSource } =
     useTagAndSource({
-      origin: Origin.TagsFilter,
+      origin: AuthTriggers.TagsFilter,
```
rebelchris, line 72: Same here

packages/shared/src/hooks/useTagAndSource.ts
```diff
@@ -22,7 +23,7 @@ export interface SourceActionArguments {
 }
 
 interface UseTagAndSourceProps {
-  origin?: string;
+  origin?: AuthTriggers;
```
rebelchris, line 26: This should stay Origin ideally.
If it's used as a Trigger we can say `Origin as Trigger` in the implementation needed

rebelchris (changes requested):
Minor remark on keeping origins as `Origin` as they ideally are typed conform our analytics system.

packages/shared/src/components/search/SearchBarSuggestionList.tsx
```diff
@@ -31,7 +31,7 @@ export function SearchBarSuggestionList({
     return (
       <PlaceholderSearchSuggestion
         className={className}
-        onClick={() => showLogin(LoginTrigger.SearchSuggestion)}
```
nensidosari, line 34: @rebelchris just checking with you since you added similar comment for Origin, should this (and the other usages) be left as LoginTrigger?

rebelchris: This should be fine as it's a LoginTrigger, just don't want to mix Origin and LoginTrigger for different purposes

nensidosari: Cool then it's ok to leave these now as AuthTriggers instead of LoginTriggers.

packages/shared/src/components/Feed.spec.tsx
```diff
@@ -336,7 +336,7 @@ it('should open login modal on anonymous upvote', async () => {
   );
   const [el] = await screen.findAllByLabelText('Upvote');
   el.click();
-  expect(showLogin).toBeCalledWith('upvote');
+  expect(showLogin).toBeCalledWith({ trigger: 'upvote' });
```
rebelchris, line 339: Not urgent, but we can simply use the `AuthTriggers` here right?

rebelchris (approved):
Nice looks great to me.
Minor suggestion for tests, but not urgent also happy with this

packages/shared/src/components/Feed.spec.tsx
nensidosari: Done! 

Related but not specific to this: 
In this one I'm always a bit confused with myself, sometimes I read it's better to actually use plain strings in tests then the enums you use in prod, just to make sure that if you change something in enums unintentionally, the tests will catch your error. 
Anyways this is something that we can check as a separate thing on the standards that we use for tests.

PR #2309, opened 2023-10-25
fix: disable search suggestions query when search is not enabled

DragosIorgulescu:
## Changes

- due to the usage of `useSearchSuggestions` in the main `FeedContainer`, this query was triggered regardless of wether the feature was enabled or not.
- adding a `disabled` argument which disables the query if `true`

## Events

N / A

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1786 #done

### DIFF
--- a/packages/shared/src/hooks/search/useSearchSuggestions.ts
+++ b/packages/shared/src/hooks/search/useSearchSuggestions.ts
@@ -8,17 +8,21 @@ import { disabledRefetch } from '../../lib/func';
 
 type UseSearchSuggestions = (data: {
   origin: SearchBarSuggestionListProps['origin'];
+  disabled?: boolean;
 }) => Pick<
   SearchBarSuggestionListProps,
   'origin' | 'suggestions' | 'isLoading'
 >;
 
-export const useSearchSuggestions: UseSearchSuggestions = (args) => {
+export const useSearchSuggestions: UseSearchSuggestions = ({
+  disabled,
+  ...args
+}) => {
   const { user } = useAuthContext();
   const { data, isLoading } = useQuery(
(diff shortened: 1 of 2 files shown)


packages/shared/src/hooks/search/useSearchSuggestions.ts
```diff
@@ -18,7 +19,7 @@ export const useSearchSuggestions: UseSearchSuggestions = (args) => {
   const { data, isLoading } = useQuery(
     generateQueryKey(RequestKey.SearchHistory, user),
     getSearchSuggestions,
-    { ...disabledRefetch, enabled: !!user },
+    { ...disabledRefetch, enabled: !args.disabled && !!user },
```
rebelchris, line 22: Since it's optional we should say `!args?.disabled` right?

DragosIorgulescu: I think `args.disabled` is optional, but `args` itself is not? https://github.com/dailydotdev/apps/blob/b94b97299a225854f6f11608cdb01da72f9c9412/packages/shared/src/hooks/search/useSearchSuggestions.ts#L9-L15

rebelchris (approved):
This will work, not props gets passed down as `args` again, not sure you need the disabled there?
If not it's fine like this.

DragosIorgulescu:
> This will work, not props gets passed down as `args` again, not sure you need the disabled there? If not it's fine like this.

@rebelchris good catch, I extracted it out just before you added the comment 😅

sshanzel (approved):
Nice! Thanks for the optimization fix! :shipit:

PR #2318, opened 2023-10-29
refactor: single useBookmarkPost

omBratteng:
- Refactors `useBookmarkPost` hook to work in feed and single post 

todo:

- [x] make it work on history page
- [x] "You might like" in modal
- [ ] Update tests

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1884 #done


packages/shared/src/components/post/PostActions.tsx
```diff
@@ -56,6 +58,21 @@ export function PostActions({
       : undefined,
   });
 
+  const { toggleBookmark } = useBookmarkPost({
```
rebelchris, line 61: We just need to make double sure wherever this previous `onBookmark` came from didn't have specific tracking we need to integrate here.

packages/shared/src/components/PostOptionsMenu.tsx
```diff
@@ -80,6 +89,7 @@ export default function PostOptionsMenu({
   const { trackEvent } = useContext(AnalyticsContext);
   const { hidePost, unhidePost } = useReportPost();
   const { openModal } = useLazyModal();
+  const { items } = useContext(ActiveFeedContext);
```
rebelchris, line 92: Since we introduced this here, perhaps we should just abstract the `feedQueryKey` here as well.
Don't really see any point in prop drilling it now.

packages/shared/src/components/post/PostHeaderActions.tsx
```diff
@@ -136,9 +137,9 @@ export function PostHeaderActions({
         </SimpleTooltip>
       )}
       <PostOptionsMenu
-        onBookmark={onBookmark}
         onShare={onShare}
         post={post}
+        feedQueryKey={feedQueryKey}
```
rebelchris, line 142: This could basically happen inside the component, as you already access the same context there.

packages/shared/src/hooks/useBookmarkPost.ts
```diff
@@ -1,82 +1,220 @@
...
+};
+
+export type UseBookmarkPostProps = {
+  onMutate?: (
+    props: UseBookmarkPostMutationProps,
+  ) => UseBookmarkPostRollback | undefined;
+  variables?: unknown;
+};
+
+const prepareVotePostAnalyticsOptions = ({
```
rebelchris, line 51:
```suggestion
const prepareBookmarkPostAnalyticsOptions = ({
```

packages/shared/src/hooks/useBookmarkPost.ts
```diff
@@ -1,82 +1,220 @@
...
-    (() => void) | undefined
-  >(
-    ({ id }) =>
-      request(graphqlUrl, ADD_BOOKMARKS_MUTATION, {
-        data: { postIds: [id] },
+  const { user, showLogin } = useContext(AuthContext);
+  const { trackEvent } = useContext(AnalyticsContext);
+
+  const defaultOnMutate = ({ id }) => {
+    updatePostCache(client, id, (post) => ({ bookmarked: !post.bookmarked }));
```
rebelchris, line 80: Don't you already send the swapped value in some places? (just checking)

packages/shared/src/hooks/useBookmarkPost.ts
```diff
@@ -1,82 +1,220 @@
...
+      if (!post) {
+        return;
+      }
+
+      if (!user) {
+        showLogin({ trigger: AuthTriggers.Bookmark });
+        return;
+      }
+
+      const analyticsOptions = prepareVotePostAnalyticsOptions({
```
rebelchris, line 134: Same naming change

rebelchris (changes requested):
Wow very neat cleanup!
Could you link it up to a preview deployment that's open (2 or 3 I think) so we can have a quick play around as well?

omBratteng:

packages/shared/src/hooks/useBookmarkPost.ts
omBratteng: This is for the default mutation, i.e. when there is no feedQueryKey, so it falls back to the mutation for a single page

packages/shared/src/hooks/useBookmarkPost.ts
omBratteng: Renamed

packages/shared/src/components/post/PostHeaderActions.tsx
omBratteng: Moved it into the `PostOptionsMenu`

packages/shared/src/components/PostOptionsMenu.tsx
capJavert: `ActiveFeedContext` already has `queryKey` which is current key for active feed.

omBratteng: It was changed to use they `queryKey` in [4464bda](https://github.com/dailydotdev/apps/pull/2318/commits/4464bdafbd5fe0368fe65303b9b8ff3199a711e0#diff-0eb690e81644b4b6b5fae075b498cec565153a663e9838c2e241f88e5e77f9aaR90)

packages/shared/src/hooks/useBookmarkPost.ts
```diff
@@ -1,82 +1,218 @@
...
+
+  if (postIndexToUpdate === -1) {
+    return undefined;
+  }
+
+  const mutationHandler = (post: Post) => ({ bookmarked: !post.bookmarked });
+  const previousState = (items[postIndexToUpdate] as PostItem)?.post
+    ?.bookmarked;
+
+  optimisticPostUpdateInFeed(
```
capJavert, line 195: I think it might be neater to sync the state with `useMutationSubscription` like `useFeedVotePost` so we don't have to bundle mutation of single and feed into single logic.

packages/shared/src/hooks/useBookmarkPost.ts
rebelchris: Ah got it! Thanks for the explanation

packages/shared/src/hooks/useBookmarkPost.ts
```diff
@@ -1,82 +1,218 @@
...
+
+export type BookmarkProps = {
+  id: string;
+};
+
+export type UseBookmarkPostProps = {
+  onMutate?: (
+    props: UseBookmarkPostMutationProps,
+  ) => Promise<UseBookmarkPostRollback> | UseBookmarkPostRollback | undefined;
+  variables?: unknown;
```
rebelchris, line 48: Don't we know what types we expect here?

rebelchris (approved):
Neat!
Ran a e2e test on preview and overall works really neat and synced.
One small things (which might be a existing issue, so let me know if it can be fixed easily or not)
When in bookmark feed, unbookmarking doesn't remove it here. (this might even be a product decision at the time though)

omBratteng:
Noticed that as well, and that is already an existing "issue", I'll take a  quick look at it and see if there are any quick fixes for it.

packages/shared/src/hooks/useBookmarkPost.ts
omBratteng: Once we get a lean feed, we should be able to slim this down to just use one mutation using a single cache key for both feed and single post.

capJavert: Lean feed is not doing that currently. It is basically doing something similar to this by updating all the caches (at least currently). It is gonna take another iteration I think revamp the caching structure. @rebelchris you know more on how it is going.

rebelchris: Correct first iteration will not fix that.
I think we could use the useMutation thing for now, it should handle logic a bit better right (more cleaner approach)

packages/shared/src/hooks/useBookmarkPost.ts
omBratteng: You're right. Going to adjust it a bit and throw in the type `MutationKey`

omBratteng: Refactored it to have a type now

packages/shared/src/components/post/PostActions.tsx
omBratteng: Went over the analytics that was sent before, and I can't find any difference from then to now.

PR #2323, opened 2023-11-01
fix: added failsafes to the animate functions based on evaluations

rebelchris:
- Added failsafe evaluations to the animate functions to prevent execution if refs don't exist/render.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-809 #done


packages/shared/src/components/RankProgress.tsx
```diff
@@ -162,74 +162,80 @@ export function RankProgress({
     const firstAnimationDuration = 400;
     const maxScale = 1.666;
 
-    const progressAnimation = showRadialProgress
-      ? progressRef.current.animate(
```
nensidosari, line 166: Probably the optional chaining would work well here too

packages/shared/src/components/RankProgress.tsx
```diff
@@ -162,74 +162,80 @@ export function RankProgress({
...
-                  transform: 'scale(1.5)',
-                  opacity: 0,
+                  transform: 'scale(1)',
+                  opacity: 1,
+                  '--stop-color1': rankToGradientStopBottom(rank),
+                  '--stop-color2': rankToGradientStopTop(rank),
                 },
               ],
-              { duration: 600, fill: 'forwards' },
+              { duration: 100, fill: 'forwards' },
```
nensidosari, line 236: Is this change intentional?

rebelchris: Oh no! not even sure how that happened 🤯

rebelchris: Ah It's actually not a change, it's just GitHub showing it on the wrong line.
This is the lastBadge which always had 100, it's showing it compared to AttentionRef though.

packages/shared/src/components/RankProgress.tsx
rebelchris: Although would probably work, I think this is safer down the line to exclude any weird things.
Let me know what you think?

nensidosari: Well that is quite useful in this cases at least when you're just checking down the chain if any of the parts is null/undefined and can make code clearer in this case. I try to use when I can.
But don't have any 100% strong opinion of using either way if we don't have any standard set.
In this case for example with it would allow you to add an extra check for progressRef as well `progressRef?.current?.animate()`

packages/shared/src/components/RankProgress.tsx
nensidosari: Ah yeah, now I see. Sometimes GitHub messes with the changes like this 🤦‍♀️ Nevermind then. :D

nensidosari (approved):
LGTM :) minor suggestion only

packages/shared/src/components/RankProgress.tsx
rebelchris: Alright, moving to optional chaining :) 
We don't need the `?.current` thought as it will exist as we use refs.

PR #2325, opened 2023-11-02
feat: referral join page

nensidosari:
## Changes
- added new key `generic` in join page for generic referral program
- created new page based on the existing AISearchInvite currently using join page
- added required images in image lib
- changed render of LazyModalElement component from MainLayout to _app.tsx in order to make it available in join page

NOTE: will add comments in the code to make sure the naming and logic are correct where I am not 100% sure.

- Creates new join page with slightly different design
- adds cta button to download extension
- adds video image that can open a youtube video when clicked

Screenshots below of all different breakpoints:
Screenshot 2023-11-02 at 2 46 26 PM
Screenshot 2023-11-02 at 2 45 50 PM
Screenshot 2023-11-02 at 2 45 35 PM
Screenshot 2023-11-02 at 2 45 18 PM

## Events

Did you introduce any new tracking events? Yes
- [x] Don't forget to update the [Analytics Taxonomy sheet](https://docs.google.com/spreadsheets/d/18Lv7zXges9QfVX5VYL1a-Hyl0e1sQ3sLr0OK8YZWKXI/edit#gid=0)

|New | accept invitation  | extra: { } |

- [x] MobileL (420px)
- [x] Tablet (656px)
- [x] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1926 #done


packages/shared/src/lib/analytics.ts
```diff
@@ -114,6 +114,9 @@ export enum AnalyticsEvent {
   DeclineContentScripts = 'decline content scripts',
   ToggleFeedPreview = 'toggle feed preview',
   CreateFeed = 'create feed',
+  // generic referral - start
+  AcceptGeneralReferralInvitation = 'accept generic referral invitation',
```
nensidosari, line 118: Still not confirmed. Conversation in thread: https://dailydotdev.slack.com/archives/C05P9ET7S9K/p1698942032420289

packages/shared/src/lib/featureManagement.ts
```diff
@@ -27,6 +28,7 @@ const feature = {
     OnboardingFilteringTitle.Control,
   ),
   lowImps: new Feature('feed_low_imps'),
+  genericReferral: new Feature('generic_referral', GenericReferral.Control),
```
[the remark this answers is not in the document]
capJavert: You need to make sure it is setup in Growthbook (our features provider). Chris can setup that if he did not already.

packages/shared/src/styles/components/buttons.css
```diff
@@ -39,6 +39,12 @@
     border-radius: 0.875rem;
   }
 
+  &.xlarge {
```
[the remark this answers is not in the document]
capJavert: Might be a design inconsistency.

packages/webapp/components/invite/Referral.tsx
```diff
@@ -0,0 +1,221 @@
...
+
+  useEffect(() => {
+    if (genericReferral === GenericReferral.Control) {
+      return;
+    }
+
+    router.push(redirectTo);
+    // router is an unstable dependency
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [redirectTo, genericReferral]);
```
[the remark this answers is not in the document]
capJavert: Hooks sounds good, we just have to make sure future join pages reuse it correctly and we don't end up with hook that does too much.

capJavert (commented):
Love to see the sense of ownership, you practically already commented/explained anything that would be pointed out in the PR review. Left the comments there. Great job 💪

packages/shared/src/hooks/referral/useReferralConfig.ts
```diff
@@ -17,19 +17,29 @@ export const useReferralConfig = ({
...
-      title: defaultTitle,
-      description: defaultDescription,
-      images: defaultImages,
-      redirectTo: '/my-feed',
-    };
-  }
+  switch (campaign as ReferralCampaignKey) {
+    case ReferralCampaignKey.Search:
+      return {
+        title: defaultTitle,
```
rebelchris, line 23: Maybe we can abstract the default to a const, and use {...default, overwrite: something } instead?
It will avoid having the same lines 3 times.

packages/shared/src/lib/analytics.ts
rebelchris: The taxonomy is different, also replied on the thread.

packages/shared/src/lib/auth.ts
```diff
@@ -54,6 +54,7 @@ export enum AuthTriggers {
   SearchInput = 'search input',
   SearchSuggestion = 'search suggestion',
   LoginPage = 'login page',
+  GenericReferral = 'generic referral',
```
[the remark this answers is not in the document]
rebelchris: Think this one is fine, it's more a extra param. maybe just note it to analytics as a comment there is a new authTrigger?
(We don't currently have a good flow, but maybe good to add)

packages/shared/src/lib/featureManagement.ts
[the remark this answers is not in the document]
rebelchris: I'm [checking](https://dailydotdev.slack.com/archives/C05P9ET7S9K/p1698997550791609) with product, actually not sure now.
Although the DR doesn't mention it.

packages/shared/src/styles/components/buttons.css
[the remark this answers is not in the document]
rebelchris: Yeah ideally we shouldn't add button version if they didn't exist before.
Probably a missed design addition.

This is the add to browser one?
I think it's indeed a out of guideline button (As the bigger icon is also not ideal)

packages/webapp/components/invite/Referral.tsx
```diff
@@ -0,0 +1,221 @@
...
+import { useAuthContext } from '@dailydotdev/shared/src/contexts/AuthContext';
+import { AuthTriggers } from '@dailydotdev/shared/src/lib/auth';
+import { useMutation } from 'react-query';
+import { acceptFeatureInvitation } from '@dailydotdev/shared/src/graphql/features';
+import { useRouter } from 'next/router';
+import Logo, { LogoPosition } from '@dailydotdev/shared/src/components/Logo';
+import { useFeature } from '@dailydotdev/shared/src/components/GrowthBookProvider';
+import { feature } from '@dailydotdev/shared/src/lib/featureManagement';
+import { GenericReferral } from '@dailydotdev/shared/src/lib/featureValues';
+import { useActions } from '@dailydotdev/shared/src/hooks/useActions';
```
rebelchris, line 16: Not urgent, but see you import a couple hook, best to use the index exports then. (general this is a preferred behaviour, but we only started doing it recently)

packages/webapp/components/invite/Referral.tsx
```diff
@@ -0,0 +1,221 @@
...
+    // we don't need to send the params here explicitly
+    trackEvent({ event_name: AnalyticsEvent.AcceptGeneralReferralInvitation });
+
+    return handleAccept();
+  };
+
+  useEffect(() => {
+    if (genericReferral === GenericReferral.Control) {
+      return;
+    }
```
rebelchris, line 104: This will depend on if it's run as experiment or not.

packages/webapp/components/invite/Referral.tsx
```diff
@@ -0,0 +1,221 @@
...
+
+  return (
+    <div
+      style={{
+        background: `url(${cloudinary.referralCampaign.genericReferral.backgroundDark})`,
+        backgroundSize: 'cover',
+      }}
+      className="flex overflow-hidden relative flex-col laptop:flex-row flex-auto items-center p-2 laptop:p-6 h-[100vh]"
+    >
+      <span className="absolute top-8 left-1/2 laptop:left-12 -translate-x-1/2 laptop:translate-x-0 scale-[1.375]">
```
rebelchris, line 119: Why the scale? did we not have the icon already in the right size?
It should not differ from the other invite pages I don't think.

packages/webapp/components/invite/Referral.tsx
```diff
@@ -0,0 +1,221 @@
...
+          <p className="my-auto text-center laptop:text-left typo-headline laptop:typo-title3 laptopL:typo-title2">
+            <span className="block laptop:inline">{referringUser.name}</span>
+            <span className="font-normal"> invited you to daily.dev</span>
+          </p>
+        </span>
+        <h1 className="w-full text-center laptop:text-left break-words-overflow typo-large-title tablet:typo-mega2 laptopL:typo-giga2">
+          The Homepage
+          <span className="block font-bold text-theme-color-cabbage">
+            {' '}
+            Developers Deserve
```
rebelchris, line 137: Not your fault, but think the [capitalization](https://dailydotdev.slack.com/archives/C05P9ET7S9K/p1698997971107649) here is wrong.

packages/webapp/components/invite/Referral.tsx
```diff
@@ -0,0 +1,221 @@
...
+          </span>
+        </h1>
+        <p className="mx-auto laptop:mx-0 mt-7 tablet:mt-8 max-w-sm laptopL:max-w-xl text-center laptop:text-left typo-title3 text-theme-label-tertiary">
+          Get a feed of the best developer news out there! Read more quality
+          articles.
```
rebelchris, line 142: Why is this on a new line? (maybe it's just the editor)

packages/webapp/components/invite/Referral.tsx
```diff
@@ -0,0 +1,221 @@
...
+          <span className="tablet:block laptopL:inline"> Stay up to date.</span>
```
rebelchris, line 143: Extra space in front of Stay

packages/webapp/components/invite/Referral.tsx
```diff
@@ -0,0 +1,221 @@
...
+        </p>
+        <Button
+          buttonSize={isMobile ? ButtonSize.Large : ButtonSize.XLarge}
+          className="p-4 mx-auto laptop:mx-0 mt-6 tablet:mt-12 max-w-[17.5rem] mobileL:max-w-[24.25rem] btn-primary"
+          tag="a"
+          href={downloadBrowserExtension}
+          onClick={() => {
+            handleAcceptClick();
+          }}
+          target="_blank"
```
rebelchris, line 153: When using external links please always add:
```
rel={anchorDefaultRel}
```

packages/webapp/components/invite/Referral.tsx
```diff
@@ -0,0 +1,221 @@
...
+          loading={isLoading}
+          disabled={isLoading || isSuccess}
+        >
+          <FlexCentered className="gap-2 mobileL:typo-title3">
+            <BrowsersIcon
+              width={isMobile ? '58px' : '80px'}
+              height={isMobile ? '30px' : '40px'}
```
rebelchris, line 160: We don't use pixels, and icons have sizes, it should reflect a icon size.

packages/webapp/components/invite/Referral.tsx
```diff
@@ -0,0 +1,221 @@
...
+        </Button>
+      </div>
+      <div className="flex fixed flex-auto w-full h-full laptop:initial laptop:bottom-[unset] bottom-[-49%]">
+        <div className="flex relative laptop:absolute justify-center self-center m-auto mx-2 z-10 laptopL:right-[-6rem] laptop:right-[-7.5rem] w-[fit-content] laptop:w-[initial]">
+          <img
+            src={
+              cloudinary.referralCampaign.genericReferral.purpleEdgeGlowTablet
+            }
+            alt="Purple glow right edge"
+            className="fixed laptop:top-0 w-full laptop:w-auto laptop:h-full bottom-[-1rem] laptop:bottom-[unset] laptop:rotate-[270deg] z-20"
```
rebelchris, line 174: Just to check is `z-20` supported?

packages/webapp/components/invite/Referral.tsx
```diff
@@ -0,0 +1,221 @@
...
+            onClick={() => {
+              setVideoOpen(true);
+            }}
+          />
+        </div>
+      </div>
+      {isVideoOpen && (
+        <Modal
+          // eslint-disable-next-line @dailydotdev/daily-dev-eslint-rules/no-custom-color
+          className="px-8 bg-[black]"
```
rebelchris, line 199: This should just be `bg-black` pretty sure we have that :)

packages/webapp/components/invite/Referral.tsx
```diff
@@ -0,0 +1,221 @@
...
+          kind={ModalKind.FlexibleCenter}
+          size={ModalSize.XLarge}
+          isOpen={isVideoOpen}
+          onRequestClose={onVideoModalClose}
+        >
+          <CloseButton
+            buttonSize={ButtonSize.Small}
+            className="top-3 right-3 text-white border-white !absolute !btn-secondary"
```
rebelchris, line 207: Buttons have a `position` attribute.
Also we force white on all states? (Think yes)

rebelchris (changes requested):
Nice!
Got some minor comments on styling/guidelines.

packages/webapp/components/invite/Referral.tsx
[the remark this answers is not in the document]
rebelchris: Hook sounds like a plan indeed

packages/shared/src/lib/auth.ts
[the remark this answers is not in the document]
nensidosari: When you say analytics you mean in that excel sheet?

packages/shared/src/styles/components/buttons.css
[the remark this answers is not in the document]
nensidosari: Yeah, will check this with design then

packages/webapp/components/invite/Referral.tsx
nensidosari: > Why the scale? did we not have the icon already in the right size?

I could not find any other one except of what was already in the join page for AISearchInvite

> It should not differ from the other invite pages I don't think.

Did doubt that as well, though the name in the DR referral program & Ux Improvements made me think maybe this could be any of the changes/improvements. 

Checking with design now in [thread](https://dailydotdev.slack.com/archives/C05P9ET7S9K/p1699012788634189?thread_ts=1699011522.350609&cid=C05P9ET7S9K), definitely best to not add custom stuff.

packages/webapp/components/invite/Referral.tsx
nensidosari: In the designs in tablet and mobile this was on a new line, based on the text size and all it wouldn't stay on a new line unless specifically arranged to stay so with the <span>... Should I check with design for this as well?

packages/webapp/components/invite/Referral.tsx
nensidosari: Yup it is

packages/webapp/components/invite/Referral.tsx
nensidosari: Took this piece of code from here https://github.com/dailydotdev/apps/blob/3489d4348c42710c91d50c7257d0351a02366f1f/packages/webapp/pages/join.tsx, probably old so I'll fix this :D

nensidosari: > Also we force white on all states? 

Which white are you referring to, text-white?

packages/shared/src/lib/auth.ts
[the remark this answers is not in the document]
rebelchris: Yes :)

packages/webapp/components/invite/Referral.tsx
rebelchris: I mean the `articles.` shows on a new line, why not behind the `quality`

packages/webapp/components/invite/Referral.tsx
rebelchris: Yeah and the border, it's probably fine, but those are non theme mode supportive (light/dark) just something to keep in mind.

packages/webapp/components/invite/Referral.tsx
nensidosari: Ah, it's the linter

packages/shared/src/styles/components/buttons.css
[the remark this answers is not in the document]
nensidosari: Removed this addition, at the moment have overridden large button padding and height  only for this button as discussed & agreed with design. Have left a comment in the code to point to this discussion.

packages/webapp/components/invite/Referral.tsx
nensidosari: Removed the scale, the icon is same across the app

packages/webapp/components/invite/Referral.tsx
nensidosari: The sizes of the icon component we have don't match up with this icon, probably because this button is in the landing pages we don't have this in the proper sizes here. 
Since we're adding overrides for the button to be a different size on what we have on the guidelines because we want to make it similar to the one in the landing page, maybe we are ok to keep these in pixels until it is decided this button will be properly introduced here? 🤔

packages/webapp/components/invite/Referral.tsx
nensidosari: So I took this code from the previous video implementation for this page referred from the DR in the video section. If we don't have specific designs on how this button should look like, I believe we can remove those `text-white` and `border-white` as they don't ruin anything, just the X will be in a lighter grey color if not hovered, and there will be no white border.

packages/webapp/components/invite/Referral.tsx
sshanzel: We must at least convert the values to `rem`. Seems like a one-off too, so I am cool if we will make it hard-coded. Though we should probably have it defined in the `className`.

packages/webapp/components/invite/Referral.tsx
```diff
@@ -0,0 +1,227 @@
...
+            <BrowsersIcon
+              width={isMobile ? '58px' : '80px'}
+              height={isMobile ? '30px' : '40px'}
+              className="text-theme-label-primary"
+            />
+            Try it now - It&apos;s free
+          </FlexCentered>
+        </Button>
+      </div>
+      <div className="flex fixed flex-auto w-full h-full laptop:initial laptop:bottom-[unset] bottom-[-49%]">
```
sshanzel, line 172: Not sure if the 1% would make a difference, but using `1/2` should result to the percentage value.
```suggestion
      <div className="flex fixed flex-auto w-full h-full laptop:initial laptop:bottom-[unset] bottom-1/2">
```

packages/webapp/components/invite/Referral.tsx
```diff
@@ -0,0 +1,227 @@
...
+        <div className="flex relative laptop:absolute justify-center self-center m-auto mx-2 z-10 laptopL:right-[-6rem] laptop:right-[-7.5rem] w-[fit-content] laptop:w-[initial]">
```
sshanzel, line 173: Kindly check if the `-right-30` exists. If not, continue with the arbitrary value.
```suggestion
        <div className="flex relative laptop:absolute justify-center self-center m-auto mx-2 z-10 laptopL:-right-24 laptop:-right-30 w-fit laptop:w-[initial]">
```

packages/webapp/components/invite/Referral.tsx
sshanzel: Why the value has to be 20? We should use a more accurate value as this produces unwanted behavior in certain cases. Such as on mobile when sidebar is opened which will have lesser value than this if I remember the value correctly.

packages/webapp/components/invite/Referral.tsx
```diff
@@ -0,0 +1,227 @@
...
+          <img
+            src={
+              cloudinary.referralCampaign.genericReferral.purpleEdgeGlowTablet
+            }
+            alt="Purple glow right edge"
+            className="fixed laptop:top-0 w-full laptop:w-auto laptop:h-full bottom-[-1rem] laptop:bottom-[unset] laptop:rotate-[270deg] z-20"
```
sshanzel, line 179:
```suggestion
            className="fixed laptop:top-0 w-full laptop:w-auto laptop:h-full -bottom-4 laptop:bottom-[unset] laptop:rotate-270 z-20"
```

sshanzel (changes requested):
Looks great overall! ✨
Well done on articulating the concerns properly 🙌 

Nothing major though I notice that we use a lot of arbitrary values, we should always try to keep using the pre-defined ones, and only when the values are not available, we switch to arbitrary ones.

packages/webapp/components/invite/Referral.tsx
nensidosari: Seems fine at 50% as well, so replacing :D

packages/webapp/components/invite/Referral.tsx
nensidosari: right-30 doesn't exist, so keeping that value. Changing others :)

packages/webapp/components/invite/Referral.tsx
nensidosari: rotate-270 so will be keeping rotate-[270deg]

packages/webapp/components/invite/Referral.tsx
```diff
@@ -0,0 +1,221 @@
...
+      return showLogin({
+        trigger: AuthTriggers.GenericReferral,
+        options: {
+          onLoginSuccess: handleAccept,
+          onRegistrationSuccess: handleAccept,
+        },
+      });
+    }
+
+    // since in the page view, query params are tracked automatically,
```
rebelchris, line 96: You double checked it on the actual event triggered? Assume so, but wanted to note it just to be safe.

packages/webapp/components/invite/Referral.tsx
```diff
@@ -0,0 +1,221 @@
...
+            aria-label="Play daily.dev introduction video"
+            onClick={() => {
+              setVideoOpen(true);
+            }}
+          />
+        </div>
+      </div>
+      {isVideoOpen && (
+        <Modal
+          // eslint-disable-next-line @dailydotdev/daily-dev-eslint-rules/no-custom-color
```
rebelchris, line 197: You no longer need this eslint above right?

packages/webapp/components/invite/Referral.tsx
```diff
@@ -0,0 +1,221 @@
...
+          />
+          <img
+            src={cloudinary.referralCampaign.genericReferral.appScreenshot}
+            alt="Daily.dev app screenshot of my feed page"
+            className="object-contain w-full laptop:w-auto h-auto laptop:h-full z-10 laptopXL:max-h-[39.5rem] laptop:max-h-[25.35rem]"
+          />
+          <img
+            src={cloudinary.referralCampaign.genericReferral.playButton}
+            alt="Play daily.dev introduction video"
+            className="absolute laptop:self-center laptopL:m-auto laptopXL:h-auto laptop:mr-[8.5rem] laptop:mt-[0.5rem] z-20 laptop:h-[8rem] tablet:h-[14.5rem] h-[7.8rem] tablet:mt-[-1.5rem]"
```
rebelchris, line 183: Not really breaking, but see a lot of custom sizes in this PR, probably all edge-cases but we really need to be warry of them and try to avoid them as much as possible where we can.
Just a note, no action needed.

rebelchris (changes requested):
Thanks for fixing all the hooks.
Some very minor semantics from my side, almost good to go.

packages/webapp/components/invite/Referral.tsx
sshanzel: I see, maybe we can use `-rotate-90` as it would produce the same result.

packages/shared/src/lib/featureManagement.ts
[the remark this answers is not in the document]
nensidosari: Based on thread decision, removing this line and the need for experiment :)

packages/webapp/components/invite/Referral.tsx
nensidosari: Removed useEffect

packages/shared/src/hooks/useSidebarRendered.ts
```diff
@@ -1,5 +1,5 @@
 import { useMemo } from 'react';
-import useMedia from './useMedia';
+import { useMedia } from './useMedia';
```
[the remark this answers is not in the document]
nensidosari: But this usage is within the shared package, do we really want import the package export from the package itself?

packages/webapp/components/invite/Referral.tsx
nensidosari: Ah with bg-black it was still needed, but replaced that now with pepper-90 and all good, can remove this.

packages/shared/src/hooks/useSidebarRendered.ts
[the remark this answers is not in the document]
rebelchris: Ah no I see now :). never mind.

packages/webapp/components/invite/Referral.tsx
nensidosari: Removed these in the end, just in case we want the border, what would be a white color that would be mode supportive?

packages/webapp/components/invite/Referral.tsx
```diff
@@ -0,0 +1,227 @@
...
+    <div
+      style={{
+        background: `url(${cloudinary.referralCampaign.genericReferral.backgroundDark})`,
+        backgroundSize: 'cover',
+      }}
+      className="flex overflow-hidden relative flex-col laptop:flex-row flex-auto items-center p-2 laptop:p-6 h-[100vh]"
+    >
+      <span className="absolute top-8 left-1/2 laptop:left-12 -translate-x-1/2 laptop:translate-x-0">
+        <Logo showGreeting={false} position={LogoPosition.Relative} />
+      </span>
```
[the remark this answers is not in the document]
nensidosari: I'm just wondering is this worth to become a component at the moment? it's quite small and Logo is already a component, so the only "big" thing we would be avoiding with this would be those classnames on the span. 
Anyways can change to a component if you think is better

packages/webapp/components/invite/Referral.tsx
```diff
@@ -0,0 +1,227 @@
...
+          />
+          <iframe
+            className="w-full border-none aspect-video"
+            src="https://www.youtube.com/embed/igZCEr3HwCg"
+            title="YouTube video player"
+            allow="encrypted-media;web-share"
+            allowFullScreen
+          />
+        </Modal>
+      )}
```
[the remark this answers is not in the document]
nensidosari: I have applied the necessary changes, but the modal still is not showing, probably I've missed something, so it would be great if we can take a look together, thanks! 🙏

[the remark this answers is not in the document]
nensidosari: Sorted, pushed the changes :D

packages/shared/src/components/MainLayout.tsx
```diff
@@ -202,7 +201,6 @@ function MainLayout({
       {customBanner}
       {isBannerAvailable && <PromotionalBanner />}
       <InAppNotificationElement />
-      <LazyModalElement />
```
sshanzel, line 205: Can you share what was the motivation for moving this element?

packages/shared/src/components/modals/VideoModal.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+export default function VideoModal({
+  onRequestClose,
+  src,
+  title,
+  ...props
+}: VideoModalProps): ReactElement {
+  return (
+    <Modal
+      className="px-8 pepper-90"
```
sshanzel, line 22: We don't have the popup on the Sketch understandably so as we are just reusing an old implementation, our modals have their own background color already, would it be not sufficient that we are applying a dark background on light mode as well?

packages/shared/src/components/modals/common.tsx
```diff
@@ -85,6 +85,10 @@ const SearchReferralModal = dynamic(
     ),
 );
 
+const VideoModal = dynamic(
```
sshanzel, line 88: I think we need to be more explicit with the filename and its component name, we can probably host another video in a modal, and this would make it confusing.

packages/webapp/components/invite/Referral.tsx
```diff
@@ -0,0 +1,181 @@
...
+        </Button>
+      </div>
+      <div className="flex fixed -bottom-1/2 flex-auto w-full h-full laptop:initial laptop:bottom-[unset]">
+        <div className="flex relative laptop:absolute laptopL:-right-24 justify-center self-center m-auto mx-2 z-10 laptop:right-[-7.5rem] w-fit laptop:w-[initial]">
+          <img
+            src={
+              cloudinary.referralCampaign.genericReferral.purpleEdgeGlowTablet
+            }
+            alt="Purple glow right edge"
+            className="fixed laptop:top-0 -bottom-4 w-full laptop:w-auto laptop:h-full laptop:bottom-[unset] laptop:rotate-[270deg] z-20"
```
sshanzel, line 151: I think we can use the negative value.
```suggestion
            className="fixed laptop:top-0 -bottom-4 w-full laptop:w-auto laptop:h-full laptop:bottom-[unset] laptop:-rotate-90 z-20"
```

sshanzel, line 151: Also, let's probably use a lesser value for the z-index as we don't want to introduce more wild numbers for it in the codebase. Not sure what would fit, but probably 1, 2, or 3 should be sufficient enough. For context, we are trying to cleanup our z-index rules as we've had many issues with it in the past.

packages/webapp/components/invite/Referral.tsx
```diff
@@ -0,0 +1,181 @@
...
+          />
+          <img
+            src={cloudinary.referralCampaign.genericReferral.appScreenshot}
+            alt="Daily.dev app screenshot of my feed page"
+            className="object-contain w-full laptop:w-auto h-auto laptop:h-full z-10 laptopXL:max-h-[39.5rem] laptop:max-h-[25.35rem]"
```
sshanzel, line 156: Same goes for the z-index here.

packages/webapp/components/invite/Referral.tsx
```diff
@@ -0,0 +1,181 @@
...
+          />
+          <img
+            src={cloudinary.referralCampaign.genericReferral.playButton}
+            alt="Play daily.dev introduction video"
+            className="absolute laptop:self-center laptopL:m-auto laptopXL:h-auto laptop:mr-[8.5rem] laptop:mt-[0.5rem] z-20 laptop:h-[8rem] tablet:h-[14.5rem] h-[7.8rem] tablet:mt-[-1.5rem]"
```
sshanzel, line 161: Let's keep using the pre-defined values.
```suggestion
            className="absolute laptop:self-center laptopL:m-auto laptopXL:h-auto laptop:mr-[8.5rem] laptop:mt-2 z-20 laptop:h-32 tablet:h-[14.5rem] h-[7.5rem] tablet:-mt-6"
```

packages/webapp/components/invite/Referral.tsx
```diff
@@ -0,0 +1,181 @@
...
+          onClick={() => {
+            handleAcceptClick();
+          }}
+          target="_blank"
+          rel={anchorDefaultRel}
+          loading={isLoading}
+          disabled={isLoading || isSuccess}
+        >
+          <FlexCentered className="gap-2 mobileL:typo-title3">
+            <BrowsersIcon className="text-theme-label-primary w-[3.7rem] h-[1.9rem] mobileL:w-[5rem] mobileL:h-[2.5rem]" />
```
sshanzel, line 139: The size is 40x80. We can use `w-20 h-10`. Not sure why GH is not allowing me to provide a suggestion :sob:

sshanzel (changes requested):
We're almost there! Just a few more arbitrary values to change 🚢

packages/shared/src/components/modals/common.tsx
nensidosari: I think that was the main idea why I made this modal not feature specific, but more generic with the main thing being hosting a video, with the idea that we might want to use it to host another video... 🤔 Do we not want this?

rebelchris: I agree like the generic approach here :)

packages/shared/src/components/modals/VideoModal.tsx
nensidosari: It does look exactly the same removing the pepper-90 there, so removing it in order for it to be defined by the modal itself :)

packages/webapp/components/invite/Referral.tsx
nensidosari: True, thank u!!

packages/shared/src/components/MainLayout.tsx
nensidosari: When creating the separate modal component for the video we show on referral page and trying to use with the LazyModal setup, realised that the modal didn't work. This because join page is not wrapped by the MainLayout component. 
In order for every page to be able to render the modal using LazyModal setup this needed to be higher up in the hierarchy so got moved to _app.tsx where it will work for every page, no matter if they use main layout or not.

I'll add this in the PR description as well when I finish all the changes :D

packages/webapp/components/invite/Referral.tsx
nensidosari: Changing these to use values z-0/1/2

packages/webapp/components/invite/Referral.tsx
nensidosari: Actually, have been checking this a couple of times trying to understand when it should fire, and apparently it doesn't fire unless the user is logged in 🤔 
This was the same logic on the AISearchInvite, so it isn't firing there as well when it is supposed to.

`if (!user) {
      return showLogin({
        trigger: AuthTriggers.SearchReferral,
        options: {
          onLoginSuccess: handleAccept,
          onRegistrationSuccess: handleAccept,
        },
      });
    }`

Because of the above code it never goes to trackEvent unless there is a user, I believe that shouldn't be like this right?

packages/shared/src/components/modals/common.tsx
sshanzel: Oh, then in that case, feel free to move forward if this was built to be generic 🚀

packages/shared/src/components/MainLayout.tsx
sshanzel: I see. Thank you for the clarification! That indeed would not work inside just the MainLayout. Sorry I reviewed way earlier, just a bit excited about this 😆

packages/webapp/components/invite/Referral.tsx
sshanzel: The login flow has its own tracking, part of the reason we don't want to send another event but only identify the trigger by passing the parameter `AuthTriggers.SearchReferral`.

packages/webapp/components/invite/Referral.tsx
rebelchris: Technically then it should just be `btn-secondary` not sure we need the overwrite.
Also if both modes have white it's fine to use this.
Otherwise, think primary is the most white one. *(salt.90) i guess?

packages/webapp/components/invite/Referral.tsx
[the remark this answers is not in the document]
rebelchris: Ideally to me this whole page is a re-used template/layout as it has many common elements.
But agree with Lee, even though it's a simple wrapper would be great to have it as a common component.

Maybe you can even get away with simply a `classed` component?

rebelchris (approved):
Looks pretty solid to me, just couple remarks:
- As this will become the default join page, do we want to try it out on preview first to see if nothing breaks?
- As we moved lazyModalElement, we really need to test out webapp/extension + companion before merging, I don't want it to effect users if something goes wrong there.
- The tracking events I'd like to see tested on set preview deployment to ensure they work.

packages/webapp/components/invite/Referral.tsx
nensidosari: Cool, got it. Quick question though, if the user is logged in, so already has an account, do we know what should happen if user is invited to join and lands on the referral page?
In the code at the moment I believe that we still show the user the page, and if the user clicks the button in this case we would send the `accept invitation` event, but would this make sense if the user already has an account?
I might have missed something, but checking just in case

packages/webapp/components/invite/Referral.tsx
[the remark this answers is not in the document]
nensidosari: I thought of creating a hook, thought the actions were not making much sense in there, was not able to find a right name for the hook to wrap up these actions, so came up with the idea of moving these higher up in the code, in join page.
You can find the refactor in https://github.com/dailydotdev/apps/pull/2337

packages/webapp/components/invite/Referral.tsx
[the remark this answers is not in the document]
nensidosari: Ah, just pushed this in the common file at the moment, I believe that should be ok. But curiosity, @rebelchris why did you specifically suggested a `classed` component?

nensidosari:
> Looks pretty solid to me, just couple remarks:
> 
> * As this will become the default join page, do we want to try it out on preview first to see if nothing breaks?
> * As we moved lazyModalElement, we really need to test out webapp/extension + companion before merging, I don't want it to effect users if something goes wrong there.
> * The tracking events I'd like to see tested on set preview deployment to ensure they work.

@rebelchris agree, I would want to properly test this as well

packages/webapp/components/invite/Referral.tsx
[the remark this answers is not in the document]
sshanzel: The reason I had it in the AISearchInvite is, there could be differences in the implementation, such as this referral is not available to logged in users only to anonymous ones, on the other hand, the search referral is available to logged in users, hence, we redirect at some point, etc. I highly suggest we make a hook that can configure that part. Probably name it as `useReferralPage`?

packages/webapp/components/invite/Referral.tsx
sshanzel: Sorry, I just got back from the gym, answered in Slack now 🥳

packages/webapp/components/invite/Referral.tsx
```diff
@@ -0,0 +1,178 @@
...
+        </h1>
+        <p className="mx-auto laptop:mx-0 mt-7 tablet:mt-8 max-w-sm laptopL:max-w-xl text-center laptop:text-left typo-title3 text-theme-label-tertiary">
+          Get a feed of the best developer news out there! Read more quality
+          articles.
+          <span className="tablet:block laptopL:inline">Stay up to date.</span>
+        </p>
+        <Button
+          buttonSize={ButtonSize.Large}
+          // important has been used is some classnames to override the default styles as agreed on [thread](https://dailydotdev.slack.com/archives/C05P9ET7S9K/p1699023366663489?thread_ts=1699011522.350609&cid=C05P9ET7S9K)
+          className="p-4 mx-auto laptop:mx-0 mt-6 tablet:mt-12 max-w-[17.5rem] mobileL:max-w-[fit-content] btn-primary mobileL:!h-[4rem] mobileL:!p-[0] mobileL:!px-[1.5rem]"
```
sshanzel, line 124: The `w-fit` should now be supported OOTB since v3 🥳 
```suggestion
          className="p-4 mx-auto laptop:mx-0 mt-6 tablet:mt-12 max-w-[17.5rem] mobileL:max-w-fit btn-primary mobileL:!h-16 mobileL:!p-0 mobileL:!px-6"
```

packages/webapp/components/invite/Referral.tsx
```diff
@@ -0,0 +1,178 @@
...
+          onClick={() => {
+            handleAcceptClick();
+          }}
+          target="_blank"
+          rel={anchorDefaultRel}
+          loading={isLoading}
+          disabled={isLoading || isSuccess}
+        >
+          <FlexCentered className="gap-2 mobileL:typo-title3">
+            <BrowsersIcon className="mobileL:w-20 mobileL:h-10 text-theme-label-primary w-[3.7rem] h-[1.9rem]" />
```
sshanzel, line 136: From the Sketch, it looks like the size change starts at tablet mode.
```suggestion
            <BrowsersIcon className="tablet:w-20 tablet:h-10 text-theme-label-primary w-[3.7rem] h-[1.9rem]" />
```

packages/webapp/components/invite/Referral.tsx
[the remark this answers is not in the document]
nensidosari: In that case I don't think this is necessary anymore, at least with the campaigns we already have they are quite different to each other. I believe it's best to leave as it is and refactor this in the future when any other campaigns are introduced that are similar. 
I'll close the refactor PR, will just get the campaign configs from there.

packages/webapp/components/invite/Referral.tsx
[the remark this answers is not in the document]
rebelchris: Because it's a dumb component without logic, but anything will work.

packages/shared/src/graphql/actions.ts
```diff
@@ -16,6 +16,7 @@ export enum ActionType {
   ExistingAnonymousBanner = 'existingAnonymousBanner',
   AcceptedSearch = 'accepted_search',
   UsedSearch = 'used_search',
+  AcceptedGenericReferral = 'accepted_generic_referral',
```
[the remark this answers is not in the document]
nensidosari: removed since the feature is no longer needed

packages/webapp/pages/join.tsx
```diff
@@ -31,6 +33,10 @@ const Page = ({
     referringUser,
   });
 
+  useEffect(() => {
+    document.body.classList.add('hidden-scrollbar');
```
sshanzel, line 37: Just for safety, we should remove this class on unmount.

sshanzel (approved):
Splendid! Well done on cleaning up some hooks too.

Had a non-blocking one, so approving the PR 🥳

packages/webapp/pages/join.tsx
rebelchris: Yep agreed, let's add the removal as well.
Tested and seems to be fine for this page also on mobile.

rebelchris (approved):

PR #2332, opened 2023-11-06
chore: manifest migration to V3

DragosIorgulescu:
## Changes

- updated `manifest.json` to match the new V3 API - [DR spec](https://dailydotdev.atlassian.net/wiki/spaces/HAN/pages/618987566/WT-1764+Manifest+V3#Updating-Manifest)
- migrate to service workers - [DR spec](https://dailydotdev.atlassian.net/wiki/spaces/HAN/pages/618987566/WT-1764+Manifest+V3#Service-Worker-Migration)
  - works out of the box with the current implementation
- `browserAction` deprecation in favor of `action` - [DR spec](https://dailydotdev.atlassian.net/wiki/spaces/HAN/pages/618987566/WT-1764+Manifest+V3#Action-API)
- validate other API changes - [DR spec](https://dailydotdev.atlassian.net/wiki/spaces/HAN/pages/618987566/WT-1764+Manifest+V3#Deprecated-APIs)
  - no changes needed, we are not using any of these
- remove `webextension-polyfill-ts` and use the main `webextension-polyfill` lib w/ included types
- extension security checklist: [DR spec](https://dailydotdev.atlassian.net/wiki/spaces/HAN/pages/618987566/WT-1764+Manifest+V3#Extension-Security)
  - remotely hosted code:  n/a
  - remote executions: n/a
  - content security policy: we are no longer pulling `script-src 'self' https://www.google-analytics.com` so this is no longer an issue
  - cross cookies in service worker: working as expected
 

## Events

- [x] validate events are firing 

- ~~Have you done sanity checks in the webapp?~~

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1907 #done


packages/extension/src/manifest.json
```diff
@@ -10,30 +10,38 @@
...
-  "__dev__permissions": ["storage", "http://localhost/", "http://*.local.com/"],
-  "optional_permissions": ["topSites", "*://*/*"],
-  "content_security_policy": "script-src 'self' https://www.google-analytics.com; object-src 'self'",
-  "__firefox__content_security_policy": "object-src 'self'",
-  "browser_action": {
+  "__dev__permissions": ["storage"],
+  "__dev__host_permissions": ["http://localhost/", "http://*.local.com/"],
+  "optional_permissions": ["topSites"],
+  "optional_host_permissions": [ "*://*/*"],
+  "content_security_policy": {
```
idoshamun, line 23: I see we removed the specific definition for firefox, is it on purpose?

idoshamun (commented):
Can't believe v3 is coming together finally! Thanks a lot

rebelchris (commented):
Oh wow, expected we really needed more changes in the background worker to be honest 🤯 
Let's move to testing phase.

I do also support Ido's question, was that a intentional removal?

packages/extension/src/background/index.ts
```diff
@@ -1,5 +1,5 @@
 import 'content-scripts-register-polyfill';
-import { browser, Runtime, Tabs } from 'webextension-polyfill-ts';
+import browser, { runtime, Runtime, Tabs } from 'webextension-polyfill';
```
capJavert, line 2: Just use `browser.runtime`? It will be easier to refactor down the line if we change the polyfill since `browser` is usually available as a global in extension context.

packages/shared/src/components/tooltips/ChangelogTooltip.tsx
```diff
@@ -56,8 +56,8 @@ function ChangelogTooltip({
           );
         }
 
-        const browser = await import('webextension-polyfill-ts').then(
-          (mod) => mod.browser,
+        const browser = await import('webextension-polyfill').then(
```
capJavert, line 59: Nice, and now we have type support for it 👌

capJavert (commented):
Nice work, at this point I have no major comments, really smooth implementation and I like the fact that aside from mocks you did not have to update any tests and are all 🟢

packages/extension/src/manifest.json
DragosIorgulescu: good catch, since we removed google analytics, there was no meaningful diff between the regular content security policy and the firefox specific one. Was there a particular reason `script-src 'self'` was not initially included there?

We can see during testing in firefox if this causes any problems or triggers a permissions reset.

packages/extension/src/background/index.ts
capJavert: I see you use `browser` top level in all other places so this is the only different case to adjust.

packages/extension/src/manifest.json
idoshamun: Fair enough, I'm not sure why we added it in the first place but we can leave it for testing

packages/extension/src/background/index.ts
DragosIorgulescu: good catch, updated 🙏

packages/extension/src/manifest.json
DragosIorgulescu: having this in also ensures protection agains cross site scripting and other things, so it's a good default to have
 
BTW, I found the answer from daildev search here very nice: https://app.daily.dev/search?id=f69e7f9a-7e23-11ee-9cee-42010a007e0c

packages/extension/src/background/index.ts
```diff
@@ -12,6 +11,8 @@ import { BOOT_LOCAL_KEY } from '@dailydotdev/shared/src/contexts/common';
 import { ExtensionMessageType } from '@dailydotdev/shared/src/lib/extension';
 import { getContentScriptPermissionAndRegister } from '../lib/extensionScripts';
 
+const client = new GraphQLClient(graphqlUrl, { fetch: globalThis.fetch });
```
sshanzel, line 14: Very neat solution. Have you checked whether we can completely avoid having to go through the process below for regular GraphQL queries? Meaning, without using the proxy for the `request` method and the queries would still work?

DragosIorgulescu: @sshanzel good point, this is only required in the background service worker due to its restricted access to various APIs (such as the `XMLHttpRequest` object). Tested it as well, of course, this works as expected outside of the background worker 🙏

packages/extension/webpack.config.js
```diff
@@ -77,9 +52,7 @@ module.exports = {
   resolve: {
     extensions: ['.svg', '.ts', '.tsx', '.js', '.json'],
     alias: {
-      'webextension-polyfill-ts': path.resolve(
-        path.join(__dirname, 'node_modules', 'webextension-polyfill-ts'),
-      ),
+      fetch: false, // or point to a file that exports the global fetch
```
capJavert, line 55: what is this for, was this culprit for xmlhttprequest usage in compiled bundle?

packages/extension/webpack.config.js
```diff
@@ -191,8 +164,6 @@ module.exports = {
     new CopyWebpackPlugin({
       patterns: [{ from: 'public', to: '.' }],
     }),
-    // plugin to enable browser reloading in development mode
-    extensionReloaderPlugin,
```
capJavert, line 195: We should open the backlog task to fix reloader down the line.

packages/extension/src/lib/extensionScripts.ts
```diff
@@ -1,29 +1,37 @@
...
+let hasInjectedScripts = false;
+
+export const registerBrowserContentScripts = async (): Promise<void> => {
+  if (hasInjectedScripts) {
+    return null;
+  }
+
+  await browser.scripting.registerContentScripts([
+    {
+      id: `companion-${Date.now()}`,
```
capJavert, line 18: Is there a better way to do this? Worried we will register a bunch of same content scripts and load them to user browser. I suspect this might be due to content script being registered in each tab/context our extension and companion are loaded.

packages/extension/webpack.config.js
```diff
@@ -6,7 +6,6 @@ const TerserPlugin = require('terser-webpack-plugin');
 const CopyWebpackPlugin = require('copy-webpack-plugin');
 const HtmlWebpackPlugin = require('html-webpack-plugin');
 const { CleanWebpackPlugin } = require('clean-webpack-plugin');
-const ExtensionReloader = require('webpack-ext-reloader');
```
capJavert, line 9: we can also remove from package.json?

capJavert (changes requested):
Almost there, some notes from latest changes.

packages/extension/src/lib/extensionScripts.ts
capJavert: This looks to me as something we need to solve before production.

DragosIorgulescu: good point @capJavert I'll do some digging into this one

packages/extension/src/background/index.ts
sshanzel: Nice, thanks for the verification. The reason I was curious is, that this will remove the overhead of using the `useBackgroundRequest` which complicates the process when making a query/mutation. This means, the requests would automatically work without using those hooks. I understand this is only to migrate the Manifest, so I think we should create a ticket for it to migrate the implementation. Would also make new implementations work automatically as opposed to introducing the said hook every time we do something. cc @rebelchris

packages/extension/src/lib/extensionScripts.ts
capJavert: @DragosIorgulescu so I was just checking something by removing `Date.now()` above. To my surprise I could not get the error we had before. Maybe it was fixed in other way? So yeah, maybe just remove it then?

packages/storybook/stories/Button.stories.tsx
```diff
@@ -1,25 +1,12 @@
-import type { Meta, StoryObj } from '@storybook/react';
-import classNames from 'classnames';
-
-import {
-  Button,
-  ButtonSize,
-} from '@dailydotdev/shared/src/components/buttons/Button';
-import ShareIcon from '@dailydotdev/shared/src/components/icons/Share';
+import type { Meta, StoryObj } from "@storybook/react";
```
rebelchris, line 1: Why are these changes needed?

packages/extension/src/lib/extensionScripts.ts
idoshamun: Agree this cannot remain as is before publishing. It's too risky

packages/storybook/stories/Button.stories.tsx
DragosIorgulescu: good point, not sure, maybe a bad merge conflict. Removed 👍

packages/extension/src/lib/extensionScripts.ts
DragosIorgulescu: this should be properly handled now 👍

packages/extension/webpack.config.js
DragosIorgulescu: ticket here: https://dailydotdev.atlassian.net/browse/MI-107

packages/extension/src/background/index.ts
sshanzel: @DragosIorgulescu we should maybe add a ticket for this. To make implementations less complicated.

packages/extension/src/manifest.json
```diff
@@ -10,30 +11,56 @@
...
     "https://dailynow.co/",
     "https://*.dailynow.co/"
   ],
-  "__dev__permissions": ["storage", "http://localhost/", "http://*.local.com/"],
-  "optional_permissions": ["topSites", "*://*/*"],
-  "content_security_policy": "script-src 'self' https://www.google-analytics.com; object-src 'self'",
-  "__firefox__content_security_policy": "object-src 'self'",
-  "browser_action": {
+  "__firefox|prod__permissions": [
+    "scripting",
```
rebelchris, line 22: This is a additional request right?
Why do we need it for firefox (and chrome for that matter)

packages/shared/src/components/ExtensionPermissionsPrompt.tsx
```diff
@@ -0,0 +1,61 @@
+import React, { ReactElement, useEffect } from 'react';
+import Logo, { LogoPosition } from './Logo';
+import { OnboardingTitleGradient } from './onboarding/common';
+import { Button, ButtonSize } from './buttons/Button';
```
rebelchris, line 4: This should be buttonV2

packages/shared/src/components/ExtensionPermissionsPrompt.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+  const handleRequestHostPermissions = async () => {
+    const success = await requestHostPermissions({ origins });
+    if (success) {
+      refetchHostPermissions();
+    }
+  };
+
+  return (
+    <div className="flex max-h-[100vh] min-h-[100vh] flex-col items-center justify-center overflow-hidden px-7 text-center antialiased">
```
rebelchris, line 30: Why don't we use `max-h-screen` here?

packages/shared/src/components/post/PostComments.tsx
```diff
@@ -58,7 +59,7 @@ export function PostComments({
         requestMethod(
           graphqlUrl,
           POST_COMMENTS_QUERY,
-          { postId: id, [initialDataKey]: comments, first: 500 },
+          { postId: id, [initialDataKey]: comments, first: 50 },
```
rebelchris, line 62: Not against it, but why did we change this here?

packages/shared/src/hooks/useHostPermissionStatus.ts
```diff
@@ -0,0 +1,35 @@
+import { useQuery } from '@tanstack/react-query';
+import { disabledRefetch } from '../lib/func';
+import { useExtensionContext } from '../contexts/ExtensionContext';
+
+export const hostKey = ['host_key'];
```
rebelchris, line 5: Why don't we use `generateQueryKey` here?

packages/shared/src/lib/constants.ts
```diff
@@ -59,3 +59,8 @@ export const authUrl =
 export const heimdallUrl = isDevelopment
   ? process.env.NEXT_PUBLIC_HEIMDALL_URL || 'http://127.0.0.1:3000'
   : authUrl;
+
+export const extensionHostPermissions = [
+  'https://daily.dev/',
+  'https://*.daily.dev/',
+];
```
rebelchris, line 66: Unused?

rebelchris (changes requested):
Some minor comments :)

packages/shared/src/components/post/PostComments.tsx
sshanzel: Our post comments are not yet paginated, this would not show comments beyond 50.

packages/extension/webpack.config.js
DragosIorgulescu: I see no harm in keeping it since we are planning on adding it back pretty soon after releasing this anyway

packages/extension/src/manifest.json
DragosIorgulescu: This is needed so we can make the request to register the companion content script: https://github.com/dailydotdev/apps/pull/2332/files/d9b5d2c91912ca948c32ea17e0a4532d336e3fce#diff-9d3bc2384cfa13933a9b00959c4da1058f54cc6338b656556dab709bc868ba5eR42

this has changed from `browser.contentScripts.register` which was used before and is no longer supported in V3.

Without this, we get the following error:
```
TypeError: D().scripting is undefined
```

We could use different APIs, but if this works on V2 as well, not sure if it's worth it to have an `if/else` for this

capJavert: It does not trigger any permission popups?

packages/extension/webpack.config.js
capJavert: Fine either way, add a TODO maybe or note so we don't forget what it is.

packages/extension/src/manifest.json
DragosIorgulescu: > It does not trigger any permission popups?

No new permission pop-ups are generated by this. We can validate this again through a test release of the firefox extension

packages/extension/webpack.config.js
DragosIorgulescu: I added a ticket for this here: https://dailydotdev.atlassian.net/browse/MI-107

packages/extension/src/manifest.json
capJavert: Fine by me.

packages/extension/src/manifest.json
```diff
@@ -10,30 +11,56 @@
     "48": "icons/48.png",
     "128": "icons/128.png"
   },
-  "__prod__permissions": [
+  "__prod__permissions": ["scripting"],
```
idoshamun, line 14: Is this needed?

packages/extension/src/lib/extensionScripts.ts
```diff
@@ -1,29 +1,63 @@
...
+export const registerBrowserContentScripts = async (): Promise<void> => {
+  if (hasInjectedScripts) {
+    return null;
+  }
+
+  const registeredScripts = await browser.scripting.getRegisteredContentScripts(
+    {
+      ids: [companionScriptId],
+    },
+  );
```
capJavert, line 35: I think this can just always call `browser.scripting.getRegisteredContentScripts` because if for any reason user disallows content script from chrome menu `hasInjectedScripts` will be outdated. I don't think actual call to `getRegisteredContentScripts` will be much of a overhead because it is a Promise to local browser APIs. 

So basically I would remove `hasInjectedScripts` optimization.

capJavert (approved):
I don't see anything blocking anymore, I would only simplify loading of content script even more to avoid edge cases. Approving either way.

packages/extension/src/lib/extensionScripts.ts
capJavert: You can remove `hasInjectedScripts` completely now right, meaning declaration and all assignments?

DragosIorgulescu: Since we are still using V2 API on FF via `browser.contentScripts.register`, we still need it for now

capJavert: You are right, tnx for clarification 🚀

packages/shared/src/hooks/useExtensionPermission.ts
```diff
@@ -1,49 +0,0 @@
-import { QueryClient, useQueryClient } from '@tanstack/react-query';
```
sshanzel, line 1: Is this file deleted already? Looks like the content was just removed but not the file entirely.

sshanzel (approved):
Nothing else to add from my end. Definitely a great task. This was no easy task and now we are ready for release 🚀

packages/shared/src/hooks/useExtensionPermission.ts
DragosIorgulescu: great catch, not sure how this happened, maybe some merge conflict

rebelchris (approved):
Nothing left from my side code wise.

PR #2334, opened 2023-11-07
feat: profile menu

sshanzel:
## Changes
- Fix the InteractivePopup placement for top and left values.
  - When the sidebar is not expanded, as part of our user settings, the left placement should follow to close up the same distance it had when it was opened.
- Introduce a `LabeledImage` component for having a label/title in the center of a faded image.
- Update the `ProfileMenu` to utilize the interactive popup for better UI support.
- Add the `color` property to the transition when hovering a button.

Note: I would love to share some screenshots but at the time of writing, I am still fixing my local env. As soon as it gets back up, I will add some.

Preview:

Screenshot 2023-11-07 at 8 54 35 PM
Screenshot 2023-11-07 at 8 54 11 PM
Screenshot 2023-11-07 at 8 54 05 PM

Screenshot 2023-11-07 at 8 55 17 PM
Screenshot 2023-11-07 at 8 55 12 PM
Screenshot 2023-11-07 at 8 55 06 PM

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1929 #done


packages/shared/src/styles/components/buttons.css
```diff
@@ -1,6 +1,6 @@
 .btn {
   transition: 0.2s ease-in-out;
-  transition-property: border, background, box-shadow;
+  transition-property: border, background, box-shadow, color;
```
sshanzel, line 3: I noticed when hovering the items in the profile menu, the switch on the font colors is quite snappy so I applied the same transition to the color as well.

packages/shared/src/styles/base.css
```diff
@@ -102,6 +102,8 @@ body {
 
   --theme-overlay-cabbage-opacity: theme('colors.cabbage.90')A3;
   --theme-overlay-onion-opacity: theme('colors.onion.90')A3;
+  --theme-fade-from: theme('colors.pepper.90');
```
sshanzel, line 105: Although this value is widely available, it needs to be consistent regardless of theme. Hence, the value is the same on both themes.

rebelchris (changes requested):
Nice work, mainly have some questions around uniforming the opening of the menu and changes inside the interactive popup.

Depending on your comment I have some follow up questions probably.

packages/shared/src/components/ProfileMenu.tsx
```diff
@@ -1,29 +1,35 @@
...
-const PortalMenu = dynamic(
-  () => import(/* webpackChunkName: "portalMenu" */ './fields/PortalMenu'),
-  {
-    ssr: false,
-  },
-);
+interface ListItem {
+  title: string;
+  buttonProps?: ButtonProps<AllowedTags>;
+}
```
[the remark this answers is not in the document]
sshanzel: Yeah, but that one did not involved our button component, so we can't pass the props we need.

packages/shared/src/components/ProfileMenu.tsx
```diff
@@ -35,85 +41,95 @@ export default function ProfileMenu(): ReactElement {
     !!referralToken &&
     campaignCtaPlacement === CampaignCtaPlacement.ProfileMenu;
 
+  const items: ListItem[] = useMemo(() => {
+    const handleReferralClick = () => {
```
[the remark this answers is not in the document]
sshanzel: Yes, it is not used anywhere, exactly the reason was to avoid adding a `useCallback` usage if we will keep it within the memo.

packages/shared/src/components/ProfileMenu.tsx
```diff
@@ -35,85 +41,95 @@ export default function ProfileMenu(): ReactElement {
...
-    <PortalMenu
-      disableBoundariesCheck
-      id="profile-context"
-      className="menu-primary"
-      animation="fade"
+    <InteractivePopup
+      onClose={onClose}
+      closeOutsideClick
+      position={InteractivePopupPosition.RightStart}
+      className="w-full border !right-4 laptop:max-w-[13.75rem] max-w-[21.25rem] roundeed-14 border-theme-divider-tertiary"
```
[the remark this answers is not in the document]
sshanzel: Oh, great catch on the typo. The thing is, it is too close to the edge, unlike our other implementation like Squad Checklist.

packages/shared/src/components/ProfileMenu.tsx
```diff
@@ -35,85 +41,95 @@ export default function ProfileMenu(): ReactElement {
...
-      </Item>
-      {showSearchReferral && (
-        <Item>
-          <button
-            type="button"
-            className="flex items-center min-w-[12.5rem]"
-            onClick={handleReferralClick}
+      <LabeledImage
+        src={user.image}
+        alt="Logged-in user"
```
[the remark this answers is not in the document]
sshanzel: Our existing implementations, iirc, contain the word image/avatar/photo, which linting does not accept, we only get away with it because we pass it through a reusable component which lint won't know we will use the alt for the `img` element. Also, here, we don't need to know which specific user is logged in. Wdyt?

packages/shared/src/components/tooltips/InteractivePopup.tsx
```diff
@@ -42,29 +44,81 @@ const positionClass: Record<InteractivePopupPosition, string> = {
...
+  }, [closeOutsideClick]);
 
   return (
     <Portal>
-      <div
-        className={classNames(
-          'fixed z-popup bg-theme-bg-primary rounded-16 overflow-hidden',
-          className,
-          classes,
+      <ConditionalWrapper
```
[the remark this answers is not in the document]
sshanzel: It is actually more complicated on Squad Checklist, when we switch to mobile, it executes openModal, and closes this. Vice-versa, the reason why I mentioned this improvement in the DR. This is rendering the same thing as opposed to two different components.

packages/shared/src/components/tooltips/InteractivePopup.tsx
```diff
@@ -42,29 +44,81 @@ const positionClass: Record<InteractivePopupPosition, string> = {
...
+  const container = useRef<HTMLDivElement>();
+  const onCloseRef = useRef(onClose);
+  const { sidebarRendered } = useSidebarRendered();
+  const { sidebarExpanded } = useSettingsContext();
+  const finalPosition = sidebarRendered
+    ? position
+    : InteractivePopupPosition.Center;
+  const classes = positionClass[finalPosition];
+
+  useEffect(() => {
```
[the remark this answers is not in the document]
sshanzel: Unfortunately, no. It only handles where the element is going to be injected. That was handled by the library.

packages/shared/src/components/ProfileMenu.tsx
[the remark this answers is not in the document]
rebelchris: Should we not uniform one of them maybe?
I don't see how the end result between a context menu for instance and this differs.

packages/shared/src/components/ProfileMenu.tsx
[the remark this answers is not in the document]
rebelchris: Alright

packages/shared/src/components/ProfileMenu.tsx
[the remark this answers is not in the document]
rebelchris: But then rather introduce a new position to be honest.
It kind of defeats the purpose of having set conditions if we override them.

packages/shared/src/components/ProfileMenu.tsx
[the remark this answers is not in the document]
rebelchris: The component is only for logged users no?
I prefer to have it more explicit as usernames photo but it's a small semantic

packages/shared/src/components/tooltips/InteractivePopup.tsx
```diff
@@ -42,29 +44,81 @@ const positionClass: Record<InteractivePopupPosition, string> = {
...
+      if (container.current && !container.current.contains(e.target as Node)) {
+        onCloseRef.current(e);
+      }
+    };
+
+    // The reason we wait for a few ms is, when the user clicks the triggering element to open this,
+    // it is considered outside the core container. Hence, triggering the close here.
+    setTimeout(() => {
+      globalThis?.document.addEventListener('click', onClickAnywhere);
+    }, WAIT_FOR_FULL_RENDER);
```
rebelchris, line 89: Can't we evaluate it by checking if container.current exist for instance?
It feels a bit like overkill.

If really no other way (which I doubt) let's use useDebounce rather.

packages/shared/src/components/sidebar/SidebarUserButton.tsx
```diff
@@ -39,15 +35,17 @@ export function SidebarUserButton({
                 <Button
```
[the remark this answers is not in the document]
sshanzel: Interesting! Let me check if they can share the same component.

packages/shared/src/components/ProfileMenu.tsx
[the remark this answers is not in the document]
sshanzel: Yeah, that makes sense. Explicit is always better.

packages/shared/src/components/ProfileMenu.tsx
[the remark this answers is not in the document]
sshanzel: I see. Yeah, that is true. Let me introduce a new one.

packages/shared/src/components/ProfileMenu.tsx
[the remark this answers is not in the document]
sshanzel: Because the context menu item has only a few defined props, this one is unique for the ProfileMenu which is not using the same base container of ContextMenu from the library. Can you share how you envision it to be? The `MenuItemProps` is kind of a bit far from what we have here:
```
export interface MenuItemProps<
  TReturn = unknown,
  TArgs extends Array<unknown> = Array<unknown>,
  TAnchorProps = AnchorHTMLAttributes<HTMLAnchorElement>,
> {
  icon: ReactElement;
  label: string;
  action?: (...args: TArgs) => TReturn;
  anchorProps?: TAnchorProps;
}
```

packages/shared/src/components/tooltips/InteractivePopup.tsx
sshanzel: Let me know if you have other alternatives to that.

With regards to using debouce, compared to now I think it became more complicated than it needs to be with all the dependency management as we have to be very strict since this will add numerous listeners if handled incorrectly. Wdyt?
Screenshot 2023-11-08 at 9 11 16 PM

packages/shared/src/components/sidebar/SidebarUserButton.tsx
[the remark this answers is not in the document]
sshanzel: What do you think of the way we merged it here? https://github.com/dailydotdev/apps/pull/2334/commits/9deb42f03160ae3f6dccee4e095cc30c4c3bd5d5

packages/shared/src/components/ProfileMenu.tsx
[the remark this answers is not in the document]
rebelchris: Hmm, let's leave it for now then, don't want to overcomplicate this PR.
Maybe something to keep in mind for future clean up if we could have 1 generic type of menu items.

packages/shared/src/components/sidebar/SidebarUserButton.tsx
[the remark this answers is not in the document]
rebelchris: Works for me :)

packages/shared/src/components/tooltips/InteractivePopup.tsx
rebelchris: To me hooks are there for exactly this reason, not re-use same logic, and handle all cleanup internally.
I think we also have a even listener hook maybe that one icw is better?

Not sure, @capJavert what do you think?

capJavert: Sorry I am not sure why the above needs a timeout in the first place, probably need some more context. 🙏 

What I do see is that with current setup we could get events leak because useEffect cleanup could fire before `setTimeout` executes in a first place. This is something we need to fix definitely. We can do that by using `clearTimeout` in cleanup function.

sshanzel: @capJavert good catch on clearing the timeout.

@rebelchris I agree we should utilize the hooks whenever possible, but I think it is better to avoid them wherever we can as it is more performant without the additional overhead behind the scenes with checking the dependencies.

rebelchris: @sshanzel That sounds very contradicting 😂
Maybe we can sync on that tomorrow quickly?

sshanzel: Let me see how we can modify the useWindowEvents to support timeouts.

sshanzel: > @sshanzel That sounds very contradicting 😂
Maybe we can sync on that tomorrow quickly?

Well, I think that is me most of the time 😂 
Sure, let's sync tomorrow.

sshanzel:
@capJavert @rebelchris have now made it to clear the timeout as well on unmount. With regards to the `useWindowEvents`, it currently doesn't support having a timeout before applying the event listeners. We can refactor it and include some mechanism, then test everything works as intended. My only concern is, that it is currently being used in a lot of Auth-related components, touching it might require testing the authentication from end-to-end as well. I would rather have a separate ticket to do the refactor including the implementation here. Wdyt?

rebelchris:
@sshanzel It still doesn't feel 100% right to me.
Perhaps we can maybe rather return the popup on desktop
And simply return a actual modal on mobile?

This way we keep using existing components, instead of mimicking them, and dealing with edge-cases.
I'm also afraid if we ever change things in our modal it wouldn't auto reflect here and bound to give headaches down the line.

For all I care it could even be a wrapped component as InteractivePopupOrModal (suck at naming)

sshanzel:
> Perhaps we can maybe rather return the popup on desktop
And simply return a actual modal on mobile?

It would not make a difference actually since the issue is in desktop. Remove the timeout, click the avatar, click anywhere, then click the avatar again.

packages/shared/src/components/tooltips/InteractivePopup.tsx
```diff
@@ -41,30 +45,83 @@ const positionClass: Record<InteractivePopupPosition, string> = {
...
+      return null;
+    }
+
+    const onClickAnywhere = (e: MouseEvent) => {
+      if (!container.current.contains(e.target as Node)) {
+        onCloseRef.current(e);
+      }
+    };
+
+    globalThis?.document.addEventListener('click', onClickAnywhere);
```
rebelchris, line 86: Cant we move this to useWindowEvents now?

packages/shared/src/hooks/utils/useInteractivePopup.ts
```diff
@@ -0,0 +1,33 @@
...
+      event.stopPropagation();
+    }
+
+    setIsOpen(value);
+  }, []);
+
+  return {
+    isOpen,
+    onUpdate,
+    wrapHandler: handleTrigger,
```
sshanzel, line 34: Example usage would be then `onClick={wrapHandler((event) => your function)}`.

packages/shared/src/components/tooltips/InteractivePopup.tsx
sshanzel: Currently, we should only enable it when the props is right, but we currently don't support it in that hook. Meaning, the listener will be added even though it is not needed.

rebelchris: Hmm should we modify it perhaps?
I really would like to reuse the hook we have to keep complex edge case friendly hooks uniform

sshanzel: I want to, but it will involve a lot of refactors on existing implementations with the type to make work with `click` as I mentioned in our discussion. That's why I was suggesting to open a ticket dedicated to it.

rebelchris: I send you a DM on slack with an example I created, that works but love to hear what you think.

sshanzel:
Ticket for cleaning up the `useWindowEvents` hook. https://dailydotdev.atlassian.net/browse/WT-1951

packages/shared/src/hooks/useWindowEvents.ts
```diff
@@ -7,13 +7,22 @@ export interface MessageEventData {
...
+interface OptionalProps {
+  validateKey: boolean;
+  enabled: boolean;
+}
+
 const useWindowEvents = <T extends MessageEventData = MessageEventData>(
   event: EventParameter,
   key: string,
   func: KeyedWindowEventHandler<T>,
-  validateKey = true,
```
rebelchris, line 14: This doesn't break any existing implementations?

rebelchris (approved):
Looks great, thanks for the refactor appreciate it.
Just tripple check that wxtra change doesnt break anything existing?
Else we rather might do that in the ticket you created

packages/shared/src/hooks/useWindowEvents.ts
sshanzel: Yes, there was no other implementation for this property it seems. Also, typescript didn't throw any issue with the type change.

kirillkurko (commented):
I don't know what is the best place to share my excitement, so I'll share it here: this is another great piece of UI (I saw it for the first time today and actually got distracted a hundred times throughout the day just because I wanted to see it again).

Kudos to everyone involved in making this small but beautiful thing happen and a special thanks to @sshanzel because this is the second time (I just started counting from the comment box, so maybe even more 😅) that you've delivered a piece of UI that sparks joy in me!

Screenshot 2023-11-14 at 20 39 29

sshanzel:
@kirillkurko great to hear from you again! Glad to hear you liked it! It was indeed a beautiful work and it wouldn't have been possible without our amazing design team! You know them, they always surprise us with these kinds of artworks. Now you made me more conscious whether what I will be putting up is level to your standards 😂

PR #2335, opened 2023-11-07
fix: add new generic referral popup

rebelchris:
- Introduces the new auto popup
- Automatically will fire event to set it to false and update the date (API handled)
- @sshanzel not sure if you want a intermediate branch so long?

image
image

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1930 #done


packages/shared/src/components/modals/referral/GenericReferralModal.tsx
```diff
@@ -0,0 +1,134 @@
...
+          <p className="typo-title3 text-theme-label-secondary">
+            And make Ido (our CTO) smile again.
+          </p>
+        </div>
+        <img
+          src={
+            !shareState
+              ? cloudinary.referralCampaign.generic.sad
+              : cloudinary.referralCampaign.generic.happy
+          }
```
omBratteng, line 65: Just wondering if it could be done like this
```suggestion
          src={
            cloudinary.referralCampaign.generic[!shareState ? 'sad' : 'happy']
          }
```

packages/shared/src/components/GrowthBookProvider.tsx
```diff
@@ -168,7 +168,7 @@ export const GrowthBookProvider = ({
 };
 
 export const useFeatureIsOn = (feature: Feature<JSONValue>): boolean =>
-  gbUseFeatureIsOn(feature.id);
+  gbUseFeatureIsOn(feature?.id);
```
capJavert, line 171: how can this happen since we have our features hardcoded in imports?

packages/shared/src/contexts/AlertContext.tsx
```diff
@@ -67,14 +74,38 @@ export const AlertContextProvider = ({
...
+      //   updateAlerts({
+      //     ...alerts,
+      //     showGenericReferral: false,
+      //   }),
+      // onError: () => {
+      //   updateAlerts({
+      //     ...alerts,
+      //     showGenericReferral: true,
+      //   });
+      // },
```
capJavert, line 90: Accidental?

packages/shared/src/hooks/referral/useReferralReminder.ts
```diff
@@ -0,0 +1,31 @@
+import { useContext, useEffect } from 'react';
+import { useLazyModal } from '../useLazyModal';
+import { LazyModal } from '../../components/modals/common/types';
+import AlertContext from '../../contexts/AlertContext';
+import useDebounce from '../useDebounce';
+
+export const useReferralReminder = (): void => {
+  const { alerts, updateLastReferralReminder } = useContext(AlertContext);
+  const { openModal } = useLazyModal();
+  // We need a small debounce to avoid it not being opened due to fast re-renders
```
capJavert, line 10: When does this happen exactly? Fixing race conditions with delays usually fails just later. I'd be happy to take a look if needed.

capJavert (commented):
This is awesome 🚀 😆 Few comments on stuff that looks off.

packages/shared/src/hooks/referral/useReferralReminder.ts
rebelchris: Just on main feed :O 
Remove debounce and simply open modal and it happens.

I got it working with other evaluation criteria, but felt even worse.
Would love to hear other thoughts

packages/shared/src/components/GrowthBookProvider.tsx
rebelchris: This is always evaluated, but for referral there are [cases](https://github.com/dailydotdev/apps/pull/2335/files#diff-612332c695635127288cbed4bd774bd6653f2fafe99e39017fbab20ae44c618fR28-R30) where they not linked to a feature.
Let me know if you think other solutions might be better?
I can't dynamic render the hook though

```
const featureFlag = campaignFeatureFlagMap[campaignKey];
  const isCampaignEnabled = useFeatureIsOn(featureFlag);
```

packages/shared/src/contexts/AlertContext.tsx
rebelchris: Ah yep, need to bring this back, my bad

packages/shared/src/components/modals/referral/GenericReferralModal.tsx
rebelchris: Should yes, let me try :)

packages/shared/src/components/GrowthBookProvider.tsx
capJavert: `gbUseFeatureIsOn` requires feature so not sure what happens when we pass undefined. If we do this we need to also make `feature` optional in our `useFeatureIsOn`.

sshanzel:
> @sshanzel not sure if you want a intermediate branch so long?

I have thought of it too. But I realized these features can go on their own as long as we ship the `join` page changes first.

packages/shared/src/hooks/referral/useReferralCampaign.ts
```diff
@@ -41,7 +42,7 @@ const useReferralCampaign = ({
   campaignKey,
 }: UseReferralCampaignProps): UseReferralCampaign => {
   const featureFlag = campaignFeatureFlagMap[campaignKey];
-  const isCampaignEnabled = useFeatureIsOn(featureFlag);
+  const isCampaignEnabled = useFeatureIsOn(featureFlag) || !featureFlag;
```
sshanzel, line 45: I think the condition should be specific whether it is the generic referral or a group of valid campaign keys. Reason being, the condition would now be true on any case.

rebelchris: We can make it based on campaignkey valid? Would that work?

sshanzel: Yes, I think that should work 🚀

rebelchris: @sshanzel Changed the evaluation, let me know what you think?

sshanzel: Looks amazing! 🤩

packages/shared/src/hooks/referral/useReferralReminder.ts
sshanzel: This is very interesting, I wonder what could be the root cause of this issue. Could it be the main content getting mounted twice, hence the lazy modal element is getting removed from the initial instance?

sshanzel (approved):
Nothing else from my side. Everything looked great! 🥳

packages/shared/src/hooks/referral/useReferralReminder.ts
rebelchris: We tried to debug it for a while, no idea.
Disabling the query (which is safe for this one) works weird enough.

capJavert (approved):

packages/shared/src/hooks/referral/useReferralReminder.ts
capJavert: It's a combination of react 17 and react-query.. sandbox worked because it was react 18, here is a sandbox that does not work https://codesandbox.io/s/admiring-surf-mjjtq4?file=/src/App.js

On 4.x and beyond it works https://codesandbox.io/s/zealous-saha-s9dzxm?file=/src/App.js

On 5.x it works also but looks like 5.x needs react 18 because it uses useSyncExternalStore hook

PR #2337, opened 2023-11-08
refactor: join page

nensidosari:
## Changes
- created component config for each referral page (generic, search)
- accepted `handleAcceptClick, isLoading, isSuccess` as props in the referral components (generic, search) and removed the corresponding code
- move mutation code and handleAcceptClick in join page

- moves join page logic code inside join page, rather then having it in different versions of join page (search invite, generic referral)
- this change was initiated in [PR](https://github.com/dailydotdev/apps/pull/2325) for adding generic referral page in [comments](https://github.com/dailydotdev/apps/pull/2325#discussion_r1380193842)
- this branch is created on top of WT-1926-referral-page and PR raised based on that branch in order to make easier the review process for this particular refactor

## Events

- [x] MobileL (420px)
- [x] Tablet (656px)
- [x] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1926 #done


packages/webapp/package.json
```diff
@@ -15,6 +15,7 @@
   },
   "prettier": "@dailydotdev/prettier-config",
   "dependencies": {
+    "@growthbook/growthbook": "https://gitpkg.now.sh/dailydotdev/growthbook/packages/sdk-js?a0d1a3add257acdbf5353dfd690547feebbada92",
```
rebelchris, line 18: Was this intentional?

packages/webapp/components/invite/common.ts
```diff
@@ -1,9 +1,29 @@
...
+  campaign: ReferralCampaignKey;
+  handleAcceptClick: () => void | Promise<void>;
+  isLoading: boolean;
+  isSuccess: boolean;
+  redirectTo: string;
+}
+
+export interface ComponentConfig {
+  actionType: ActionType;
+  feature?: Feature<JSONValue>;
```
rebelchris, line 25:
```suggestion
  feature?: Feature;
```
This would work right?

rebelchris (commented):
Nice, in general does the join page styling also not overlap quite a bit?
Just wondering if it could be a generic page with some addition copy?

packages/webapp/components/invite/common.ts
nensidosari: Nope, it requires an argument. That's why the dependency above:
`    "@growthbook/growthbook": "https://gitpkg.now.sh/dailydotdev/growthbook/packages/sdk-js?a0d1a3add257acdbf5353dfd690547feebbada92",
`

packages/webapp/package.json
nensidosari: Yes, related to https://github.com/dailydotdev/apps/pull/2337#discussion_r1386268280

nensidosari:
> Nice, in general does the join page styling also not overlap quite a bit?
> Just wondering if it could be a generic page with some addition copy?

@rebelchris U mean the generic referral and search? If yes, they do a bit in the layout, but still there is a lot of custom changes between the 2, that at the moment I think it would still be a bit messy if made generic... 🤔

packages/webapp/components/invite/common.ts
DragosIorgulescu: so did we need to fork the growthbook library? Sry if I'm missing some context, but if that is the case, why did we need to do this?

rebelchris: @nensidosari Let's move these types to shared then?
It will make more sense since that already has GB

nensidosari:
Pages not that similar anymore based on discussion on [comments](https://github.com/dailydotdev/apps/pull/2325#discussion_r1386438581)

PR #2338, opened 2023-11-09
chore: start react query v4 upgrade

rebelchris:
- Start on the react query -> tanstack query v4 migration
- Basically whole setup changed and had to ensure correct queryKeys everywhere

## TODO:
- [ ] Test locally 
- [ ] Test on preview deployment
- [ ] Test critical flows
- [ ] Test extension
- [ ] Test companion

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done


packages/shared/src/hooks/useRequestProtocol.ts
```diff
@@ -1,12 +1,12 @@
...
 import request from 'graphql-request';
-import { useQueryClient } from 'react-query';
+import { useQueryClient } from '@tanstack/react-query';
 import { RequestProtocol, REQUEST_PROTOCOL_KEY } from '../graphql/common';
 
 export const useRequestProtocol = (): RequestProtocol => {
   const client = useQueryClient();
   const { requestMethod, fetchMethod, isCompanion } =
-    client.getQueryData<RequestProtocol>(REQUEST_PROTOCOL_KEY) || {};
+    client.getQueryData<RequestProtocol>([REQUEST_PROTOCOL_KEY]) || {};
```
capJavert, line 9: I these kind of places I would just export the array from `REQUEST_PROTOCOL_KEY`

capJavert (commented):
Yeah the change to required array is major, but good thing we mostly use them already.

packages/shared/src/hooks/useRequestProtocol.ts
capJavert: It will also minimize amount of code changes probably?

rebelchris: Yeah did it for other ones, just missed this one 😢

packages/shared/src/hooks/post/useBlockPostPanel.ts
```diff
@@ -75,10 +75,11 @@ export const useBlockPostPanel = (
   const client = useQueryClient();
   const { user } = useAuthContext();
   const { checkHasCompleted, completeAction } = useActions();
-  const key = generateStorageKey(
-    StorageTopic.Post,
-    `block:${post?.id}`,
-    user?.id,
+  const key = useMemo(
+    () => [
+      generateStorageKey(StorageTopic.Post, `block:${post?.id}`, user?.id),
```
sshanzel, line 80: Our function always returns an array right? Probably just a matter of type?

sshanzel (commented):
Looks great with the changes! 🚀

packages/shared/src/hooks/post/useBlockPostPanel.ts
rebelchris: generateStorageKey returns a string, it's used for storage parts, but mixed use in this specific implementation as a react query key as well.

capJavert: Probably better to use wrapped in `generateQueryKey`?

rebelchris: Hmm ideally yes, this edge-case has local state as well which uses storage key, so to keep them sync I choose to leave it.

Ideally secondary task should be: move everything to use generateQueryKey (also the hardcoded string arrays now)

capJavert: But keep the storage key and when used in `useQuery` just wrap it in `generateQueryKey`. This way if react-query down the line changes query keys to what ever we just need to change `generateQueryKey` implementation vs looking for all array initializations.

rebelchris: But then let's just do this right away, I just wanted to avoid all this work now 😅

capJavert: I don't think we need to do all, we touch this part so let's change it, but that works as well🙏

rebelchris: Actually moving this to `queryKey` now will introduce a key change, which no longer matches with the stored one, as our generate function will always add `annonymous` as extra prop.

capJavert: Don't want to block it by additional work if we can do it down the line.

rebelchris: Changed all together 😂

packages/shared/src/hooks/companion/useBackgroundRequest.ts
```diff
@@ -14,7 +14,7 @@ export const useBackgroundRequest = (
   const client = useQueryClient();
 
   useRawBackgroundRequest(({ key, ...args }) => {
-    if (!enabled || !isQueryKeySame(key, queryKey)) {
+    if (!enabled || !isQueryKeySame([key], queryKey)) {
```
capJavert, line 17: Could this already be array in some cases, looking at `CompanionEngagements` and usage it has the `Array.isArray` check. Could you check to make sure it does not break?

packages/extension/babel.config.js
```diff
@@ -1,48 +1,34 @@
...
-  ],
-  env: {
-    production: {
-      plugins: [
+  // remove this part when https://github.com/vercel/next.js/issues/24566 is closed
+  if (isTest) {
+    return {
+      presets: [
+        '@babel/preset-env',
+        '@babel/preset-typescript',
```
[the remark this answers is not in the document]
rebelchris: Mainly cleanup our babel configs to be inline, ended up deleting quite a lot of unused stuff here.

packages/extension/jest.config.js
```diff
@@ -9,7 +9,7 @@ module.exports = {
     '<rootDir>/__tests__/helpers/',
     '<rootDir>/__tests__/fixture/',
   ],
-  transformIgnorePatterns: ['node_modules/(?!@dailydotdev)'],
+  transformIgnorePatterns: ['<rootDir>/node_modules/(?!@dailydotdev)'],
```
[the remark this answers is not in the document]
rebelchris: This is needed actually to support the main `node_modules` as well.
Not sure why I read this in a jest blog and it fixed the companion tests from not reading the node_modules in root

packages/shared/src/components/modals/SubmitArticleModal.spec.tsx
```diff
@@ -53,11 +53,7 @@ const renderComponent = (
         loadedUserFromCache
       >
         <NotificationsContextProvider>
-          <SubmitArticleModal
-            headerCopy="Submit article"
```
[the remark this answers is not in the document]
rebelchris: Header copy is no longer supported actually, not sure why the test didn't fail before

packages/shared/src/hooks/companion/useBackgroundRequest.ts
rebelchris: Good point yes!

packages/extension/jest.config.js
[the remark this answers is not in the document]
capJavert: 👍

packages/shared/src/hooks/companion/useBackgroundRequest.ts
```diff
@@ -14,12 +14,13 @@ export const useBackgroundRequest = (
   const client = useQueryClient();
 
   useRawBackgroundRequest(({ key, ...args }) => {
-    if (!enabled || !isQueryKeySame(key, queryKey)) {
+    const validKey = Array.isArray(key) ? key : [key];
+    if (!enabled || !isQueryKeySame([validKey], queryKey)) {
```
capJavert, line 18: Isn't this now double array since `key` is always wrapped in array above as `validKey`?

capJavert (approved):
Don't see anything else alarming, nice work on cleaning up usages, only one comment.

packages/shared/src/hooks/companion/useBackgroundRequest.ts
rebelchris: Good point, yes, modified! 🙏

packages/shared/src/hooks/companion/useBackgroundRequest.ts
```diff
@@ -14,12 +14,13 @@ export const useBackgroundRequest = (
...
   useRawBackgroundRequest(({ key, ...args }) => {
-    if (!enabled || !isQueryKeySame(key, queryKey)) {
+    const validKey = Array.isArray(key) ? key : [key];
+    if (!enabled || !isQueryKeySame(validKey, queryKey)) {
       return;
     }
 
     if (callback) {
-      callback({ key, ...args });
+      callback({ validKey, ...args });
```
capJavert, line 23: should be `key: validKey` since callback expects `key`?

rebelchris: I guess so, probably should make it typed eventually, but see it's widely used as generic now.

capJavert (approved):
I think good to go, nice work 🚀 , we should look for any broken pages if any invalid query key go through or cache bugs due to different keys when merged, but I think you got them.

nensidosari (commented):
LGTM. Great work! 🚀

nensidosari (approved):

packages/webapp/package.json
```diff
@@ -34,7 +34,8 @@
     "react-intersection-observer": "^8.34.0",
     "react-modal": "^3.15.1",
     "react-parallax-tilt": "^1.7.10",
-    "react-query": "3.39.1",
+    "@tanstack/react-query": "^4.36.1",
+    "@tanstack/react-query-devtools": "^4.35.3",
```
omBratteng, line 38: My only comment is about this, is it intended to add the devtools to production dependencies?

rebelchris: That's a good point, it's from their migration script, but indeed could be in dev deps for all I know.
I'll try and refactor :)

capJavert: I think you will have problems with imports in production then, worth a try. But there is no need to put it in dev dependencies since it is not included in production bundle [see here](https://tanstack.com/query/v4/docs/react/devtools#:~:text=By%20default%2C%20React%20Query%20Devtools%20are%20only%20included%20in%20bundles%20when%20process.env.NODE_ENV%20%3D%3D%3D%20%27development%27%2C%20so%20you%20don%27t%20need%20to%20worry%20about%20excluding%20them%20during%20a%20production%20build.)

rebelchris: Oh that's even better, let's leave it for now then.

omBratteng: That's good to see.

I do miss the possibility to wrap a import statement in a try...catch like I did in python when I wanted to import a dependency only during development

omBratteng (approved):

PR #2342, opened 2023-11-10
feat: generic referral campaign analytics

sshanzel:
~~Blocked: once the profile page changes is ready, I will update this PR for the relevant tracking events.~~
We will apply the proper analytics on the profile page widget on its own PR.

## Changes
- Impression when the Generic Referral Popup is seen by the user.
  - The copy link button has its built-in tracking and already added the required properties.
- Invite page
  - Same copy referral event like the popup but with a different value for the target id
  - On social share click, we send an event
- The Generic Referral Page should have an event of `download extension` instead of `accept invitation`.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1931 #done

### DIFF
--- a/packages/shared/src/components/modals/referral/GenericReferralModal.tsx
+++ b/packages/shared/src/components/modals/referral/GenericReferralModal.tsx
@@ -1,4 +1,4 @@
-import React, { ReactElement, useState } from 'react';
+import React, { ReactElement, useEffect, useState } from 'react';
 import { Modal, ModalProps } from '../common/Modal';
 import { cloudinary } from '../../../lib/image';
 import CloseButton from '../../CloseButton';
@@ -17,8 +17,9 @@ import TwitterIcon from '../../icons/Twitter';
 import { useShareOrCopyLink } from '../../../hooks/useShareOrCopyLink';
 import { link } from '../../../lib/links';
 import { labels } from '../../../lib';
-import { AnalyticsEvent, TargetId } from '../../../lib/analytics';
+import { AnalyticsEvent, TargetId, TargetType } from '../../../lib/analytics';
 import { ReferralCampaignKey, useReferralCampaign } from '../../../hooks';
+import { useAnalyticsContext } from '../../../contexts/AnalyticsContext';
 
 function GenericReferralModal({
   onRequestClose,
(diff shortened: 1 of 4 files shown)


packages/shared/src/components/modals/referral/GenericReferralModal.tsx
```diff
@@ -37,11 +38,19 @@ function GenericReferralModal({
...
     onShareOrCopyLink();
     setShareState(true);
   };
 
+  useEffect(() => {
+    trackEvent({
+      event_name: AnalyticsEvent.Impression,
+      target_type: TargetType.ReferralPopup,
+    });
+  }, [trackEvent]);
```
rebelchris, line 52: For impression we always leave this empty for now. 
I know it has edge-cases but rather keep everything uniform with code base.

packages/webapp/components/invite/Referral.tsx
```diff
@@ -31,7 +31,7 @@ export function Referral({
   const handleAcceptClick = () => {
     // since in the page view, query params are tracked automatically,
     // we don't need to send the params here explicitly
```
rebelchris, line 33: I'm still not sure what this even means, perhaps simply remove it for now.

rebelchris (changes requested):
Please mention a comment on the `accept invitation` in taxonomy for Dave so he knows we don't need it.

sshanzel:
> Please mention a comment on the accept invitation in taxonomy for Dave so he knows we don't need it.

Yes, made a comment on it while making this PR.

packages/shared/src/components/modals/referral/GenericReferralModal.tsx
sshanzel: Ah, right. To ensure we trigger it only once. Let me refactor.

packages/webapp/components/invite/Referral.tsx
sshanzel: Because in the taxonomy, there was a mention of additional properties that are found in the query params.

rebelchris (approved):

PR #2343, opened 2023-11-10
feat: referral profile widget

nensidosari:
## Changes
- Added ReferralWidget component
- added new analytics target id "profile page"
- Conditionally render ReferralWidget on user profile if user is viewing their own profile
- used useReferralCampaign to get the invite url

- Adds referral widget on user profile so user can share their invite link 

https://github.com/dailydotdev/apps/assets/36197304/1a5ad198-c3da-484d-8eb2-390d8362b93f

## Events

Did you introduce any new tracking events? Yes
Don't forget to update the [Analytics Taxonomy sheet](https://docs.google.com/spreadsheets/d/18Lv7zXges9QfVX5VYL1a-Hyl0e1sQ3sLr0OK8YZWKXI/edit#gid=0) - This was already in the exel sheet

The event existed only target id "profile page" was added when event sent through profile

| Change | copy referral link  | target_id: 'profile page' |

- [ ] Does this not break anything in companion? No

- [x] MobileL (420px)
- [x] Tablet (656px)
- [x] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1927 #done


packages/shared/src/components/cards/ReferralWidget.tsx
```diff
@@ -0,0 +1,112 @@
...
+import { TextField } from '../fields/TextField';
+import { ReferralCampaignKey, useReferralCampaign } from '../../hooks';
+import { link } from '../../lib/links';
+import { useShareOrCopyLink } from '../../hooks/useShareOrCopyLink';
+import { labels } from '../../lib';
+import { AnalyticsEvent, TargetId } from '../../lib/analytics';
+import { SimpleTooltip } from '../tooltips';
+
+const ReferralWidget = (): ReactElement => {
+  const { isReady, url } = useReferralCampaign({
```
nensidosari, line 20: Based on this in the DR:

> to get the right URL we send a query which won’t be at the same time we fetch the user information, so rendering the user details might come in first then the campaign URL which would produce the layout shift, make sure to wait for the campaign query to finish before displaying the content.

We want to display user profile after the url is fetched, but also fetch the url after the profile query has finished so we can validate if the user is seeing their own profile.

I think having these both would place us in a loop and would require us to add this rule again:
```
// eslint-disable-next-line react-hooks/rules-of-hooks
```
since we would want to render the hook to get the referral url conditionally after the profile query has finished, and I don't know if we would want to introduce more of this...
I'm thinking more of we render the profile and the referral widget, just render that disabled until the url is fetched...which would not introduce a layout shift?

Would like to hear your thoughts on this, can definitely quickly change to whatever we think is best.

capJavert: I think you just need to use `useReferralCampaign`  also in the `ProfileLayout` and once it `isReady` you render its content.

packages/webapp/components/layouts/ProfileLayout/index.tsx
```diff
@@ -92,6 +93,7 @@ export default function ProfileLayout({
 
   // Needed because sometimes initialProfile is defined and fetchedProfile is not
   const profile = fetchedProfile ?? initialProfile;
+  const isCurrentUserProfile = profile && profile.id === user?.id;
```
capJavert, line 96: To continue on the comment https://github.com/dailydotdev/apps/pull/2343/files#r1389695102

Basically here you can call `const { isReady: isReferralReady } = useReferralCampaign` and then before below add `isReferralReady` check and return null until data is ready.

capJavert: Per DR this will delay render of profile but it will avoid layout shift.

packages/shared/src/components/cards/ReferralWidget.tsx
```diff
@@ -0,0 +1,112 @@
...
+    }),
+  });
+  const onShareClick = () => {
+    onShareOrCopyLink();
+  };
+
+  return (
+    <div className="flex laptopL:absolute flex-col tablet:flex-row laptopL:flex-col flex-wrap justify-between p-4 laptopL:m-6 mt-6 laptopL:mt-0 mb-4 laptopL:max-w-widget laptopL:h-auto rounded-2xl border border-theme-divider-tertiary laptopL:left-[40rem] max-w-fit bg-theme-bg-primary">
+      <h3 className="mb-2 font-bold typo-title3">Invite friends</h3>
+      <p className="contents text-salt-50 typo-callout">
```
rebelchris, line 39: We don't support named colors.
```suggestion
      <p className="contents text-theme-label-secondary typo-callout">
```

packages/shared/src/components/cards/ReferralWidget.tsx
```diff
@@ -0,0 +1,112 @@
...
+        Tell your dev friends how easy is it to learn, collaborate, and grow
+        together
+      </p>
+      <div className="flex flex-col my-5 w-auto tablet:w-70">
```
rebelchris, line 43: It looks more like this is always rendered 100% of the parent no?

packages/shared/src/components/cards/ReferralWidget.tsx
```diff
@@ -0,0 +1,112 @@
...
+              onClick={onShareClick}
+              disabled={!isReady}
+            >
+              Copy link
+            </Button>
+          }
+          readOnly
+        />
+      </div>
+      <div className="flex justify-between items-center text-salt-90 typo-callout">
```
rebelchris, line 67: Same color issue
```suggestion
      <div className="flex justify-between items-center text-theme-label-tertiary typo-callout">
```

packages/shared/src/components/cards/ReferralWidget.tsx
```diff
@@ -0,0 +1,112 @@
...
+      event_name: AnalyticsEvent.CopyReferralLink,
+      target_id: TargetId.ProfilePage,
+    }),
+  });
+  const onShareClick = () => {
+    onShareOrCopyLink();
+  };
+
+  return (
+    <div className="flex laptopL:absolute flex-col tablet:flex-row laptopL:flex-col flex-wrap justify-between p-4 laptopL:m-6 mt-6 laptopL:mt-0 mb-4 laptopL:max-w-widget laptopL:h-auto rounded-2xl border border-theme-divider-tertiary laptopL:left-[40rem] max-w-fit bg-theme-bg-primary">
```
rebelchris, line 37: In general there are elements here that are re-used between the modal and this place, perhaps we should re-use them?

packages/webapp/components/layouts/ProfileLayout/index.tsx
rebelchris: Also we already have this evaluation on the `pages/userId/index` file as:
```
const isSameUser = profile?.id === user?.id;
```

Not sure if you can re-use it?

rebelchris (changes requested):
Nice work, some minor comments from my side 🙏

packages/shared/src/components/cards/ReferralWidget.tsx
nensidosari: Hmm, no not always, image below shows where it's not 100% of parent in tablet + the width for laptop and above of the whole widget is defined by the width of this element, that w-70. Unless u meant something else... 🤔 
Screenshot 2023-11-13 at 7 10 10 AM

packages/shared/src/components/cards/ReferralWidget.tsx
nensidosari: Before using a div I was using Card component for this wrapper, because with it it is also similar, but then I had to use a lot of `!important` to override some stuff so switched to this.. Will check with modal as well but I believe it could be the same case, but will check. Unless we just extract the common elements to a variable somewhere... 🤔

rebelchris: I mean between this implementation and the Referral modal, maybe we can share some of those :)

packages/shared/src/components/cards/ReferralWidget.tsx
rebelchris: I think it's more of a design flaw to be honest.
We should just have the minimal space requirement between the elements the rest should hit automatically I think.

packages/shared/src/components/cards/ReferralWidget.tsx
nensidosari: That wouldn't be enough as it would mean we would be fetching the referral url even if the user profile wasn't their own. Meaning to use the referral hook we would need to wait until user profile info is fetched and compared if it is their own, if so fetch the referral link and then once the referral link is ready show the profile. 
@sshanzel suggested I use `enabled` option in useQuery for referral invite, so am looking at that opportunity now.

sshanzel: There are two things here I believe:
- The first is identifying whether the user is viewing his own profile page which we will pass the value if the query should be enabled or not.
- Second thing is, to avoid the layout shift, for that, we will wait if the referral campaign query to be finished which we can utilize the `isReady` to see if the query is finished.

nensidosari: Yup, the second is quite straightforward, the first one is the complex bit, as the `enabled` should change after the profile info is fetched

sshanzel: Also, just be wary with regards to the `isReady` evaluation, our current `react-query`, I believe would return `isSuccess: false` if it is `enabled: false` as opposed to the latest version which would return `isSuccess: true`.

nensidosari: I have a question though, are we really against showing profile info if the url is not ready? Any reasons behind this except layout shift? 
Personally I believe making a whole profile wait for the referral url is too much, what if there is sth wrong with the api for referral url, will we never show the profile? I'm quite pro to not delaying profile until referral url is there, but rather we just manage the design to not introduce layout shift (things like make copy button disabled until ready, have a placeholder widget until url ready). I might be missing context, but would be curious to know.

packages/webapp/components/layouts/ProfileLayout/index.tsx
nensidosari: We can make some util fn that checks both

packages/shared/src/components/cards/ReferralWidget.tsx
sshanzel: Layout shift is bad UX, and part of the reason our design team is always going for the extra mile to build the skeletons is to avoid it.

I agree that the page can get held up due to some failing services and we also have done what you suggested, in the past, where we set a specific height to some invisible element, though that is only for elements available every time for everyone. This is only available to you when viewing your own profile, so I don't think the impact would be high in terms of the slight delay.

sshanzel: Would love to hear others' thoughts too. cc @dailydotdev/web-team

nensidosari: > where we set a specific height to some invisible element

Similar though I'm not suggesting this, I'm suggesting either skeleton for the widget or a disabled state to the copy button, or a loading state.. Sth like this, and this would be for sure only for users who are viewing their own profile

rebelchris: For now let's not slow down the development and keep it as blocking.
Feel free to open a separate discussion for improvement with design, in the end it's a UX issue and they need to advise.

nensidosari: I'll move this conversation at referral channel then

packages/shared/src/components/cards/ReferralWidget.tsx
nensidosari: Hm, the link and invite are not within a separate tag with the other header and description, which changes a bit on how to arrange the elements, if they were in a separate tag then yeah what you are saying would be doable. I just didn't want to introduce another tag when we could do with what we have. Or maybe I'm not envisioning properly how to do this 🤔 
What would be wrong with what we have already?

packages/webapp/components/layouts/ProfileLayout/index.tsx
nensidosari: Agreed to keep this as is

packages/shared/src/components/cards/ReferralWidget.tsx
nensidosari: Agreed to keep this as is

packages/shared/src/components/cards/ReferralWidget.tsx
nensidosari: Have moved render of the share icons in separate component

packages/shared/src/components/widgets/ReferralWidget.tsx
```diff
@@ -0,0 +1,68 @@
...
+    }),
+  });
+  const onShareClick = () => {
+    onShareOrCopyLink();
+  };
+
+  return (
+    <div
+      data-testid="referral-widget"
+      className="flex laptopL:absolute flex-col tablet:flex-row laptopL:flex-col flex-wrap justify-between p-4 laptopL:m-6 mt-6 laptopL:mt-0 mb-4 laptopL:max-w-widget laptopL:h-auto rounded-2xl border border-theme-divider-tertiary laptopL:left-[40rem] max-w-fit bg-theme-bg-primary"
```
sshanzel, line 27: Probably just a personal preference, but I think it is safer to use percentages than fixed values. Results are usually more predictable.

```suggestion
      className="flex laptopL:absolute flex-col tablet:flex-row laptopL:flex-col flex-wrap justify-between p-4 laptopL:m-6 mt-6 laptopL:mt-0 mb-4 laptopL:max-w-widget laptopL:h-auto rounded-2xl border border-theme-divider-tertiary laptopL:left-full max-w-fit bg-theme-bg-primary"
```

packages/shared/src/components/widgets/ReferralWidget.tsx
```diff
@@ -0,0 +1,68 @@
...
+    >
+      <h3 className="mb-2 font-bold typo-title3">Invite friends</h3>
```
sshanzel, line 29: From the Sketch I think it says title 2.
```suggestion
      <h3 className="mb-2 font-bold typo-title2">Invite friends</h3>
```

packages/shared/src/components/widgets/ReferralWidget.tsx
```diff
@@ -0,0 +1,68 @@
...
+      <p className="contents text-theme-label-secondary typo-callout">
```
sshanzel, line 30: Can you elaborate why do need this class? Seems to be not available on tailwindcss.
```suggestion
      <p className="text-theme-label-secondary typo-callout">
```

packages/shared/src/components/widgets/ReferralWidget.tsx
```diff
@@ -0,0 +1,68 @@
...
+        Tell your dev friends how easy is it to learn, collaborate, and grow
+        together
+      </p>
+      <div className="flex flex-col my-5 w-auto tablet:w-70">
```
sshanzel, line 34: Is it necessary for us to wrap the component in a div?

packages/shared/src/hooks/referral/useReferralCampaign.ts
```diff
@@ -40,12 +41,14 @@ const campaignFeatureFlagMap: Partial<
 
 const useReferralCampaign = ({
   campaignKey,
+  enabled = true,
```
rebelchris, line 44: Why did we add this here?

packages/shared/src/components/widgets/ReferralSocialShareButtons.tsx
```diff
@@ -0,0 +1,56 @@
...
+  getWhatsappShareLink,
+} from '../../lib/share';
+import WhatsappIcon from '../icons/Whatsapp';
+import FacebookIcon from '../icons/Facebook';
+import TwitterIcon from '../icons/Twitter';
+import { link } from '../../lib/links';
+import { labels } from '../../lib';
+import { SimpleTooltip } from '../tooltips';
+
+const ReferralSocialShareButtons = ({ url }: { url: string }): ReactElement => {
```
rebelchris, line 15: I think I didn't add this yet, but this should have tracking @sshanzel not sure you wanted to add it seperate?

rebelchris (approved):
Nice 2 non blocking comments from my side.
The tracking can be seperate ticket.

packages/shared/src/components/widgets/ReferralWidget.tsx
```diff
@@ -0,0 +1,68 @@
...
+    text: labels.referral.generic.inviteText,
+    link: inviteLink,
+    trackObject: () => ({
+      event_name: AnalyticsEvent.CopyReferralLink,
+      target_id: TargetId.ProfilePage,
+    }),
+  });
+  const onShareClick = () => {
+    onShareOrCopyLink();
+  };
```
sshanzel, line 22: No need to wrap this, we can just do an inline function.
```suggestion
  const onShareClick = () => {
    onShareOrCopyLink();
  };
```

packages/webapp/components/layouts/ProfileLayout/index.tsx
```diff
@@ -148,7 +156,9 @@ export default function ProfileLayout({
     }
   }, [profile]);
 
-  if (isFallback && !initialProfile) {
+  const isReadyProfile = isCurrentUserProfile && !isReady;
```
sshanzel, line 159: I think the variable name is misleading. If the profile is ready, I am expecting to see the content not an empty page. You can use the suggestion below or something similar.
```suggestion
  const isLoadingReferral = isCurrentUserProfile && !isReady;
```

packages/shared/src/hooks/referral/useReferralCampaign.ts
nensidosari: In order to not fire the query if the profile is not the logged in user profile https://github.com/dailydotdev/apps/blob/b7853ddca89e22739b898a322c2e3e193c11491e/packages/webapp/components/layouts/ProfileLayout/index.tsx#L79

sshanzel (changes requested):
Looks great! Just had a few very minor comments. Then the only thing missing would be the analytics for clicking the social provider buttons which we send `invite referral` event.

packages/shared/src/components/widgets/ReferralWidget.tsx
nensidosari: Yes but the size in sketch is 20px which is title 3, from design it also looks like it makes sense to be title3 as it's the same as the user name in the profile... 🤔

packages/shared/src/components/widgets/ReferralWidget.tsx
nensidosari: we need the classes but moved those to the container classnames for TextField components

packages/shared/src/components/widgets/ReferralSocialShareButtons.tsx
sshanzel: I think there is nothing left for us, so I thought of having the analytics here too.

packages/shared/src/components/widgets/ReferralWidget.tsx
rebelchris: I see it's set as typo-2 but design overwritten.
The general rule of thumb would be to take whatever the name is set to, we don't really look at pixel values if that makes sense?

But when we do see these overwrites we should notify design it just means something went wrong in the sync most of the time. (See little warning signs)

Screenshot 2023-11-14 at 08 42 18

packages/webapp/components/layouts/ProfileLayout/index.tsx
nensidosari: Agree, I think your suggestion is quite good so going with that

packages/shared/src/components/widgets/ReferralWidget.tsx
nensidosari: True true, I've missed this in my list of questions for design

packages/shared/src/components/widgets/ReferralSocialShareButtons.tsx
nensidosari: Yeah sure can add that here if we want

packages/shared/src/components/widgets/ReferralWidget.tsx
sshanzel: @rebelchris thank you for clearing it up! Seems like they wanted title3 anyway. So we should be good here 🚀

packages/shared/src/components/widgets/ReferralSocialShareButtons.tsx
nensidosari: Just pushed the events as well now.

packages/shared/src/components/widgets/ReferralSocialShareButtons.tsx
```diff
@@ -0,0 +1,76 @@
...
+import { SimpleTooltip } from '../tooltips';
+import { AnalyticsEvent, TargetType } from '../../lib/analytics';
+import AnalyticsContext from '../../contexts/AnalyticsContext';
+
+const ReferralSocialShareButtons = ({
+  url,
+  origin,
+}: {
+  url: string;
+  origin: TargetType;
```
sshanzel, line 23: Origin is a different property we send on certain events. Let's use `targetType` to avoid confusion. Also maybe we should introduce `ReferralSocialShareButtonsProps` to clean it a bit.

```
interface ReferralSocialShareButtonsProps {
  url: string;
  origin: TargetType;
}
```

```suggestion
  targetType: TargetType;
```

sshanzel (approved):
Amazing work! Just had one non blocking comment, but everything looks to be good to go! :shipit:

packages/shared/src/components/widgets/ReferralSocialShareButtons.tsx
rebelchris: Agree with Lee in this case.
Did open a [discussion](https://dailydotdev.slack.com/archives/C02JAUF8HJL/p1699946558816669) for experiment team to give us better clarity, seems like they randomly assign the properties now.

nensidosari: On it

nensidosari: Done, thank u!

rebelchris (approved):

PR #2349, opened 2023-11-14
fix: missing close button on profile menu mobile

sshanzel:
## Changes
- Interactive popup should show the close button on mobile or when the user decides to always show it, similar to Squad Checklist.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1958 #done

### DIFF
--- a/packages/shared/src/components/tooltips/InteractivePopup.tsx
+++ b/packages/shared/src/components/tooltips/InteractivePopup.tsx
@@ -5,6 +5,8 @@ import { useSettingsContext } from '../../contexts/SettingsContext';
 import useSidebarRendered from '../../hooks/useSidebarRendered';
 import ConditionalWrapper from '../ConditionalWrapper';
 import useWindowEvents from '../../hooks/useWindowEvents';
+import { Button, ButtonSize } from '../buttons/Button';
+import CloseIcon from '../icons/MiniClose';
 
 export enum InteractivePopupPosition {
   Center = 'center',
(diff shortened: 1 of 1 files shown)


packages/shared/src/components/tooltips/InteractivePopup.tsx
```diff
@@ -25,6 +27,7 @@ interface InteractivePopupProps {
   className?: string;
   position?: InteractivePopupPosition;
   closeOutsideClick?: boolean;
+  alwaysShowClose?: boolean;
```
rebelchris, line 30: Did we nee do introduce this, does it not always exist?
And would maybe rename it then to `showClose`?

rebelchris (approved):
Non breaking question

packages/shared/src/components/tooltips/InteractivePopup.tsx
sshanzel: For the case of the profile menu it doesn't show all the time for laptop, or should we make it not show when it is profile menu only rather than optional? Wdyt?

rebelchris: Since interactive is a generic popup, i'm not sure.
EG how is it handled in squad tour?

Perhaps it's a component inside rather, where we can add the logic?

sshanzel: The Squad tour is the old implementation where it uses a modal on mobile. Though on mobile we always have close button, so I guess the better way is to always display it, except Profile menu on desktop? Wdyt?

rebelchris: Yeah agreed :)

sshanzel: Confirmed to still work 🥳 

Screenshot 2023-11-14 at 9 57 42 PM

PR #2352, opened 2023-11-15
refactor: referral program copy changes

sshanzel:
## Changes
- Copy changes from the Miro link below:

https://miro.com/app/board/uXjVNe2aNY8=/?messenger=slack&utm_source=slack&utm_medium=messenger&utm_campaign=notifications&utm_content=unfurled_chat_link&utm_term=messenger_analytics

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1972 #done

### DIFF
--- a/packages/shared/src/components/referral/InviteLinkInput.tsx
+++ b/packages/shared/src/components/referral/InviteLinkInput.tsx
@@ -51,7 +51,7 @@ export function InviteLinkInput({
     }
 
     if (hasCopied) {
-      return text?.copied ?? copy;
+      return text?.copied ?? 'Copied';
     }
 
     return copy;
--- a/packages/shared/src/components/referral/InviteLinkInput.tsx
+++ b/packages/shared/src/components/referral/InviteLinkInput.tsx
@@ -1,4 +1,4 @@
-import React, { ReactElement } from 'react';
+import React, { ReactElement, useState } from 'react';
 import { Button, ButtonSize } from '../buttons/Button';
 import { TextField } from '../fields/TextField';
 import { AnalyticsEvent, TargetId } from '../../lib/analytics';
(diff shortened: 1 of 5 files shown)


packages/shared/src/components/referral/InviteLinkInput.tsx
```diff
@@ -29,20 +36,25 @@ export function InviteLinkInput({
...
     }
   };
 
   const renderText = () => {
-    const copy = 'Copy link';
+    const copy = text?.initial ?? 'Copy link';
+
+    if (text?.copying) {
+      return copying ? text?.copying : copy;
+    }
```
rebelchris, line 51: Where did this come from, I don't see this part in the miro

rebelchris (changes requested):
Just one I don't see

packages/shared/src/components/referral/InviteLinkInput.tsx
sshanzel: When click a "Copied" button, nothing really changed, kind of a bad UX, but I can remove this if we don't want it.

rebelchris: Rather check this UX with Sab, I think they are not aware of how it works maybe.

rebelchris: Or they intended to have:
Copy link -> click -> Copied -> return back to Copy link 
As it is in current behaviour

sshanzel: Ah, I see now why there is a short delay with returning it back. In that case, the previous one was correct.

sshanzel: Fixed with the copied state now 🚀

rebelchris (approved):

PR #2361, opened 2023-11-22
fix: add wrapped useMedia hook

rebelchris:
- Created a wrapped useMedia hook to evaluate the breakpoints we use now.
- Want to see what you think before I refactor all implementations.

Usage would be as:

```
const isMobile = useViewSize(ViewSize.MobileL); // true or false output
```

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1938 #done


packages/shared/src/hooks/useViewSize.ts
```diff
@@ -0,0 +1,26 @@
...
+  isMobile: boolean;
+  isTablet: boolean;
+  isLaptop: boolean;
+  isLaptopL: boolean;
+};
+const useViewSize = (): ViewSize => {
+  const isMobile = !useMedia([tablet.replace('@media ', '')], [true], false);
+  const isTablet = useMedia([tablet.replace('@media ', '')], [true], false);
+  const isLaptop = useMedia([laptop.replace('@media ', '')], [true], false);
+  const isLaptopL = useMedia([laptopL.replace('@media ', '')], [true], false);
```
omBratteng, line 15: What about laptopXL, desktop and desktopL?

rebelchris: Didn't really see usage of those for this hook. (they mostly CSS only)
As this hook can become quite heavy on window events wanted to keep it to the bare minimum.

packages/shared/src/hooks/useViewSize.ts
```diff
@@ -0,0 +1,26 @@
...
+
+  return useMemo(() => {
+    return {
+      isMobile,
+      isTablet,
+      isLaptop,
+      isLaptopL,
```
capJavert, line 22: My issue is that this is really inefficient because we register one media change handler for each of those `useMedia` hook when we could just evaluate everything in a single `useMedia`. Also on what @omBratteng commented, what about other breakpoints, having a bunch of different computed values again sounds inefficient. 

Do you think we could do it other way around by having something in lines of `useViewSize('laptop')` and that returns `true|false` and under the hood we sort out the parameters needed for `useMedia` which is the real part that is "heavy" and duplicated all over codebase, WDYT?

capJavert: The version with array would be the same as calling two hooks so I think just calling two hooks is better in this case and also not sure you could reuse `useMedia` in array manner.  Quick search around codebase shows only two files `MainFeedLayout` and `OnboardingHeader` where we `useMedia` two times in a single file if I looked correctly.

Also agree we should always try and do it through CSS because the JS check is not SSR compatible so it would bite us down the line. 👌

capJavert: I think most of the usages are boolean based which this the approach solves.

rebelchris: I see so you propose a wrapping hook that can act on single sizes, but simply uses useMedia wrapped?
So 

```
useViewSize(ENUMLAPTOP)
```

```
// inside the hook
return useMedia(FORTHEENUM)
```

We can also simply change the useMedia to do the evals inside probably? (that way we only have the one hook)

capJavert: For sure, basically if we eliminate ugly API from useMedia we solve everything.. way to go 👌

rebelchris: I see we have 2 edge-cases were we use it for other purposes so I'll still do wrapper, but differently.
Proposing change soon, thanks for the discussion.

packages/shared/src/hooks/useViewSize.ts
```diff
@@ -0,0 +1,48 @@
...
+  MobileL = 'mobileL',
+  Tablet = 'tablet',
+  Laptop = 'laptop',
+  LaptopL = 'laptopL',
+  LaptopXL = 'laptopXL',
+  Desktop = 'desktop',
+  DesktopL = 'desktopL',
+}
+
+const reversedEvaluatedSizes = [ViewSize.MobileM, ViewSize.MobileL];
```
capJavert, line 24: What is this for?

capJavert (approved):
Looks great 👌 only one question

packages/shared/src/hooks/useViewSize.ts
rebelchris: So for those two we actually check if they are NOT one of those.
EG: MobileL checks if it's not Tablet.

This is a downside of us using `min-width` if we were to add `max-width` as types we could do it, but i was a bit hesitant as it also affect CSS usage.

Hope that explains it well?

You can see the final use in the return where we either return the check or the negated check.

Which is the same as this from the previous:
```
  const isMobile = !useMedia([tablet.replace('@media ', '')], [true], false);
  const isTablet = !useMedia([laptop.replace('@media ', '')], [true], false);
  const isLaptop = useMedia([laptop.replace('@media ', '')], [true], false);
```

capJavert: Hmm, would negation on the hook just solve it `!useViewSize(mobileM)`? Or more verbose:
```js
const isMobile = useViewSize(mobileM) === false
```

Feels wrong to have that kind of internal logic that is not obvious from hooks public interface.

rebelchris: Wont work as the evaluation is against another query basically.
The only solution is i think to use max width then, but we already go against this pattern now

capJavert: I personally think below is clear:
```js
const isMobile = useViewSize(mobileM) === false
```

It's just comparing two boolean values in javascript? Let's ask others, I am fine with going either way.

rebelchris: Let's wait for others indeed.
From my point of view it's a bit unclear you have to do this:

```js
// To evaluate if isMobile is true
const isMobile = useViewSize(mobileM) === false

// To evaluate if tablet is true
const isTablet = useViewSize(tablet)
```

I feel like for someone new you wouldn't know you need to manually negate for mobile sizes right?
This is why I wanted to solve it inside the hook so the return values are reflecting what you are requesting.

sshanzel: > I feel like for someone new you wouldn't know you need to manually negate for mobile sizes right?

Same thoughts. Hence, I would much prefer if it would automatically be ready right from the instance call.

DragosIorgulescu: I am on the same boat that we should not expose internal implementation on the public interface of the function. Let's keep it consistent?

DragosIorgulescu (approved):
👌

packages/shared/src/hooks/useViewSize.ts
rebelchris: @DragosIorgulescu Not sure I get the right idea, you like the proposed idea or Ante's idea?

rebelchris: @capJavert Are you ok with keeping it inside the hook as proposed?
See discussion above.

capJavert: 🚀

PR #2362, opened 2023-11-22
refactor: unnecessary calls of tags categories

sshanzel:
## Changes
- After further investigation, one of the reasons why we have a high number of requests from this query is due to the context menu. The nature of our context menu is rendered all the time, and with it being rendered, it would send a request as the query is always enabled. Since the menu is present in both the feed and post page (our most highly visited pages), it is apparent that the number of calls for this query would be high as well.

Note that this is only partial of the graphql loads task, but thought shipping out bits of fixes doesn't sound bad.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done


omBratteng (commented):
Just as an FIY, the `TagCategories` was actually only requested once (from what I could see), but because it has 3 sub queries inside it (tagsCategories, feedSettings and advancedSettings) the metrics counter in API incremented the counter for `TagCategories` for each sub-query.

See this PR https://github.com/dailydotdev/daily-api/pull/1540

It has now stabilised to more normal levels, on par with `SearchSessionHistory` and `UserReadingRank`

sshanzel:
> Just as an FIY, the TagCategories was actually only requested once (from what I could see), but because it has 3 sub queries inside it (tagsCategories, feedSettings and advancedSettings) the metrics counter in API incremented the counter for TagCategories for each sub-query.

Yes, thankfully react-query consolidates the request and only fires once. Just that it is always requested even when unnecessary.

I think we should also set some standard/benchmark where we can consider whether a request is way over the threshold and when it should be alright.

packages/shared/src/hooks/useFeedSettings.ts
```diff
@@ -64,7 +69,7 @@ export default function useFeedSettings(): FeedSettingsReturnType {
 
       return { ...req, feedSettings };
     },
-    { ...disabledRefetch },
+    { ...disabledRefetch, enabled },
```
rebelchris, line 72: Should we consider adding staleTime here as well? 
My only note on that field would be to validate if we manually reset/mutate the settings on change.

sshanzel: Yeah, I was thinking of that too. Maybe we should start considering it on most of our queries, especially the tooltip-related query.

Would 60 seconds be sufficient for this one?

rebelchris: We should indeed, it can also be part of the global research for the loads, so no need to do for all.
The time totally depends on the type of query, this one needs to "behave" realtime, meaning if people make changes it should reflect.
So the main thing is we need to validate that happens (any tag changes etc should reflect invaldiation/mutation) if that happens we can go with 60 (I think I set a default on another query already, which we can reuse)

sshanzel: Awesome. I will check if invalidation happens whenever we send the relevant mutations.

sshanzel: After checking, the `useMutateFilters` manually updates the local cache so no need for invalidation.

advanced settings:
https://github.com/dailydotdev/apps/blob/dfa97065f51a017db5babf2f8bda03570c6f5dda/packages/shared/src/hooks/useMutateFilters.ts#L85-L99

tags:
https://github.com/dailydotdev/apps/blob/dfa97065f51a017db5babf2f8bda03570c6f5dda/packages/shared/src/hooks/useMutateFilters.ts#L106-L120

sources:
https://github.com/dailydotdev/apps/blob/dfa97065f51a017db5babf2f8bda03570c6f5dda/packages/shared/src/hooks/useMutateFilters.ts#L127-L141

packages/shared/src/hooks/useFeedSettings.ts
```diff
@@ -54,17 +61,17 @@ export default function useFeedSettings(): FeedSettingsReturnType {
         { loggedIn: !!user?.id },
       );
 
-      if (user) {
-        return req;
-      }
-
```
sshanzel, line 60: Realized we don't show the feed to anonymous users anymore so we don't need this condition here but rather to the `enabled` property.

packages/shared/src/hooks/useContextMenu.ts
```diff
@@ -21,7 +22,9 @@ export default function useContextMenu({
   );
   const client = useQueryClient();
   const { show, hideAll } = useContexifyContextMenu({ id });
-  const { data: isOpen } = useQuery(key, () => client.getQueryData(key));
+  const { data: isOpen } = useQuery(key, () =>
+    client.getQueryData<boolean>(key),
+  );
```
rebelchris, line 27: Thinking about this now, I think in react query 4 we actually need to return something here as a default.
Did it not complain about this?

packages/shared/src/hooks/useFeedSettings.ts
```diff
@@ -41,10 +41,17 @@ const isObjectEmpty = (obj: unknown) => {
   return Object.keys(obj).length === 0;
 };
 
-export default function useFeedSettings(): FeedSettingsReturnType {
+interface UseFeedSettingsProps {
+  enabled?: boolean;
+}
+
+const FEED_SETTINGS_STALE_TIME = 1000 * 60;
```
rebelchris, line 48: Maybe put it where we have the react query key generator fn?
And do something like:
```
const ONE_MINUTE = 1000 * 60
export const FEED_SETTINGS_STALE_TIME = ONE_MINUTE
```

(will allow us to be more flexible in the future. (and maybe define: DEFAULT_STALE_TIME = X_MINUTE

rebelchris (commented):
Minor suggetions, also when you do merge please do a E2E test on production :)

packages/shared/src/hooks/useFeedSettings.ts
sshanzel: Yeah, that's better.

packages/shared/src/hooks/useContextMenu.ts
sshanzel: It did not surprisingly. Do you know the specific doc page about this so I can look it up?

capJavert: We just need to make sure it is not `undefined` https://tanstack.com/query/v4/docs/react/guides/migrating-to-react-query-4#undefined-is-an-illegal-cache-value-for-successful-queries

rebelchris: Yep as Ante mentions, I did a fix in a couple of items already in my PR.
(think modal and toast for instance)

capJavert (approved):
Looks good overall, just the comments.

sshanzel:
> Minor suggetions, also when you do merge please do a E2E test on production :)

Definitely. Before and after, I will run some tests.

packages/shared/src/hooks/useContextMenu.ts
sshanzel: Updated. Thank you both!

sshanzel:
Also did a code cleanup, will do a thorough testing end to end before and after merge.

packages/shared/src/graphql/feedSettings.ts
```diff
@@ -72,14 +65,14 @@ export const SEARCH_TAGS_QUERY = gql`
 `;
 
 export const FEED_SETTINGS_QUERY = gql`
-  query TagCategories($loggedIn: Boolean!) {
+  query TagCategories {
```
omBratteng, line 68: While we're at it, could we maybe rename this to something else?
Like FeedSettings? I feel like the name `TagCategories` is a bit misleading, as it also fetches `feedSettings` and `advancedSettings` in addition the the actual `tagsCategories`?

Maybe call it FeedSettings or something.
```suggestion
  query FeedSettings {
```

I was thinking about sneaking it into one of my YouTube changes, but since you're already here. Thoughts @rebelchris?

sshanzel: Yeah, sure, maybe `FeedConfig` can work? Just a bit odd when calling it since it has a property called the exact same thing (feedSettings). Wdyt?

omBratteng: Didn't really like FeedSettings tbh, so
FeedPreferences is a much better name I think.

rebelchris: Works for me :)

omBratteng (approved):

PR #2363, opened 2023-11-23
chore: gitpod env for api to mock user

sshanzel:
## Changes
- Add the user id for which user will be mocked when using our docker image (which is used in GitPod).

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -70,3 +70,4 @@ services:
       - JWT_PUBLIC_KEY_PATH=/opt/app/.cert/public.pem
       - JWT_PRIVATE_KEY_PATH=/opt/app/.cert/key.pem
       - GROWTHBOOK_CLIENT_KEY='local'
+      - MOCK_USER_ID='404'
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -70,3 +70,4 @@ services:
       - JWT_PUBLIC_KEY_PATH=/opt/app/.cert/public.pem
       - JWT_PRIVATE_KEY_PATH=/opt/app/.cert/key.pem
       - GROWTHBOOK_CLIENT_KEY='local'
+      - MOCK_USER_ID='404'
```
rebelchris, line 73: I think we should probably add another user for this, the 404 is technically a deleted user and might eventually be unclear to people/cause issues.
Let me know what you think?

sshanzel: Ah, in that case, yes, it is probably better to introduce a more defined user for this purpose.

capJavert (approved):

PR #2364, opened 2023-11-23
fix(tools): exclude searching in dist and node_modules in the VS Code workspace

omBratteng:
## Changes

Excludes searching from the dist or node_modules directories.

Always annoyed me when I tried to search for something, and VS Code would also show me lots of results from the compiled code from the `dist` directory for extension or `.next` for webapp.

This just sets the setting for the workspace to exclude searching in those directories.

### DIFF
--- a/.vscode/apps.code-workspace
+++ b/.vscode/apps.code-workspace
@@ -18,6 +18,11 @@
     "editor.quickSuggestions": {
       "strings": true
     },
+    "search.exclude": {
+      "**/node_modules": true,
+      "**/dist": true,
+      "**/.next": true,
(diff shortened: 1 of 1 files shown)


capJavert (approved):

rebelchris (approved):

rebelchris:
Just use WS we already had this exclude :)

omBratteng:
I wanted to say it would be a cold day in Hell before I use any JetBrains tools, but the issue with that is that Hell is a place 10 minutes from my home.
So when it is cold at my home, it is cold in Hell too.

PR #2365, opened 2023-11-23
fix: disallow non-existing rank indexes to be returned by `getRank`

DragosIorgulescu:
- may be part of the issues causing some users to experience authentication issues - as per slack thread: https://dailydotdev.slack.com/archives/C01L0QXQJNB/p1700645470832879

## Events

N / A

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done

### DIFF
--- a/packages/shared/src/lib/rank.ts
+++ b/packages/shared/src/lib/rank.ts
@@ -104,15 +104,13 @@ interface GetNextRankTextProps {
 }
 
 export const getRank = (rank: number): number => {
-  if (rank < 0) {
-    return 0;
-  }
+  const computedRank = Math.max(0, rank - 1);
 
-  if (rank >= RANKS.length) {
+  if (computedRank >= RANKS.length) {
     return RANKS.length - 1;
   }
 
-  return rank;
+  return computedRank;
 };
 
 export const getNextRankText = ({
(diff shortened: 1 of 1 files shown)


packages/shared/src/lib/rank.ts
```diff
@@ -104,7 +104,15 @@ interface GetNextRankTextProps {
...
-  return rank === 0 ? rank : rank - 1;
+  if (rank < 0) {
+    return 0;
+  }
+
+  if (rank >= RANKS.length) {
+    return RANKS.length - 1;
+  }
+
+  return rank;
```
rebelchris, line 115: I think you need to do the initial calculation first, it needs to do the -1 from what I recall. (as it's based on previous rank technically)

```suggestion
const calculatedRank = rank === 0 ? rank : rank - 1
  if (calculatedRank < 0) {
    return 0;
  }

  if (calculatedRank >= RANKS.length) {
    return RANKS.length - 1;
  }

  return calculatedRank;
```

rebelchris (changes requested):
Small suggestion :)

packages/shared/src/lib/rank.ts
DragosIorgulescu: ah, great catch, updated 🙏

rebelchris (approved):
Yep this works :)

PR #2366, opened 2023-11-23
fix: encode the email passed in query params

DragosIorgulescu:
- covers more complex addresses such as alias email addresses

## Events

N / A

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done

### DIFF
--- a/packages/shared/src/lib/kratos.ts
+++ b/packages/shared/src/lib/kratos.ts
@@ -246,8 +246,9 @@ export const getKratosError = async (id: string): Promise<ErrorData> => {
 export const checkKratosEmail = async (
   email: string,
 ): Promise<KratosEmailData> => {
-  const params = new URLSearchParams({ email_address: email });
-  const res = await fetch(`${heimdallUrl}/api/check_email?${params}`, {
+  const url = new URL(`${heimdallUrl}/api/check_email`);
+  url.searchParams.append('email_address', email);
+  const res = await fetch(url.toString(), {
     method: 'POST',
   });
   return res.json();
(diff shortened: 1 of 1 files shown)


omBratteng (approved):

packages/shared/src/lib/kratos.ts
```diff
@@ -246,12 +246,10 @@ export const getKratosError = async (id: string): Promise<ErrorData> => {
...
-  const res = await fetch(
-    `${heimdallUrl}/api/check_email?email_address=${email}`,
-    {
-      method: 'POST',
-    },
-  );
+  const params = new URLSearchParams({ email });
+  const res = await fetch(`${heimdallUrl}/api/check_email?${params}`, {
+    method: 'POST',
+  });
```
capJavert, line 252: Why not use the whole `URL` interface? Like this if `params` was empty you would also get trailing `?`.

DragosIorgulescu: I think params can't be empty here, at most what can happen is `email` is empty, which would result in `/check_email?email=`

I'm open to updating this, though the end result would be the same?

capJavert: Actually worse, if `email` was undefined the result would be `/check_email?email=undefined`. But either way, don't see why not to use dedicated URL API to assemble the URL.

DragosIorgulescu: oh wow, you're right, damn JS man 😆 how is that a thing?

capJavert: Because URLSearchParams (and URL as well) calls `toString` when adding params through constructor or `params.append`. And `toString` of `undefined` is `"undefined"` 😆

sshanzel (approved):
Looks great! 🚀

PR #2368, opened 2023-11-23
feat: collections intro

sshanzel:
## Changes
- Collections intro about the feature.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1969 #done


packages/shared/src/components/post/widgets/CollectionsIntro.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+export function CollectionsIntro(): ReactElement {
+  const { checkHasCompleted, completeAction } = useActions();
+
+  if (checkHasCompleted(ActionType.CollectionsIntro)) {
+    return null;
+  }
+
+  return (
+    <span className="flex relative flex-row p-3 pr-2 w-full rounded-10 border border-theme-color-cabbage">
```
rebelchris, line 15: The rounded-10 we always use like this? Cant recall

capJavert:
@sshanzel we have epic branch `WT-1939-collections` so if you can point to that.

sshanzel:
> @sshanzel we have epic branch WT-1939-collections so if you can point to that.

I was receiving some build errors due to our cards list requiring every enum item to have an implementation. I can do that but I will have to add null values to those variables for which already exists on where this is pointing.

capJavert:
@sshanzel ok, I will merge it when that one is done. 👌

rebelchris:
@sshanzel Think you can rebase and get access to all the missing parts now.

packages/shared/src/components/Pill.tsx
```diff
@@ -34,7 +34,7 @@ const Pill = ({
     <div
       className={classNames(
         pillSizeToClassName[size],
-        'rounded-xl',
+        'rounded-12',
```
rebelchris, line 37: This should be 10px according to design which is `rounded-10` I think I asked dragos before if we had a named variant, but see the numbers are actually our overwrites.

packages/shared/src/components/post/PostWidgets.tsx
```diff
@@ -32,6 +34,7 @@ export function PostWidgets({
         className="hidden tablet:flex pt-6"
         contextMenuId="post-widgets-context"
       />
+      {post.type === PostType.Collection && <CollectionsIntro />}
```
rebelchris, line 37: Is there anyway we could inject this only in the collection modal for instance?

packages/shared/src/components/post/widgets/CollectionsIntro.tsx
```diff
@@ -0,0 +1,29 @@
...
+      <div className="flex flex-col flex-1 typo-callout">
```
rebelchris, line 16: I think it's subhead, but unsure as the names seem mixed.
Asking [here](https://dailydotdev.slack.com/archives/C0381BQUEN9/p1701157996730849).

rebelchris (changes requested):
Almost there, some minor things :)

packages/shared/src/components/Pill.tsx
sshanzel: I see, in this case, the original value was incorrect being `xl` (12px). Let me fix this.

packages/shared/src/components/post/PostWidgets.tsx
sshanzel: We don't have yet the modal/page for rendering collections. We can probably add it if we get a separate implementation of the PostContent for Collections.

rebelchris: Cool let's leave it for now, maybe can be a end refactor once we have that :)

packages/shared/src/components/post/widgets/CollectionsIntro.tsx
rebelchris: @sshanzel Please note new comment in the thread

rebelchris (approved):

packages/shared/src/components/Pill.tsx
sshanzel: Updated the value to 10 now 🥳

PR #2369, opened 2023-11-24
fix: tooltip request overload

sshanzel:
## Changes
We have a feature that when you hovers on a user profile image/name/handle, it would send a request to our server to fetch the user information. While doing so, every in and out of the cursor triggers the process, which means it would trigger the fetch again as well. By setting a stale time, we can avoid sending the request which data does not change very often. Users hardly update their user information once set, so setting a stale time of 30 minutes should be reasonable. Watch the network requests before and after on the section below: https://dailydotdev.atlassian.net/wiki/spaces/HAN/pages/691339518/Managing+GraphQL+loads#Profile-tooltip-sending-requests-on-every-hover

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done

### DIFF
--- a/packages/shared/src/hooks/useFeedSettings.ts
+++ b/packages/shared/src/hooks/useFeedSettings.ts
@@ -12,7 +12,7 @@ import AuthContext from '../contexts/AuthContext';
 import { graphqlUrl } from '../lib/config';
 import { LoggedUser } from '../lib/user';
 import { disabledRefetch } from '../lib/func';
-import { FEED_SETTINGS_STALE_TIME } from '../lib/query';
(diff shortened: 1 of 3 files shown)


omBratteng (approved):

rebelchris (approved):
Nice one! 👍
Also curious to hear @capJavert thoughts

capJavert (approved):
👌 we can do wonders with proper stale times, love to see it!

PR #2372, opened 2023-11-28
feat: Add a host permissions request page to the extension 

DragosIorgulescu:
## Changes

- first thing triggered when extension is loaded if host permissions are missing -- since without these nothing works
- don't trigger the boot request until we have host permissions enabled
- add an additional check for host permissions when using the features boundary HoC
- attempt to refactor extension functions into a context rather than query cache
  - currently causing errors when bundling the non-extension webapp due to imports triggering the inclusion of the `exteionScripts` file when bundling -- this needs to be resolved

## Preview

extension-permission-request

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1992 #done


packages/extension/package.json
```diff
@@ -1,13 +1,14 @@
...
-  "version": "3.28.2",
+  "version": "3.28.4",
   "scripts": {
     "dev:chrome": "cross-env NODE_ENV=development cross-env TARGET_BROWSER=chrome webpack --watch",
     "dev:firefox": "cross-env NODE_ENV=development cross-env TARGET_BROWSER=firefox webpack --watch",
     "dev:opera": "cross-env NODE_ENV=development cross-env TARGET_BROWSER=opera webpack --watch",
     "build:chrome": "cross-env NODE_ENV=production cross-env TARGET_BROWSER=chrome webpack",
     "build:firefox": "cross-env NODE_ENV=production cross-env TARGET_BROWSER=firefox webpack",
     "build:opera": "cross-env NODE_ENV=production cross-env TARGET_BROWSER=opera webpack",
+    "build:edge": "cross-env NODE_ENV=production cross-env TARGET_BROWSER=edge webpack",
```
rebelchris, line 11: Why was this added?

packages/extension/src/companion/CompanionPermission.tsx
```diff
@@ -4,18 +4,18 @@ import {
...
+import { ExtensionContext } from '@dailydotdev/shared/src/contexts/common';
 
 const CompanionSection = classed('div', 'flex flex-col max-w-full');
 
 const CompanionPermissionComponent = (
   _,
   ref: Ref<HTMLDivElement>,
 ): ReactElement => {
-  const { requestContentScripts } = useExtensionPermission();
+  const { requestContentScripts } = useContext(ExtensionContext);
```
rebelchris, line 18: Perhaps export as direct use hook, it's a common practice we recently started doing.
Example being: `useAnalyticsContext`

packages/extension/src/lib/extensionScripts.ts
```diff
@@ -1,16 +1,23 @@
 import browser from 'webextension-polyfill';
-import {
-  CreateRequestContentScripts,
-  contentScriptKey,
-} from '@dailydotdev/shared/src/hooks';
+import { contentScriptKey } from '@dailydotdev/shared/src/hooks';
 import { companionPermissionGrantedLink } from '@dailydotdev/shared/src/lib/constants';
-import { AnalyticsEvent } from '@dailydotdev/shared/src/lib/analytics';
+import { AnalyticsEvent as AnalyticsEventType } from '@dailydotdev/shared/src/lib/analytics';
```
rebelchris, line 4:
```suggestion
import { AnalyticsEvent as AnalyticsEventName } from '@dailydotdev/shared/src/lib/analytics';
```

Probably should have named these as such a long time ago 😮‍💨

packages/extension/src/lib/extensionScripts.ts
```diff
@@ -96,37 +103,12 @@ export const getHostPermission = (): Promise<boolean> =>
     origins: HOST_PERMISSIONS,
   });
 
-export const getHostPermissionAndRegister = async (): Promise<void> => {
+export const getHostPermissionAndRegister = async (): Promise<boolean> => {
```
rebelchris, line 106: Intentional we now expect a return here?

packages/extension/src/newtab/App.tsx
```diff
@@ -128,12 +109,16 @@ function InternalApp({
       : DEFAULT_TAB_TITLE;
   }, [unreadCount]);
 
-  console.log({ hostGranted });
+  if (!hostGranted) {
+    // eslint-disable-next-line no-param-reassign
+    pageRef.current = '/permissions';
```
rebelchris, line 114: Not really your mistake, but we should probably use forwardRef.
We can log it as seperate.
@capJavert What do you think?

packages/shared/src/components/ExtensionPermissionsPrompt.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+  const handleRequestHostPermissions = async () => {
+    const success = await requestHostPermissions({ origins });
+    if (success) {
+      onSuccess();
+    }
+  };
+
+  return (
+    <div className="flex overflow-hidden flex-col justify-center items-center px-7 antialiased text-center min-h-[100vh] max-h-[100vh]">
```
rebelchris, line 23: Why don't we use `min-h-screen` here?
Also not sure why the px-7 is needed, mobile?

packages/shared/src/components/withFeaturesBoundary.tsx
```diff
@@ -1,12 +1,20 @@
...
   options?: { fallback?: ReactElement },
 ): ComponentType<Props> => {
   const WithFeaturesBoundary = (props: Props): ReactElement => {
     const { ready: areFeaturesReady } = useContext(FeaturesReadyContext);
+    const { hostGranted } = useHostStatus();
+    const isExtension = checkIsExtension();
+
+    if (isExtension && !hostGranted) {
+      return <WrappedComponent {...props} />;
+    }
```
rebelchris, line 17: Is this needed?
Just not quite sure why we need it in this spot already

packages/shared/src/contexts/BootProvider.tsx
```diff
@@ -96,8 +102,9 @@ export const BootDataProvider = ({
     isFetched,
     dataUpdatedAt,
   } = useQuery(BOOT_QUERY_KEY, () => getBootData(app), {
-    refetchOnWindowFocus: loggedUser,
+    refetchOnWindowFocus: loggedUser && hostGranted,
```
rebelchris, line 105: This shouldn't be needed if enabled already disables it

packages/shared/src/lib/boot.ts
```diff
@@ -80,6 +80,8 @@ export async function getBootData(app: string, url?: string): Promise<Boot> {
   if (url) {
     params.append('url', url);
   }
+
+  console.log('CALLED');
```
rebelchris, line 84: This should be removed

packages/shared/src/contexts/common.ts
```diff
@@ -1,2 +1,20 @@
...
+
+export interface ExtensionContextData {
+  getContentScriptPermission?: () => Promise<boolean>;
+  requestContentScripts?: RequestContentScripts;
+  registerBrowserContentScripts?: () => Promise<void>;
+  getHostPermission?: () => Promise<boolean>;
+  requestHostPermissions?: Permissions.Static['request'];
+  origins?: string[];
+}
+export const ExtensionContext = React.createContext<ExtensionContextData>({});
```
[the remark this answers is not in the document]
rebelchris: I'd love to here other opinions here if this is a practive we want to follow.
My concern would be it could be harder to see what actually gets used/not.

CC: @dailydotdev/web-team

packages/extension/src/newtab/App.tsx
capJavert: Yeah but not crucial, agree it can be done separate 👍

rebelchris: Will log it separately :)

packages/shared/src/contexts/common.ts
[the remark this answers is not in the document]
DragosIorgulescu: one potential argument for this is that we would no longer depend on the `QueryClient` provider being set first in order to use these -- which was the main reason I made the change, since we have a situation here where we need to make calls to these scripts before anything else.

packages/extension/src/lib/extensionScripts.ts
DragosIorgulescu: yes, we want to know in case the user clicked `deny` on this pop-up (which would result in a `false` returned by the fn call`

packages/shared/src/components/ExtensionPermissionsPrompt.tsx
DragosIorgulescu: I picked up the same classes that were used in the [ExtensionOnboarding](https://github.com/dailydotdev/apps/blob/1f19c3434f66f81d2e11a2c2ecdc94a3cb187301/packages/shared/src/components/ExtensionOnboarding.tsx#L11) component, I'll double check these and clean them up 👍

packages/shared/src/contexts/common.ts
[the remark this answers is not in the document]
rebelchris: Yeah quite like the context overall, just the way in which some cases it might be emty/some it might be set might not be super clear.
Also perhaps we should consider having it as a proper context file in shared?
And then the one in extension will extend/override the data?

idoshamun:
@DragosIorgulescu don't wait for my review on this 🙏 
If there are specifics you'd like to me to look at just let me know

packages/shared/src/contexts/common.ts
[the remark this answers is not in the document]
DragosIorgulescu: fair point, it is actually always empty outside the extension (the provider isn't even mounted in the [webapp](https://github.com/dailydotdev/apps/blob/65b6c8cd36a5a45247682a1dc273caa620e54bf6/packages/webapp/pages/_app.tsx#L184-L202)

But I agree with your point that the way the `Context` and `Provider` are split here can be confusing -- I extracted it out into it's own `ExtensionContext` file in shared, as you suggested, which is then overridden in the extension

packages/shared/src/components/withFeaturesBoundary.tsx
DragosIorgulescu: Actually, I think there is a better way to do this by just using the fallback and not worrying about extension or host permissions inside this wrapper

packages/shared/src/components/ExtensionPermissionsPrompt.tsx
DragosIorgulescu: Tested `min-h-screen` and it does not look like it works properly

packages/shared/src/components/ExtensionOnboarding.tsx
```diff
@@ -26,16 +22,15 @@ const ExtensionOnboarding = ({ onContinue }: Props): ReactElement => {
...
 
-      {/* <Link href={onboardingUrl} passHref> */}
-      <Button
-        // tag="a"
-        className="z-1 w-full btn-primary max-w-[18.75rem]"
-        buttonSize={ButtonSize.Large}
-        onClick={() => onContinue(onboardingUrl)}
-      >
-        Continue ➔
-      </Button>
```
DragosIorgulescu, line 37: Need to double check that this was not doing some extra things on `onContinue` which made the `Link` usage not an option

packages/shared/src/components/ExtensionPermissionsPrompt.tsx
rebelchris: Let's leave it for now. perhaps we override something.

rebelchris (approved):
Nice solution with the context!
Code wise looks good to me, let's wait for 1 more review from @dailydotdev/web-team before we merge.

packages/extension/src/contexts/ExtensionContext.tsx
```diff
@@ -0,0 +1,37 @@
...
+  const client = useQueryClient();
+  const { trackEvent } = useContext(AnalyticsContext);
+
+  const contextData = {
+    requestContentScripts: requestContentScripts?.(client, trackEvent),
+    getContentScriptPermission,
+    getHostPermission,
+    requestHostPermissions: browser.permissions.request,
+    origins: HOST_PERMISSIONS,
+  };
```
capJavert, line 30: Memoize if you can to avoid re-renders of subscribed components.

capJavert (approved):
Code wise looks fine, just a single comment for context memoization.

packages/shared/src/components/ExtensionOnboarding.tsx
DragosIorgulescu: checked, this was just a temporary change I made. `Link` is the way it was originally here.

PR #2377, opened 2023-11-29
#1000 fix BUG: on mobile the header is missing the app's logo

whovishnu:
Hi,
Fix App's Logo Missing in Mobile.

### DIFF
--- a/packages/shared/src/components/MainLayout.tsx
+++ b/packages/shared/src/components/MainLayout.tsx
@@ -55,11 +55,9 @@ const feeds = Object.values(SharedFeedPage);
 
 function MainLayout({
   children,
-  showOnlyLogo,
   greeting,
   activePage,
   isNavItemsButton,
-  mobileTitle,
   showDnd,
   dndActive,
   customBanner,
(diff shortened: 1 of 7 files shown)


rebelchris (changes requested):
Thanks for the adjustments. 🥳 

It looks like with these changes we actually don't use `mobileTitle` anymore, I think we can safely remove it.
Perhaps also refactor the `showOnlyLogo`? (not sure if we only use it sometimes)

whovishnu:
@rebelchris welcome,
so i will remove `mobileTitle` and by default will be  `showOnlyLogo`.

rebelchris (approved):
Thank you so much,
looks good to me, just waiting for @sshanzel to review as he manages the OS contributions :)

sshanzel (approved):
Wow! Thank you so much for the fix and the cleanup itself. I always love when the deleted code is greater than the new ones. Well done indeed 👏

sshanzel:
@rebelchris have verified on the preview, it works as intended 🥳

PR #2370, opened 2023-11-25
Wt 1968 post page modal collections

nensidosari:
### Describe what this PR does
- Short and concise, bullet points can help
- Screenshots if applicable can also help

## Events

Did you introduce any new tracking events?
Don't forget to update the [Analytics Taxonomy sheet](https://docs.google.com/spreadsheets/d/18Lv7zXges9QfVX5VYL1a-Hyl0e1sQ3sLr0OK8YZWKXI/edit#gid=0)

Log the new/changed events below:

| Type   | event_name  | value |
|--------|-------------|-------|
| Change/New | event name  | extra: { ... } |

### On those affected packages:
- [ ] Have you done sanity checks in the webapp?
- [ ] Have you done sanity checks in the extension?
- [ ] Does this not break anything in companion?

### Did you test the modified components media queries?
- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

#### Did you test on actual mobile devices?
- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done

PR #2378, opened 2023-11-30
feat: collection update notification type support

DragosIorgulescu:
## Changes

- Added support for the `collection_updated` notification type
- Using the profile picture group to display notification avatars

image

## Events

N / A

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1970 #done


DragosIorgulescu:
@capJavert I wasn't sure how to generate some collection notifications to test with here, do you have an easy way to do this so I can verify the UI here?

packages/shared/src/components/notifications/NotificationItem.tsx
```diff
@@ -49,17 +50,32 @@ function NotificationItem({
...
-      ?.map?.((avatar) => (
-        <NotificationItemAvatar
-          key={avatar.referenceId}
-          className="z-1"
-          {...avatar}
-        />
-      ))
-      .filter((avatar) => avatar) ?? [];
-  const hasAvatar = avatarComponents.length > 0;
+    type === NotificationType.CollectionUpdated ? (
```
rebelchris, line 56: Is there anyway we can uniform the two to use the ProfilePictureGroup?
(Think only difference is basically the rounded state.

rebelchris (changes requested):
One request if it's possible? 🙏

packages/shared/src/components/notifications/NotificationItem.tsx
rebelchris: Hmm I see the spacing is also overlapping for collecitons which is not for the others.

rebelchris (approved):
Looks good to me on second though :)

packages/shared/src/components/notifications/NotificationItem.tsx
DragosIorgulescu: good point, this is already modified on the collection post page branch though: https://github.com/dailydotdev/apps/pull/2370/files#diff-1dbb3736e496fd2b9830576074141b705cc949e7eef12d17eb6c0cb5c77211e9

I'll consolidate with those changes here as well

capJavert:
@DragosIorgulescu should be possible to insert some for your user. I can help with that once we merge https://github.com/dailydotdev/daily-api/pull/1558

sshanzel (approved):
Looks good to me! :shipit:

PR #2381, opened 2023-12-01
feat: new button implementation

kopancek:
- new Button implementation, [based on DR](https://dailydotdev.atlassian.net/wiki/spaces/HAN/pages/696221697/WT-2017+Enhance+Button+component)
- all look and feel stays the same as the original button
- added storybook
- New component is used in `InviteLinkInput` and `ReferralSocialShareButtons` and `ProfileLayout`. 

## Related issue
- [WT-2017](https://dailydotdev.atlassian.net/browse/WT-2017)

## Events

No changes

## Screenshots

Screenshot 2023-12-08 at 12 31 28
Screenshot 2023-12-08 at 13 41 19
Screenshot 2023-12-08 at 12 35 37
Screenshot 2023-12-08 at 12 35 43

## Preview deployment

New button can be seen in action on the user profile page on [preview3](https://preview3.app.daily.dev/kopancek.) You need to modify the following URL to land on your own user profile: `https://preview3.app.daily.dev/YOUR_USER_NAME` (otherwise the link from the user dropdown menu points to app.daily.dev instead).

## Running storybook locally
```
cd packages/storybook
pnpm i
npm run dev
```
And open your browser on http://localhost:6006

## TODO
- [x] Make sure the icons work

- [x] MobileL (420px)
- [x] Tablet (656px)
- [x] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [x] Android

[WT-2017]: https://dailydotdev.atlassian.net/browse/WT-2017?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


kopancek:
This is a diff of the 2 button implementations, for easier understanding of what was actually changed so far:
```diff
diff -y packages/shared/src/components/buttons/Button.tsx packages/shared/src/components/buttons/ButtonV2.tsx
import React, {                                                         import React, {
  forwardRef,                                                             forwardRef,
  HTMLAttributes,                                                         HTMLAttributes,
  ReactElement,                                                           ReactElement,
  ReactNode,                                                              ReactNode,
  Ref,                                                                    Ref,
} from 'react';                                                         } from 'react';
import classNames from 'classnames';                                    import classNames from 'classnames';
import { IconProps, IconSize } from '../Icon';                          import { IconProps, IconSize } from '../Icon';
import { Loader } from '../Loader';                                     import { Loader } from '../Loader';
import { combinedClicks } from '../../lib/click';                       import { combinedClicks } from '../../lib/click';

                                                                >       import { ColorName } from '../../styles/colors';
                                                                >
export enum ButtonSize {                                                export enum ButtonSize {
  XXSmall = 'xxsmall',                                                    XXSmall = 'xxsmall',
  XSmall = 'xsmall',                                                      XSmall = 'xsmall',
  Small = 'small',                                                        Small = 'small',
  Medium = 'medium',                                                      Medium = 'medium',
  Large = 'large',                                                        Large = 'large',
  XLarge = 'xlarge',                                                      XLarge = 'xlarge',
}                                                                       }

const buttonSizeToIconSize: Record<ButtonSize, IconSize>                const buttonSizeToIconSize: Record<ButtonSize, IconSize>
  [ButtonSize.XXSmall]: IconSize.XXSmall,                                 [ButtonSize.XXSmall]: IconSize.XXSmall,
  [ButtonSize.XSmall]: IconSize.XSmall,                                   [ButtonSize.XSmall]: IconSize.XSmall,
  [ButtonSize.Small]: IconSize.Small,                                     [ButtonSize.Small]: IconSize.Small,
  [ButtonSize.Medium]: IconSize.Medium,                                   [ButtonSize.Medium]: IconSize.Medium,
  [ButtonSize.Large]: IconSize.Large,                                     [ButtonSize.Large]: IconSize.Large,
  [ButtonSize.XLarge]: IconSize.XLarge,                                   [ButtonSize.XLarge]: IconSize.XLarge,
};                                                                      };

export interface StyledButtonProps {                            |       export type IconType = React.ReactElement<IconProps>;
  buttonSize?: ButtonSize;                                      |
  iconOnly?: boolean;                                           |       export type ButtonColor = ColorName;
                                                                >
                                                                >       export enum ButtonKind {
                                                                >         Button = 'button',
                                                                >         Link = 'link',
}                                                                       }

export type IconType = React.ReactElement<IconProps>;           |       export enum ButtonVariant {
                                                                >         Primary = 'primary',
                                                                >         Secondary = 'secondary',
                                                                >         Tertiary = 'tertiary',
                                                                >       }

export interface BaseButtonProps {                                      export interface BaseButtonProps {
  buttonSize?: ButtonSize;                                      |         kind?: ButtonKind;
  loading?: boolean;                                            |         variant?: ButtonVariant;
  pressed?: boolean;                                            |         size?: ButtonSize;
  tag?: React.ElementType;                                      |         color?: ButtonColor;
  icon?: IconType;                                                        icon?: IconType;
  rightIcon?: IconType;                                                   rightIcon?: IconType;
  children?: ReactNode;                                         |         loading?: boolean;
  displayClass?: string;                                        |         pressed?: boolean;
  textPosition?: string;                                        <
  position?: string;                                            <
  disabled?: boolean;                                                     disabled?: boolean;
  spanClassName?: string;                                       |         children?: ReactNode;
}                                                                       }

const useGetIconWithSize = (size: ButtonSize, iconOnly: b               const useGetIconWithSize = (size: ButtonSize, iconOnly: b
  return (icon: React.ReactElement<IconProps>) =>                         return (icon: React.ReactElement<IconProps>) =>
    React.cloneElement(icon, {                                              React.cloneElement(icon, {
      size: icon.props?.size ?? buttonSizeToIconSize[size                     size: icon.props?.size ?? buttonSizeToIconSize[size
      className: classNames(icon.props.className, !iconOn                     className: classNames(icon.props.className, !iconOn
    });                                                                     });
};                                                                      };

export type AllowedTags = keyof Pick<JSX.IntrinsicElement               export type AllowedTags = keyof Pick<JSX.IntrinsicElement
export type AllowedElements = HTMLButtonElement | HTMLAnc               export type AllowedElements = HTMLButtonElement | HTMLAnc
export type ButtonElementType<Tag extends AllowedTags> =                export type ButtonElementType<Tag extends AllowedTags> =
  ? HTMLAnchorElement                                                     ? HTMLAnchorElement
  : HTMLButtonElement;                                                    : HTMLButtonElement;

export type ButtonProps<Tag extends AllowedTags> = BaseBu               export type ButtonProps<Tag extends AllowedTags> = BaseBu
  HTMLAttributes<AllowedElements> &                             |         HTMLAttributes<Tag> &
  JSX.IntrinsicElements[Tag] & {                                          JSX.IntrinsicElements[Tag] & {
    ref?: Ref<ButtonElementType<Tag>>;                                      ref?: Ref<ButtonElementType<Tag>>;
    readOnly?: boolean;                                         <
  };                                                                      };

function ButtonComponent<TagName extends AllowedTags>(          |       function ButtonComponent<T extends AllowedTags>(
  {                                                                       {
    loading,                                                    |           kind = ButtonKind.Button,
    pressed,                                                    |           variant,
                                                                >           size,
                                                                >           color,
                                                                >           className,
    icon,                                                                   icon,
    rightIcon,                                                              rightIcon,
    buttonSize = ButtonSize.Medium,                             |           loading,
                                                                >           pressed,
    children,                                                               children,
    tag: Tag = 'button',                                        <
    className,                                                  <
    displayClass,                                               <
    position = 'relative',                                      <
    textPosition = 'justify-center',                            <
    readOnly,                                                   <
    iconOnly: showIconOnly,                                     <
    onClick,                                                                onClick,
    spanClassName,                                              <
    ...props                                                                ...props
  }: StyledButtonProps & ButtonProps<TagName>,                  |         }: ButtonProps<T>,
  ref?: Ref<ButtonElementType<TagName>>,                        |         ref?: Ref<ButtonElementType<T>>,
): ReactElement {                                                       ): ReactElement {
  const iconOnly = (icon && !children && !rightIcon) || s       |         const iconOnly = icon && !children && !rightIcon;
  const getIconWithSize = useGetIconWithSize(buttonSize,        |         const getIconWithSize = useGetIconWithSize(size, iconOn
  const isAnchor = Tag === 'a';                                 |         const isAnchor = kind === ButtonKind.Link;
                                                                >         const Tag = isAnchor ? 'a' : 'button';

  return (                                                                return (
    <Tag                                                                    <Tag
      {...(props as StyledButtonProps)}                         |             {...props}
      {...(isAnchor ? combinedClicks(onClick) : { onClick                     {...(isAnchor ? combinedClicks(onClick) : { onClick
      aria-busy={loading}                                                     aria-busy={loading}
      aria-pressed={pressed}                                                  aria-pressed={pressed}
      ref={ref}                                                               ref={ref}
      className={classNames(                                                  className={classNames(
        { iconOnly, readOnly },                                 |               { iconOnly },
        buttonSize,                                             |               size,
        'btn flex-row items-center border typo-callout fo       |               'btn flex flex-row items-center justify-center bo
        textPosition,                                           |               {
        displayClass || 'flex',                                 |                 [`btn-${variant}`]: variant,
        position,                                               |                 [`btn-${variant}-${color}`]: color && variant,
                                                                >               },
        className,                                                              className,
      )}                                                                      )}
    >                                                                       >
      {icon && getIconWithSize(icon)}                                         {icon && getIconWithSize(icon)}
      {children && <span className={spanClassName}>{child       |             {children && <span>{children}</span>}
      {rightIcon && getIconWithSize(rightIcon)}                               {rightIcon && getIconWithSize(rightIcon)}
      {loading && (                                                           {loading && (
        <Loader                                                                 <Loader
          data-testid="buttonLoader"                                              data-testid="buttonLoader"
          className="hidden absolute top-0 right-0 bottom                         className="hidden absolute top-0 right-0 bottom
        />                                                                      />
      )}                                                                      )}
    </Tag>                                                                  </Tag>
  );                                                                      );
}                                                                       }

export const Button = forwardRef(ButtonComponent);                      export const Button = forwardRef(ButtonComponent);
```

kopancek:
Updated diff after the color prop was changed to an enum as well: 
```diff
diff -y packages/shared/src/components/buttons/Button.tsx packages/shared/src/components/buttons/ButtonV2.tsx
import React, {                                                         import React, {
  forwardRef,                                                             forwardRef,
  HTMLAttributes,                                                         HTMLAttributes,
  ReactElement,                                                           ReactElement,
  ReactNode,                                                              ReactNode,
  Ref,                                                                    Ref,
} from 'react';                                                         } from 'react';
import classNames from 'classnames';                                    import classNames from 'classnames';
import { IconProps, IconSize } from '../Icon';                          import { IconProps, IconSize } from '../Icon';
import { Loader } from '../Loader';                                     import { Loader } from '../Loader';
import { combinedClicks } from '../../lib/click';                       import { combinedClicks } from '../../lib/click';

                                                                >       import { Colors as ButtonColor } from '../../styles/color
                                                                >
export enum ButtonSize {                                                export enum ButtonSize {
  XXSmall = 'xxsmall',                                                    XXSmall = 'xxsmall',
  XSmall = 'xsmall',                                                      XSmall = 'xsmall',
  Small = 'small',                                                        Small = 'small',
  Medium = 'medium',                                                      Medium = 'medium',
  Large = 'large',                                                        Large = 'large',
  XLarge = 'xlarge',                                                      XLarge = 'xlarge',
}                                                                       }

const buttonSizeToIconSize: Record<ButtonSize, IconSize>                const buttonSizeToIconSize: Record<ButtonSize, IconSize>
  [ButtonSize.XXSmall]: IconSize.XXSmall,                                 [ButtonSize.XXSmall]: IconSize.XXSmall,
  [ButtonSize.XSmall]: IconSize.XSmall,                                   [ButtonSize.XSmall]: IconSize.XSmall,
  [ButtonSize.Small]: IconSize.Small,                                     [ButtonSize.Small]: IconSize.Small,
  [ButtonSize.Medium]: IconSize.Medium,                                   [ButtonSize.Medium]: IconSize.Medium,
  [ButtonSize.Large]: IconSize.Large,                                     [ButtonSize.Large]: IconSize.Large,
  [ButtonSize.XLarge]: IconSize.XLarge,                                   [ButtonSize.XLarge]: IconSize.XLarge,
};                                                                      };

export interface StyledButtonProps {                            |       export type IconType = React.ReactElement<IconProps>;
  buttonSize?: ButtonSize;                                      |
  iconOnly?: boolean;                                           |       export { ButtonColor };
                                                                >
                                                                >       export enum ButtonKind {
                                                                >         Button = 'button',
                                                                >         Link = 'link',
}                                                                       }

export type IconType = React.ReactElement<IconProps>;           |       export enum ButtonVariant {
                                                                >         Primary = 'primary',
                                                                >         Secondary = 'secondary',
                                                                >         Tertiary = 'tertiary',
                                                                >       }

export interface BaseButtonProps {                                      export interface BaseButtonProps {
  buttonSize?: ButtonSize;                                      |         kind?: ButtonKind;
  loading?: boolean;                                            |         variant?: ButtonVariant;
  pressed?: boolean;                                            |         size?: ButtonSize;
  tag?: React.ElementType;                                      |         color?: ButtonColor;
  icon?: IconType;                                                        icon?: IconType;
  rightIcon?: IconType;                                                   rightIcon?: IconType;
  children?: ReactNode;                                         |         loading?: boolean;
  displayClass?: string;                                        |         pressed?: boolean;
  textPosition?: string;                                        <
  position?: string;                                            <
  disabled?: boolean;                                                     disabled?: boolean;
  spanClassName?: string;                                       |         children?: ReactNode;
}                                                                       }

const useGetIconWithSize = (size: ButtonSize, iconOnly: b               const useGetIconWithSize = (size: ButtonSize, iconOnly: b
  return (icon: React.ReactElement<IconProps>) =>                         return (icon: React.ReactElement<IconProps>) =>
    React.cloneElement(icon, {                                              React.cloneElement(icon, {
      size: icon.props?.size ?? buttonSizeToIconSize[size                     size: icon.props?.size ?? buttonSizeToIconSize[size
      className: classNames(icon.props.className, !iconOn                     className: classNames(icon.props.className, !iconOn
    });                                                                     });
};                                                                      };

export type AllowedTags = keyof Pick<JSX.IntrinsicElement               export type AllowedTags = keyof Pick<JSX.IntrinsicElement
export type AllowedElements = HTMLButtonElement | HTMLAnc               export type AllowedElements = HTMLButtonElement | HTMLAnc
export type ButtonElementType<Tag extends AllowedTags> =                export type ButtonElementType<Tag extends AllowedTags> =
  ? HTMLAnchorElement                                                     ? HTMLAnchorElement
  : HTMLButtonElement;                                                    : HTMLButtonElement;

export type ButtonProps<Tag extends AllowedTags> = BaseBu               export type ButtonProps<Tag extends AllowedTags> = BaseBu
  HTMLAttributes<AllowedElements> &                             |         HTMLAttributes<Tag> &
  JSX.IntrinsicElements[Tag] & {                                          JSX.IntrinsicElements[Tag] & {
    ref?: Ref<ButtonElementType<Tag>>;                                      ref?: Ref<ButtonElementType<Tag>>;
    readOnly?: boolean;                                         <
  };                                                                      };

function ButtonComponent<TagName extends AllowedTags>(          |       function ButtonComponent<T extends AllowedTags>(
  {                                                                       {
    loading,                                                    |           kind = ButtonKind.Button,
    pressed,                                                    |           variant,
                                                                >           size,
                                                                >           color,
                                                                >           className,
    icon,                                                                   icon,
    rightIcon,                                                              rightIcon,
    buttonSize = ButtonSize.Medium,                             |           loading,
                                                                >           pressed,
    children,                                                               children,
    tag: Tag = 'button',                                        <
    className,                                                  <
    displayClass,                                               <
    position = 'relative',                                      <
    textPosition = 'justify-center',                            <
    readOnly,                                                   <
    iconOnly: showIconOnly,                                     <
    onClick,                                                                onClick,
    spanClassName,                                              <
    ...props                                                                ...props
  }: StyledButtonProps & ButtonProps<TagName>,                  |         }: ButtonProps<T>,
  ref?: Ref<ButtonElementType<TagName>>,                        |         ref?: Ref<ButtonElementType<T>>,
): ReactElement {                                                       ): ReactElement {
  const iconOnly = (icon && !children && !rightIcon) || s       |         const iconOnly = icon && !children && !rightIcon;
  const getIconWithSize = useGetIconWithSize(buttonSize,        |         const getIconWithSize = useGetIconWithSize(size, iconOn
  const isAnchor = Tag === 'a';                                 |         const isAnchor = kind === ButtonKind.Link;
                                                                >         const Tag = isAnchor ? 'a' : 'button';

  return (                                                                return (
    <Tag                                                                    <Tag
      {...(props as StyledButtonProps)}                         |             {...props}
      {...(isAnchor ? combinedClicks(onClick) : { onClick                     {...(isAnchor ? combinedClicks(onClick) : { onClick
      aria-busy={loading}                                                     aria-busy={loading}
      aria-pressed={pressed}                                                  aria-pressed={pressed}
      ref={ref}                                                               ref={ref}
      className={classNames(                                                  className={classNames(
        { iconOnly, readOnly },                                 |               { iconOnly },
        buttonSize,                                             |               size,
        'btn flex-row items-center border typo-callout fo       |               'btn flex flex-row items-center justify-center bo
        textPosition,                                           |               {
        displayClass || 'flex',                                 |                 [`btn-${variant}`]: variant,
        position,                                               |                 [`btn-${variant}-${color}`]: color && variant,
                                                                >               },
        className,                                                              className,
      )}                                                                      )}
    >                                                                       >
      {icon && getIconWithSize(icon)}                                         {icon && getIconWithSize(icon)}
      {children && <span className={spanClassName}>{child       |             {children && <span>{children}</span>}
      {rightIcon && getIconWithSize(rightIcon)}                               {rightIcon && getIconWithSize(rightIcon)}
      {loading && (                                                           {loading && (
        <Loader                                                                 <Loader
          data-testid="buttonLoader"                                              data-testid="buttonLoader"
          className="hidden absolute top-0 right-0 bottom                         className="hidden absolute top-0 right-0 bottom
        />                                                                      />
      )}                                                                      )}
    </Tag>                                                                  </Tag>
  );                                                                      );
}                                                                       }

export const Button = forwardRef(ButtonComponent);                      export const Button = forwardRef(ButtonComponent);
```

packages/storybook/stories/ButtonV2.stories.tsx
```diff
@@ -0,0 +1,45 @@
...
+      control: "object",
+    },
+  },
+};
+
+export default meta;
+
+type Story = StoryObj<typeof Button>;
+
+export const Basic: Story = {
```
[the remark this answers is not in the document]
nensidosari: Yeah that for sure, I was more on having this as an additional thing rather then replacement of what we have now. Or we might want to keep in docs all the components rendered and then in the story have it customisable. It would need some work though this was since at this moment we are using auto generation.

packages/storybook/stories/button.stories.tsx
```diff
@@ -1,40 +0,0 @@
-import type { Meta, StoryObj } from "@storybook/react";
```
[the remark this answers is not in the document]
kopancek: The story is now only for ButtonV2. I can keep the old one intact, but it was not doing much...

packages/shared/src/components/buttons/ButtonV2.tsx
```diff
@@ -0,0 +1,134 @@
...
+
+export enum ButtonVariant {
+  Primary = 'primary',
+  Secondary = 'secondary',
+  Tertiary = 'tertiary',
+}
+
+export interface BaseButtonProps {
+  kind?: ButtonKind;
+  variant?: ButtonVariant;
```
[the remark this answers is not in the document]
kopancek: The old button supports not providing a variant/color/size, so I kept the same interface. 

I need to figure out, if we actually use the button without the variant somewhere, maybe it does not make sense to not provide a button variant 🤔 

Not providing color should be supported, as colors only override the default button theme color. 

Size defaults to medium in CSS, we can as well make it the default for the prop itself 👍

packages/shared/src/components/buttons/ButtonV2.spec.tsx
```diff
@@ -0,0 +1,120 @@
...
+} from './ButtonV2';
+import UpvoteIcon from '../icons/Upvote';
+
+const renderComponent = <Tag extends AllowedTags>(
+  props: Partial<ButtonProps<Tag>> = {},
+): RenderResult => {
+  return render(<Button {...props} />);
+};
+
+it('should render children', async () => {
```
sshanzel, line 20: Should we maybe wrap these sections in a `describe`? Just thinking of making it easier to isolate related tests. Not required just a thought.

For example we will have a `describe('button variants', () => { // tests go here })`

packages/shared/src/components/buttons/ButtonV2.tsx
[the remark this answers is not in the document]
sshanzel: We always have used the Button with a variant. We can maybe use `primary` for the default. We used to have a `default` variant (without `btn-` class) but we never really used it.

packages/shared/src/styles/colors.ts
```diff
@@ -1,20 +1,23 @@
...
+  Onion = 'onion',
+  Water = 'water',
+  Salt = 'salt',
+  Pepper = 'pepper',
+  Twitter = 'twitter',
+  Whatsapp = 'whatsapp',
+  Facebook = 'facebook',
+}
+
+export type ColorName = Uncapitalize<keyof typeof Colors>;
```
sshanzel, line 20: We don't need this anymore I believe, on line 23 we can just `Record<Colors, Color>`. We can even retain the `ColorName` on the enum part.

packages/shared/src/components/buttons/ButtonV2.tsx
```diff
@@ -0,0 +1,150 @@
...
+          [`btn-${variant}`]: variant,
+          [`btn-${variant}-${color}`]: color && variant,
+        },
+        className,
+      )}
+    >
+      {icon &&
+        iconPosition === ButtonIconPosition.Left &&
+        getIconWithSize(icon)}
+      {children && <span>{children}</span>}
```
sshanzel, line 136: I thought we were supposed to not use the `span` from within the Button component reading the DR.

packages/shared/src/components/buttons/ButtonV2.tsx
```diff
@@ -0,0 +1,150 @@
...
+      {icon &&
+        iconPosition === ButtonIconPosition.Right &&
+        getIconWithSize(icon)}
+      {loading && (
```
sshanzel, line 140: Can you check the behavior of our loading state here whether it renders the content still or just the spinner when it is true.

sshanzel (commented):
Just some few minor initial comments but great start nonetheless! 🚀

packages/shared/src/components/buttons/ButtonV2.tsx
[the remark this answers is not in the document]
kopancek: The problem is for example QuarternaryButton, which is using the button without a variant. So it's not true that we always provide the variant. Maybe there's also another use case that I didn't catch already.

packages/shared/src/components/buttons/ButtonV2.spec.tsx
kopancek: I tried, please take a look again? :)

packages/shared/src/styles/colors.ts
kopancek: Removed, thanks for pointing that out 🙇

packages/shared/src/components/buttons/ButtonV2.tsx
kopancek: There are some style definitions in button.css that rely on the button component to have a child component. Otherwise the box shadow was not applied correctly for example. There is another solution for this of course, but this seemed like the least amount of change.

The DR itself is talking about removing the `spanClassName` prop, not about removing the span itself.

packages/shared/src/components/buttons/ButtonV2.tsx
kopancek: It renders just the spinner, which is the same behavior as old button. Do we want to change it?

packages/storybook/stories/button.stories.tsx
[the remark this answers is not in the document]
rebelchris: I'm ok with removing it, it was a placeholder at best :)

packages/storybook/stories/ButtonV2.stories.tsx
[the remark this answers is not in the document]
rebelchris: Great suggestion for a follow up task for someone when they are completely stuck :)

packages/storybook/.storybook/preview.ts
```diff
@@ -1,19 +1,18 @@
-import { Preview, ReactRenderer } from '@storybook/react';
-import { withThemeByClassName } from '@storybook/addon-themes';
+import { Preview, ReactRenderer } from "@storybook/react";
+import { withThemeByClassName } from "@storybook/addon-themes";
 
-import '@dailydotdev/shared/src/styles/globals.css';
+import "@dailydotdev/shared/src/styles/globals.css";
```
rebelchris, line 4: I think our linting is not set right on this mono, it should prefer single quotes actually. following the rest of the monorepo.

rebelchris (commented):
Nice work so far.
Just early feedback/request:

Could you potentially perhaps add the button in 1 space instead of the default one and perhaps point it to a preview2/3 domain so we can validate it?

packages/shared/src/components/buttons/ButtonV2.tsx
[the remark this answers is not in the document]
nensidosari: Maybe that quaternary button is a variant of it's own?

packages/storybook/.storybook/preview.ts
kopancek: Found the issue - it was bad `package.json` file in the storybook directory. Should be fixed now 👍

packages/shared/src/components/buttons/ButtonV2.tsx
```diff
@@ -0,0 +1,152 @@
...
+      aria-pressed={pressed}
+      ref={ref}
+      className={classNames(
+        { iconOnly },
+        size,
+        'btn flex flex-row items-center justify-center border typo-callout font-bold no-underline shadow-none cursor-pointer select-none focus-outline relative',
+        {
+          [`btn-${variant}`]: variant,
+          [`btn-${variant}-${color}`]: color && variant,
+        },
```
rebelchris, line 131: Be aware this won't work.
[Tailwind](https://stackoverflow.com/a/76193787/1719919) doesn't allow mixed variables, you should always use fully qualified names.

Thus I think we need some mapping function that based on variant-color can return the final `btn-` class.

rebelchris (changes requested):

packages/shared/src/components/buttons/ButtonV2.tsx
kopancek: I refactored it anyway, so that the code uses similar conventions as for button sizes. Please take a look again 👍

packages/shared/src/components/buttons/ButtonV2.tsx
```diff
@@ -0,0 +1,193 @@
...
+export type IconType = React.ReactElement<IconProps>;
+
+export { ButtonColor };
+
+export enum ButtonVariant {
+  Primary = 'primary',
+  Secondary = 'secondary',
+  Tertiary = 'tertiary',
+  Float = 'tertiaryFloat',
+}
```
rebelchris, line 40: We currently also have `btn-quaternary` as a class what should happen to that one?

packages/shared/src/components/buttons/ButtonV2.tsx
```diff
@@ -0,0 +1,193 @@
...
+  [ButtonSize.Small]: 'h-8 w-8 p-0 rounded-[0.63rem]',
+  [ButtonSize.XSmall]: 'h-6 w-6 p-0 rounded-lg',
+};
+
+const variantToClassName: Record<ButtonVariant, string> = Object.values(
+  ButtonVariant,
+).reduce((acc, variant) => {
+  acc[variant] = `btn-${variant}`;
+  return acc;
+}, {} as Record<ButtonVariant, string>);
```
rebelchris, line 119: I'm still not a big fan of this, it goes against tailwinds own guidelines and we generally always avoid it.
It's maybe better to consider a mapping array? (I think that's the proposal here, but it's still concatenated)

Although mostly it should probably work, tailwind might also change the JIT compiler, then this will break as it can't pre-render the strings for instance.

packages/shared/src/components/buttons/ButtonV2.tsx
```diff
@@ -0,0 +1,193 @@
...
+
+export type ButtonProps<T extends AllowedTags> = BaseButtonProps &
+  HTMLAttributes<T> &
+  JSX.IntrinsicElements[T] & {
+    ref?: Ref<ButtonElementType<T>>;
+  };
+
+const sizeToClassName: Record<ButtonSize, string> = {
+  [ButtonSize.XLarge]: 'border-none',
+  [ButtonSize.Large]: 'h-12 px-6 rounded-[0.88rem]',
```
rebelchris, line 100: Rounded should actually have set variants we already have.
EG `rounded-10` for a 10 pixel round, could you adjust those to be inline?

packages/shared/src/components/buttons/ButtonV2.tsx
```diff
@@ -0,0 +1,193 @@
...
+      ref={ref}
+      className={classNames(
+        `btn inline-flex flex-row items-center justify-center border
+        typo-callout font-bold
+        no-underline shadow-none cursor-pointer select-none focus-outline relative
+        transition duration-200 ease-in-out`,
+        { iconOnly },
+        iconOnly ? iconOnlySizeToClassName[size] : sizeToClassName[size],
+        variantToClassName[variant],
+        variantColorToClassName[variant]?.[color],
```
rebelchris, line 172: Not 100% sure but does this add both in case of variant + color?
EG:
`btn-primary btn-primary-cabbage`?

packages/shared/src/components/buttons/ButtonV2.tsx
```diff
@@ -0,0 +1,193 @@
...
+        iconPosition === ButtonIconPosition.Left &&
+        getIconWithSize(icon)}
+      {children}
+      {icon &&
+        iconPosition === ButtonIconPosition.Right &&
+        getIconWithSize(icon)}
+      {loading && (
+        <Loader
+          data-testid="buttonLoader"
+          className="hidden absolute top-0 right-0 bottom-0 left-0 m-auto btn-loader"
```
rebelchris, line 186:
```suggestion
          className="hidden absolute inset-0 m-auto btn-loader"
```
This should do I think

packages/shared/src/components/buttons/ButtonV2.tsx
```diff
@@ -0,0 +1,193 @@
...
+    <Tag
+      {...props}
+      {...(isAnchor ? combinedClicks(onClick) : { onClick })}
+      aria-busy={loading}
+      aria-pressed={pressed}
+      ref={ref}
+      className={classNames(
+        `btn inline-flex flex-row items-center justify-center border
+        typo-callout font-bold
+        no-underline shadow-none cursor-pointer select-none focus-outline relative
```
rebelchris, line 167: With the link one we add underline right in some cases? 
Just want to check it doesn't interfear

packages/shared/src/styles/components/buttons.css
```diff
@@ -71,14 +77,10 @@
...
-    width: 2.5rem;
-    padding: 0;
-    border-radius: 0.75rem;
-
-    &.xxsmall {
+    &.none {
       width: 1.25rem;
       padding: 0;
-      border-radius: 0.375rem;
+      border-radius: 0.25rem;
```
rebelchris, line 83: Do we still need these as you also have them as tailwind classes?
Not quite sure I followed the suggestion here right.

rebelchris (changes requested):
Amazing work on the type safety!
I think we are almost there, just some minor comments on keeping the implementations up to our guidelines/best practices.

packages/shared/src/components/buttons/ButtonV2.tsx
kopancek: I'd like to keep it separate from this PR, as we have the QuarternaryButton component. Ideally, we create a followup PR where the QuarternaryButton and ClickableText components are merged with the Button component and we decide via variant which type of button we want to render.

packages/shared/src/components/buttons/ButtonV2.tsx
kopancek: I can give it another go and avoid the string concatenation 👍

packages/shared/src/components/buttons/ButtonV2.tsx
kopancek: Yes, but it does not hurt, it works as expected if both classes are added. I can change the logic so that only one class is added if that's what you prefer.

packages/shared/src/styles/components/buttons.css
kopancek: We don't need the `.none` styles for now, I'll remove them.

packages/shared/src/components/buttons/ButtonV2.tsx
rebelchris: I mean sometimes we use `Quarternarybutton` which is a seperate component.
But I see we also use `<Button className="btn-quarternary" />`

Just something to check perhaps?

packages/shared/src/components/buttons/ButtonV2.tsx
kopancek: Yes, but since the link button is currently implemented as <ClickableText> component, I think we should be safe. `no-underline` is also part of the old `<Button>` component.

packages/shared/src/components/buttons/ButtonV2.tsx
rebelchris: Prefer to only have one.
Tailwind keeps changing priority of evaluation so we might end up the wrong one taking priority.

packages/shared/src/components/buttons/ButtonV2.tsx
kopancek: I checked the existing usage of `btn-quaternary`. Other than in QuaternaryButton, it's used in `MultiLevelMenuDetail` component and in `EmailForm`. In both cases, removing the `btn-quaternary` class does not have any effect on the appearance of the button. I think we can ignore that for now and tackle it when doing refactoring of the Button usage in a followup PR.

rebelchris: Perfect, thanks for checking, so we can pretty much just remove it there then.

packages/shared/src/components/buttons/ButtonV2.tsx
```diff
@@ -0,0 +1,256 @@
...
+  return (icon: React.ReactElement<IconProps>) =>
+    React.cloneElement(icon, {
+      size: icon.props?.size ?? buttonSizeToIconSize[size],
+      className: classNames(
+        icon.props.className,
+        !iconOnly && 'text-base !w-6 !h-6',
+        !iconOnly && iconPosition === ButtonIconPosition.Left && '-ml-2 mr-1',
+        !iconOnly && iconPosition === ButtonIconPosition.Right && 'ml-1 -mr-2',
+        // add margin to social icons
+        icon.props.className?.split(' ').includes('socialIcon') && 'mr-3',
```
rebelchris, line 81: I feel this could be on the implementation itself no as a extra component?

packages/shared/src/components/buttons/ButtonV2.tsx
```diff
@@ -0,0 +1,256 @@
...
+export type ButtonProps<T extends AllowedTags> = BaseButtonProps &
+  HTMLAttributes<T> &
+  JSX.IntrinsicElements[T] & {
+    ref?: Ref<ButtonElementType<T>>;
+  };
+
+const sizeToClassName: Record<ButtonSize, string> = {
+  [ButtonSize.XLarge]: 'border-none',
+  [ButtonSize.Large]: 'h-12 px-6 rounded-14',
+  [ButtonSize.Medium]: 'h-10 px-5 rounded-xl',
```
rebelchris, line 101:
```suggestion
  [ButtonSize.Medium]: 'h-10 px-5 rounded-12',
```

packages/shared/src/components/buttons/ButtonV2.tsx
```diff
@@ -0,0 +1,256 @@
...
+  [ButtonSize.Small]: 'h-8 px-3 rounded-10',
+  [ButtonSize.XSmall]: 'h-6 px-2 rounded-lg',
```
rebelchris, line 103:
```suggestion
  [ButtonSize.XSmall]: 'h-6 px-2 rounded-8',
```

packages/shared/src/components/buttons/ButtonV2.tsx
```diff
@@ -0,0 +1,256 @@
...
+};
+
+const iconOnlySizeToClassName: Record<ButtonSize, string> = {
+  [ButtonSize.XLarge]: 'h-16 w-16 p-0 rounded-[1.375rem]',
```
rebelchris, line 107: What amount is this?
We need to check if we support it maybe, or else add it :)

packages/shared/src/components/buttons/ButtonV2.tsx
```diff
@@ -0,0 +1,256 @@
...
+  [ButtonSize.Large]: 'h-12 w-12 p-0 rounded-14',
+  [ButtonSize.Medium]: 'h-10 w-10 p-0 rounded-xl',
```
rebelchris, line 109:
```suggestion
  [ButtonSize.Medium]: 'h-10 w-10 p-0 rounded-12',
```

packages/shared/src/components/buttons/ButtonV2.tsx
```diff
@@ -0,0 +1,256 @@
...
+  [ButtonSize.Small]: 'h-8 w-8 p-0 rounded-10',
+  [ButtonSize.XSmall]: 'h-6 w-6 p-0 rounded-lg',
```
rebelchris, line 111:
```suggestion
  [ButtonSize.XSmall]: 'h-6 w-6 p-0 rounded-8',
```

packages/shared/src/components/buttons/ButtonV2.tsx
```diff
@@ -0,0 +1,256 @@
...
+  Ref,
+} from 'react';
+import classNames from 'classnames';
+import { IconProps, IconSize } from '../Icon';
+import { Loader } from '../Loader';
+import { combinedClicks } from '../../lib/click';
+
+import { ColorName as ButtonColor } from '../../styles/colors';
+
+export enum ButtonSize {
```
rebelchris, line 15: Maybe we can move some of the function definitions to a `/buttons/common` file just to less bloat this page?

rebelchris (commented):
Neat work, left some minor comments.
Feel free to ping others on web as well to review it.

packages/shared/src/components/buttons/ButtonV2.tsx
kopancek: We don't have this at the moment. I was on the fence about adding it vs just using a custom value here, since we do not use it anywhere else right now. But I'll add it 👍

packages/shared/src/components/buttons/ButtonV2.tsx
kopancek: Sorry, I don't think I understand what you meant here :)

packages/shared/src/components/buttons/ButtonV2.tsx
rebelchris: Thanks, it's 22px? Yeah let's add it, as we might use it for profiles or images in the future.

packages/shared/src/components/buttons/ButtonV2.tsx
kopancek: Moved most of constants and the hook into the `common.ts` file.

packages/shared/src/components/buttons/ButtonV2.tsx
kopancek: OK, removed this special casing, we can tackle the styling properly when we apply ButtonV2 to social login buttons.

packages/shared/src/components/buttons/ButtonV2.tsx
```diff
@@ -0,0 +1,119 @@
...
+  useGetIconWithSize,
+  IconOnlySizeToClassName,
+  SizeToClassName,
+  VariantColorToClassName,
+  VariantToClassName,
+} from './common';
+
+export type IconType = React.ReactElement<IconProps>;
+
+export { ButtonColor, ButtonSize, ButtonVariant, ButtonIconPosition };
```
rebelchris, line 27: This feels a bit odd, is it only because we supporting both now and you want to be able to import from 1?
Meaning this should go away once we delete the old one?

rebelchris (approved):
Nice Just one generic question, but think we made it!
Looks amazing, love to see the common file having all the types and the implementation being clean.

Please wait with merging till we get atleast one more review though.

packages/shared/src/components/buttons/ButtonV2.tsx
kopancek: I wanted to be able to control the `ButtonV2` with only one import. If I did not export these fields here, we would end up with 2 imports - one for Button component itself and one for the common button types. I think this is clearer from the consumer perspective, the consumer does not need to care where the `ButtonSize` enum is defined.

rebelchris: Yeah that was my understanding indeed.
So when we remove `Button` (v1) we can remove this as well correct?

nensidosari (approved):
Looks amazing, nthn to add! 🚀

sshanzel:
Just wanted to check if this is now the new behavior of the Button component. Previously, the content is not displayed, I guess it was modified with the fix?

Screenshot 2023-12-13 at 6 32 41 AM

kopancek:
> Just wanted to check if this is now the new behavior of the Button component. Previously, the content is not displayed, I guess it was modified with the fix?
> 
> Screenshot 2023-12-13 at 6 32 41 AM

Yes, should be fixed with commit [0a03969](https://github.com/dailydotdev/apps/pull/2381/commits/0a0396982cc209503248ed47f04ba7bb1088a709)

sshanzel:
> Yes, should be fixed with commit [0a03969](https://github.com/dailydotdev/apps/pull/2381/commits/0a0396982cc209503248ed47f04ba7bb1088a709) and behave the same as before.

Nice, very clever with the invisible class. Makes the content size persist.

PR #2383, opened 2023-12-01
feat: video advanced settings

sshanzel:
## Changes
- Introduced use advanced settings to manage users advanced settings.
- Refactor the Settings page for filters to use the said hook.
- Introduce a new tab for content_types.
  - The selection contains just the video only, but once our advanced settings with `group` grows, this should trigger multiple mutations to update each item. Otherwise we must introduce a new mutation to update all at once.

Note: do we already have an intermediate branch to point this to?

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1985 #done

### DIFF
--- a/packages/shared/src/components/fields/FilterCheckbox.tsx
+++ b/packages/shared/src/components/fields/FilterCheckbox.tsx
@@ -1,0 +1,23 @@
+import React, { ReactElement } from 'react';
+import { Checkbox, CheckboxProps } from './Checkbox';
+
+interface FilterCheckboxProps extends CheckboxProps {
+  description: string;
+}
+
+export function FilterCheckbox({
+  description,
+  children,
+  ...props
+}: FilterCheckboxProps): ReactElement {
+  return (
+    <Checkbox {...props} className="!items-start !typo-callout">
+      <div className="flex flex-col">
+        <span className="mb-2">{children}</span>
+        <span className="font-normal text-theme-label-secondary">
+          {description}
+        </span>
+      </div>
+    </Checkbox>
+  );
+}
--- a/packages/shared/src/components/filters/ContentTypesFilter.tsx
+++ b/packages/shared/src/components/filters/ContentTypesFilter.tsx
@@ -6,6 +6,11 @@ import useFeedSettings from '../../hooks/useFeedSettings';
 export function ContentTypesFilter(): ReactElement {
   const { advancedSettings, isLoading } = useFeedSettings();
   const { selectedSettings, onToggleSettings } = useAdvancedSettings();
+  /*
+   * At the moment, we are only referencing the Video entity, but it should be based from the group that it is in.
+   * Point being, we have to send multiple mutation requests at the same time if the user toggles the switch.
+   * We should instead introduce a new mutation to handle an array of settings to toggle.
+   * */
   const videos = advancedSettings.find(({ title }) => title === 'Videos');
 
   return (
--- a/packages/shared/src/components/filters/ContentTypesFilter.tsx
+++ b/packages/shared/src/components/filters/ContentTypesFilter.tsx
@@ -10,14 +10,18 @@ export function ContentTypesFilter(): ReactElement {
 
   return (
     <section className="flex flex-col px-6" aria-busy={isLoading}>
-      <FilterSwitch
-        name={videos.title}
-        label={videos.title}
-        inputId={videos.title}
-        description={videos.description}
-        checked={selectedSettings[videos.id] ?? videos.defaultEnabledState}
-        onToggle={() => onToggleSettings(videos.id, videos.defaultEnabledState)}
-      />
+      {videos && (
+        <FilterSwitch
+          name={videos.title}
+          label={videos.title}
+          inputId={videos.title}
+          description={videos.description}
+          checked={selectedSettings[videos.id] ?? videos.defaultEnabledState}
+          onToggle={() =>
+            onToggleSettings(videos.id, videos.defaultEnabledState)
+          }
+        />
+      )}
       <FilterSwitch
         name="Articles"
         label="Articles"
--- a/packages/shared/src/components/filters/ContentTypesFilter.tsx
+++ b/packages/shared/src/components/filters/ContentTypesFilter.tsx
@@ -1,5 +1,5 @@
 import React, { ReactElement } from 'react';
-import { FilterSwitch } from './FilterSwitch';
+import { FilterCheckbox } from '../fields/FilterCheckbox';
 import { useAdvancedSettings } from '../../hooks/feed';
 import useFeedSettings from '../../hooks/useFeedSettings';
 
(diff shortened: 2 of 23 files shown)


packages/shared/src/graphql/feedSettings.ts
```diff
@@ -1,17 +1,23 @@
...
 export interface AdvancedSettings {
   id: number;
   title: string;
   description: string;
   defaultEnabledState: boolean;
+  group: AdvancedSettingsGroup;
 }
 
-export interface FeedAdvancedSettings
-  extends Partial<Omit<AdvancedSettings, 'defaultEnabledState'>> {
```
sshanzel, line 12: We're really not using the extended properties, so no point having it here, but instead, it should be from the `advancedSettings` property.

omBratteng (commented):
Throw this up on one of the preview URLs now that the API side was merged?
EOD here, so I'll review this on monday

packages/shared/src/components/filters/ContentTypesFilter.tsx
```diff
@@ -0,0 +1,35 @@
+import React, { ReactElement } from 'react';
+import { FilterSwitch } from './FilterSwitch';
+import { useAdvancedSettings } from '../../hooks/feed';
+import useFeedSettings from '../../hooks/useFeedSettings';
+
+export function ContentTypesFilter(): ReactElement {
+  const { advancedSettings, isLoading } = useFeedSettings();
+  const { selectedSettings, onToggleSettings } = useAdvancedSettings();
+  const videos = advancedSettings.find(({ title }) => title === 'Videos');
```
rebelchris, line 9: Should we not match them on groups perhaps?
Just wondering what happens now if we introduce `Collections` for instance as we filterable type?

rebelchris (changes requested):
I'm kind of missing the place where you leverage this newly introduced `group` setting?
Also it's not a group yet on production?
Screenshot 2023-12-04 at 10 59 42

packages/shared/src/components/filters/ContentTypesFilter.tsx
rebelchris: Guess for now we can leave it, if it's out of scope that's fine.
Something to note for a future change.

Just wondering what we need the group for now then?

sshanzel:
> I'm kind of missing the place where you leverage this newly introduced group setting?
Also it's not a group yet on production?

I based it on any existing type of video group but none was found (not yet in DB). Though it is probably safer that way until we merge this.

packages/shared/src/components/filters/ContentTypesFilter.tsx
sshanzel: Yeah, maybe I need to add a comment on the implementation.

In order for the group to display, we should add the row in our DB.

omBratteng:
Sounds good, probably raise it to design and hear their thoughts.

sshanzel:
> Sounds good, probably raise it to design and hear their thoughts.

Their design should be good (though I will send a message to the channel for confirmation), just our implementation of the base component is wrong 😅

packages/shared/src/components/fields/FilterCheckbox.tsx
```diff
@@ -0,0 +1,23 @@
...
+  description: string;
+}
+
+export function FilterCheckbox({
+  description,
+  children,
+  ...props
+}: FilterCheckboxProps): ReactElement {
+  return (
+    <Checkbox {...props} className="!items-start !typo-callout">
```
rebelchris, line 14: Maybe we can move these to the default className and merge it?
I think moving forward that's a better practive then overwrites.

So:

```
export const Checkbox = forwardRef(function Checkbox(
  {
    className = 'item-center typo-body font-bold,
    ...props
  }: CheckboxProps,
  ref: LegacyRef<HTMLInputElement>,
): ReactElement {
```

Here we would only add the overwrite classes so they default correctly.
Love to here what you think.

packages/shared/src/components/filters/ContentTypesFilter.tsx
```diff
@@ -0,0 +1,40 @@
...
+import { useAdvancedSettings } from '../../hooks/feed';
+import useFeedSettings from '../../hooks/useFeedSettings';
+
+export function ContentTypesFilter(): ReactElement {
+  const { advancedSettings, isLoading } = useFeedSettings();
+  const { selectedSettings, onToggleSettings } = useAdvancedSettings();
+  /*
+   * At the moment, we are only referencing the Video entity, but it should be based from the group that it is in.
+   * Point being, we have to send multiple mutation requests at the same time if the user toggles the switch.
+   * We should instead introduce a new mutation to handle an array of settings to toggle.
```
rebelchris, line 12: Although for the others we already do individual mutations right?
The other advanced settings I mean.

Else you get that you need to toggle multiple and then press save somewhere.

rebelchris (approved):
Nice, minor non breaking comments from my side.

packages/shared/src/components/filters/ContentTypesFilter.tsx
rebelchris: My point being: Wonder if we could already leverage group here and simply use the same mutation?
It's not needed for this implementation PR, but could probably be a very quick solution.

packages/shared/src/components/fields/FilterCheckbox.tsx
sshanzel: It can be a good option but my only concern would be when we just want to add margins and spacing. It will automatically remove the base classes. Wdyt?

rebelchris: Hmm yeah good point, not sure how often we override them.
Another approach could be className as object, but it would require more refactor.

Mostly down the line this is pretty bad to keep doing as it will come back to bite us.

sshanzel: I agree. Hopefully we can figure out a way to avoid the `important` overrides.

packages/shared/src/components/filters/ContentTypesFilter.tsx
sshanzel: I see, in this case, we will pass another parameter that would say toggle based on the group? And then tweak the mutation from the API to toggle all items under it.

rebelchris: For now it's fine, not sure I understood it right.
It could be individual updates right?

sshanzel: Yeah our endpoint supports individual updates but for this checkbox it needs to update all advanced settings under the group in talk.

rebelchris: Why though? It's only video as being one setting no?
I'm a bit confused here.

sshanzel: For now, there is one only `video:youtube`, but it can grow to more than one right? Let's say for example we will get another `video:vimeo` content type which will fall under the same category. This tab is designed to be the categories if I understand correctly.

rebelchris: I think you turn `video` on/off not the specific types.
They are just subtypes for rendering purposes.

But let's fix it once we have another one.

sshanzel: Yep, agreed, let's tackle it when it comes.

PR #2385, opened 2023-12-01
feat: support video in feed and private squads feed

sshanzel:
## Changes
- Updated the base types for which types to display in our regular feed.
- Though I have also updated the types required on private Squads feed, let me know if we proceed with that.

Also, there are other places where we generate feed data and supply it with supported types not based from the `baseSupportedFeedTypes`, such as:
- Bookmark (uses `supportedTypesForPrivateSources`).
- FurtherReading
- FeedPreview
- Sources
- Tags

Let me know if the list above needs to be updated as well. I feel like we should somehow unify this.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1986 #done

### DIFF
--- a/packages/shared/src/graphql/posts.ts
+++ b/packages/shared/src/graphql/posts.ts
@@ -38,6 +38,7 @@ export const supportedTypesForPrivateSources = [
   PostType.Share,
   PostType.Welcome,
   PostType.Freeform,
+  PostType.VideoYouTube,
 ];
 
 type PostFlags = {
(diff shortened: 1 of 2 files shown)


packages/shared/src/graphql/posts.ts
```diff
@@ -38,6 +38,7 @@ export const supportedTypesForPrivateSources = [
   PostType.Share,
   PostType.Welcome,
   PostType.Freeform,
+  PostType.VideoYouTube,
```
rebelchris, line 41: Wondering if we should simply extend `baseFeedSupportedTypes` here?
And include welcome?

rebelchris (approved):
Non breaking comment.

packages/shared/src/graphql/posts.ts
sshanzel: That makes sense, so it is going to be a lot easier.

PR #2390, opened 2023-12-05
fix: feed card feedback actions

sshanzel:
## Changes
- The reason there is no change with the UI is, that the `useVotePost` updates the local cache for specific post query, not the feed query.
- The upvote and downvote are inherently available from the cards so utilizing that instead as it follows the correct update of cache.

Deployed at https://preview3.app.daily.dev/

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done

### DIFF
--- a/packages/shared/src/components/FeedItemComponent.tsx
+++ b/packages/shared/src/components/FeedItemComponent.tsx
@@ -150,10 +150,10 @@ export default function FeedItemComponent({
             ...item.post,
           }}
           data-testid="postItem"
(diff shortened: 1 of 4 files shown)


rebelchris (approved):
Great fix, did you check the other mention of the upvoters list as well as do some research to see if perhaps anything else has the same issue?
(Mostly perhaps look at comments or something like that)

PR #2391, opened 2023-12-05
feat: extending article card to video content

nensidosari:
There are few concerns being checked with design at the moment in slack [thread](https://dailydotdev.slack.com/archives/C059GQ5RFP0/p1701789755348789)

## Changes
- Added PostType VideoYoutube and added wherever PostType is referenced (left TODO's for things yet to be done like post modal and content)
- added check isVideoType based on post type to differentiate between article and video (read post - watch video, read time - watch time, play icon on cover image)
- added play icon in cover image if the post is video type
- added article navigation to postType videoYoutube
- added tests

Added PostType VideoYoutube and adjusted Article card to handle it since the layout and functionality is almost the same. 

<details>

<summary>Screenshots for reference</summary>
Screenshot 2023-12-05 at 9 17 53 AM
Screenshot 2023-12-05 at 9 18 05 AM
Screenshot 2023-12-05 at 9 18 18 AM
Screenshot 2023-12-05 at 4 03 01 PM
Screenshot 2023-12-05 at 4 02 49 PM
Screenshot 2023-12-07 at 12 05 29 AM

</details>

## Events

- [x] MobileL (420px)
- [x] Tablet (656px)
- [x] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1988 #done


packages/shared/src/components/FeedItemComponent.tsx
```diff
@@ -86,6 +86,7 @@ const PostTypeToTag: Record<PostType, FunctionComponent> = {
   [PostType.Share]: SharePostCard,
   [PostType.Welcome]: WelcomePostCard,
   [PostType.Freeform]: WelcomePostCard,
+  [PostType.VideoYouTube]: ArticlePostCard,
```
nensidosari, line 89: Reusing this at the moment since it doesn't require much changes, let me know if this is not the best approach for this moment.

packages/shared/src/components/image/Image.tsx
```diff
@@ -21,8 +24,17 @@ const ImageComponent = (
     }
   };
 
-  // eslint-disable-next-line jsx-a11y/alt-text
-  return <img {...props} ref={ref} onError={onError} />;
+  return (
+    <>
+      {isVideoType && (
+        <div className="flex absolute flex-col justify-center my-2 w-full h-40 rounded-xl bg-overlay-float-pepper">
```
nensidosari, line 30: Did use the colored pepper background as it's dark in design, if I used the themed overflow quarternary that would be white in dark more and pepper in light mode.

packages/shared/src/components/cards/ArticlePostCard.tsx
```diff
@@ -103,6 +106,7 @@ export const ArticlePostCard = forwardRef(function PostCard(
             <PostMetadata
               createdAt={post.createdAt}
               readTime={post.readTime}
```
nensidosari, line 108: Mentioned in [DR](https://dailydotdev.atlassian.net/wiki/spaces/HAN/pages/679149569/YouTube+content#Post-card:~:text=duration-,readTime,-video_id) this is the DB field where the duration is going to be saved, so didn't add any field.
Highlighting to make sure.

packages/shared/src/components/cards/PostMetadata.tsx
```diff
@@ -18,24 +22,35 @@ export default function PostMetadata({
...
+  const timeActionContent = isVideoType ? 'watch' : 'read';
+
   return (
     <div
       className={classNames(
         'flex items-center text-theme-label-tertiary typo-footnote',
         className,
       )}
     >
+      {isVideoType && insaneMode && (
```
nensidosari, line 42: Icon here is smaller then the design, but the XSmall version is bigger, so checking this with design.

packages/shared/src/components/cards/PostList.tsx
```diff
@@ -15,6 +15,7 @@ import PostAuthor from './PostAuthor';
 import FeedItemContainer from './FeedItemContainer';
```
nensidosari, line 15: This looks a bit different then the design for article cards.
Article cards:
Screenshot 2023-12-05 at 4 02 49 PM

New design:
Screenshot 2023-12-05 at 4 08 41 PM

Not sure if it is intended to be changed here as it's not mentioned in the DR either, will check with design.

packages/shared/src/components/cards/PostMetadata.tsx
```diff
@@ -18,24 +22,35 @@ export default function PostMetadata({
...
+        <PlayIcon secondary size={IconSize.XXSmall} className="mr-1" />
+      )}
       {!!description && <span>{description}</span>}
       {!!createdAt && !!description && <Separator />}
       {!!createdAt && <time dateTime={createdAt}>{date}</time>}
```
nensidosari, line 47: Order of created at and readTime is swapped in design, checking with them if this is intended.

capJavert: It was like that for collections too, so I think its design inconsistency.

sshanzel (changes requested):
Overall, it looks great, though I think we missed the part where a post can either be a scraped video post from our end or shared from a Squad. We have covered the case when it is scraped but we also need to cover the case where the video link is shared from a Squad.

Screenshot 2023-12-06 at 8 18 59 AM

packages/shared/src/components/cards/PostMetadata.tsx
nensidosari: Confirmed with design for XXSmall

packages/shared/src/components/cards/PostList.tsx
nensidosari: Confirmed with design & @rebelchris  - this will stay with the old version and the change will be tackled globally for all cards in a different ticket

rebelchris: Should be able to merge it now to new design :) (Actually should just work out of the box now)

packages/shared/src/components/image/Image.tsx
rebelchris: Let's go with `bg-overlay-tertiary-black` we should just add it to our `overlay` file.
Let me know if you need help with that.

rebelchris (changes requested):
Minor adjustment required :)

packages/shared/src/components/image/Image.tsx
rebelchris: If it's a pain you can set it to full height, but might become an issue with the re-layout task as images become flexible.

nensidosari: It needed some tweaks but it's fixed now

nensidosari: > Let's go with bg-overlay-tertiary-black we should just add it to our overlay file.
Let me know if you need help with that.

For this, trying to understand why should we go with "black" ? any reasons to not use twitter as that is our black color name?

nensidosari: This raises issues with eslint "ESLint: bg-overlay-tertiary-black is not a theme color(@dailydotdev/daily-dev-eslint-rules/no-custom-color)" so it would need any changes in the themes as well.

nensidosari: This is solved and pushed now, it was cache issue. Thanks Chris!!

packages/shared/src/components/cards/Card.tsx
```diff
@@ -43,7 +43,31 @@ export const ListCardTitle = classed(Title, 'mr-2');
...
+    wrapper?: string;
+  };
+}
+
+export const CardImage = ({
+  className,
+  ...props
+}: CardImageProps): ReactElement => {
+  return (
+    <div
```
sshanzel, line 58: Just curious, why do we always have to wrap it now?

packages/shared/src/components/cards/Card.tsx
```diff
@@ -43,7 +43,31 @@ export const ListCardTitle = classed(Title, 'mr-2');
...
-export const CardImage = classed(Image, 'rounded-xl h-40');
+interface CardImageProps extends Omit<ImageProps, 'className'> {
+  className?: {
+    image?: string;
+    wrapper?: string;
+  };
+}
+
+export const CardImage = ({
+  className,
```
sshanzel, line 54: For safety, since the object is not using optional chaining on the implementation part.
```suggestion
  className = {},
```

packages/shared/src/components/image/Image.tsx
```diff
@@ -21,8 +23,13 @@ const ImageComponent = (
     }
   };
 
-  // eslint-disable-next-line jsx-a11y/alt-text
-  return <img {...props} ref={ref} onError={onError} />;
+  return (
+    <>
+      {isVideoType && <VideoPlayOverlay />}
```
sshanzel, line 28: Now I understand why the other component has a wrapper, I think it should be done here rather than the `CardImage` component. Then we should utilize the `ConditionalWrapper` being enabled when it is a `videoType`.

packages/shared/src/components/video/VideoPlayOverlay.tsx
```diff
@@ -0,0 +1,23 @@
+import React, { ReactElement } from 'react';
+import PlayIcon from '../icons/Play';
+import { IconSize } from '../Icon';
+
+interface VideoPlayOverlayProps {
+  size?: IconSize;
+}
+
+const VideoPlayOverlay = ({
```
sshanzel, line 9: After updating the `Image` component, we can probably turn this to just a `classed` component so its children will be centered `PlayIcon`. Feels like there's a good chance for things to break when you don't control the container of the absolute elements. Just not really a fan of absolute elements without the container defined side-by-side as it is less predictable. Probably just a personal preference, but would love to hear other people's thoughts. cc @dailydotdev/web-team

sshanzel (commented):
Great work on adding the variant for a Squad post. I don't have anything major but mostly about the approach on how we define components to make it more predictable.

packages/shared/src/components/cards/AdCard.tsx
```diff
@@ -43,17 +43,19 @@ export const AdCard = forwardRef(function AdCard(
...
             alt="Ad image"
             src={ad.image}
-            className={classNames(
-              'w-full z-1',
-              showBlurredImage && 'inset-0 m-auto absolute',
-            )}
+            className={{
+              image: classNames(
+                'w-full z-1',
+                showBlurredImage && 'm-auto absolute top-full',
```
rebelchris, line 49: You don't need the inset anymore to span it? (Just double checking)

packages/shared/src/components/image/Image.tsx
rebelchris: Second opinion:
To me the videoPlayOverlay has nothing to do with the generic image component.
Perhaps we should rather have a `VideoImage` component?

Else we have this logic spread everywhere as images are used widely and simply render a image with fallback basically.

Let me know what either of you think?

rebelchris (changes requested):
Almost there I think.
My suggestion is that we should remove some of the logic and rather create a wrapped component for this sake?

nensidosari:
@rebelchris @sshanzel referring both opinions on VideoPlayOverlay:

Why this component is created is to reuse the logic in Image component, but also when the image is actually a div with background (in the case of [shared post](https://github.com/dailydotdev/apps/blob/ccd42a5342d4919a6fb7ab28cfc2f5c26009a8ba/packages/shared/src/components/cards/SharedPostCardFooter.tsx#L34)). Since we have used 2 ways to display an image in a card both with the same goal we have 2 ways here I believe:
- modify both as done so far and keep the design logic already there
- refactor this to using the same image component (or a new video image component as mentioned by Chris)

I think I went for the first approach to minimise the changes done for shipping this feature by reusing the functionality there, but we can definitely refactor that if we see value at this moment. 

I do like the Video image idea too, separates both concerns and no much need for prop drilling to the very basic image component

packages/shared/src/components/image/Image.tsx
rebelchris: In that case let's maybe go for the video image component?
It just moved the evaluation up and less prop drilling/polluting other components.

sshanzel: Yep. Video component is the way to go. As long as we have everything in one place, I'm cool with it.

packages/shared/src/components/cards/AdCard.tsx
nensidosari: Have reverted this

packages/shared/src/components/cards/SharedPostCardFooter.tsx
```diff
@@ -2,39 +2,51 @@ import classNames from 'classnames';
...
 export const SharedPostCardFooter = ({
   sharedPost,
   isShort,
+  isVideoType,
 }: SharedPostCardFooterProps): ReactElement => {
+  const ImageComponent = isVideoType ? CardVideoImage : CardImage;
   return (
     <div
       className={classNames(
-        'flex flex-1 gap-2 p-3 mb-2 rounded-12 border border-theme-divider-tertiary',
```
nensidosari, line 17: Should we be able to use the defined tailwind classes for min or max height? I wasn't able to, thus max-h-[13.5rem] & min-h-[0]

packages/shared/src/components/cards/PostCardFooter.tsx
```diff
@@ -18,13 +18,16 @@ type PostCardFooterProps = {
...
+  isVideoType?: boolean;
 };
 
 export const PostCardFooter = ({
   post,
   showImage,
   className,
+  isVideoType,
 }: PostCardFooterProps): ReactElement => {
+  const ImageComponent = isVideoType ? CardVideoImage : CardImage;
```
rebelchris, line 30: Nice fix!

packages/shared/src/components/cards/SharePostCard.tsx
```diff
@@ -67,10 +73,11 @@ export const SharePostCard = forwardRef(function SharePostCard(
         title={post.title}
         onHeightChange={onSharedPostTextHeightChange}
       />
-      <Container ref={containerRef}>
+      <Container ref={containerRef} className="justify-end min-h-[0]">
```
rebelchris, line 76: We'll just have to be careful this doesn't introduce edge-cases for other implementations.

packages/shared/src/components/cards/SharedPostCardFooter.tsx
```diff
@@ -2,39 +2,51 @@ import classNames from 'classnames';
...
         isShort ? 'flex-row items-center' : 'flex-col',
       )}
     >
-      <span
+      <div
         className={classNames(
-          'typo-footnote text-theme-label-secondary',
-          isShort ? 'line-clamp-4 flex-1' : 'line-clamp-2',
+          'typo-footnote text-theme-label-secondary flex',
+          isShort ? 'line-clamp-4 pr-3 w-8/12' : 'line-clamp-2',
```
rebelchris, line 29: Why do we need the w-8/12 here? Feels a bit odd to use.

packages/shared/src/components/cards/SharedPostCardFooter.tsx
rebelchris: What did you try to achieve, missing the context a bit.
Feel you can probably use defaults yes.

packages/shared/src/components/cards/SharedPostCardFooter.tsx
```diff
@@ -2,39 +2,51 @@ import classNames from 'classnames';
...
+      </div>
+
+      <div className={classNames('flex flex-auto h-auto overflow-auto')}>
+        <ImageComponent
+          size={isShort ? IconSize.XLarge : IconSize.XXXLarge}
+          alt="Shared Post Cover image"
+          src={sharedPost.image}
+          fallbackSrc={cloudinary.post.imageCoverPlaceholder}
+          className={classNames(
+            'object-cover h-auto min-h-[0px]',
```
rebelchris, line 42: The should be a default better alternative for the min-h, also is it really neeeded?

packages/shared/tailwind.config.js
```diff
@@ -222,6 +222,9 @@ module.exports = {
       minHeight: {
         card: '23.75rem',
       },
+      maxHeight: {
+        card: '23.75rem',
+      },
```
rebelchris, line 227: I'm not sure cards should ever have a max-h? 
As they can grow based on content

rebelchris (changes requested):
Nice, some minor comments but looks great.

packages/shared/src/components/cards/SharePostCard.tsx
nensidosari: This is just used in shared post card, so shouldn't be messing much with other cards... I could test with the data I had and created locally, but could check this if we place it in preview

packages/shared/src/components/cards/SharedPostCardFooter.tsx
nensidosari: So from design the title takes almost 66% of the container and 33% the image (when there are more then 5 lines of post title text), so used this to make sure. If there is any other way to do this top of your mind, please let me know 🙏

packages/shared/src/components/cards/SharedPostCardFooter.tsx
nensidosari: With my implementation, really needed. Allows the image to shrink in height when there is less room in the container. Can't really think of another way to do this at the moment, but happy to hear any other ideas.

If we don't like this and don't have any ideas, I'll have to go at it again and try to re write

packages/shared/tailwind.config.js
nensidosari: Do they?

rebelchris: NVM :)

packages/shared/src/components/cards/SharedPostCardFooter.tsx
rebelchris: And it should apply to all shared post?
I can't really recall

nensidosari: Yes, to all shared post whose main post title is more then 5 lines. So when the main title is more then 5 lines (this is what the user writes when they share a post) then the shared post actual title and image should show side by side.
Screenshot 2023-12-12 at 4 10 48 PM

packages/shared/src/components/cards/SharedPostCardFooter.tsx
nensidosari: I tried deleting .next and all possible things for this, but didn't work, let me try one more time 🤷‍♀️

packages/shared/src/components/cards/SharedPostCardFooter.tsx
rebelchris: But this is a existing implementation right?
Just wondering how this worked alright before then?

packages/shared/src/components/cards/SharedPostCardFooter.tsx
nensidosari: removed usage of `max-h-[13.5rem]` as it was no longer needed.

packages/shared/src/components/cards/SharedPostCardFooter.tsx
nensidosari: it was not using the image component, it was using a div with the post image as a background image

nensidosari: We decided we wanted to unify this in the comments above when the suggestion for VideoImage component was made

packages/shared/src/components/cards/SharePostCard.tsx
nensidosari: should be ok as soon as we update the branches with main

packages/shared/src/components/cards/SharedPostCardFooter.tsx
rebelchris: Oh I see, ok no worries in that case :) 
Alternative would maybe be the image component to have height of parent with square aspect ratio.
Then flex-1 on the text block to fill remaining space?

But also ok with this.
I'm not sure if the above has downsides/issues.

rebelchris (approved):
Nice nothing breaking from my side.
1 suggestive comment, but will leave that up to you.

omBratteng (approved):

packages/shared/src/components/image/VideoImage.tsx
```diff
@@ -0,0 +1,38 @@
...
+    >
+      <>
+        <span className="absolute w-full h-full rounded-xl bg-overlay-tertiary-black" />
+        <PlayIcon
+          secondary
+          size={size}
+          data-testid="playIconVideoPost"
+          className="absolute"
+        />
+      </>
```
sshanzel, line 32: I think we don't need the Fragment anymore. Also we should not be using the tailwind defined rounded classes.
```suggestion
      <span className="absolute w-full h-full rounded-12 bg-overlay-tertiary-black" />
      <PlayIcon
        secondary
        size={size}
        data-testid="playIconVideoPost"
        className="absolute"
      />
```

sshanzel (approved):
Kindly fix the last comment but everything looks great to me! 🚀

packages/shared/src/components/image/VideoImage.tsx
nensidosari: Ah yes, missed it.

nensidosari: As for the rounded classes, I saw the xl used in Card components so went with the same class, wasn't aware we shouldn't use them

sshanzel: Yep, legacy stuff. It is easier for us to understand as the classes with explicit numbers came from our design system.

packages/shared/src/components/cards/SharedPostCardFooter.tsx
nensidosari: I tried and the image takes a a bigger height then the parent, as the parent doesn't have a set height, maybe I'm missing something there. 
But thinking of merging this as is so I don't block the feature, and we can check together later if you want.

PR #2393, opened 2023-12-05
fix: make `getRank` resilient to undefined arguments being passed in

DragosIorgulescu:
- why the `rank` may be coming back as undefined is to be determined separately

## Events

N / A

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-2036 #done

### DIFF
--- a/packages/shared/src/lib/rank.ts
+++ b/packages/shared/src/lib/rank.ts
@@ -103,7 +103,7 @@ interface GetNextRankTextProps {
   showNextLevel?: boolean;
 }
 
-export const getRank = (rank: number | undefined): number => {
+export const getRank = (rank = 0): number => {
   const computedRank = rank ? Math.max(0, rank - 1) : 0;
 
   if (computedRank >= RANKS.length) {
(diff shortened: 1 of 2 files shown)


packages/shared/src/lib/rank.ts
```diff
@@ -103,8 +103,8 @@ interface GetNextRankTextProps {
   showNextLevel?: boolean;
 }
 
-export const getRank = (rank: number): number => {
-  const computedRank = Math.max(0, rank - 1);
+export const getRank = (rank: number | undefined): number => {
```
rebelchris, line 106: Would it not work like this:
```suggestion
export const getRank = (rank = 0): number => {
```

DragosIorgulescu: I think that only works if the `rank` argument is not passed at all. But in our case, the `rank` is an actual variable that is passed which has the value of `undefined`

DragosIorgulescu: scratch that, your suggestion works even in the case I mentioned 🤝

rebelchris (approved):

PR #2394, opened 2023-12-05
fix: editing comments cache update

sshanzel:
## Changes
- The reason why the cache is not updating, we should now deep-clone the data we are trying to update. Else, if it is considered to be the same, no re-render is triggered.
- Another issue shows, that when editing the same comment for the second time in one session, the latest value is not displayed, so in order to always base from the query, we need to add another parameter to ensure the value and date match.

Seems like after the react-query upgrade we now need to deep clone objects.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-2022 #done

### DIFF
--- a/packages/shared/src/components/comments/MainComment.tsx
+++ b/packages/shared/src/components/comments/MainComment.tsx
@@ -78,7 +78,9 @@ export default function MainComment({
               commentId: selected.id,
             })
           }
-          onEdit={(selected) => onEdit({ commentId: selected.id })}
+          onEdit={({ id, lastUpdatedAt }) =>
+            onEdit({ commentId: id, lastUpdatedAt })
+          }
         />
       )}
       {editProps && (
--- a/packages/shared/src/components/comments/SubComment.tsx
+++ b/packages/shared/src/components/comments/SubComment.tsx
@@ -32,9 +32,10 @@ function SubComment({
           key={comment.id}
           parentId={parentComment.id}
           comment={comment}
-          onEdit={(selected) =>
+          onEdit={({ id, lastUpdatedAt }) =>
             onEdit({
-              commentId: selected.id,
+              commentId: id,
+              lastUpdatedAt,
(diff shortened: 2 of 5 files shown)


rebelchris (approved):
Wonder if it has impact on more things now 🫠

sshanzel:
> Wonder if it has impact on more things now 🫠

I'll be sure to be on the lookout 👀

capJavert:
Tnx for the fix 👍 

Catching this a bit late, I would not say this issue is with react-query since they for sure do not traverse the whole object tree to find nested differences but it could be due to fact that some of nested objects were referentially equal and then component or useMemo would not re-render even if child properties like arrays changed. For example `postComments` which was mutated but its reference stayed the same, and we do use that object to actually render comments. 

> Wonder if it has impact on more things now 🫠

Using cloneDeep all over codebase can quickly become quite performance intensive so we should probably pull out this case by itself and look at the parts we are mutating inline.

sshanzel:
> Catching this a bit late, I would not say this issue is with react-query since they for sure do not traverse the whole object tree to find nested differences but it could be due to fact that some of nested objects were referentially equal and then component or useMemo would not re-render even if child properties like arrays changed. For example postComments which was mutated but its reference stayed the same, and we do use that object to actually render comments.

I had the same thought initially, but I manually checked the components where the comments are rendered, I even switched the `key` for mapping the content to use the updated field in order to force react to render the new value but still, no luck.

> Using cloneDeep all over codebase can quickly become quite performance intensive so we should probably pull out this case by itself and look at the parts we are mutating inline and then make sure it is done in immutable way.

Agree with this, I actually hate lodash if I'm being honest, but looking at the initial implementation we already update the object in an immutable way (single level). I also even tried to re-create the `edges` part since that is where the map happens.

PR #2397, opened 2023-12-06
feat: youtube content for post modal/page

sshanzel:
## Changes
- Refactored share post content
  - Moved `PostLink` as a reusable component
  - Created a new container component for content with collapsible TLDR.
- Updated `PostContent` to manage YT content.
- Introduced post content component for YT-related posts. 

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1989 #done


packages/shared/src/components/post/common/SharedLinkContainer.tsx
```diff
@@ -0,0 +1,69 @@
...
+}
+
+export function SharedLinkContainer({
+  children,
+  summary,
+  className,
+}: SharedLinkContainerProps): ReactElement {
+  const [height, setHeight] = useState<number>(null);
+  const [shouldShowSummary, setShouldShowSummary] = useState(true);
+  const tldrHeight = useMemo(() => {
```
sshanzel, line 19: Same implementation with the toggle. Just moved to its own file.

packages/shared/src/components/Feed.tsx
```diff
@@ -338,7 +338,11 @@ export default function Feed<T>({
     nextPost: (items[postMenuIndex + 1] as PostItem)?.post,
   };
 
-  const ArticleModal = PostModalMap[selectedPost?.type];
+  const PostModal =
+    post?.sharedPost?.type === PostType.VideoYouTube &&
+    post?.source?.type === SourceType.Squad
+      ? SharePostModal
+      : PostModalMap[selectedPost?.type];
```
rebelchris, line 345: Why do you need this weird evaluation?
I'm not sure why it makes a difference if the sharedPost is a video or not?

packages/shared/src/components/post/SquadPostContent.tsx
```diff
@@ -17,11 +17,13 @@ import MarkdownPostContent from './MarkdownPostContent';
 import { SquadPostWidgets } from './SquadPostWidgets';
 import { isSourcePublicSquad } from '../../graphql/squads';
 import { useAuthContext } from '../../contexts/AuthContext';
+import ShareYouTubeContent from './ShareYouTubeContent';
 
 const ContentMap = {
   [PostType.Freeform]: MarkdownPostContent,
   [PostType.Welcome]: MarkdownPostContent,
   [PostType.Share]: SharePostContent,
+  [PostType.VideoYouTube]: ShareYouTubeContent,
```
rebelchris, line 26: Can't we simply reuse the article here as it's pretty much the same?

packages/shared/src/components/post/ShareYouTubeContent.tsx
```diff
@@ -0,0 +1,44 @@
...
+import { SharedLinkContainer } from './common/SharedLinkContainer';
+import { SharedPostLink } from './common/SharedPostLink';
+import { YouTubeVideo } from './common/YouTubeVideo';
+
+interface ShareYouTubeContentProps {
+  post: Post;
+  onReadArticle: () => Promise<void>;
+}
+
+function ShareYouTubeContent({ post }: ShareYouTubeContentProps): ReactElement {
```
rebelchris, line 14: This is for the shared post card?
I'm a bit confused by the naming so not 100% sure here.

packages/shared/src/components/post/common/YouTubeVideo.tsx
```diff
@@ -0,0 +1,24 @@
+import React, { ReactElement } from 'react';
+import classNames from 'classnames';
+
+interface YouTubeVideoProps {
+  title: string;
+  videoId: string;
+  className?: string;
+}
+
+export function YouTubeVideo({
```
rebelchris, line 10: This already existed right?

packages/webapp/pages/posts/[id]/index.tsx
```diff
@@ -73,7 +74,7 @@ const CONTENT_MAP: Record<PostType, typeof PostContent> = {
   welcome: SquadPostContent,
   freeform: SquadPostContent,
   // TODO: remove this once we know what content to use
-  [PostType.VideoYouTube]: null,
+  [PostType.VideoYouTube]: PostContent,
```
Comment (rebelchris) [inline packages/webapp/pages/posts/[id]/index.tsx:77]: Still not the final?

packages/shared/src/components/post/SquadPostContent.tsx
sshanzel: This is for Shared post with a type of Video. The article post is reused when it came from a direct source.

packages/shared/src/components/post/ShareYouTubeContent.tsx
sshanzel: This is still for the post modal for shared youtube video. You will notice there are two Sketch boards for Post Modal that is because there can be two sources. Either directly from YT or a link shared through a Squad.

packages/shared/src/components/post/common/YouTubeVideo.tsx
sshanzel: Ah, right. Should remove the TODO. This is where the article modal components are used.

packages/webapp/pages/posts/[id]/index.tsx
Comment (sshanzel) [reply packages/webapp/pages/posts/[id]/index.tsx]: Should be removed now 🥳

packages/shared/src/components/Feed.tsx
rebelchris: But should it not just be a shared post then?
Nensi's PR should already reflect the rendering difference I think?

The only difference I think is the image is now a video, so it should just be another evaluation of the shared post content I personally think.

Let me know what you think, else maybe we can sync on it quickly?

sshanzel: > But should it not just be a shared post then?

I was thinking of that too. I thought of checking the spec but it seemed to be empty. The way it was presented in the design looks like you should be able to play the video right away.

> Let me know what you think, else maybe we can sync on it quickly?

We can probably talk about it in our upcoming 1-1.

packages/shared/src/components/post/SquadPostContent.tsx
rebelchris: This will actually never evaluate.
The share post can't be directly video, it has to be Share itself.
The subtype will be video:youtube that's being shared.

packages/shared/src/components/post/common/YouTubeVideo.tsx
sshanzel: Not quite sure, do you know the exact component?

packages/shared/src/components/Feed.tsx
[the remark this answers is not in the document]
sshanzel: Yep, should work as how we discussed.

packages/shared/src/components/post/common/YouTubeVideo.tsx
sshanzel: Thanks! Removed this file now 🚢

rebelchris (approved):
Looks good to me :)

PR #2399, opened 2023-12-07
feat: adding hook and experiment values

nensidosari:
## Changes
- added new hook for feed layout changes to return object based on the changes required for the feed to keep them centralised
- added experiment values and management

- Sets the stage for the new feed layout changed

## Events

## Manual Testing - No

WT-1997 #done

### DIFF
--- a/packages/shared/src/hooks/useFeedLayout.ts
+++ b/packages/shared/src/hooks/useFeedLayout.ts
@@ -6,7 +6,10 @@ interface UseFeedLayoutProps {
   feedName: string;
 }
 
-export const useFeedLayout = ({ feedName }: UseFeedLayoutProps) => {
+// TODO: return type to be changed once more info is added to the return object
+export const useFeedLayout = ({
+  feedName,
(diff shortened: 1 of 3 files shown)


packages/shared/src/hooks/useFeedLayout.ts
```diff
@@ -0,0 +1,24 @@
+import { useFeature } from '../components/GrowthBookProvider';
+import { feature } from '../lib/featureManagement';
+import { FeedLayout } from '../lib/featureValues';
+
+interface UseFeedLayoutProps {
+  feedName: string;
+}
+
+// TODO: return type to be changed once more info is added to the return object
+export const useFeedLayout = ({
```
rebelchris, line 10: I'm not 100% sure why we need the hook again might just miss the context or forgot about it.
In either case the naming is a little bit confusing, could you maybe help me understand why we can't simply use the feature as the evaluation?

rebelchris (approved):
After seeing the usage this seems good

PR #2400, opened 2023-12-07
feat: single column feed

nensidosari:
## Changes
- update hook to check for the right feed to change and return right values for isSingleColumnFeedLayout
- use the returned value from hook for how to render feed 

- Converts feed from multi column to single column layout

<details>

<summary>Video for webapp single col feed</summary>

https://github.com/dailydotdev/apps/assets/36197304/31703cd4-da4e-4773-b38b-4aa59bb2cad7

</details>

<details>

<summary>Video for extension single col feed</summary>

https://github.com/dailydotdev/apps/assets/36197304/2a8100de-cfc8-4f73-8638-0d0e4d341b80

</details>

## Events

- [x] MobileL (420px)
- [x] Tablet (656px)
- [x] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1995 #done

### DIFF
--- a/packages/shared/src/components/Feed.tsx
+++ b/packages/shared/src/components/Feed.tsx
@@ -142,7 +142,7 @@ export default function Feed<T>({
       items,
       feedName,
     };
-  }, [feedQueryKey, items]);
+  }, [feedQueryKey, items, feedName]);
 
   const { ranking } = (variables as RankVariables) || {};
   const {
--- a/packages/shared/src/components/Feed.tsx
+++ b/packages/shared/src/components/Feed.tsx
@@ -305,7 +305,7 @@ export default function Feed<T>({
     );
   };
 
-  const onAdClick = (ad: Ad, index: number, row: number, column: number) => {
+  const onAdClick = (ad: Ad, _: number, row: number, column: number) => {
     trackEvent(
       adAnalyticsEvent('click', ad, {
         columns: virtualizedNumCards,
@@ -356,8 +356,9 @@ export default function Feed<T>({
         showSearch={showSearch && isValidFeed}
         besideSearch={besideSearch}
         actionButtons={actionButtons}
+        feedName={feedName}
       >
-        {items.map((item, index) => (
+        {items.map((_, index) => (
           <FeedItemComponent
             items={items}
             index={index}
--- a/packages/shared/src/components/Feed.tsx
+++ b/packages/shared/src/components/Feed.tsx
@@ -140,6 +140,7 @@ export default function Feed<T>({
     return {
       queryKey: feedQueryKey,
       items,
+      feedName,
     };
   }, [feedQueryKey, items]);
 
@@ -305,7 +306,7 @@ export default function Feed<T>({
     );
   };
 
-  const onAdClick = (ad: Ad, _: number, row: number, column: number) => {
+  const onAdClick = (ad: Ad, row: number, column: number) => {
     trackEvent(
       adAnalyticsEvent('click', ad, {
         columns: virtualizedNumCards,
@@ -356,7 +357,6 @@ export default function Feed<T>({
         showSearch={showSearch && isValidFeed}
         besideSearch={besideSearch}
         actionButtons={actionButtons}
-        feedName={feedName}
       >
         {items.map((_, index) => (
           <FeedItemComponent
--- a/packages/shared/src/components/feeds/FeedContainer.tsx
+++ b/packages/shared/src/components/feeds/FeedContainer.tsx
@@ -14,10 +14,6 @@ import SettingsContext from '../../contexts/SettingsContext';
 import FeedContext from '../../contexts/FeedContext';
 import ScrollToTopButton from '../ScrollToTopButton';
 import styles from '../Feed.module.css';
-import {
-  ToastSubject,
-  useToastNotification,
-} from '../../hooks/useToastNotification';
 import { SearchBarInput, SearchBarSuggestionList } from '../search';
 import { useFeature } from '../GrowthBookProvider';
 import { feature } from '../../lib/featureManagement';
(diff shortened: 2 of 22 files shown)


packages/shared/src/components/feeds/FeedContainer.tsx
```diff
@@ -39,6 +41,7 @@ export interface FeedContainerProps {
   showSearch?: boolean;
   besideSearch?: ReactNode;
   actionButtons?: ReactNode;
+  feedName?: AllFeedPages;
```
capJavert, line 44: You might not needs this, you can use `ActiveFeedContext`, add `feetName` there and use it without prop drilling.

capJavert (commented):
Looking good 👌 , some optimization ideas.

sshanzel (commented):
Looks great! Nothing to add on top of Ante's comments.

packages/shared/src/hooks/useFeedLayout.ts
```diff
@@ -1,24 +1,33 @@
...
 
-// TODO: return type to be changed once more info is added to the return object
 export const useFeedLayout = ({
   feedName,
-}: UseFeedLayoutProps): Record<string, unknown> => {
+}: UseFeedLayoutProps): UseFeedLayoutReturn => {
   const feedLayoutVersion = useFeature(feature.feedLayout);
   const isFeedLayoutVersion = feedLayoutVersion === FeedLayout.V1;
-  const shouldUseFeedLayout = isFeedLayoutVersion && feedName !== 'squad';
+  const isIncludedFeed = Object.values(SharedFeedPage).includes(
```
rebelchris, line 20: We did say squad page would be excluded only because of the header right?
Just trying to remember why we didn't apply it there.

packages/shared/src/hooks/useFeedLayout.ts
```diff
@@ -1,24 +1,33 @@
...
+  isSingleColumnFeedLayout: boolean;
 }
 
-// TODO: return type to be changed once more info is added to the return object
 export const useFeedLayout = ({
   feedName,
-}: UseFeedLayoutProps): Record<string, unknown> => {
+}: UseFeedLayoutProps): UseFeedLayoutReturn => {
   const feedLayoutVersion = useFeature(feature.feedLayout);
   const isFeedLayoutVersion = feedLayoutVersion === FeedLayout.V1;
```
rebelchris, line 19: I'm not sure this naming is clear in this case.

packages/shared/src/hooks/useFeedLayout.ts
```diff
@@ -1,24 +1,33 @@
...
-  const shouldUseFeedLayout = isFeedLayoutVersion && feedName !== 'squad';
+  const isIncludedFeed = Object.values(SharedFeedPage).includes(
+    feedName as SharedFeedPage,
+  );
+  const shouldUseFeedLayout = isFeedLayoutVersion && isIncludedFeed;
```
rebelchris, line 23: The name is confusing as everything is a feed layout no?
You want to evaluate if something is v1 and applicable?

rebelchris (changes requested):
I see why you need the hook now, though think the naming can be optimized.

rebelchris (changes requested):
I see why you need the hook now, though think the naming can be optimized.

packages/shared/src/components/Feed.tsx
```diff
@@ -305,7 +305,7 @@ export default function Feed<T>({
     );
   };
 
-  const onAdClick = (ad: Ad, index: number, row: number, column: number) => {
+  const onAdClick = (ad: Ad, _: number, row: number, column: number) => {
```
kopancek, line 308: If we don't need the number, can we not send it as a parameter?

packages/shared/src/hooks/useFeedLayout.ts
kopancek: Agreed, maybe `isV1` or similar would be a better name here.

packages/shared/src/hooks/useFeedLayout.ts
```diff
@@ -0,0 +1,33 @@
...
+
+  if (shouldUseFeedLayout) {
+    return {
+      isSingleColumnFeedLayout: true,
+    };
+  }
+  return {
+    isSingleColumnFeedLayout: false,
+  };
```
kopancek, line 32: I think this logic can be simplified to a lot less lines (maybe even one, if we inline the `isSingleColumnFeedLayout` variable):
```suggestion
  const isSingleColumnFeedLayout = isFeedLayoutVersion && isIncludedFeed;
  return { isSingleColumnFeedLayout };
```

kopancek (commented):
Added some small comments for improvements. 

Overall would also be nice to have a test, where we render multi-column feed layout if the feature flag is off and single-column if the feature flag is on. But I'm not sure how much effort that would be or if it's common to write such tests in dailydev.

packages/shared/src/hooks/useFeedLayout.ts
```diff
@@ -1,24 +1,33 @@
...
     return {
-      // the values of the object can also be specific to
-      // certain components and pages and can be separated in the return
+      isSingleColumnFeedLayout: true,
     };
   }
-  return {};
+  return {
+    isSingleColumnFeedLayout: false,
+  };
```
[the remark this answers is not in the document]
nensidosari: Yeah for sure we can

packages/shared/src/hooks/useFeedLayout.ts
nensidosari: So we are only applying it to shared feed pages https://dailydotdev.slack.com/archives/C061T644XV3/p1700218672945989 based on this discussion.

So basically there could be more edge cases there and product wants to experiment with these feeds first..

packages/shared/src/hooks/useFeedLayout.ts
nensidosari: Yeah true, would this be ok `shouldUseFeedLayoutV1` ?

packages/shared/src/components/Feed.tsx
nensidosari: Yeah sure, didn't properly notice

packages/shared/src/hooks/useFeedLayout.ts
nensidosari: Yeah, based on Ante's comment as well will change this, my mind was set on separating the cases because other props might not be only related to the bool value of `isFeedLayoutVersion && isIncludedFeed`, but will change based on suggestions as it definitely makes sense now :D Thank u

kopancek (approved):
Nothing to add, LGTM 🚢

rebelchris (approved):
Looks great!

PR #2402, opened 2023-12-11
feat: youtube content

omBratteng:
## Changes

YouTube content

WT-1905

### DIFF
--- a/packages/shared/src/components/cards/ArticlePostCard.tsx
+++ b/packages/shared/src/components/cards/ArticlePostCard.tsx
@@ -97,13 +97,13 @@ export const ArticlePostCard = forwardRef(function PostCard(
       >
         <CardTextContainer>
           <PostCardHeader
+            post={post}
             className={showFeedback ? 'hidden' : 'flex'}
             openNewTab={openNewTab}
             source={post.source}
             postLink={post.permalink}
             onMenuClick={(event) => onMenuClick?.(event, post)}
             onReadArticleClick={onReadArticleClick}
-            isVideoType={isVideoType}
           />
           <CardTitle lineClamp={showFeedback ? 'line-clamp-2' : undefined}>
             {post.title}
--- a/packages/shared/src/components/cards/ArticlePostCard.tsx
+++ b/packages/shared/src/components/cards/ArticlePostCard.tsx
@@ -19,6 +19,7 @@ import { useFeedPreviewMode, usePostFeedback } from '../../hooks';
 import styles from './Card.module.css';
 import { FeedbackCard } from './FeedbackCard';
 import { Origin } from '../../lib/analytics';
+import { PostType } from '../../graphql/posts';
 
 export const ArticlePostCard = forwardRef(function PostCard(
   {
@@ -47,6 +48,7 @@ export const ArticlePostCard = forwardRef(function PostCard(
   const customStyle = !showImage ? { minHeight: '15.125rem' } : {};
   const { showFeedback } = usePostFeedback({ post });
   const isFeedPreview = useFeedPreviewMode();
+  const isVideoType = post.type === PostType.VideoYouTube;
 
   if (data?.showTagsPanel && post.tags.length > 0) {
     return (
@@ -101,6 +103,7 @@ export const ArticlePostCard = forwardRef(function PostCard(
             postLink={post.permalink}
             onMenuClick={(event) => onMenuClick?.(event, post)}
             onReadArticleClick={onReadArticleClick}
+            isVideoType={isVideoType}
           />
           <CardTitle lineClamp={showFeedback ? 'line-clamp-2' : undefined}>
             {post.title}
@@ -112,7 +115,9 @@ export const ArticlePostCard = forwardRef(function PostCard(
             <PostMetadata
               createdAt={post.createdAt}
               readTime={post.readTime}
+              isVideoType={isVideoType}
               className="mx-4"
+              insaneMode={insaneMode}
             />
           </Container>
         )}
@@ -122,6 +127,7 @@ export const ArticlePostCard = forwardRef(function PostCard(
             openNewTab={openNewTab}
             post={post}
             showImage={showImage}
+            isVideoType={isVideoType}
             className={{
               image: classNames(showFeedback && 'mb-0'),
             }}
--- a/packages/shared/src/components/cards/ArticlePostCard.tsx
+++ b/packages/shared/src/components/cards/ArticlePostCard.tsx
@@ -19,7 +19,7 @@ import { useFeedPreviewMode, usePostFeedback } from '../../hooks';
 import styles from './Card.module.css';
 import { FeedbackCard } from './FeedbackCard';
 import { Origin } from '../../lib/analytics';
-import { PostType } from '../../graphql/posts';
+import { isVideoPost } from '../../graphql/posts';
 
 export const ArticlePostCard = forwardRef(function PostCard(
   {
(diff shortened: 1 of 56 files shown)


packages/webapp/pages/posts/[id]/index.tsx
```diff
@@ -72,8 +72,7 @@ const CONTENT_MAP: Record<PostType, typeof PostContent> = {
   share: SquadPostContent,
   welcome: SquadPostContent,
   freeform: SquadPostContent,
-  // TODO: remove this once we know what content to use
-  [PostType.VideoYouTube]: null,
+  [PostType.VideoYouTube]: PostContent,
```
Comment (omBratteng) [inline packages/webapp/pages/posts/[id]/index.tsx:75]: @sshanzel there might be an issue here. I can open a squad post directly here, and it would render the wrong layout.
image

Comment (sshanzel) [reply packages/webapp/pages/posts/[id]/index.tsx:75]: You mean a Squad post for video? If that is the case, it would be handled in a separate file (SquadPostContent) where it decided whether to display the layout for Video.

Comment (omBratteng) [reply packages/webapp/pages/posts/[id]/index.tsx:75]: You can see it in the squad I created.
https://app.daily.dev/posts/ce4uyPVi8
https://app.daily.dev/squads/youtubecontent

Comment (sshanzel) [reply packages/webapp/pages/posts/[id]/index.tsx:75]: Can you invite me to the Squad? Though, yes, we need to [update something](https://dailydotdev.slack.com/archives/C059GQ5RFP0/p1702440788590979?thread_ts=1702306710.702739&cid=C059GQ5RFP0) in the API side in order to reflect the right layout. Currently, the post would be considered as an ArticlePost rather than VideoYouTube.

Comment (sshanzel) [reply packages/webapp/pages/posts/[id]/index.tsx:75]: This is the reason why.

https://github.com/dailydotdev/daily-api/blob/655c6af2fed9e082f8ce7bee3d66fb73dece8035/src/entity/posts/utils.ts#L241

Comment (omBratteng) [reply packages/webapp/pages/posts/[id]/index.tsx:75]: Invite [link here](https://dailydotdev.slack.com/archives/C059GQ5RFP0/p1702463442840419)

packages/shared/src/components/cards/ArticlePostCard.tsx
```diff
@@ -47,6 +48,7 @@ export const ArticlePostCard = forwardRef(function PostCard(
   const customStyle = !showImage ? { minHeight: '15.125rem' } : {};
   const { showFeedback } = usePostFeedback({ post });
   const isFeedPreview = useFeedPreviewMode();
+  const isVideoType = post.type === PostType.VideoYouTube;
```
rebelchris, line 51: I noticed we have multiple definitions of `isVideoType` sometimes it's like this, sometimes it also checks if the sharedPost is a video.
Wondering if it will cause confusion down the line, and maybe we should turn this into a utility functions like we do to determine the sharedPost for instance?
It could take a prop of `alsoCheckSharedTyped` (suck at naming)

packages/shared/src/components/cards/PostList.tsx
```diff
@@ -33,7 +34,10 @@ export const PostList = forwardRef(function PostList(
 ): ReactElement {
   const { data } = useBlockPostPanel(post);
   const onPostCardClick = () => onPostClick(post);
-  const { trending, pinnedAt } = post;
+  const { trending, pinnedAt, type, sharedPost } = post;
+  const isVideoType =
```
rebelchris, line 38: Here is the other definition for example

packages/shared/src/components/post/PostContent.tsx
```diff
@@ -154,32 +153,45 @@ export function PostContent({
               {post.title}
             </a>
           </h1>
+          {post.type === PostType.VideoYouTube && (
```
rebelchris, line 156: Here we do it inline, could be the same util function for safety

packages/shared/src/components/post/PostContent.tsx
```diff
@@ -74,6 +72,7 @@ export function PostContent({
   const { onSharePost: onShare, onReadArticle } = engagementActions;
 
   const hasNavigation = !!onPreviousPost || !!onNextPost;
+  const isVideoType = post.type === PostType.VideoYouTube;
```
rebelchris, line 75: We have the same fn here.

packages/shared/src/components/post/PostContent.tsx
```diff
@@ -154,32 +153,45 @@ export function PostContent({
...
-          >
-            <LazyImage
-              imgSrc={post.image}
-              imgAlt="Post cover image"
-              ratio="49%"
-              eager
-              fallbackSrc={cloudinary.post.imageCoverPlaceholder}
-            />
-          </a>
+          {post.type !== PostType.VideoYouTube && (
```
rebelchris, line 176: Inline again

packages/shared/src/components/post/PostHeaderActions.tsx
```diff
@@ -96,13 +97,15 @@ export function PostHeaderActions({
       }
     }
   };
+  const readButtonText =
+    post.type === PostType.VideoYouTube ? 'Watch video' : 'Read post';
```
rebelchris, line 101: Seen two cases of this already, perhaps also a good util function.
Let's say we want to adjust the copy, would now need to happen in multiple places.

packages/shared/src/components/post/SquadPostContent.tsx
```diff
@@ -70,7 +72,9 @@ function SquadPostContent({
     onSendViewPost(post.id);
   }, [post.id, onSendViewPost, user?.id]);
 
-  const Content = ContentMap[post?.type];
+  const isVideoShared = post?.sharedPost?.type === PostType.VideoYouTube;
```
rebelchris, line 75: Maybe this could be part of the util export.
And then a combined one?

rebelchris (approved):
Some minor comments, nothing blocking, but might be nice to quick fix.

packages/shared/src/components/cards/ArticlePostCard.tsx
omBratteng: I think just checking like this works. Kinda what I did in #2434 (at first, I see I removed it)

```js
post.type === PostType.VideYouTube ||
(post.type === PostType.Shared && post.sharedPost.type === PostType.VideoYouTube)
`

Comment (omBratteng) [reply packages/shared/src/components/cards/ArticlePostCard.tsx]: @rebelchris refactored the check if post is youtube or shared youtube to a dedicated util function in b7794e1

File: packages/shared/src/components/post/PostHeaderActions.tsx (hunk 7 above)
Comment (rebelchris) [reply packages/shared/src/components/post/PostHeaderActions.tsx]: Yes exactly, just so it's managed from 1 spot :)

File: packages/shared/src/graphql/posts.ts [hunk 9]
```diff
@@ -42,6 +42,11 @@ export const isInternalReadType = (post: Post): boolean =>
 export const isSharedPostSquadPost = (post: Post): boolean =>
   post.sharedPost?.source.type === SourceType.Squad;
 
+export const isVideoPost = (post: Post): boolean =>
+  post?.type === PostType.VideoYouTube ||
+  (post?.type === PostType.Share &&
+    post?.sharedPost?.type === PostType.VideoYouTube);
+
```
rebelchris, line 49: In the previous examples we sometimes only checked the first line, just wondering if there where examptions where it was needed to not verify the share?

omBratteng: I don't think so, but I'll double check

packages/shared/src/components/post/PostHeaderActions.tsx
omBratteng: Refactored in a7ac75a, now just one place to update it

rebelchris: Nice love it 🍡

packages/shared/src/graphql/posts.ts
omBratteng: Because we use `||`, the second part would only be checked if the first is false. And I don't think there are any exemptions to not verify the share.

PR #2403, opened 2023-12-12
fix: missing analytics for go to link action

sshanzel:
## Changes
- Introduced the [same component](https://github.com/dailydotdev/apps/blob/f8e5b0aea4564937115b9c41399782b4579657dc/packages/shared/src/components/post/common/SharedPostLink.tsx) found in the YT intermediate branch.
- Fix the missing analytics for go to link actions.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-2054 #done

### DIFF
--- a/packages/shared/src/components/post/SharePostContent.tsx
+++ b/packages/shared/src/components/post/SharePostContent.tsx
@@ -1,6 +1,5 @@
 import React, { ReactElement, useContext, useMemo, useState } from 'react';
 import classNames from 'classnames';
-import Link from 'next/link';
 import PostSourceInfo from './PostSourceInfo';
 import { ReadArticleButton } from '../cards/ReadArticleButton';
 import { LazyImage } from '../LazyImage';
@@ -12,23 +11,7 @@ import SettingsContext from '../../contexts/SettingsContext';
 import { SharePostTitle } from './share';
 import { combinedClicks } from '../../lib/click';
 import { SourceType } from '../../graphql/sources';
-
-interface PostLinkProps {
-  href: string;
-  as?: string;
-  target?: string;
-  rel?: string;
-  className?: string;
-  children: React.ReactNode;
-}
-
-const PostLink = ({ href, as, children, ...props }: PostLinkProps) => {
-  return (
-    <Link href={href} as={as}>
-      <a {...props}>{children}</a>
-    </Link>
-  );
-};
+import { SharedPostLink } from './common/SharedPostLink';
 
 interface SharePostContentProps {
   post: Post;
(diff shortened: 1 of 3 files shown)


packages/shared/src/components/post/common/SharedPostLink.tsx
```diff
@@ -0,0 +1,42 @@
+import Link from 'next/link';
+import React, { ReactElement, ReactNode } from 'react';
+import { Post } from '../../../graphql/posts';
+import { CombinedClicks } from '../../../lib/click';
```
rebelchris, line 4: Alternative would have been to import type from.
We have some examples of that in the codebase.

packages/shared/src/components/post/SharePostContent.tsx
```diff
@@ -51,21 +34,10 @@ function SharePostContent({
...
-      }
-    : {
-        href: `${post.sharedPost.commentsPermalink}?squad=${post.source.handle}&n=${post.source.name}`,
-        as: post.sharedPost.commentsPermalink,
-      };
-
   const openArticle = (e: React.MouseEvent) => {
     e.stopPropagation();
     onReadArticle();
+    console.log('on click open article go to link');
```
rebelchris, line 40: Log

packages/shared/src/hooks/useOnPostClick.ts
```diff
@@ -53,6 +53,7 @@ export default function useOnPostClick({
   return useMemo(
     () =>
       async ({ post, row, column, optional }): Promise<void> => {
+        console.log('tracking event: ', eventName);
```
rebelchris, line 56: Log

packages/shared/src/hooks/usePostContent.ts
```diff
@@ -35,11 +35,13 @@ const usePostContent = ({
   const { user } = useAuthContext();
   const { trackEvent } = useAnalyticsContext();
   const onPostClick = useOnPostClick({ origin });
-  const onReadArticle = () =>
-    onPostClick({
+  const onReadArticle = () => {
+    console.log('reading article');
```
rebelchris, line 39: Log

rebelchris (changes requested):
Please fix the logs removal before merging.
When merging please tripple check the analytics on prod as this is sensitive change

sshanzel:
@rebelchris still WIP that's why I have so many logs right now 😅

packages/shared/src/components/post/common/SharedPostLink.tsx
sshanzel: > Alternative would have been to import type from.
We have some examples of that in the codebase.

Not exactly sure what you mean here, can you give an example?

rebelchris:
@sshanzel Oh sorry totally missed it was in draft as i got an email for some reason 🤯

packages/shared/src/components/post/common/SharedPostLink.tsx
rebelchris: You can import types specifically, probably something we should leverage more:

```
import type { FilterOnboardingProps } from './FilterOnboarding';
```

sshanzel: This is actually importing the type as shown below. Reason why we updated it to become an exported type.

Screenshot 2023-12-12 at 5 55 01 PM

rebelchris (approved):
Keep in mind to double test on production 🙏

packages/shared/src/components/post/common/SharedPostLink.tsx
sshanzel: Ah, I understand the `import type` now. Looks to be something we can definitely apply more and more. We just probably need to have a rule when to use it. Like would it still be beneficial if you are already importing something from the same file, etc.

Discussion would be great, indeed.

PR #2404, opened 2023-12-12
fix: react rules of hooks

rebelchris:
- We abused the rules of hooks on this page a lot, which resulted in breaking it now.
- Decided to follow the right approach, confirmed it now works without breaking.

Screenshot 2023-12-12 at 10 04 19
Screenshot 2023-12-12 at 10 04 13

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-2040 #done

### DIFF
--- a/packages/webapp/components/layouts/ProfileLayout/index.tsx
+++ b/packages/webapp/components/layouts/ProfileLayout/index.tsx
@@ -78,16 +78,11 @@ export default function ProfileLayout({
     campaignKey: ReferralCampaignKey.Generic,
     enabled: initialProfile?.id === user?.id,
   });
-
-  if (!isFallback && !initialProfile) {
-    return <Custom404 />;
-  }
-
-  const selectedTab = tabs.findIndex((tab) => tab.path === router?.pathname);
-
+  const [twitterHandle, setTwitterHandle] = useState<string>();
+  const [githubHandle, setGithubHandle] = useState<string>();
+  const [hashnodeHandle, setHashnodeHandle] = useState<string>();
+  const [portfolioLink, setPortfolioLink] = useState<string>();
   const queryKey = ['profile', initialProfile?.id];
-  // @NOTE see https://dailydotdev.atlassian.net/l/cp/dK9h1zoM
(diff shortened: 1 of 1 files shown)


sshanzel (approved):
Definitely so much better now without all the comments.

Also, arguably we can store the handles in a single object since we only update it once for sanitization. This is optional though, not really part of the scope of this PR.

packages/webapp/components/layouts/ProfileLayout/index.tsx
```diff
@@ -96,14 +91,18 @@ export default function ProfileLayout({
...
   const profile = fetchedProfile ?? initialProfile;
-  const isCurrentUserProfile = profile && profile.id === user?.id;
-
+  useEffect(() => {
+    if (profile) {
+      const purify = DOMPurify(window);
+      setTwitterHandle(sanitizeOrNull(purify, profile.twitter));
+      setGithubHandle(sanitizeOrNull(purify, profile.github));
+      setHashnodeHandle(sanitizeOrNull(purify, profile.hashnode));
+      setPortfolioLink(sanitizeOrNull(purify, profile.portfolio));
```
capJavert, line 102: I don't think you need effect + state here, you can just `useMemo`.

capJavert (commented):
Nice, just one optimization to avoid re-renders.

packages/webapp/components/layouts/ProfileLayout/index.tsx
rebelchris: You mean memo on the profile or the actual states?

rebelchris: Hmm probably indeed, it's pretty much a edge case anyway not used that much.
I'll log it as a extra ticket if that's ok.

sshanzel: Oh, right. Yep. One less re-render.

omBratteng:
My face when something finally broke after we disregarded the warnings from the linter about hooks

PR #2408, opened 2023-12-12
feat: add block state for post option

capJavert:
- We were missing actual block state when user clicks "Don't show posts from..:" 
- We use `feedSettings` to check the state and show it
- Added option to "Show posts from" when source is blocked which remove filter eg. follows source

image
image

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-2013 #done

### DIFF
--- a/packages/shared/src/components/PostOptionsMenu.tsx
+++ b/packages/shared/src/components/PostOptionsMenu.tsx
@@ -1,4 +1,4 @@
-import React, { ReactElement, useContext } from 'react';
+import React, { ReactElement, useContext, useMemo } from 'react';
 import { Item } from '@dailydotdev/react-contexify';
 import dynamic from 'next/dynamic';
 import { useQueryClient } from '@tanstack/react-query';
@@ -208,6 +208,21 @@ export default function PostOptionsMenu({
     );
   };
 
+  const onUnblockSource = async (): Promise<void> => {
+    const { successful } = await onFollowSource({
+      source: post?.source,
+      requireLogin: true,
+    });
+
+    if (!successful) {
+      return;
+    }
+
+    displayToast(`${post?.source?.name} unblocked`, {
+      subject: ToastSubject.Feed,
+    });
+  };
+
   const onBlockTag = async (tag: string): Promise<void> => {
     const { successful } = await onBlockTags({
       tags: [tag],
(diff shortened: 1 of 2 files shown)


rebelchris (approved):
Nice work!

packages/shared/src/components/PostOptionsMenu.tsx
```diff
@@ -208,6 +208,21 @@ export default function PostOptionsMenu({
...
   };
 
+  const onUnblockSource = async (): Promise<void> => {
+    const { successful } = await onFollowSource({
+      source: post?.source,
+      requireLogin: true,
+    });
+
+    if (!successful) {
+      return;
```
kopancek, line 218: Should we display an error to the user if it does not succeed?

packages/shared/src/components/PostOptionsMenu.tsx
```diff
@@ -208,6 +208,21 @@ export default function PostOptionsMenu({
...
+    }
+
+    displayToast(`${post?.source?.name} unblocked`, {
```
kopancek, line 221: The label on the button says `Show posts from`, but here we say `unblocked`. Should we change the label or the description here to align?

packages/shared/src/components/PostOptionsMenu.tsx
```diff
@@ -278,8 +299,10 @@ export default function PostOptionsMenu({
     },
     {
       icon: <MenuIcon Icon={BlockIcon} />,
```
kopancek, line 301: I would say we should have 2 different icons, one for blocking and an opposite for unblocking. Otherwise it's quite confusing to the user, that we say `Show posts from`, but show the same icon as for `Not interested in...` actions.

kopancek (commented):
Great functionality. A few comments to hopefully make it even better :)

packages/shared/src/components/PostOptionsMenu.tsx
rebelchris: I thought about leaving the same feedback, but see we already always use the same even for existing block/unblocking actions.
Can probably be something to raise separate to design team.

packages/shared/src/components/PostOptionsMenu.tsx
capJavert: Not succeeding in this case can also be just user not logged in. Also follows the same logic as block so don't want to complicate it.

packages/shared/src/components/PostOptionsMenu.tsx
capJavert: Also follows the same spec as "Don't show" => blocked. We wanted to make a quick win here and will update this if needed as papercut.

packages/shared/src/components/PostOptionsMenu.tsx
sshanzel: Also, this is a pre-existing issue where the state is not changed even after the action.

PR #2409, opened 2023-12-12
refactor: feed card to experiment bookmark button

sshanzel:
## Changes
- Experiment with the card whether to include the Bookmark button or not.
- Refactor `ActionButtons` to not use `ConditionalWrapper` when unnecessary.
- Also removed the `children` props as it is not used.
- Experiment key is `bookmark_on_card`. Values is either true or false.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-2052 #done


packages/shared/src/components/Feed.tsx
```diff
@@ -316,6 +316,18 @@ export default function Feed<T>({
...
+    onBookmark({
+      post,
+      origin,
+      opts: {
+        row,
+        column,
+        columns: virtualizedNumCards,
+        ...feedAnalyticsExtra(feedName, ranking),
+      },
+    });
```
rebelchris, line 329: We already have `toggleBookmark` right?
This is preferred to use in this case.

packages/shared/src/components/Feed.tsx
```diff
@@ -376,6 +388,7 @@ export default function Feed<T>({
             user={user}
             feedName={feedName}
             ranking={ranking}
+            onBookmark={onCardBookmark}
```
rebelchris, line 391: Probably should be `toggleBookmark` to be inline

packages/shared/src/components/cards/ActionButtons.tsx
```diff
@@ -25,10 +28,10 @@ export interface ActionButtonsProps {
   onUpvoteClick?: (post: Post) => unknown;
   onCommentClick?: (post: Post) => unknown;
   onShare?: (post: Post) => unknown;
+  onBookmark?: (post: Post) => unknown;
```
rebelchris, line 31: Perhaps we should keep the naming similar to `onBookmarkClick`

rebelchris (changes requested):
Nice work, mainly some cleanup on naming.
And think you can re-use the existing `toggleBookmark` that's already on the feed.

I'd also like to run a e2e test locally.
wiIll do so after the quiz

packages/shared/src/components/Feed.tsx
sshanzel: Yes, that is used here, was just renamed as `onBookmark`.

packages/shared/src/components/Feed.tsx
sshanzel: As mentioned above, this one uses the `toggleBookmark`.

packages/shared/src/components/Feed.tsx
rebelchris: Why did you rename it, the actual action is toggling the bookmark that seemed more explicit no?

packages/shared/src/components/cards/ActionButtons.tsx
sshanzel: Cool, I based this off the `onShare` which didn't have the `Click` part, but we can do that here.

packages/shared/src/components/Feed.tsx
sshanzel: I did not renamed it. Has been defined as an alias of that here..

packages/shared/src/components/cards/ActionButtons.tsx
```diff
@@ -106,9 +106,8 @@ export default function ActionButtons({
   return (
     <div
       className={classNames(
-        styles.actionButtons,
-        'flex flex-row items-center',
-        insaneMode && 'justify-between',
+        'flex flex-row items-center justify-between',
+        !insaneMode && !bookmarkOnCard && 'mx-4',
```
rebelchris, line 110: It looks like we always need the mx-4 actually.

rebelchris (changes requested):

rebelchris (approved):
Looks good!

PR #2410, opened 2023-12-12
feat: add shortcuts events

capJavert:
- Implementing shortcuts events

TODO
- [x] check when `save shortcut access` should fire, [thread](https://dailydotdev.slack.com/archives/C061T644XV3/p1702390607014719) 

## Events

| New | impression | target_type: shortcuts |
| New | click | target_type: shortcuts |
| New | revoke shortcut access | target_type: shortcuts |
| New | save shortcut access | target_type: shortcuts |
| New | open shortcut config | target_type: shortcuts |

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-2013 #done


capJavert:
Ready for review, will update the last event when I get feedback from product.

packages/extension/src/newtab/ShortcutLinks.tsx
```diff
@@ -29,6 +42,21 @@ export default function ShortcutLinks({
...
+      return;
+    }
+
+    mountedRef.current = true;
+
+    trackEvent({
+      event_name: AnalyticsEvent.Impression,
+      target_type: TargetType.Shortcuts,
+    });
+  }, [trackEvent]);
```
rebelchris, line 58: So far we always just avoid the `trackEvent` I totally agree with this though.
Wondering if we should make this a custom `useOnceOffImpression` hook for now?

Just so it becomes easier to maintain if we want to expend?
Totally optional though.

rebelchris (commented):
Minor comments.
I raised some concerns on the taxonomy.
It feels like we could get more info by knowing if the shortcuts are manual or not.

packages/extension/src/newtab/ShortcutLinks.spec.tsx
```diff
@@ -212,6 +229,11 @@ describe('shortcut links component', () => {
 
     const addShortcuts = await screen.findByText('Add shortcuts');
     expect(addShortcuts).toBeVisible();
+
+    expect(trackEvent).toHaveBeenCalledWith({
+      event_name: AnalyticsEvent.RevokeShortcutAccess,
+      target_type: TargetType.Shortcuts,
+    });
```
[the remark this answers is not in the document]
capJavert: 🤣

packages/extension/src/newtab/ShortcutLinks.tsx
capJavert: It's a tricky subject. I personally am not a fan of those because they hide implementation detail. Probelm is does not strictly mean it will fire once. It will fire each mount which is once per component lifecycle, but if component unmounts (lifecycle ends) and mounts again it will fire again. 

Also with react concepts as concurrent rendering the above might also not be "once off" as we think.

capJavert: Basically [useEffectEvent](https://react.dev/learn/separating-events-from-effects#declaring-an-effect-event) which would solve reference for `trackEvent` and the we could just use `useEffect` without dependencies.

packages/extension/src/newtab/ShortcutLinks.tsx
```diff
@@ -29,6 +42,21 @@ export default function ShortcutLinks({
     onSaveChanges,
   } = useShortcutLinks();
 
+  const mountedRef = useRef(false);
+
+  useEffect(() => {
```
idoshamun, line 47: Shouldn't we check for showTopSites? Otherwise, we may track this impression without actually showing it

capJavert: It says in taxonomy sheet it should fire when component renders?

capJavert: Do you want it only when some shortcuts are actually rendered?

rebelchris: I raised this question on taxonomy as well :) 
It basically goes for everything, do we want to track it in general or only for custom, browser ones, etc?

idoshamun: Sorry my question was confusing. I meant we need to make sure we actually render the shortcuts. Because I see right after the `useEffect` a condition to make sure we should render it

capJavert: Got it, fixed.

sshanzel (approved):
Looks great!

nensidosari (approved):
Looks good to me as well! Nothing to add.

packages/extension/src/newtab/ShortcutLinks.spec.tsx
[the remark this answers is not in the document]
sshanzel: Oh, nice. I forgot we actually discussed new events should be tested 😆 

Well done 👏

rebelchris (approved):
Amazing

capJavert:
Verified and merging so we can release with current extension today.

idoshamun (approved):

PR #2411, opened 2023-12-12
fix: move max and min height to extend rather then override

nensidosari:
move max and min height to extend rather then override

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done


rebelchris (approved):
Thanks,
Some background I checked with Nensi and it felt we used `overwrite` instead of `extends` wrongly here.
Surely we want existing tailwind min-max height :)

So this retro fixes it.
(with small change it might break some things or add more space now 😅)

capJavert (approved):
Makes sense 👍

PR #2413, opened 2023-12-12
fix: the reverse of default state

rebelchris:
- Persist the don't show as the default should be true
- **Note: If the default was meant to be false for video content disregard this PR**

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done


packages/shared/src/components/PostOptionsMenu.tsx
```diff
@@ -230,9 +230,9 @@ export default function PostOptionsMenu({
   };
   const video = advancedSettings?.find(({ title }) => title === 'Videos');
   const onToggleVideo = async () => {
-    await onToggleSettings(video.id, false);
+    await onToggleSettings(video.id, true);
     await showMessageAndRemovePost(`⛔️ Video content blocked`, postIndex, () =>
-      onToggleSettings(video.id, true),
+      onToggleSettings(video.id, false),
```
omBratteng, line 235: I need to double check, but I am sure that it was correct before.
The default value in database is `true` when it is not set to anything, i.e. video content is allowed.
So if you disable video content, you set the `enabled` column in `feed_advanced_settings` table to `false`.

rebelchris: Ah yes but onToggleSettings reverses what you pass afaik (from what i tested) it accepts a default state

sshanzel: This value is the `defaultEnabledState` and should not be changed. Maybe that parameter should not even be there in the first place, I believe we had that for some experiment in the past but the value there should reflect what we have in our DB for this entity which is `true`. That is the reason why this was strongly-typed.

sshanzel: The whole issue here is that, it is a toggle, and the post options menu does not reflect the correct state, so even when it says it should block the option, it would then enable it if you re-click it.

rebelchris: It also didnt reflect right in the database though.
It added it as "true" because it didnt exist yet for this user.

Secondary the option disappears from the post options menu when false so think thats fine right?

Maybe we can sync on it today?

sshanzel: > It added it as "true" because it didnt exist yet for this user.

That is the default state from the AdvancedSettings not necessarily for the user.

sshanzel (approved):
Thanks for the fix!

packages/shared/src/components/PostOptionsMenu.tsx
rebelchris: Validated with Lee this was the right fix.
CC: @omBratteng

sshanzel: Just to note: the issue was after clicking the post options menu to block Videos, then clicked the Undo, it would fail.

PR #2415, opened 2023-12-13
fix: ad spot adjustment under experiment

sshanzel:
## Changes
- Fix the ad spot placement when the user is under our feed layout experiment.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-2027 #done


packages/shared/src/components/Feed.tsx
```diff
@@ -123,11 +123,12 @@ export default function Feed<T>({
   const insaneMode = !forceCardMode && listMode;
   const numCards = currentSettings.numCards[spaciness ?? 'eco'];
   const isSquadFeed = feedName === 'squad';
+  const { shouldUseFeedLayoutV1 } = useFeedLayout({ feedName });
```
rebelchris, line 126: Why do you need to pass it here, the context will have it right?

rebelchris (changes requested):
we dont need to pass the feedmame i think.
Otherwise looks great

packages/shared/src/components/Feed.tsx
nensidosari: Yup the context should have it, that's why we removed it from the hook props

packages/shared/src/components/Feed.tsx
```diff
@@ -123,11 +123,12 @@ export default function Feed<T>({
...
   const { items, updatePost, removePost, fetchPage, canFetchMore, emptyFeed } =
     useFeed(
       feedQueryKey,
       currentSettings.pageSize,
-      isSquadFeed ? 3 : currentSettings.adSpot,
+      isSquadFeed || shouldUseFeedLayoutV1 ? 3 : currentSettings.adSpot,
```
nensidosari, line 131:
```suggestion
      isSquadFeed || shouldUseFeedLayoutV1 ? 2 : currentSettings.adSpot,
```

This should also be adjusted to use the 3rd spot - which in this case would be nr 2. 

Nr.2 is the default in the settings so just check to align these 🙏

packages/shared/src/hooks/useFeedLayout.ts
```diff
@@ -4,18 +4,25 @@ import { feature } from '../lib/featureManagement';
...
+  feedName?: AllFeedPages;
+}
 
 interface UseFeedLayout {
   shouldUseFeedLayoutV1: boolean;
 }
 
-export const useFeedLayout = (): UseFeedLayout => {
+export const useFeedLayout = ({
+  feedName: feedNameProp,
```
nensidosari, line 18: These shouldn't be added based on the comment above about feedName being taken from context.

nensidosari (changes requested):

packages/shared/src/hooks/useFeedLayout.ts
sshanzel: Unfortunately the Context starts from the components after the `Feed`. So from where we called it, it is undefined.

packages/shared/src/components/Feed.tsx
sshanzel: This component is not yet under that context, so it is not familiar with the value.

packages/shared/src/components/Feed.tsx
sshanzel: Thanks for the suggestion! 🚀

packages/shared/src/hooks/useFeedLayout.ts
rebelchris: In that case we better of moving the context perhaps?
Which area is not under it, basically all feeds should be.

This feels like a very weird fix now, either it's automagical or we pass it.

sshanzel: Part of the requirements of the context is the `items` object, moving it would involve refactoring a lot of the logic and drilling at least 6 props to this Feed component. Not sure if would be beneficial to do that amount of work and changes for this experiment.

nensidosari (approved):
Looks great! No other issues on my side :D

packages/shared/src/hooks/useFeedLayout.ts
rebelchris: I see indeed, that's a bit sad.
Anyway maybe with leanfeed we can refactor it a bit :)

sshanzel: Yes, agree with that. Hopefully we can introduce a context when the cards/list items can utilize to avoid all the drillings.

PR #2419, opened 2023-12-13
chore: refactor profile page and account settings to ButtonV2

kopancek:

First iteration of moving to ButtonV2 in our components, more to come...

This one tackles ProfileButton, Profile Page and Account Settings pages.

Related: https://dailydotdev.atlassian.net/browse/WT-2017 

### Screenshots

Screenshot 2023-12-13 at 10 36 38
<br/>
Screenshot 2023-12-13 at 11 42 40

Screenshot 2023-12-13 at 10 28 26

Screenshot 2023-12-13 at 11 35 59

Screenshot 2023-12-13 at 11 34 58
Screenshot 2023-12-13 at 14 53 18

- [x] MobileL (420px)
- [x] Tablet (656px)
- [x] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [x] Android

### DIFF
--- a/packages/shared/src/components/profile/AccountDangerZone.tsx
+++ b/packages/shared/src/components/profile/AccountDangerZone.tsx
@@ -1,6 +1,11 @@
 import classNames from 'classnames';
 import React, { ReactElement, ReactNode } from 'react';
-import { Button, ButtonSize } from '../buttons/Button';
+import {
+  Button,
+  ButtonColor,
+  ButtonSize,
+  ButtonVariant,
+} from '../buttons/ButtonV2';
 
 interface AccountDangerZoneProps {
   onDelete: () => void;
@@ -46,8 +51,10 @@ function AccountDangerZone({
       {children}
       <Button
         onClick={onDelete}
-        buttonSize={ButtonSize.Small}
-        className="self-start mt-6 btn-primary-ketchup text-theme-label-primary"
+        size={ButtonSize.Small}
+        variant={ButtonVariant.Primary}
+        color={ButtonColor.Ketchup}
+        className="self-start mt-6 text-theme-label-primary"
       >
         Delete account
       </Button>
--- a/packages/shared/src/components/profile/AccountDangerZone.tsx
+++ b/packages/shared/src/components/profile/AccountDangerZone.tsx
@@ -54,7 +54,7 @@ function AccountDangerZone({
         size={ButtonSize.Small}
         variant={ButtonVariant.Primary}
         color={ButtonColor.Ketchup}
-        className="self-start mt-6 text-theme-label-primary"
+        className="self-start mt-6"
       >
         Delete account
       </Button>
--- a/packages/shared/src/components/profile/ProfileButton.tsx
+++ b/packages/shared/src/components/profile/ProfileButton.tsx
@@ -5,7 +5,7 @@ import AuthContext from '../../contexts/AuthContext';
 import { ProfilePicture } from '../ProfilePicture';
 import { SimpleTooltip } from '../tooltips/SimpleTooltip';
 import SettingsIcon from '../icons/Settings';
-import { Button } from '../buttons/Button';
+import { Button, ButtonSize, ButtonVariant } from '../buttons/ButtonV2';
 import { useInteractivePopup } from '../../hooks/utils/useInteractivePopup';
 
 const ProfileMenu = dynamic(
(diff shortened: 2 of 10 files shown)


packages/shared/src/components/profile/AccountDangerZone.tsx
```diff
@@ -46,8 +51,10 @@ function AccountDangerZone({
       {children}
       <Button
         onClick={onDelete}
-        buttonSize={ButtonSize.Small}
-        className="self-start mt-6 btn-primary-ketchup text-theme-label-primary"
+        size={ButtonSize.Small}
+        variant={ButtonVariant.Primary}
+        color={ButtonColor.Ketchup}
+        className="self-start mt-6 text-theme-label-primary"
```
nensidosari, line 57: Maybe I missed something like this discussed in buttonV2, or something like this was out of scope, so just checking. 
For `text-theme-label-primary` wouldn't we think a button with a certain color/variant combination has a set label theme color?
Could be something that we can improve later on, but I believe it would be great if a default color for a certain combination is already set and consumers don't have to define that all the time.

No need to take action on this on this PR just raising for discussion and further work improvements we can do on the button.

packages/webapp/components/layouts/AccountLayout/Security/index.tsx
```diff
@@ -209,16 +210,17 @@ function AccountSecurityDefault({
...
+            className="mt-6 w-fit"
             onClick={() => onSwitchDisplay(Display.ChangeEmail)}
           >
             Change email
           </Button>
         )}
       </AccountContentSection>
       <AccountLoginSection
-        className={{ button: 'btn-primary' }}
+        variant={ButtonVariant.Primary}
```
nensidosari, line 223: Would this be best as buttonVariant?

nensidosari (approved):
Looks great, just one comment & suggestion on my side in general idea, but definitely non blocking.

packages/shared/src/components/profile/AccountDangerZone.tsx
kopancek: Yes, that's already the case. In this case, it was overriden to white font, so I kept the same visuals as we had before. But agree, ideally we do not add such a classname and keep the default label color.

packages/webapp/components/layouts/AccountLayout/Security/index.tsx
kopancek: Good point, renaming :)

packages/shared/src/components/profile/AccountDangerZone.tsx
rebelchris: Ah that's actually a good catch.
In the beginning the red color was always with white text we overruled that for accesibility concerns.
So technically you shouldn't have custom colors on buttons.

But maybe double check with design and give them a pre-after screenshot?

packages/shared/src/components/profile/ProfileButton.tsx
```diff
@@ -14,40 +14,38 @@ const ProfileMenu = dynamic(
...
         <SimpleTooltip placement="left" content="Profile settings">
-          <button
-            type="button"
+          <Button
+            variant={ButtonVariant.Float}
+            size={ButtonSize.Small}
             className={classNames(
-              'items-center p-0 ml-0.5 font-bold no-underline rounded-lg border-none cursor-pointer text-theme-label-primary bg-theme-bg-secondary typo-callout focus-outline',
-              className ?? 'flex',
+              'items-center !p-0 ml-0.5 text-theme-label-primary gap-2',
```
rebelchris, line 41: This might be an exception because it's so custom. what's the default color we add on float buttons?

packages/shared/src/components/profile/ProfileButton.tsx
```diff
@@ -14,40 +14,38 @@ const ProfileMenu = dynamic(
...
+              className,
             )}
             onClick={wrapHandler(() => onUpdate(!isOpen))}
           >
-            <span className="hidden laptop:block mr-2 ml-3">
-              {user.reputation ?? 0}
-            </span>
+            <span className="block ml-3">{user.reputation ?? 0}</span>
```
rebelchris, line 46: Why did you remove the responsive styles here?

packages/shared/src/components/profile/ProfileButton.tsx
```diff
@@ -14,40 +14,38 @@ const ProfileMenu = dynamic(
...
-          className="btn btn-tertiary"
+          variant={ButtonVariant.Tertiary}
           onClick={wrapHandler(() => onUpdate(!isOpen))}
           icon={<SettingsIcon />}
         />
       ) : (
         <SimpleTooltip placement="left" content="Profile settings">
-          <button
-            type="button"
+          <Button
```
rebelchris, line 37: After thinking, maybe this shouldn't even be a button tbh 😅
It's like a custom element with a onClick really.
But open to keep it for simplicity

packages/webapp/components/layouts/AccountLayout/EmailForm.tsx
```diff
@@ -55,7 +59,11 @@ function EmailForm({
           label={passwordProps?.label || 'Password'}
         />
       )}
-      <Button className="mt-6 ml-auto w-fit bg-theme-color-cabbage btn-quaternary">
+      <Button
+        className="mt-6 ml-auto w-fit text-theme-label-primary"
```
rebelchris, line 63: I don't think the text override is needed, but maybe also tripple check with design they ok with it.

rebelchris (changes requested):
Minor comments, but overall great work.

You might find a couple places where we in existing places override text colors, imho this shouldn't happen and it's bad practice, we can probably just stick with button guidelines, but in doubt double-check with design team.

packages/shared/src/components/profile/ProfileButton.tsx
kopancek: The default is the same as the bell icon on the notifications button. So it would be greyish - #a8b3cf.

packages/webapp/components/layouts/AccountLayout/EmailForm.tsx
kopancek: I checked and the result is we should keep the default: https://dailydotdev.slack.com/archives/C0381BQUEN9/p1702475962350259?thread_ts=1702475717.398979&cid=C0381BQUEN9

rebelchris: Thanks for checking, I thought so :)

packages/shared/src/components/profile/ProfileButton.tsx
kopancek: Here is another response from design, we should keep it white: https://dailydotdev.slack.com/archives/C0381BQUEN9/p1702476773654429?thread_ts=1702475717.398979&cid=C0381BQUEN9

As for keeping it as a button or not, I think keeping the Button has better accessibility than having a `<span with onClick>`, e.g. for screen readers it's much easier if it's a `<button` if it's a clickable element. So my vote is for keeping it as a button.

rebelchris: Ah yes but I mean using a `button` rather then our `Button` component?
I do think this is where `none` variant should actually be used almost.

rebelchris: Let's do that, it avoids issues down the line if we make buttonv2 even more strict etc.

rebelchris (approved):
Looks good :)

PR #2421, opened 2023-12-13
feat: when building locally sw files get generated in public

capJavert:
Just annoying so wanted to add.


kopancek (approved):
Nice :shipit:

PR #2422, opened 2023-12-13
feat: tree shaking barrel files webapp

capJavert:

In last few weeks/moths (since react migration) I noticed our bundle size was rising incrementally with each PR. In partial this might be for not using lazy loading for some components and probably due to barrel files usage in project. But while looking at it I noticed we might be able to save on those KBs with enabling tree shaking of all unused code. 

Read more at: https://webpack.js.org/guides/tree-shaking/#mark-the-file-as-side-effect-free

Basically by declaring no side effects all unused code is remove from final bundle. I don't think we have any side effects because at the end our shared is bundled during webapp build so actually all the code needed is pulled there. 

Only side effect I could find was `lazysizes` import which at this point is legacy and we already mentioned we should remove it in favor of native lazy loading. 🤞 

I also added `.css` to side effects for now since we use global stylesheet.

Available on [preview](https://preview3.app.daily.dev/) if anyone wants take a look.

Difference (shared js bundle, affects all pages):
Current main: 
image

This PR:
image

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done

### DIFF
--- a/packages/shared/package.json
+++ b/packages/shared/package.json
@@ -130,5 +130,5 @@
     "web-vitals": "^3.5.0",
     "webextension-polyfill-ts": "^0.22.0"
   },
-  "sideEffects": false
+  "sideEffects": ["./src/lib/lazysizesImport.ts"]
 }
--- a/packages/shared/package.json
+++ b/packages/shared/package.json
@@ -130,5 +130,5 @@
     "web-vitals": "^3.5.0",
     "webextension-polyfill-ts": "^0.22.0"
   },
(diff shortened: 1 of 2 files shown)


idoshamun (approved):
Nice gains!!

rebelchris (approved):
Amazing how easy some wins are :)

capJavert:
Gonna merge this now, I'll let the channel know to look out for any issues, but I think it should be good.

PR #2425, opened 2023-12-13
fix: do not render PostUpvotesCommentsCount if there are no comments or upvotes

omBratteng:
## Changes

Stops rendering an empty div if there are no comments or upvotes, reducing the black gap

<table>
<tr>
 <td>Before
 <td>After
<tr>
 <td>

 <td>
</table>


rebelchris (approved):

sshanzel (approved):
Looks great! 🚀

PR #2429, opened 2023-12-14
fix: margins shared content

nensidosari:
## Changes
- fix margins in shared content

WT-2065 #done


omBratteng (commented):
I just one of these is enough, though I will double check later

nensidosari:
> I just one of these is enough, though I will double check later

i checked and for normal shared post as well this was an issue, that's why I added at the other one as well.

omBratteng (approved):
Of course, they're used two different places.

PR #2432, opened 2023-12-15
fix: check if shared post and use sharedPost permalink in insane mode

omBratteng:
- When in insane mode, it checks if the post is of type `shared`, and then gets the permalink from the sharedPost

### DIFF
--- a/packages/shared/src/components/cards/ActionButtons.tsx
+++ b/packages/shared/src/components/cards/ActionButtons.tsx
@@ -1,7 +1,7 @@
 import React, { ReactElement } from 'react';
 import classNames from 'classnames';
 import dynamic from 'next/dynamic';
-import { Post, UserPostVote } from '../../graphql/posts';
+import { Post, PostType, UserPostVote } from '../../graphql/posts';
 import InteractionCounter from '../InteractionCounter';
 import { QuaternaryButton } from '../buttons/QuaternaryButton';
 import UpvoteIcon from '../icons/Upvote';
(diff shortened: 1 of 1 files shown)


capJavert (approved):

packages/shared/src/components/cards/ActionButtons.tsx
```diff
@@ -163,7 +163,11 @@ export default function ActionButtons({
         >
           <ReadArticleButton
             className="mr-2 btn-primary"
-            href={post.permalink}
+            href={
+              post.type === PostType.Share
+                ? post.sharedPost.permalink
+                : post.permalink
+            }
```
rebelchris, line 170: I think we also need to change the openNewTab if we look at the `SharePostContent` handler for instance:
```
href={
                isSharedPostSquadPost
                  ? post.sharedPost.commentsPermalink
                  : post.sharedPost.permalink
              }
              openNewTab={isSharedPostSquadPost ? false : openNewTab}
```

rebelchris (approved):
Not sure, but think we need to ensure the openNewTab (it might be always good already though, I can't reproduce the original issue)

packages/shared/src/components/cards/ActionButtons.tsx
omBratteng: That's when for when the shared post is to a squad post. Don't think this is an issue here, haven't encountered it. If I do, I'll raise a PR to fix it.

PR #2433, opened 2023-12-18
feat: render watch time as 1h 23m

omBratteng:
## Changes

Creates a new function that formats the readTime as [discussed here](https://dailydotdev.slack.com/archives/C059GQ5RFP0/p1702562731056579).

I decided to also do it for read time, in the extreme cases the read time is over 60 minutes


rebelchris (approved):
Looks good to me!

capJavert (approved):
Not sure why `date-fns` does not have a util to do this 😆 .. though it was this thing but not really.. https://date-fns.org/v3.0.0/docs/formatDuration

PR #2434, opened 2023-12-18
fix: insane mode read post button

omBratteng:
## Changes

It just hides the read post button if the post is a freeform, as clicking on the card opens it. I was thinking about mimicking that on the button, but thought removing it made more sense. Let me know if I should re-introduce it.

Also updated the target link of the button if the article is a shared post that points to another squad.

### DIFF
--- a/packages/shared/src/components/cards/ActionButtons.tsx
+++ b/packages/shared/src/components/cards/ActionButtons.tsx
@@ -166,7 +166,7 @@ export default function ActionButtons({
             className="mr-2 btn-primary"
             href={getReadArticleLink(post)}
             onClick={onReadArticleClick}
-            openNewTab={isSharedPostSquadPost(post) ? false : openNewTab}
+            openNewTab={!isSharedPostSquadPost(post) && openNewTab}
           />
           <OptionsButton
             className={visibleOnGroupHover}
--- a/packages/shared/src/components/cards/ActionButtons.tsx
+++ b/packages/shared/src/components/cards/ActionButtons.tsx
@@ -1,12 +1,7 @@
 import React, { ReactElement } from 'react';
 import classNames from 'classnames';
 import dynamic from 'next/dynamic';
-import {
-  Post,
-  PostType,
-  UserPostVote,
-  isInternalReadType,
-} from '../../graphql/posts';
+import { Post, UserPostVote, isInternalReadType } from '../../graphql/posts';
 import InteractionCounter from '../InteractionCounter';
 import { QuaternaryButton } from '../buttons/QuaternaryButton';
 import UpvoteIcon from '../icons/Upvote';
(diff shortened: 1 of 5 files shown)


packages/shared/src/components/cards/ActionButtons.tsx
```diff
@@ -157,19 +170,15 @@ export default function ActionButtons({
         </SimpleTooltip>
         {insaneMode && lastActions}
       </ConditionalWrapper>
-      {insaneMode ? (
+      {insaneMode && post.type !== PostType.Freeform ? (
```
capJavert, line 173: We can probably look at `internalReadTypes` because all of those don't need read button. For example when collections is merged this would need to be updated.

packages/shared/src/components/cards/ActionButtons.tsx
```diff
@@ -103,6 +104,18 @@ export default function ActionButtons({
...
+    post.sharedPost?.source.type === SourceType.Squad;
+
+  const insaneReadArticleLink = () => {
+    if (post.type === PostType.Share) {
+      return isSharedPostSquadPost
+        ? post.sharedPost.commentsPermalink
+        : post.sharedPost.permalink;
+    }
+    return post.permalink;
+  };
```
capJavert, line 117: This looks like something that we can have as util and then use it with arguments vs having an inline function in component. `getReadArticleLink({ post })`

packages/shared/src/components/cards/ActionButtons.tsx
omBratteng: Good idea, will update it.

omBratteng: Refactored it now.

packages/shared/src/components/cards/ActionButtons.tsx
```diff
@@ -157,19 +158,15 @@ export default function ActionButtons({
...
             className="mr-2 btn-primary"
-            href={
-              post.type === PostType.Share
-                ? post.sharedPost.permalink
-                : post.permalink
-            }
+            href={getReadArticleLink(post)}
             onClick={onReadArticleClick}
-            openNewTab={openNewTab}
+            openNewTab={isSharedPostSquadPost(post) ? false : openNewTab}
```
kopancek, line 169: I think this can be simplified to
```suggestion
            openNewTab={!isSharedPostSquadPost(post) && openNewTab}
```

kopancek (approved):
Looks great, suggested 1 improvement to simplify the conditional logic

rebelchris:
Tbh I don't think freeform can show up on your feed (as far as I remember?)
So this could never be applied?

Did you manage to see freeform on the main feed? If so never mind my comment.

rebelchris (approved):
Nice fix, thanks for that :)

capJavert (approved):

PR #2435, opened 2023-12-18
refactor: use ButtonV2 in modals, empty screens and search bar

kopancek:
## Changes

Use ButtonV2 component instead of Button in more places (102 files left to convert).

Deleted the `ModalCloseButton` as it was a duplicate of `ModalClose`, that was used in more places.

### Screenshots

Screenshot 2023-12-15 at 09 46 31
<br/>
Screenshot 2023-12-15 at 12 07 26
<br/>
Screenshot 2023-12-15 at 12 10 04
<br/>
Screenshot 2023-12-15 at 12 11 40
<br/>
Screenshot 2023-12-15 at 12 12 00
<br/>
Screenshot 2023-12-15 at 12 16 47
<br/>
Screenshot 2023-12-15 at 12 17 52
<br/>
Screenshot 2023-12-15 at 12 22 32
<br/>
Screenshot 2023-12-15 at 12 26 59
<br/>
Screenshot 2023-12-15 at 13 26 10
<br/>
Screenshot 2023-12-15 at 14 02 12
<br/>
Screenshot 2023-12-15 at 14 11 44
<br/>
Screenshot 2023-12-18 at 09 37 37
<br/>
Screenshot 2023-12-18 at 09 42 41
<br/>
Screenshot 2023-12-18 at 09 43 47
<br/>
Screenshot 2023-12-18 at 10 36 32
<br/>
Screenshot 2023-12-18 at 10 41 53
<br/>
Screenshot 2023-12-18 at 10 46 16
<br/>
Screenshot 2023-12-18 at 10 49 40
<br/>
Screenshot 2023-12-18 at 10 50 11
<br/>
Screenshot 2023-12-18 at 13 10 11
<br/>
Screenshot 2023-12-18 at 13 15 23
<br/>
Screenshot 2023-12-18 at 13 20 40
<br/>
Screenshot 2023-12-18 at 13 34 39
<br/>
Screenshot 2023-12-18 at 13 35 15
<br/>
Screenshot 2023-12-18 at 13 36 31
<br/>
Screenshot 2023-12-18 at 14 56 57
<br/>
Screenshot 2023-12-18 at 15 06 09
<br/>
Screenshot 2023-12-18 at 15 17 58
<br/>
Screenshot 2023-12-18 at 15 18 54

- [X] MobileL (420px)
- [X] Tablet (656px)
- [X] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android


packages/shared/src/components/checklist/ChecklistCard.tsx
```diff
@@ -30,8 +31,9 @@ const ChecklistCard = ({
           <p className="text-white typo-callout">{description}</p>
           {typeof onRequestClose === 'function' && (
             <CloseButton
-              buttonSize={ButtonSize.Small}
-              className="top-3 right-3 text-white border-white !absolute btn-secondary"
+              size={ButtonSize.Small}
+              variant={ButtonVariant.Secondary}
+              className="top-3 right-3 text-white border-white !absolute"
```
rebelchris, line 36: We have position added for `!absolute` right?
Also the border overrides and text are needed on this modal?

packages/shared/src/components/modals/SquadTourModal.tsx
```diff
@@ -25,8 +25,9 @@ function SquadTourModal({
     >
       <SquadTour onClose={onModalClose} />
       <CloseButton
-        buttonSize={ButtonSize.Small}
-        className="top-3 right-3 !absolute !btn-secondary"
+        size={ButtonSize.Small}
+        variant={ButtonVariant.Secondary}
+        className="top-3 right-3 !absolute"
```
rebelchris, line 30: Same here, think we should leverage the position prop.

packages/shared/src/components/modals/VideoModal.tsx
```diff
@@ -26,10 +25,9 @@ export default function VideoModal({
       {...props}
     >
       <CloseButton
-        buttonSize={ButtonSize.Small}
-        className="top-3 right-3"
+        size={ButtonSize.Small}
+        className="top-3 right-3 !absolute"
```
rebelchris, line 29: Here as well.

packages/shared/src/components/modals/common/ModalClose.tsx
```diff
@@ -19,11 +13,9 @@ function ModalCloseComponent(
   return (
     <CloseButton
       {...props}
-      buttonSize={buttonSize}
       onClick={onClick}
       ref={ref}
-      className={classNames('right-2 z-1', className)}
-      style={{ position: 'absolute', ...style }}
+      className={classNames('right-2 z-1 !absolute', className)}
```
rebelchris, line 18: And here

packages/shared/src/components/modals/referral/GenericReferralModal.tsx
```diff
@@ -36,9 +36,9 @@ function GenericReferralModal({
     <Modal {...props} onRequestClose={onRequestClose} size={ModalSize.Small}>
       <CloseButton
         onClick={onRequestClose}
-        position="absolute"
-        buttonSize={ButtonSize.Small}
-        className="top-4 right-4 z-2 !btn-secondary"
+        variant={ButtonVariant.Secondary}
+        size={ButtonSize.Small}
+        className="top-4 right-4 z-2 !absolute"
```
rebelchris, line 41: Here as well.

packages/shared/src/components/modals/referral/SearchReferralModal.tsx
```diff
@@ -65,8 +64,8 @@ function SearchReferralModal({
     >
       <CloseButton
         onClick={handleRequestClose}
-        position="absolute"
-        className="top-3 right-3 z-1 !btn-secondary"
+        variant={ButtonVariant.Secondary}
+        className="top-3 right-3 z-1 !absolute"
```
rebelchris, line 68: Another one

packages/shared/src/components/modals/squads/SquadPromotionModal.tsx
```diff
@@ -31,8 +31,9 @@ export function SquadPromotionModal({
     >
       <PromotionTour onClose={onRequestClose} source={squad} />
       <CloseButton
-        buttonSize={ButtonSize.Small}
-        className="top-3 right-3 !absolute !btn-secondary"
+        size={ButtonSize.Small}
+        variant={ButtonVariant.Secondary}
+        className="top-3 right-3 !absolute"
```
rebelchris, line 36: Here also

packages/shared/src/components/notifications/EnableNotification.tsx
```diff
@@ -87,17 +92,14 @@ function EnableNotification({
...
           onClick={onEnable}
         >
           Subscribe
         </Button>
-        <CloseButton
-          className="right-3"
-          position="absolute"
-          onClick={onDismiss}
-        />
+        <CloseButton className="right-3 !absolute" onClick={onDismiss} />
```
rebelchris, line 102: And here

packages/shared/src/components/notifications/EnableNotification.tsx
```diff
@@ -135,17 +137,19 @@ function EnableNotification({
       <div className="flex mt-4 align-center">
         {!hasEnabled && (
           <Button
-            buttonSize={ButtonSize.Small}
-            className="mr-4 min-w-[7rem] btn-primary-cabbage"
+            size={ButtonSize.Small}
+            variant={ButtonVariant.Primary}
+            color={ButtonColor.Cabbage}
+            className="mr-4 min-w-[7rem]"
```
rebelchris, line 143: This might have some use-case but was the `min-w` needed here?

packages/shared/src/components/notifications/EnableNotification.tsx
```diff
@@ -166,10 +170,9 @@ function EnableNotification({
       />
       {!showTextCloseButton && (
         <CloseButton
-          buttonSize={ButtonSize.XSmall}
-          className="top-1 laptop:top-3 right-1 laptop:right-3"
+          size={ButtonSize.XSmall}
+          className="top-1 laptop:top-3 right-1 laptop:right-3 !absolute"
```
rebelchris, line 174: Absolute :)

packages/shared/src/components/notifications/InAppNotification.tsx
```diff
@@ -106,10 +106,9 @@ export function InAppNotificationElement(): ReactElement {
       onMouseLeave={() => startTimer(payload.timer)}
     >
       <CloseButton
-        buttonSize={ButtonSize.XSmall}
-        className="top-3 right-3"
+        size={ButtonSize.XSmall}
+        className="top-3 right-3 !absolute"
```
rebelchris, line 110: Absolute

packages/webapp/components/CookieBanner.tsx
```diff
@@ -22,7 +22,7 @@ export default function CookieBanner({
 
   return (
     <div className="flex fixed laptop:right-6 bottom-0 laptop:bottom-6 left-0 flex-col laptop:items-center laptop:p-6 py-4 pr-14 pl-4 w-full laptop:w-48 laptop:text-center laptop:rounded-2xl laptop:border border-t text-theme-label-secondary bg-theme-bg-tertiary border-theme-divider-secondary typo-footnote z-[200] laptop:left-[unset]">
-      <ModalCloseButton onClick={close} />
+      <ModalClose onClick={close} className="top-2" />
```
rebelchris, line 25: I see we add `top-2` here, but no absolute position?
I see the `ModalCloseButton` actually have some inline styling for it, so maybe double check what needs to be added here.

```
      className={classNames('right-4 z-1', className)}
      style={{ position: 'absolute', ...style }}
```

packages/shared/src/components/modals/OnboardingModal.tsx
```diff
@@ -140,22 +139,15 @@ function OnboardingModal({
...
-            className="h-full"
-            onClose={onClose}
-            formRef={formRef}
-            onSuccessfulLogin={onRegistrationSuccess}
-            onSuccessfulRegistration={onRegistrationSuccess}
-            trigger={AuthTriggers.CreateFeedFilters}
-            onDisplayChange={(display: AuthDisplay) => setScreenValue(display)}
-          />
-        </>
+        <AuthOptions
```
rebelchris, line 142: Why did we remove the CloseButton here?

rebelchris (changes requested):
Amazing work.
Have some minor optimizations for the position we can do.

Thanks for the screenshots helped a lot.
Did notice one that should have a close button:
Screenshot 2023-12-19 at 08 58 51

packages/shared/src/components/checklist/ChecklistCard.tsx
kopancek: No, the `position` prop was removed in ButtonV2 in favor of using the absolute className from tailwind, as done here.

rebelchris: Ah but then we don't need the important right?
We should avoid having both `relative` and `absolute` at the same time.

packages/shared/src/components/modals/OnboardingModal.tsx
kopancek: We already render CloseButton in the `AuthHeader` component. This is rendered for each tab in `AuthOptions`. So we were effectively rendering 2 close buttons on top of each other. I decided to remove this one to only have 1 button there.

rebelchris: Got it, thanks for the info :)

kopancek:
> Amazing work. Have some minor optimizations for the position we can do.
> 
> Thanks for the screenshots helped a lot. Did notice one that should have a close button: Screenshot 2023-12-19 at 08 58 51

I actually noticed this and fixed the code, but forgot to upgrade the screenshot in the PR description. Replacing now...

packages/shared/src/components/notifications/EnableNotification.tsx
kopancek: Looking at the visuals, it seems like the size of the button is more than the specified min width in all of the viewports (large, tablet, phone). So removing the min-w makes sense 👍

packages/webapp/components/CookieBanner.tsx
kopancek: `ModalClose` is the one that won the war of the close buttons. And [it already has `!absolute` as part of the `className` prop](https://github.com/dailydotdev/apps/pull/2435/files#diff-9a6560e0e53738b292f2a629bf9396d7f35f656211f270c65cd9cbc360dbfaacR18).

packages/shared/src/components/checklist/ChecklistCard.tsx
kopancek: The original [`Button.tsx` defaults to `position = 'relative'`](https://github.com/dailydotdev/apps/blob/main/packages/shared/src/components/buttons/Button.tsx#L83)

I kept the same behavior for ButtonV2, because I didn't know if we already leveraged that somewhere.

Now that I think about it, it's probably safe to remove that. Any styles that we have can be changed while we migrate to ButtonV2 to match existing behavior. That way we can move to `absolute` instead of `!absolute`. WDYT?

rebelchris: Not sure what the proposal would be to only rely on classnames?
My fear is we are going to have to add `relative` everywhere then?

Or did you mean to re-use this position prop?

kopancek: From what I can see, we do not apply any of the properties `top, left, right, bottom, inset, z-index` to Buttons by default. Which means, we do not actually need to render the Button in relative positioning. 

So we can treat adding the `relative` prop where needed, based on existing usage. I can do it while I refactor the rest of the usages. It should be fairly uncommon to use `relative` positioning for Button IMO.

So the proposal is:
```
--- a/packages/shared/src/components/buttons/ButtonV2.tsx
+++ b/packages/shared/src/components/buttons/ButtonV2.tsx
@@ -90,7 +90,7 @@ function ButtonComponent<TagName extends AllowedTags>(
       className={classNames(
         `btn inline-flex flex-row items-center justify-center border
         typo-callout font-bold
-        no-underline shadow-none cursor-pointer select-none focus-outline relative
+        no-underline shadow-none cursor-pointer select-none focus-outline

--- a/packages/shared/src/components/checklist/ChecklistCard.tsx
+++ b/packages/shared/src/components/checklist/ChecklistCard.tsx
@@ -33,7 +33,7 @@ const ChecklistCard = ({
             <CloseButton
               size={ButtonSize.Small}
               variant={ButtonVariant.Secondary}
-              className="top-3 right-3 text-white border-white !absolute"
+              className="top-3 right-3 text-white border-white absolute"
               onClick={onRequestClose}
```

rebelchris: In some cases relative might have been added only for z-issues (even just adding relative without z-x could be a benefit.
So I'm open to it, but it would mean we need to double check a lot of things that they didn't get broken.

kopancek: Based on my understanding, unless z-index is specified, `relative` by itself does not create a new stacking order: https://developer.mozilla.org/en-US/docs/Web/CSS/position#relative

rebelchris: Let's proceed with your suggestion of removing it and removing the important note on absolutes

kopancek: Should be done in the latest commit

packages/shared/src/components/modals/VideoModal.tsx
```diff
@@ -26,10 +25,9 @@ export default function VideoModal({
       {...props}
     >
       <CloseButton
```
rebelchris, line 27: Should this not be `ModalCloseButton` then?
(it's not really blocking tbh

rebelchris (approved):
Just wondering why some modals don't use the `ModalCloseButton` though?
Otherwise happy with it.

packages/shared/src/components/modals/VideoModal.tsx
kopancek: That's a good question. Not sure why, maybe again something from the past that we can now change to be consistent. Let me try...

kopancek: OK, I refactored to `ModalClose` where it made sense (in Modals mostly). Please take a look again 🙏

packages/shared/src/components/checklist/ChecklistCard.tsx
```diff
@@ -29,9 +30,10 @@ const ChecklistCard = ({
           <p className="mb-1 font-bold text-white typo-body">{title}</p>
           <p className="text-white typo-callout">{description}</p>
           {typeof onRequestClose === 'function' && (
-            <CloseButton
-              buttonSize={ButtonSize.Small}
-              className="top-3 right-3 text-white border-white !absolute btn-secondary"
+            <ModalClose
+              size={ButtonSize.Small}
+              variant={ButtonVariant.Secondary}
+              className="top-3 right-3"
```
rebelchris, line 36: We should just be careful as `modalClose` already has `right-2` by default.
Meaning we could end up with both classes and depending on tailwind version the order changes it executes them.

packages/shared/src/components/modals/common/ModalClose.tsx
```diff
@@ -19,11 +17,9 @@ function ModalCloseComponent(
   return (
     <CloseButton
       {...props}
-      buttonSize={buttonSize}
       onClick={onClick}
       ref={ref}
-      className={classNames('right-2 z-1', className)}
-      style={{ position: 'absolute', ...style }}
+      className={classNames('right-2 z-1', { absolute }, className)}
```
rebelchris, line 22: Perhaps we can have the position as a default prop?
`right-2` and then override for others?

rebelchris (commented):
Nice just have one small request to avoid duplicate classes

packages/shared/src/components/modals/common/ModalClose.tsx
kopancek: I just did that in [1b2d07a](https://github.com/dailydotdev/apps/pull/2435/commits/1b2d07a83d32f72e5b63a607d860a921955545b3)

packages/shared/src/components/checklist/ChecklistCard.tsx
kopancek: See [1b2d07a](https://github.com/dailydotdev/apps/pull/2435/commits/1b2d07a83d32f72e5b63a607d860a921955545b3)

rebelchris (approved):
That works :)

PR #2436, opened 2023-12-18
Fix: shared postcard width

nensidosari:
## Changes

Potential solution for https://dailydotdev.slack.com/archives/C059GQ5RFP0/p1702912778957629

### DIFF
--- a/packages/shared/src/components/cards/SharedPostCardFooter.tsx
+++ b/packages/shared/src/components/cards/SharedPostCardFooter.tsx
@@ -19,7 +19,7 @@ export const SharedPostCardFooter = ({
   return (
     <div
       className={classNames(
-        'p-3 gap-3 mb-2 rounded-12 border border-theme-divider-tertiary flex w-fit flex-auto h-auto min-h-0 w-full',
+        'p-3 gap-3 mb-2 rounded-12 border border-theme-divider-tertiary flex flex-auto h-auto min-h-0 w-full',
         isShort ? 'flex-row items-center' : 'flex-col',
       )}
     >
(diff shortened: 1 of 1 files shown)


kopancek (approved):
LGTM :shipit:

packages/shared/src/components/cards/SharedPostCardFooter.tsx
```diff
@@ -19,7 +19,7 @@ export const SharedPostCardFooter = ({
   return (
     <div
       className={classNames(
-        'p-3 gap-3 mb-2 rounded-12 border border-theme-divider-tertiary flex w-fit flex-auto h-auto min-h-0',
+        'p-3 gap-3 mb-2 rounded-12 border border-theme-divider-tertiary flex w-fit flex-auto h-auto min-h-0 w-full',
```
rebelchris, line 22: There are contradicting widths now.
In our tailwind version only 1 will win and think it's currently based on alphabetical order, maybe double check which one get's rendered and delete the other?

Refering to `w-fit` and `w-full`

rebelchris (changes requested):
One small request.

omBratteng:
@capJavert you encountered this issue, could you please check if this fixes it?

packages/shared/src/components/cards/SharedPostCardFooter.tsx
omBratteng: That's what's a bit annoying with using `classNames` function, the linter doesn't catch it 😬
That's why I (try to) always paste the classes in a regular `className` to get the linter to do its thing.

capJavert (approved):
I don't really see what the changes are because of rebase but when I throttle locally I can not see the jump anymore 👍

packages/shared/src/components/cards/SharedPostCardFooter.tsx
nensidosari: Removed w-fit, that was my intention, just missed removing it 😓

omBratteng (approved):

rebelchris (approved):

PR #2441, opened 2023-12-20
fix: check if next page exists and disable pagination

capJavert:
- currently the next arrow in related posts widget will always be active until user clicks on it and request checks if there is data
- this is due to using cursor pagination which does not include total
- with this we always prefetch next page so we can preemptively disable the button if there is no next page

image

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done


rebelchris (approved):
Should work :)
Not 100% sure if we need a cancel on the cleanup in useEffect as it's a running query?

capJavert:
@rebelchris we can not really but since it is internal of react-query it will handle it.

PR #2443, opened 2023-12-20
fix: console warning when using wrapperClassName prop on plain image

kopancek:

Get's rid of the following react warning in dev tools console:
```
Warning: React does not recognize the `wrapperClassName` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `wrapperclassname` instead. If you accidentally passed it from a parent component, remove it from the DOM element.
```

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

### DIFF
--- a/packages/shared/src/components/cards/PostCardFooter.tsx
+++ b/packages/shared/src/components/cards/PostCardFooter.tsx
@@ -27,11 +27,6 @@ export const PostCardFooter = ({
 }: PostCardFooterProps): ReactElement => {
   const isVideoType = isVideoPost(post);
   const ImageComponent = isVideoType ? CardVideoImage : CardImage;
-  const videoImageProps: { wrapperClassName?: string } = {};
-  if (isVideoType) {
-    videoImageProps.wrapperClassName = 'my-2';
-  }
-
   return (
     <>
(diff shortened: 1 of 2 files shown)


nensidosari (approved):
LGTM :)

omBratteng (commented):
Could take a look at how I did it in #2444, that way you don't need to set types

packages/shared/src/components/cards/PostCardFooter.tsx
```diff
@@ -50,7 +50,7 @@ export const PostCardFooter = ({
           )}
           loading="lazy"
           data-testid="postImage"
-          wrapperClassName="my-2"
+          {...(isVideoType && { wrapperClassName: 'overflow-hidden' })}
```
omBratteng, line 53: Was this supposed to be changed to `'overflow-hidden'` 😬

kopancek: Damn, bad copy paste, thanks!

PR #2444, opened 2023-12-20
fix: image wrapper width in notifications

omBratteng:

Fixes the width of video image in notifications, and reduces the size of the play icon

<table>
<tr>
 <td>Before
 <td>After
<tr>
 <td>
image
</td>
 <td>image
</table>

### DIFF
--- a/packages/shared/src/components/image/VideoImage.tsx
+++ b/packages/shared/src/components/image/VideoImage.tsx
@@ -7,21 +7,18 @@ import { ImageProps, Image } from './Image';
 export interface VideoImageProps extends ImageProps {
   size?: IconSize;
   wrapperClassName?: string;
-  wrapperWidth?: 'w-full' | 'w-fit';
 }
 
 const VideoImage = ({
   size = IconSize.XLarge,
   wrapperClassName,
-  wrapperWidth = 'w-full',
   ...props
 }: VideoImageProps): ReactElement => {
   return (
     <div
       className={classNames(
         wrapperClassName,
-        wrapperWidth,
-        'flex relative justify-center items-center h-auto rounded-12',
+        'flex relative justify-center items-center h-auto rounded-12 w-full',
       )}
     >
       <span className="absolute w-full h-full rounded-12 bg-overlay-tertiary-black" />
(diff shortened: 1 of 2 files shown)


packages/shared/src/components/image/VideoImage.tsx
```diff
@@ -7,18 +7,21 @@ import { ImageProps, Image } from './Image';
 export interface VideoImageProps extends ImageProps {
   size?: IconSize;
   wrapperClassName?: string;
+  wrapperWidth?: 'w-full' | 'w-fit';
```
rebelchris, line 10: Isn't that what the wrappedClassName should do?

rebelchris (changes requested):
I'm not 100% sure why we needed the change, maybe check with the person who originally added it what it supposed to solve.
Else we go back and forth between issues maybe.

packages/shared/src/components/image/VideoImage.tsx
omBratteng: First thought about adding `!w-fit` to wrapperClassName, but then you get an extra classname.

But I will do some more testing and see if w-full is actually needed in production

rebelchris: Cool yeah please check as I'm not sure for which place it was added as it's a generic component basically.

omBratteng: The change from w-fit to w-full was in a different component, done in #2436.

If I set it to be w-fit always, the normal cards looks wrong
image

omBratteng: Though, I think I actually can just add an empty div around the imagecomponent here to fix it.

packages/shared/src/components/image/VideoImage.tsx
```diff
@@ -10,7 +10,7 @@ export interface VideoImageProps extends ImageProps {
 }
 
 const VideoImage = ({
-  size = IconSize.XXXLarge,
+  size = IconSize.XLarge,
```
nensidosari, line 13: Does this mix up the size for the video images on cards? I think they require the XXXLarge by default?

nensidosari (approved):
LGTM as a solution for the moment, though not a big fan of unnecessary wrapping divs only for layout fixes. 
Maybe some solution can be given modifying the already wrapping div of notification, but we can leave that for future improvement if we want this fix shipped fast.

rebelchris (approved):
Let's give it a go, but monitor impact :)

PR #2447, opened 2023-12-21
fix: cleanup hide modal class on hook unmount

capJavert:
- Original [issue](https://dailydotdev.slack.com/archives/C01L0QXQJNB/p1701346072422259) 
- Basically the issue was that we add `hide-on-modal` class on next root when modal is active, it is controlled in `useHideOnModal` hook
- This class (after opening post modal once) would not be cleaned, which then would hide the feed when any other modal like share is active
- This was due to modal which called the hook `useHideOnModal` being unmounted because it was closed
- Here I add the cleanup to remove the class on unmount which fixes the issue

Before:
image

After:
image

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-2034 #done

### DIFF
--- a/packages/shared/src/hooks/useHideOnModal.ts
+++ b/packages/shared/src/hooks/useHideOnModal.ts
@@ -4,12 +4,16 @@ export function useHideOnModal(predicate: boolean): void {
   useEffect(() => {
     const root = document.querySelector('#__next');
     if (!root) {
-      return;
+      return undefined;
     }
     if (predicate) {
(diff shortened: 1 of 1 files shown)


rebelchris (approved):
Nice find!
What a edge-case 🤯

omBratteng (approved):

PR #2448, opened 2023-12-21
refactor: use ButtonV2 in PromotionalBanner and SocialShareButton

kopancek:
## Changes

Uses the ButtonV2 in more components - see screenshots.
Renamed the `SocialShareIcon` file to `SocialShareButton` and tidied up the typings and color definitions.

## Screenshots

Screenshot 2023-12-19 at 16 59 57
Screenshot 2023-12-19 at 17 06 41
Screenshot 2023-12-20 at 13 39 04
Screenshot 2023-12-21 at 10 02 35
Screenshot 2023-12-21 at 10 06 48
Screenshot 2023-12-21 at 10 08 08
Screenshot 2023-12-21 at 10 08 44
Screenshot 2023-12-21 at 10 09 17
Screenshot 2023-12-21 at 10 09 37
Screenshot 2023-12-21 at 10 10 25
Screenshot 2023-12-21 at 10 10 45
Screenshot 2023-12-21 at 10 11 07
Screenshot 2023-12-21 at 10 11 30
Screenshot 2023-12-21 at 10 11 49
Screenshot 2023-12-21 at 11 45 02
Screenshot 2023-12-21 at 13 37 59
Screenshot 2023-12-21 at 13 45 26

- [X] MobileL (420px)
- [X] Tablet (656px)
- [X] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android


packages/shared/src/components/ProfileMenu.tsx
```diff
@@ -128,13 +128,12 @@ export default function ProfileMenu({
         <span className="font-bold typo-title3">{user.name}</span>
         <span className="mt-1 typo-callout">@{user.username}</span>
       </LabeledImage>
-      <div className="flex flex-col -mt-16">
+      <div className="flex relative flex-col -mt-16">
```
rebelchris, line 131: Not super urgent, but actually these probably shouldn't be `Button` button `button` as they have no overlap styling and just render for the onClicks right?

packages/shared/src/components/post/PostContent.tsx
```diff
@@ -195,7 +195,7 @@ export function PostContent({
         onShare={onShare}
         onReadArticle={onReadArticle}
         post={post}
-        className="pb-8"
+        className="pb-8 post-widgets"
```
rebelchris, line 198: Did you mean to add this one?

packages/shared/src/components/widgets/SocialShareButton.tsx
```diff
@@ -22,42 +18,36 @@ const sizeToText = {
...
-        iconOnly
+        size={size}
         icon={icon}
-        pressed={pressed}
         ref={button}
       />
       <ShareText
         className={classNames(
-          'mt-1.5 max-w-[4rem] overflow-hidden overflow-ellipsis text-center',
+          'mt-1.5 max-w-16 overflow-hidden overflow-ellipsis text-center', // max-w-[4rem]
```
rebelchris, line 50: You have a comment here still :)
Also you verified the max-w-16 works? (not sure why we had the arbitraty one before then.

packages/shared/tailwind/colors.js
```diff
@@ -180,6 +182,43 @@ const colors = {
     70: '#2E4886',
     80: '#2E4886',
     90: '#253B76',
+    darkLabel: '#FFFFFF',
+  },
+  reddit: {
```
rebelchris, line 187: With this I think we need to modify the `tailwind.config` variants of reddit/linkedin etc?
Although maybe that was more preferred as they only have 1 color really?
So you don't really need the color schema right.

rebelchris (changes requested):
Nice work!
Some minor suggestions

packages/shared/src/components/widgets/SocialShareButton.tsx
kopancek: Good catch. I added the 16 as an arbitrary value in [tailwind.config.js](https://github.com/dailydotdev/apps/pull/2448/files#diff-23b6dee9318cba31995fad41ac7e2579aa617078b5963bcb8d94e6d75cda6159R212)

packages/shared/tailwind/colors.js
kopancek: True, these colors are only used for buttons and we don't really need the color schema. We had kind of a dichotomy here, where twitter for example only defined 1 color and whatsapp and facebook more of a color scheme.

I think keeping the values here makes more sense than in the `tailwind.config.js`, so I'm going to remove the themes defined there in favor of the colors defined in this file.

packages/shared/src/components/widgets/SocialShareButton.tsx
kopancek: It seems like `max-w-16` class was not being precomputed for us by tailwind though. But it should. not matter, since the parent component has `w-16` class, so this child will never exceed that anyway. I think it should be safe to remove the `max-w-` altogether, so that's what I'll do here.

packages/shared/src/components/post/PostContent.tsx
kopancek: Nope 🤦

packages/shared/src/components/ProfileMenu.tsx
kopancek: Well, they do have some hover state styling, which actually is the same as tertiary Button variant.

packages/shared/tailwind/colors.js
rebelchris: Ah I think Twitter had multiple when it was blue probably just something missed/choose the easy way when doing the black

rebelchris: I meant more should we not rather go for single entries in `tailwind.config` vs having the same value here everywhere.
Unless we use the specific implementation anywhere?

rebelchris (changes requested):
Just the color use I think we can optimize a bit.

rebelchris (approved):
Good to go

PR #2449, opened 2023-12-21
fix: replace useRef with state

omBratteng:
- Replaces useRef with useState and adds it to `ExtensionContextProvider` to prevent prop drilling
- Render nothing while checking for host permissions

WT-2021 #done

### DIFF
--- a/packages/extension/src/contexts/ExtensionContext.tsx
+++ b/packages/extension/src/contexts/ExtensionContext.tsx
@@ -1,10 +1,4 @@
-import React, {
-  ReactElement,
-  ReactNode,
-  useContext,
-  useMemo,
-  useState,
-} from 'react';
+import React, { ReactElement, ReactNode, useContext, useMemo } from 'react';
 
 import browser from 'webextension-polyfill';
 import { useQueryClient } from '@tanstack/react-query';
--- a/packages/extension/src/contexts/ExtensionContext.tsx
+++ b/packages/extension/src/contexts/ExtensionContext.tsx
@@ -12,11 +12,13 @@ import {
 } from '../lib/extensionScripts';
 
 export type ExtensionContextProviderProps = {
+  currentPage: string;
   setCurrentPage: React.Dispatch<React.SetStateAction<string>>;
   children?: ReactNode;
 };
 
 export const ExtensionContextProvider = ({
+  currentPage,
   setCurrentPage,
   children,
 }: ExtensionContextProviderProps): ReactElement => {
@@ -30,9 +32,10 @@ export const ExtensionContextProvider = ({
       getHostPermission,
       requestHostPermissions: browser.permissions.request,
       origins: HOST_PERMISSIONS,
+      currentPage,
       setCurrentPage,
     }),
-    [client, setCurrentPage, trackEvent],
+    [client, currentPage, setCurrentPage, trackEvent],
   );
 
   return (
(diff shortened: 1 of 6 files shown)


packages/shared/src/components/ExtensionOnboarding.tsx
```diff
@@ -1,12 +1,22 @@
...
 
 const ExtensionOnboarding = (): ReactElement => {
+  const { setCurrentPage } = useExtensionContext();
+
+  useEffect(() => {
+    setCurrentPage('/hijacking');
+    return () => {
+      setCurrentPage('/');
+    };
+  });
```
capJavert, line 18: I feel like `ExtensionOnboarding` should be rendered based on current page, not other way around where mounting of component triggers page change.

packages/shared/src/components/ExtensionPermissionsPrompt.tsx
```diff
@@ -1,21 +1,24 @@
...
+    useExtensionContext();
   const { refetch: refetchHostPermissions } = useHostStatus();
 
+  useEffect(() => {
+    setCurrentPage('/permissions');
+
+    return () => {
+      setCurrentPage('/');
+    };
+  });
```
capJavert, line 20: Same as my other comment.

packages/extension/src/contexts/ExtensionContext.tsx
```diff
@@ -28,8 +30,9 @@ export const ExtensionContextProvider = ({
       getHostPermission,
       requestHostPermissions: browser.permissions.request,
       origins: HOST_PERMISSIONS,
+      setCurrentPage,
```
capJavert, line 33: Maybe also expose current page through context?

packages/shared/src/components/ExtensionOnboarding.tsx
capJavert: Also using `useEffect` without deps array here would trigger it on each render

omBratteng: I always mix them up, my idea was only trigger once, but that is empty deps array, not no-deps array.

omBratteng: > I feel like `ExtensionOnboarding` should be rendered based on current page, not other way around where mounting of component triggers page change.

Not sure I get it. Set the current page, then render?

rebelchris (approved):
I guess that works as well :)

packages/shared/src/components/ExtensionOnboarding.tsx
capJavert: Basically with this kind of approach when this component renders and sets the current page. Now if another component eg. `ExtensionPermissionsPrompt` renders after it, it will overwrite the current page as well. 

Now you can argue that they are never rendered at the same time but this kind of things get forgotten easily 😆 

So if we are going in this kind of direction where we are basically creating a little router for the extension I would rather have some other condition like `shouldRedirectOnboarding` actually update the `currentPage` and then in `App` we only render ExtensionOnboarding when `currentPage === '/onboarding`.

omBratteng: Small issue with that, because of with FeaturesBoundary happening outside InternalApp, we would then not set currentPage to the right page for permissions.

omBratteng: I think it works fine like this for now. If we end up adding more pages, we probably need to reconsider the router

PR #2452, opened 2023-12-22
refactor: use ButtonV2 in auth forms

kopancek:
## Changes

Move to ButtonV2 in all of the auth forms.

### Screenshots

Screenshot 2023-12-22 at 07 55 32
Screenshot 2023-12-22 at 07 57 07
Screenshot 2023-12-22 at 07 57 26
Screenshot 2023-12-22 at 07 57 49
Screenshot 2023-12-22 at 09 34 42
Screenshot 2023-12-22 at 09 37 46
Screenshot 2023-12-22 at 09 50 54

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [X] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android


packages/shared/src/components/auth/LoginForm.tsx
```diff
@@ -119,13 +120,9 @@ function LoginForm({
           <AlertParagraph className="flex-1 !mt-0">
             The email or password you entered doesn&apos;t match our records.
             Please try again or{' '}
-            <button
-              type="button"
-              onClick={onSignup}
-              className="font-bold underline"
-            >
+            <ClickableText onClick={onSignup} className="font-bold underline">
```
rebelchris, line 123: I think this should use `inverseUnderline` prop of ClickableText
Although not 100% sure if it adds more unwanted styles maybe?

nensidosari (approved):
Looks great to me! 🥳

rebelchris (approved):
Nice work :D
One small non blocking comment

packages/shared/src/components/auth/LoginForm.tsx
kopancek: If I use `inverseUnderline`, it is not underlined by default and only on hover. I think we want the opposite based on the current implementation. I changed it so that we underline, but remove underline on hover. This is aligned with the behavior of other similar buttons on the login forms.

This is how it looks now:
Screenshot 2023-12-22 at 10 46 24

PR #2460, opened 2023-12-26
refactor: post page redesign/redirect

sshanzel:
## Changes
- Missing left border on widgets for public squad posts.
- On post click open the modal only when the feed layout is control.
- Instead of using router push when clicking a post on feed layout v1, we instead change the overlay used to be a link. This way we can leverage the `prefetch={true}` set by default by nextjs to make redirection to the new page, faster. This is currently applied on ArticlePostCard only, once agreed, I will apply it to other Card types too.
- New Pill-like label on top of the post content below the navigation.
- Refactor the navigation button on post pages to utilize the new Back button.

TODO:
- [x] Check the color applied on the pill-like element as it is not part of the design system.

Preview:
Screenshot 2023-12-26 at 11 20 37 PM

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-2003 #done


packages/shared/src/components/post/BasePostContent.tsx
```diff
@@ -51,6 +59,11 @@ export function BasePostContent({
       {customNavigation ?? (
         <PostNavigation {...navigationProps} className={className.navigation} />
       )}
+      {withPillTypes.includes(post?.type) && (
+        <span className="py-2 px-3 mt-6 font-bold capitalize rounded-8 bg-[#BF29F014] text-theme-color-cabbage typo-footnote w-fit">
```
sshanzel, line 63: The usage of `bg-[#BF29F014]` is temporary, I have already raised a discussion with the design team.

rebelchris: Bg-cabbage-float i think will do,
There should be some examples for the exact classname.

Also should this always show or only v1 feed?

packages/shared/src/components/post/SquadPostPageNavigation.tsx
```diff
@@ -28,7 +34,24 @@ function SquadPostPageNavigation({
   squadLink,
 }: SquadPostPageNavigationProps): ReactElement {
   const isFromNavigation = checkIsFromNavigation();
+  const router = useRouter();
   const link = isFromNavigation ? NOTIFICATIONS_LINK : squadLink;
+  const layout = useFeature(feature.feedLayout);
+
+  if (layout === FeedLayout.V1) {
```
rebelchris, line 41: For squad it should still always show though right? No matter v1 or not

rebelchris: Oh see it below, should we not uniform those?

rebelchris (changes requested):
Nice work, some minor checks on v1 vs not

packages/shared/src/components/post/BasePostContent.tsx
sshanzel: Was not available unfortunately. The available one was `overlay-float` which doesn't produce any color.

packages/shared/src/components/post/SquadPostPageNavigation.tsx
sshanzel: Yeah, was thinking of it too, though the only overlap they had was the history check and I didn't want to include the logic of `routedFromSquad` to be part of the other implementations. Wdyt?

packages/shared/src/components/post/BasePostContent.tsx
rebelchris: This one is the one:
`bg-theme-overlay-float-cabbage`

Overlay means the .50 version and float means 8%

packages/shared/src/components/post/SquadPostPageNavigation.tsx
rebelchris: Yeah I was indeed thinking about a generic `PostBackButton` which could hold both logic parts?
Let me know if it's a quick win or not.
Seems easier to maintain

sshanzel: Makes sense. Let's do it. Would also lessen some logic from the original files.

packages/shared/src/components/post/BasePostContent.tsx
```diff
@@ -51,6 +59,11 @@ export function BasePostContent({
       {customNavigation ?? (
         <PostNavigation {...navigationProps} className={className.navigation} />
       )}
+      {withPillTypes.includes(post?.type) && (
```
nensidosari, line 62: Should this change be dependant on the experiment?

packages/shared/src/components/cards/ArticlePostCard.tsx
```diff
@@ -60,6 +67,26 @@ export const ArticlePostCard = forwardRef(function PostCard(
     );
   }
 
+  const renderOverlay = () => {
+    if (isFeedPreview) {
+      return null;
+    }
+
+    if (!shouldUseFeedLayoutV1) {
```
nensidosari, line 75: I personally prefer using positive conditions as it seems more readable, in this case I would go better with 
`if (shouldUseFeedLayoutV1) {
return (
      <Link href={post.commentsPermalink}>
        <CardLink
          title={post.title}
          onClick={onPostCardClick}
          href={post.commentsPermalink}
        />
      </Link>
    );
return <CardButton title={post.title} onClick={onPostCardClick} />;
}`

Nothing blocking, just a matter of preference.

packages/shared/src/components/post/SquadPostPageNavigation.tsx
```diff
@@ -29,6 +33,11 @@ function SquadPostPageNavigation({
 }: SquadPostPageNavigationProps): ReactElement {
   const isFromNavigation = checkIsFromNavigation();
   const link = isFromNavigation ? NOTIFICATIONS_LINK : squadLink;
+  const layout = useFeature(feature.feedLayout);
```
nensidosari, line 36: Any reason here for not using the useFeedLayout?

packages/shared/src/components/post/common/PostBackButton.tsx
```diff
@@ -0,0 +1,34 @@
...
+import { useFeature } from '../../GrowthBookProvider';
+import { feature } from '../../../lib/featureManagement';
+
+interface PostBackButtonProps {
+  link?: string;
+}
+
+export function PostBackButton({ link }: PostBackButtonProps): ReactElement {
+  const router = useRouter();
+  const layout = useFeature(feature.feedLayout);
```
nensidosari, line 16: Same as above here, or does it not have the feed context?

packages/shared/src/components/post/common/PostBackButton.tsx
```diff
@@ -0,0 +1,34 @@
...
+  const canGoBack = !!globalThis?.window?.history?.length;
+  const onClick = () => router.push(link);
+
+  if (layout === FeedLayout.Control || (!canGoBack && !link)) {
```
nensidosari, line 20: Why do we check for control rather then not being on V1?

packages/webapp/pages/posts/[id]/index.tsx
```diff
@@ -88,6 +92,7 @@ const PostPage = ({ id, initialData }: Props): ReactElement => {
   const [position, setPosition] =
     useState<CSSProperties['position']>('relative');
   const router = useRouter();
+  const layout = useFeature(feature.feedLayout);
```
Comment (nensidosari) [inline packages/webapp/pages/posts/[id]/index.tsx:95]: Same here? Sorry for these, just trying to evaluate if that hook is necessary or not, if we're not using it.
Or is it because here we don't care on what feed we are in?

packages/webapp/pages/posts/[id]/index.tsx
```diff
@@ -154,16 +159,28 @@ const PostPage = ({ id, initialData }: Props): ReactElement => {
...
+
+  const articleNavigation = (() => {
+    const routedFromSquad = router?.query?.squad;
+    const squadLink = `/squads/${router.query.squad}`;
+
+    if (layout === FeedLayout.V1) {
+      return <PostBackButton link={routedFromSquad ? squadLink : undefined} />;
+    }
+
+    if (!routedFromSquad) {
```
Comment (nensidosari) [inline packages/webapp/pages/posts/[id]/index.tsx:171]: Same here on the positive condition note, nothing to worry & non need to change, just want to highlight in case you think is better as well :)

nensidosari (commented):
Overall looks great, left some minor comments, mostly for my understanding

packages/webapp/pages/posts/[id]/index.tsx
Comment (sshanzel) [reply packages/webapp/pages/posts/[id]/index.tsx:95]: Yep, this part is regardless of which feed you are in since this is the post page which is never in a feed.

packages/shared/src/components/post/common/PostBackButton.tsx
sshanzel: These parts are regardless of which feed you are in since this is the post page which is never in a feed.

packages/webapp/pages/posts/[id]/index.tsx
Comment (sshanzel) [reply packages/webapp/pages/posts/[id]/index.tsx:171]: It is more of a defensive programming where I return early then I leave the most complicated result at the bottom of the function to make the early returns cleaner and easier to read.

packages/shared/src/components/post/SquadPostPageNavigation.tsx
sshanzel: Using the useFeedLayout would always result in `false` since there is no context of feed on the post page.

packages/shared/src/components/post/BasePostContent.tsx
sshanzel: This actually exists already on the pages where it is needed. Will remove this. Thank you!

packages/shared/src/components/cards/ArticlePostCard.tsx
sshanzel: I do defensive programming which returns early, one of the main points is to reduce indents to make it easier to read and follow. The more complicated (multi-line) the return is, the more indent will be applied. In this case, I leave the most complicated return on the bottom-most part to not have the additional indent of the if statement.

Kind of a personal preference, I guess. But if there is a strong point for better devex, I'd be happy to apply changes.

packages/shared/src/components/post/common/PostBackButton.tsx
sshanzel: Yes, there is no context of feed here.

packages/shared/src/components/post/common/PostBackButton.tsx
sshanzel: To return early and render null.

packages/shared/src/components/cards/ArticlePostCard.tsx
kopancek: I'm also a fan of defensive programming and returning early, I vote for keeping it as is...

packages/webapp/pages/posts/[id]/index.tsx
```diff
@@ -154,16 +159,28 @@ const PostPage = ({ id, initialData }: Props): ReactElement => {
...
+      return <></>;
```
Comment (kopancek) [inline packages/webapp/pages/posts/[id]/index.tsx:172]: Why do we return an empty fragment here instead of `null`? Is it because of type checks?

kopancek (approved):
LGTM, added one non-blocking question.

rebelchris (approved):
agree with the early return favorism.

packages/webapp/pages/posts/[id]/index.tsx
Comment (sshanzel) [reply packages/webapp/pages/posts/[id]/index.tsx:172]: We can do that, it was because of the way we handle the default value by using `??` which then detects `null` and falls back to the default value but the intention was to actually have the `null` value instead of the fallback.

Just going to have a bit of refactoring, but I am up for it to make it easier to understand overall.

nensidosari (approved):
Looks great to me as well. Great work 🚀

PR #2462, opened 2023-12-27
fix: card on mobile layout

sshanzel:
## Changes
- Fixed a merge issue.
- Fixed the card layout for mobile - there was a `mb-8` which is not really necessary due to the fact that we have an empty div with `flex-1` class.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done

### DIFF
--- a/packages/shared/src/components/cards/ArticlePostCard.tsx
+++ b/packages/shared/src/components/cards/ArticlePostCard.tsx
@@ -110,7 +110,7 @@ export const ArticlePostCard = forwardRef(function PostCard(
           </CardTitle>
         </CardTextContainer>
         {!showFeedback && (
-          <Container className="mb-8 tablet:mb-0">
+          <Container>
             <CardSpace />
             <PostMetadata
               createdAt={post.createdAt}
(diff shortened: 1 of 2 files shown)


packages/shared/src/components/cards/common/SquadPostCardHeader.tsx
```diff
@@ -32,7 +32,7 @@ export const SquadPostCardHeader = ({
   };
 
   return (
-    <div className="flex relative flex-row gap-2 m-2 mb-3">
+    <div className="flex relative flex-row gap-2 m-2">
```
sshanzel, line 35: The merge conflict mentioned.

idoshamun (approved):
Thx for the quick fix

packages/shared/src/components/cards/ArticlePostCard.tsx
```diff
@@ -110,7 +110,7 @@ export const ArticlePostCard = forwardRef(function PostCard(
           </CardTitle>
         </CardTextContainer>
         {!showFeedback && (
-          <Container className="mb-8 tablet:mb-0">
```
sshanzel, line 113: Since we have now introduced the `max-h-card` on all cards to be strict with consistency, this margin causes the unexpected expansion of the content which is not needed due to the element exactly below this (`CardSpace`).

PR #2464, opened 2023-12-28
refactor: feed layout heading to work with new experiment

sshanzel:
## Changes
- Refactor the prop named `besideSearch` to `shortcuts` as the former is a bit vague. The main purpose of that property, as seen, is to hold the shortcuts element.
- New container With border to hold the settings related to the current feed and my feed.
- As per the task, this is just purely about the feed cards wrapper with the proper border and radius along with the action buttons beside the title. Including the shortcut links being moved below the suggestions list.

Previews of all the variations are as below (to display things are still working as intended).

Extension:
Screenshot 2024-01-03 at 9 30 31 PM
Screenshot 2024-01-03 at 9 30 08 PM
Screenshot 2024-01-03 at 9 29 50 PM
Screenshot 2024-01-03 at 9 29 31 PM

Webapp:
Screenshot 2024-01-03 at 9 34 17 PM
Screenshot 2024-01-03 at 9 31 41 PM
Screenshot 2024-01-03 at 9 31 10 PM
Screenshot 2024-01-03 at 9 30 52 PM

Mobile and tablet:

Screenshot 2024-01-03 at 10 54 00 PM
Screenshot 2024-01-03 at 10 53 21 PM
Screenshot 2024-01-03 at 10 47 54 PM
Screenshot 2024-01-03 at 10 47 47 PM
Screenshot 2024-01-03 at 10 47 02 PM
Screenshot 2024-01-03 at 10 46 52 PM
Screenshot 2024-01-03 at 10 46 02 PM
Screenshot 2024-01-03 at 10 45 49 PM

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1998 #done


packages/shared/src/components/MainFeedLayout.tsx
```diff
@@ -238,16 +243,13 @@ export default function MainFeedLayout({
...
-        (isUpvoted || isSortableFeed) ? (
-          <SearchControlHeader {...searchProps} />
-        ) : null,
-      besideSearch,
+      header: searchVersion === SearchExperiment.Control && !isSearchOn && (
+        <SearchControlHeader {...searchProps} navChildren={navChildren} />
+      ),
+      actionButtons: isV1Search && feedWithActions && (
+        <SearchControlHeader {...searchProps} />
+      ),
```
sshanzel, line 251: Looks much easier to read without the ternaries.

packages/shared/src/components/filters/MyFeedHeading.tsx
```diff
@@ -66,29 +73,20 @@ function MyFeedHeading({
         ? AlertPlacement.Right
         : AlertPlacement.Bottom,
   };
-
-  if (searchVersion === SearchExperiment.V1) {
-    return (
-      <AlertPointer {...alertProps} offset={[0, 0]}>
-        <Button
-          variant={ButtonVariant.Float}
```
sshanzel, line 74: The only difference is the variant and offset, no need for us to have the whole block of code.

sshanzel:
@dailydotdev/web-team should now be ready for review 🥳

nensidosari:
Something I see missing from the images attached in the desc @sshanzel:

- in the search background, that is supposed to be transparent based on the new design. 
- the glow seems to be starting before the feed, on the main component, it now should be within the feed container if you check with the designs
- There is one feed sorting button that needs to be adjusted with this new [design](https://www.figma.com/file/nmfWPS7x3kzLUvYMkBx2kW/daily.dev---Dev-Mode?node-id=275%3A10533&mode=dev), if you see in the design, on the right of the feed settings, if the feed sorting is enabled a new button should show on the header to sort the feed.

packages/extension/src/newtab/CustomLinks.tsx
```diff
@@ -27,7 +27,8 @@ export function CustomLinks({
   return (
     <div
       className={classNames(
-        'ml-auto hidden h-fit flex-row gap-2 rounded-14 border border-theme-divider-secondary p-2 laptop:flex',
+        'hidden laptop:flex flex-row gap-2 p-2 rounded-14 border border-theme-divider-secondary h-fit',
+        'hidden h-fit flex-row gap-2 rounded-14 border border-theme-divider-secondary p-2 laptop:flex',
```
nensidosari, line 31: Why is this doubled? 👀

packages/extension/src/newtab/MainFeedPage.tsx
```diff
@@ -119,7 +123,13 @@ export default function MainFeedPage({
             />
           }
           navChildren={!isSearchOn && <ShortcutLinks />}
-          besideSearch={<ShortcutLinks className="ml-auto" />}
+          shortcuts={
+            <ShortcutLinks
+              className={
+                layout === FeedLayoutEnum.Control ? 'ml-auto' : 'mt-4 w-fit'
```
nensidosari, line 129: wouldn't this depend on the feedName as well?

packages/shared/src/components/MainFeedLayout.tsx
```diff
@@ -254,20 +256,38 @@ export default function MainFeedLayout({
     return isLaptop ? cloudinary.feed.bg.laptop : cloudinary.feed.bg.tablet;
   };
 
-  const FeedPageComponent = shouldUseFeedLayoutV1 ? FeedPageLayoutV1 : FeedPage;
+  const getPadding = () => {
+    if (isV1Search) {
+      return '!pt-2';
+    }
+
+    return shouldUseFeedLayoutV1 && '!pt-0';
```
nensidosari, line 264: I don't think this was the only change between the FeedPageLayoutV1 and FeedPage styled components. Any reason to remove it?

packages/shared/src/components/MainFeedLayout.tsx
```diff
@@ -254,20 +256,38 @@ export default function MainFeedLayout({
...
       )}
       {isSearchOn && search}
-      {feedProps && <Feed {...feedProps} />}
+      {feedProps && (
+        <Feed
+          {...feedProps}
+          className={
+            (shouldUseFeedLayoutV1 || isV1Search) &&
+            !isFinder &&
+            'laptop:!max-w-[36.5rem]'
```
nensidosari, line 286: I'm not sure I understand why this is needed

packages/webapp/components/layouts/MainFeedPage.tsx
```diff
@@ -51,9 +51,7 @@ export default function MainFeedPage({
       router.replace('/');
     } else if (isFinderPage) {
       setIsSearchOn(true);
-      if (!feedName) {
-        setFeedName('popular');
-      }
+      setFeedName('search');
```
nensidosari, line 54: Just checking, I've seen this too and wasn't sure why it was like this, so just making sure we've checked this properly

packages/shared/src/components/feeds/FeedContainer.tsx
```diff
@@ -164,7 +176,7 @@ export const FeedContainer = ({
       )}
     >
       <ScrollToTopButton />
-      <div className="flex w-full flex-col pt-2 laptopL:mx-auto" style={style}>
+      <div className="flex w-full flex-col laptopL:mx-auto" style={style}>
```
nensidosari, line 179: Is the `pt-2` removal intentional?

nensidosari (changes requested):
Few comments on some changes I'm not 100% sure of, and also something I see missing from the images attached in the desc @sshanzel:

- in the search background, that is supposed to be transparent based on the new design.

- the glow seems to be starting before the feed, on the main component, it now should be within the feed container if you check with the designs

- There is one feed sorting button that needs to be adjusted with this new [design](https://www.figma.com/file/nmfWPS7x3kzLUvYMkBx2kW/daily.dev---Dev-Mode?node-id=275%3A10533&mode=dev), if you see in the design, on the right of the feed settings, if the feed sorting is enabled a new button should show on the header to sort the feed.

Other then that, looks good so far :D

packages/extension/src/newtab/CustomLinks.tsx
sshanzel: Probably some merge conflict 😅

packages/shared/src/components/MainFeedLayout.tsx
sshanzel: Yeah, this is not the only difference, but after checking, the new one is not accurate. After merging from main, this branch became broken as well.

packages/shared/src/components/MainFeedLayout.tsx
sshanzel: If you keep this, the content will be maxed out, which should not be the case.

packages/webapp/components/layouts/MainFeedPage.tsx
sshanzel: Yep. This is a bug fix.

packages/shared/src/components/feeds/FeedContainer.tsx
sshanzel: Yes. We don't need that top padding here.

packages/extension/src/newtab/MainFeedPage.tsx
sshanzel: These shortcuts are naturally not available on search feed and are only seen on the extension which is only on all the valid feeds.

sshanzel:
> I see screenshots from testing in devtools as a mobile device, but have you tested on an actual phone? Sometimes there might be a difference.

Not yet, we currently don't have a way to test experiments on preview. Will have to check how to connect my phone to local env.

packages/shared/src/components/MainFeedLayout.tsx
sshanzel: Seems to be broken on mobile.

https://github.com/dailydotdev/apps/assets/13744167/af00c3a6-e98d-416e-b01b-552b490ccbc4

sshanzel: Found a fix to keep the `FeedPageLayoutV1`.

sshanzel:
> in the search background, that is supposed to be transparent based on the new design.

Right. We might have to ask the design team to re-upload the latest in the cloudinary.

sshanzel:
> in the search background, that is supposed to be transparent based on the new design.

Right. We might have to ask the design team to re-upload the latest in the cloudinary.

> the glow seems to be starting before the feed, on the main component, it now should be within the feed container if you check with the designs

Right. We can do that, just wanted to avoid that we will have to conditionally render two image backgrounds instead of one. Should be in this commit: https://github.com/dailydotdev/apps/pull/2464/commits/d7fd3c13f42f9e1178a499a4da11718605631dfd

> There is one feed sorting button that needs to be adjusted with this new [design](https://www.figma.com/file/nmfWPS7x3kzLUvYMkBx2kW/daily.dev---Dev-Mode?node-id=275%3A10533&mode=dev), if you see in the design, on the right of the feed settings, if the feed sorting is enabled a new button should show on the header to sort the feed.

Yes, this should already be part of the new header itself.

packages/shared/src/components/MainFeedLayout.tsx
```diff
@@ -122,6 +124,7 @@ export default function MainFeedLayout({
   });
   const feedVersion = useFeature(feature.feedVersion);
   const searchVersion = useFeature(feature.search);
+  const isV1Search = searchVersion === SearchExperiment.V1;
```
rebelchris, line 127: Here you also don't have access to the hook?

packages/shared/src/components/MainFeedLayout.tsx
```diff
@@ -199,6 +202,7 @@ export default function MainFeedLayout({
     };
 
     const variables = getVariables();
+    const feedWithActions = isUpvoted || isSortableFeed;
```
rebelchris, line 205: Not sure I get the naming, what is the intended difference here?

packages/shared/src/components/MainFeedLayout.tsx
```diff
@@ -243,30 +245,28 @@ export default function MainFeedLayout({
...
 
   return (
-    <FeedPageComponent className="relative">
-      {searchVersion === SearchExperiment.V1 && !isFinder && (
-        <img
-          className="absolute left-0 top-0 w-full max-w-[58.75rem]"
-          src={getImage()}
-          alt="Gradient background"
+    <FeedPageComponent className={classNames('relative', getPadding())}>
+      {isV1Search && !shouldUseFeedLayoutV1 && !isFinder && <FeedGradientBg />}
```
rebelchris, line 260: The evaluations become a bit tricky to maintain I feel like, let's see if we conclude experiments if it becomes easier. (no action needed)

packages/shared/src/components/feeds/FeedContainer.tsx
```diff
@@ -181,14 +207,26 @@ export const FeedContainer = ({
           )}
           {inlineHeader && header}
           {isV1Search && (
-            <span className="flex flex-row gap-3">
+            <ConditionalWrapper
+              condition={!shouldUseFeedLayoutV1}
```
rebelchris, line 211: Since we have this edge-case with conditional wrapper we need to test what happens if someone is not on v1 then suddenly get's added.
What I remember the conditional wrapper might omit that evaluation.

packages/shared/src/components/feeds/FeedContainer.tsx
```diff
@@ -197,32 +235,55 @@ export const FeedContainer = ({
...
+            wrapper={(child) => (
+              <div
+                className={classNames(
+                  'flex flex-col rounded-16 border border-theme-divider-tertiary laptop:mx-0 laptop:mt-6',
+                  isV1Search && 'mt-6',
+                )}
+              >
+                <span className="flex w-full flex-row items-center justify-between px-6 py-4">
+                  <strong className="typo-title3">
+                    {feedNameToHeading[feedName] ?? ''}
```
rebelchris, line 268: This is only shown on v1 layout?

rebelchris (commented):
Some minor comments, nothing blocking from my side.
Seeing as we definitely need to try this out on some preview domain (control and v1) I don't think we blocked by many things.

We'll see during internal testing if there are discrepancies.

packages/shared/src/components/MainFeedLayout.tsx
nensidosari: I'd rather keep FeedPageLayoutV1, and fix what is broken from whatever has been merged after. As if we want to remove it, then we would need to go back to the ticket this was created, to make sure we're considering every detail.

nensidosari: Curiosity, did u realised what was breaking it?

packages/shared/src/components/MainFeedLayout.tsx
nensidosari: This is the search experiment, no hook for it

packages/shared/src/components/MainFeedLayout.tsx
sshanzel: The feeds where you have dropdown options for sorting.

packages/shared/src/components/MainFeedLayout.tsx
sshanzel: Yeah, hopefully, we get past at least one of the experiments to make it all easier.

packages/shared/src/components/feeds/FeedContainer.tsx
```diff
@@ -158,13 +174,23 @@ export const FeedContainer = ({
...
         styles.container,
         className,
       )}
     >
+      {isV1Search && shouldUseFeedLayoutV1 && (
+        <span
+          className="absolute left-1/2 top-0 -mt-4 h-24 w-3/5 -translate-x-1/2 -translate-y-1/2"
+          style={{
+            background:
+              'radial-gradient(ellipse farthest-side, #4E2BD7, #AC1CE4)',
```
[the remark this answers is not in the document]
sshanzel: Was a prototype, I have this removed now and went back to the original. It was mentioned that we can stick with what we have for now.

packages/shared/src/components/feeds/FeedContainer.tsx
sshanzel: They need to get refreshed right? Since experiment values are part of the boot, they need to be re-triggered.

packages/shared/src/components/feeds/FeedContainer.tsx
sshanzel: Yes, the layout with the new header.

packages/shared/src/components/feeds/FeedGradientBg.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+    return isLaptop ? cloudinary.feed.bg.laptop : cloudinary.feed.bg.tablet;
+  };
+
+  return (
+    <img
+      className={classNames(
+        'absolute left-0 top-0 w-full laptop:max-w-[58.75rem]',
+      )}
+      src={getImage()}
```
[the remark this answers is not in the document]
sshanzel: Let me check what's with that `srcSet`. Have not heard of it yet. Will get back to this.

packages/shared/src/components/feeds/FeedContainer.tsx
rebelchris: Yeah think this case will be alright, just not sure on edge-cases.
But maybe we'll see on testing mode.

packages/shared/src/components/feeds/FeedContainer.tsx
rebelchris: Ah ok, just making sure it doesn't show on control :)

packages/shared/src/components/feeds/FeedContainer.tsx
sshanzel: Awesome. Thank you for the insight with regards to the known issue.

packages/shared/src/components/MainFeedLayout.tsx
sshanzel: Some margins somewhere, can't exactly remember on which place.

packages/shared/src/components/MainFeedLayout.tsx
```diff
@@ -243,30 +245,28 @@ export default function MainFeedLayout({
...
+      {isSearchOn && search}
+      {feedProps && (
+        <Feed
+          {...feedProps}
+          className={
+            shouldUseFeedLayoutV1 && !isFinder && 'laptop:!max-w-[36.5rem]'
```
nensidosari, line 266: I believe the content should be maxed out, we're just missing paddings on desktop screens. I think I'd rather introduce paddings then another set size width.

packages/shared/src/components/utilities/common.tsx
```diff
@@ -111,7 +111,7 @@ export const FeedPage = classed(
 export const FeedPageLayoutV1 = classed(
   BasePageContainer,
   pageContainerClassNames,
-  'pt-10 tablet:px-6 !ml-auto !px-0',
+  'pt-10 tablet:px-6 !ml-auto !px-0 !max-w-full',
```
nensidosari, line 114: Why is this `!max-w-full` needed?

nensidosari (commented):
Few more concerns on some styles 🙏

packages/shared/src/components/MainFeedLayout.tsx
sshanzel: This is from Figma's artboard.

Screenshot 2024-01-04 at 6 58 44 PM

With regards to the padding, we have the right spacing as seen below. It is just that the margin was used over the padding but should not make a difference when either is used.

Screenshot 2024-01-04 at 7 00 29 PM

sshanzel: We even have a bigger padding than what is in Figma.

Screenshot 2024-01-04 at 7 02 20 PM

packages/shared/src/components/utilities/common.tsx
sshanzel: We need it, or else the tablet mode would result differently. From Figma, the tablet mode should consume the whole screen width.

Without the max width:
Screenshot 2024-01-04 at 7 03 50 PM

Figma:
Screenshot 2024-01-04 at 7 04 25 PM

packages/shared/src/components/MainFeedLayout.tsx
sshanzel: This is an experiment to introduce a new layout that might be better overall, part of introducing the improvements can be the size of the width itself. I think we are accurate in specifying the exact values rather than always adjusting how much should be the padding.

nensidosari: But in case we stay with this, we have already introduced a max width for the higher up container - 680px and aligned that with the width on profile pages as well, in case we keep this and want to change this max size of the higher up container later on to let's say 700px - then we would need to change this size as well here to something bigger, rather then having a 24px padding left and right, in which case we wouldn't have to do anything.

sshanzel: That actually makes sense. Let's go with that.

packages/shared/src/components/utilities/common.tsx
nensidosari: Are we sure we want that from 656px to 1020px? 
Screenshot 2024-01-04 at 12 23 05 PM
Looks weird to me in larger screens.

I think when setting the breakpoints I had taken this as, it should look like this from 656px and downwards, not upwards. But might be me mistaken. 
Shall we check with design to make sure?

sshanzel: Yeah, this case will have to be checked with the design team.

nensidosari: Cool, added here https://dailydotdev.slack.com/archives/C061T644XV3/p1704368093654569, lets see what they say and move on accordingly

nensidosari: Nice, we can leave this then, it was my misunderstanding 🙌

nensidosari: After checking with design, we want the feed to have a max width of 680px, which it already has, the fullwidth will come naturally in screens smaller then 680px.

packages/shared/src/components/MainFeedLayout.tsx
nensidosari: Yup, that bigger margin is because of this `laptop:!max-w-[36.5rem]`. That's why I'm suggesting to remove it, and just add paddings/margins whatever works best.

nensidosari: Conversation continued here https://github.com/dailydotdev/apps/pull/2464#discussion_r1441600408

packages/shared/src/components/MainFeedLayout.tsx
rebelchris: Hmm ok open to leave it alternative `feedWithActions` but also not much clearer

sshanzel: Agree, it is a bit vague. Open for any suggestions 🚀

packages/shared/src/components/feeds/FeedGradientBg.tsx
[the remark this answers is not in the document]
sshanzel: Worked great! I learned something new today 🚀

packages/shared/src/components/MainFeedLayout.tsx
sshanzel: It is basically the same thing, by applying paddings which will equate to the total margin displayed here.

packages/shared/src/components/MainFeedLayout.tsx
rebelchris: No better suggestion, let's proceed with this.

rebelchris (approved):
Nice, amazing work on the srcSet nothing from my side.

nensidosari (approved):
Looks great on my side as well!! 😄 🥳 

Did we want to give this a quick try on preview first?

sshanzel:
> Did we want to give this a quick try on preview first?

These are under experiment flags, so if we really want to test out the variations, we're gonna have to pull this locally. Though I have already tried all the variations, we can probably see this in action during the testing phase.

PR #2468, opened 2023-12-29
fix: centered profile and profile 404 page

omBratteng:
## Changes

Centers the profile page (and the new 404 page) by setting `screenCentered=false` on `MainLayout`

<table>
<tr>
 <td>Before</td>
 <td>After</td>
<tr>
 <td></td>
 <td></td>
<tr>
 <td>
<td>
</table>

### DIFF
--- a/packages/webapp/components/layouts/ProfileLayout/index.tsx
+++ b/packages/webapp/components/layouts/ProfileLayout/index.tsx
@@ -279,7 +279,9 @@ export const getLayout = (
   page: ReactNode,
   props: ProfileLayoutProps,
 ): ReactNode =>
-  getMainLayout(<ProfileLayout {...props}>{page}</ProfileLayout>, null);
+  getMainLayout(<ProfileLayout {...props}>{page}</ProfileLayout>, null, {
+    screenCentered: false,
+  });
 
 interface ProfileParams extends ParsedUrlQuery {
   userId: string;


rebelchris (commented):
How does the profile impact the 404 though? 
Seems im missing something.

Also not 100% if that had some negative effects on smaller/mobile screens cant recall

omBratteng:
Because we have the `[userId]` page, any page that does not exist, is tried through the profile

rebelchris:
Ah right, technically the 404 can show on multiple so wondered if only that had the issue.
If it didnt add any side effects we can try this

rebelchris (approved):

omBratteng:
Only saw that it was affected on this 404 page, all others that might get rendered looks centered.

Will double check side effects on mobile devices

omBratteng:
Actually, it might be a bigger issue than just the profile page. Noticed that other pages like history page doesn't center the content properly either. And there's some weird margin-left on the PageContainer/ResponsivePageContainer that skews the content a little bit to the left too.

https://github.com/dailydotdev/apps/blob/9a7edfdc20c3aaa396f1918bfa9d999206158802/packages/shared/src/components/utilities/utilities.module.css#L7

rebelchris:
@omBratteng Yeah I think this is a bit messy at the moment.
Perhaps we rather log it, as already it's not perfect and come up with a decent cross pages fix?

omBratteng:
Yeah, I'll log it instead and just close this.

PR #2470, opened 2023-12-31
feat: profile v2

idoshamun:

Profile page is revamped! This is our new year's surprise: more tech debt! 🎁 

There's no official spec, but you can look at the [figma design](https://www.figma.com/file/Ywxj2MFtLVi2lnRuhgeS80/daily.dev---Mockup?node-id=1842%3A23808&mode=dev)

New features that were added:
* Profile readme - Users can now write markdown that will be featured in their profile
* Cover image - Users can upload a cover image alongside their profile image

The new UI was built as a mobile-first, and I did my best to optimize the experience.

It's deployed to preview3, so you can check that out. There are mobile and tablet+ designs and different edge cases if you visit your profile or someone else's.

Here's a screenshot of my profile on mobile:

It's almost ready, I have some missing bits:
- [x] Tests
- [x] Edge cases support for no squads/no bio/no social links
- [x] When navigating through tabs, the scroll is reset
- [ ] Some more QA and bug hunting

Feel free to give it a high-level review in the meantime, or even just play around

## Events

Nothing new

- [x] MobileL (420px)
- [x] Tablet (656px)
- [x] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android


packages/shared/__tests__/fixture/feed.ts
```diff
@@ -70,7 +70,7 @@ const feed: Connection<Post> = {
     {
       node: {
         id: '0e4005b2d3cf191f8c44c2718a457a1e',
-        title: 'Learn SQL & MongoDb Simultaneously — The Easy Way (Part 1)',
+        title: 'Learn SQL & MongoDb Simultaneously The Easy Way (Part 1)',
```
idoshamun, line 73: The unicode messed my test so I had to remove it

packages/shared/src/components/MainLayout.tsx
```diff
@@ -81,11 +80,6 @@ function MainLayout({
   useAuthVerificationRecovery();
   useNotificationParams();
   useReferralReminder();
-  const handlers = useSwipeableSidebar({
```
idoshamun, line 84: Nimrod asked to remove completely the swipes for now

packages/shared/src/components/UserBadge.tsx
```diff
@@ -8,25 +8,34 @@ export type UserBadgeProps = {
...
+  removeMargins?: boolean;
+  disableResponsive?: boolean;
 };
 
 const UserBadge = ({
   className,
   content,
   Icon,
   iconProps,
+  removeMargins,
```
idoshamun, line 20: The squad component in the new profile doesn't need the responsiveness

packages/shared/src/components/comments/CommentBox.tsx
```diff
@@ -74,13 +78,18 @@ function CommentBox({
       )}
       data-testid="comment"
     >
+      {linkToComment && (
```
idoshamun, line 81: The replies feed need to navigate to the comment on click, just like cards

packages/shared/src/components/fields/MarkdownInput/index.tsx
```diff
@@ -276,6 +276,7 @@ function MarkdownInput(
               )}
               icon={icon}
               onClick={() => uploadRef?.current?.click()}
+              type="button"
```
idoshamun, line 279: This fixed an issue that upon clicking on any icon on the markdown bar it submitted the form

packages/shared/src/components/profile/RanksWidget.tsx
```diff
@@ -0,0 +1,73 @@
...
+  >
+    <Rank className="w-8 h-8" rank={rank} colorByRank />
+    <span className="hidden tablet:block ml-1 text-theme-label-tertiary">
+      {rankName}
+    </span>
+    <span className="tablet:ml-auto">{count}</span>
+  </div>
+);
+
+export function RanksWidget({
```
idoshamun, line 34: Copied from the previous implementation

packages/shared/src/components/profile/ReadingHeatmapWidget.tsx
```diff
@@ -0,0 +1,111 @@
...
+export interface ReadingHeatmapWidgetProps {
+  fullHistory: boolean;
+  selectedHistoryYear: number;
+  readHistory: UserReadHistory[];
+  before: Date;
+  after: Date;
+  yearOptions: string[];
+}
+
+export function ReadingHeatmapWidget({
```
idoshamun, line 57: Copied from the previous implementation

packages/shared/src/components/profile/ReadingTagsWidget.tsx
```diff
@@ -0,0 +1,29 @@
...
+import { MostReadTag } from '../../graphql/users';
+import { ActivityContainer, ActivitySectionHeader } from './ActivitySection';
+import { weeklyGoal } from '../../lib/constants';
+import { ReadingTagProgress } from './ReadingTagProgress';
+
+export interface ReadingTagsWidgetProps {
+  mostReadTags: MostReadTag[];
+}
+
+export function ReadingTagsWidget({
```
idoshamun, line 11: Copied from the previous implementation

packages/shared/src/components/profile/Readme.tsx
```diff
@@ -0,0 +1,130 @@
...
+  const isSameUser = loggedUser?.id === user.id;
+  const readme = user?.readmeHtml;
+
+  // Markdown is supported only in the client due to sanitization
+  const isClient = typeof window !== 'undefined';
+
+  const [editMode, setEditMode] = useState(false);
+
+  const queryKey = generateQueryKey(RequestKey.Readme, user);
+  const { data: remoteReadme, isLoading } = useQuery<{
```
idoshamun, line 39: Fetches the readme as markdown only when needed. Lessons learned from the comments

packages/shared/src/graphql/common.ts
```diff
@@ -124,3 +125,12 @@ export const errorMessage = {
 export interface ReferredUsersData {
   referredUsers: Connection<PublicProfile>;
 }
+
+export interface MutationError {
```
idoshamun, line 129: Moved to a shared place

packages/shared/src/graphql/posts.ts
```diff
@@ -1,5 +1,5 @@
 import request, { gql } from 'graphql-request';
-import { Author, Comment, Scout } from './comments';
+import type { Author, Comment, Scout } from './comments';
```
idoshamun, line 2: To prevent circular dependency

packages/shared/src/hooks/comments/useDeleteComment.ts
```diff
@@ -0,0 +1,56 @@
...
+  description:
+    'Are you sure you want to delete your comment? This action cannot be undone.',
+  okButton: {
+    title: 'Delete',
+    variant: ButtonVariant.Primary,
+    color: ButtonColor.Cabbage,
+  },
+};
+
+export function useDeleteComment(): UseDeleteCommentRet {
```
idoshamun, line 34: Refactored to a standalone hook so we can reuse it in multiple places

packages/shared/src/useDynamicHeader.ts
```diff
@@ -0,0 +1,52 @@
+import { Ref, useEffect, useRef, useState } from 'react';
+
+export type UseDynamicHeaderRet<T extends HTMLElement> = {
+  ref: Ref<T>;
+  progress: number | undefined;
+};
+
+export const useDynamicHeader = <T extends HTMLElement>(
```
idoshamun, line 8: Used to implement the sticky header on scroll with smooth transition

packages/webapp/pages/[userId]/index.tsx
```diff
@@ -1,306 +1,104 @@
...
+  const {
+    selectedHistoryYear,
+    setSelectedHistoryYear,
+    before,
+    after,
+    yearOptions,
+    fullHistory,
+  } = useActivityTimeFilter();
+
+  const { data: user } = useQuery(
```
Comment (idoshamun) [inline packages/webapp/pages/[userId]/index.tsx:43]: I introduced this query also here for users who visit their own profile to make sure the readme is up to date and not cached. Otherwise, it creates a weird UX where you update the readme and it still shows the cached version

packages/webapp/pages/account/profile.tsx
```diff
@@ -157,15 +209,6 @@ const AccountProfilePage = (): ReactElement => {
             name="github"
             value={user.github}
           />
-          <AccountTextField
```
idoshamun, line 160: Requested to remove

packages/shared/src/components/MainLayout.tsx
rebelchris: Let's see how long before people start complaining 🤣

packages/shared/src/components/ProfilePicture.tsx
```diff
@@ -103,7 +104,7 @@ function ProfilePictureComponent(
         alt={`${user.username || user.id}'s profile`}
         onError={onError}
         className={classes}
-        loading={isCompanion ? 'eager' : 'lazy'}
+        loading={eager || isCompanion ? 'eager' : 'lazy'}
```
rebelchris, line 107: We actually prefer/wanted to move all to `nativeLazyLoading` would it not work on this one?

packages/shared/src/components/comments/MainComment.tsx
```diff
@@ -61,15 +65,21 @@ export default function MainComment({
...
     >
       {!editProps && (
         <CommentBox
           {...props}
           comment={comment}
           parentId={comment.id}
-          className={{ container: 'border-b' }}
+          className={{
+            container: 'border-b',
+            ...commentBoxClassName,
```
rebelchris, line 81: This won't do much right?
className here is a object, we read it as `className.container` etc.
Not sure you need it altogether, if so you need to adjust the passing.

packages/shared/src/components/comments/MainComment.tsx
```diff
@@ -32,6 +35,7 @@ const shouldShowBannerOnComment = (
 
 export default function MainComment({
   className,
+  commentBoxClassName,
```
rebelchris, line 38: Open for this, but generally we prefer the object method of classNames when more then 1 are needed.

packages/shared/src/components/profile/Header.tsx
```diff
@@ -0,0 +1,81 @@
...
+  const [, onShareOrCopyLink] = useShareOrCopyLink({
+    text: `Check out ${user.name}'s profile on daily.dev`,
+    link: user.permalink,
+    trackObject: () => ({ event_name: 'share profile', target_id: user.id }),
+  });
+
+  return (
+    <header
+      className={classNames('flex items-center px-4 h-12', className)}
+      style={style}
```
rebelchris, line 35: We can't do this with tailwind?

packages/shared/src/components/profile/Header.tsx
```diff
@@ -0,0 +1,81 @@
...
+    >
+      {sticky ? (
+        <>
+          <ProfilePicture user={user} nativeLazyLoading size="medium" />
+          <div className="flex flex-col ml-2 typo-footnote">
+            <div className="font-bold">{user.name}</div>
+            <div className="text-theme-label-tertiary">
+              {largeNumberFormat(user.reputation)} Reputation
+            </div>
```
rebelchris, line 44: Do you need divs here or would `p` do to be more semantic as they only hold copy.

packages/shared/src/components/profile/Header.tsx
```diff
@@ -0,0 +1,81 @@
...
+        >
+          Edit profile
+        </Button>
+      )}
+      <Button
+        className={classNames('ml-auto', isSameUser && 'tablet:ml-0')}
+        variant={ButtonVariant.Float}
+        size={ButtonSize.Small}
+        icon={<ShareIcon />}
+        onClick={() => onShareOrCopyLink()}
```
rebelchris, line 66:
```suggestion
        onClick={onShareOrCopyLink}
```

packages/shared/src/components/profile/Header.tsx
```diff
@@ -0,0 +1,81 @@
...
+      />
+      {isSameUser && (
+        <Button
+          className="tablet:hidden ml-2"
+          variant={ButtonVariant.Float}
+          size={ButtonSize.Small}
+          icon={<SettingsIcon />}
+          tag="a"
+          href={`${process.env.NEXT_PUBLIC_WEBAPP_URL}account/profile`}
+          aria-label="Edit profile"
```
rebelchris, line 76: These get added automatic right (from what I recall)

packages/shared/src/components/profile/HeroImage.tsx
```diff
@@ -0,0 +1,38 @@
...
+  ref: Ref<HTMLDivElement>,
+): ReactElement {
+  return (
+    <div className="flex relative mx-4 h-24" ref={ref}>
+      {cover ? (
+        <Image
+          src={cover}
+          alt="cover"
+          loading="eager"
+          className="object-cover absolute top-0 left-0 -z-1 w-full h-full rounded-26"
```
rebelchris, line 21: We should now have access to `size-full` in favor of `w-full + h-full`

packages/shared/src/components/profile/HeroImage.tsx
```diff
@@ -0,0 +1,38 @@
...
+        />
+      ) : (
+        <div className="absolute top-0 left-0 -z-1 w-full h-full rounded-26 bg-theme-bg-secondary" />
```
rebelchris, line 24: Same here:
We should now have access to `size-full` in favor of `w-full + h-full`

packages/shared/src/components/profile/ProfileWidgets.tsx
```diff
@@ -0,0 +1,101 @@
...
+      className={classNames('flex flex-col gap-6 my-4 tablet:my-0', className)}
+    >
+      <Header user={user} isSameUser={isSameUser} className="-mb-2" />
+      {!hideSticky && (
+        <Header
+          user={user}
+          isSameUser={isSameUser}
+          sticky
+          className="fixed top-0 left-0 z-3 w-full transition-transform duration-75 bg-theme-bg-primary"
+          style={{ transform: `translateY(${(stickyProgress - 1) * 100}%)` }}
```
rebelchris, line 54: Ah I see :).

packages/shared/src/components/profile/RanksWidget.tsx
```diff
@@ -0,0 +1,73 @@
...
+const RankHistory = ({
+  rank,
+  rankName,
+  count,
+}: RankHistoryProps): ReactElement => (
+  <div
+    className="flex flex-col tablet:flex-row items-center p-2 tablet:py-1 tablet:pr-4 tablet:pl-2 font-bold rounded-12 border typo-callout border-theme-bg-secondary"
+    aria-label={`${rankName}: ${count}`}
+  >
+    <Rank className="w-8 h-8" rank={rank} colorByRank />
```
rebelchris, line 26: Should also have `size-8` now

packages/shared/src/components/profile/Readme.tsx
rebelchris: Optional could have had these in a useProfile hook or something, but not blocking the PR

packages/shared/src/components/profile/Readme.tsx
```diff
@@ -0,0 +1,130 @@
...
+      }),
+    {
+      enabled: editMode,
+      refetchOnWindowFocus: false,
+      refetchOnReconnect: false,
+      refetchOnMount: false,
+    },
+  );
+
+  const { mutateAsync: updateReadme, isLoading: submitting } = useMutation<
```
rebelchris, line 55: This part as well in that case above of a hook

packages/shared/src/components/profile/Readme.tsx
```diff
@@ -0,0 +1,130 @@
...
+      if (!message) {
+        return;
+      }
+
+      displayToast(message);
+    },
+  });
+
+  let children: ReactNode;
+  if (editMode) {
```
rebelchris, line 79: This feels a bit messy to me, should we not prefer 2 components in this case?
1 view and 1 edit one.

packages/shared/src/components/profile/SocialChips.tsx
```diff
@@ -0,0 +1,72 @@
...
+    label: (x) => `@${x}`,
+  },
+  portfolio: {
+    icon: <LinkIcon />,
+    href: (x) => x,
+    // Strip protocol from url
+    label: (x) => x.replace(/(^\w+:|^)\/\//, ''),
+  },
+};
+const order: (keyof SocialChipsProps['links'])[] = [
```
rebelchris, line 36: could also have added `order` in the above, but this is fine to me.

packages/shared/src/components/profile/SocialChips.tsx
```diff
@@ -0,0 +1,72 @@
...
+  'github',
+  'twitter',
+  'portfolio',
+];
+
+export function SocialChips({ links }: SocialChipsProps): ReactElement {
+  const elements = order
+    .filter((key) => !!links[key])
+    .map((key) => (
+      <Button
```
rebelchris, line 46: Did you choose the old button v1 on purpose?
Technically it should be unused soon as we plan to remove it.

packages/shared/src/components/profile/SquadsList.tsx
```diff
@@ -0,0 +1,204 @@
...
+
+      return;
+    }
+
+    await joinSquad();
+  };
+
+  return (
+    <div
+      className="flex relative items-center p-2 first:ml-4 tablet:first:ml-0 w-40 tablet:w-auto bg-theme-float hover:bg-theme-hover rounded-2xl"
```
rebelchris, line 78: `rounded-2xl` should be the numeric version not sure how much it comes to but something like `rounded-12`

packages/shared/src/components/profile/SquadsList.tsx
```diff
@@ -0,0 +1,204 @@
...
+    >
+      <Link href={squad.permalink} prefetch={false} passHref>
+        <CardLink />
+      </Link>
+      <div className="flex overflow-hidden flex-col flex-1">
+        <div className="flex gap-2 items-center">
+          <Image
+            src={squad.image}
+            alt={squad.name}
+            className="w-8 h-8 rounded-full"
```
rebelchris, line 89: size-8 should do

packages/shared/src/components/profile/SquadsList.tsx
```diff
@@ -0,0 +1,204 @@
...
+          />
+          <div className="flex flex-col flex-1">
+            <div className="overflow-hidden font-bold whitespace-nowrap typo-caption1 text-ellipsis">
+              {squad.name}
+            </div>
+            <div className="overflow-hidden whitespace-nowrap text-theme-label-quaternary typo-caption2 text-ellipsis">
+              @{squad.handle}
+            </div>
```
rebelchris, line 97: These should be divs or can we get away with paragraphs?

packages/shared/src/components/profile/SquadsList.tsx
```diff
@@ -0,0 +1,204 @@
...
+          </span>
+          {showJoin && (
+            <Button
+              className="tablet:hidden z-1 ml-auto"
+              variant={ButtonVariant.Secondary}
+              size={ButtonSize.XSmall}
+              icon={<PlusIcon />}
+              onClick={onJoin}
+              loading={isLoading}
+              aria-label="Join Squad"
```
rebelchris, line 122: Don't think you need this specifically? (but could mistake it)

packages/shared/src/components/profile/SquadsList.tsx
```diff
@@ -0,0 +1,204 @@
...
+        <Button
+          variant={ButtonVariant.Float}
+          size={ButtonSize.Large}
+          tag="a"
+          href={`${webappUrl}squads/new?origin=profile`}
+          icon={<PlusIcon />}
+          aria-label="Create a new Squad"
+        />
+        <div className="rounded-2xl border flex-[3] border-theme-divider-tertiary" />
+        <div className="rounded-l-2xl border flex-[2] border-theme-divider-tertiary" />
```
rebelchris, line 183: These can also have the radius numeric preferably

packages/shared/src/hooks/profile/useActivityTimeFilter.ts
```diff
@@ -0,0 +1,58 @@
...
+      const start = startOfTomorrow();
+      return [start, subYears(subDays(start, 2), 1)];
+    }
+    const selected = parseInt(dropdownOptions[selectedHistoryYear], 10);
+    const startYear = new Date(selected, 0, 1);
+
+    return [addDays(endOfYear(startYear), 1), startYear];
+    // @NOTE see https://dailydotdev.atlassian.net/l/cp/dK9h1zoM
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [selectedHistoryYear]);
```
rebelchris, line 48: Should we not add all deps here to avoid breaking the rules.
Shouldn't make too much difference I think.

packages/webapp/components/layouts/ProfileLayout/v2.tsx
```diff
@@ -0,0 +1,142 @@
+import React, { ReactElement, ReactNode } from 'react';
```
rebelchris, line 1: With removal of index should we not just call this `index`?

packages/webapp/pages/[userId]/index.tsx
```diff
@@ -1,306 +1,104 @@
...
+    generateQueryKey(RequestKey.Profile, initialUser),
+    () => getProfile(initialUser?.id),
+    {
+      placeholderData: initialUser,
+      refetchOnReconnect: false,
+      refetchOnMount: false,
+      refetchOnWindowFocus: false,
+      // Make sure the logged in user's profile is always up to date
+      enabled: !!initialUser && isSameUser,
```
Comment (rebelchris) [inline packages/webapp/pages/[userId]/index.tsx:52]: This would just mean it will always refetch even if you didn't change, not needed for this PR, but actually had a thought about it.

1. Can we not manually manipulate the cache as we do in other places, that should work I think
2. If issues we could actually start using back redirects with query params like ?r=true as when `r` is set we refresh enable?

Although think we prefer 1 with a staleTime?

rebelchris (changes requested):
Amazing amount of work!
Some minor comments, you can take some with a grain of salt as suggestions :)

packages/shared/src/components/ProfilePicture.tsx
idoshamun: Wdym? It's still native lazy loading, but I wanted to add an option to eagerly load them. The default is lazy load but to optimize the profile page, I wanted above the fold images to load immediately

packages/shared/src/components/profile/Header.tsx
idoshamun: It's not a `p` really but maybe a `span` would be better then. Right?

packages/shared/src/components/profile/Header.tsx
idoshamun: TypeScript complains because onClick has some args

packages/shared/src/components/profile/Header.tsx
idoshamun: Wdym?

packages/shared/src/components/profile/HeroImage.tsx
idoshamun: Neat! Didn't know that

packages/shared/src/components/profile/Header.tsx
idoshamun: I see that you already figured it out. So I guess it's ok :)

packages/shared/src/components/profile/SocialChips.tsx
idoshamun: Yes I copied an existing implementation that Tsahi referred me to. We'll be able to migrate this button only when we introduce the "Subtle" variation. Don't know the timeline for that though

packages/shared/src/components/ProfilePicture.tsx
rebelchris: We try to avoid the `lazy` one as it uses js lazy loading (from top of my head) we have a option for `nativeLazyLoading` which uses plain `img` elements props.

packages/shared/src/components/profile/SquadsList.tsx
idoshamun: Span perhaps?

packages/shared/src/components/profile/SquadsList.tsx
idoshamun: It's needed for accessibility since this button has nothing but icon

packages/shared/src/components/profile/Header.tsx
rebelchris: To me text is most of the time `p` but not sure how others view it.
Span could also work, but I tend to use it for inline things because of it's native inline behaviour

packages/shared/src/components/profile/Header.tsx
rebelchris: Ah yes seen this before :) nvm then

packages/shared/src/hooks/profile/useActivityTimeFilter.ts
idoshamun: I can. Just copied whatever existed into a standalone hook. Didn't touch it

packages/webapp/components/layouts/ProfileLayout/v2.tsx
idoshamun: Good catch!

packages/shared/src/components/profile/SocialChips.tsx
rebelchris: Ah ok got it :) 
Yeah there are 3/4 exclusions now we need to find out still so no problem for this PR.

packages/webapp/pages/[userId]/index.tsx
Comment (idoshamun) [reply packages/webapp/pages/[userId]/index.tsx:52]: This will not fix all the caching issues I'm trying to solve. The primary one is given we use next ISR, the page is cached for quite some time. But let's say I just updated my readme and then refresh the page, without this I'll see the old readme. With this I see always fresh data

Comment (rebelchris) [reply packages/webapp/pages/[userId]/index.tsx:52]: Ah yes, but in that case would the 2nd option not work of leveraging query params to force refreshes?
(it's not really blocking, more opening the discussion for future things as we have quite a few of these issues already)

Comment (idoshamun) [reply packages/webapp/pages/[userId]/index.tsx:52]: When the user manually decides to visit their profile again I cannot force them into adding query param, right? Or am I missing something?

packages/shared/src/components/ProfilePicture.tsx
idoshamun: I believe this is exactly what we do here

packages/webapp/pages/[userId]/index.tsx
Comment (rebelchris) [reply packages/webapp/pages/[userId]/index.tsx:52]: That case no, my suggestion was:
- user edit profile
- on save we invoke toast + redirect back to profile + ?r=true

(but now I'm not sure if we didn't redirect for specific reason maybe)

packages/shared/src/components/ProfilePicture.tsx
rebelchris: Oh sigh, I for some reason thought this was the bottom evaluation 🤦

packages/webapp/pages/[userId]/index.tsx
Comment (idoshamun) [reply packages/webapp/pages/[userId]/index.tsx:52]: I feel like this is not a coherent solution for my use case then. If that's ok can we keep it as is for now?

Comment (rebelchris) [reply packages/webapp/pages/[userId]/index.tsx:52]: Yeah 100% ok for this PR :)

packages/shared/src/hooks/profile/useProfile.ts
```diff
@@ -0,0 +1,25 @@
...
+
+export function useProfile(initialUser?: PublicProfile): PublicProfile {
+  const { user: loggedUser } = useContext(AuthContext);
+  const isSameUser = loggedUser?.id === initialUser?.id;
+
+  const { data: user } = useQuery(
+    generateQueryKey(RequestKey.Profile, initialUser),
+    () => getProfile(initialUser?.id),
+    {
+      placeholderData: initialUser,
```
rebelchris, line 15: This initialUser is already fetched data right, I think you might want to use `initialData` in that case?
It's a small nuance, but initial will persist already.

https://tkdodo.eu/blog/placeholder-and-initial-data-in-react-query

packages/shared/src/components/profile/Header.tsx
rebelchris: Ah no never mind, I thought we added aria-label automatically in the button, but we don't 😢

rebelchris (approved):
Nothing blocking from my side 🚀

packages/shared/src/hooks/profile/useProfile.ts
idoshamun: My default is initial data but here I specifically wanted placeholder to make sure I fetch fresh data for a specific use case. Otherwise, I may have shown cached data to the user

PR #2473, opened 2024-01-02
fix: freeform card

sshanzel:
## Changes
- Fixed the freeform card's max clamp. From the design it should have been 3 lines only, similar to article post card.

From Sketch:
image

Before fix:
Screenshot 2024-01-02 at 11 17 19 AM

After the fix:
Screenshot 2024-01-02 at 11 20 25 AM

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done

### DIFF
--- a/packages/shared/src/components/cards/AdCard.tsx
+++ b/packages/shared/src/components/cards/AdCard.tsx
@@ -31,9 +31,7 @@ export const AdCard = forwardRef(function AdCard(
     <Card {...domProps} data-testid="adItem" ref={ref}>
       <AdLink ad={ad} onLinkClick={onLinkClick} />
       <CardTextContainer>
-        <CardTitle
-          className={classNames('my-4 line-clamp-4 font-bold typo-title3')}
-        >
+        <CardTitle className="my-4 font-bold line-clamp-4 typo-title3">
           {ad.description}
         </CardTitle>
       </CardTextContainer>
--- a/packages/shared/src/components/cards/AdCard.tsx
+++ b/packages/shared/src/components/cards/AdCard.tsx
@@ -31,7 +31,7 @@ export const AdCard = forwardRef(function AdCard(
     <Card {...domProps} data-testid="adItem" ref={ref}>
       <AdLink ad={ad} onLinkClick={onLinkClick} />
       <CardTextContainer>
-        <CardTitle className="my-4 font-bold line-clamp-4 typo-title3">
+        <CardTitle className="my-4 line-clamp-4 font-bold typo-title3">
           {ad.description}
         </CardTitle>
       </CardTextContainer>
--- a/packages/shared/src/components/cards/AdList.tsx
+++ b/packages/shared/src/components/cards/AdList.tsx
@@ -12,7 +12,7 @@ export const AdList = forwardRef(function AdList(
     <ListCard {...domProps} data-testid="adItem" ref={ref}>
       <AdLink ad={ad} onLinkClick={onLinkClick} />
       <ListCardMain>
-        <ListCardTitle className="font-bold line-clamp-4 typo-title3">
+        <ListCardTitle className="line-clamp-4 font-bold typo-title3">
           {ad.description}
         </ListCardTitle>
         <AdAttribution ad={ad} className="mt-2" />
--- a/packages/shared/src/components/cards/AdList.tsx
+++ b/packages/shared/src/components/cards/AdList.tsx
@@ -1,5 +1,4 @@
 import React, { forwardRef, ReactElement, Ref } from 'react';
-import classNames from 'classnames';
 import { AdCardProps } from './AdCard';
 import { ListCard, ListCardMain, ListCardTitle } from './Card';
 import AdLink from './AdLink';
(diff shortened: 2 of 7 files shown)


packages/shared/src/components/cards/AdCard.tsx
```diff
@@ -31,9 +31,7 @@ export const AdCard = forwardRef(function AdCard(
     <Card {...domProps} data-testid="adItem" ref={ref}>
       <AdLink ad={ad} onLinkClick={onLinkClick} />
       <CardTextContainer>
-        <CardTitle
-          className={classNames('my-4 line-clamp-4 font-bold typo-title3')}
-        >
+        <CardTitle className="my-4 font-bold line-clamp-4 typo-title3">
```
sshanzel, line 34: I was checking through implementations of `line-clamp-4` and just noticed it is using `classnames` unnecessarily.

packages/shared/src/components/cards/AdList.tsx
```diff
@@ -13,9 +12,7 @@ export const AdList = forwardRef(function AdList(
     <ListCard {...domProps} data-testid="adItem" ref={ref}>
       <AdLink ad={ad} onLinkClick={onLinkClick} />
       <ListCardMain>
-        <ListCardTitle
-          className={classNames('line-clamp-4 font-bold typo-title3')}
-        >
+        <ListCardTitle className="font-bold line-clamp-4 typo-title3">
```
sshanzel, line 15: I was checking through implementations of `line-clamp-4` and just noticed it is using `classnames` unnecessarily. Nothing really changed here.

packages/shared/src/components/cards/CollectionCard/CollectionCard.tsx
```diff
@@ -34,7 +34,7 @@ export const CollectionCard = forwardRef(function CollectionCard(
       return 'line-clamp-3';
     }
 
-    return post.contentHtml ? 'line-clamp-4' : 'line-clamp-9';
+    return post.contentHtml ? 'line-clamp-3' : 'line-clamp-9';
```
rebelchris, line 37: Only with image should it be 3 from the design.
- title only: 9
- title + description: 4
- title + image: 3

Could you add local test + screenshots for those versions?

packages/shared/src/components/cards/WelcomePostCard.tsx
```diff
@@ -52,7 +52,7 @@ export const WelcomePostCard = forwardRef(function SharePostCard(
       return 'line-clamp-3';
     }
 
-    return post.contentHtml ? 'line-clamp-4' : 'line-clamp-9';
+    return post.contentHtml ? 'line-clamp-3' : 'line-clamp-9';
```
rebelchris, line 55: See same as above :)

rebelchris (changes requested):
Minor feedback from the sketch.
Also noticed the spacing is missing, could you double check that one? (see [slack](https://dailydotdev.slack.com/archives/C01L0QXQJNB/p1704178299973289?thread_ts=1704102670.284819&cid=C01L0QXQJNB))

packages/shared/src/components/cards/CollectionCard/CollectionCard.tsx
sshanzel: Definitely. Let me check this.

sshanzel: After further checking, the issue is not with the rules inside the code but rather how we check the image. In here, there is no cover image, so it returns the clamp for 4 which is correct, but then we have a process of how we define the image to use which was not utilized here (which will then say we should have used line clamp 3).

sshanzel:
To centralize the image to use for the cover and verification of clamps, I introduced a hook and then passed the value through props.

image

packages/shared/src/components/cards/CollectionCard/CollectionCard.tsx
```diff
@@ -29,6 +30,7 @@ export const CollectionCard = forwardRef(function CollectionCard(
   }: PostCardProps,
   ref: Ref<HTMLElement>,
 ) {
+  const image = usePostImage(post);
   const clamp = (() => {
     if (post.image) {
```
rebelchris, line 35: Should this not check on `image` then?

packages/shared/src/hooks/post/usePostImage.ts
```diff
@@ -0,0 +1,18 @@
...
+    }
+
+    const parser = new DOMParser();
+    const doc = parser.parseFromString(post?.contentHtml, 'text/html');
+    const imgTag = doc.querySelector('img');
+    if (imgTag) {
+      return imgTag.getAttribute('src');
+    }
+
+    return undefined;
```
rebelchris, line 17: Do we want to return undefined or null?

sshanzel: I think undefined should be alright. Though I just copy-pasted this from the original implementation.

packages/shared/src/components/cards/CollectionCard/CollectionCard.tsx
sshanzel: Right, forgot to update the other one. Maybe I should introduce a function that generates this since they are of the same result.

sshanzel:
> Nice could you perhaps also check that spacing between the title and text I see in your screenshot it's still wrong.

Right. Completely forgot about this. What a sloppy start from the holidays 😅

sshanzel:
Applied the top margin required for description.

Sketch:
Screenshot 2024-01-02 at 9 14 38 PM

Implementation:
Screenshot 2024-01-02 at 9 13 23 PM

PR #2476, opened 2024-01-02
fix overflowing action buttons on feed cards

kopancek:
3rd try - this time we are using different media queries depending on if the sidebar is rendered closed, rendered open or not rendered at all.

We do not need different configurations this way and align to the minimal size of 272px for the single card, as defined in Figma.

## Screenshot

Before:
image

After:
Screenshot 2024-01-03 at 10 45 23

## Manual Testing

Manually checked all the combinations of sidebar closed, open and not rendered at all and settings eco, cozy and roomy as well.

- [X] MobileL (420px)
- [X] Tablet (656px)
- [X] Laptop (1020px)
- [X] Laptop XL (2156px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-2079 #done

### DIFF
--- a/packages/shared/src/lib/featureManagement.ts
+++ b/packages/shared/src/lib/featureManagement.ts
@@ -16,7 +16,7 @@ const feature = {
   feedVersion: new Feature('feed_version', 15),
   search: new Feature('search', SearchExperiment.Control),
   lowImps: new Feature('feed_low_imps'),
-  bookmarkOnCard: new Feature('bookmark_on_card', false),
+  bookmarkOnCard: new Feature('bookmark_on_card', true),
   onboardingCopy: new Feature('onboarding_copy', {
     title: 'Where developers grow together',
     description: 'Get one personalized feed for all the knowledge you need.',
--- a/packages/shared/src/lib/featureManagement.ts
+++ b/packages/shared/src/lib/featureManagement.ts
@@ -16,7 +16,7 @@ const feature = {
   feedVersion: new Feature('feed_version', 15),
   search: new Feature('search', SearchExperiment.Control),
   lowImps: new Feature('feed_low_imps'),
-  bookmarkOnCard: new Feature('bookmark_on_card', true),
+  bookmarkOnCard: new Feature('bookmark_on_card', false),
   onboardingCopy: new Feature('onboarding_copy', {
     title: 'Where developers grow together',
     description: 'Get one personalized feed for all the knowledge you need.',
--- a/packages/shared/src/styles/media.ts
+++ b/packages/shared/src/styles/media.ts
@@ -6,3 +6,4 @@ export const laptopL = `@media (min-width: 1360px)`;
 export const laptopXL = `@media (min-width: 1668px)`;
 export const desktop = `@media (min-width: 1976px)`;
 export const desktopL = `@media (min-width: 2156px)`;
+export const desktopXL = `@media (min-width: 2294px)`;
--- a/packages/shared/src/components/FeedLayout.tsx
+++ b/packages/shared/src/components/FeedLayout.tsx
@@ -99,7 +99,7 @@ const replaceDigitsWithIncrement = (str: string, increment: number): string => {
   if (!match) {
     return str;
   }
-  return str.replace(match[0], `${parseInt(match[0]) + increment}`);
+  return str.replace(match[0], `${parseInt(match[0], 10) + increment}`);
 };
 
 const sidebarRenderedWidth = 44;
@@ -118,11 +118,12 @@ const determineBreakPoints = ({
   sidebarExpanded,
   sidebarRendered,
 }: DetermineFeedSettingsProps): string[] => {
-  return sidebarRendered
-    ? sidebarExpanded
+  if (sidebarRendered) {
+    return sidebarExpanded
       ? reversedBreakpointsSidebarOpen
-      : reversedBreakpointsSidebarRendered
-    : reversedBreakpoints;
+      : reversedBreakpointsSidebarRendered;
+  }
+  return reversedBreakpoints;
 };
 
 export default function FeedLayout({
--- a/packages/shared/src/components/FeedLayout.tsx
+++ b/packages/shared/src/components/FeedLayout.tsx
@@ -1,7 +1,7 @@
 import React, { ReactElement, ReactNode, useContext } from 'react';
 import {
   desktop,
-  desktopL,
+  desktopXL,
   laptop,
   laptopL,
   laptopXL,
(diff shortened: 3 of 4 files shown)


sshanzel (changes requested):
I believe the value 6 is part of our design system. I think the primary issue is with the card's `min-width` that it is smaller than how it should be.

Similar to the issue with the card's height recently, there were some discrepancies with the spacings. Maybe it is a good idea to check whether they fit the guideline. Kindly check whether we have the right image/content size and their respective paddings for the card.

kopancek:
> I believe the value 6 is part of our design system. I think the primary issue is with the card's `min-width` that it is smaller than how it should be.
> 
> Similar to the issue with the card's height recently, there were some discrepancies with the spacings (vertically). Maybe it is a good idea to check whether they fit the guidelines (horizontally). Kindly check whether we have the right image/content size and their respective paddings for the card.
> 
> Link to the guideline: https://www.figma.com/file/C7n8EiXBwV1sYIEHkQHS8R/daily.dev---Design-System?node-id=483%3A1131&mode=dev

yes, value 6 is part of the design system, but the actual implementation differs a lot.

E.g. we use the same media queries on the whole viewport width, regardless of whether sidebar is rendered or not, whether it is expanded or not. But the design system breakpoints for column counts are based on the size of the feed container.

Maybe a better way would be to define the minmax of the column width based on the design system and let the browser decide how many columns it can fit. Something like this:
```
grid-template-columns: repeat(auto-fit, minmax(272px, 320px));
```

We can then have a media query to limit the number of columns displayed to the maximum specified by the design system. WDYT about that solution?

kopancek:
> > I believe the value 6 is part of our design system. I think the primary issue is with the card's `min-width` that it is smaller than how it should be.
> > Similar to the issue with the card's height recently, there were some discrepancies with the spacings (vertically). Maybe it is a good idea to check whether they fit the guidelines (horizontally). Kindly check whether we have the right image/content size and their respective paddings for the card.
> > Link to the guideline: https://www.figma.com/file/C7n8EiXBwV1sYIEHkQHS8R/daily.dev---Design-System?node-id=483%3A1131&mode=dev
> 
> yes, value 6 is part of the design system, but the actual implementation differs a lot.
> 
> E.g. we use the same media queries on the whole viewport width, regardless of whether sidebar is rendered or not, whether it is expanded or not. But the design system breakpoints for column counts are based on the size of the feed container.
> 
> Maybe a better way would be to define the minmax of the column width based on the design system and let the browser decide how many columns it can fit. Something like this:
> 
> ```
> grid-template-columns: repeat(auto-fit, minmax(272px, 320px));
> ```
> 
> We can then have a media query to limit the number of columns displayed to the maximum specified by the design system. WDYT about that solution?

But even if we keep to that definition, the buttons do overflow if the comment or upvote numbers are high enough, as seen on this screenshot. I think we need to talk to design about what to do about the lower bound there, as the buttons simply do not fit into 272px, unless we make the spacing between them smaller or something.
Screenshot 2024-01-02 at 16 32 14

sshanzel:
> Maybe a better way would be to define the minmax of the column width based on the design system and let the browser decide how many columns it can fit. Something like this:

You will have to confer that with the design team.

> But even if we keep to that definition, the buttons do overflow if the comment or upvote numbers are high enough, as seen on this screenshot. I think we need to talk to design about what to do about the lower bound there, as the buttons simply do not fit into 272px, unless we make the spacing between them smaller or something.

This is the exact reason I was asking to have the exact elements checked is because sometimes when implementing from design to app, the result needs to be a bit bigger than expected due to how the browser calculate the actual elements vs from design apps. If you check the implementation of the height, it does not really match the exact number defined in the artboard but validated the exact sizes of the elements.

sshanzel:
Also, I noticed some paddings with the Quaternary button, not sure they are really needed for this type.
Screenshot 2024-01-02 at 11 41 29 PM

rebelchris:
The last change wont work, it will introduce layout shifts, we put the widths there on propose.

My suggestion, for now lets push the 5 vs 6 fix and log the rest for research, else we introduce too many new bugs.

rebelchris (approved):
Happy to give this a try 🍻

packages/shared/src/lib/featureManagement.ts
```diff
@@ -16,7 +16,7 @@ const feature = {
   feedVersion: new Feature('feed_version', 15),
   search: new Feature('search', SearchExperiment.Control),
   lowImps: new Feature('feed_low_imps'),
-  bookmarkOnCard: new Feature('bookmark_on_card', false),
+  bookmarkOnCard: new Feature('bookmark_on_card', true),
```
sshanzel, line 19: Just a reminder to revert this before merge.

packages/shared/src/styles/media.ts
```diff
@@ -6,3 +6,4 @@ export const laptopL = `@media (min-width: 1360px)`;
 export const laptopXL = `@media (min-width: 1668px)`;
 export const desktop = `@media (min-width: 1976px)`;
 export const desktopL = `@media (min-width: 2156px)`;
+export const desktopXL = `@media (min-width: 2294px)`;
```
sshanzel, line 9: Is this new breakpoint part of the design system?

sshanzel (commented):
Highly impactful changes, kindly check that nothing is out of the ordinary. Looks great to me code-wise. Kindly return the feature default value, then I think we should be good to go 🥳

packages/shared/src/styles/media.ts
kopancek: No, but the design system is broken. In figma the definition is: 2156 width should fit 7 columns. But if you multiply 7 columns with minimal width of 272px and add the column gaps of 32px and the padding of the container, you get to  `272px*7 + 32px*6 + 64px*2 = 2224px` which is more than the 2156px that the design system is talking about. 

2294 is basically just 2224 + 70px as a safety margin. This way, we keep the column size at at least 272px. It is similar to what we did with other breakpoints. E.g. for 6 columns, you only need `272px*6 + 32px*5 + 64px*2 = 1920px`, but the breakpoint is defined at 1976px instead.

packages/shared/src/lib/featureManagement.ts
kopancek: Oh no 🤦

packages/shared/src/styles/media.ts
sshanzel: I see. That's a great catch. And I see that you already messaged the design team. Thank you for the fix! 🚀

sshanzel (approved):
We should be good to go 🥳

PR #2477, opened 2024-01-03
fix(profile): apply ellipsis on long social link

idoshamun:

Long portfolio links broke the design of the page

Fixed version:
image

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done

### DIFF
--- a/packages/shared/src/components/profile/SocialChips.tsx
+++ b/packages/shared/src/components/profile/SocialChips.tsx
@@ -44,11 +44,11 @@ export function SocialChips({ links }: SocialChipsProps): ReactElement {
     .filter((key) => !!links[key])
     .map((key) => (
       <Button
-        spanClassName="w-fit my-2 font-normal"
+        spanClassName="my-2 font-normal text-ellipsis whitespace-nowrap overflow-hidden flex-1"
         textPosition="justify-start"
         icon={handlers[key].icon}
         buttonSize={ButtonSize.Small}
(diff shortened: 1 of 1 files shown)


packages/shared/src/components/profile/SocialChips.tsx
```diff
@@ -44,11 +44,11 @@ export function SocialChips({ links }: SocialChipsProps): ReactElement {
     .filter((key) => !!links[key])
     .map((key) => (
       <Button
-        spanClassName="w-fit my-2 font-normal"
+        spanClassName="my-2 font-normal text-ellipsis whitespace-nowrap overflow-hidden flex-1"
```
rebelchris, line 47: Dang see this is passed through, else we could use `getTextEllipsis` function/component.

rebelchris (approved):
Thanks for the quick fix.

packages/shared/src/components/profile/SocialChips.tsx
idoshamun: Sorry I auto merged. But I see that `getTextEllipsis` returns a component which I can't use as I'm using the button element. With that said, I believe having it by default in the button element can be super beneficial

rebelchris: Hmm yeah could be useful as the default on buttons indeed!

PR #2479, opened 2024-01-03
fix: add text truncating to sidebar items

omBratteng:
## Changes

<table>
<tr>
 <td>Before
 <td>After
<tr>
 <td>
 <td>
<tr>
  <td colspan="2">Bonus, added popover to truncated items<br></td>
</table>

- Fixed truncating on sidebar items
- Added title to sidebar items to show popover on hover

### DIFF
--- a/packages/shared/src/components/sidebar/Section.tsx
+++ b/packages/shared/src/components/sidebar/Section.tsx
@@ -67,6 +67,7 @@ export function Section({
                 : null
             }
             isButton={isItemsButton && !item?.isForcedLink}
+            className="truncate"
           >
             <ItemInner
               item={item}
--- a/packages/shared/src/components/sidebar/SquadSection.tsx
+++ b/packages/shared/src/components/sidebar/SquadSection.tsx
@@ -45,9 +45,6 @@ export function SquadSection(props: SectionCommonProps): ReactElement {
         ),
       title: name,
       path: permalink,
-      className: {
-        text: '!block',
-      },
     });
   });
 
--- a/packages/shared/src/components/sidebar/SquadSection.tsx
+++ b/packages/shared/src/components/sidebar/SquadSection.tsx
@@ -45,6 +45,9 @@ export function SquadSection(props: SectionCommonProps): ReactElement {
         ),
       title: name,
       path: permalink,
+      className: {
+        text: '!block',
+      },
     });
   });
 
--- a/packages/shared/src/components/sidebar/common.tsx
+++ b/packages/shared/src/components/sidebar/common.tsx
@@ -142,7 +142,7 @@ export const ItemInner = ({
       <Icon {...item} active={active} />
       <span
         className={classNames(
-          'truncate text-left transition-opacity',
+          'flex-1 truncate text-left transition-opacity',
           shouldShowLabel ? 'opacity-100 delay-150' : 'opacity-0',
           item?.className?.text,
         )}
--- a/packages/shared/src/components/sidebar/common.tsx
+++ b/packages/shared/src/components/sidebar/common.tsx
@@ -146,6 +146,7 @@ export const ItemInner = ({
           shouldShowLabel ? 'opacity-100 delay-150' : 'opacity-0',
           item?.className?.text,
         )}
+        title={item.title}
       >
         {item.title}
       </span>
--- a/packages/shared/src/components/sidebar/common.tsx
+++ b/packages/shared/src/components/sidebar/common.tsx
@@ -142,7 +142,7 @@ export const ItemInner = ({
       <Icon {...item} active={active} />
       <span
         className={classNames(
-          'flex flex-1 flex-row items-center truncate text-left transition-opacity',
+          'flex-1 truncate text-left transition-opacity',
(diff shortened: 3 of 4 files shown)


packages/shared/src/components/sidebar/Section.tsx
```diff
@@ -67,6 +67,7 @@ export function Section({
                 : null
             }
             isButton={isItemsButton && !item?.isForcedLink}
+            className="truncate"
```
rebelchris, line 70: Ah didn't realize we have access to this shorthand these days 🙈

packages/shared/src/components/sidebar/SquadSection.tsx
```diff
@@ -45,6 +45,9 @@ export function SquadSection(props: SectionCommonProps): ReactElement {
         ),
       title: name,
       path: permalink,
+      className: {
+        text: '!block',
```
rebelchris, line 49: What is the default here that we need the override? This feels like it might add a edge case somehow.

packages/shared/src/components/sidebar/common.tsx
```diff
@@ -146,6 +146,7 @@ export const ItemInner = ({
           shouldShowLabel ? 'opacity-100 delay-150' : 'opacity-0',
           item?.className?.text,
         )}
+        title={item.title}
```
rebelchris, line 149: This was just missing?

rebelchris (commented):
Just one blocking question on the block element, as potentially it might introduce a edge case on mobile etc where the rendering of the layout changes.

packages/shared/src/components/sidebar/common.tsx
omBratteng: Yeah, to get the popover (native in browser)

packages/shared/src/components/sidebar/SquadSection.tsx
omBratteng: `flex`. Couldn't get it to add the ellipsis with flex. Works with everything else but flex and inline flex.

This just adds `block` span that wraps around the squad name. Shouldn't affect anything else.

rebelchris: Hmm since that already has `flex` and alignments it might though.
Either those become redundant or we removed vertical centering for instance?

I think we should either fully remove the flex on that last child or see if we can fix it with the [min-width fix](https://css-tricks.com/flexbox-truncated-text/) described here?
Let me know what you think.

omBratteng: I tried the `min-width: 0` hack on all of the items in that hierarchy, but none seems to do the trick.

The flex that is being overridden in this case should not have any use-case anyway, as it's just a single span wrapping text, no html inside it. 
image
image

Removing `items-center` on that `span` from this span does not seem to have any effect either, as the parent `a` both has `flex` and `items-center`.

omBratteng: I thought about removing flex entirely from the `span` inside the `ItemInner` entirely, but didn't want to do that in case the `flex` might have an effect the other places it's being used.

rebelchris: Ah yes so let;s then remove the flex related classes as they might confuse people right (if not used now)

omBratteng: Removed them, now I also don't need to add the block class

omBratteng: Just to be sure, I will assign a preview to the pr and go through it to verify nothing is broken.

omBratteng: Checked on desktop and mobile, nothing seems to have been broken.

packages/shared/src/components/sidebar/common.tsx
```diff
@@ -142,10 +142,11 @@ export const ItemInner = ({
       <Icon {...item} active={active} />
       <span
         className={classNames(
-          'flex flex-1 flex-row items-center truncate text-left transition-opacity',
+          'flex-1 truncate text-left transition-opacity',
```
rebelchris, line 145: flex-1 also doesn't do anything anymore right?

rebelchris (approved):
non blocking

packages/shared/src/components/sidebar/common.tsx
omBratteng: Don't think so. I'll try to remove it to verify.
It could have an effect as the parent `a` has `flex` on it, but in this case I don't think it actually have any effect.

omBratteng: Pushed a change, will verify in preview as soon as it has been deployed.

rebelchris: On interesting as it's not even a flex item, but ok let's leave it and go for it :)

omBratteng: But the parent is using flex. And `flex-1` is for controlling the flex size of the children of that parent.
image

PR #2484, opened 2024-01-05
fix: code block paddings

sshanzel:
## Changes
- The fix on the multi-line code block padding should now be applied.
- Though on the inline ones, I can't seem to find the right pseudo-class to apply the proper paddings.

Fix:
image

Sketch:
Screenshot 2024-01-05 at 8 11 24 PM

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-1975 #done

### DIFF
--- a/packages/shared/src/components/markdown.module.css
+++ b/packages/shared/src/components/markdown.module.css
@@ -31,7 +31,7 @@
     @apply typo-markdown italic pl-7 my-5 border-l border-theme-divider-secondary;
   }
   & :where(pre) {
-    @apply typo-markdown relative bg-theme-float py-0 px-0 my-5 rounded-12 border border-theme-divider-quaternary;
+    @apply typo-markdown relative bg-theme-float py-7 px-6 my-5 rounded-12 border border-theme-divider-quaternary;
   }
 
   & :where(pre) > code {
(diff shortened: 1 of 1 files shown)


rebelchris (approved):
Just need to ensure it doesn't break content for freeform posts as they re-use this.
As for the inline, it does already look good no?

capJavert (approved):

omBratteng (approved):

omBratteng:
@rebelchris the inline is almost there, the only difference is the left/right padding when the inline codeblock breaks over multiple lines

omBratteng:
@sshanzel could apply the `box-decoration-break: clone` (`-webkit-box-decoration-break: clone`), it would look like this
image

compared to this without
image

[1] https://developer.mozilla.org/en-US/docs/Web/CSS/box-decoration-break

sshanzel:
> @sshanzel could apply the box-decoration-break: clone (-webkit-box-decoration-break: clone), it would look like this

Nice! Have been looking for the right property for some minutes already. Thank you!

sshanzel:
> Just need to ensure it doesn't break content for freeform posts as they re-use this.

Will verify this before merging.

sshanzel:
Freeform posts look to be fixed now 🥳 

Before:

Screenshot 2024-01-08 at 3 39 29 PM

After:
Screenshot 2024-01-08 at 3 39 20 PM

PR #2486, opened 2024-01-05
fix: adding a redirect from source changelog to squad changelog as th…

rebelchris:
To test visit this link:
```
https://daily-webapp-on5rccuj1-dailydotdev.vercel.app/sources/daily_updates
```

- Added a redirect route to go from `source` to `squad` for the changelog as both technically are valid and we have links in places already.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done

### DIFF
--- a/packages/webapp/pages/sources/[source].tsx
+++ b/packages/webapp/pages/sources/[source].tsx
@@ -184,7 +184,6 @@ export async function getStaticProps({
       revalidate: 60,
     };
   } catch (err) {
-    console.log('some err');
     if (err?.response?.errors?.[0].extensions.code === ApiError.NotFound) {
       return {
         props: {
--- a/packages/webapp/pages/sources/[source].tsx
+++ b/packages/webapp/pages/sources/[source].tsx
@@ -168,6 +168,15 @@ export async function getStaticProps({
       id: params.source,
     });
 
+    if (res.source?.type === 'squad') {
+      return {
+        redirect: {
+          destination: `/squads/${params.source}`,
+          permanent: false,
+        },
+      };
+    }
+
     return {
       props: {
         source: res.source,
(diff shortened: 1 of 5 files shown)


omBratteng (approved):

omBratteng (approved):

idoshamun (commented):
@rebelchris don't we want a more general approach in the static props?

rebelchris:
@idoshamun Guess it depends how many we plan to convert, we can convert the next one into a redirector function that can check type of source `squad|machine` and add redirect rules accordingly?

rebelchris:
@idoshamun Refactored to static paths (both ways, way neater I think)
I left them as permanent false in case we decide to swap back and forth maybe.

@capJavert Would also love to see your view on this approach :) 
(The main idea is that current sources are converted to public squads (or visa-versa))

capJavert (approved):
To me this looks good 👌 Potentially this kind of thing could be a done in middleware as well but since we have limited number of source types I would not complicate with that. 🤘

idoshamun (approved):
Perfect!

packages/webapp/pages/sources/[source].tsx
```diff
@@ -168,6 +168,15 @@ export async function getStaticProps({
       id: params.source,
     });
 
+    if (res.source?.type === 'squad') {
+      return {
+        redirect: {
+          destination: `/squads/${params.source}`,
```
Comment (capJavert) [inline packages/webapp/pages/sources/[source].tsx:174]: Looking at another pass we might want to reapply any query params since this does not do it automatically as redirect in `next.config.js`.

Comment (rebelchris) [reply packages/webapp/pages/sources/[source].tsx:174]: Very good point!
For now it should be ok as staticProps doesn't have access to them, we should move to getServerSideProps as discussed.

I think we should push now, and add a ticket to refactor it slowly.

Note: we don't really advertise these with UTM's much so shouldn't block much I don't think.
@idoshamun agreed?

Comment (idoshamun) [reply packages/webapp/pages/sources/[source].tsx:174]: I'm cool with that 😀

omBratteng (approved):

PR #2488, opened 2024-01-08
feat: feed layout v1 card layout implementation

kopancek:
## Changes

- Added copies of cards and other needed components into `/v1` folder
- Using the copied v1 cards based on type and feed v1 being on in `FeedItemComponent.tsx`
- Added proper borders and spacing around cards
- Changed the header of the card to show the type (for video and collection cards)
- Changed the display of the hot/pinned flag, based on figma designs

### Out of scope

- Please do not focus your review on inner card details. I did some work already (mistakenly), but we do have subtasks to tackle the card contents, the card header and card footer (action buttons).
- The decision to make card copies in another directory was made in the DR.

### Screenshots

Screenshot 2024-01-08 at 14 55 32
Screenshot 2024-01-08 at 14 52 09
Screenshot 2024-01-08 at 14 49 58
Screenshot 2024-01-08 at 14 49 28
Screenshot 2024-01-08 at 14 48 56
Screenshot 2024-01-08 at 14 43 56
Screenshot 2024-01-08 at 14 43 43
Screenshot 2024-01-08 at 14 43 08

- [X] MobileL (420px)
- [X] Tablet (656px)
- [X] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-2004 #done

### DIFF
--- a/packages/shared/src/components/FeedItemComponent.tsx
+++ b/packages/shared/src/components/FeedItemComponent.tsx
@@ -10,7 +10,7 @@ import { AdCard as AdCardV1 } from './cards/v1/AdCard';
 import { PlaceholderList } from './cards/PlaceholderList';
 import { PlaceholderCard } from './cards/PlaceholderCard';
 import { PlaceholderCard as PlaceholderCardV1 } from './cards/v1/PlaceholderCard';
-import { Ad, Post, PostType } from '../graphql/posts';
+import { Ad, Post, PostItem, PostType } from '../graphql/posts';
 import { LoggedUser } from '../lib/user';
 import { CommentOnData } from '../graphql/comments';
 import useTrackImpression from '../hooks/feed/useTrackImpression';
@@ -107,11 +107,23 @@ const PostTypeToTagV1: Record<PostType, FunctionComponent> = {
   [PostType.Collection]: CollectionCardV1,
 };
 
-const getPlaceHolder = (useList: boolean, isFeedLayoutV1: boolean) => {
+const getTags = (
+  isList: boolean,
+  isFeedLayoutV1: boolean,
+  postType: PostType,
+) => {
   if (isFeedLayoutV1) {
-    return PlaceholderCardV1;
+    return {
+      PostTag: PostTypeToTagV1[postType] ?? ArticlePostCardV1,
+      AdTag: AdCardV1,
+      PlaceholderTag: PlaceholderCardV1,
+    };
   }
-  return useList ? PlaceholderList : PlaceholderCard;
+  return {
+    PostTag: isList ? PostList : PostTypeToTag[postType] ?? ArticlePostCard,
+    AdTag: isList ? AdList : AdCard,
+    PlaceholderTag: isList ? PlaceholderList : PlaceholderCard,
+  };
 };
 
 export default function FeedItemComponent({
@@ -120,7 +132,7 @@ export default function FeedItemComponent({
   row,
   column,
   columns,
-  useList,
+  useList: isList,
   insaneMode,
   openNewTab,
   postMenuIndex,
@@ -142,7 +154,6 @@ export default function FeedItemComponent({
   onAdClick,
   onReadArticleClick,
 }: FeedItemComponentProps): ReactElement {
-  const AdTag = useList ? AdList : AdCard;
   const item = items[index];
   const inViewRef = useTrackImpression(
     item,
@@ -155,13 +166,17 @@ export default function FeedItemComponent({
   );
 
   const { shouldUseFeedLayoutV1 } = useFeedLayout();
-  const PlaceholderTag = getPlaceHolder(useList, shouldUseFeedLayoutV1);
+  const { PostTag, AdTag, PlaceholderTag } = getTags(
+    isList,
+    shouldUseFeedLayoutV1,
+    (item as PostItem).post?.type,
+  );
 
   let postTypeToTag = PostTypeToTag;
   let articlePostCard = ArticlePostCard;
   let EffectiveAdTag = AdTag;
 
-  if (feedName !== 'squad' && shouldUseFeedLayoutV1) {
+  if (shouldUseFeedLayoutV1) {
     postTypeToTag = PostTypeToTagV1;
     articlePostCard = ArticlePostCardV1;
     EffectiveAdTag = AdCardV1;
@@ -176,9 +191,6 @@ export default function FeedItemComponent({
         return null;
       }
 
-      const PostTag = useList
-        ? PostList
-        : postTypeToTag[item.post.type] ?? articlePostCard;
       return (
         <PostTag
           enableSourceHeader={
@@ -235,8 +247,8 @@ export default function FeedItemComponent({
                 comment({ post: item.post, content, row, column, columns })
               }
               loading={isSendingComment}
-              compactCard={!useList && insaneMode}
-              listMode={useList}
+              compactCard={!isList && insaneMode}
+              listMode={isList}
             />
           )}
         </PostTag>
--- a/packages/shared/src/components/FeedItemComponent.tsx
+++ b/packages/shared/src/components/FeedItemComponent.tsx
@@ -172,16 +172,6 @@ export default function FeedItemComponent({
     (item as PostItem).post?.type,
   );
 
-  let postTypeToTag = PostTypeToTag;
-  let articlePostCard = ArticlePostCard;
-  let EffectiveAdTag = AdTag;
-
-  if (shouldUseFeedLayoutV1) {
-    postTypeToTag = PostTypeToTagV1;
-    articlePostCard = ArticlePostCardV1;
-    EffectiveAdTag = AdCardV1;
-  }
-
   switch (item.type) {
     case 'post': {
       if (
@@ -256,7 +246,7 @@ export default function FeedItemComponent({
     }
     case 'ad':
       return (
-        <EffectiveAdTag
+        <AdTag
           ref={inViewRef}
           ad={item.ad}
           onLinkClick={(ad) => onAdClick(ad, row, column)}
--- a/packages/shared/src/components/FeedItemComponent.tsx
+++ b/packages/shared/src/components/FeedItemComponent.tsx
@@ -3,20 +3,26 @@ import dynamic from 'next/dynamic';
 import { FeedItem } from '../hooks/useFeed';
 import { PostList } from './cards/PostList';
 import { ArticlePostCard } from './cards/ArticlePostCard';
+import { ArticlePostCard as ArticlePostCardV1 } from './cards/v1/ArticlePostCard';
 import { AdList } from './cards/AdList';
 import { AdCard } from './cards/AdCard';
+import { AdCard as AdCardV1 } from './cards/v1/AdCard';
 import { PlaceholderList } from './cards/PlaceholderList';
 import { PlaceholderCard } from './cards/PlaceholderCard';
+import { PlaceholderCard as PlaceholderCardV1 } from './cards/v1/PlaceholderCard';
 import { Ad, Post, PostType } from '../graphql/posts';
 import { LoggedUser } from '../lib/user';
 import { CommentOnData } from '../graphql/comments';
 import useTrackImpression from '../hooks/feed/useTrackImpression';
 import { FeedPostClick } from '../hooks/feed/useFeedOnPostClick';
 import { SharePostCard } from './cards/SharePostCard';
+import { SharePostCard as SharePostCardV1 } from './cards/v1/SharePostCard';
 import { WelcomePostCard } from './cards/WelcomePostCard';
+import { WelcomePostCard as WelcomePostCardV1 } from './cards/v1/WelcomePostCard';
 import { Origin } from '../lib/analytics';
-import { UseVotePost } from '../hooks';
+import { UseVotePost, useFeedLayout } from '../hooks';
 import { CollectionCard } from './cards/CollectionCard';
+import { CollectionCard as CollectionCardV1 } from './cards/v1/CollectionCard';
 
 const CommentPopup = dynamic(
   () => import(/* webpackChunkName: "commentPopup" */ './cards/CommentPopup'),
@@ -92,6 +98,22 @@ const PostTypeToTag: Record<PostType, FunctionComponent> = {
   [PostType.Collection]: CollectionCard,
 };
 
+const PostTypeToTagV1: Record<PostType, FunctionComponent> = {
+  [PostType.Article]: ArticlePostCardV1,
+  [PostType.Share]: SharePostCardV1,
+  [PostType.Welcome]: WelcomePostCardV1,
+  [PostType.Freeform]: WelcomePostCardV1,
+  [PostType.VideoYouTube]: ArticlePostCardV1,
+  [PostType.Collection]: CollectionCardV1,
+};
+
+const getPlaceHolder = (useList: boolean, isFeedLayoutV1: boolean) => {
+  if (isFeedLayoutV1) {
+    return PlaceholderCardV1;
+  }
+  return useList ? PlaceholderList : PlaceholderCard;
+};
+
 export default function FeedItemComponent({
   items,
   index,
@@ -121,7 +143,6 @@ export default function FeedItemComponent({
   onReadArticleClick,
 }: FeedItemComponentProps): ReactElement {
   const AdTag = useList ? AdList : AdCard;
-  const PlaceholderTag = useList ? PlaceholderList : PlaceholderCard;
   const item = items[index];
   const inViewRef = useTrackImpression(
     item,
@@ -132,6 +153,19 @@ export default function FeedItemComponent({
     feedName,
     ranking,
   );
+
+  const { shouldUseFeedLayoutV1 } = useFeedLayout();
+  const PlaceholderTag = getPlaceHolder(useList, shouldUseFeedLayoutV1);
+
+  let postTypeToTag = PostTypeToTag;
+  let articlePostCard = ArticlePostCard;
+  let EffectiveAdTag = AdTag;
+
+  if (feedName !== 'squad' && shouldUseFeedLayoutV1) {
+    postTypeToTag = PostTypeToTagV1;
+    articlePostCard = ArticlePostCardV1;
+    EffectiveAdTag = AdCardV1;
+  }
 
   switch (item.type) {
     case 'post': {
@@ -144,7 +178,7 @@ export default function FeedItemComponent({
 
       const PostTag = useList
         ? PostList
-        : PostTypeToTag[item.post.type] ?? ArticlePostCard;
+        : postTypeToTag[item.post.type] ?? articlePostCard;
       return (
         <PostTag
           enableSourceHeader={
@@ -210,7 +244,7 @@ export default function FeedItemComponent({
     }
     case 'ad':
       return (
-        <AdTag
+        <EffectiveAdTag
           ref={inViewRef}
           ad={item.ad}
           onLinkClick={(ad) => onAdClick(ad, row, column)}
--- a/packages/shared/src/components/cards/AdAttribution.tsx
+++ b/packages/shared/src/components/cards/AdAttribution.tsx
@@ -2,11 +2,16 @@ import React, { ReactElement } from 'react';
 import classNames from 'classnames';
 import { Ad } from '../../graphql/posts';
 
-export type AdAttributionProps = {
+interface AdClassName {
+  main?: string;
+  typo?: string;
+}
+
+export interface AdAttributionProps {
   ad: Ad;
-  className?: string;
+  className?: AdClassName;
   typoClassName?: string;
-};
+}
 
 export default function AdAttribution({
   ad,
@@ -15,8 +20,8 @@ export default function AdAttribution({
 }: AdAttributionProps): ReactElement {
   const elementClass = classNames(
     'text-theme-label-quaternary no-underline',
-    typoClassName ?? 'typo-footnote',
-    className,
+    className.typo ?? 'typo-footnote',
+    className.main,
   );
   if (ad.referralLink) {
     return (
--- a/packages/shared/src/components/cards/AdAttribution.tsx
+++ b/packages/shared/src/components/cards/AdAttribution.tsx
@@ -10,13 +10,11 @@ interface AdClassName {
 export interface AdAttributionProps {
   ad: Ad;
   className?: AdClassName;
-  typoClassName?: string;
 }
 
 export default function AdAttribution({
   ad,
   className,
-  typoClassName,
 }: AdAttributionProps): ReactElement {
   const elementClass = classNames(
     'text-theme-label-quaternary no-underline',
--- a/packages/shared/src/components/cards/AdAttribution.tsx
+++ b/packages/shared/src/components/cards/AdAttribution.tsx
@@ -5,14 +5,17 @@ import { Ad } from '../../graphql/posts';
 export type AdAttributionProps = {
   ad: Ad;
   className?: string;
+  typoClassName?: string;
 };
 
 export default function AdAttribution({
   ad,
   className,
+  typoClassName,
 }: AdAttributionProps): ReactElement {
   const elementClass = classNames(
-    'text-theme-label-quaternary no-underline typo-footnote',
+    'text-theme-label-quaternary no-underline',
+    typoClassName ?? 'typo-footnote',
     className,
   );
   if (ad.referralLink) {
--- a/packages/shared/src/components/cards/v1/ActionButtons.tsx
+++ b/packages/shared/src/components/cards/v1/ActionButtons.tsx
@@ -10,29 +10,17 @@ import { SimpleTooltip } from '../../tooltips/SimpleTooltip';
 import { useFeedPreviewMode } from '../../../hooks';
 import BookmarkIcon from '../../icons/Bookmark';
 import DownvoteIcon from '../../icons/Downvote';
+import { ActionButtonsProps } from '../ActionButtons';
 
 const ShareIcon = dynamic(
   () => import(/* webpackChunkName: "shareIcon" */ '../../icons/Share'),
 );
 
-export interface ActionButtonsProps {
-  post: Post;
-  onMenuClick?: (e: React.MouseEvent) => unknown;
-  onUpvoteClick?: (post: Post) => unknown;
-  onDownvoteClick?: (post: Post) => unknown;
-  onCommentClick?: (post: Post) => unknown;
-  onBookmarkClick?: (post: Post) => unknown;
+interface ShareButtonProps {
   onShareClick?: (event: React.MouseEvent, post: Post) => unknown;
-  onReadArticleClick?: (e: React.MouseEvent) => unknown;
-  className?: string;
-  insaneMode?: boolean;
-  openNewTab?: boolean;
+  post: Post;
 }
 
-type ShareButtonProps = {
-  onShareClick?: (event: React.MouseEvent, post: Post) => unknown;
-  post: Post;
-};
 function ShareButton(props: ShareButtonProps) {
   const { onShareClick, post } = props;
   const onClickShare = (event) => onShareClick?.(event, post);
@@ -49,6 +37,10 @@ function ShareButton(props: ShareButtonProps) {
   );
 }
 
+interface ActionButtonsPropsV1 extends ActionButtonsProps {
+  onDownvoteClick?: (post: Post) => unknown;
+}
+
 export default function ActionButtons({
   post,
   onUpvoteClick,
@@ -57,7 +49,7 @@ export default function ActionButtons({
   onBookmarkClick,
   onShareClick,
   className,
-}: ActionButtonsProps): ReactElement {
+}: ActionButtonsPropsV1): ReactElement {
   const isFeedPreview = useFeedPreviewMode();
 
   if (isFeedPreview) {
--- a/packages/shared/src/components/cards/v1/ActionButtons.tsx
+++ b/packages/shared/src/components/cards/v1/ActionButtons.tsx
@@ -1,0 +1,150 @@
+import React, { ReactElement } from 'react';
+import classNames from 'classnames';
+import dynamic from 'next/dynamic';
+import { Post, UserPostVote } from '../../../graphql/posts';
+import InteractionCounter from '../../InteractionCounter';
+import UpvoteIcon from '../../icons/Upvote';
+import CommentIcon from '../../icons/Discuss';
+import { Button, ButtonColor, ButtonVariant } from '../../buttons/ButtonV2';
+import { SimpleTooltip } from '../../tooltips/SimpleTooltip';
+import { useFeedPreviewMode } from '../../../hooks';
+import BookmarkIcon from '../../icons/Bookmark';
+import DownvoteIcon from '../../icons/Downvote';
+
+const ShareIcon = dynamic(
+  () => import(/* webpackChunkName: "share" */ '../../icons/Share'),
+);
+
+export interface ActionButtonsProps {
+  post: Post;
+  onMenuClick?: (e: React.MouseEvent) => unknown;
+  onUpvoteClick?: (post: Post) => unknown;
+  onDownvoteClick?: (post: Post) => unknown;
+  onCommentClick?: (post: Post) => unknown;
+  onBookmarkClick?: (post: Post) => unknown;
+  onShareClick?: (event: React.MouseEvent, post: Post) => unknown;
+  onReadArticleClick?: (e: React.MouseEvent) => unknown;
+  className?: string;
+  insaneMode?: boolean;
+  openNewTab?: boolean;
+}
+
+type ShareButtonProps = {
+  onShareClick?: (event: React.MouseEvent, post: Post) => unknown;
+  post: Post;
+};
+function ShareButton(props: ShareButtonProps) {
+  const { onShareClick, post } = props;
+  const onClickShare = (event) => onShareClick?.(event, post);
+  return (
+    <SimpleTooltip content="Share post">
+      <Button
+        className="ml-2"
+        icon={<ShareIcon />}
+        onClick={onClickShare}
+        color={ButtonColor.Cabbage}
+        variant={ButtonVariant.Float}
+      />
+    </SimpleTooltip>
+  );
+}
+
+export default function ActionButtons({
+  post,
+  onUpvoteClick,
+  onDownvoteClick,
+  onCommentClick,
+  onBookmarkClick,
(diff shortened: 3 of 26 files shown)


packages/shared/src/components/cards/v1/RaisedLabel.tsx
```diff
@@ -0,0 +1,55 @@
...
+  className,
+}: RaisedLabelProps): ReactElement {
+  return (
+    <div
+      className={classNames(
+        'absolute -top-2 right-2 flex flex-row px-2',
+        className,
+      )}
+    >
+      <Spacer />
```
sshanzel, line 41: Can you tell us more of the purpose of this spacer?

packages/shared/src/components/cards/v1/PostMetadata.tsx
```diff
@@ -0,0 +1,68 @@
...
+          </span>
+        )}
+        {!!createdAt && showReadTime && <Separator />}
+        {!!createdAt && <time dateTime={createdAt}>{date}</time>}
+        {/* {(!!createdAt || showReadTime) && !!numUpvotes && <Separator />}
+      {!!numUpvotes && (
+        <span data-testid="numUpvotes">
+          {numUpvotes} upvote{numUpvotes > 1 ? 's' : ''}
+        </span>
+      )} */}
```
sshanzel, line 63: Can be removed? Or should be brought back?

packages/shared/src/components/cards/v1/PostCardFooter.tsx
```diff
@@ -0,0 +1,68 @@
...
+    <>
+      {/* {!showImage && post.author && (
+        <PostAuthor
+          author={post.author}
+          className={classNames(
+            'mx-4 mt-2 hidden tablet:flex laptop:hidden',
+            visibleOnGroupHover,
+          )}
+        />
+      )} */}
```
sshanzel, line 40: Can be removed? Or should be brought back?

packages/shared/src/components/cards/v1/PostCardFooter.tsx
```diff
@@ -0,0 +1,68 @@
...
+          className={classNames(
+            'absolute z-1 mt-2 flex w-full items-center rounded-t-xl bg-theme-bg-primary px-3 py-2 font-bold text-theme-label-secondary typo-callout',
+            visibleOnGroupHover,
+          )}
+        >
+          <ProfilePicture size="small" user={post.author} />
+          <span className="mx-3 flex-1 truncate">{post.author.name}</span>
+          <FeatherIcon secondary className="text-2xl text-theme-status-help" />
+        </div>
+      )} */}
```
sshanzel, line 65: Can be removed? Or should be brought back?

packages/shared/src/components/cards/v1/PostCardFooter.tsx
```diff
@@ -0,0 +1,68 @@
...
+        src={post.image}
+        fallbackSrc={cloudinary.post.imageCoverPlaceholder}
+        className={classNames(
+          'w-full object-cover',
+          className.image,
+          !isVideoType && 'mt-4',
+        )}
+        loading="lazy"
+        data-testid="postImage"
+        {...(isVideoType && { wrapperClassName: 'mt-4' })}
```
sshanzel, line 52: Might break the app if the value is false. Wdyt?

packages/shared/src/components/cards/v1/PostCardFooter.tsx
```diff
@@ -0,0 +1,68 @@
...
+  image?: string;
+}
+
+type PostCardFooterProps = {
+  insaneMode: boolean;
+  openNewTab: boolean;
+  showImage: boolean;
+  post: Post;
+  className: PostCardFooterClassName;
+};
```
sshanzel, line 17: To be inline with our guidelines, let's proceed with using interface for new ones.
```suggestion
interface PostCardFooterProps {
  insaneMode: boolean;
  openNewTab: boolean;
  showImage: boolean;
  post: Post;
  className: PostCardFooterClassName;
}
```

packages/shared/src/components/cards/v1/FeedbackCard.tsx
```diff
@@ -0,0 +1,89 @@
...
+  post,
+  onUpvoteClick,
+  onDownvoteClick,
+  showImage = true,
+}: FeedbackCardProps): ReactElement => {
+  const { dismissFeedback } = usePostFeedback({ post });
+  const feedbackCopy = useFeature(feature.cardFeedbackCopy);
+
+  return (
+    <div className="flex-1 space-y-4">
```
sshanzel, line 34: Is it necessary to have this wrapped with an extra div?

packages/shared/src/components/cards/v1/FeedItemContainer.tsx
```diff
@@ -0,0 +1,64 @@
...
+  const showTypeLabel = !!adAttribution || !!type;
+
+  return (
+    <Card
+      {...domProps}
+      data-testid="postItem"
+      ref={ref}
+      className={domProps?.className}
+    >
+      <fieldset>
```
sshanzel, line 48: Semantically speaking, isn't `fieldset` used in a form?

packages/shared/src/components/cards/v1/ArticlePostCard.tsx
```diff
@@ -0,0 +1,132 @@
...
+      {showFeedback ? (
+        <FeedbackCard
+          post={post}
+          onUpvoteClick={() => onUpvoteClick(post, Origin.FeedbackCard)}
+          onDownvoteClick={() => onDownvoteClick(post, Origin.FeedbackCard)}
+          showImage={showImage}
+        />
+      ) : (
+        <>
+          <div>
```
sshanzel, line 80: Just wanted to check whether the `div` was necessary considering there is no stylings attached to it.

sshanzel (commented):
Looks great so far! Just had a few comments but most are very minor and due to the fact that some code blocks were copy and pasted, so there are some legacy stuff that we should probably cleaning out as well (at least on the new implementations).

Can't imagine the amount of effort you put in here, and to test different types of cards must have been time consuming. Great work! 💪 🎉

packages/shared/src/components/FeedItemComponent.tsx
```diff
@@ -133,6 +154,19 @@ export default function FeedItemComponent({
     ranking,
   );
 
+  const { shouldUseFeedLayoutV1 } = useFeedLayout();
+  const PlaceholderTag = getPlaceHolder(useList, shouldUseFeedLayoutV1);
+
+  let postTypeToTag = PostTypeToTag;
+  let articlePostCard = ArticlePostCard;
+  let EffectiveAdTag = AdTag;
```
rebelchris, line 162: I'd try to avoid this `let` basically and maybe we can opt for a better mapping function that has 2 dimensions? (version/cards)

packages/shared/src/components/cards/AdAttribution.tsx
```diff
@@ -5,14 +5,17 @@ import { Ad } from '../../graphql/posts';
 export type AdAttributionProps = {
   ad: Ad;
   className?: string;
+  typoClassName?: string;
```
rebelchris, line 8: It's better to opt for `className` as object, you can see some existing implementations already or @sshanzel can support you as well.

packages/shared/src/components/cards/v1/ActionButtons.tsx
```diff
@@ -0,0 +1,150 @@
...
+import UpvoteIcon from '../../icons/Upvote';
+import CommentIcon from '../../icons/Discuss';
+import { Button, ButtonColor, ButtonVariant } from '../../buttons/ButtonV2';
+import { SimpleTooltip } from '../../tooltips/SimpleTooltip';
+import { useFeedPreviewMode } from '../../../hooks';
+import BookmarkIcon from '../../icons/Bookmark';
+import DownvoteIcon from '../../icons/Downvote';
+
+const ShareIcon = dynamic(
+  () => import(/* webpackChunkName: "share" */ '../../icons/Share'),
```
rebelchris, line 15: The chunkname should be shareIcon to avoid conflicts down the line.

packages/shared/src/components/cards/v1/PlaceholderCard.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+export const PlaceholderCard = forwardRef(function PlaceholderCard(
+  { className, showImage, ...props }: PlaceholderCardProps,
+  ref: Ref<HTMLElement>,
+): ReactElement {
+  return (
+    <Card aria-busy className={classNames(className)} {...props} ref={ref}>
+      <CardTextContainer>
+        <CardSpace className="mb-2 flex flex-row">
+          <ElementPlaceholder className="h-10 w-10 rounded-full" />
```
rebelchris, line 21: For these elements with equal w-h we can actually use `size-10` these days, it just got added to our tailwind version 🎉

packages/shared/src/components/cards/v1/RaisedLabel.tsx
```diff
@@ -0,0 +1,55 @@
...
+export enum RaisedLabelType {
+  Hot = 'Hot',
+  Pinned = 'Pinned',
+  Beta = 'Beta',
+}
+
+const typeToClassName: Record<RaisedLabelType, string> = {
+  [RaisedLabelType.Hot]: 'bg-theme-status-error',
+  [RaisedLabelType.Pinned]: 'bg-theme-bg-bun',
+  [RaisedLabelType.Beta]: 'bg-cabbage-40',
```
rebelchris, line 15: Is this the right color, feels odd not sure it's legacy.
Should be something like `bg-theme-bg-cabbage` I feel like.

packages/shared/src/components/cards/v1/RaisedLabel.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+export interface RaisedLabelProps {
+  type: RaisedLabelType;
+  description?: string;
+  className?: string | undefined;
+}
+
+const Spacer = classed(
+  'div',
+  'mt-[7px] w-2 h-1 bg-theme-bg-primary  group-hover:bg-theme-bg-secondary group-focus:bg-theme-bg-secondary group-focus:mt-1.5',
```
rebelchris, line 26: 7px can't be right, our rule of thumbs is that's it's dividable
CC: @sshanzel

packages/shared/src/components/cards/v1/TypeLabel.tsx
```diff
@@ -0,0 +1,50 @@
+import React, { ReactElement } from 'react';
+import classNames from 'classnames';
+import { PostType } from '../../../graphql/posts';
+
+type TypeLabelType = PostType | ReactElement | string;
+
+const typeToClassName: Record<PostType, string> = {
```
rebelchris, line 7: Why do we have the ones that don't have one defined still in here?

rebelchris (commented):
Just some minor comments from my side.

packages/shared/src/components/cards/v1/RaisedLabel.tsx
kopancek: If you look at the designs for hover/focus with the flag shown:
Screenshot 2024-01-08 at 16 27 28

There is a gap between the flag and the top border on the card. The spacer is there to cover the border, so that it looks the same as in the design. I could maybe do the same trick by having a px-2 and playing with a background gradient on the flag div as well, if you'd prefer that instead. This just was faster to come up with on my end.

packages/shared/src/components/cards/v1/PlaceholderCard.tsx
kopancek: Nice, will do that 👍

packages/shared/src/components/cards/v1/RaisedLabel.tsx
kopancek: Yes, I was playing with it a bit and `bg-theme-bg-cabbage` is the same. I guess this is just copy&paste from the old component.

packages/shared/src/components/cards/v1/TypeLabel.tsx
```diff
@@ -0,0 +1,50 @@
...
+export function TypeLabel({
+  type = PostType.Article,
+  className,
+}: TypeLabelProps): ReactElement {
+  // do not show tag label for these types
+  if (
+    type === PostType.Article ||
+    type === PostType.Freeform ||
+    type === PostType.Share ||
+    type === PostType.Welcome
```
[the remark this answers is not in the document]
kopancek: The code is much neater in that case, agreed. But this should be more efficient. Maybe if I create a set instead, it would be more efficient than `includes`. I know, we only have 4 elements in the array here, but still :)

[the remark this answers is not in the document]
sshanzel: Yep, long as we won't have to add a new line for each type, I would be glad to have it 🚀

packages/shared/src/components/cards/v1/FeedItemContainer.tsx
kopancek: I would tend to agree. Was following the task definition on the issue here: https://dailydotdev.atlassian.net/browse/WT-2004

packages/shared/src/components/cards/v1/RaisedLabel.tsx
kopancek: 7px is exactly right for what I want to achieve with that spacer. [See the comment here](https://github.com/dailydotdev/apps/pull/2488#discussion_r1444833110).

packages/shared/src/components/FeedItemComponent.tsx
```diff
@@ -133,6 +154,19 @@ export default function FeedItemComponent({
     ranking,
   );
 
+  const { shouldUseFeedLayoutV1 } = useFeedLayout();
+  const PlaceholderTag = getPlaceHolder(useList, shouldUseFeedLayoutV1);
```
nensidosari, line 158:
```suggestion
  const PlaceholderTag = getPlaceholder(useList, shouldUseFeedLayoutV1);
```

packages/shared/src/components/cards/v1/RaisedLabel.tsx
sshanzel: Yeah, usually, the issue with the design apps is due to the borders offsetting the value.

packages/shared/src/components/cards/v1/PostCardFooter.tsx
kopancek: Let's leave this for the implementation of the [card footer in WT-2007](https://dailydotdev.atlassian.net/browse/WT-2007).

packages/shared/src/components/FeedItemComponent.tsx
```diff
@@ -133,6 +154,19 @@ export default function FeedItemComponent({
...
+
+  let postTypeToTag = PostTypeToTag;
+  let articlePostCard = ArticlePostCard;
+  let EffectiveAdTag = AdTag;
+
+  if (feedName !== 'squad' && shouldUseFeedLayoutV1) {
```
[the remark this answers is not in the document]
kopancek: If you mean this function:
```
const checkShouldUseFeedLayoutV1 = (feedName: SharedFeedPage): boolean =>
  Object.values(SharedFeedPage)
    .filter((feedPage) => feedPage !== SharedFeedPage.Search)
    .includes(feedName) && feedName !== SharedFeedPage.Search;
```

As far as I can see, it only excludes search feed. Do you mean I extend it and also exclude the squad feed there?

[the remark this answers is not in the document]
nensidosari: shouldUseFeedLayoutV1 should be directly avoiding the squad feed

packages/shared/src/components/feeds/FeedContainer.tsx
```diff
@@ -124,7 +132,8 @@ export const FeedContainer = ({
   const numCards = currentSettings.numCards[spaciness ?? 'eco'];
   const insaneMode = !forceCardMode && listMode;
   const isList = (insaneMode && numCards > 1) || shouldUseFeedLayoutV1;
-  const feedGapPx = getFeedGapPx[gapClass(isList, spaciness)];
+  const feedGapPx =
```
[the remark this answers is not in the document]
kopancek: If we already have a function that does business logic to assign a variable, I'd rather keep all the logic in the function. But that's my preference, not sure what is the convention in the code 🤷

packages/shared/src/components/cards/v1/FeedItemContainer.tsx
sshanzel: I see. Let's probably seek opinion of the others as well @dailydotdev/web-team

packages/shared/src/components/cards/v1/PostCardFooter.tsx
sshanzel: This is a new implementation, not sure we would have to wait the other ticket for this part.

packages/shared/src/components/feeds/FeedContainer.tsx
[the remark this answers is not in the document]
sshanzel: We don't have any specific convention for this one, just thought of not touching the function and having it lesser parameters than it needs to when we can make the decision outside. But yes, I also agree that since it already handles the business logic, why not have it inside? I am cool with what you have here, I just threw some thoughts.

packages/shared/src/components/FeedItemComponent.tsx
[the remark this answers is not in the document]
kopancek: Ah, now I see. The function already filters out squad page by not even including the squads feed in the `SharedFeedPage` enum. Smart. So I can just remove the `feedName !== 'squad'` check completely.

packages/shared/src/components/cards/v1/PostCardFooter.tsx
kopancek: Same as above

packages/shared/src/components/cards/v1/PostCardFooter.tsx
kopancek: Same as [above](https://github.com/dailydotdev/apps/pull/2488#discussion_r1444859638)

packages/shared/src/components/cards/v1/ArticlePostCard.tsx
kopancek: No, it was not. Probably some leftover from when I was refactoring a bit.

packages/shared/src/components/cards/v1/FeedItemContainer.tsx
nensidosari: Fieldset can be used outside a form as well, though it's intend is to group form elements together. Is that the best practice, not sure at this moment, we shouldn't really be using it only from design perspective most probably, that's why it's a recommendation rather then a set way on how to do it, so if we feel it's a bad practice, let's design with normal divs.

nensidosari (commented):
From the images:
- let's make sure we don't use the collection tag again within the card now with the new design.

Left few comments, haven't checked the rest of the PR through, will get back to it a bit later tonight.

packages/shared/src/components/cards/v1/PostCardFooter.tsx
kopancek: I don't think I get what might break here if `isVideoType` is `false` :)

packages/shared/src/components/cards/v1/TypeLabel.tsx
kopancek: Originally I thought that we will have a style for every type and later forgot to remove the unused ones.

kopancek:
> From the images:
> 
> * let's make sure we don't use the collection tag again within the card now with the new design.

I intentionally let it be, since this task is not about dealing with card contents. And the inner collection tag should probably be handled in [WT-2005](https://dailydotdev.atlassian.net/browse/WT-2005)

nensidosari:
> I intentionally let it be, since this task is not about dealing with card contents. And the inner collection tag should probably be handled in [WT-2005](https://dailydotdev.atlassian.net/browse/WT-2005)

True, that's fine, because of all the code change it was just confusion by my side 👍

packages/shared/src/components/cards/v1/AdCard.tsx
```diff
@@ -0,0 +1,74 @@
...
+              'z-1 w-full',
+              showBlurredImage && 'absolute inset-0 m-auto',
+            )}
+            style={{ objectFit: showBlurredImage ? 'contain' : 'cover' }}
+          />
+          {showBlurredImage && (
+            <CardImage
+              alt="Ad image background"
+              src={ad.image}
+              className={classNames('-z-1 w-full')}
```
nensidosari, line 58: There used to be a `styles.blur` is it intentionally removed? Just checking to not unintentionally remove functionality

packages/shared/src/components/cards/v1/ArticlePostCard.tsx
```diff
@@ -0,0 +1,130 @@
...
+          />
+        </Link>
+      )}
+
+      {showFeedback ? (
+        <FeedbackCard
+          post={post}
+          onUpvoteClick={() => onUpvoteClick(post, Origin.FeedbackCard)}
+          onDownvoteClick={() => onDownvoteClick(post, Origin.FeedbackCard)}
+          showImage={showImage}
```
nensidosari, line 76: I know this is not part of this task, but since the changes are here, trying to understand why the change of logic on how we show the feedback card?

packages/shared/src/components/cards/v1/FeedbackCard.tsx
nensidosari: For this questions to have a chance to answer, maybe having and TODO in the code to point here would be helpful. We just need to make sure we clear those out in the end

packages/shared/src/components/cards/v1/PostCardFooter.tsx
nensidosari: If we want to leave these commented I think we better add a TODO for the person working on the task after to check, as just commented it might go unnoticed. Or we bring it back and leave it for the right card to owner to decide what needs to be done.

packages/shared/src/components/cards/v1/PostCardFooter.tsx
nensidosari: If we want to leave these commented I think we better add a TODO for the person working on the task after to check, as just commented it might go unnoticed. Or we bring it back and leave it for the right card to owner to decide what needs to be done.

packages/shared/src/components/cards/v1/PostMetadata.tsx
nensidosari: If we want to leave these commented I think we better add a TODO for the person working on the task after to check, as just commented it might go unnoticed. Or we bring it back and leave it for the right card to owner to decide what needs to be done.

packages/shared/src/components/cards/v1/RaisedLabel.tsx
nensidosari: Could be done with a wrapper div as well rather then with 2 different divs - `Spacer` set in the beginning and end

nensidosari (commented):
Looks good so far :) Few concerns and comments on my side

One additional one, I was courious to try the focus on the cards but this happened. Maybe worth checking.

Screenshot 2024-01-09 at 6 58 58 AM

packages/shared/src/components/cards/v1/FeedbackCard.tsx
sshanzel: Ah, cool. Yep, after checking it again, doesn't seem to be an extra one 🚀

packages/shared/src/components/cards/v1/PostCardFooter.tsx
sshanzel: You are spreading the result of this line, meaning if false, you will be spreading the boolean false which is not an object you can iterate similar to object literal/arrays.

packages/shared/src/components/cards/v1/RaisedLabel.tsx
kopancek: Yup. I'll probably play with the background gradient. That way we can avoid having additional DOM elements altogether (hopefully).

packages/shared/src/components/FeedItemComponent.tsx
```diff
@@ -201,16 +247,16 @@ export default function FeedItemComponent({
...
+              listMode={isList}
             />
           )}
         </PostTag>
       );
     }
     case 'ad':
       return (
-        <AdTag
+        <EffectiveAdTag
```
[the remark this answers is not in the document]
kopancek: Seems like my brain was not working properly anymore 😭

packages/shared/src/components/FeedItemComponent.tsx
```diff
@@ -133,6 +165,23 @@ export default function FeedItemComponent({
...
+    isList,
+    shouldUseFeedLayoutV1,
+    (item as PostItem).post?.type,
+  );
+
+  let postTypeToTag = PostTypeToTag;
+  let articlePostCard = ArticlePostCard;
+  let EffectiveAdTag = AdTag;
+
+  if (shouldUseFeedLayoutV1) {
```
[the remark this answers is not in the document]
kopancek: Same as above 😭

packages/shared/src/components/cards/v1/AdCard.tsx
kopancek: Not intentional. I planned on moving to tailwind and then didn't get to it. Good catch!

packages/shared/src/components/cards/v1/ArticlePostCard.tsx
kopancek: It's mostly the implementation is cleaner this way. Instead of hiding components with CSS we do not even render it, if it's not shown.

packages/shared/src/components/cards/v1/PostCardFooter.tsx
kopancek: We are doing the same thing in the original component. I think this part of the code is handled by react and not the typescript compiler / javascript runtime. Hence why it does not fail in production. It's actually a pretty common pattern that I've seen elsewhere as well 🤔

packages/shared/src/components/cards/v1/RaisedLabel.tsx
kopancek: I did away with the spacers and instead implemented it with wrapper div being absolutely positioned to `top: -1px` (or `top: -2px` on focus) and relatively positioning the child div with the flag to `top: -8px`.

kopancek:
> Looks good so far :) Few concerns and comments on my side
> 
> One additional one, I was courious to try the focus on the cards but this happened. Maybe worth checking.
> 
> Screenshot 2024-01-09 at 6 58 58 AM

Sorry I didn't catch this before. It's showing OK if you manually set focus in dev tools. I will need to take a look on how to resolve this.

nensidosari (approved):
Looks good on my side 🥳  left one small comment

packages/shared/src/components/cards/v1/ActionButtons.tsx
```diff
@@ -0,0 +1,150 @@
...
+import { SimpleTooltip } from '../../tooltips/SimpleTooltip';
+import { useFeedPreviewMode } from '../../../hooks';
+import BookmarkIcon from '../../icons/Bookmark';
+import DownvoteIcon from '../../icons/Downvote';
+
+const ShareIcon = dynamic(
+  () => import(/* webpackChunkName: "shareIcon" */ '../../icons/Share'),
+);
+
+export interface ActionButtonsProps {
```
sshanzel, line 18: I just realized, that the only difference here would be the layout, maybe we should import the same type used on the other ones to not bloat the amount of types/interfaces we're using. Considering it is a bit tedious sometimes to make small changes.

packages/shared/src/components/cards/v1/ActionButtons.tsx
```diff
@@ -0,0 +1,150 @@
...
+  onReadArticleClick?: (e: React.MouseEvent) => unknown;
+  className?: string;
+  insaneMode?: boolean;
+  openNewTab?: boolean;
+}
+
+type ShareButtonProps = {
+  onShareClick?: (event: React.MouseEvent, post: Post) => unknown;
+  post: Post;
+};
```
sshanzel, line 35: Just converting the new ones.
```suggestion
interface ShareButtonProps {
  onShareClick?: (event: React.MouseEvent, post: Post) => unknown;
  post: Post;
}

```

packages/shared/src/components/cards/v1/AdCard.tsx
```diff
@@ -0,0 +1,74 @@
...
+      linkProps={getLinkProps(ad, onLinkClick)}
+    >
+      <CardTextContainer>
+        <CardTitle className="line-clamp-4 font-bold typo-title3">
+          {ad.description}
+        </CardTitle>
+      </CardTextContainer>
+      <CardSpace />
+      {showImage && (
+        <div className="relative overflow-hidden rounded-xl">
```
sshanzel, line 44:
```suggestion
        <div className="relative overflow-hidden rounded-12">
```

packages/shared/src/components/cards/v1/FeedItemContainer.tsx
```diff
@@ -0,0 +1,90 @@
...
+  return (
+    <Card
+      {...domProps}
+      data-testid="postItem"
+      ref={ref}
+      className={classNames(domProps?.className, focus && 'bg-theme-float')}
+    >
+      {linkProps && (
+        <ConditionalWrapper
+          condition={!linkProps.href.startsWith('https://')}
```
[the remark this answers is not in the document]
sshanzel: Maybe we can use our existing function of `isValidHttpUrl` if you are trying to ensure only valid URLs to be passed.

packages/shared/src/components/cards/v1/RaisedLabel.tsx
```diff
@@ -0,0 +1,51 @@
...
+  type = RaisedLabelType.Hot,
+  description,
+  className,
+  focus = false,
+}: RaisedLabelProps): ReactElement {
+  return (
+    <div
+      className={classNames(
+        'absolute right-2 flex flex-row px-2 group-hover:bg-theme-bg-secondary group-focus:-top-0.5 group-focus:bg-theme-bg-secondary',
+        !focus && '-top-[1px] bg-theme-bg-primary',
```
sshanzel, line 34: This should be available.
```suggestion
        !focus && '-top-px bg-theme-bg-primary',
```

packages/shared/src/components/cards/v1/RaisedLabel.tsx
```diff
@@ -0,0 +1,51 @@
...
+        focus && '-top-0.5 bg-theme-bg-secondary',
+        className,
+      )}
+    >
+      <SimpleTooltip content={description}>
+        <div
+          className={classNames(
+            'relative -top-2 h-4 rounded-sm px-2 font-bold uppercase text-white typo-caption2',
```
sshanzel, line 42: Just to be inline with what we have.
```suggestion
            'relative -top-2 h-4 rounded-2 px-2 font-bold uppercase text-white typo-caption2',
```

packages/shared/src/components/cards/v1/PostCardFooter.tsx
sshanzel: Oh, if that is the case, feel free to move on. I learned something today 🚀

sshanzel (approved):
Looks great to me overall! 💪 

I just had a few non-blocking comments, most of it have suggestions already, and some were just questions.

packages/shared/src/components/cards/v1/RaisedLabel.tsx
rebelchris: I think we probably can move `radius` to overwrite not extend to be honest, this will avoid this in the future.
I'll raise it in web-guild.

packages/shared/src/components/cards/v1/FeedItemContainer.tsx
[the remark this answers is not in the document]
rebelchris: Yeah this conditional render feels a bit odd, also could it not be handled inside `CardLink`?

packages/shared/src/components/cards/v1/TypeLabel.tsx
```diff
@@ -0,0 +1,51 @@
...
+}: TypeLabelProps): ReactElement {
+  // do not show tag label for excluded types
+  if (excludedTypes.has(type as PostType)) {
+    return null;
+  }
+
+  return (
+    <legend
+      className={classNames(
+        'rounded bg-theme-bg-primary px-2 font-bold capitalize typo-caption1 group-hover:bg-theme-bg-secondary group-focus:bg-theme-bg-secondary',
```
rebelchris, line 43: The 4px rounded is ok here?

rebelchris (approved):
Besides Lee's comments nothing blocking from my side, so approving already.
Do note to adjust some of the comments there.

packages/shared/src/components/cards/v1/FeedItemContainer.tsx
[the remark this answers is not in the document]
kopancek: This condition is there so that we do not render a nextjs `<Link>` for links that point outside of our app. E.g. if it's a global link, we just want the `<a>` element there. It's a bit naive, but don't know if we have a better solution for that somewhere already.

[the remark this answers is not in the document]
sshanzel: I see, there are places where links are with http or https for internal pages, this is because it was built to support the extension. Since this is for the feed card's wrapper, it will always be pointing to internal links when clicking the card.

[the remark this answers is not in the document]
kopancek: I just checked and nextjs already has this check as part of the `<Link` component, so I'll remove the conditional rendering. Should make it simpler hopefully :)

packages/shared/src/components/cards/v1/RaisedLabel.tsx
kopancek: Oh, that's how. Thanks, today I learned.

packages/shared/src/components/cards/v1/RaisedLabel.tsx
kopancek: should be 4px, so changing to `rounded`

packages/shared/src/components/cards/v1/TypeLabel.tsx
kopancek: Yes, same as in Figma.

PR #2491, opened 2024-01-09
feat: use event listener hook and refactor usages

capJavert:
- Introduce new `useEventListener` hook that supports all different event handlers and iterations of `addEventListener`
- Refactor all usages of `useWindowEvents` to `useEventListener`
- Move `validateKey` functionality to handlers for better visibility vs hiding it in hook implementation (also they are used in less places
- Available on [preview](https://preview.app.daily.dev/)

TODO:
- [x] end to end test places where hooks were used to assure everything is working with `data.eventKey` handling
- [x] tested social login/registration
- [x] tested email login and registration
- [x] tested push notification enable
- [x] tested interactive popup close from outside click (global click handler)
- [x] tested companion login data refetch
- [x] tested popstate reload 
- [x] tested logout 

AS-25 #done


packages/shared/src/hooks/useEventListener.ts
```diff
@@ -0,0 +1,158 @@
...
+export type TUseEventListenerOptions =
+  | boolean
+  | AddEventListenerOptions
+  | undefined;
+
+export interface MessageEventData {
+  eventKey?: string;
+}
+
+const useEventListener = <
```
capJavert, line 95: Whole hook and tests are mostly the same logic from [here](https://github.com/kickassCoderz/kickass-toolkit/blob/16d2b6264c7dd6b09b2412cc41ea9df0b7815f49/packages/react/src/hooks/useEventListener/useEventListener.ts), could not use the package since we are still on react@17.x so copied everything for now.

packages/shared/src/components/auth/AuthOptions.tsx
```diff
@@ -239,70 +238,67 @@ function AuthOptions({
...
-  useWindowEvents<SocialRegistrationFlow>(
-    'message',
-    AuthEvent.SocialRegistration,
-    async (e) => {
-      if (e.data?.flow) {
-        const connected = await getKratosFlow(
-          AuthFlow.Registration,
-          e.data.flow,
-        );
+  useEventListener(globalThis, 'message', async (e) => {
```
rebelchris, line 241: This one will require some manual/preview verification before we launch it.
Mostly on both email registration and social, just to exclude edge-cases.

packages/shared/src/hooks/useEventListener.spec.ts
```diff
@@ -0,0 +1,71 @@
...
+    expect(removeSpy).toHaveBeenCalledTimes(0);
+
+    rerender();
+
+    expect(addSpy).toHaveBeenCalledTimes(1);
+    expect(removeSpy).toHaveBeenCalledTimes(0);
+
+    unmount();
+
+    expect(addSpy).toHaveBeenCalledTimes(1);
```
rebelchris, line 68: should this be called on unmount?

packages/webapp/pages/posts/[id]/index.tsx
```diff
@@ -95,14 +92,10 @@ const PostPage = ({ id, initialData }: Props): ReactElement => {
...
-  useWindowEvents(
-    'popstate',
-    CHECK_POPSTATE,
-    // @NOTE see https://dailydotdev.atlassian.net/l/cp/dK9h1zoM
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-    useCallback(() => router.reload(), []),
-    { validateKey: false },
-  );
+
+  useEventListener(globalThis, 'popstate', () => {
```
Comment (rebelchris) [inline packages/webapp/pages/posts/[id]/index.tsx:96]: what did we use the validateKey and callback for again?

rebelchris (approved):
Nothing to add as hook existed on your side.
we definitely need to test extensively and the critical flows before/after merging

packages/shared/src/components/auth/AuthOptions.tsx
```diff
@@ -239,70 +238,67 @@ function AuthOptions({
...
-      if (!bootResponse.data.user || !('email' in bootResponse.data.user)) {
-        trackEvent({
-          event_name: AuthEventNames.SubmitSignUpFormError,
-          extra: JSON.stringify({
-            error: 'Could not find email on social registration',
-          }),
-        });
-        return displayToast(labels.auth.error.generic);
+      if (
+        [4010002, 4010003, 4000007].includes(connected?.ui?.messages?.[0]?.id)
```
kopancek, line 261: What do these magic numbers mean? Should we add const variable with more meaningful names?

packages/shared/src/components/auth/AuthOptions.tsx
```diff
@@ -239,70 +238,67 @@ function AuthOptions({
...
-    'message',
-    AuthEvent.SocialRegistration,
-    async (e) => {
-      if (e.data?.flow) {
-        const connected = await getKratosFlow(
-          AuthFlow.Registration,
-          e.data.flow,
-        );
+  useEventListener(globalThis, 'message', async (e) => {
+    if (e.data?.eventKey !== AuthEvent.SocialRegistration) {
```
kopancek, line 242: Can this condition be added directly to the `useEventListener`? E.g. by passing another parameter to the `useEventListener` hook. I see that we have the same behavior everywhere where the hook is used (returning undefined), just the condition is different...

packages/shared/src/hooks/useAuthErrors.ts
```diff
@@ -1,10 +1,15 @@
...
 export function useAuthErrors(): void {
   const { displayToast } = useToastNotification();
-  useWindowEvents<ErrorEvent>('message', AuthEvent.Error, async (e) => {
+
+  useEventListener(globalThis, 'message', async (e) => {
+    if (e.data?.eventKey !== AuthEvent.Error) {
+      return;
+    }
+
     if (!e.data?.id) {
```
kopancek, line 13: Since the return value is the same, these conditions can be merged, e.g. 
```suggestion
    if (e.data?.eventKey !== AuthEvent.Error || !e.data?.id) {
```

kopancek (commented):
I don't really have context here. Looks good to me, just spotted some potential places for improving the code or making it easier to understand.

packages/shared/src/components/auth/AuthOptions.tsx
sshanzel: These are error codes from Kratos. We should have an enum to hold these values `KRATOS_ERROR`.

packages/shared/src/components/auth/AuthOptions.tsx
capJavert: This was what the old hook did and was hidden with `validateKey`. I think this (while adding some code) at least keeps the condition visible when looking at the related code. And we don't really use it on much of the places.

packages/shared/src/hooks/useAuthErrors.ts
capJavert: Would it not be less readable then?

packages/shared/src/components/auth/AuthOptions.tsx
capJavert: Yeah I could add the enum here 👌 @sshanzel do you know from where are those codes taken/documented?

packages/webapp/pages/posts/[id]/index.tsx
Comment (capJavert) [reply packages/webapp/pages/posts/[id]/index.tsx:96]: Here it was not used because `validateKey: false`, on other places I moved checks to the callbacks. From what I seen the kratos `message` events had `data.eventKey` to differentiate so it was a guard to execute logic only for specific event.

packages/shared/src/components/auth/AuthOptions.tsx
capJavert: I could add the new hook like `useKratosMessageListener` that would encapsulate that logic and subscribe to specific key?

packages/shared/src/hooks/useEventListener.spec.ts
capJavert: It checks that listener was not added after unmount eg. it was still called only once during mount.

packages/shared/src/components/auth/AuthOptions.tsx
capJavert: Yes!

packages/shared/src/hooks/useEventListener.spec.ts
rebelchris: I see :) perfect in that case

packages/webapp/pages/posts/[id]/index.tsx
Comment (rebelchris) [reply packages/webapp/pages/posts/[id]/index.tsx:96]: I see, ok no worries in that case.

packages/shared/src/components/auth/AuthOptions.tsx
sshanzel: Will this be attached to the window with just this?

packages/shared/src/components/auth/AuthOptions.tsx
sshanzel: Should be available through this document: sendAudioData

packages/shared/src/components/auth/AuthOptions.tsx
capJavert: Yes, and using `globalThis` nomenclature instead of window as we do recently.

packages/shared/src/components/auth/AuthOptions.tsx
capJavert: Added in 52ab383713c09115c39c2380ed2304d8ba290c11

packages/shared/src/components/auth/AuthOptions.tsx
capJavert: I tested the cases above (updated description).

packages/shared/src/hooks/useAuthErrors.ts
kopancek: I don't have a strong preference. To me the readability is the same. Having 2 conditions below each other when it can be defined with OR operator, I always prefer the OR operator.

If readability is a concern, you can split the condition into two lines, e.g.:
```suggestion
    if (e.data?.eventKey !== AuthEvent.Error ||
        !e.data?.id)
      return;
    }
```

kopancek: Also, do not feel obliged. This is not blocking the merge from my point of view, just to be clear :)

capJavert: Above would not be allowed by our linter but basically not about code readability, its about logic readability. 

Multiple `||` and `&&` etc. quickly become hard to parse logically when looking the code. Throw in a negation `!` and you are looking at something that becomes a monster really fast. 

Early return pattern like above solves it in 99% of the time while keeping conditions nice and lean. Would love to talk about it more at some other time 🙏

PR #2492, opened 2024-01-10
fix: post options menu state

sshanzel:
## Changes
- When viewing a post page, and you try to follow/unfollow certain tags/sources/settings, the values do not update when you do any action.
- In this PR, we now consider the current state of the user's preference to reflect better the actual status.
- Introduced a new settings function to update the value based on the desired state. This was due to a bug encountered were the value is not getting updated after clicking undo (because of the memoized function).

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-2057 #done

### DIFF
--- a/packages/shared/src/hooks/feed/useAdvancedSettings.ts
+++ b/packages/shared/src/hooks/feed/useAdvancedSettings.ts
@@ -8,6 +8,7 @@ import useMutateFilters from '../useMutateFilters';
 interface UseAdvancedSettings {
   selectedSettings: Record<string, boolean>;
   onToggleSettings(id: number, state: boolean): void;
+  onUpdateSettings(id: number, state: boolean): void;
 }
 
 export const useAdvancedSettings = (
@@ -29,6 +30,24 @@ export const useAdvancedSettings = (
     [feedSettings?.advancedSettings],
   );
 
+  const onUpdateSettings = useCallback(
+    (id: number, enabled: boolean) => {
+      trackEvent({
+        event_name: `toggle ${enabled ? 'on' : 'off'}`,
+        target_type: 'advanced setting',
+        target_id: id.toString(),
+        extra: JSON.stringify({ origin: 'advanced settings filter' }),
+      });
+
+      return updateAdvancedSettings({
+        advancedSettings: [{ id, enabled }],
+      });
+    },
+    [trackEvent, updateAdvancedSettings],
+  );
+
+  // We still need to support this function as this is used for experiments
+  // and currently used in simpler implementations of updating the settings
   const onToggleSettings = useCallback(
     (id: number, defaultEnabledState: boolean) => {
       if (alerts?.filter && user) {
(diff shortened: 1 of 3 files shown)


packages/shared/src/hooks/useFeedSettings.ts
```diff
@@ -53,14 +54,29 @@ export default function useFeedSettings({
   );
 
   const { tagsCategories, feedSettings, advancedSettings } = feedQuery;
+  const checkEnabledSettings = useCallback(
+    (value: number) => {
```
rebelchris, line 58: Value is the advancedSettingsId?
I would maybe call it that or `id` just so it's clear.

packages/shared/src/hooks/useFeedSettings.ts
```diff
@@ -53,14 +54,29 @@ export default function useFeedSettings({
   );
 
   const { tagsCategories, feedSettings, advancedSettings } = feedQuery;
+  const checkEnabledSettings = useCallback(
```
rebelchris, line 57: Also the naming contradict it searches one value.
checkSettingsEnabledState or something might be better? (WDYT)

packages/shared/src/hooks/feed/useAdvancedSettings.ts
```diff
@@ -29,34 +30,36 @@ export const useAdvancedSettings = (
...
-      selectedSettings,
-      trackEvent,
-      updateAdvancedSettings,
-      updateAlerts,
-      user,
-    ],
+    [trackEvent, updateAdvancedSettings],
+  );
+
+  // We still need to support this function as this is used for experiments
```
rebelchris, line 49: We can't migrate both into one?

rebelchris (commented):
Nice work, nothing blocking really, but let's wait for 2 reviews :)

packages/shared/src/hooks/useFeedSettings.ts
sshanzel: Right, I was reserving the `id` on the destructure part, but we can do that.

packages/shared/src/hooks/feed/useAdvancedSettings.ts
sshanzel: We can, probably. Just not sure if we should do it as we've had two experiments on separate occasions on this part. Now that we have new advanced settings, we might again introduce the same experiment. Wdyt?

packages/shared/src/hooks/useFeedSettings.ts
sshanzel: Makes sense. Let's update it.

packages/shared/src/hooks/feed/useAdvancedSettings.ts
rebelchris: Hmm yeah let's not complicate it for now :) 
We can adjust later as well.

packages/shared/src/hooks/useFeedSettings.ts
```diff
@@ -53,14 +54,31 @@ export default function useFeedSettings({
   );
 
   const { tagsCategories, feedSettings, advancedSettings } = feedQuery;
+  const checkSettingsEnabledState = useCallback(
+    (idParam: number) => {
+      const advancedSetting = advancedSettings?.find(
+        ({ id }) => id === idParam,
+      );
+
+      if (!advancedSettings) {
```
kopancek, line 63: I think this should be 
```suggestion
      if (!advancedSetting) {
```

kopancek (changes requested):

packages/shared/src/hooks/useFeedSettings.ts
sshanzel: Oh, good catch! 🚀

kopancek (approved):
Nothing more to add, looks good to me, great job! 👍 :shipit:

packages/shared/src/hooks/useFeedSettings.ts
sshanzel: Maybe a good sign that I should rename the variable name entirely to void confusion.

PR #2493, opened 2024-01-11
refactor: post options menu removing bookmark

sshanzel:
## Changes
- Bookmark on card should be based on the experiment not on the feed layout.
- For the post options menu, it should be the same.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

MI-20 #done

### DIFF
--- a/packages/shared/src/components/PostOptionsMenu.tsx
+++ b/packages/shared/src/components/PostOptionsMenu.tsx
@@ -17,7 +17,7 @@ import useTagAndSource from '../hooks/useTagAndSource';
 import AnalyticsContext from '../contexts/AnalyticsContext';
 import { postAnalyticsEvent } from '../lib/feed';
 import { MenuIcon } from './MenuIcon';
-import { ToastSubject, useToastNotification } from '../hooks';
+import { ToastSubject, useFeedLayout, useToastNotification } from '../hooks';
 import {
   AllFeedPages,
   generateQueryKey,
@@ -289,6 +289,7 @@ export default function PostOptionsMenu({
     },
   ];
 
+  const { shouldUseFeedLayoutV1 } = useFeedLayout();
   const bookmarkOnCard = useFeature(feature.bookmarkOnCard);
 
   if (!bookmarkOnCard) {
@@ -305,8 +306,8 @@ export default function PostOptionsMenu({
     });
   }
 
-  postOptions.push(
-    {
+  if (shouldUseFeedLayoutV1) {
+    postOptions.push({
       icon: (
         <MenuIcon
           className={classNames(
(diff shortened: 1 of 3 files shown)


packages/shared/src/components/PostOptionsMenu.tsx
```diff
@@ -285,7 +287,12 @@ export default function PostOptionsMenu({
       label: 'Hide',
       action: onHidePost,
     },
-    {
+  ];
+
+  const bookmarkOnCard = useFeature(feature.bookmarkOnCard);
```
rebelchris, line 292: We have this currently in the post option menu dynamically?
Cant recall if we just always show it or based on experiment

sshanzel: From the ticket Nensi created, she [mentioned](https://dailydotdev.slack.com/archives/C061T644XV3/p1700551627401439) it to be dynamically available.

nensidosari: Yup, this should be dynamically available also in the share post options menu

packages/shared/src/components/PostOptionsMenu.tsx
```diff
@@ -295,8 +303,11 @@ export default function PostOptionsMenu({
       ),
       label: `${post?.bookmarked ? 'Remove from' : 'Save to'} bookmarks`,
       action: onToggleBookmark,
-    },
-    {
+    });
+  }
+
+  if (shouldUseFeedLayoutV1) {
```
nensidosari, line 309: shouldn't this be `!shouldUseFeedLayoutV1` ?

packages/shared/src/components/PostOptionsMenu.tsx
sshanzel: Cool. Let me apply it on that end too.

packages/shared/src/components/PostOptionsMenu.tsx
sshanzel: Oh, right. Will update quickly.

packages/shared/src/components/PostOptionsMenu.tsx
```diff
@@ -109,24 +100,6 @@ export default function PostOptionsMenu({
     shouldInvalidateQueries: false,
   });
 
-  const { toggleBookmark } = useBookmarkPost({
```
nensidosari, line 112: Can we make this at the moment removed if in feed layout experiment just like we're doing for downvote?

packages/shared/src/components/PostOptionsMenu.tsx
```diff
@@ -285,18 +258,12 @@ export default function PostOptionsMenu({
       label: 'Hide',
       action: onHidePost,
     },
-    {
-      icon: (
```
nensidosari, line 289: Same as above here

packages/shared/src/components/ShareOptionsMenu.tsx
```diff
@@ -59,17 +56,6 @@ export default function ShareOptionsMenu({
       text: 'Share post via...',
       action: () => onShare(post),
     },
-    {
```
nensidosari, line 62: Same as above

nensidosari (commented):
Looks great, just one small ask 🙏 We want these removals to be behind the feed layout experiment, as at the moment we don't know what the decision is going to be from other experiments.

packages/shared/src/components/PostOptionsMenu.tsx
sshanzel: The bookmark is going to be regardless of feed layout. Either control or v1 will have the bookmark on card once it wins.

sshanzel: I thought we were under the assumption that the bookmark on card wins? That was already what we had before.

sshanzel: Ah, right. We only do that assumption on v1.

nensidosari (approved):
Looks great, we'll check those conditions once the bookmark experiment is concluded. 🥳 
Apologies for the confusion, having multiple experiments introducing the same thing is confusing.

PR #2494, opened 2024-01-12
fix: loading more posts on tall screens

omBratteng:
## Changes

Fixes the edge case of being unable to load more in feed when height of screen is too large.

It forces removal of the infinite scroll trigger while loading, so that the `inView` state gets reset.

Deployed to https://preview3.app.daily.dev/

WT-2087


packages/shared/src/hooks/feed/useFeedInfiniteScroll.ts
```diff
@@ -9,7 +9,7 @@ export interface UseFeedInfiniteScrollProps {
 
 export const InfiniteScrollScreenOffset = classed(
   'div',
-  'absolute bottom-screen desktopL:bottom-screen-20 left-0 w-px h-px opacity-0 pointer-events-none',
+  'bottom-screen desktopL:bottom-screen-20 left-0 w-px h-px opacity-0 pointer-events-none',
```
kopancek, line 12: Was removal of the `absolute` class here intentional? If yes, why do we still need `left-0`?

kopancek (commented):
Looks mostly good to me, just have one question.

packages/shared/src/hooks/feed/useFeedInfiniteScroll.ts
omBratteng: Yeah, it was intentional.
Probably don't need `left-0`. Will double check.

omBratteng:
```suggestion
  'bottom-screen desktopL:bottom-screen-20 w-px h-px opacity-0 pointer-events-none',
```

rebelchris: To add to that, without absolute `bottom-screen` also doesn't get added, so might want to clean it up.
Screenshot 2024-01-15 at 13 40 18

rebelchris (approved):
One minor addition to Milan's comment, non blocking for me.

packages/shared/src/hooks/feed/useFeedInfiniteScroll.ts
omBratteng: 🙈

packages/shared/src/hooks/feed/useFeedInfiniteScroll.ts
```diff
@@ -9,7 +9,7 @@ export interface UseFeedInfiniteScrollProps {
 
 export const InfiniteScrollScreenOffset = classed(
   'div',
-  'absolute bottom-screen desktopL:bottom-screen-20 left-0 w-px h-px opacity-0 pointer-events-none',
+  'bottom-screen desktopL:bottom-screen-20 w-px h-px opacity-0 pointer-events-none',
```
omBratteng, line 12:
```suggestion
  'w-px h-px opacity-0 pointer-events-none',
```

capJavert (approved):

PR #2495, opened 2024-01-12
fix: header button back to plain element

sshanzel:
## Changes
- As mentioned in the Slack, we are moving back to using a plain `button` element rather than our `Button` element. However, it can be argued that can probably just override the x-padding.

Preview:
Screenshot 2024-01-12 at 12 55 12 PM

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

MI-1 #done

### DIFF
--- a/packages/shared/src/components/referral/SearchReferralButton.tsx
+++ b/packages/shared/src/components/referral/SearchReferralButton.tsx
@@ -9,7 +9,6 @@ import { KeyReferralIcon } from '../icons';
 import { IconSize } from '../Icon';
 import { useAnalyticsContext } from '../../contexts/AnalyticsContext';
 import { AnalyticsEvent, TargetId, TargetType } from '../../lib/analytics';
-import { Button, ButtonVariant } from '../buttons/ButtonV2';
 
 interface SearchReferralButtonProps {
   className?: string;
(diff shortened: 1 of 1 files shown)


packages/shared/src/components/referral/SearchReferralButton.tsx
```diff
@@ -40,14 +39,16 @@ export function SearchReferralButton({
...
+      className={classNames(
+        'flex h-10 flex-row items-center rounded-12 bg-theme-overlay-from px-3 font-bold text-theme-label-tertiary typo-callout hover:text-theme-label-primary',
+        className,
+      )}
       onClick={handleClick}
     >
       {availableCount}
       <KeyReferralIcon size={IconSize.Medium} className="-mr-1 ml-1 mt-1" />
-    </Button>
+    </button>
```
kopancek, line 52: I would honestly prefer if we force the margins instead. This way if we change the button styling in the future, this will need to be adjusted as well. And I doubt we will remember, until we see again that it looks different...

kopancek (approved):
Approving to unblock, but would prefer the other solution - forcing the margin. It's also smaller code change :)

packages/shared/src/components/referral/SearchReferralButton.tsx
sshanzel: Do you mean the padding? Margins left and right look to be okay. I agree with your point, though I think this is a special case where we only have one of this kind of button. So I guess, we can move forward with this for now.

kopancek: Yup, sorry, padding is what I meant. Next time I'll say spacing :D

kopancek: And as I said above, feel free to merge this. It's just my preference for future-proofness. Which I'm not sure if it's that big of a deal, since it's just visual change.

PR #2496, opened 2024-01-12
fix: unwanted right border on safari

sshanzel:
## Changes
- The border seems to be intended for laptop, but since the laptop pseudo was applied on the color instead of the border width, there is a primary border but only on safari.
- Moved the borders classes in a single place to avoid confusion on where things are coming from.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

MI-64 #done

### DIFF
--- a/packages/shared/src/components/post/PostContentContainer.tsx
+++ b/packages/shared/src/components/post/PostContentContainer.tsx
@@ -11,7 +11,7 @@ const BodyContainer = classed('div', 'flex flex-col w-full pb-6');
 
 const PageBodyContainer = classed(
   BodyContainer,
-  'm-auto w-full laptop:border-l laptop:border-theme-divider-tertiary',
+  'm-auto w-full laptop:border-l laptop:border-r laptop:border-theme-divider-tertiary',
 );
 
 function PostContentContainer({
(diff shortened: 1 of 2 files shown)


packages/shared/src/components/post/PostContentContainer.tsx
```diff
@@ -11,7 +11,7 @@ const BodyContainer = classed('div', 'flex flex-col w-full pb-6');
 
 const PageBodyContainer = classed(
   BodyContainer,
-  'm-auto w-full laptop:border-l laptop:border-theme-divider-tertiary',
+  'm-auto w-full laptop:border-l laptop:border-r laptop:border-theme-divider-tertiary',
```
kopancek, line 14: I think we can simplify to `border-x` here:

```suggestion
  'm-auto w-full laptop:border-x laptop:border-theme-divider-tertiary',
```

kopancek (approved):
Approving, but think we can leverage `border-x` for borders on both sides.

packages/shared/src/components/post/PostContentContainer.tsx
sshanzel: Oh, didn't know about that. Thanks! 🚀

PR #2499, opened 2024-01-12
feat: reading streak popup

sshanzel:
Note: as mentioned in the ticket, we won't be pulling any live data for now. This is just the popup component.

## Changes
- Reading streaks experiment.
- Introduced a type for streak whether it is:
  - Completed (filled)
  - Pending (broken border lines)
  - Freeze (disabled state)
  - Upcoming (transparent background)
- Reading streak popup.
- Instead of having a margin-right on every element, I've decided to use gap to make spacing easier to manage.
- New icons
  - Reading streak.
  - Triangle arrow.

Preview:
Screenshot 2024-01-12 at 11 03 19 PM

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

MI-11 #done


packages/shared/src/components/layout/MainLayoutHeader.tsx
```diff
@@ -77,23 +86,44 @@ function MainLayoutHeader({
...
     return (
       <>
-        <CreatePostButton
-          className={classNames(
-            'mr-0 laptop:mr-3',
-            optOutWeeklyGoal ? 'tablet:mr-0' : 'tablet:mr-3',
-          )}
-        />
+        <CreatePostButton />
+        {user && readingStreak === ReadingStreaksExperiment.V1 && (
```
rebelchris, line 95: For now this is fine, but it will get a custom hook I guess?

packages/shared/src/components/layout/MainLayoutHeader.tsx
```diff
@@ -77,23 +86,44 @@ function MainLayoutHeader({
...
+            visible={shouldShowStreaks}
+            container={{
+              paddingClassName: 'p-4',
+              bgClassName: 'bg-theme-bg-tertiary',
+              textClassName: 'text-theme-label-primary typo-callout',
+              className: 'border border-theme-divider-tertiary',
+            }}
+            content={<ReadingStreakPopup />}
+          >
+            <Button
```
rebelchris, line 108: Should this not be part of the `ReadingStreakPopup`?

packages/shared/src/components/streak/popup/DayStreak.tsx
```diff
@@ -0,0 +1,62 @@
...
+interface DayStreakProps {
+  streak: Streak;
+  day: number;
+  shouldShowArrow?: boolean;
+}
+
+const dayInitial = ['S', 'M', 'T', 'W', 'T', 'F', 'S'];
+const Circle = classed(
+  'div',
+  'w-7 h-7 rounded-full border border-theme-divider-tertiary',
```
rebelchris, line 23:
```suggestion
  'size-7 rounded-full border border-theme-divider-tertiary',
```

packages/shared/src/components/streak/popup/DayStreak.tsx
```diff
@@ -0,0 +1,62 @@
...
+        <ReadingStreakIcon
+          secondary={streak === Streak.Completed}
+          size={IconSize.Medium}
+        />
+      );
+    }
+
+    return (
+      <Circle
+        className={streak === Streak.Freeze && 'bg-theme-label-disabled'}
```
rebelchris, line 43: Technically you could have completed a freeze right?
(Might be an edge-case we need to tackle in development)
CC: @DragosIorgulescu

rebelchris (changes requested):
Nice, looks great!
Have some minor comments 🙏

packages/shared/src/components/streak/popup/DayStreak.tsx
sshanzel: Right. We have this now. Thanks!

packages/shared/src/components/streak/popup/DayStreak.tsx
sshanzel: Yes, that should be possible and should be handled by the consuming component (ReadingStreakPopup). This kind of dumb component where will only act based on props, and not necessarily contain the logic to decide which day is completed or not.

packages/shared/src/components/layout/MainLayoutHeader.tsx
sshanzel: This is the CTA. Maybe introduce a separate component `ReadingStreakButton` to contain the configurations and not make this header component messy. Wdyt?

packages/shared/src/components/layout/MainLayoutHeader.tsx
sshanzel: Yeah, if we ever need additional logic to determine the result and keep everything inline.

packages/shared/src/components/layout/MainLayoutHeader.tsx
rebelchris: Hmm why are they wrapped then, feels a bit weird somehow?
Expected to have the button, and "modal/tippy" seperate from each other

sshanzel: Ah, usually on modal that's the case. Though for tippy, this has been the case, to wrap the component so we can have a reference on where we are going to display the tooltip.

rebelchris (approved):
Nothing blocking from my side, please await epic owner

packages/shared/src/components/layout/MainLayoutHeader.tsx
DragosIorgulescu: should we bite the bullet and add a hook now? We will need to make this check in other places as well, so might as well have a hook we can go to make changes if we ever need to in the future

packages/shared/src/components/layout/MainLayoutHeader.tsx
```diff
@@ -77,23 +86,44 @@ function MainLayoutHeader({
...
       <>
-        <CreatePostButton
-          className={classNames(
-            'mr-0 laptop:mr-3',
-            optOutWeeklyGoal ? 'tablet:mr-0' : 'tablet:mr-3',
-          )}
-        />
+        <CreatePostButton />
+        {user && readingStreak === ReadingStreaksExperiment.V1 && (
+          <SimpleTooltip
```
DragosIorgulescu, line 96: I would go with adding this logic to a separate component called `ReadingStreakButton`, similar to `CompanionPopupButton`, but up to you in the end

packages/shared/src/lib/featureManagement.ts
```diff
@@ -18,6 +22,10 @@ const feature = {
   lowImps: new Feature('feed_low_imps'),
   bookmarkOnCard: new Feature('bookmark_on_card', false),
   feedLayout: new Feature('feed_layout', FeedLayout.Control),
+  readingStreaks: new Feature(
+    'reading_streaks',
+    ReadingStreaksExperiment.Control,
+  ),
```
DragosIorgulescu, line 28: nice, you knocked out [this ticket](https://dailydotdev.atlassian.net/browse/MI-77) as well with these changes (almost, if you add a hook as well)

DragosIorgulescu (approved):
Great work! I don't have any blocking comments, just some questions / suggestions.

packages/shared/src/components/layout/MainLayoutHeader.tsx
sshanzel: Yeah, that makes sense. Would be easier to update everything down the line.

packages/shared/src/components/layout/MainLayoutHeader.tsx
sshanzel: Agreed. This is to avoid cluttering the main component.

packages/shared/src/lib/featureManagement.ts
sshanzel: Well. I will add the hook then 😂

PR #2501, opened 2024-01-15
feat: first streak popup

sshanzel:
## Changes
- Popup to show once the user has reached the first streak.
- As mentioned in the ticket, connecting with live data/mutation should be part of another ticket.

Preview:

Screenshot 2024-01-15 at 4 50 08 PM

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

MI-12 #done


packages/shared/src/components/modals/common.tsx
```diff
@@ -96,6 +96,13 @@ const GenericReferralModal = dynamic(
     ),
 );
 
+const FirstStreakModal = dynamic(
+  () =>
+    import(
+      /* webpackChunkName: "genericReferralModal" */ './streaks/FirstStreakModal'
```
rebelchris, line 102:
```suggestion
      /* webpackChunkName: "firstStreakModal" */ './streaks/FirstStreakModal'
```

packages/shared/src/components/modals/streaks/FirstStreakModal.tsx
```diff
@@ -0,0 +1,52 @@
...
+  onRequestClose,
+  ...props
+}: LazyModalCommonProps): ReactElement {
+  return (
+    <Modal
+      {...props}
+      onRequestClose={onRequestClose}
+      kind={Modal.Kind.FlexibleCenter}
+      size={Modal.Size.XSmall}
+      className="p-6"
```
rebelchris, line 18: This seems odd, shouldn't it just maintain modal guidelines?

packages/shared/src/components/modals/streaks/FirstStreakModal.tsx
```diff
@@ -0,0 +1,52 @@
...
+    >
+      <CloseButton className="right-2 top-2" onClick={onRequestClose} />
```
rebelchris, line 20: Think you need absolute here as well with buttonV2 changes.
@kopancek ?

packages/shared/src/components/streak/popup/DayStreak.tsx
```diff
@@ -13,34 +15,43 @@ export enum Streak {
...
           secondary={streak === Streak.Completed}
-          size={IconSize.Medium}
+          className={className}
+          size={size}
         />
       );
     }
 
     return (
       <Circle
```
rebelchris, line 49: Should we maybe not use this as a Icon in this case?
Because it feels like we are abusing icon props to make it appear as one now?

rebelchris (commented):
Nice some minor comments

DragosIorgulescu (approved):
Looks good overall, I think once existing comments are resolved this should be good to go

packages/shared/src/components/modals/streaks/FirstStreakModal.tsx
sshanzel: There was actually no padding probably because I did not use the Modal.Body. Let me check if I should actually be using them.

packages/shared/src/components/streak/popup/DayStreak.tsx
sshanzel: But this is an actual icon that renders an svg file.

packages/shared/src/components/streak/popup/DayStreak.tsx
```diff
@@ -13,34 +15,43 @@ export enum Streak {
...
-        className={streak === Streak.Freeze && 'bg-theme-label-disabled'}
+        className={classNames(
+          className,
+          iconSizeToClassName[size],
```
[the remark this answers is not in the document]
sshanzel: Probably the primary issue is this, I felt like we don't have to turn this into an icon when we can simply create this as a div - and utilized the css just to make them consistent.

packages/shared/src/components/modals/streaks/FirstStreakModal.tsx
sshanzel: Moved to ModalClose. It should have a default `absolute` value.

packages/shared/src/components/streak/popup/DayStreak.tsx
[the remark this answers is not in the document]
rebelchris: Hmm just feels like it's a bit abusing things to me.
But if everyone else is happy not a blocker for me.

rebelchris (approved):
Non blocking left.

packages/shared/src/components/streak/popup/DayStreak.tsx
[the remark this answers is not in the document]
sshanzel: Yeah, totally understandable. Thoughts before merging? @dailydotdev/web-team - the image below is the elements we are referring to.
Screenshot 2024-01-16 at 4 54 58 PM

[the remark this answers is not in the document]
sshanzel: Merging this to unblock. Will then open a new PR if we find an alternative.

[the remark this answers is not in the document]
nensidosari: Personally think this should be a styled div...

nensidosari (approved):
On the Icon/Div discussions, I personally think that should be a simple styled Div rather then an icon, but non-blocking if people think it's best to be an Icon

PR #2502, opened 2024-01-15
feat: update cards body to fit with the new layout

DragosIorgulescu:
## Changes

- covered welcome, share, article & ad card types

Notes:
- ~~squad feed layout seems broken~~
- ~~mobileXL not being picked up properly~~

## Preview

**Article & Ad cards** 
Screenshot 2024-01-15 at 12 49 24
Screenshot 2024-01-15 at 12 49 12

**Freeform & Share cards**
Screenshot 2024-01-15 at 12 49 46

**Mobile**

https://github.com/dailydotdev/apps/assets/1225100/8526cfab-1377-43d9-ac5d-3e327e6711dd

## Events

N / A

### On those affected packages:
- [ ] Have you done sanity checks in the webapp?
- [ ] Have you done sanity checks in the extension?
- [ ] Does this not break anything in companion?

### Did you test the modified components media queries?
- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

#### Did you test on actual mobile devices?
- [ ] iOS (Chrome and Safari)
- [ ] Android

MI-16 #done

PR #2509, opened 2024-01-16
feat: add feedback CTA loop from post page for feed layout v1

kopancek:

Change the `useOnPostClick` hook implementation.
The flow is as following:
- If `ActiveFeedContext` provides `feedQueryKey`, behavior is the same as before
- If `shouldUseFeedLayoutV1` is `false`, behavior is the same as before
- If `shouldUseFeedLayoutV1` is `true` and `feedQueryKey` is undefined, we do the following:
  1. Get all feed query keys from react query cache
  2. For each feed query key:
    1. Find the correct `page`, `index` and `post` item based on the `post.id`
    2. Call the `updatePost` function with the `page`, `index`, `post`, and apply `read: true` 
  3. Use the go back button on the post page
  4. Profit from the updated cache 💪 

### Loom video
https://www.loom.com/share/461dd564a737480daca28722be801be1?sid=222b6c95-43fa-442c-a221-81bb4115007f 

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [X] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

MI-32 #done

### DIFF
--- a/packages/shared/src/hooks/useFeedLayout.ts
+++ b/packages/shared/src/hooks/useFeedLayout.ts
@@ -1,10 +1,8 @@
 import { useMemo } from 'react';
-import { useFeature } from '../components/GrowthBookProvider';
-import { feature } from '../lib/featureManagement';
-import { FeedLayout } from '../lib/featureValues';
 import { SharedFeedPage } from '../components/utilities';
 import { AllFeedPages } from '../lib/query';
 import { useActiveFeedNameContext } from '../contexts';
+import { useFeatureFeedLayoutV1 } from './useFeatureFeedLayoutV1';
 
 interface UseFeedLayoutProps {
   feedName?: AllFeedPages;
@@ -28,8 +26,7 @@ export const useFeedLayout = ({
 }: UseFeedLayoutProps = {}): UseFeedLayout => {
   const { feedName } = useActiveFeedNameContext();
   const name = (feedNameProp ?? feedName) as SharedFeedPage;
-  const feedLayoutVersion = useFeature(feature.feedLayout);
-  const isV1 = feedLayoutVersion === FeedLayout.V1;
+  const isV1 = useFeatureFeedLayoutV1();
   const isIncludedFeed = useMemo(
     () => checkShouldUseFeedLayoutV1(name),
     [name],
--- a/packages/shared/src/hooks/useFeedLayout.ts
+++ b/packages/shared/src/hooks/useFeedLayout.ts
@@ -14,10 +14,14 @@ interface UseFeedLayout {
   shouldUseFeedLayoutV1: boolean;
 }
 
+export const FeedLayoutV1FeedPages = new Set(
+  Object.values(SharedFeedPage).filter(
+    (feedPage) => feedPage !== SharedFeedPage.Search,
+  ),
+);
+
 const checkShouldUseFeedLayoutV1 = (feedName: SharedFeedPage): boolean =>
-  Object.values(SharedFeedPage)
-    .filter((feedPage) => feedPage !== SharedFeedPage.Search)
-    .includes(feedName) && feedName !== SharedFeedPage.Search;
+  FeedLayoutV1FeedPages.has(feedName) && feedName !== SharedFeedPage.Search;
 
 export const useFeedLayout = ({
   feedName: feedNameProp,
--- a/packages/shared/src/hooks/useOnPostClick.ts
+++ b/packages/shared/src/hooks/useOnPostClick.ts
@@ -15,11 +15,11 @@ import {
 import { Post, PostType } from '../graphql/posts';
 import { Origin } from '../lib/analytics';
 import { ActiveFeedContext } from '../contexts';
-import { AllFeedPages, updateCachedPagePost } from '../lib/query';
+import { updateCachedPagePost } from '../lib/query';
 import { usePostFeedback } from './usePostFeedback';
-import { FeedLayoutV1FeedPages, useFeedLayout } from './useFeedLayout';
-import { SharedFeedPage } from '../components/utilities';
+import { FeedLayoutV1FeedPages } from './useFeedLayout';
 import { FeedData } from '../graphql/feed';
+import { useFeatureFeedLayoutV1 } from './useFeatureFeedLayoutV1';
 
 interface PostClickOptionalProps {
   skipPostUpdate?: boolean;
@@ -99,10 +99,7 @@ export default function useOnPostClick({
   const { trackEvent } = useContext(AnalyticsContext);
   const { incrementReadingRank } = useIncrementReadingRank();
   const { queryKey: feedQueryKey, items } = useContext(ActiveFeedContext);
-  const { shouldUseFeedLayoutV1 } = useFeedLayout({
-    // need to provide something, in case feedName is undefined (such as on post page)
-    feedName: (feedName as AllFeedPages) || SharedFeedPage.MyFeed,
-  });
+  const isFeedLayoutV1 = useFeatureFeedLayoutV1();
 
   const { isLowImpsEnabled } = usePostFeedback();
 
@@ -160,7 +157,7 @@ export default function useOnPostClick({
             }
 
             updateFeedPostCache({ index: postIndex });
-          } else if (!feedName && shouldUseFeedLayoutV1) {
+          } else if (!feedName && isFeedLayoutV1) {
             const trySetPostRead = (queryKey: QueryKey, id: string) => {
               const updateFeedPost = updateCachedPagePost(
                 queryKey as unknown[],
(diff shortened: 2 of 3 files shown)


packages/shared/src/hooks/useOnPostClick.ts
```diff
@@ -30,13 +38,52 @@ export type FeedPostClick = ({
...
 }) => Promise<void>;
 
+const getFeedQueryKeys = (client: QueryClient): QueryKey[] => {
+  const queryKeys = client
+    .getQueryCache()
+    .findAll()
+    .map((query) => query.queryKey);
+
+  // filter only query keys for supported feed layout v1 feeds
+  const keys = queryKeys.filter(
```
capJavert, line 48: Since this can happen many times while user clicks on articles `reduce` could help to do filter + map in the same iteration.

capJavert (commented):
Looks good, nice work on caching 👍 

We just have to be careful to test control and make sure the events are sent in the same way.

packages/shared/src/hooks/useOnPostClick.ts
```diff
@@ -48,6 +95,11 @@ export default function useOnPostClick({
   const { trackEvent } = useContext(AnalyticsContext);
   const { incrementReadingRank } = useIncrementReadingRank();
   const { queryKey: feedQueryKey, items } = useContext(ActiveFeedContext);
+  const { shouldUseFeedLayoutV1 } = useFeedLayout({
+    // need to provide something, otherwise it's always false
+    feedName: SharedFeedPage.MyFeed,
+  });
```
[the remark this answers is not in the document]
kopancek: The problem is, that we don't have the current feed in context on the post page. Same with the `feedName` prop, it's not sed on the post page. This is a poor attempt to get the feature flag value.

If you have better idea on how to get the feed name in there, please let me know. I wanted to avoid passing a query parameter to the URL (seems ugly).

packages/shared/src/hooks/useOnPostClick.ts
```diff
@@ -81,27 +133,53 @@ export default function useOnPostClick({
...
+              }
+            };
+
+            getFeedQueryKeys(client).forEach((key) => {
+              trySetPostRead(key, post.id);
+            });
+          }
         }
       },
     // @NOTE see https://dailydotdev.atlassian.net/l/cp/dK9h1zoM
```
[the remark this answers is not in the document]
kopancek: Right, I need to take a look on the memoization of this. Since the handler is being used on every card, I think it still might make sense to memoize it, but maybe the parameter list needs to be updated 👍

packages/shared/src/hooks/useOnPostClick.ts
capJavert: Yeah, sure, non blocking.

packages/shared/src/hooks/useOnPostClick.ts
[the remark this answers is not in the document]
capJavert: Yeah, if you remove the eslint disable it should auto recommend all the dependencies.

packages/shared/src/hooks/useOnPostClick.ts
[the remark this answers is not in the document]
capJavert: The hook does have `feedName` prop so I would look into passing it somehow. `useFeedLayout` is mostly here to contain the logic of checking the feeds that should show v1.  But bottom line we just need to check if `useFeature(feature.feedLayout) === FeedLayout.V1` for post page and since we don't care which feed it is on that page. So I would just evaluate `useFeature` here. Would that work?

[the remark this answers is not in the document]
nensidosari: Yup if we don't care for this condition what feed we are one we should be good using useFeature, we've done so in other parts of the code changes

nensidosari (commented):
Have no further comments on what @capJavert has mentioned. Please also take a look at my previous comment [here](https://github.com/dailydotdev/apps/pull/2509#issuecomment-1894020118) on the missing image, I believe we might have removed that by mistake, would be great if we could bring it back in this PR 🙏 as it completes the ticket.

packages/shared/src/hooks/useOnPostClick.ts
```diff
@@ -48,6 +95,11 @@ export default function useOnPostClick({
   const { trackEvent } = useContext(AnalyticsContext);
   const { incrementReadingRank } = useIncrementReadingRank();
   const { queryKey: feedQueryKey, items } = useContext(ActiveFeedContext);
+  const { shouldUseFeedLayoutV1 } = useFeedLayout({
+    // need to provide something, otherwise it's always false
```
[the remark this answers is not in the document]
rebelchris: Maybe we can adjust the hook?
That if you don't provide the feedname it simply returns the value of the experiment key?

packages/shared/src/hooks/useFeedLayout.ts
```diff
@@ -14,10 +14,14 @@ interface UseFeedLayout {
   shouldUseFeedLayoutV1: boolean;
 }
 
+export const FeedLayoutV1FeedPages = new Set(
+  Object.values(SharedFeedPage).filter(
+    (feedPage) => feedPage !== SharedFeedPage.Search,
```
rebelchris, line 19: Search and Squad should be excluded right?
@nensidosari

packages/shared/src/hooks/useFeedLayout.ts
```diff
@@ -14,10 +14,14 @@ interface UseFeedLayout {
   shouldUseFeedLayoutV1: boolean;
 }
 
+export const FeedLayoutV1FeedPages = new Set(
```
rebelchris, line 17: I think we can maybe modify this hook to either return the raw value as well?
(also open to simply using it directly)

DragosIorgulescu:
> One thing I can see before looking at code, the image seems to be missing in the feedback loop post, was it like this? Apparently we've missed this too? 🤔 
> 
> 
> 
> Seems to have been done on this PR https://github.com/dailydotdev/apps/pull/2502/files#diff-621b28d595a78140d2ae6f1a1f816ec6fe5de3db072aaf76ecb52de90d984dfa
> 
> 
> 
> @DragosIorgulescu any reason that was removed?

I removed the component because it was no longer needed, since I moved most card images out of the footer. I did not realize at the time this affects the feedback card, my bad. Pushing a fix in a little bit 🙏

rebelchris (approved):
Nice work.
This is a very touchy element of the code, we need to tripple check that none of the tracking gets affected as they might not show directly.

DragosIorgulescu:
FYI, I added the image back in this PR: https://github.com/dailydotdev/apps/pull/2511 -- no need to do it here

packages/shared/src/hooks/useFeedLayout.ts
kopancek: Squad feed and search should not be part of the v1 feeds. See this thread: https://dailydotdev.slack.com/archives/C061T644XV3/p1705320332454659?thread_ts=1705316572.801329&cid=C061T644XV3

packages/shared/src/hooks/useFeedLayout.ts
kopancek: Do you mean to have another variable, like `export const FeedLayoutV1Feeds` that would export the array of values that are part of the feed layout v1? If not, please elaborate :)

packages/shared/src/hooks/useFeedLayout.ts
nensidosari: Yes yes, why you only see Search mentioned there @rebelchris  is because Squad is not part of SharedFeedPage.

kopancek:
> This is a very touchy element of the code, we need to tripple check that none of the tracking gets affected as they might not show directly.

Agreed, that's why I was cautious and I tried to make the new code path only work if:
1. `feedName` is not defined - this should only happen on post page
2. v1 feed layout feature is turned on

In all other cases, existing code path will be taken, which means any other existing features should work as they did before I touched this hook.

In all honesty, it makes the code uglier and harder to understand, since we now have a lot more conditional logic. But I think it's worth it to mitigate the potential risks.

packages/shared/src/hooks/useOnPostClick.ts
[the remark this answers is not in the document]
kopancek: I refactored out a new hook called `useFeatureFeedLayoutV1` and use that in both the `useFeedLayout` and `useOnPostClick` now. Should make the code more readable and easier to understand.

packages/shared/src/hooks/useOnPostClick.ts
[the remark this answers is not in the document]
kopancek: All the dependencies should now be added.

packages/shared/src/hooks/useFeedLayout.ts
rebelchris: Ah bit confusing to read then :)

packages/shared/src/hooks/useFeedLayout.ts
rebelchris: Yeah, but after thinking probably best to just use `useFeature` as we already do this in places.

DragosIorgulescu (approved):
nice work 💪

PR #2512, opened 2024-01-17
feat: handle zeros and colors for card action buttons in feed layout v1

kopancek:
### Screenshots && video

Classic cards not affected
Screenshot 2024-01-17 at 15 30 46

Feed layout v1 cards now behave well
Screenshot 2024-01-17 at 15 32 22

Screenshot 2024-01-17 at 15 27 46

https://www.loom.com/share/c92e27d866ef47e18c54989ce0e7d33f

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

MI-20 #done

### DIFF
--- a/packages/shared/src/components/InteractionCounter.tsx
+++ b/packages/shared/src/components/InteractionCounter.tsx
@@ -1,7 +1,6 @@
 import React, { ReactElement, useEffect, useState } from 'react';
 import { requestIdleCallback } from 'next/dist/client/request-idle-callback';
 import classNames from 'classnames';
-import styles from './InteractionCounter.module.css';
 
 export type InteractionCounterProps = {
   className?: string;
@@ -15,7 +14,6 @@ export default function InteractionCounter({
 }: InteractionCounterProps): ReactElement {
   const [shownValue, setShownValue] = useState(value);
   const [animate, setAnimate] = useState(false);
-
   useEffect(() => {
     if (value !== shownValue) {
       if (value < shownValue) {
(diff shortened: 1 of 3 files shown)


packages/shared/src/components/InteractionCounter.tsx
```diff
@@ -42,26 +45,24 @@ export default function InteractionCounter({
...
       >
         {shownValue}
       </span>
       <span
-        className={`absolute left-0 top-0 ${
-          animate ? 'translate-y-0 opacity-100' : 'translate-y-full opacity-0'
-        }`}
+        className={classNames(
+          childClassName,
+          animate ? '-translate-y-full opacity-100' : 'translate-y-0 opacity-0',
```
nensidosari, line 64: Is this change from `'translate-y-0 opacity-100' : 'translate-y-full opacity-0'` to `'-translate-y-full opacity-100' : 'translate-y-0 opacity-0'` intended?

nensidosari (commented):
Just one comment for one not sure intended/not intended change. Otherwise looks good. Happy to approve once that is resolved.

Also I see we don't check the testing on extension much here, missing changes that need to be done at extension as well. Still for this since extension is broken starting from the card, we should be good checking this in the other ticket specified for it.

packages/shared/src/components/InteractionCounter.tsx
kopancek: Yes. Before it was using absolute positioning and that doesn't work if we want to center the value element.

So I switched to a `flex-col` layout instead. The `span` elements are rendered below each other. Hence when we animate, we need to `translate-y` both of the spans 100% up. Does that make sense?

nensidosari: Makes sense in idea. For as long as you've tested this on both v1 and not v1 where it is used we should be fine 👍

nensidosari (approved):

PR #2515, opened 2024-01-18
fix: collection pills not clickable

kopancek:
## Changes

Removes pointer events from collection pill sources on CollectionCard in feed layout v1. This replicates the same behavior as we had with the existing layout.

Removes relative div from PostCardHeader
Adds relative positioning to the SourceButton and Squad source picture in the PostCardHeader, so that you can still get to the source when you click on the image.

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

MI-90 #done


nensidosari (approved):
LGTM, let's just make sure this doesn't break anything in non V1 feed 🙏

PR #2520, opened 2024-01-18
fix: set card images to `object-fill` on desktop

DragosIorgulescu:
## Changes

Note: 
- setting a max height that maintains 1:1 aspect ratio messes up with the `fill` setting, so we would need to set `cover` or `fill` depending on the image height, is this something we want to do? Or maybe someone has other suggestions?

----

Screenshot 2024-01-18 at 15 11 03

----

Screenshot 2024-01-18 at 15 10 47

----

Screenshot 2024-01-18 at 15 10 38

## Events

n / a

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done


packages/shared/src/components/cards/v1/WelcomePostCard.tsx
```diff
@@ -125,7 +125,7 @@ export const WelcomePostCard = forwardRef(function SharePostCard(
               alt="Post Cover image"
               src={image}
               fallbackSrc={cloudinary.post.imageCoverPlaceholder}
-              className="my-2 object-cover"
+              className="my-2 object-cover mobileXXL:self-start mobileXXL:object-fill"
```
nensidosari, line 128: Just suggestion to add this line in the component if it's being added to all components 🙏

nensidosari (approved):
LGTM, one non blocking comment.

Also 🙏 can you please post the note in the web channel, maybe someone has any ideas but just hasn't seen the PRs

packages/shared/src/components/cards/v1/WelcomePostCard.tsx
DragosIorgulescu: sry, missed your review. That's a good point, but these cards are used for the `share` and `feedback` cards as well, and we don't want these classes there 🙏

nensidosari (changes requested):
This PR does not fix issues with large images as it doesn't allow for max height: https://dailydotdev.slack.com/archives/C02E2C3C13R/p1705657476639799?thread_ts=1705590446.763629&cid=C02E2C3C13R

Will come back to this.

nensidosari (approved):
This is good to go, large images to be tackled in another PR

PR #2529, opened 2024-01-22
fix: change icon color to purple

nensidosari:
- Changed icon color of play to purple and default size to XXL

- [x] Does this not break anything in companion? Changes play icon used there to default secondary with purple play icon

- [x] MobileL (420px)
- [x] Tablet (656px)
- [x] Laptop (1020px)

MI-3 #done


rebelchris (approved):
Works, not sure why the icon size is bigger now? (it's a svg so would have scalled)
Also assumed at first the color would be programatically, but guess this can do.

PR #2530, opened 2024-01-22
fix: unnecessary video overlay

sshanzel:
## Changes
- The `h-auto` is consuming more height than intended.
- Applying the `h-fit` makes the container fit to the actual content.
- The control value uses the same component. Hence, should fix both ends.

I have thought of removing the floating `span` overlay at first, but it is probably needed in some cases. ~~Though not sure now whether that background overlay will ever be seen.~~

@dailydotdev/web-team does anyone know why we need that overlay? The color has a bit of opacity, which is why it is turning out to be **not** pitch-black.

Edit: is it probably for the actual overlay that makes the thumbnail a bit dark?

Before:

Screenshot 2024-01-22 at 3 21 02 PM

After:

Screenshot 2024-01-22 at 3 24 35 PM

## Events

Did you introduce any new tracking events?
Don't forget to update the [Analytics Taxonomy sheet](https://docs.google.com/spreadsheets/d/18Lv7zXges9QfVX5VYL1a-Hyl0e1sQ3sLr0OK8YZWKXI/edit#gid=0)

Log the new/changed events below:

| Type   | event_name  | value |
|--------|-------------|-------|
| Change/New | event name  | extra: { ... } |

### On those affected packages:
- [ ] Have you done sanity checks in the webapp?
- [ ] Have you done sanity checks in the extension?
- [ ] Does this not break anything in companion?

### Did you test the modified components media queries?
- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

#### Did you test on actual mobile devices?
- [ ] iOS (Chrome and Safari)
- [ ] Android

MI-93 #done

PR #2531, opened 2024-01-22
fix: image overlay on both control and v1

sshanzel:
## Changes
Reverted back to h-auto as default then override on v1. To fix the control, we set the `max-height: fit-content`. 
Tested 3 cases:
- Feed control
- Feed control shared video post
- Feed v1

Fixes the issue here: https://github.com/dailydotdev/apps/pull/2530#issuecomment-1903415156

Screenshot 2024-01-22 at 3 51 05 PM
Screenshot 2024-01-22 at 3 50 55 PM
Screenshot 2024-01-22 at 3 50 48 PM

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done

### DIFF
--- a/packages/shared/src/components/cards/v1/ArticlePostCard.tsx
+++ b/packages/shared/src/components/cards/v1/ArticlePostCard.tsx
@@ -139,7 +139,7 @@ export const ArticlePostCard = forwardRef(function PostCard(
                 loading="lazy"
                 data-testid="postImage"
                 {...(isVideoType && {
-                  wrapperClassName: 'mt-4 mobileXL:w-40 mobileXXL:w-56',
+                  wrapperClassName: 'mt-4 mobileXL:w-40 mobileXXL:w-56 !h-fit',
                 })}
               />
             </CardContent>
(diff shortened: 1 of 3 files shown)


packages/shared/src/components/cards/v1/ArticlePostCard.tsx
```diff
@@ -139,7 +139,7 @@ export const ArticlePostCard = forwardRef(function PostCard(
                 loading="lazy"
                 data-testid="postImage"
                 {...(isVideoType && {
-                  wrapperClassName: 'mt-4 mobileXL:w-40 mobileXXL:w-56',
+                  wrapperClassName: 'mt-4 mobileXL:w-40 mobileXXL:w-56 !h-fit',
```
nensidosari, line 142: This is changing it only in ArticlePostCard for v1, did we want to change for Collections, Shared, Welcome cards as well in V1 - or those are not affected?

nensidosari (commented):
Just one small check. See comment

packages/shared/src/components/cards/v1/ArticlePostCard.tsx
sshanzel: Let me deploy in preview to further check.

sshanzel: For Welcome, and Collections, we should not need the fix for v1 there since the type itself can never be video. For Shared, I have now pushed the same just to be on the safe side.

nensidosari (approved):

PR #2533, opened 2024-01-22
fix: autofill input background

binalajudiya:
- The username input field now maintains a consistent and proper styling, ensuring a seamless appearance within the form when use Autocomplete. 
- Just applied to fields on account-> profile page. 
Screenshot 2024-01-22 at 9 59 18 AM

## Events

Did you introduce any new tracking events? - No

- [ ] Have you done sanity checks in the webapp? - Yes
- [ ] Have you done sanity checks in the extension? - Yes
- [ ] Does this not break anything in companion? - Yes

- [ ] MobileL (420px)  - Yes
- [ ] Tablet (656px)  - Yes
- [ ] Laptop (1020px)  - Yes

- [ ] iOS (Chrome and Safari)  - Yes
- [ ] Android - Yes

#done


sshanzel (commented):
Nice! This looks great! 🚀 

Though the fix here is correct, we were wanting it to be applied across the app for all input fields.

We have a `base.css` file, we can move the classes applied on that end using the `input` selector.

As for the plugin we should set the fill color with `var(--theme-label-primary)` in order to make it work for both light and dark themes.

packages/shared/tailwind.config.js
```diff
@@ -238,7 +238,21 @@ module.exports = {
...
+  plugins: [
+    caret, 
+    typography, 
+    buttons,
+    function ({ addUtilities }) {
+      const newUtilities = {
+        '.bg-clip-text': {
+          'background-clip': 'text',
+          '-webkit-background-clip': 'text',
+          '-webkit-text-fill-color': '#ffffff',
```
sshanzel, line 250: Just to make it work with both themes 🙌 
```suggestion
          '-webkit-text-fill-color': 'var(--theme-label-primary)',
```

packages/shared/tailwind.config.js
```diff
@@ -238,7 +238,21 @@ module.exports = {
...
   },
   // eslint-disable-next-line global-require
-  plugins: [caret, typography, buttons],
+  plugins: [
+    caret, 
+    typography, 
+    buttons,
+    function ({ addUtilities }) {
+      const newUtilities = {
+        '.bg-clip-text': {
```
rebelchris, line 247: This is already a tailwind class right?
https://tailwindcss.com/docs/background-clip

sshanzel: Yeah, but the color fill doesn't match our requirements. Should we maybe let go of the plugin and add the missing CSS property on the base.css input selector?

binalajudiya: @sshanzel Thanks, I've removed `bg-clip-text` from the plugin. As it is already exists, I used it and applied `var(--theme-label-primary)` to the `input` selector on base.css, so now it is consistent. Also, I am sure there must be a way to add `bg-clip-text` on the global level to the `input` selector instead of adding separate.

packages/shared/src/styles/base.css
```diff
@@ -487,3 +487,7 @@ details.right-icon {
     border-radius: 14px;
   }
 }
+
+input {
+  -webkit-text-fill-color: var(--theme-label-primary);
+}
```
sshanzel, line 493: @binalajudiya this should apply the available class from tailwind 🥳 

After this, I think you can revert the changes from the `profile` file, and see if it works as intended.
```suggestion
input {
  @apply bg-clip-text;

  -webkit-text-fill-color: var(--theme-label-primary);
}

```

packages/shared/tailwind.config.js
sshanzel: Kindly see this comment https://github.com/dailydotdev/apps/pull/2533#discussion_r1463008425.

You are almost there! 🚀

packages/shared/src/styles/base.css
binalajudiya: @sshanzel I applied, tested and it worked  🚀

Reverted `profile` file.

sshanzel (approved):
Looking great code-wise! Well done! 🥳 

Will check the preview and see how it looks personally.

sshanzel:
Tested locally, both login form and profile page inputs are now aligned with the changes here 🥳  Wdyt? @rebelchris

rebelchris (approved):
Looks good :)

sshanzel:
@binalajudiya thank you so much for your contribution! 🚀 ✨

PR #2534, opened 2024-01-22
fix: set 16px font size for article summary

kopancek:
## Changes

Changes font size for summary to be consistent on extension as well.

### Screenshots

Before
Screenshot 2024-01-22 at 11 33 08

After
Screenshot 2024-01-22 at 12 24 58

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [X] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

MI-114 #done

### DIFF
--- a/packages/shared/src/components/cards/v1/ArticlePostCard.tsx
+++ b/packages/shared/src/components/cards/v1/ArticlePostCard.tsx
@@ -116,7 +116,7 @@ export const ArticlePostCard = forwardRef(function PostCard(
                 </CardTitle>
 
                 {post.summary && (
-                  <CardTextContainer className="mt-4 text-theme-label-secondary text-base">
+                  <CardTextContainer className="mt-4 text-theme-label-secondary typo-body">
                     {summary}
                   </CardTextContainer>
                 )}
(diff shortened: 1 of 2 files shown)


omBratteng (approved):

packages/shared/src/components/cards/v1/ArticlePostCard.tsx
```diff
@@ -116,7 +116,7 @@ export const ArticlePostCard = forwardRef(function PostCard(
                 </CardTitle>
 
                 {post.summary && (
-                  <CardTextContainer className="mt-4 text-theme-label-secondary">
```
rebelchris, line 119: We should use our typo-body class I think, just because line-heights might not match. 
(actually noticed it's slightly different on webapp as well then, so better inline with design)

rebelchris (changes requested):
Small request to be more inline with figma

packages/shared/src/components/cards/v1/ArticlePostCard.tsx
kopancek: TIL, thanks!

kopancek: Should be fixed now (although I didn't bother to change the screenshot, hope that's OK).

nensidosari (approved):
LGTM

rebelchris (approved):
Amazing amazing

PR #2539, opened 2024-01-22
fix: comments button color and type

nensidosari:
- removed secondary type from button to get the type based on figma
- on button hover the color of nr of comments should be same color as the highlighted comments button

Screenshot 2024-01-22 at 11 23 18 PM
Screenshot 2024-01-22 at 11 23 12 PM

- [x] MobileL (420px)
- [x] Tablet (656px)
- [x] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

MI-116 #done

### DIFF
--- a/packages/shared/src/components/cards/v1/ActionButtons.tsx
+++ b/packages/shared/src/components/cards/v1/ActionButtons.tsx
@@ -152,9 +152,9 @@ export default function ActionButtons({
           <Link href={post.commentsPermalink}>
             <Button
               id={`post-${post.id}-comment-btn`}
-              className="pointer-events-auto ml-2"
+              className="pointer-events-auto ml-2 hover:text-theme-color-blueCheese"
               color={ButtonColor.BlueCheese}
-              icon={<CommentIcon secondary={post.commented} />}
+              icon={<CommentIcon />}
               tag="a"
               href={post.commentsPermalink}
               pressed={post.commented}
@@ -163,12 +163,7 @@ export default function ActionButtons({
             >
               {post?.numComments > 0 ? (
                 <InteractionCounter
-                  className={classNames(
-                    'tabular-nums',
-                    post.commented
-                      ? 'text-theme-color-blueCheese'
-                      : 'text-theme-label-tertiary',
-                  )}
+                  className="tabular-nums"
                   value={post.numComments}
                 />
               ) : null}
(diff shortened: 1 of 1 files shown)


packages/shared/src/components/cards/v1/ActionButtons.tsx
```diff
@@ -152,23 +152,17 @@ export default function ActionButtons({
           <Link href={post.commentsPermalink}>
             <Button
               id={`post-${post.id}-comment-btn`}
-              className="pointer-events-auto ml-2"
+              className="pointer-events-auto ml-2 hover:text-theme-color-blueCheese"
               color={ButtonColor.BlueCheese}
-              icon={<CommentIcon secondary={post.commented} />}
```
sshanzel, line 157: I think this was intended right? When the user has made a comment we fill it.

nensidosari: Ah yes, could be, it's not in designs, but if this is what the behaviour used to be, should be the same. 
Would `post.commented` check if the user itself has commented? Or if the post is commented by any user?

packages/shared/src/components/cards/v1/ActionButtons.tsx
```diff
@@ -152,23 +152,17 @@ export default function ActionButtons({
...
+              icon={<CommentIcon />}
               tag="a"
               href={post.commentsPermalink}
-              pressed={post.commented}
               variant={ButtonVariant.Float}
               {...combinedClicks(() => onCommentClick?.(post))}
             >
               {post?.numComments > 0 ? (
                 <InteractionCounter
-                  className={classNames(
```
nensidosari, line 166: I'll change this back as well based on the question/answer above

packages/shared/src/components/cards/v1/ActionButtons.tsx
sshanzel: Yeah, I checked the Figma and it doesn't seem to be there. It checks whether the logged in user has made a comment.

nensidosari: Have updated to only make the required change. thank u!!

sshanzel (approved):
Looks great! 🚀

omBratteng (approved):

PR #2547, opened 2024-01-23
feat: current search provider

capJavert:
- [x] current search provider icon
- [x] current search provider label
- [x] also add hover and tab functionality to toggle between providers

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

AS-34 #done


packages/shared/src/components/KeyboardShortcutLabel.tsx
```diff
@@ -10,6 +10,19 @@ export const KeyboadShortcutLabel = ({
   className,
   keys,
 }: KeyboadShortcutLabelProps): ReactElement => {
+  const [mounted, setMounted] = useState(false);
+
+  useEffect(() => {
+    // keyboard shortcut buttons are most of the times dependant
+    // on browser features so we don't render them in SSR to
+    // avoid hydration mismatches
```
rebelchris, line 18: Not for now, but should we maybe add a backlog ticket to make this into it's own hook for clarity down the line?

packages/shared/src/components/MainFeedLayout.tsx
```diff
@@ -247,11 +254,27 @@ export default function MainFeedLayout({
...
     <FeedPageComponent
       className={classNames('relative', shouldUseFeedLayoutV1 && '!pt-0')}
     >
-      <div id={searchPanelGradientElementId} />
+      {isSearchPanelGradientActive && (
+        <FeedGradientBg
+          className={classNames(
+            'right-0 -z-1 hidden translate-x-0 opacity-0 transition-opacity duration-500 laptop:flex',
+            sidebarExpanded
+              ? '!-left-32 !max-w-[calc(100%)]'
```
rebelchris, line 272: Why do you need the calc 100% here?

packages/shared/src/components/search/SearchPanel/SearchPanelAction.tsx
```diff
@@ -16,8 +18,16 @@ const iconToProviderMap: Record<
   FunctionComponent<IconProps>
 > = {
   [SearchProviderEnum.Posts]: SearchIcon,
-  // TODO AS-3-search-merge replace with correct icon
-  [SearchProviderEnum.Chat]: AiIcon,
+  [SearchProviderEnum.Chat]: ({ className, ...rest }: IconProps) => (
+    <MagicIcon
+      className={classNames(
+        className,
+        'rounded-6 bg-gradient-to-t from-onion-40 to-cabbage-40 p-0.5 text-white',
```
rebelchris, line 25: Same comment from other PR on the colors

rebelchris (changes requested):
Minor queries so far :)

packages/shared/src/components/KeyboardShortcutLabel.tsx
capJavert: Feel like hook would hide it and make it more magical and then in hindsight used whenever someone encountered any SSR issue vs most of the time you can solve it by other mechanisms. Agree we should keep it in check this is why I added comment for now to explain.

packages/shared/src/components/MainFeedLayout.tsx
capJavert: Good catch, I think I was playing with values and then left it.

packages/shared/src/components/KeyboardShortcutLabel.tsx
rebelchris: Yeah comment is perfect for now :)

rebelchris (approved):
Looks good to me

PR #2551, opened 2024-01-24
fix: search highlight effect

sshanzel:
## Changes
- Replace the image by using CSS to avoid the square cut from both ends.

Preview:

Screenshot 2024-01-24 at 8 33 37 PM
Screenshot 2024-01-24 at 8 33 51 PM
Screenshot 2024-01-24 at 8 37 01 PM
Screenshot 2024-01-24 at 8 36 54 PM
Screenshot 2024-01-24 at 8 36 22 PM
99)

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done

### DIFF
--- a/packages/shared/src/components/feeds/FeedContainer.tsx
+++ b/packages/shared/src/components/feeds/FeedContainer.tsx
@@ -187,11 +187,10 @@ export const FeedContainer = ({
       )}
     >
       {isV1Search && shouldUseFeedLayoutV1 && (
-        <img
-          className="absolute left-0 top-0 w-full -translate-y-1/2"
-          src={cloudinary.feed.bg.layoutV1}
-          alt="Gradient background"
-        />
+        <div className="absolute -top-24 left-1/2 flex -translate-x-1/2 flex-row justify-center">
+          <span className="z-1 size-40 translate-x-6 rounded-full bg-theme-color-cabbage blur-[5rem]" />
+          <span className="z-0 h-40 w-70 -translate-x-6 rounded-[50%] bg-theme-color-onion blur-[5rem]" />
+        </div>
       )}
       <ScrollToTopButton />
       <div className="flex w-full flex-col laptopL:mx-auto" style={style}>
--- a/packages/shared/src/components/feeds/FeedContainer.tsx
+++ b/packages/shared/src/components/feeds/FeedContainer.tsx
@@ -33,7 +33,7 @@ import {
 import ConditionalWrapper from '../ConditionalWrapper';
 import { SharedFeedPage } from '../utilities';
 import { useActiveFeedNameContext } from '../../contexts';
-import { cloudinary } from '../../lib/image';
+import { FeedGradientBg } from './FeedGradientBg';
 
 export interface FeedContainerProps {
   children: ReactNode;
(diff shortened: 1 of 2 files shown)


packages/shared/src/components/feeds/FeedContainer.tsx
```diff
@@ -187,11 +187,10 @@ export const FeedContainer = ({
       )}
     >
       {isV1Search && shouldUseFeedLayoutV1 && (
-        <img
-          className="absolute left-0 top-0 w-full -translate-y-1/2"
-          src={cloudinary.feed.bg.layoutV1}
-          alt="Gradient background"
-        />
+        <div className="absolute -top-24 left-1/2 flex -translate-x-1/2 flex-row justify-center">
```
kopancek, line 190: why do we use both `-top-24` and `-translate-x-1/2`?

packages/shared/src/components/feeds/FeedContainer.tsx
```diff
@@ -187,11 +187,10 @@ export const FeedContainer = ({
...
+          <span className="z-1 size-40 translate-x-6 rounded-full bg-theme-color-cabbage blur-[5rem]" />
```
kopancek, line 191: Using translate instead of margin seems unnecessary to me. We can also get rid of the z-index here by swapping the order of the elements, e.g. ellipse first and circle second in the DOM.

kopancek (approved):
Looks good to me, just added a few non-blocking discussion comments.

capJavert (approved):
Not blocking but we could probably move it to component like `FeedGradientBg` for reuse?

packages/shared/src/components/feeds/FeedContainer.tsx
sshanzel: We don't display the full size, instead, we show partial.

The `-translate-x-1/2` is to make it on absolute centered position working with `left-1/2`.

kopancek: OK, makes sense. I would maybe prefer to not mix absolute positioning with CSS transforms, but whatever works :)

packages/shared/src/components/feeds/FeedContainer.tsx
```diff
@@ -187,11 +187,10 @@ export const FeedContainer = ({
...
+          <span className="z-0 h-40 w-70 -translate-x-6 rounded-[50%] bg-theme-color-onion blur-[5rem]" />
```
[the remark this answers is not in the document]
kopancek: Since you define the width to be different than height, it will not become a circle, even if you specify `rounded-full`.

packages/shared/src/components/feeds/FeedContainer.tsx
sshanzel: Yeah, we can probably remove z-index by doing flow reverse then swap the elements around.

With regards to the translate vs margin. I think it is more appropriate to use translate when fixing the actual placement of the element. Margins, I believe, are used for spacing, just probably usable when moving elements as well. Wdyt?

sshanzel:
@rebelchris I tried to remove the image and make it fit the layout for Control value for feed layout. Wdyt?

Old implementation:
Screenshot 2024-01-24 at 9 36 18 PM
Screenshot 2024-01-24 at 9 36 13 PM
Screenshot 2024-01-24 at 9 35 54 PM

CSS implementation:
Screenshot 2024-01-24 at 9 34 35 PM
Screenshot 2024-01-24 at 9 34 29 PM
Screenshot 2024-01-24 at 9 34 19 PM

packages/shared/src/components/feeds/FeedContainer.tsx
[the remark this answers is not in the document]
sshanzel: Yeah, but it looks different.
Screenshot 2024-01-24 at 9 39 21 PM

rebelchris (approved):
works for me :)

sshanzel:
@kopancek I've merged now to unblock, but we can discuss more of the technical details which can come as a follow-up PR. At least we can start testing everything while the discussion is ongoing.

PR #2553, opened 2024-01-24
fix: scroll to top button position on iOS

kopancek:
## Changes

Fixed by moving the ScrollToTop button to be part of the footer navbar container instead. Also changed the positioning from fixed to absolute. The navbar is already fixed positioned. This way, if the footer navbar height or position changes, the button moves as well. So we should never cover the scroll to top button with the navbar anymore.

### Screenshots

Screenshot 2024-01-24 at 15 52 30
Screenshot 2024-01-24 at 15 52 08

- [X] MobileL (420px)
- [X] Tablet (656px)
- [X] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

MI-100 #done


packages/extension/src/newtab/MainFeedPage.tsx
```diff
@@ -97,45 +98,50 @@ export default function MainFeedPage({
...
-                  : 'mt-4 w-fit [@media(width<=680px)]:mx-6',
-              )}
-            />
-          }
-        />
-      </FeedLayout>
-      <DndModal isOpen={showDnd} onRequestClose={() => setShowDnd(false)} />
-    </MainLayout>
+    <>
+      <div className="fixed bottom-0 left-0 z-2 w-full">
```
rebelchris, line 102: I'm not really sure what this PR is intended to fix/what the issue was.

But setting a main wrapped to fixed is 100% guaranteed to give issues if you ask me.
Can you maybe show me the issue, perhaps we can solve it otherwise? 🙏

rebelchris (changes requested):

packages/extension/src/newtab/MainFeedPage.tsx
kopancek: Well, this is not setting main to fixed. It is adding another element into main that is fixed. Does that make sense?

Originally - the scroll to top button has a fixed position with fixed bottom coordinates. But on iOS in some cases, the browser does not respect that, e.g. the bottom navbar appears to be taller (this happens for example if you display the browser in multitasking on right half of the screen on the iPad). So in cases like that, the bottom navbar can cover the scroll to top button (at least partially). See https://dailydotdev.atlassian.net/browse/MI-100. 

The obvious fix here was to add the scroll to top button to the bottom navbar, hence the coordinates will be always right and the navbar will never cover the button.

Let me know what you think, we can also sync on it together, but you need a physical iOS device for this, since it does not happen on a simulator.

nensidosari (commented):
Looks alright to me. 
Maybe check with @rebelchris one more time if he still has any concerns or better idea. 🙏

packages/extension/src/newtab/MainFeedPage.tsx
sshanzel: You won't have to about mobile on the extension since the extension is only for desktops. We don't normally cover the case.

rebelchris: As Lee mentioned if we are only introducing this for extension mobile view I'd rather not do it.
Adds more edge-case then solves them I think?

rebelchris:
@kopancek // @nensidosari I'm really not sure about the origin on this ticket.
*(would have to see it in action)

But to me wrapping our main layout as fixed seems more likely to introduce weird other edge-case we haven't covered now.
So to me rather not solve it as presented. 

Maybe we can rather sync on it, as I'm totally missing the issue here, and might review it wrong.

nensidosari:
@rebelchris , not sure if you've missed this part of @kopancek [comment](https://github.com/dailydotdev/apps/pull/2553#discussion_r1465960965) on it:

> Well, this is not setting main to fixed. It is adding another element into main that is fixed. 

So there is another fixed element outside of main being added, not setting main to fixed. Does this solve your concern?
But yeah free to sync whenever you have time

rebelchris (approved):
Missed the latest change of non wrapping :)

nensidosari (approved):
Since concerns are solved, looks good

PR #2561, opened 2024-01-26
feat: email verification

rebelchris:
- Introduces the new code verification flow

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

AS-5 #done


packages/shared/src/components/auth/OnboardingRegistrationForm.tsx
```diff
@@ -39,20 +53,88 @@ const OnboardingRegistrationForm = ({
...
+      target_type: AuthEventNames.SignUpProvider,
+      target_id: 'email',
+      extra: JSON.stringify({ trigger }),
+    });
+
+    const email = getFormEmail(e);
+
+    if (!email) {
+      return null;
+    }
```
capJavert, line 73: What happens here if there is no email, is error messages shown?

packages/shared/src/components/auth/EmailCodeVerification.tsx
```diff
@@ -0,0 +1,106 @@
...
+      flow: AuthFlow.Verification,
+      flowId,
+      onError: setHint,
+      onVerifyCodeSuccess: () => {
+        onSubmit();
+      },
+    });
+  if (autoResend && !hint) {
+    setHint('Please click resend code to get a new code.');
+  }
```
capJavert, line 41: I think needs to go to `useEffect` because you are setting state during render.

packages/shared/src/hooks/useAccountEmailFlow.ts
```diff
@@ -34,35 +47,58 @@ type EmailFlow = AuthFlow.Recovery | AuthFlow.Verification;
...
-        : { enabled: isQueryEnabled }),
+      ...disabledRefetch,
+      enabled,
     },
   );
 
+  if (emailFlow?.id && !activeFlow) {
+    setActiveFlow(emailFlow.id);
+    setAutoResend(true);
+  }
```
capJavert, line 93: This also needs to go to `useEffect` or potentially you can just memoize and avoid extra state since it just depends on emailFlow value.

capJavert (changes requested):
Looks good, little bit harder to track with all the callbacks but I think you got it 👌

Some stuff to cleanup to potentially avoid some weird edge cases.

packages/shared/src/components/auth/AuthDefault.tsx
```diff
@@ -135,6 +136,14 @@ const AuthDefault = ({
     return <EmailSignupForm onSubmit={onEmailSignup} isReady={isReady} />;
   };
 
+  const getOrDivider = () => {
```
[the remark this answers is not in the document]
omBratteng: Could add a new prop `show` or `display` to `OrDivider` that accepts a boolean statement, and that decides wether to render it or not?

packages/shared/src/components/auth/CodeVerificationForm.tsx
```diff
@@ -46,12 +46,12 @@ function CodeVerificationForm({
 
   const onCodeVerification = async (e) => {
     e.preventDefault();
-    trackEvent({
-      event_name: AuthEventNames.SubmitForgotPassword,
-    });
+    // trackEvent({
+    //   event_name: AuthEventNames.SubmitForgotPassword,
+    // });
```
omBratteng, line 51: Just commenting to check if this is intended to be removed in prod

packages/shared/src/lib/kratos.ts
```diff
@@ -340,11 +363,25 @@ export const getKratosSession = async (): Promise<AuthSession> => {
   return res.json();
 };
 
+export const getVerificationSession =
+  async (): Promise<VerificationSession> => {
+    const res = await fetch(`${heimdallUrl}/api/get_verification_flow`, {
+      credentials: 'include',
+    });
+
+    if (res.status === 401) {
```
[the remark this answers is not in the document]
rebelchris: I think it's specific for kratos errors as they are maintained by us.
Can't fully recall but it's not a new addition basically.

packages/shared/src/components/auth/AuthDefault.tsx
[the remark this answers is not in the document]
rebelchris: It's not really a new thing, but can check what I can do.
Wanted to avoid to many changes that don't really play part.

packages/shared/src/hooks/useAccountEmailFlow.ts
rebelchris: Good point, I kind of assumed the values are available on the re-render which seems to be enough.
What case would they not be up to date then? (where you need the useEffect) (kind of feels like that would just trigger yet another re-render?)

packages/shared/src/components/auth/CodeVerificationForm.tsx
rebelchris: Ah good point actually needed to check with @idoshamun if we wanted specific tracking?
(this above is just a copy of existing tracking)

packages/shared/src/components/auth/AuthDefault.tsx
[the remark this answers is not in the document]
capJavert: I was just saying the above can be a component to make it more Reac-tish, no need to call as function.

packages/shared/src/components/auth/CodeVerificationForm.tsx
idoshamun: Yes let's add 
image

packages/shared/src/hooks/useAccountEmailFlow.ts
capJavert: Setting state during render will do another render which can then do another and another so you can end up in infinite loop really fast. This is why it is not recommended. Usually it can be replaced with:
- `useEffect` + `useState`
- `useMemo` - just compute new values from current state, which I think you can do in this case.

packages/shared/src/lib/kratos.ts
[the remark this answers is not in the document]
capJavert: Still I think we should treat any error as no active session right?

packages/shared/src/hooks/useAccountEmailFlow.ts
rebelchris: Alright let me memo it for now :) 
I don't like to many useEffects lol

packages/shared/src/components/auth/AuthDefault.tsx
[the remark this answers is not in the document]
rebelchris: Went for inline evaluation basically, seems about right.
Just wondering how do you feel about the getForm function here? (Multiple conditions/mixes)

packages/shared/src/components/auth/EmailSignupForm.tsx
```diff
@@ -8,9 +14,35 @@ interface EmailSignupFormProps {
 }
 
 function EmailSignupForm({
```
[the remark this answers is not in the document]
rebelchris: Screenshot 2024-01-29 at 19 14 16
Is used though?

packages/shared/src/components/auth/AuthDefault.tsx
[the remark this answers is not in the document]
rebelchris: Can log it for tech debt for sure.
But anyway you could pass props to the function right?

Not sure besides semantics if it adds anything?

capJavert (approved):

packages/shared/src/components/auth/AuthDefault.tsx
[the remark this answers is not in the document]
capJavert: semantics yes

idoshamun:
@rebelchris anything specific to review here? I'm a bit out of context in the apps already 🙈

rebelchris:
@idoshamun From your perspective helping on a E2E test would probably be best.
I'll log it down once ready in the channel with where it's deployed.

Alternative perhaps the tracking only?

packages/shared/src/components/auth/OnboardingRegistrationForm.tsx
```diff
@@ -39,20 +53,88 @@ const OnboardingRegistrationForm = ({
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [shouldLogin]);
 
+  const onEmailSignup = async (e: React.FormEvent) => {
+    e.preventDefault();
+    if (isLoading) {
+      return null;
+    }
+
+    trackEvent({
```
idoshamun, line 62: Let's add here also the new pixel tracking Lee added

PR #2564, opened 2024-01-28
feat: reading streak button and popup integration

sshanzel:
## Changes
- Utilized an existing query used in the heatmap to check whether the user has read from previous days but limited the amount.
- Fixed the issue when the popup is opened, it should close when the user clicks anywhere outside.
- Integrates the popup content with the live data.

The complete days indicated in the image below is true from the live data.

Screenshot 2024-01-29 at 4 40 20 AM

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

MI-78 #done

### DIFF
--- a/packages/shared/src/components/streak/popup/ReadingStreakPopup.tsx
+++ b/packages/shared/src/components/streak/popup/ReadingStreakPopup.tsx
@@ -1,9 +1,13 @@
-import React, { ReactElement } from 'react';
-import { addDays, subDays } from 'date-fns';
+import React, { ReactElement, useMemo } from 'react';
+import { addDays, isSameDay, subDays } from 'date-fns';
+import { useQuery } from '@tanstack/react-query';
 import { Button, ButtonVariant } from '../../buttons/ButtonV2';
 import { StreakSection } from './StreakSection';
 import { DayStreak, Streak } from './DayStreak';
 import { ButtonSize } from '../../buttons/Button';
+import { generateQueryKey, RequestKey } from '../../../lib/query';
+import { getReadingStreak30Days, UserStreak } from '../../../graphql/users';
+import { useAuthContext } from '../../../contexts/AuthContext';
 
 const today = new Date();
 const dateToday = today.getDate();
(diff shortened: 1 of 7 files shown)


nensidosari (approved):
LGTM

sshanzel:
@DragosIorgulescu let me know if you want to add anything here 🚀

packages/shared/src/components/streak/popup/ReadingStreakPopup.tsx
```diff
@@ -19,52 +23,74 @@ const streakDays = [
...
-              return Streak.Completed;
-            }
+      const getStreak = () => {
+        if (isCompleted) {
+          return Streak.Completed;
+        }
 
-            return Streak.Upcoming;
-          };
+        if (day === 0 || day === 6) {
```
DragosIorgulescu, line 58: should this not be 5 & 6?
```suggestion
        if (day === 5 || day === 6) {
```

packages/shared/src/graphql/users.ts
```diff
@@ -423,3 +434,46 @@ export const UNSUBSCRIBE_PERSONALIZED_DIGEST_MUTATION = gql`
...
+interface ReadingDay {
+  date: string;
+  reads: number;
+}
+
+export const getReadingStreak30Days = async (
+  id: string,
+): Promise<ReadingDay[]> => {
+  const today = new Date();
+  const last30Days = subDays(today, 30);
```
DragosIorgulescu, line 447: should we allow this to be passed as an argument rather than being hard-coded? I am thinking that we might want to just get the last 7 days here since we don't display more than that in the `ReadingStreakPopup` at the moment

DragosIorgulescu (approved):
nice work, I added a couple non-blocking comments 🙏

packages/shared/src/components/streak/popup/ReadingStreakPopup.tsx
kopancek: 0 is Sunday, 6 is Saturday, [see the `getDay` function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getDay#return_value).

Maybe we can add const or enum values for these here as well, like you did in the other PR @sshanzel ?

packages/shared/src/components/streak/popup/ReadingStreakPopup.tsx
```diff
@@ -19,52 +23,74 @@ const streakDays = [
...
 
-          // this is just a dummy logic to simulate the popup with different values
-          const getStreak = () => {
-            if (date === dateToday) {
-              return Streak.Pending;
-            }
+  const streaks = useMemo(() => {
+    return streakDays.map((value) => {
+      const day = value.getDay();
+      const date = value.getDate();
```
kopancek, line 42: Fun with timezones never ends. By default, the browser uses the timezone setting from the OS. This can potentially be different from the timezone the user has configured on their user profile.

I think we should be consistent with our calculation and need to convert all of these dates to the timezone the user specified on their user profile. [Since the calculation on the graphql endpoint is already doing that as well (see the `timestampAtTimezone`)](https://github.com/dailydotdev/daily-api/blob/mi-71-user-streak-counter/src/schema/users.ts#L843)

packages/shared/src/graphql/users.ts
kopancek: Plus one on the above.

packages/shared/src/graphql/users.ts
```diff
@@ -423,3 +434,46 @@ export const UNSUBSCRIBE_PERSONALIZED_DIGEST_MUTATION = gql`
...
+  const res = await request(graphqlUrl, USER_STREAK_HISTORY, {
+    after: last30Days.toISOString(),
+    before: today,
```
kopancek, line 450: Since `after`  is given as an ISO string - so basically in UTC, I think we should do the same here
```suggestion
    before: today.toISOString(),
```

kopancek (changes requested):
Great progress, I think we are almost there.

There is another problem with timezones I think. The GraphQL endpoint for getting the userReadHistory uses the timezone of the user on the user profile, but the popup is using the timezone from the browser/OS, which might lead to unexpected results in case those are different.

Other comments from me are non-blocking.

packages/shared/src/components/streak/popup/ReadingStreakPopup.tsx
sshanzel: 6 and 0 is Saturday and Sunday. Unfortunately, Sunday starts at `0` in JS 😅

packages/shared/src/graphql/users.ts
sshanzel: I was thinking of even just the last 14 days. But as per RFC we will pull 30 days.

packages/shared/src/components/streak/popup/ReadingStreakPopup.tsx
sshanzel: > Maybe we can add const or enum values for these here as well, like you did in the other PR @sshanzel ?

Yeah, makes sense. Let's introduce it here as well.

packages/shared/src/components/streak/popup/ReadingStreakPopup.tsx
sshanzel: The values being processed here are all the result of the BE's response. FE-wise, the checks here should be good to go.

packages/shared/src/graphql/users.ts
sshanzel: Introduced a second parameter to decide when should the query start.

packages/shared/src/graphql/users.ts
sshanzel: Right. Totally missed that. Thanks! 🚀

packages/shared/src/components/streak/popup/ReadingStreakPopup.tsx
kopancek: Let's try it out with different timezones in the future and see what happens 👍

PR #2565, opened 2024-01-29
feat: add pixel & gtm scripts

DragosIorgulescu:
- possible naive approach, there may be cleaner ways to do this
- I did not want to install third-party libraries for these, although there are some out there if that would be the easier route

TODO:
- [ ] trigger the signup events when the sign up button is clicked
- [ ] check if there is a possibly cleaner way of adding these (without requiring too much work)

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done

### DIFF
--- a/packages/webapp/pages/onboarding.tsx
+++ b/packages/webapp/pages/onboarding.tsx
@@ -314,6 +314,73 @@ export function OnboardPage(): ReactElement {
     return <ProgressBar percentage={isAuthenticating ? percentage : 0} />;
   };
 
+  const getPixelTracking = () => {
+    return (
+      <>
+        <script
+          dangerouslySetInnerHTML={{
+            __html: `
+            !function(f,b,e,v,n,t,s)
+            {if(f.fbq)return;n=f.fbq=function(){n.callMethod?
+              n.callMethod.apply(n,arguments):n.queue.push(arguments)};
+              if(!f._fbq)f._fbq=n;n.push=n;n.loaded=!0;n.version='2.0';
+              n.queue=[];t=b.createElement(e);t.async=!0;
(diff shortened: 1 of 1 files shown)


sshanzel:
@rebelchris @nensidosari @DragosIorgulescu applied the tracking based on [NextJS example](https://github.com/vercel/next.js/tree/canary/examples/with-facebook-pixel) on how to implement these kinds of scripts/tracking.

We now track on click of any provider. Just wanted to check if that was the desired behavior?

image


cc @idoshamun

capJavert (approved):
Looking good, nice usage of `Script` 🤘

idoshamun (approved):
Looks great! Also happy with the fact we track all the buttons

nensidosari (approved):
Looks great.

nensidosari:
As on verifying that the events are being sent, I would say just checking network 🤷‍♀️ unless we have any developer gtag id we can use to check?

capJavert (commented):
Now that @nensidosari said it, I got remineded. I think pixel/gtag should not even load in development so we might want to add `isDev` checks and not render `Script` there?

sshanzel:
Good point. I can do an early return when it is in Dev mode.

@nensidosari I tried checking the network, however, nothing seems to be fired even though the functions are executed.

sshanzel:
@nensidosari @capJavert added a check for if it is not production to return null so that we don't have to worry about the test environment as well.

sshanzel:
Events are fired from the network tab 🥳 🚀

PR #2573, opened 2024-01-31
feat: adjust search page and empty state

capJavert:
- Adjustments to empty state of the new search page
- Now by default the `/search` page does not show anything until you search
- Also made adjustments to no results page from AS-40 since it was connected closely

TODO:
- [ ] just missing gradient at the bottom on mobile

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

AS-69 #done

### DIFF
--- a/packages/shared/src/components/SearchEmptyScreen.tsx
+++ b/packages/shared/src/components/SearchEmptyScreen.tsx
@@ -6,7 +6,7 @@ export default function SearchEmptyScreen(): ReactElement {
   return (
     <div className="flex w-full max-w-[32rem] flex-col items-center gap-4 self-center px-6">
       <MagnifyingIcon
-        className="text-overlay-tertiary-salt"
+        className="text-theme-label-disabled"
         size={IconSize.XXXLarge}
       />
       <h2 className="text-center typo-title2">No results found</h2>
--- a/packages/shared/src/components/SearchEmptyScreen.tsx
+++ b/packages/shared/src/components/SearchEmptyScreen.tsx
@@ -1,19 +1,16 @@
 import React, { ReactElement } from 'react';
 import { SearchIcon as MagnifyingIcon } from './icons';
-import { EmptyScreenIcon } from './EmptyScreen';
+import { IconSize } from './Icon';
 
 export default function SearchEmptyScreen(): ReactElement {
   return (
-    <div
-      className="flex w-full flex-col self-center px-6"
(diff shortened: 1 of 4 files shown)


packages/shared/src/components/SearchEmptyScreen.tsx
```diff
@@ -1,19 +1,16 @@
...
   return (
-    <div
-      className="flex w-full flex-col self-center px-6"
-      style={{ maxWidth: '32.5rem' }}
-    >
+    <div className="flex w-full max-w-[32rem] flex-col items-center gap-4 self-center px-6">
       <MagnifyingIcon
-        className="self-center text-theme-label-disabled"
-        style={EmptyScreenIcon.style}
+        className="text-overlay-tertiary-salt"
```
rebelchris, line 9: Was it not text-theme-label-disabled?
It seems it's named that way in figma (maybe we have a mismatch here naming wise, or they changed it)

packages/webapp/components/search/SearchControlPage.tsx
```diff
@@ -36,6 +42,26 @@ const Search = (): ReactElement => {
...
       <NextSeo {...seo} {...baseSeo} />
+      {!searchQuery && (
+        <div
+          className={classNames(
+            'flex w-full max-w-[32rem] flex-col items-center gap-4  self-center px-6',
+            sidebarExpanded ? '!-left-32' : '!-left-6',
+          )}
+        >
+          <MagicIcon
+            className="text-overlay-tertiary-salt"
```
rebelchris, line 53: Same here I guess

rebelchris (approved):
Non breaking, but can you maybe double check with design if they changed whatever we have as disabled now?

Why:
I think we use it in more places, so maybe they changed the guidelines on it, thus be good for us to address the disabled text to represent more closely what figma has.

packages/shared/src/components/SearchEmptyScreen.tsx
capJavert: Basically they are both `salt.90` with `32%` opacity, I'll update with disabled. Just the matter of colors being confusing to find I guess. Good catch 👍

capJavert: Updated 3c2008e164e663dad4bf68ad28646486eb8965b2

PR #2595, opened 2024-02-08
feat: customize tab and downloading image

sshanzel:
## Changes
- Customize the DevCard itself.

Preview:

Note: it was intentionally made to enable all options just for preview.

https://github.com/dailydotdev/apps/assets/13744167/794362c6-76d6-440e-ae25-1815dc9f9e35

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done

### DIFF
--- a/packages/webapp/pages/devcard.tsx
+++ b/packages/webapp/pages/devcard.tsx
@@ -160,7 +160,7 @@ const Step2 = ({
   return (
     <div className="mx-2 mt-14 flex flex-col self-stretch">
       <main className="z-2 flex flex-col gap-10 laptop:flex-row laptopL:gap-20">
-        <section className="align-center flex max-w-[600px] flex-col">
+        <section className="align-center flex flex-col">
           <h1 className="mx-3 mb-8 text-center font-bold typo-title1">
             Share your #DevCard!
           </h1>
--- a/packages/webapp/pages/devcard.tsx
+++ b/packages/webapp/pages/devcard.tsx
@@ -29,6 +29,7 @@ import { AuthTriggers } from '@dailydotdev/shared/src/lib/auth';
 import { labels } from '@dailydotdev/shared/src/lib';
 import {
   DevCard,
+  DevCardTheme,
   DevCardType,
   themeToLinearGradient,
 } from '@dailydotdev/shared/src/components/profile/devcard';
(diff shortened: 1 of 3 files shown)


packages/webapp/pages/devcard.tsx
```diff
@@ -347,59 +396,18 @@ const seo: NextSeoProps = {
...
-    },
-  );
-
   const stepProps: StepProps = {
-    onGenerateImage: generateDevCard,
-    devCardSrc,
-    isLoadingImage,
-    error: imageError,
+    onGenerateImage: () => {
+      setShouldShowCard(true);
```
nensidosari, line 405: Why are we using both these checks?

packages/webapp/pages/devcard.tsx
```diff
@@ -409,7 +417,11 @@ const DevCardPage = (): ReactElement => {
       )}
     >
       <NextSeo {...seo} />
-      {isDevCardGenerated ? <Step2 {...stepProps} /> : <Step1 {...stepProps} />}
+      {isDevCardGenerated || shouldShowCard ? (
```
nensidosari, line 420: Related to above comment

nensidosari (commented):
All looks great. Have just one question on some checks being used to render which step of the devcard. Once that is cleared, happy to approve 🚀

packages/webapp/pages/devcard.tsx
sshanzel: The first one is just to switch to the other screen since as per design we have this first "Generate devcard" screen.

sshanzel: The second one, I can't quite remember but I believe it is supposed to be some sort of highlighting pulse to some element relate

sshanzel: After checking the codebase, it seems we only use it here, this would make the screen stop showing the first page once the user has seen it already.

sshanzel: We might actually not need the state in this case.

nensidosari: Yup, the isDevCardGenerated is related to the action we set when user has seen the generate page, so I don't think we need `shouldShowCard` this state. Let's remove

nensidosari (approved):
Looks great

PR #2596, opened 2024-02-08
feat: email verification

rebelchris:
Introduces email verification


packages/shared/src/components/auth/CodeVerificationForm.tsx
```diff
@@ -31,7 +31,7 @@ function CodeVerificationForm({
   const [emailSent, setEmailSent] = useState(false);
   const { sendEmail, verifyCode, resendTimer, isLoading } = useAccountEmailFlow(
     {
-      flow: AuthFlow.Recovery,
+      flow: AuthFlow.Verification,
```
nensidosari, line 34: Just checking that we don't wan't these other events anymore but are replacing these with these new ones.

packages/shared/src/components/auth/EmailCodeVerification.tsx
```diff
@@ -0,0 +1,126 @@
...
+function EmailCodeVerification({
+  code: codeProp,
+  email: emailProp,
+  flowId,
+  onSubmit,
+  className,
+}: EmailCodeVerificationProps): ReactElement {
+  const { trackEvent } = useContext(AnalyticsContext);
+  const [hint, setHint] = useState('');
+  const [alert, setAlert] = useState({ firstAlert: true, alert: false });
```
nensidosari, line 30: I find these alert and firstAlert names really confusing to understand what they are about 😅

packages/shared/src/components/auth/EmailCodeVerification.tsx
```diff
@@ -0,0 +1,126 @@
...
+      flow: AuthFlow.Verification,
+      flowId,
+      onError: setHint,
+      onVerifyCodeSuccess: () => {
+        onSubmit();
+      },
+    });
+
+  useEffect(() => {
+    if (autoResend && !alert.alert && alert.firstAlert === true) {
```
nensidosari, line 44: If i understood this right, firstAlert seems to be to keep this running only once. I think it's better if alert can have states maybe:

Maybe is better to leave it as null or undefined and look specifically for undefined type in the first try, rather then a boolean value.

nensidosari (approved):
Looks all good to me. Left 1 comment to check and another non blocking, but that I think might make code more understandable.

packages/shared/src/components/auth/EmailCodeVerification.tsx
rebelchris: Ah yes, totally get your message, I initially wanted to solve it with one state, but because everything is so intertwined from propdrilling this became super messy and introduced frontend bugs.
I'll see if I can quickly make it a bit easier.

packages/shared/src/components/auth/CodeVerificationForm.tsx
rebelchris: Thanks for raising this, checked with Dave and we decided to add some more/different ones.

PR #2603, opened 2024-02-12
feat: download twitter cover

sshanzel:
## Changes
- Use the same mutation to generate the devcard

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done

### DIFF
--- a/packages/shared/src/components/profile/devcard/DevCard.tsx
+++ b/packages/shared/src/components/profile/devcard/DevCard.tsx
@@ -9,6 +9,7 @@ import { devCardBoxShadow, DevCardTheme, DevCardType } from './common';
 import { DevCardFooter } from './DevCardFooter';
 import { DevCardContainer } from './DevCardContainer';
 import { useDevCard } from '../../../hooks/profile/useDevCard';
+import { DevCardTwitterCover } from './DevCardTwitterCover';
 
 interface DevCardProps {
   type?: DevCardType;
@@ -19,10 +20,15 @@ export function DevCard({
   type = DevCardType.Vertical,
   userId,
 }: DevCardProps): ReactElement {
-  const { devcard, isLoading, coverImage } = useDevCard(userId);
+  const data = useDevCard(userId);
+  const { devcard, isLoading, coverImage } = data ?? {};
 
   if (isLoading || !devcard) {
     return null;
+  }
+
+  if (type === DevCardType.Twitter) {
+    return <DevCardTwitterCover data={data} />;
   }
 
   const { theme, user, articlesRead, tags, sources, showBorder } = devcard;
--- a/packages/shared/src/components/profile/devcard/DevCard.tsx
+++ b/packages/shared/src/components/profile/devcard/DevCard.tsx
@@ -28,7 +28,7 @@ export function DevCard({
   }
 
   if (type === DevCardType.Twitter) {
-    return <DevCardTwitterCover data={data} />;
+    return <DevCardTwitterCover {...data} />;
   }
 
   const { theme, user, articlesRead, tags, sources, showBorder } = devcard;
--- a/packages/shared/src/components/profile/devcard/DevCardTwitterCover.tsx
+++ b/packages/shared/src/components/profile/devcard/DevCardTwitterCover.tsx
@@ -6,15 +6,10 @@ import { DevCardFooter } from './DevCardFooter';
 import Logo from '../../Logo';
 import { UseDevCard } from '../../../hooks/profile/useDevCard';
 
-interface DevCardTwitterCoverProps {
-  data: UseDevCard;
-}
-
 export function DevCardTwitterCover({
-  data,
-}: DevCardTwitterCoverProps): ReactElement {
-  const { devcard, coverImage } = data ?? {};
-
+  devcard,
+  coverImage,
+}: Pick<UseDevCard, 'devcard' | 'coverImage'>): ReactElement {
   if (!devcard) {
     return null;
   }
--- a/packages/shared/src/components/profile/devcard/DevCardTwitterCover.tsx
+++ b/packages/shared/src/components/profile/devcard/DevCardTwitterCover.tsx
@@ -4,16 +4,16 @@ import { devCardBoxShadow, DevCardTheme, DevCardType } from './common';
 import { DevCardStats } from './DevCardStats';
 import { DevCardFooter } from './DevCardFooter';
 import Logo from '../../Logo';
-import { useDevCard } from '../../../hooks/profile/useDevCard';
+import { UseDevCard } from '../../../hooks/profile/useDevCard';
 
 interface DevCardTwitterCoverProps {
-  userId: string;
+  data: UseDevCard;
 }
 
 export function DevCardTwitterCover({
-  userId,
+  data,
 }: DevCardTwitterCoverProps): ReactElement {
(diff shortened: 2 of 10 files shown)


packages/shared/src/hooks/utils/useDownloadUrl.ts
```diff
@@ -0,0 +1,35 @@
...
+interface UseDownloadUrl {
+  onDownloadUrl: (props: DownloadProps) => Promise<void>;
+}
+
+interface UseDownloadUrlProps {
+  onError?: (error: Error) => void;
+}
+
+export const useDownloadUrl = ({
+  onError,
```
capJavert, line 15: I would move this to `onDownloadUrl` function call instead since this is where handling of error actually happens.

packages/shared/src/hooks/utils/useDownloadUrl.ts
```diff
@@ -0,0 +1,35 @@
...
+}: UseDownloadUrlProps = {}): UseDownloadUrl => {
+  const onDownloadUrl = async ({ filename, url }: DownloadProps) => {
```
capJavert, line 17: memoize with useCallback

packages/shared/src/hooks/utils/useDownloadUrl.ts
capJavert: Thinking about it more, it might even be better to just do the handling with try catch in place you call `onDownloadUrl`. Since onError is optional errors would be swallowed like this which is bad pattern.

packages/shared/src/hooks/utils/useDownloadUrl.ts
capJavert: This could also be react-query mutation where you would get loading, error and all other states for free which might be even better?

packages/webapp/pages/devcard.tsx
```diff
@@ -277,6 +272,15 @@ const Step2 = (): ReactElement => {
...
                   Level up your Twitter game with a DevCard header image!
                 </p>
+                <Button
+                  className="mt-5"
+                  variant={ButtonVariant.Secondary}
+                  size={ButtonSize.Small}
+                  onClick={() => onGenerate({ type: DevCardType.Twitter })}
+                  loading={downloading || isLoading}
+                >
+                  Download X cover image
```
omBratteng, line 282: Really hate the name `X`...

packages/shared/src/hooks/utils/useDownloadUrl.ts
capJavert: Yeah, that is why I think this actually might be better https://github.com/dailydotdev/apps/pull/2603#discussion_r1486076862

packages/shared/src/hooks/utils/useDownloadUrl.ts
sshanzel: Totally agree with that. Didnt realize it at first as i just copy pasted this part.

packages/shared/src/components/profile/devcard/DevCardTwitterCover.tsx
```diff
@@ -4,16 +4,16 @@ import { devCardBoxShadow, DevCardTheme, DevCardType } from './common';
 import { DevCardStats } from './DevCardStats';
 import { DevCardFooter } from './DevCardFooter';
 import Logo from '../../Logo';
-import { useDevCard } from '../../../hooks/profile/useDevCard';
+import { UseDevCard } from '../../../hooks/profile/useDevCard';
 
 interface DevCardTwitterCoverProps {
-  userId: string;
+  data: UseDevCard;
```
kopancek, line 10: We only use `devcard` and `coverImage` from the `UseDevCard` type. Maybe it's worth narrowing down to those. See my comment above. Here we could do:
```suggestion
  devcard?: DevCardData;
  coverImage?: string;
```

packages/shared/src/hooks/profile/useDevCard.ts
```diff
@@ -21,7 +21,7 @@ export interface DevCardData {
   sources: Source[];
 }
 
-interface UseDevCard {
+export interface UseDevCard {
```
kopancek, line 24: If you decide to follow my suggestion above, we no longer need to export this type.

kopancek (commented):
Looks great. Added some non-blocking comments for what I think would be a better developer experience.

packages/shared/src/hooks/utils/useDownloadUrl.ts
sshanzel: Yep. Let's go for that.

packages/shared/src/components/profile/devcard/DevCardTwitterCover.tsx
sshanzel: Yes, that's the only thing that can be pulled from that object. So essentially, they are the same. The primary reason for doing this was that the component initially used the hook in itself.

packages/shared/src/components/profile/devcard/DevCard.tsx
```diff
@@ -19,12 +20,17 @@ export function DevCard({
...
-  const { devcard, isLoading, coverImage } = useDevCard(userId);
+  const data = useDevCard(userId);
+  const { devcard, isLoading, coverImage } = data ?? {};
 
   if (isLoading || !devcard) {
     return null;
   }
 
+  if (type === DevCardType.Twitter) {
+    return <DevCardTwitterCover data={data} />;
```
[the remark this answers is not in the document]
sshanzel: Yeah, to be honest, I'm not a fan of passing a whole object of hooks, etc.

packages/shared/src/hooks/utils/useDownloadUrl.ts
```diff
@@ -0,0 +1,22 @@
+import { useMutation, MutationOptions } from '@tanstack/react-query';
```
sshanzel, line 1: @capJavert do you think we still need this hook? Or we can just define the reusable function and then do `useMutation` on-demand?

packages/shared/src/components/profile/devcard/DevCardTwitterCover.tsx
sshanzel: Refactored here https://github.com/dailydotdev/apps/pull/2603/commits/532f2c8a2fffe4889f4c537cd0ae7ea711c5b835

packages/shared/src/hooks/profile/useDevCard.ts
sshanzel: I think we can keep this, I used `Pick` on this type for the props.

nensidosari (approved):
LGTM 🚀

packages/shared/src/hooks/utils/useDownloadUrl.ts
capJavert: Since does not need anything from context/react scope that would require a hook, yes, having it as function sounds good as well 👍

sshanzel: Made the changes here https://github.com/dailydotdev/apps/pull/2613/commits/3c779a73e5c04783c7b6e0c4d6f6f454b1a5ba50

PR #2608, opened 2024-02-12
fix: vertical align of tags + border width of profile image

kopancek:
The tags are now vertically aligned
Profile picture border is set to 8px in all cases, not just for horizontally laid out card

Screenshot 2024-02-12 at 15 50 59

### DIFF
--- a/packages/shared/src/components/profile/devcard/DevCard.tsx
+++ b/packages/shared/src/components/profile/devcard/DevCard.tsx
@@ -85,7 +85,7 @@ export function DevCard({
               alt={`avatar of ${user.name}`}
               className={classNames(
                 '-rotate-3 border-white object-cover',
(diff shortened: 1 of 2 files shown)


nensidosari (approved):

packages/shared/src/components/profile/devcard/DevCard.tsx
```diff
@@ -85,7 +85,7 @@ export function DevCard({
               alt={`avatar of ${user.name}`}
               className={classNames(
                 '-rotate-3 border-white object-cover',
-                showBorder && (isHorizontal ? 'border-8' : 'border-4'),
```
sshanzel, line 88: Not sure why we removed this. They have different size of borders from the design.

PR #2611, opened 2024-02-12
feat: sidebar overflows search panel dropdown

capJavert:
- See [discussion](https://dailydotdev.slack.com/archives/C01L0QXQJNB/p1707437951151969)
- make header above sidebar so that panel dropdown is also above
- pad our custom z-index values so we can easily add other values between
  - also scaled other values in css to match padded values
- available for test on [preview](https://preview3.app.daily.dev)

TODO
- [x] test on preview for different tooltips/modals

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

AS-107 #done


packages/shared/tailwind.config.js
```diff
@@ -127,14 +127,15 @@ module.exports = {
...
-      max: '100',
+      rank: '30',
+      header: '75',
+      postNavigation: '76',
+      sidebar: '70',
+      tooltip: '80',
+      popup: '90',
+      modal: '100',
+      cookie: '2000',
+      max: '1000',
```
rebelchris, line 138: Max is lower then cookie?

packages/shared/src/styles/components/menu.css
```diff
@@ -2,7 +2,7 @@
   position: fixed;
   opacity: 0;
   user-select: none;
-  z-index: 100;
+  z-index: 1000;
```
rebelchris, line 5: Shouldnt we move these to named?

rebelchris: Also above 10 and 100 both resolve to 1000 now?

rebelchris (changes requested):
Minor comments

packages/shared/tailwind.config.js
capJavert: It was before as well, it was `z-[200]` vs z-100 that was previous max.

packages/shared/src/styles/components/menu.css
capJavert: I am not sure what else I need to change to move them, so just updated for now. 

> Also above 10 and 100 both resolve to 1000 now?

not sure what you mean by this?

packages/shared/tailwind.config.js
capJavert: Last cookie value comes from here https://github.com/dailydotdev/apps/pull/1936. From What I see `z-max` should be ok as well so we can change to that.

packages/shared/src/styles/components/menu.css
rebelchris: Got it on the changes.
As for the other point, never mind my phone version showed one under 10 and one 100

packages/shared/tailwind.config.js
rebelchris: Hmm yeah let's maybe try that and see if it works?

rebelchris (approved):

packages/shared/src/styles/components/menu.css
capJavert: Yeah, all are basically multiplied by 10, so 10 -> 100, 100 -> 1000.

packages/shared/tailwind.config.js
```diff
@@ -127,14 +127,15 @@ module.exports = {
...
-      sidebar: '7',
-      tooltip: '8',
-      popup: '9',
-      modal: '10',
-      max: '100',
+      rank: '30',
+      header: '75',
+      postNavigation: '76',
+      sidebar: '70',
+      tooltip: '80',
```
sshanzel, line 134: Tooltip should probably be higher? Like hovering a user on post modal. Or would it work ootb since it is inside the modal?

packages/shared/tailwind.config.js
capJavert: I tested already locally, it works, only things that have the z-max as well and could interfeer with banner are:
- toast
- context menu 

since those are temp and parts of user interaction it makes sense they would be on the same level, it is not like they obscure premaritally.

packages/shared/tailwind.config.js
capJavert: Only thing that uses `z-tooltip` is this thing:
image

everything else is tippy or contexify which have their own z-index to 1000 or 9999.

sshanzel: Ahh, nice. Thanks for the verification.

PR #2616, opened 2024-02-13
MI-151-share-customize-page

kopancek:
All the changes for share and customize page go here


packages/shared/src/lib/strings.ts
```diff
@@ -8,3 +8,6 @@ export const capitalize = (value: string): string =>
   (value && value[0].toUpperCase() + value.slice(1)) || '';
 
 export const anchorDefaultRel = 'noopener noreferrer';
+
+export const checkStringEquality = (value1: string, value2: string): boolean =>
```
rebelchris, line 12: @sshanzel you might know, did we not already have such function somewhere?

rebelchris (approved):
Nothing blocking from my side, just minor questions

packages/shared/src/lib/strings.ts
capJavert: Do we really need a function for this? Just use `===`

sshanzel: Can't exactly remember, though using `===`, it won't work when the other one is on uppercase and the other is on lowercase.

capJavert: Yeah in those cases you just do `toLowerCase()` inline where you need it, with above we are hiding implementation without any obvious reasons right?

sshanzel: I think it's just a personal preference in terms of "code cleanliness", there was a lot of it in a single file which felt it was all over the place.

sshanzel: We were comparing two values, one is probably lowercase already, but can't exactly remember which one, so in order to avoid testing whether we need lowercasing the other or not, we just do it both here.

capJavert: The function name could include "lowercase" to reflect what it does since implementation is hidden.

packages/shared/src/components/profile/devcard/DevCard.tsx
```diff
@@ -113,41 +133,71 @@ export function DevCard({
...
+                @{user.username}
+              </span>
               <Separator />
               <time
                 className={classNames(
-                  'text-opacity-64 typo-caption1',
-                  isIron ? 'text-white' : 'text-pepper-90',
+                  'typo-caption1',
+                  isIron && 'text-white',
+                  isDefault && 'text-pepper-10/[.64]',
```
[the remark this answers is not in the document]
sshanzel: Just checked the overlay, I think it has a value of 90.

capJavert (approved):

PR #2621, opened 2024-02-13
feat: optimize post pages

capJavert:
- optimize rendering on pages with many comments
- the main bottleneck was the actual number of nodes render in the DOM and JavaScript main thread work due to React rendering markdown and other nodes

Main penalties:
image
image

Fixed by lazy rendering comments that are not in view. One issue was that we also have comment hash links and in those cases the lazy rendering is disabled for now. This will allow us to not be penalized for SEO and users can still use comment deep links.  

Everything else should work the same since `useInView` hook is configured to trigger only once so once render comment will stay render which allows editing any all other actions. 

Before (desktop):
image
After (desktop):
image

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

MI-164 #done

### DIFF
--- a/packages/shared/src/components/comments/MainComment.tsx
+++ b/packages/shared/src/components/comments/MainComment.tsx
@@ -1,5 +1,6 @@
 import React, { ReactElement, useContext, useMemo } from 'react';
 import classNames from 'classnames';
+import { useInView } from 'react-intersection-observer';
 import EnableNotification from '../notifications/EnableNotification';
 import CommentBox, { CommentBoxProps } from './CommentBox';
 import SubComment from './SubComment';
@@ -69,8 +70,14 @@ export default function MainComment({
   const { commentId, inputProps, onReplyTo } = useComments(props.post);
   const { inputProps: editProps, onEdit } = useCommentEdit();
 
+  const { ref: inViewRef, inView } = useInView({
+    triggerOnce: true,
+    initialInView: position < lazyCommentThreshold,
+  });
+
   return (
     <section
+      ref={inViewRef}
       className={classNames(
         'flex scroll-mt-16 flex-col items-stretch rounded-24 border border-theme-divider-tertiary',
         className?.container,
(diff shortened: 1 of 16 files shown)


packages/shared/src/components/comments/MainComment.tsx
```diff
@@ -66,15 +69,25 @@ export default function MainComment({
...
     <section
+      ref={inViewRef}
       className={classNames(
         'flex scroll-mt-16 flex-col items-stretch rounded-24 border border-theme-divider-tertiary',
         className?.container,
       )}
       data-testid="comment"
+      style={{
+        contentVisibility: initialInView ? 'visible' : 'auto',
+      }}
```
capJavert, line 88: This additionally allows better scroll performance since views not visible will not be rendered.

rebelchris: sadness tailwind doesn't support it yet.
https://github.com/tailwindlabs/tailwindcss/discussions/2162

But yeah nice thinking, totally didn't think about this.

rebelchris (approved):
Love it, nothing to add from my side :) 
@idoshamun if you don't see anything blocking here, let's give it a try. (maybe you know some downsides to this method)

nensidosari (approved):
Looks great on my side as well!! Thank u!!

idoshamun (approved):
Wow that's a massive uplift!!! My only concern is would google bot will be able to use all the comments for SEO?

idoshamun:
@capJavert @rebelchris @nensidosari out of scope but we should consider pagination at some point as well

capJavert:
@idoshamun we can play with lazy load threshold, now it is 5 comments but it would affect performance again. I think in order for google better indexing comments it would be better to maybe add https://schema.org/Comment in the future as we do for base post?

idoshamun:
@capJavert yap we can do that. it's ok for now. also something to consider down the line is finding a more performant purification library, out of scope for now

PR #2626, opened 2024-02-14
fix: og image url random parameter

kopancek:
## Changes

- Added random parameter to og image URL, so that when the image is changed, we get around the caching
- Use the same random parameter generation as we do in the API side.

As before, we can only test this in preview.

### DIFF
--- a/packages/webapp/components/layouts/ProfileLayout/index.tsx
+++ b/packages/webapp/components/layouts/ProfileLayout/index.tsx
@@ -57,6 +57,7 @@ export default function ProfileLayout({
     process.env.NEXT_PUBLIC_API_URL,
   );
   ogImageUrl.searchParams.set('type', 'wide');
+  ogImageUrl.searchParams.set('r', Math.random().toString(36).substring(2, 5));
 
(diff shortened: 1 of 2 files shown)


packages/webapp/components/layouts/ProfileLayout/index.tsx
```diff
@@ -57,6 +57,7 @@ export default function ProfileLayout({
     process.env.NEXT_PUBLIC_API_URL,
   );
   ogImageUrl.searchParams.set('type', 'wide');
```
sshanzel, line 59: Not part of this PR but wanted to ask, are we only going to use the wide version for OG?

sshanzel (approved):
Looks good to me 🚀

packages/webapp/components/layouts/ProfileLayout/index.tsx
kopancek: As far as I understand, yes: https://www.figma.com/file/nmfWPS7x3kzLUvYMkBx2kW/daily.dev---Dev-Mode?node-id=4234%3A17279&mode=dev

PR #2630, opened 2024-02-14
feat: onboarding v4.5

omBratteng:


packages/webapp/next.config.js
```diff
@@ -91,6 +91,11 @@ module.exports = withTM(
               source: '/posts/finder',
               destination: '/search?provider=posts',
               permanent: false
+            },
+            {
+              source: '/signup',
+              destination: '/onboarding',
+              permanent: true
```
omBratteng, line 98:
```suggestion
              permanent: false
```

packages/shared/src/components/auth/AuthOptions.tsx
```diff
@@ -372,7 +375,7 @@ function AuthOptions({
             onPasswordLogin={onPasswordLogin}
             loginHint={loginHint}
             isLoading={isPasswordLoginLoading}
-            isLoginFlow={isForgotPasswordReturn || isLoginFlow}
+            isLoginFlow={isForgotPasswordReturn || isLoginFlow || isLogin}
```
nensidosari, line 378: What is the difference between isLoginFlow and isLogin? I think it's quite confusing to understand

packages/shared/src/components/onboarding/OnboardingFeedHeader.tsx
```diff
@@ -0,0 +1,154 @@
...
+    updateAlerts({ filter: false, myFeed: 'created' });
+
+    trackEvent({
+      event_name: AnalyticsEvent.CreateFeed,
+    });
+
+    return router.replace({
+      pathname: router.route,
+      query: {
+        welcome: 'true',
```
nensidosari, line 68: We're removing this in another onboarding experiment, but since this is now added in a different place, not sure if we would want that experiment available here as well.

packages/shared/src/lib/featureValues.ts
```diff
@@ -25,3 +25,11 @@ export enum PostPageOnboarding {
   Control = 'control',
   V1 = 'v1',
 }
+
+export enum OnboardingV4dot5 {
+  Control = 'control',
+  V4dot5 = 'v4dot5',
+}
+
+export const isOnboardingV4dot5 = (flag: string): boolean =>
```
nensidosari, line 34: Do we want to add this here or in any utils? Seems like it messes one of the only files we have structured 😅

nensidosari (commented):
Looks good overall, few comments to check 🙌

packages/shared/src/components/auth/AuthOptions.tsx
omBratteng: Discussed this here https://github.com/dailydotdev/apps/pull/2665#discussion_r1499212196

> `isLoginFlow` is controlled via the `loginState` in `AuthContext`, so I would have to import it to update that state
https://github.com/dailydotdev/apps/blob/4585cc58ef9fb3c936139394ef9ba9486008dc60/packages/shared/src/components/auth/AuthModal.tsx#L72C22-L72C34

packages/shared/src/lib/featureValues.ts
omBratteng: I know it's not the best place. This was one of the places we didn't have a common file 😬 Maybe move it into the neighboring `func.ts` file?

packages/shared/src/components/onboarding/OnboardingFeedHeader.tsx
omBratteng: Not sure, maybe product can answer?

packages/shared/src/lib/featureValues.ts
capJavert: Do you really need this though, you can just use `===`? I don't think this helper adds anything.

omBratteng: Just DX I guess, quicker to auto complete on isOnboardingV4dot5 than writing the === check. I'm indifferent really.

packages/shared/src/components/auth/AuthOptions.tsx
omBratteng: The entire auth is in need of some overhaul to cleanup the mishmash of solutions.

packages/shared/src/components/onboarding/FeedReadyMessage.tsx
```diff
@@ -15,6 +16,11 @@ export const FeedReadyMessage = ({
   className,
 }: FeedReadyMessageProps): ReactElement => {
   const { user } = useContext(AuthContext);
+  const { alerts } = useAlertsContext();
+
+  if (alerts?.filter) {
+    return null;
+  }
```
capJavert, line 23: Should this be under feature flag, does it apply for control?

omBratteng: Was thinking to apply it to everything. In case you somehow stumble upon https://app.daily.dev/?welcome=true if you have an account, but haven't created your feed, you shouldn't really see the "your feed is ready message". And it probably shouldn't be possible to view it on the other feeds as well 😈

- https://app.daily.dev/popular?welcome=true
- https://app.daily.dev/squads/watercooler?welcome=true

packages/shared/src/lib/featureValues.ts
nensidosari: I'd say we either move or do what Ante says it's what we commonly do ATM.

capJavert: Hmm for me DX would be to then have the `useFeature` and other logic wrapped so you don't need to import `features`, `useFeature`, your helper and then use it. Eg. how we have `useFeatureFeedLayoutV1` or `useStreakExperiment`.

At this point this just adds another function call to the stack. So I would personally prefer not to add anything but if you want the hook would be better option I guess.

omBratteng: I'll remove it.

packages/shared/src/components/onboarding/FeedReadyMessage.tsx
capJavert: Ah gotcha, makes sense. One edge case comes to mind though. Do we set `alerts.filter` in onboarding so then it might happen that we pass `welcome=true` but because you already set `alerts.filter` it would not show? That would happen in control I guess.

capJavert: So just double check if you can 🙏

omBratteng: I think that setting the `alerts.filter` happens on account creation. And that alert should be removed when you choose a tag (be it onboarding or anywhere else).
https://github.com/dailydotdev/apps/blob/2f390ae941d70f31ddf5870cd16c5f248e4d86b3/packages/shared/src/hooks/useTagAndSource.ts#L88-L90

I just added so that in this experiment, we don't clear the filter alerts until the click "create feed".

Does that answer your edge case?

packages/shared/src/components/onboarding/OnboardingFeedHeader.tsx
omBratteng: Will leave as is in this experiment.

packages/shared/src/components/auth/AuthOptions.tsx
nensidosari: Yup agree, anyway let's leave that for another time.

omBratteng: Another time™

rebelchris: Yeah there's too many abuse in the whole auth.
Also noticed this morning we actually have 3x the code verification components for different placements

capJavert (approved):

packages/shared/src/components/onboarding/OnboardingFeedHeader.tsx
```diff
@@ -0,0 +1,154 @@
...
+        return;
+      }
+      router.events.emit('routeChangeError');
+
+      showPrompt(promptConfig).then((leave) => {
+        if (leave) {
+          stopNav = false;
+          trackEvent({
+            event_name: AnalyticsEvent.OnboardingSkip,
+          });
```
rebelchris, line 87: Not sure if this is right here?
Was it checked with Dave/Ido?

I think we might want one, but a new one?

rebelchris (changes requested):
One small tracking concern potentially.

Also would you mind checking locally this prompts the code verification flow correct for emails? (as it's newly merged)

packages/shared/src/components/onboarding/OnboardingFeedHeader.tsx
omBratteng: I'll check with Dave

omBratteng: Will check with Dave

omBratteng: We'll re-use

rebelchris (approved):

nensidosari (approved):

omBratteng:
Tested with email auth code flow, and it works.

PR #2633, opened 2024-02-14
fix: missing download devcard button

kopancek:
## Changes

~~Ugly fix~~ Thanks to @capJavert now a nice fix to always make sure we have devcard data defined, regardless of the order of execution.

### DIFF
--- a/packages/shared/src/hooks/profile/useDevCard.ts
+++ b/packages/shared/src/hooks/profile/useDevCard.ts
@@ -27,10 +27,7 @@ export interface UseDevCard {
   coverImage: string;
 }
 
-export const useDevCard = (
-  userId: string,
-  onSuccess?: (data: DevCardData) => void,
-): UseDevCard => {
+export const useDevCard = (userId: string): UseDevCard => {
   const { requestMethod } = useRequestProtocol();
   const { data: devcard, isLoading } = useQuery<DevCardData>(
     generateQueryKey(RequestKey.DevCard, { id: userId }),
@@ -41,7 +38,7 @@ export const useDevCard = (
 
       return res.devCard;
     },
-    { staleTime: StaleTime.Default, enabled: !!userId, onSuccess },
+    { staleTime: StaleTime.Default, enabled: !!userId },
   );
 
   const { isProfileCover, user } = devcard ?? {};
--- a/packages/shared/src/hooks/profile/useDevCard.ts
+++ b/packages/shared/src/hooks/profile/useDevCard.ts
@@ -27,7 +27,10 @@ export interface UseDevCard {
   coverImage: string;
 }
 
-export const useDevCard = (userId: string): UseDevCard => {
+export const useDevCard = (
+  userId: string,
+  onSuccess?: (data: DevCardData) => void,
+): UseDevCard => {
   const { requestMethod } = useRequestProtocol();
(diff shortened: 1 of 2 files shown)


packages/webapp/pages/devcard.tsx
```diff
@@ -150,7 +150,13 @@ const Step2 = ({ initialDevCardSrc }: Step2Props): ReactElement => {
     () => generateQueryKey(RequestKey.DevCard, { id: user?.id }),
     [user],
   );
-  const devcard = client.getQueryData<GenerateDevCardParams>(key);
+  const [devcard, setDevCard] = useState(
+    client.getQueryData<GenerateDevCardParams>(key),
+  );
+  useEffect(() => {
```
rebelchris, line 156: https://dailydotdev.slack.com/archives/C02E2C3C13R/p1707919572326829

rebelchris: BTW we have `usePersistentContext` would that solve this maybe?
It's just a wrapper

capJavert: Why do we even use state here, we can just always update query cache from mutation and use that as state with `useQuery`?

kopancek: Well, I didn't say it's a good fix 😅

nensidosari: So @kopancek planning on aligning with feedback, or we want to merge this as is to fix the bug for the moment? Not sure we're clear on the action?

kopancek: Working on a simplified version now :)

nensidosari (approved):
Looks good

packages/webapp/pages/devcard.tsx
capJavert: What I mean the whole `setUpdatePreference` and its `useState` can just always be pulled from devcard right, its the same params? And then instead of setting state you would always mutate query cache. The initial data can be added from `useQuery. initialData`. And then you would not need the `onSuccess` since it would happen as soon the devcard is fetched for the first time and exists.

packages/shared/src/hooks/profile/useDevCard.ts
```diff
@@ -27,7 +27,10 @@ export interface UseDevCard {
   coverImage: string;
 }
 
-export const useDevCard = (userId: string): UseDevCard => {
+export const useDevCard = (
+  userId: string,
+  onSuccess?: (data: DevCardData) => void,
+): UseDevCard => {
```
capJavert, line 33: We use object syntax for hook params eg. `{ userId }` to allow new params to be added without reordering or updating.

packages/webapp/pages/devcard.tsx
kopancek: Hmm, that's true. Except the `type` param, which is not part of the query, but we can keep that as a separate state. I can rewrite again with this in mind 👍

kopancek: @capJavert should be changed now accordingly 🙇

sshanzel (approved):
Code-wise, everything looks great. Just kindly check whether switching themes and other things would still work. I did it something close to this in the past and it was not re-rendering. After that, I think this should be good to go 🚀

PR #2635, opened 2024-02-14
feat: authentication banner

sshanzel:
## Changes
- Made the `AuthOptions` accept a `className` object to be applied on certain elements.
- Introduced `ignoreMessages` prop. This is because we will now be rendering two `AuthOptions` component. One for the banner, and one for the modal. We want the banner to only be utilized when clicking the provider buttons. The other (modal) auth options will be listening to the messages which will continue the process.
- Tracking events are intact since we just reused the same component.

Preview in action:
https://dailydotdev.slack.com/archives/C06K1LYU716/p1707947399158029?thread_ts=1707940049.888179&cid=C06K1LYU716

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

MI-169 #done

### DIFF
--- a/packages/shared/src/components/auth/AuthOptions.tsx
+++ b/packages/shared/src/components/auth/AuthOptions.tsx
@@ -75,6 +75,11 @@ export interface AuthProps {
   defaultDisplay?: AuthDisplay;
 }
 
+interface ClassName {
+  container?: string;
+  onboardingSignup?: string;
+}
+
 export interface AuthOptionsProps {
   onClose?: CloseAuthModalFunc;
   onAuthStateUpdate?: (props: Partial<AuthProps>) => void;
@@ -84,7 +89,7 @@ export interface AuthOptionsProps {
   trigger: AuthTriggersType;
   defaultDisplay?: AuthDisplay;
   forceDefaultDisplay?: boolean;
-  className?: string;
+  className?: ClassName;
   simplified?: boolean;
   isLoginFlow?: boolean;
   onDisplayChange?: (value: string) => void;
@@ -97,7 +102,7 @@ function AuthOptions({
   onAuthStateUpdate,
   onSuccessfulLogin,
   onSuccessfulRegistration,
-  className,
+  className = {},
   formRef,
   trigger,
   defaultDisplay = AuthDisplay.Default,
@@ -344,12 +349,12 @@ function AuthOptions({
       className={classNames(
         'z-1 flex w-full max-w-[26.25rem] flex-col overflow-y-auto rounded-16',
         !simplified && 'bg-theme-bg-tertiary',
-        className,
+        className?.container,
       )}
     >
       <TabContainer<AuthDisplay>
         onActiveChange={(active) => onSetActiveDisplay(active)}
-        controlledActive={activeDisplay}
+        controlledActive={forceDefaultDisplay ? defaultDisplay : activeDisplay}
         showHeader={false}
       >
         <Tab label={AuthDisplay.Default}>
@@ -419,6 +424,7 @@ function AuthOptions({
             isReady={isReady}
             simplified={simplified}
             targetId={targetId}
+            className={className?.onboardingSignup}
           />
         </Tab>
         <Tab label={AuthDisplay.SignBack}>
--- a/packages/shared/src/components/auth/AuthOptions.tsx
+++ b/packages/shared/src/components/auth/AuthOptions.tsx
@@ -95,6 +95,7 @@ export interface AuthOptionsProps {
   onDisplayChange?: (value: string) => void;
   initialEmail?: string;
   targetId?: string;
+  ignoreMessages?: boolean;
 }
 
 function AuthOptions({
@@ -112,6 +113,7 @@ function AuthOptions({
   targetId,
   simplified = false,
   initialEmail = '',
+  ignoreMessages = false,
 }: AuthOptionsProps): ReactElement {
   const { displayToast } = useToastNotification();
   const { syncSettings } = useContext(SettingsContext);
@@ -247,7 +249,7 @@ function AuthOptions({
   };
 
   useEventListener(globalThis, 'message', async (e) => {
-    if (e.data?.eventKey !== AuthEvent.SocialRegistration) {
+    if (e.data?.eventKey !== AuthEvent.SocialRegistration || ignoreMessages) {
       return undefined;
     }
 
--- a/packages/shared/src/components/auth/AuthenticationBanner.tsx
+++ b/packages/shared/src/components/auth/AuthenticationBanner.tsx
@@ -1,0 +1,54 @@
+import React, { ReactElement } from 'react';
+import classed from '../../lib/classed';
+import { OnboardingHeadline } from './OnboardingHeadline';
+import AuthOptions, { AuthDisplay } from './AuthOptions';
+import { AuthTriggers } from '../../lib/auth';
+import { MemberAlready } from '../onboarding/MemberAlready';
+import { useAuthContext } from '../../contexts/AuthContext';
+
+const Container = classed('div', 'flex flex-col w-[23.25rem]');
+
+export function AuthenticationBanner(): ReactElement {
+  const { showLogin } = useAuthContext();
+
+  return (
+    <div
+      className="fixed bottom-0 left-0 z-modal flex w-full flex-row justify-center gap-6 py-10"
+      style={{
+        background: `linear-gradient(270deg, rgba(206, 61, 243, 0.16) 0%, rgba(113, 71, 237, 0.16) 100%), var(--Colors-Background-default, #0E1217)`,
+      }}
+    >
+      <Container>
+        <OnboardingHeadline
+          className={{ title: '!typo-mega3', description: '!typo-title3' }}
+        />
+      </Container>
+      <Container className="pt-2">
+        <AuthOptions
+          ignoreMessages
+          formRef={null}
+          trigger={AuthTriggers.Onboarding}
+          simplified
+          defaultDisplay={AuthDisplay.OnboardingSignup}
+          forceDefaultDisplay
+          className={{ onboardingSignup: '!gap-4' }}
+          onAuthStateUpdate={() =>
+            showLogin({ trigger: AuthTriggers.Onboarding })
+          }
+        />
+        <MemberAlready
+          onLogin={() =>
+            showLogin({
+              trigger: AuthTriggers.Onboarding,
+              options: { isLogin: true },
+            })
+          }
+          className={{
+            container: 'justify-center text-theme-label-secondary typo-callout',
+            login: 'font-bold',
+          }}
+        />
+      </Container>
+    </div>
+  );
+}
--- a/packages/shared/src/components/auth/AuthenticationBanner.tsx
+++ b/packages/shared/src/components/auth/AuthenticationBanner.tsx
@@ -15,7 +15,7 @@ export function AuthenticationBanner(): ReactElement {
     <div
       className="fixed bottom-0 left-0 z-modal flex w-full flex-row justify-center gap-6 py-10"
       style={{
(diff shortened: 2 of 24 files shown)


packages/shared/src/components/auth/AuthOptions.tsx
```diff
@@ -344,12 +351,12 @@ function AuthOptions({
...
         'z-1 flex w-full max-w-[26.25rem] flex-col overflow-y-auto rounded-16',
         !simplified && 'bg-theme-bg-tertiary',
-        className,
+        className?.container,
       )}
     >
       <TabContainer<AuthDisplay>
         onActiveChange={(active) => onSetActiveDisplay(active)}
-        controlledActive={activeDisplay}
+        controlledActive={forceDefaultDisplay ? defaultDisplay : activeDisplay}
```
sshanzel, line 359: Fixed the `forceDefaultDisplay` to always show it, the primary reason we are forcing it, rather than having it in the initial state.

packages/shared/src/components/auth/AuthenticationBanner.tsx
```diff
@@ -0,0 +1,54 @@
...
+import { MemberAlready } from '../onboarding/MemberAlready';
+import { useAuthContext } from '../../contexts/AuthContext';
+
+const Container = classed('div', 'flex flex-col w-[23.25rem]');
+
+export function AuthenticationBanner(): ReactElement {
+  const { showLogin } = useAuthContext();
+
+  return (
+    <div
```
nensidosari, line 15: We might want to use similar setup for other experiments, I already know one coming up soon which had the same idea, so can we have this as a styled component maybe? We can leave the background part out of the styled component though as that might change.

packages/shared/src/components/auth/AuthenticationBanner.tsx
```diff
@@ -0,0 +1,54 @@
...
+      className="fixed bottom-0 left-0 z-modal flex w-full flex-row justify-center gap-6 py-10"
+      style={{
+        background: `linear-gradient(270deg, rgba(206, 61, 243, 0.16) 0%, rgba(113, 71, 237, 0.16) 100%), var(--Colors-Background-default, #0E1217)`,
```
nensidosari, line 18: Can we change the color to pepper 90?
```suggestion
        background: `linear-gradient(270deg, rgba(206, 61, 243, 0.16) 0%, rgba(113, 71, 237, 0.16) 100%), var(--Colors-Background-default, ${colors.pepper['90']})`,
```

packages/shared/src/components/auth/AuthenticationBanner.tsx
```diff
@@ -0,0 +1,54 @@
...
+      </Container>
+      <Container className="pt-2">
+        <AuthOptions
+          ignoreMessages
+          formRef={null}
+          trigger={AuthTriggers.Onboarding}
+          simplified
+          defaultDisplay={AuthDisplay.OnboardingSignup}
+          forceDefaultDisplay
+          className={{ onboardingSignup: '!gap-4' }}
```
nensidosari, line 34: We're adding a lot of !!! everywhere 😢

packages/shared/src/components/post/BasePostContent.tsx
```diff
@@ -29,20 +30,18 @@ export function BasePostContent({
   customNavigation,
 }: BasePostContentProps): ReactElement {
   const { id } = post ?? {};
+  const { onCloseShare, sharePost, onSharePost } = engagementProps;
+  const { onStartArticleOnboarding, showArticleOnboarding } =
```
nensidosari, line 34: Glad another of thesee is removed

`    // @NOTE see https://dailydotdev.atlassian.net/l/cp/dK9h1zoM
    // eslint-disable-next-line react-hooks/rules-of-hooks`

packages/shared/src/lib/featureManagement.ts
```diff
@@ -33,6 +34,10 @@ const feature = {
     webm: cloudinary.onboarding.video.webm,
     mp4: cloudinary.onboarding.video.mp4,
   }),
+  postPageOnboarding: new Feature(
+    'post_page_onboarding',
```
nensidosari, line 38: @nensidosari for me to remember to setup the experiment

packages/webapp/pages/onboarding.tsx
```diff
@@ -375,16 +378,7 @@ export function OnboardPage(): ReactElement {
...
-              Where developers suffer together
-            </OnboardingTitleGradient>
-
-            <h2 className="mb-8 typo-body tablet:typo-title2">
-              We know how hard it is to be a developer. It doesn&apos;t have to
-              be. Personalized news feed, dev community and search, much better
-              than what&apos;s out there. Maybe ;)
-            </h2>
-
+            <OnboardingHeadline />
```
nensidosari, line 381: Maybe to avoid the ! in the banner we're adding, we can actually accept the specific classnames for the title from both places.

nensidosari (commented):
Few comments, but nothing blocking

packages/shared/src/components/auth/AuthenticationBanner.tsx
sshanzel: Ah, good point. Though I think we should use the one that varies based on theme. Will update this quickly.

packages/shared/src/components/auth/AuthenticationBanner.tsx
sshanzel: Yeah, I agree, though I've actually done that in a separate PR as I felt the need there 😅 

So maybe we can resolve this one out?
https://github.com/dailydotdev/apps/pull/2636/commits/97bde42f56ee86f9062c71670012e17d3f8b2749

packages/shared/src/components/auth/AuthOptions.tsx
```diff
@@ -242,7 +249,7 @@ function AuthOptions({
   };
 
   useEventListener(globalThis, 'message', async (e) => {
-    if (e.data?.eventKey !== AuthEvent.SocialRegistration) {
+    if (e.data?.eventKey !== AuthEvent.SocialRegistration || ignoreMessages) {
```
rebelchris, line 252: Because of the ignore here don't you miss below set error?

packages/webapp/pages/onboarding.tsx
sshanzel: You mean like adding a `typoClassName`?

packages/shared/src/components/auth/AuthenticationBanner.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+const Container = classed('div', 'flex flex-col w-[23.25rem]');
+
+export function AuthenticationBanner(): ReactElement {
+  const { showLogin } = useAuthContext();
+
+  return (
+    <div
+      className="fixed bottom-0 left-0 z-modal flex w-full flex-row justify-center gap-6 py-10"
+      style={{
```
rebelchris, line 17: Custom colors?
Maybe double verify with Tsahi, probably a design mistake?

rebelchris (changes requested):
@sshanzel How does it now popup the modal on click?
Feels somehow I'm missing where this is invoked (*although see it working)

packages/shared/src/components/auth/AuthOptions.tsx
sshanzel: Yes, we won't since the modal is opened, it has its own listener (with `ignoreMessages = false`) which allows the whole process to be completed.

packages/shared/src/components/auth/AuthenticationBanner.tsx
nensidosari: Ah yes, in the mobile PR, yes that resolves it. I would suggest though a more generic name like BottomBannerContainer. Do we use something similar for cookies banner as well?

sshanzel:
> @sshanzel How does it now popup the modal on click?
Feels somehow I'm missing where this is invoked (*although see it working)

Through here:
https://github.com/dailydotdev/apps/blob/9c843023f384eaa98a5f8c43ebba80e572603484/packages/shared/src/components/auth/AuthenticationBanner.tsx#L35-L37

packages/webapp/pages/onboarding.tsx
nensidosari: No I mean not having default typo setup in the classnames in OnboardingHeadline, but actually accepting the typos from classname props here just like we do on the other place we use OnboardingHeadline
`ex. className={{ title: '!typo-mega3', description: '!typo-title3' }}` If we do the same here we avoid the !

sshanzel: Ah, yep. That makes sense. Let's do that 🚀

packages/shared/src/components/auth/AuthenticationBanner.tsx
sshanzel: Cool. Would it be okay for me to tackle that in the other PR?

packages/shared/src/components/auth/AuthOptions.tsx
rebelchris: Got it :)

packages/shared/src/components/auth/AuthenticationBanner.tsx
sshanzel: Sent a Slack message. Will wait for their response 🚢

packages/shared/src/components/auth/AuthenticationBanner.tsx
nensidosari: Yup

nensidosari (approved):
All good on my side 🚀

packages/shared/src/lib/featureManagement.ts
nensidosari: experiment feature is setup as well

sshanzel: Thank you!

packages/shared/src/components/auth/AuthenticationBanner.tsx
sshanzel: Found some equivalent values in our codebase, will update quickly.

sshanzel: Applied the missing styles 🥳 

Screenshot 2024-02-15 at 5 59 26 PM

rebelchris: Not urgent, but would it be possible to use the tailwind variation of this?
If too much work never mind :)

rebelchris (approved):
Non blocking comment, otherwise looks good to me :)

packages/shared/src/components/auth/AuthenticationBanner.tsx
sshanzel: Let's see, I will check tailwinds docs for linear gradient as it can definitely be used again in the future.

sshanzel: Doesn't look exactly the same after converting. Not sure if we go for this.

Screenshot 2024-02-15 at 6 35 35 PM

sshanzel: Merging now to unblock, but so far this is what I have tried:

```
 <BottomBannerContainer
      // style={style}
      className="gap-6 bg-gradient-to-l from-theme-overlay-active-cabbage from-0% via-theme-overlay-active-onion via-100% bg-theme-bg-primary py-10 shadow-3"
    >
```

I've also tried adding `to-theme-bg-primary` but doesn't change anything since the stop mentions 0 to 100 already. Not sure how it behaves differently on CSS style.

sshanzel: Found the issue and fixed here: https://github.com/dailydotdev/apps/pull/2638

PR #2640, opened 2024-02-15
refactor: use noindex instead of robots.txt

idoshamun:
Chris suggested a much better and failsafe approach. We should utilize noindex on the page itself.

I added a property for the profile which indicates whether it should add the noindex property.


rebelchris (approved):
Amazing amazing

capJavert (approved):

PR #2645, opened 2024-02-15
chore: remove tailwind base classes for border radius

sshanzel:
## Changes
- Set the defaults to undefined. Once undefined, it becomes excluded from the classes list. Proven by the first commit which had a failing build for using `@apply` on an unknown class.
- Create a reusable function to remove a whole object.
- Start removing undefined from rounded.
- This also removed `rounded` as it apparently has a value of `4px` which is not intuitive with our current setup. With that, I introduced a rounded-4

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-80 #done

### DIFF
--- a/packages/shared/tailwind.config.js
+++ b/packages/shared/tailwind.config.js
@@ -1,25 +1,10 @@
 /* eslint-disable @typescript-eslint/no-var-requires */
-// eslint-disable-next-line import/no-extraneous-dependencies
-const defaultTheme = require('tailwindcss/defaultTheme');
 const colors = require('./tailwind/colors');
 const overlay = require('./tailwind/overlay');
 const boxShadow = require('./tailwind/boxShadow');
 const caret = require('./tailwind/caret');
 const typography = require('./tailwind/typography');
 const buttons = require('./tailwind/buttons');
-
-// by setting a key to undefined, it gets removed in tailwind
-const removeDefaults = (obj, excluded, overwrites) => {
-  const removed = Object.keys(obj).reduce((acc, key) => {
-    if (excluded.includes(key)) {
-      return acc;
-    }
-
-    return { ...acc, [key]: undefined };
-  }, {});
-
-  return { ...removed, ...overwrites };
-};
 
 module.exports = {
   theme: {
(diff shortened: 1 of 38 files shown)


packages/shared/tailwind.config.js
```diff
@@ -186,24 +201,30 @@ module.exports = {
...
-        18: '1.125rem',
-        22: '1.375rem',
-        24: '1.5rem',
-        26: '1.625rem',
-        32: '2rem',
-        40: '2.5rem',
-        48: '3rem',
-        50: '3.125rem',
-      },
+      borderRadius: removeDefaults(
```
rebelchris, line 204: We should probably not use extend, but overwrite then you don't need to remove defaults :)

packages/shared/tailwind.config.js
```diff
@@ -186,24 +201,30 @@ module.exports = {
...
+        defaultTheme.borderRadius,
+        ['full', 'none'],
+        {
+          half: '50%',
```
rebelchris, line 208: Probably better to stick to 1/2 naming as it's common for tailwind

rebelchris (changes requested):
Minor suggestion we discussed, but might have gotten lost in translation :)

packages/shared/tailwind.config.js
sshanzel: I see, I can't seem to find the docs for overwrite, do you maybe know a sample of how to do it?

packages/shared/tailwind.config.js
sshanzel: Yep. Agree with that.

packages/shared/tailwind.config.js
sshanzel: Ahh, I see it now. It is under the `extend` object.

rebelchris: Yeah: https://tailwindcss.com/docs/theme#overriding-the-default-theme

rebelchris (approved):
Looks great, thanks Lee!

omBratteng (approved):

PR #2648, opened 2024-02-16
fix: change copy on devcard share

nensidosari:
- changes description for twitter share

MI-172 #done

### DIFF
--- a/packages/shared/src/lib/labels.ts
+++ b/packages/shared/src/lib/labels.ts
@@ -33,7 +33,7 @@ export const labels = {
   },
   devcard: {
     generic: {
-      shareText: `Excited to share that I've created my personalized devcard on daily.dev, highlighting my favorite tech topics and publications. Create yours now!`,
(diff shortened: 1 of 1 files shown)


packages/shared/src/lib/labels.ts
```diff
@@ -33,7 +33,7 @@ export const labels = {
   },
   devcard: {
     generic: {
-      shareText: `Excited to share that I've created my personalized devcard on daily.dev, highlighting my favorite tech topics and publications. Create yours now!`,
+      shareText: `Check out my #DevCard by @dailydotdev! Flex yours (if it's flex worthy)`,
```
rebelchris, line 36: This would still not add the actual image, wasn't sure, but that was how I interpreted the message?

rebelchris (approved):
Non breaking, but did I get it right?

packages/shared/src/lib/labels.ts
rebelchris: Oh wait maybe it;'s a different prop?

kopancek (approved):

PR #2649, opened 2024-02-16
feat: show streak milestones modals when alert is on

kopancek:
## Changes

- Adds a new hook to show the reading streaks milestone modals

### Screenshots

Screenshot 2024-02-16 at 16 41 08
Screenshot 2024-02-16 at 16 40 40

## Events

- [X] MobileL (420px)
- [X] Tablet (656px)
- [X] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

MI-75 #done


packages/shared/src/hooks/streaks/useStreakMilestone.ts
```diff
@@ -0,0 +1,55 @@
...
+
+  useEffect(() => {
+    if (
+      !loadedAlerts ||
+      alerts?.showStreakMilestone !== true ||
+      isLoading ||
+      !isEnabled ||
+      !loadedSettings ||
+      optOutWeeklyGoal ||
+      !streak?.current
```
rebelchris, line 27: Are the ors really what we want?
Feels like some are or and some should be and no?

It's a bit hard to follow due to the big evaluation.

@capJavert Possibly would like your review on this hook as well :)

rebelchris: Alternative was thinking if we could simply put the high level exits (as optOut above the useEffect even?

packages/shared/src/hooks/streaks/useStreakMilestone.ts
```diff
@@ -0,0 +1,55 @@
...
+
+    openModal({
+      type: modalType,
+      props: {
+        currentStreak: streak?.current,
+        onRequestClose: () => {
+          updateAlerts({ showStreakMilestone: false });
+          closeModal();
+        },
+      } as StreakModalProps,
```
rebelchris, line 40: Because the dynamic modalType Ts issue?

packages/shared/src/hooks/useIncrementReadingRank.ts
```diff
@@ -16,14 +17,27 @@ export default function useIncrementReadingRank(): ReturnType {
   const { user } = useContext(AuthContext);
   const queryKeyRef = useRef<string[]>();
   const queryClient = useQueryClient();
-  const [clearQueries] = useDebounce(() => {
-    queryClient.invalidateQueries(queryKeyRef.current);
-  }, 100);
+  const [clearQueries] = useDebounce(
+    async (readToday = false): Promise<void> => {
+      const promises = [queryClient.invalidateQueries(queryKeyRef.current)];
```
rebelchris, line 22: Is there any change this could result in accidentally clearing all queries (eg if it evaluates null?) (not sure just asking to be safe)

packages/shared/src/hooks/useIncrementReadingRank.ts
```diff
@@ -16,14 +17,27 @@ export default function useIncrementReadingRank(): ReturnType {
...
+      if (!readToday) {
+        promises.push(
+          queryClient.invalidateQueries(
+            generateQueryKey(RequestKey.UserStreak, user),
+          ),
+        );
+      }
+      await Promise.all(promises);
+    },
+    100,
```
rebelchris, line 32: Why do we debounce it actually? (product req or engineering thing)

packages/shared/src/lib/featureManagement.ts
```diff
@@ -24,10 +24,7 @@ const feature = {
   lowImps: new Feature('feed_low_imps'),
   bookmarkOnCard: new Feature('bookmark_on_card', false),
   feedLayout: new Feature('feed_layout', FeedLayout.Control),
-  readingStreaks: new Feature(
-    'reading_streaks',
-    ReadingStreaksExperiment.Control,
-  ),
+  readingStreaks: new Feature('reading_streaks', ReadingStreaksExperiment.V1),
```
rebelchris, line 27: Just be careful to not push this please :)

packages/shared/src/hooks/streaks/useStreakMilestone.ts
capJavert: yeah, basically two things:

1. Minimize amount of dependencies by precomputing conditions outside of hook. Eg. instead of doing:
```js
useEffect(() => {
  if (cond1 && cond2 && !cond3 && cond4) {
    return
  }

  // logic
}, [cond1, cond2, cond3, cond4]);
```

do this:
```js
const condition = cond1 && cond2 && !cond3 && cond4

useEffect(() => {
  if (!condition) {
    return
  }

  // logic
}, [condition]);
```
`condition` variable name should reflect what the condition means. Also it lowers amount of dependencies that React has to check thus lowering number of `useEffect` executions.

2. As Chris said, the condition sometimes might be complicated for single evaluation and few early returns make it more readable so then it would look something like:
```js
const condition1 = cond1 && cond2

const condition2 = !cond3 && cond4

useEffect(() => {
  if (!condition1) {
    return
  }

  if (!condition2) {
    return
  }

  // logic
}, [condition1, condition2]);
```

Point 2. can be optional here but you can probably decide better since you are more in line on what code needs to do 👍

packages/shared/src/hooks/streaks/useStreakMilestone.ts
capJavert: We have a generic available on openModal right, or does that also not work?

packages/shared/src/hooks/useIncrementReadingRank.ts
capJavert: Yep, if `queryKeyRef.current` is `[]` it will invalidate all queries, we should probably check if empty to avoid it.

nensidosari (commented):
Nth to add except what's already being discussed. 🙌

packages/shared/src/hooks/streaks/useStreakMilestone.ts
sshanzel: This is probably because the modal has not yet been introduced to the list of modals at `/components/modals/common.tsx`

packages/shared/src/components/Feed.tsx
```diff
@@ -263,6 +264,8 @@ export default function Feed<T>({
     }
   }, [selectedPost]);
 
+  useStreakMilestone();
```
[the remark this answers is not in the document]
sshanzel: Should probably be at MainLayout? This is so when the user is visiting a post page, and clicked read article, it would count as view - the milestone popup would be possible to be triggered.

sshanzel (commented):
Looking great overall! 🥳

packages/shared/src/hooks/streaks/useStreakMilestone.ts
kopancek: We cannot put opt outs above hooks, because all hoooks need to be evaluated, they cannot be conditionally skipped: https://legacy.reactjs.org/docs/hooks-rules.html#only-call-hooks-at-the-top-level

packages/shared/src/components/Feed.tsx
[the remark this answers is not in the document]
kopancek: To show the modals, the hook needs to be actually executed, which is done on render. If we put it on `MainLayout`, how do we trigger the render? AFAIK we rarely render the whole layout.

packages/shared/src/hooks/streaks/useStreakMilestone.ts
kopancek: Both of the modal types were introduced in `components/modals/common.tsx`. This is the error I was getting from TS compiler:
```
Object literal may only specify known properties, and 'onRequestClose' does not exist in type 'LazyModalComponentType<LazyModal.SquadMember> | LazyModalComponentType<LazyModal.SquadTour> | LazyModalComponentType<LazyModal.UpvotedPopup> | ... 13 more ... | LazyModalComponentType<...>'.ts(2353)
common.tsx(161, 9): The expected type comes from property 'props' which is declared here on type 'LazyModalType<LazyModal>'
```

packages/shared/src/hooks/useIncrementReadingRank.ts
kopancek: There's no change in behavior here from the existing one, so this risk was already there. Do we want to fix it?

packages/shared/src/hooks/useIncrementReadingRank.ts
kopancek: I didn't dig deeper here, I just replicated the existing behavior. Will try to understand it now...

packages/shared/src/hooks/streaks/useStreakMilestone.ts
kopancek: Please take a look again. I removed unneeded dependencies in the condition, split it into a separate variable and rewritten the code into multiple lines with comments.

packages/shared/src/hooks/useIncrementReadingRank.ts
kopancek: The setTimeout and clearing of queries was introduced here: https://github.com/dailydotdev/apps/commit/630e6ca4d5799be56d87067314b787a2e09b2857#diff-2ca4b5627b297d8adac0ec5fa73626249c395ceb53225a067d128ed3fad2a6d5

Maybe it was because of animations? 🤔

packages/shared/src/hooks/useIncrementReadingRank.ts
kopancek: Looking at the code again, `queryKeyRef.current` should never be empty, as we assign it before calling the clearQueries function.

```typescript
const queryKey = getRankQueryKey(user);
queryKeyRef.current = queryKey;	      queryKeyRef.current = queryKey;
```

packages/shared/src/components/Feed.tsx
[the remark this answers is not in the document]
kopancek: Asking product if it's OK to show these modals on feed pages only: https://dailydotdev.slack.com/archives/C02E2C3C13R/p1708421020306569?thread_ts=1708099375.192589&cid=C02E2C3C13R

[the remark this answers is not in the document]
rebelchris: Not sure I follow your comment, we have many of such hooks in the past on mainLayout afaik :) 
(the modals that show on action)

@sshanzel will know a lot about this.
Don't see why we would have to limit ourselves.

packages/shared/src/hooks/useIncrementReadingRank.ts
rebelchris: Perhaps potentially if user is null?
Not sure but if it's easy to simply check for null value it would be appreciated.

packages/shared/src/hooks/useIncrementReadingRank.ts
rebelchris: Hmm probably a edge-case before feel free to remove for your needs if we can get away with not having it for the new one.
But guess we support both so then let's leave to avoid it (can't fully recall)

packages/shared/src/hooks/streaks/useStreakMilestone.ts
rebelchris: It's because we omit them.
```
type LazyModalComponentType<K extends LazyModal> = Omit<
  GetComponentProps<ModalsType[K]>,
  'isOpen' | 'onRequestClose'
>;
```

I think the idea was that these should never be interfered with as they do a lot of magic stuff.
Some alternatives is to pass a onAfterClose *(some modal has this already)

packages/shared/src/hooks/streaks/useStreakMilestone.ts
capJavert: From my point of view good idea with `||=`. I would only check if we polyfill it correctly. The adoption is good but since some browsers got the support pretty recently https://caniuse.com/?search=Logical%20OR%20assignment I would want to be sure. Maybe check in compiled bundle how it looks?

kopancek: Oh, right. TIL that browsers didn't support that assignment before 😊 

I think I'll change to assigning directly. It's just syntactic sugar...

packages/shared/src/hooks/streaks/useStreakMilestone.ts
kopancek: TIL, thanks for the input 👍

packages/shared/src/components/Feed.tsx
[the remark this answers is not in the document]
sshanzel: Yeah, also, there can be a case when the user visits a post page link -> clicks read article -> reads the article in a different tab -> comes back to the post page tab -> popup should show.

> AFAIK we rarely render the whole layout.

The `MainLayout` is rendered in almost every page, probably 96% of all pages (sample excluded ones are `onboarding`, referral page, etc.).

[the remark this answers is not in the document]
kopancek: What I meant by that comment is that when you click on the read post button, the main layout is not re-rendered, so the hook to show the modals would not be executed, increasing the delay that we show the modals.

I have moved the hook back to the `MainLayout` though, please take another look 🙏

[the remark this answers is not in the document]
sshanzel: It would be re-rendered when you go back to the app (window focus) since the boot query will refetch, and it is inside the boot provider which will trigger the re-render and MainLayout is wrapped by the BootProvider.

packages/shared/src/components/MainLayout.tsx
```diff
@@ -87,6 +88,7 @@ function MainLayoutComponent({
   useAuthVerificationRecovery();
   useNotificationParams();
   useReferralReminder();
+  useStreakMilestone();
```
rebelchris, line 91: Nice!!

packages/shared/src/hooks/streaks/useStreakMilestone.ts
```diff
@@ -0,0 +1,45 @@
...
+  useEffect(() => {
+    if (shouldHideModal) {
+      return;
+    }
+
+    openModal({
+      type: modalType,
+      props: {
+        currentStreak: streak?.current,
+        onAfterClose: () => {
```
rebelchris, line 39: Was this a default one we had? Else im missing the prop. But might have missed this nuance myself 😂

rebelchris (approved):
Nothing blocking, basic question

packages/shared/src/hooks/streaks/useStreakMilestone.ts
kopancek: It's defined on `ReactModal.Props` that all our modals extend. So there's more that we probably could be using and didn't know about 😊

rebelchris: Good to know indeed!
Something storybook could enhance 🥲

PR #2650, opened 2024-02-16
feat: new onboarding modal

capJavert:

Adding new onboarding modal needed to show on the flow:
- its a modification of original modal
- added back email field/signup
- moved to separate component so we can easily manipulate until feature is done
- only adding view in this PR so that it can be used where needed per new spec

Some changes from design:
- modal background in design is popover, all our auth modals are tertiary
- email field has placeholder vs everywhere else its with label animating to top (so left it as that)
- title gradient is a slightly different, left it as in app currently

image

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

AS-87 #done

### DIFF
--- a/packages/shared/src/components/auth/AuthOptions.tsx
+++ b/packages/shared/src/components/auth/AuthOptions.tsx
@@ -54,6 +54,7 @@ import { labels } from '../../lib';
 import OnboardingRegistrationForm from './OnboardingRegistrationForm';
 import { useEventListener } from '../../hooks';
 import { trackAnalyticsSignUp } from './OnboardingAnalytics';
+import { OnboardingRegistrationForm4d5 } from './OnboardingRegistrationForm4d5';
 
 export enum AuthDisplay {
   Default = 'default',
@@ -66,6 +67,7 @@ export enum AuthDisplay {
   EmailSent = 'email_sent',
   VerifiedEmail = 'VerifiedEmail',
   OnboardingSignup = 'onboarding_signup',
+  OnboardingSignupV4d5 = 'onboarding_signup_v4.5',
 }
 
 export interface AuthProps {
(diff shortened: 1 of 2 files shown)


capJavert:
@omBratteng since @rebelchris and I were not sure about how it should work since different initiatives as mobile ux, code verification etc. entangled we decided I am just gonna add the view and then we can push it in wherever needed functionality wise.

packages/shared/src/components/auth/OnboardingRegistrationForm4d5.tsx
```diff
@@ -0,0 +1,185 @@
...
+  useEffect(() => {
+    trackEvent({
+      event_name: shouldLogin
+        ? AuthEventNames.OpenLogin
+        : AuthEventNames.OpenSignup,
+      extra: JSON.stringify({ trigger }),
+      target_id: targetId,
+    });
+    // @NOTE see https://dailydotdev.atlassian.net/l/cp/dK9h1zoM
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
nensidosari, line 63: Any issues with adding all dependencies here? If not, let's add and remove this 🙏

nensidosari (approved):
Looks good, just a tiny comment 🙏

packages/shared/src/components/auth/OnboardingRegistrationForm4d5.tsx
capJavert: yeah, our trackEvent function is not stable and we only wanted this event when login state changes so if we change it here it would change how many of events it sends thus impacting analytics.

nensidosari: We can always compare the previous version of login state to the current and if changed we can track within an if condition while having all deps in place. 
But anyways should be fine, we already have a ton of these disables anyways.

packages/shared/src/components/auth/AuthOptions.tsx
```diff
@@ -508,6 +510,34 @@ function AuthOptions({
             )}
           </EmailVerified>
         </Tab>
+        <Tab label={AuthDisplay.OnboardingSignupV4d5}>
+          <OnboardingRegistrationForm4d5
```
rebelchris, line 514: Was there any way we can adjust the `default` to become this?
(can be done later as you suggested)

packages/shared/src/components/auth/OnboardingRegistrationForm4d5.tsx
capJavert: I agree, If you ask me this should just have all the dependencies inside but it was requested like this before from analytics. 👍

rebelchris (approved):
Nothing really blocking here, one remark :)

omBratteng (approved):

PR #2656, opened 2024-02-19
feat: move the tag selection from onboarding to my-feed page

omBratteng:
Moves the tag selection from onboarding page to my-feed page
image


packages/shared/src/components/MainFeedLayout.tsx
```diff
@@ -122,11 +126,15 @@ export default function MainFeedLayout({
     hasFiltered: !alerts?.filter,
     hasUser: !!user,
   });
+  // console.log(feedName);
```
rebelchris, line 129:
```suggestion

```

packages/shared/src/components/MainFeedLayout.tsx
```diff
@@ -122,11 +126,15 @@ export default function MainFeedLayout({
...
   const feedVersion = useFeature(feature.feedVersion);
   const searchVersion = useFeature(feature.search);
+  const onboardingV4dot5 = useFeature(feature.onboardingV4dot5);
```
rebelchris, line 132: @idoshamun Prime example, this will evaluate everyone now, technically we should only enroll new users, but it's almost impossible as to on which level we need this.

Love to hear your opinion on this matter. (non urgent)

packages/shared/src/components/MainFeedLayout.tsx
```diff
@@ -255,6 +263,41 @@ export default function MainFeedLayout({
...
+  const isFeedPreviewEnabled = tagsCount >= REQUIRED_TAGS_THRESHOLD;
+  const [isPreviewFeedVisible, setPreviewFeedVisible] = useState(false);
+
+  if (
+    isOnboardingV4dot5 &&
+    alerts?.filter &&
+    feedName === SharedFeedPage.MyFeed
+  ) {
+    return (
+      <FeedPage className="relative !p-0">
```
rebelchris, line 276: I'd personally avoid having 2 different renders.
Can't we simply make some props dynamic and only render the header dynamic?

Love to hear @capJavert thoughts as well :)

packages/webapp/pages/onboarding.tsx
```diff
@@ -159,7 +163,13 @@ export function OnboardPage(): ReactElement {
   };
 
   const onSuccessfulRegistration = () => {
-    setIsFiltering(true);
+    if (onboardingV4dot5 === OnboardingV4dot5.Control) {
+      setIsFiltering(true);
+    } else {
+      router.replace({
+        pathname: '/my-feed',
```
rebelchris, line 170: I think we need the `welcome` param here as well right?

rebelchris (changes requested):
Minor comments on first review :)

packages/shared/src/components/MainFeedLayout.tsx
omBratteng: 😬

packages/webapp/pages/onboarding.tsx
omBratteng: Not really, this is just the redirect from after signup to get to my-feed page.
After the feed has been generated, we add the welcome to their current url.

packages/shared/src/components/MainFeedLayout.tsx
capJavert: I would agree yes, mostly because it would take two places to update Feed props for any change.

packages/webapp/pages/onboarding.tsx
omBratteng: Noticed now that if I add welcome after signup, the welcome banner is shown in the preview feed 😬 
image

packages/shared/src/components/onboarding/OnboardingFeedHeader.tsx
```diff
@@ -0,0 +1,161 @@
...
+    },
+    [preventNavigation, router.events, stopNav],
+  );
+
+  useEffect(() => {
+    router.events.on('routeChangeStart', routeHandler);
+    return () => {
+      router.events.off('routeChangeStart', routeHandler);
+    };
+  }, [preventNavigation, routeHandler, router.events, stopNav]);
```
capJavert, line 105: Better to move `routeHandler` inside `useEffect` since they care about same dependencies. `preventNavigation` can also go inline then.

capJavert: Also like this the `routeHandler` reference can change before useEffect cleanup which would leak event handlers.

packages/shared/tailwind/colors.js
```diff
@@ -127,6 +127,7 @@ const colors = {
     90: '#0D31D9',
   },
   salt: {
+    0: '#FFFFFF',
```
capJavert, line 130: Just checking if this and other color additions are intended? 🙏

capJavert (changes requested):
Some comments but overall short and sweet 👌

packages/shared/src/components/onboarding/OnboardingFeedHeader.tsx
omBratteng: Refactored it now, and still works as intended 🙏

packages/shared/src/components/onboarding/OnboardingFeedHeader.tsx
```diff
@@ -0,0 +1,154 @@
...
+  tagsCount: number;
+};
+
+export const OnboardingFeedHeader = ({
+  isPreviewFeedVisible,
+  setPreviewFeedVisible,
+  isFeedPreviewEnabled,
+  tagsCount,
+}: OnboardingFeedHeaderProps): JSX.Element => {
+  let stopNav = true;
```
capJavert, line 51: This needs to go inside the useEffect right?

packages/webapp/pages/onboarding.tsx
rebelchris: Oh right, sorry missed the fact that the feed won't be ready yet 😅

packages/shared/src/components/onboarding/OnboardingFeedHeader.tsx
omBratteng: That works too.

packages/shared/src/components/MainFeedLayout.tsx
rebelchris: Everything inside is a whole feed component thats also used by other things..

This is what i mentioned many cases its not just a simple new thing, but logic on existing components.

Making it two makes it almost impossible to maintain (imo)

rebelchris: Well guess the main issue is the rule for evaluation here. (bit of a edge-case)
Target only people who just started onboarding.

Basically you want to say, if you didn't get this feature from onboarding don't do anything 😅

capJavert: Does the "new user" evaluation not come from GB targeting?

packages/webapp/pages/onboarding.tsx
omBratteng: Also just did a change to hide the welcome banner if the feed is actually not ready, but the user has found the welcome url param 😬

packages/shared/src/components/MainFeedLayout.tsx
omBratteng: Changed it to only use one render

packages/shared/src/components/MainFeedLayout.tsx
idoshamun: It's possible to that via GB. But the ideal rule of thumb is to use the feature only when you need it. I should think maybe there are exclusion that we can relax it a bit, new user as an example

packages/shared/src/components/MainFeedLayout.tsx
capJavert: Looks good to me 👍

capJavert (approved):
Aside from discussion when to trigger GB I think good to go, its going to epic so work can still be done on that 🚀

packages/shared/src/components/MainFeedLayout.tsx
omBratteng: @rebelchris thoughts about the change?

rebelchris (approved):
Nothing blocking this merge :)

packages/shared/src/components/MainFeedLayout.tsx
rebelchris: Yeah it's good for me :)

PR #2665, opened 2024-02-20
feat: use the new onboarding auth display as default for all signups

omBratteng:
Sets the default auth display to be the new onboarding 4.5 view on signups

### DIFF
--- a/packages/shared/src/components/auth/AuthOptions.tsx
+++ b/packages/shared/src/components/auth/AuthOptions.tsx
@@ -146,6 +146,7 @@ function AuthOptions({
   const [handleLoginCheck, setHandleLoginCheck] = useState<boolean>(null);
   const [chosenProvider, setChosenProvider] = useState<string>(null);
   const [isRegistration, setIsRegistration] = useState(false);
+  const [isLogin, setIsLogin] = useState(false);
   const windowPopup = useRef<Window>(null);
   const onLoginCheck = () => {
     if (isRegistration) {
@@ -374,7 +375,7 @@ function AuthOptions({
             onPasswordLogin={onPasswordLogin}
             loginHint={loginHint}
             isLoading={isPasswordLoginLoading}
-            isLoginFlow={isForgotPasswordReturn || isLoginFlow}
+            isLoginFlow={isForgotPasswordReturn || isLoginFlow || isLogin}
             trigger={trigger}
             isReady={isReady}
             simplified={simplified}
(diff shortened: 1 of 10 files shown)


rebelchris (approved):

packages/shared/src/components/auth/AuthOptions.tsx
```diff
@@ -146,6 +146,7 @@ function AuthOptions({
   const [handleLoginCheck, setHandleLoginCheck] = useState<boolean>(null);
   const [chosenProvider, setChosenProvider] = useState<string>(null);
   const [isRegistration, setIsRegistration] = useState(false);
+  const [isLogin, setIsLogin] = useState(false);
```
rebelchris, line 149: Isn't this exactly what isLoginFlow should do?
Feels like we now have 2 states for the same

packages/shared/src/components/auth/OnboardingRegistrationForm4d5.tsx
```diff
@@ -43,6 +44,7 @@ export const OnboardingRegistrationForm4d5 = ({
   trigger,
   className,
   onClose,
+  onShowLoginOptions,
 }: OnboardingRegistrationForm4d5Props): ReactElement => {
   const { trackEvent } = useContext(AnalyticsContext);
   const [shouldLogin, setShouldLogin] = useState(false);
```
rebelchris, line 50: Is this one even used here now then?

packages/shared/src/components/onboarding/OnboardingFeedHeader.tsx
```diff
@@ -121,8 +121,8 @@ export const OnboardingFeedHeader = ({
         </div>
       </div>
 
-      <div className="flex max-w-full flex-col">
-        <FilterOnboardingV4 className="mt-44 pt-6 tablet:pt-0" />
+      <div className="flex w-full max-w-full flex-col">
+        <FilterOnboardingV4 className="mt-44 px-10 pt-6 tablet:pt-0" />
```
rebelchris, line 125: I think on mobile it should be less from the design.

rebelchris (changes requested):
Some minor things.

packages/shared/src/components/auth/OnboardingRegistrationForm4d5.tsx
omBratteng: Yeah, it's set to true of signing up with email and it exists, it shows an error message.

packages/shared/src/components/auth/AuthOptions.tsx
omBratteng: Yeah 😅
`isLoginFlow` is controlled via the `loginState` in `AuthContext`, so I would have to import it to update that state
https://github.com/dailydotdev/apps/blob/4585cc58ef9fb3c936139394ef9ba9486008dc60/packages/shared/src/components/auth/AuthModal.tsx#L72C22-L72C34

packages/shared/src/components/auth/OnboardingRegistrationForm4d5.tsx
rebelchris: Oh ok, but not the actual screen, more asking from the event point of view.

packages/shared/src/components/auth/AuthOptions.tsx
rebelchris: Hmm if it's a mayhem let's just do this and have tech debt for later

packages/shared/src/components/auth/OnboardingRegistrationForm4d5.tsx
omBratteng: Since this screen is only signups, it should only send signup events. Once you click the login button, you're navigated to a new screen which would then send a login even
image
image

rebelchris (approved):
no blockers :)

PR #2666, opened 2024-02-20
feat: drawer component

sshanzel:
## Changes
- Drawer component.
- For the moment, this is only used on mobile so I did not introduce multiple position options such as left and right (probably useful on desktop but I don't think the design team has plans for it at the moment).
- Title property to add a pre-style header.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

MI-124 #done


packages/shared/src/components/drawers/Drawer.tsx
```diff
@@ -0,0 +1,74 @@
...
+  onClose,
+}: DrawerProps): ReactElement {
+  const [hasAnimated, setHasAnimated] = useState(false);
+  const [animate] = useDebounce(() => setHasAnimated(true), 1);
+
+  useEffect(() => {
+    if (!hasAnimated) {
+      animate();
+    }
+  }, [animate, hasAnimated]);
```
kopancek, line 48: Why do we need this code here? Seems rather hard to follow.

kopancek (commented):
Looks mostly great to me. The storybook 😍 👍  

Added 1 question. Also noticed when looking at the storybook, that the element only animates when toggled on the screen, should we also have animation for toggling off - e.g. going back where it came from?

packages/shared/src/components/drawers/Drawer.tsx
```diff
@@ -0,0 +1,74 @@
...
+interface ClassName {
+  overlay?: string;
+  drawer?: string;
+}
+
+export interface DrawerProps {
+  children: ReactNode;
+  className?: ClassName;
+  position?: DrawerPosition;
+  closeOnOutsideClick?: boolean;
```
rebelchris, line 20: From the storybook when this is true it also closes if you click on the items, not sure this is intended behaviour?

packages/shared/src/components/drawers/Drawer.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+export function Drawer({
+  children,
+  className = {},
+  position = DrawerPosition.Bottom,
+  closeOnOutsideClick = true,
+  onClose,
+}: DrawerProps): ReactElement {
+  const [hasAnimated, setHasAnimated] = useState(false);
+  const [animate] = useDebounce(() => setHasAnimated(true), 1);
```
rebelchris, line 42: Why do we need the debounce here?

packages/shared/src/components/drawers/Drawer.tsx
```diff
@@ -0,0 +1,74 @@
...
+    if (!hasAnimated) {
+      animate();
+    }
+  }, [animate, hasAnimated]);
+
+  return (
+    <RootPortal>
+      <div
+        className={classNames(
+          'first: fixed inset-0 z-max bg-overlay-quaternary-onion',
```
rebelchris, line 54: `first:` does nothing here?

packages/shared/src/components/drawers/Drawer.tsx
```diff
@@ -0,0 +1,74 @@
...
+        onClick={closeOnOutsideClick && onClose}
+        role="presentation"
+      >
+        <div
+          className={classNames(
+            'absolute flex flex-col overflow-y-auto bg-theme-bg-primary px-4 py-3 transition-transform duration-200 ease-in-out',
+            !hasAnimated && animatePositionClassName[position],
+            drawerPositionToClassName[position],
+            className?.drawer,
+            'max-h-[calc(100vh-5rem)] w-full',
```
rebelchris, line 66: What is the 5rem neg?

packages/shared/src/components/drawers/StorybookDrawerTrigger.tsx
```diff
@@ -0,0 +1,30 @@
+import React, { ReactElement, useState } from 'react';
+import { QueryClient, QueryClientProvider } from '@tanstack/react-query';
+import { Button, ButtonVariant } from '../buttons/Button';
+import { Drawer, DrawerProps } from './Drawer';
+
+export function StorybookDrawerTrigger({
```
rebelchris, line 6: Can't we add this directly to storybook package?
Feels a bit misplaced now.

packages/storybook/stories/drawers/Drawer.stories.tsx
```diff
@@ -0,0 +1,47 @@
+import type { Meta, StoryObj } from '@storybook/react';
```
rebelchris, line 1: Are we expecting more types of Drawers?
If not I'd place it in the root to avoid the double folder structure.

rebelchris (changes requested):
Wow amazing work!
Have some minor suggestions

packages/shared/src/components/drawers/StorybookDrawerTrigger.tsx
sshanzel: It is missing quite a lot of packages, one for react-query, if we want, we can install those deps as well.

packages/shared/src/components/drawers/Drawer.tsx
sshanzel: This part. The screen should not be filled as max height.

Screenshot 2024-02-21 at 5 08 54 PM

packages/shared/src/components/drawers/StorybookDrawerTrigger.tsx
rebelchris: Hmm let's see what @nensidosari thinks.
Don't want to overcomplicate, might just look weird to newcomers

sshanzel: Yeah, agree with that.

packages/shared/src/components/drawers/Drawer.tsx
sshanzel: When animating the entrance of conditionally rendered elements, we need to ensure the element is available before running the animation itself. Otherwise, it will not work.

packages/shared/src/components/drawers/Drawer.tsx
sshanzel: Same here https://github.com/dailydotdev/apps/pull/2666#discussion_r1497149472

packages/shared/src/components/drawers/StorybookDrawerTrigger.tsx
sshanzel: Also don't want this here. I'm open to suggestions on how we can tackle this. Maybe the right way is by installing the deps? Not sure yet at the moment 😅

packages/shared/src/components/drawers/Drawer.tsx
kopancek: Should we use a ref instead? How does this make sure the element is available?

packages/shared/src/components/drawers/StorybookDrawerTrigger.tsx
kopancek: Why do we need `QueryClientProvider` in here?

packages/shared/src/components/drawers/Drawer.tsx
sshanzel: Oh, is it. That should not be the case. Let me check quickly.

packages/shared/src/components/drawers/Drawer.tsx
sshanzel: On line 42, we wrapped the function in a debounce with 1ms, which ensures the elements are already written in the DOM.

We need the re-render to execute the animation, hence we can only use `useState`.

sshanzel:
> Added 1 question. Also noticed when looking at the storybook, that the element only animates when toggled on the screen, should we also have animation for toggling off - e.g. going back where it came from?

Oh, I thought we are only animating the entrance. Yeah we can do the exit as well. Though let's check first with the design team as it will have additional engineering effort (similar to the entrance, we need to wait for the animation to finish before conditionally rendering out)

packages/shared/src/components/drawers/StorybookDrawerTrigger.tsx
sshanzel: The Storybook is crashing without it. We have some internal hooks that utilize the query client.

packages/shared/src/components/drawers/Drawer.tsx
kopancek: If you use a `ref` on the main drawer element, we don't really need a debounce and it actually ensures that we have the element in the DOM. Or am I missing something? :)

sshanzel: Ah, I thought you meant `useRef` over `useState`. That is actually a great idea. Let me check that quickly.

packages/shared/src/components/drawers/StorybookDrawerTrigger.tsx
kopancek: I think this is the reason: https://github.com/dailydotdev/apps/blob/MI-124-drawer-component/packages/shared/src/components/tooltips/Portal.tsx#L25

Do we. need to use the `RootPortal`? Maybe that could be a quick fix :)

packages/shared/src/components/drawers/Drawer.tsx
sshanzel: We would still need the debounce though that was a better approach since we are now one less `useEffect`. Thanks for the suggestion! 🚀

packages/shared/src/components/drawers/StorybookDrawerTrigger.tsx
sshanzel: Arguably, should be possible not to use it. There could be some implementations in the future where the Drawer might not work without it, but we can probably tackle that when it comes.

packages/shared/src/components/drawers/Drawer.tsx
rebelchris: Oh ok got it :)

packages/shared/src/components/drawers/StorybookDrawerTrigger.tsx
sshanzel: Yep. Have decided to remove the Portal and bring it back once necessary.

packages/shared/src/components/drawers/Drawer.tsx
sshanzel: Should be fixed now 🚀

rebelchris: Yes validated! Thanks :D
Really nice to see Storybook being so easy to test now :D

rebelchris (approved):
Wow, love how you made it so much cleaner now and works like a charm!

packages/shared/src/components/drawers/Drawer.tsx
sshanzel: True, we are now able to say confidently that things are working 🚀

sshanzel:
@kopancek have added the exit animation now 🥳

kopancek (approved):
Approving to unblock 👍

nensidosari (approved):
Love it!

sshanzel:
Refactored a bit and moved away from using a hook and instead made it easier to use for everyone.

PR #2670, opened 2024-02-21
Fix scrool to top

nensidosari:

reduce scroll to top button to 1

### DIFF
--- a/packages/shared/src/components/ScrollToTopButton.tsx
+++ b/packages/shared/src/components/ScrollToTopButton.tsx
@@ -7,6 +7,7 @@ import {
   ButtonVariant,
 } from './buttons/Button';
 import { ArrowIcon } from './icons';
+import { useViewSize, ViewSize } from '../hooks';
 
 const baseStyle: CSSProperties = {
   transition: 'transform 0.1s ease-out, opacity 0.1s ease-out',
@@ -15,6 +16,7 @@ const baseStyle: CSSProperties = {
 
 export default function ScrollToTopButton(): ReactElement {
   const [show, setShow] = useState(false);
+  const isLaptop = useViewSize(ViewSize.Laptop);
 
   useEffect(() => {
     const callback = () => {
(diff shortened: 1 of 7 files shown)


packages/shared/src/components/ScrollToTopButton.tsx
```diff
@@ -36,29 +38,17 @@ export default function ScrollToTopButton(): ReactElement {
...
-        style={style}
-      />
-    </>
+    <Button
+      aria-label="scroll to top"
+      {...props}
+      className={classNames(
+        'absolute -top-18 right-4 z-2 laptop:-top-24 laptop:right-8',
+        !show && 'hidden',
+        show && isLaptop && 'laptop:flex',
```
capJavert, line 47: Can we also just not render if `show=false`?

kopancek, line 47: Do we need the `isLaptop` check? We want to show the button for all the screen sizes, just different styling right?

kopancek: Is the laptop:flex even doing anything? 🤔

nensidosari: 🤦‍♀️  no don't think so, nice catch 😅

rebelchris (approved):
Nice one :) 
Besides Milan's comment nothing to add.

Approving so long

packages/shared/src/components/ScrollToTopButton.tsx
nensidosari: yeah good ctach

kopancek (approved):
Looks great :shipit:

sshanzel (approved):
Looks great! 🥳

packages/shared/src/components/ScrollToTopButton.tsx
capJavert: I mean, if there is no other reason it is always better to not render instead of hide.

capJavert (approved):

packages/shared/src/components/ScrollToTopButton.tsx
nensidosari: Done

PR #2671, opened 2024-02-21
feat: remove onboarding feed message and pulsating animations

nensidosari:
## Changes
- added new feature onboarding_optimizations
- removing tooltip, animation from feed settings
- removing welcome message

- Based on [specs](https://dailydotdev.atlassian.net/wiki/spaces/HAN/pages/795770908/Onboarding+optimizations) here, we want to try and optimise onboarding by removing tooltip and animation on feed settings, and welcome message

## Events

- [x] MobileL (420px)
- [x] Tablet (656px)
- [x] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

MI-180 #done

### DIFF
--- a/packages/shared/src/components/filters/MyFeedHeading.tsx
+++ b/packages/shared/src/components/filters/MyFeedHeading.tsx
@@ -13,7 +13,6 @@ import AlertPointer, {
   AlertPointerProps,
   OffsetXY,
 } from '../alert/AlertPointer';
-import { filterAlertMessage } from './FeedFilters';
 import { Alerts } from '../../graphql/alerts';
 import AnalyticsContext from '../../contexts/AnalyticsContext';
 import { AnalyticsEvent } from '../../lib/analytics';
@@ -21,6 +20,8 @@ import { useFeature } from '../GrowthBookProvider';
 import { feature } from '../../lib/featureManagement';
 import { SearchExperiment } from '../../lib/featureValues';
 import { useFeedLayout } from '../../hooks';
+
+export const filterAlertMessage = 'Edit your personal feed preferences here';
 
 interface MyFeedHeadingProps {
   isAlertDisabled: boolean;
@@ -41,6 +42,7 @@ function MyFeedHeading({
   const shouldShowHighlightPulse = router.query?.hset === 'true';
   const { shouldUseFeedLayoutV1 } = useFeedLayout();
   const isV1Search = searchVersion === SearchExperiment.V1;
+  const onboardingOptimizations = useFeature(feature.onboardingOptimizations);
 
   const onClick = () => {
     trackEvent({ event_name: AnalyticsEvent.ManageTags });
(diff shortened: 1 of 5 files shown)


packages/shared/src/components/onboarding/FeedReadyMessage.tsx
```diff
@@ -15,6 +17,11 @@ export const FeedReadyMessage = ({
   className,
 }: FeedReadyMessageProps): ReactElement => {
   const { user } = useContext(AuthContext);
+  const onboardingOptimizations = useFeature(feature.onboardingOptimizations);
```
nensidosari, line 20: Adding check here and not where this component is rendered because that is the FeedContainer that is used in every feed and would enroll also users who view the feed not for the first time. 
Based on [this](https://dailydotdev.slack.com/archives/C02E2C3C13R/p1708508556884909)

packages/shared/src/components/filters/MyFeedHeading.tsx
```diff
@@ -41,6 +42,7 @@ function MyFeedHeading({
   const shouldShowHighlightPulse = router.query?.hset === 'true';
   const { shouldUseFeedLayoutV1 } = useFeedLayout();
   const isV1Search = searchVersion === SearchExperiment.V1;
+  const onboardingOptimizations = useFeature(feature.onboardingOptimizations);
```
nensidosari, line 45: I'm not sure though how to exclude anyone coming at this component not directly from onboarding...

sshanzel: Maybe through query params? Or browser history? But still, yeah, this needs to be a component deeper if we are going to check those conditions.

sshanzel (commented):
Looking great! Nothing major. Would love to get other's opinion as well.

packages/shared/src/components/filters/MyFeedHeading.tsx
nensidosari: > this needs to be a component deeper if we are going to check those conditions.

Yup no matter what condition, means we might need to separate more on components if we want to do that...

rebelchris (changes requested):
For both flows you could have actually gone one up to onboarding and just not set `welcome=true&hset=true` that way you ensure it's only from onboarding and it's almost easier to not render these things.

The only one we don't have then is the tooltip, but maybe we could even add a new param for that?

nensidosari:
> For both flows you could have actually gone one up to onboarding and just not set welcome=true&hset=true that way you ensure it's only from onboarding and it's almost easier to not render these things.

If we do that then it means we cannot hook the share tracking feature with these params...which we are currently. But yeah maybe easier if we set a separate param for that.

> The only one we don't have then is the tooltip, but maybe we could even add a new param for that?

Could easily set one for it.

@rebelchris thanks, great suggestion.

packages/shared/src/components/filters/MyFeedHeading.tsx
```diff
@@ -82,7 +83,7 @@ function MyFeedHeading({
 
   const alertProps: Omit<AlertPointerProps, 'children'> = {
     offset: getOffset(),
-    isAlertDisabled,
+    isAlertDisabled: shouldHighlightFeedSettings ? isAlertDisabled : true,
```
rebelchris, line 86: When do we set the isAlertDisabled?
Nothing really to add, just wanted to ensure we remove it correctly.

packages/webapp/pages/onboarding.tsx
```diff
@@ -146,10 +147,12 @@ export function OnboardPage(): ReactElement {
...
-      query: {
-        welcome: 'true',
-        hset: 'true',
-      },
+      query: onboardingOptimizations
+        ? {}
+        : {
+            welcome: 'true',
+            hset: 'true',
+          },
```
rebelchris, line 155: Not important, but different suggestion:

```suggestion
     ...(onboardingOptimizations && {
    query: {
      welcome: 'true',
      hset: 'true',
    },
  }
```

rebelchris (approved):
Nothing blocking :)

packages/webapp/pages/onboarding.tsx
nensidosari: We'll most propably add something to the query params here for share tracking which means we would need to change this back to the previous. But yeah until then, can change to this.

packages/shared/src/components/filters/MyFeedHeading.tsx
nensidosari: https://github.com/dailydotdev/apps/blob/1af7586539ce01bd150cbe8f6c167d21f3fb6df9/packages/shared/src/components/layout/common.tsx#L82

It's related to some alert we apparently have for this, we could possibly remove completely as an alert, but it's used in one alert dot somewhere in mobile sidebar as well, so decided to leave as is.

packages/webapp/pages/onboarding.tsx
rebelchris: Ah ok yeah then it's fine either way

PR #2674, opened 2024-02-21
feat: form wrapper

sshanzel:
## Changes
- Form wrapper with built-in buttons that copy is modifiable.
- Storybook is also available.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

MI-137 #done

### DIFF
--- a/packages/shared/src/components/fields/form/FormWrapper.tsx
+++ b/packages/shared/src/components/fields/form/FormWrapper.tsx
@@ -1,0 +1,44 @@
+import React, { ReactElement, ReactNode } from 'react';
+import classNames from 'classnames';
+import { Button, ButtonVariant } from '../../buttons/Button';
+
+interface Copy {
+  left?: string;
+  right?: string;
+}
+
+interface FormWrapperProps {
+  children: ReactNode;
+  className?: string;
+  form: string;
+  copy?: Copy;
+  onLeftClick?(): void;
+  onRightClick?(): void;
+}
+
+export function FormWrapper({
+  children,
+  className,
+  form,
+  copy = {},
+  onLeftClick,
+  onRightClick,
+}: FormWrapperProps): ReactElement {
+  return (
+    <div className={classNames('flex flex-col', className)}>
+      <div className="flex flex-row justify-between border-b border-theme-divider-tertiary px-4 py-2">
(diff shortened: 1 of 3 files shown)


packages/shared/src/components/fields/form/FormWrapper.tsx
```diff
@@ -0,0 +1,44 @@
...
+  className,
+  form,
+  copy = {},
+  onLeftClick,
+  onRightClick,
+}: FormWrapperProps): ReactElement {
+  return (
+    <div className={classNames('flex flex-col', className)}>
+      <div className="flex flex-row justify-between border-b border-theme-divider-tertiary px-4 py-2">
+        <Button variant={ButtonVariant.Tertiary} onClick={onLeftClick}>
```
rebelchris, line 30: These are always solid Tertiary? was wondering if it would be easier to simply accept left and right buttons as complete elements?

packages/shared/src/components/fields/form/FormWrapper.tsx
```diff
@@ -0,0 +1,44 @@
...
+interface FormWrapperProps {
+  children: ReactNode;
+  className?: string;
+  form: string;
+  copy?: Copy;
+  onLeftClick?(): void;
+  onRightClick?(): void;
+}
+
+export function FormWrapper({
```
rebelchris, line 19: The idea is that this will be a seperate render for mobile, so we'll dynamic wrap or something?

rebelchris (commented):
Only questions so far, no remarks on the code.

packages/shared/src/components/fields/form/FormWrapper.tsx
nensidosari: If we do that then I don't even think we need a wrapper in the first place, just styled components.

If we're going to be using this the same way in a couple of places I think it doesn't hurt to have it non customisable at the moment, we can customise later base on our use cases.

nensidosari (commented):
Looks good. Let's just agree on the component itself what we want to do. Happy to approve

nensidosari (approved):
Suggestion to wait on @rebelchris confirmation as well. Looks good on my side

rebelchris (approved):

PR #2675, opened 2024-02-22
fix: vscode eslint argument

omBratteng:
Keep getting this change whenever I open apps in VSCode, so I presume it means that the extension has changed the default value from `true` to `explicit`.


capJavert (approved):
Yep, tnx for fixing!

rebelchris (approved):

PR #2676, opened 2024-02-22
refactor: remove signup page

omBratteng:
- Removes signup page and adds redirect to onboarding page.

WT-2105


rebelchris (approved):
perfect

sshanzel (approved):
Nice. Finally, have this removed. Also, great thinking with the redirection 🤓

packages/webapp/next.config.js
```diff
@@ -91,6 +91,11 @@ module.exports = withTM(
               source: '/posts/finder',
               destination: '/search?provider=posts',
               permanent: false
+            },
+            {
+              source: '/signup',
+              destination: '/onboarding',
+              permanent: true
```
capJavert, line 98: Do we want it permanent `/signup` is a common path so in the future we might want to do something else on it?

omBratteng: Don't think it's much of a problem. It would only affect new users, and if we start using signup page again, new users (never visited before), won't have ever visited the redirect before. 

But I can change it to temporary.

omBratteng: I'll just change it to temporary.

omBratteng: Done, it has been updated in #2630

PR #2677, opened 2024-02-22
fix: calculate the local timezone offset when generating the heatmap

omBratteng:
- Adds local timezone offset when generating heat map

The heatmap generator does not take into account the users timezone, so when they have read a post today, it looks like it was yesterday.

I set my local timezone to GMT-8 (Vancouver, Canada), and ran the following line, which assumes GMT+0 time, and then it outputs it as local time. 

```
> new Date("2024-02-22")
Wed Feb 21 2024 16:00:00 GMT-0800
```

<tr>
<td>
<td>

```json
{
  "data": {
    "userReadingRankHistory": [
      {
        "rank": 1,
        "count": 1
      }
    ],
    "userReadHistory": [
      {
        "date": "2024-02-22",
        "reads": 1
      }
    ],
    "userMostReadTags": []
  }
}
```

### DIFF
--- a/packages/shared/src/components/CalendarHeatmap.tsx
+++ b/packages/shared/src/components/CalendarHeatmap.tsx
@@ -118,7 +118,10 @@ export function CalendarHeatmap<T extends { date: string }>({
   >(
     () =>
       values.reduce((acc, value) => {
-        const date = new Date(value.date);
+        const localDate = new Date(value.date);
+        const date = new Date(
+          localDate.valueOf() + localDate.getTimezoneOffset() * 60 * 1000,
+        );
         const index = differenceInDays(date, startDateWithEmptyDays);
         if (index < 0) {
           return acc;


omBratteng:
Marking as draft, might need a bit more thinking

kopancek:
To me this looks like we are fixing it in a wrong spot. Because even if you fix the dates, the calculation is still inherently wrong, since it will count posts that you read on one day into either tomorrow, yesterday or tomorrow depending on the actual timezone you are in.

So I think a better solution would be to change the heatmap generator to take the timezone of the user into account...

omBratteng:
Working with time is 💩.
On the API side, each view is stored as UTC, and when we get the reads for the user, we take their saved timezone into account. So if you read at `2023-12-31 23:00 GMT-1`, your read is saved as `2024-01-01 00:00 UTC`, the API response would be `2023-12-31`.

But because JS parses date strings with no time as being at midnight, it would be shown in the heat map on the 30th, because `2023-12-31 00:00 GMT` is `2023-12-30 23:00 GMT-1`

Time 😢 

I like the way that GitHub has solved it. It shows the heat map entries with the visitors timezone taken into account

PR #2681, opened 2024-02-23
chore: cleanup post page onboarding experiment

nensidosari:
- cleanup post page onboarding experiment since experiment was concluded ([thread](https://dailydotdev.slack.com/archives/C06K1LYU716/p1708690792715929?thread_ts=1708660588.090829&cid=C06K1LYU716))

### DIFF
--- a/packages/shared/src/hooks/auth/useOnboarding.tsx
+++ b/packages/shared/src/hooks/auth/useOnboarding.tsx
@@ -1,6 +1,3 @@
-import { useFeature } from '../../components/GrowthBookProvider';
-import { feature } from '../../lib/featureManagement';
-import { PostPageOnboarding } from '../../lib/featureValues';
 import { useAuthContext } from '../../contexts/AuthContext';
 
 interface UseOnboarding {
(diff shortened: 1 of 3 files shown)


rebelchris (approved):

packages/shared/src/hooks/auth/useOnboarding.tsx
```diff
@@ -9,9 +6,7 @@ interface UseOnboarding {
 
 export const useOnboarding = (): UseOnboarding => {
   const { isAuthReady, user } = useAuthContext();
-  const postPageOnboarding = useFeature(feature.postPageOnboarding);
-  const shouldShowAuthBanner =
-    postPageOnboarding === PostPageOnboarding.V1 && isAuthReady && !user;
+  const shouldShowAuthBanner = isAuthReady && !user;
```
omBratteng, line 9: So it should always be shown I guess. We can remove the `PostFeedFiltersOnboarding` component then?

nensidosari: Yes the banner will always be shown on post page when user is not authenticated. But hmm not sure, which component is that and what does it do?

sshanzel: This is what I was referring to that we need to clean up. The banner on top of the post page/modal that we matched this experiment against.

PR #2682, opened 2024-02-26
chore: upgrade OneSignal package to v3

sshanzel:
One of the primary reasons why we had to upgrade to v3 was, Safari doesn't work anymore on v2 (if you visit our app on Safari, you won't be able to toggle on/off the subscription at the accounts page. In the console, there is going to be an error message about fetching the scripts.

What made this task difficult was, OneSignal won't run with our local environment since it needs to be either under `localhost` or running with HTTPS. Both are not possible at the moment, though Ole-Martin might look into making our local environment HTTPS-friendly in the future. I had to turn to the preview deployment every time I make a significant change to monitor whether things are still working or not.

Other than that, it is pretty complicated to work with the NotificationsContext due to functions and initializations found at the provider itself.

## Changes
- One thing stands out when working with the NotificationsContext. Everything is in a single place which is the context provider which makes it hard to follow the code.
- I have structured it into 3 blocks through some hooks and separation of concerns:
  - I introduced a `useOneSignal` hook, this is to contain all initialization processes and set up listeners when permissions/subscriptions change. This should handle whether the user is subscribed and whether we support it.
  - Another hook is `usePushNotification` which manages the mutations that decide whether to display certain alerts or to hide them based on the current state.
  - Now, everything left at the `NotificationsContext` should be related to the actual in-app notification which should be as simple as pulling data and setting unread count.
  - Everything related to push notification is now under the `push` object in the NotificationsContext to contain the logs behind it in a single object.
- Another thing I refactored is the way we handle mutations. In the past, there was only one function that handled whether the user wanted to subscribe or toggle off.
  - I've made the function be split into two to make it straightforward.
    - A function named `onEnable` - should support the "Enable permission" buttons easily without having to worry if we need to have additional changes to support different initiatives.
    - I kept the `onTogglePermission` but simplified it by just deciding whether to opt in or out of the subscription.

I've also tested on [preview](https://preview.app.daily.dev) whether the events are sent from at least three locations (all of these areas use the same component which is why the events naturally work the same way across all of them):

Screenshot 2024-02-26 at 2 50 20 AM
Screenshot 2024-02-26 at 2 49 57 AM
Screenshot 2024-02-26 at 2 49 06 AM
Screenshot 2024-02-26 at 2 47 56 AM
Screenshot 2024-02-26 at 2 42 29 AM

Note: nevermind the lint and tests, I will fix them soon 😅 

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done


packages/shared/src/hooks/notifications/useOneSignal.ts
```diff
@@ -0,0 +1,97 @@
...
+  const { trackEvent } = useAnalyticsContext();
+  const [isSubscribed, setIsSubscribed] = useState(false);
+  const client = useQueryClient();
+  const key = generateQueryKey(RequestKey.OneSignal, user);
+  const isEnabled = !!user && !isExtension && !isTesting;
+  const {
+    data: OneSignal,
+    isFetched,
+    isLoading,
+  } = useQuery<typeof OSR>(
```
sshanzel, line 43: Instead of a `useEffect`, we are now initializing through `useQuery`.

packages/shared/src/contexts/NotificationsContext.tsx
```diff
@@ -1,39 +1,28 @@
...
-import AuthContext from './AuthContext';
 import { AnalyticsEvent, NotificationPromptSource } from '../lib/analytics';
 import { useAnalyticsContext } from './AnalyticsContext';
+import { SubscriptionCallback } from '../hooks/notifications/useOneSignal';
+import {
+  UsePushNotification,
+  usePushNotification,
+} from '../hooks/notifications/usePushNotification';
 
-export interface NotificationsContextData
```
sshanzel, line 22: Simplified the returned properties by either encapsulating it through the single object or by introducing a hook if it is something that can be accessed through `useQuery`.

packages/shared/src/lib/lifecycle.ts
```diff
@@ -2,7 +2,8 @@
 
 declare global {
   interface Window {
-    safari?: { pushNotification?: unknown };
+    /* eslint-disable  @typescript-eslint/no-explicit-any */
+    safari?: { pushNotification: any };
```
sshanzel, line 6: Had to resort to `any` with `eslint-disable`. Using `unknown` or `never` is not getting accepted by the build.

packages/shared/package.json
```diff
@@ -121,7 +121,7 @@
     "lodash.clonedeep": "^4.5.0",
     "node-fetch": "^2.6.6",
     "react-markdown": "^8.0.7",
-    "react-onesignal": "^2.0.4",
+    "react-onesignal": "^3.0.1",
```
rebelchris, line 124: Would we also need to update the API side?

packages/shared/src/contexts/NotificationsContext.tsx
rebelchris: I actually debated moving this to it's own context called `InApp` or something or `Push` since it's only used on webapp .
Let me know if you think that makes sense.

packages/shared/src/contexts/NotificationsContext.tsx
```diff
@@ -177,102 +79,25 @@ export const NotificationsContextProvider = ({
...
+    [],
   );
 
+  const data: NotificationsContextData = {
+    isNotificationsReady,
+    unreadCount: currentUnreadCount,
+    clearUnreadCount,
+    incrementUnreadCount,
+    trackPermissionGranted,
+    push,
```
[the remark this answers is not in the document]
rebelchris: See my comment above, might be good to abstract into seperate context as they don't have anything to do with each other basically.
(I think only the invocation one is shared between apps)

packages/shared/src/hooks/useEnableNotification.ts
```diff
@@ -92,17 +78,7 @@ export const useEnableNotification = ({
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [shouldShowCta]);
 
-  useEffect(() => {
-    return () => {
-      if (onAcceptedPermissionJustNow) {
-        onAcceptedPermissionJustNow(false);
-      }
-    };
-  }, [onAcceptedPermissionJustNow]);
```
rebelchris, line 101: This one is replaced elsewhere or how do we prompt this?

rebelchris (approved):
Not blocking, but open for discussion with the recent "hotfix" to remove OneSignal from the build via webpack magic.
Perhaps we can here split out the context into 2?
Basically the extension never uses the Push context and thus can be empty or omitted?

Let me know what you think.
Would also love to see atleast 2 reviews on this 🙏

And will try it out today somewhere on preview.

packages/shared/src/contexts/NotificationsContext.tsx
sshanzel: Yeah, that makes sense too. Especially they seem like have nothing to do with one another.

packages/shared/package.json
sshanzel: Good question. This should be independent in terms of usage, but since there is nothing wrong yet with the API it should be good. Probably worth to check as well if we need to upgrade it as well.

sshanzel:
> Basically the extension never uses the Push context and thus can be empty or omitted?

I really like this idea. Let's see how we can move forward.

packages/shared/src/contexts/NotificationsContext.tsx
rebelchris: Yeah if it's not a "major" effort I would suggest to do it here since you already abstracted it mostly.

packages/shared/src/hooks/useEnableNotification.ts
sshanzel: This one is to ensure the banner is removed when the user moves out of the page/component. By using `useQuery` with a function that returns a default value false, this automatically happens.

packages/shared/src/contexts/NotificationsContext.tsx
sshanzel: Yeah, mostly copy-paste, would just require a bit of work. Let me fix the blockers first, then I will get back to this.

packages/shared/src/hooks/useEnableNotification.ts
rebelchris: Ah cool got it, yeah just wanted to ensure it still happens on UI end.

packages/webapp/pages/account/notifications.tsx
```diff
@@ -129,34 +138,34 @@ const AccountNotificationsPage = (): ReactElement => {
     }
   };
 
+  const label = isSubscribed ? 'On' : 'Off';
+
   return (
     <AccountPageContainer title="Notifications">
-      {isNotificationSupported && (
```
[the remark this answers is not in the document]
nensidosari: I'd say mentioned it as a thing to them that it's like this now, but move on with it. Whenever they think they what to improve this we can do/prioritise after.

nensidosari (approved):
Looks good on my side. Just checking, would you be fixing the Alert that shows up on this PR as well?

packages/webapp/__tests__/AccountNotificationsPage.tsx
```diff
@@ -131,20 +131,6 @@ const renderComponent = (user = defaultLoggedUser): RenderResult => {
   );
 };
 
-it('should change user push notification', async () => {
```
sshanzel, line 134: This won't work anymore since we are basing our subscription state on OneSignal's response and it needs integration.

rebelchris (approved):
Also validated we'll still need the webpack magic :( 
Nothing blocking, but if you could perhaps add a small note just to avoid confusion down the line.

Let's also get another vision on this :)

packages/shared/src/contexts/NotificationsContext.tsx
```diff
@@ -45,234 +23,32 @@ export interface NotificationsContextProviderProps {
...
-            permission: 'granted',
-            ...(existing_permission && { existing_permission }),
-          }),
-        });
-      }
-    };
-  }, [trackEvent, notificationSourceRef]);
 
   useEffect(() => {
     setCurrentUnreadCount(unreadCount);
```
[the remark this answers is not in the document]
sshanzel: Yeah, basically this is for on load since fetching the count is async, the initial state is 0 and it should have been 3, after the fetch.

packages/shared/src/contexts/PushNotificationContext.tsx
```diff
@@ -0,0 +1,150 @@
...
+  isPushSupported: boolean;
+  isInitialized: boolean;
+  isSubscribed: boolean;
+  isLoading: boolean;
+  shouldOpenPopup: boolean;
+  onSourceChange: (source: string) => void;
+  trackPermissionGranted: (source: NotificationPromptSource) => void;
+}
+
+export const PushNotificationsContext =
```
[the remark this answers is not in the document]
sshanzel: Yeah, we can leave a comment here I believe.

packages/shared/src/hooks/notifications/useAcceptedPushNow.ts
```diff
@@ -0,0 +1,31 @@
...
+import { generateQueryKey, RequestKey } from '../../lib/query';
+import { useAuthContext } from '../../contexts/AuthContext';
+import { disabledRefetch } from '../../lib/func';
+
+interface UseAcceptedPushNow {
+  onAcceptedJustNow: (accepted: boolean) => void;
+  acceptedJustNow: boolean;
+}
+
+export const useAcceptedPushNow = (): UseAcceptedPushNow => {
```
capJavert, line 12: Just to understand why "now" is needed, on what specific flow do we need to say user just now accepted push vs accepted in general?

packages/webapp/__tests__/AccountNotificationsPage.tsx
capJavert: Maybe we can mock it? But yeah, probably better of as e2e backlog test once we have it.

packages/shared/src/lib/lifecycle.ts
capJavert: Where is this used? Or does it break build due to internal usage by onesignal?

packages/shared/src/contexts/PushNotificationContext.tsx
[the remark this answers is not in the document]
capJavert: We can also throw in addition to comment if extension, so in dev you notice it immediately

packages/shared/src/contexts/PushNotificationContext.tsx
```diff
@@ -0,0 +1,150 @@
...
+          event_name: AnalyticsEvent.ClickEnableNotification,
+          extra: JSON.stringify({
+            origin: source || notificationSourceRef.current,
+            permission: 'granted',
+            ...(existing_permission && { existing_permission }),
+          }),
+        });
+      }
+    };
+  }, [trackEvent]);
```
capJavert, line 126: Is this only to keep `subscriptionCallbackRef.current` in sync with `trackEvent` reference? If so we can just assign inline during render without useEffect.

packages/shared/src/contexts/PushNotificationContext.tsx
```diff
@@ -0,0 +1,150 @@
...
+
+      setIsSubscribed(OneSignal.User.PushSubscription.optedIn);
+
+      OneSignal.User.PushSubscription.addEventListener(
+        'change',
+        ({ current }) => {
+          setIsSubscribed(() => current.optedIn);
+          subscriptionCallbackRef.current?.(current.optedIn);
+        },
+      );
```
capJavert, line 93: When do we clean this listener? This could introduce memory leak with listener being added each time queryFn is evaluated.

packages/shared/src/contexts/PushNotificationContext.tsx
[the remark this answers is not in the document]
sshanzel: Oh, how is that? Do you maybe have some sample snippets?

packages/shared/src/hooks/notifications/useAcceptedPushNow.ts
sshanzel: We display some sort of UI but keeps the same banner. When coming back, the banner will be gone completely.

packages/shared/src/contexts/PushNotificationContext.tsx
sshanzel: Oh, right. We currently don't clean it up, but the evaluation only happens once and we ensure it by having some early checks..

packages/shared/src/contexts/PushNotificationContext.tsx
sshanzel: Right. We don't wait for anything, so we can just add this.

packages/shared/src/lib/lifecycle.ts
capJavert: `global.safari` is now added in onesignal SDK from what I can see so you can remove it from our code.

packages/shared/src/hooks/notifications/useAcceptedPushNow.ts
capJavert: Ah, ok, was thinking do we need to cleanup that value on page navigation or something? Since it will stay true until next full page refresh. If that was meant to happen then it is ok 👍

packages/shared/src/contexts/PushNotificationContext.tsx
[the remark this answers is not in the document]
capJavert: Just `throw new Error('do not use in extension`) 😆

packages/shared/src/contexts/PushNotificationContext.tsx
capJavert: Hmm, I that is a lot of assumptions. Better way would be to do `useEffect` on `OneSignal` value that query returns and then properly subscribe/unsubscribe when you want.

```js
useEffect(() => {
  if (!OneSignal) {
    return undefined;
  }

  const onChange = ({ current }) => setIsSubscribed(() => current.optedIn);

  OneSignal.User.PushSubscription.addEventListener('change', onChange);

  return () => {
    OneSignal.User.PushSubscription.removeEventListener('change', onChange);
  };
}, [OneSignal]);
```

sshanzel: Yeah, we can do that.

packages/shared/src/contexts/PushNotificationContext.tsx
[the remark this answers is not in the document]
sshanzel: Oh, thought it was something different. Yeah we can add that.

packages/shared/src/contexts/PushNotificationContext.tsx
sshanzel: I think it is more than an assumption as it is pretty straightforward. Also OneSignal throws an error whenever you initialize it more than once. From that alone, we understand that the evaluation only happened once.

packages/shared/src/lib/lifecycle.ts
sshanzel: Seems like this is a legacy code, but I removed it since it doesn't seem to be necessary anymore.

packages/shared/src/contexts/PushNotificationContext.tsx
```diff
@@ -54,10 +54,23 @@ export function PushNotificationContextProvider({
...
+          event_name: AnalyticsEvent.ClickEnableNotification,
+          extra: JSON.stringify({
+            origin: source || notificationSourceRef.current,
+            permission: 'granted',
+            ...(existing_permission && { existing_permission }),
+          }),
+        });
+      }
+    },
+  );
```
capJavert, line 74: You are going, to need to add line of `subscriptionCallbackRef.current = () => ...`, because `useRef` initial value is set only first time. You can see it in `useMutationSubscription`.

capJavert (approved):
Just one more note, looks great 🚀

packages/shared/src/contexts/PushNotificationContext.tsx
sshanzel: Ahhh. use ref. Yeah, good catch.

capJavert (approved):

sshanzel:
Also tested an extension build. Accepting through modals available for extension, works. We will merge in a few minutes

PR #2684, opened 2024-02-26
fix: update targetId for onboarding v4.5

omBratteng:
- update the target_id for events if onboarding v4.5

### DIFF
--- a/packages/shared/src/components/auth/AuthModal.tsx
+++ b/packages/shared/src/components/auth/AuthModal.tsx
@@ -72,6 +72,7 @@ export default function AuthModal({
         trigger={trigger}
         isLoginFlow={isLogoutFlow || loginState?.isLogin}
         defaultDisplay={defaultDisplay}
+        targetId={isOnboardingV4dot5 ? OnboardingV4dot5.V4dot5 : undefined}
         onDisplayChange={(display: Display) => setScreenValue(display)}
       />
     </Modal>
--- a/packages/webapp/pages/onboarding.tsx
+++ b/packages/webapp/pages/onboarding.tsx
@@ -121,7 +121,10 @@ export function OnboardPage(): ReactElement {
   );
   const onboardingOptimizations = useFeature(feature.onboardingOptimizations);
   const onboardingV4dot5 = useFeature(feature.onboardingV4dot5);
-  const targetId = ExperimentWinner.OnboardingV4;
+  let targetId: string = ExperimentWinner.OnboardingV4;
+  if (onboardingV4dot5 === OnboardingV4dot5.Control) {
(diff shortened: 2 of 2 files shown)


packages/webapp/pages/onboarding.tsx
```diff
@@ -121,7 +121,10 @@ export function OnboardPage(): ReactElement {
   );
   const onboardingOptimizations = useFeature(feature.onboardingOptimizations);
   const onboardingV4dot5 = useFeature(feature.onboardingV4dot5);
-  const targetId = ExperimentWinner.OnboardingV4;
+  let targetId: string = ExperimentWinner.OnboardingV4;
+  if (onboardingV4dot5 === OnboardingV4dot5.Control) {
+    targetId = OnboardingV4dot5.V4dot5;
+  }
```
rebelchris, line 127: We can probably do this in one go?
```suggestion
  const targetId: string = onboardingV4dot5 === OnboardingV4dot5.Control ? OnboardingV4dot5.V4dot5 : ExperimentWinner.OnboardingV4;
```

rebelchris: Not the biggest fan of lets :)

packages/shared/src/components/auth/AuthModal.tsx
```diff
@@ -72,6 +72,7 @@ export default function AuthModal({
         trigger={trigger}
         isLoginFlow={isLogoutFlow || loginState?.isLogin}
         defaultDisplay={defaultDisplay}
+        targetId={isOnboardingV4dot5 ? OnboardingV4dot5.V4dot5 : undefined}
```
rebelchris, line 75: Does this not need a fallback to v4 instead of undefined? (not entirely sure how it's used further)

rebelchris (approved):
Minor non blocking ones.

packages/shared/src/components/auth/AuthModal.tsx
omBratteng: Well, the `AuthDisplay.OnboardingSignup` is also used in the AuthenticationBanner, but that doesn't have targetId set on AuthOptions

rebelchris: Hmm yeah to many mixins going on.
I'm afraid we are missing some tracking everywhere.

rebelchris (approved):
Raised [seperate](https://dailydotdev.slack.com/archives/C02JAUF8HJL/p1708944307732589) topic on experiment values.

packages/shared/src/components/auth/AuthModal.tsx
omBratteng: Yeah, definitely possible that something is lost somewhere somehow.

PR #2687, opened 2024-02-26
feat: experiment to hide or show email login on banner post page

nensidosari:
## Changes
- adds v2 on postpageonboarding experiment
- gives the ability to hide email login on onboarding auth options for the banner

- Based on this [request](https://dailydotdev.slack.com/archives/C06K1LYU716/p1708931322030979)
Gives the ability to experiment on the post page banner with and without the email option.
It was added back after the verification got released but we’re not sure how it affects the conversion rates.

MI-187 #done


kopancek (approved):
:shipit:

packages/shared/src/lib/featureValues.ts
```diff
@@ -20,3 +20,8 @@ export enum OnboardingV4dot5 {
   Control = 'control',
   V4dot5 = 'v4dot5',
 }
+
+export enum PostPageOnboarding {
+  Control = 'control',
+  V2 = 'v2',
```
rebelchris, line 26: Why is this V2 not V1 🤔

packages/shared/src/components/auth/AuthenticationBanner.tsx
```diff
@@ -7,13 +7,17 @@ import { AuthTriggers } from '../../lib/auth';
...
+import { feature } from '../../lib/featureManagement';
+import { PostPageOnboarding } from '../../lib/featureValues';
 
 const Section = classed('div', 'flex flex-col w-[23.25rem]');
 export const authGradientBg =
   'bg-theme-bg-primary bg-gradient-to-l from-theme-overlay-active-cabbage from-0% to-theme-overlay-active-onion to-100%';
 
 export function AuthenticationBanner(): ReactElement {
   const { showLogin } = useAuthContext();
+  const postPageOnboarding = useFeature(feature.postPageOnboarding);
```
rebelchris, line 20: Why do you prop drill it and not directly access the feature on the last component?
Seems we don't really do anything else with it right?

rebelchris (commented):
Minor quesitions

packages/shared/src/lib/featureValues.ts
nensidosari: I'm not quite sure, because we already had a v1 🤷‍♀️ though to me it makes sense to have different if they run in parallel. That's why have asked here in [thread](https://dailydotdev.slack.com/archives/C06K1LYU716/p1708942468197959?thread_ts=1708931322.030979&cid=C06K1LYU716) and am waiting for confirmation.

packages/shared/src/components/auth/AuthenticationBanner.tsx
nensidosari: We do with AuthOptions. The onboarding auth options show in the banner and in the onboarding page. So if I use it in auth options we would be enrolling users landing on the onboarding page, not only post page.

packages/shared/src/lib/featureValues.ts
rebelchris: Oh right, I didn't see it here, but guess it was a experiment winner already?
Maybe we want to be careful and rather use new experiments in that case? 

It might impact already allocated users

nensidosari: We already also cleaned up for the v1 experiment, so v1 is already the default in our code. https://github.com/dailydotdev/apps/pull/2681

packages/shared/src/components/auth/AuthenticationBanner.tsx
rebelchris: Oh got it, ok :)

rebelchris (approved):

PR #2690, opened 2024-02-26
feat: add reputation badges

capJavert:
- Add badges all over the places

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

AS-138 #done


packages/shared/src/components/sidebar/SidebarUserButton.tsx
```diff
@@ -23,7 +25,15 @@ export function SidebarUserButton({
                   className="focus-outline flex h-10 cursor-pointer items-center rounded-12 border-none bg-theme-bg-secondary p-0 font-bold text-theme-label-primary no-underline typo-callout"
                 >
                   <ProfilePicture user={user} size="large" />
-                  <span className="ml-2 mr-3">{user.reputation ?? 0}</span>
+                  <ReputationUserBadge
+                    className="ml-2 mr-3 !typo-callout"
+                    user={{
+                      reputation: user.reputation ?? 0,
```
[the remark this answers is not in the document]
capJavert: Basically on the ProfileButton (and one in sidebar) it should show zero.

packages/shared/src/components/profile/ProfileButton.tsx
```diff
@@ -42,7 +44,15 @@ export default function ProfileButton({
             )}
             onClick={wrapHandler(() => onUpdate(!isOpen))}
           >
-            <span className="ml-3 block">{user.reputation ?? 0}</span>
+            <ReputationUserBadge
+              className="ml-3 !typo-callout"
```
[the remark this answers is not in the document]
rebelchris: Nah its ok, perhaps we could rather define types (big/medium) etc and apply styling but it would require a guidelined component.

Lets leave it for now 👍

packages/shared/src/components/profile/ProfileButton.tsx
```diff
@@ -42,7 +44,15 @@ export default function ProfileButton({
...
+              user={{
+                reputation: user.reputation ?? 0,
```
nensidosari, line 50: In other usages we're sending the whole user object, why only reputation here? If only reputation is needed, can we send only that in all use cases?

nensidosari (approved):
Looks good, just one non blocking comment

packages/shared/src/components/profile/ProfileButton.tsx
capJavert: `ReputationUserBadge` needs only `reputation`. The `??` was from all code but since then I removed it 898401e701e81929b3a7df7bbbdb179025306491 because it is not really needed. So overall I think cleaner now.

rebelchris: Not much to add but indeed we could have opted to have only `reputation` passed as we don't use anything else.
But either way works for me :)

rebelchris: The good part if we have `reputation` as prop we can default it to 0 :)

capJavert: I wanted to keep it open you never know when they want to show different reputation icon based on role 😆

PR #2692, opened 2024-02-27
feat: acquisition channel card

sshanzel:
## Changes
- Display the form card when the query param is available.
- Refactored the Radio component to be type-friendly.

### Screenshot
Screenshot 2024-02-27 at 6 30 28 PM

### Loom Video
https://www.loom.com/share/03fd8ea060584a0b856beed8fe25fa1f?sid=7225de36-8e10-492c-9327-8ee8efb07313

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [x] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

MI-178 #done


sshanzel:
Are we probably missing the UI for when the user is under FeedLayout v1 experiment?

packages/shared/src/components/cards/AcquisitionFormCard.tsx
```diff
@@ -0,0 +1,85 @@
...
+    onSuccess: onRemoveQueryParams,
+  });
+
+  if (isDismissed) {
+    return null;
+  }
+
+  return (
+    <Card data-testid="acquisitionFormCard" className="p-4">
+      <OnboardingTitleGradient className="flex w-full flex-row items-center whitespace-nowrap pb-1 typo-body">
```
rebelchris, line 56: Maybe rename this to be more generic?

rebelchris (changes requested):

packages/shared/src/components/cards/AcquisitionFormCard.tsx
sshanzel: Which one? The `OnboardingTitleGradient `?

packages/shared/src/components/feeds/FeedContainer.tsx
```diff
@@ -202,6 +211,7 @@ export const FeedContainer = ({
                 cardClass(isList, numCards),
               )}
             >
+              {showAcquisitionForm && <AcquisitionFormCard />}
```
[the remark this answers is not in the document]
sshanzel: Right. Good point. Let's see what we end up with in the RFC.

packages/shared/src/components/cards/AcquisitionFormCard.tsx
```diff
@@ -0,0 +1,85 @@
...
+        <Button
+          className="ml-auto"
+          size={ButtonSize.Small}
+          variant={ButtonVariant.Tertiary}
+          icon={<MiniCloseIcon />}
+          onClick={onDismiss}
+          aria-label="Close acquisition form"
+        />
+      </OnboardingTitleGradient>
+      <Radio<AcquisitionChannel>
```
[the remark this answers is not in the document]
sshanzel: What do you mean? That is to support enum values.

packages/shared/src/components/cards/AcquisitionFormCard.tsx
rebelchris: Yes :D maybe just TitleGradient would do?

packages/shared/src/components/cards/AcquisitionFormCard.tsx
```diff
@@ -0,0 +1,85 @@
...
+];
+
+export const acquisitionKey = 'ua';
+
+export function AcquisitionFormCard(): ReactElement {
+  const [isDismissed, setIsDismissed] = useState(false);
+  const [value, setValue] = useState<AcquisitionChannel>();
+  const router = useRouter();
+  const onRemoveQueryParams = () => {
+    const removed = removeQueryParam(window.location.href, acquisitionKey);
```
nensidosari, line 38: newUrl, adjustedUrl?

packages/shared/src/components/fields/Radio.tsx
```diff
@@ -46,7 +46,7 @@ export function Radio({
           disabled={option.disabled}
           checked={value === option.value}
           onChange={() => onChange(option.value)}
-          className={classNames('my-0.5 truncate', className.content)}
+          className={classNames('truncate', className.content)}
```
nensidosari, line 49: does removing this my-0.5 mess with the design of any other usage of radio?

nensidosari (commented):
Looks good. Few checks.

- Should we be adding the ua=true param in onboarding in this ticket/PR?
- If yes in the above, are we adding the experiment values in this ticket/PR as well?
- Are we adding the analytics events we need to send on dismiss/submit in this ticket/PR or other?

If we're not adding here, let's make sure we have other tickets for these. 
cc. @kopancek

packages/shared/src/components/cards/AcquisitionFormCard.tsx
kopancek: The `OnboardingTitleGradient` was already named like this before. Renaming is tangential to this PR. If we want to do it, I'd rather do it in a followup, since it's not related to changes here.

kopancek:
> Looks good. Few checks.
> 
> * Should we be adding the ua=true param in onboarding in this ticket/PR?
Yes and should be done alread

> * If yes in the above, are we adding the experiment values in this ticket/PR as well?
Yes, also should be done already

> * Are we adding the analytics events we need to send on dismiss/submit in this ticket/PR or other?
Still need to add that. And fix tests (but tests refuse to run on my machine 😢 )

packages/shared/src/components/fields/Radio.tsx
kopancek: As far as I understand, we removed the `-my-0.5` on the container div and also removed `my-0.5` here. 
The `my-0.5` was used here to add gap between elements, which is now handled by `gap-1` on the container itself instead. 

The negative margin was used on the container so that the first item did not have unnecessary y margin. There should be no effect of this change as far as I can see.

sshanzel: Most of the time, we now use `gap-` to apply the correct spacing in-between elements, which is what the margin is doing here. This should be safe.

packages/shared/src/components/feeds/FeedContainer.tsx
```diff
@@ -107,10 +113,30 @@ export const FeedContainer = ({
     insaneMode: listMode,
     loadedSettings,
   } = useContext(SettingsContext);
+  const { user, isLoggedIn } = useAuthContext();
   const { shouldUseFeedLayoutV1 } = useFeedLayout();
   const { feedName } = useActiveFeedNameContext();
   const router = useRouter();
   const onboardingV4dot5 = useFeature(feature.onboardingV4dot5);
+  const userAcquisitionVersion = useFeature(feature.userAcquisition);
```
nensidosari, line 121: Using useFeature here, it would mean we are enrolling all users landing in all feed pages (since all feed pages use feed container) to the experiment. We want to avoid that, we want to enroll only users landing on feed container only after onboarding(so on my feed page). 
In order to do this, we can check for the feature on onboarding page, where we set ua=true.

packages/shared/src/components/feeds/FeedContainer.tsx
```diff
@@ -107,10 +113,30 @@ export const FeedContainer = ({
...
+  const showAcquisitionForm =
+    isLoggedIn &&
+    feedName === SharedFeedPage.MyFeed &&
+    userAcquisitionVersion === UserAcquisition.V1 &&
```
nensidosari, line 125: Based on the above comment, this line can be removed from here

packages/webapp/pages/onboarding.tsx
```diff
@@ -158,12 +158,13 @@ export function OnboardPage(): ReactElement {
 
     return router.replace({
       pathname: '/',
-      ...(!onboardingOptimizations && {
-        query: {
+      query: {
+        ua: 'true',
```
nensidosari, line 162: Based on the above comment, just as with usage of `onboardingOptimizations`, we can use useFeature here for the acquisition and only set the query param if we are on the right version.

packages/shared/src/components/Feed.spec.tsx
```diff
@@ -1058,3 +1073,126 @@ it('should report irrelevant tags', async () => {
...
+
+describe('acquisition form card', () => {
+  const replaceRouter = jest.fn();
+  const url = new URL(`http://localhost:5002?${acquisitionKey}=true`);
+
+  let mockUserAcquisitionVersion = UserAcquisition.V1;
+
+  beforeEach(() => {
+    /* eslint-disable @typescript-eslint/no-var-requires,global-require */
+    const useRouter = jest.spyOn(require('next/router'), 'useRouter');
```
nensidosari, line 1085: U can use this way to mock useRouter to avoid the above eslint issues - https://github.com/dailydotdev/apps/blob/dfa97065f51a017db5babf2f8bda03570c6f5dda/packages/shared/src/components/SidebarRankProgress.spec.tsx#L34

packages/shared/src/components/Feed.spec.tsx
```diff
@@ -1058,3 +1073,126 @@ it('should report irrelevant tags', async () => {
...
+      configurable: true,
+    });
+  });
+
+  it('should show the form card when the right url query param is present', async () => {
+    renderComponent([createFeedMock()], defaultUser);
+    await screen.findByTestId('acquisitionFormCard');
+  });
+
+  it('should hide the card once form is submitted', async () => {
```
nensidosari, line 1126: Since the test is saying hide the card, we're not actually checking anymore if the card is shown in the end. Maybe we can either say remove query param, or add the test in the end for the card not being found.

nensidosari: expect(card).not.toBeInTheDocument();

packages/shared/src/components/Feed.spec.tsx
```diff
@@ -1058,3 +1073,126 @@ it('should report irrelevant tags', async () => {
...
+  });
+
+  it('should not show the card if the feed is different than My Feed', async () => {
+    renderComponent(undefined, undefined, SharedFeedPage.Popular);
+
+    const card = screen.queryByTestId('acquisitionFormCard');
+    expect(card).not.toBeInTheDocument();
+  });
+
+  it('should not show the card if the userAcquisition feature is in Control version', async () => {
```
nensidosari, line 1190: We usually don't add tests for experiments, but nice to see these.

nensidosari (changes requested):
Looking great, few comments, only the useFeature part concerning that would need to be changed.

Just want to check, now the feed layout mobile variation won (before this I know we agreed to keep this out of scope), did we see how this looks there, can we check if we will be adding it for the new feed layout now? No need to make the change in this ticket if we need to consider it, we can add it in another.

Ad place in mobile is on the third slot, how are we dealing with it? Still the solution can be in another ticket/PR, just raising it here.

packages/shared/src/components/cards/AcquisitionFormCard.tsx
```diff
@@ -0,0 +1,104 @@
...
+  });
+
+  useEffect(() => {
+    if (!isDismissed) {
+      trackEvent({
+        event_name: AnalyticsEvent.Impression,
+        target_type: acquisitionKey,
+      });
+    }
+  }, [isDismissed, trackEvent]);
```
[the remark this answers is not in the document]
kopancek: It can, but I wanted to be safe. Maybe a needless complication. Will change the code for more simplicity...

packages/shared/src/components/feeds/FeedContainer.tsx
kopancek: OK, so if I understand it correctly, we want to add `useFeature` hook to the onboarding page as well. That way, we make sure to enroll first, and the `useFeature` here is just going to check the enrolled value. Is that a correct assumption?

packages/shared/src/components/Feed.spec.tsx
kopancek: We decided to add the experiment gate only after these tests were already written, so I decided to modify them instead of removing.

packages/shared/src/components/feeds/FeedContainer.tsx
nensidosari: Yup, so using useFeature just in onboarding page

packages/shared/src/components/cards/AcquisitionFormCard.tsx
rebelchris: Happy with that :)
Just want to avoid not logging it somewhere 🙏

packages/shared/src/components/feeds/FeedContainer.tsx
```diff
@@ -107,10 +113,30 @@ export const FeedContainer = ({
...
+    (router.query?.[acquisitionKey] as string)?.toLocaleLowerCase() ===
+      'true' &&
+    !user?.acquisitionChannel;
+  const childItems = useMemo(() => {
```
rebelchris, line 129: Should we consider handling the logic in useFeed rather?
I'm coming from a point of the other epic where we also need to plug into this spot so would be great if it's uniform already.

That also handles memo etc already and could pretty much avoid even adding the first ad for instance. (maybe even not loading on first page? (bonus))

rebelchris (commented):
Minor request if it doesn't add too much scope would be appreciated.
Nothing otherwise blocking then Nensi's comment on moving the enrollment

packages/shared/src/components/feeds/FeedContainer.tsx
kopancek: I was considering this as well and the implementation is now in a state, where we can easily morph it that way. Let me know if you would prefer that.

packages/shared/src/components/cards/AcquisitionFormCard.tsx
kopancek: Added to our backlog: https://dailydotdev.atlassian.net/browse/MI-190

packages/shared/src/components/feeds/FeedContainer.tsx
kopancek: I went ahead and moved it to `useFeed` hook, it was quite simple to do now. Thanks for the suggestion 👍

packages/shared/src/components/Feed.spec.tsx
```diff
@@ -1058,3 +1072,108 @@ it('should report irrelevant tags', async () => {
...
+    await waitFor(() => {
+      expect(mutationCalled).toBeTruthy();
+    });
+
+    // set mocked query to false
+    mockedQuery[acquisitionKey] = 'false';
+
+    // rerender
+    unmount();
+    renderComponent([createFeedMock()], defaultUser);
```
rebelchris, line 1143: More of a note, we do have a actual re-render if you need:
https://github.com/dailydotdev/apps/blob/ea73e392446905fba18348b9d5629418b24f04ec/packages/shared/src/contexts/AnalyticsContext.spec.tsx#L270

packages/shared/src/components/Feed.tsx
```diff
@@ -137,6 +140,12 @@ export default function Feed<T>({
   const numCards = currentSettings.numCards[spaciness ?? 'eco'];
   const isSquadFeed = feedName === 'squad';
   const { shouldUseFeedLayoutV1 } = useFeedLayout();
+  const showAcquisitionForm =
+    isLoggedIn &&
```
rebelchris, line 144: Anyone on myFeed is already logged in :)

packages/shared/src/hooks/useFeed.ts
```diff
@@ -152,7 +158,9 @@ export default function useFeed<T>(
             index,
           }));
 
-          if (isAdsQueryEnabled) {
+          if (settings.showAcquisitionForm) {
+            posts.splice(adSpot, 0, { type: 'userAcquisition' });
```
rebelchris, line 162: Sweet!

rebelchris: In this case why removing it it will do a full rerender of the feed now taking ad into account? (this is more for my understanding)

packages/webapp/pages/onboarding.tsx
```diff
@@ -156,14 +157,21 @@ export function OnboardPage(): ReactElement {
       });
     }
 
+    const { id, defaultValue } = feature.userAcquisition;
+    const userAcquisitionVersion =
+      growthbook?.getFeatureValue(id, defaultValue) ?? UserAcquisition.Control;
```
rebelchris, line 162: This could just be `useFeature(` no? 
It handles all this magic internally.

rebelchris (dismissed):
Just one blocking on the feature value retrieval, and some questions more for myself.

packages/shared/src/components/Feed.spec.tsx
kopancek: Yeah, but that expects a react element to be passed in. So I would need to create another function that returns the component. Seems easier this way :)

packages/shared/src/components/Feed.tsx
kopancek: Good to know, we can remove that check then 👍

packages/shared/src/hooks/useFeed.ts
kopancek: Yes, unfortunately removing the ad will cause a full re-render with this approach. However most of the items in the array stay the same, so it should be a rerender of list, where all of the items (except one) do not get rerendered, since the items are the same reference. At least that's my understanding.

I think we had the same problem in the previous iteration of this PR - e.g. we had a children list, which we changed, so all of the feed container got rerendered. But all of the children except one were the same reference.

On trying it out locally, I didn't see any perf hiccups, but I didn't run any benchmarks or check if my assumptions above are correct.

packages/webapp/pages/onboarding.tsx
rebelchris: That seems like a engineering decision.
But yes in that case rather have a hook, or adjust the hook or something.

Basically the whole idea is that it handles the fallback itself, so actually not sure if it won't do the same already?

This is also conditional of the gb value, thus enrolls?

packages/shared/src/hooks/useFeed.ts
rebelchris: Ah cool, if it looks alright locally it's fine we can give it a go.

packages/webapp/pages/onboarding.tsx
kopancek: Or do we automatically enroll the user when GrowthBookContext is involved already? Sorry, I don't really understand how all of this works :)

rebelchris: AFAIK when you want to retrieve a feature value for someone.

This is the function we have:
```js
export function useFeatureValue<T extends JSONValue = any>(
  id: string,
  fallback: T
): WidenPrimitives<T> {
  const growthbook = useGrowthBook();
  return growthbook
    ? growthbook.getFeatureValue(id, fallback)
    : (fallback as WidenPrimitives<T>);
}
```

Thus would pretty much do the same no?

packages/shared/src/components/cards/AcquisitionFormCard.tsx
```diff
@@ -0,0 +1,110 @@
...
+      },
+    },
+  );
+
+  useEffect(() => {
+    trackEvent({
+      event_name: AnalyticsEvent.Impression,
+      target_type: acquisitionKey,
+    });
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
nensidosari, line 67: Why are we keeping eslint disable here?

nensidosari (approved):
Looks great, all good on my side. Just one small non blocking comment

packages/shared/src/components/cards/AcquisitionFormCard.tsx
kopancek: oversight counts as a reason? 😅

nensidosari: Hahah yeah, let's remove 🙏

kopancek: Hmm, so it was serving a purpose after all 😢 
```
warning  React Hook useEffect has a missing dependency: 'trackEvent'. Either include it or remove the dependency array  react-hooks/exhaustive-deps
```

Returning it back...

PR #2696, opened 2024-02-28
fix: introduce changes to profile menu

rebelchris:
- Tried to reuse as much of what we have already.
- Engineering decision: We left `Profile` option as otherwise you had no way of getting there
- Added some dynamic props to elements to change to accomodate both designs.

Screenshot 2024-02-28 at 09 53 45
Screenshot 2024-02-28 at 09 53 39

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

AS-135 #done


packages/shared/src/components/ProfileMenu.tsx
```diff
@@ -53,9 +72,6 @@ export default function ProfileMenu({
           href: `${webappUrl}devcard`,
         },
       },
-    ];
-
-    list.push(
```
rebelchris, line 58: Not sure why this was here to being with 😅

packages/shared/src/components/ProfileMenu.tsx
```diff
@@ -85,17 +101,31 @@ export default function ProfileMenu({
       onClose={onClose}
       closeOutsideClick
       position={InteractivePopupPosition.ProfileMenu}
-      className="w-full max-w-[21.25rem] !rounded-14 border border-theme-divider-tertiary laptop:max-w-[13.75rem]"
+      className="w-full max-w-64 !rounded-24 border border-theme-divider-tertiary"
```
rebelchris, line 104: The width was changed and looked nicer smaller (with the full width header)

packages/shared/src/lib/constants.ts
```diff
@@ -8,7 +8,7 @@ export const reportIssue = 'https://r.daily.dev/report-issue';
 export const termsOfService = 'https://r.daily.dev/tos';
 export const privacyPolicy = 'https://r.daily.dev/privacy-policy';
 export const cookiePolicy = 'https://r.daily.dev/cookie-policy';
-export const reputationGuide = 'https://r.daily.dev/reputation-guide';
```
rebelchris, line 11: This one was unused before :)

packages/shared/src/components/ProfileMenu.tsx
sshanzel: I think we had some items that go in-between when the user has certain feature flag. Should be good to be moved 🚀

packages/shared/src/components/tooltips/InteractivePopup.tsx
```diff
@@ -65,9 +71,14 @@ function InteractivePopup({
   position = InteractivePopupPosition.Center,
   closeOutsideClick,
   onClose,
-  closeButtonVariant = ButtonVariant.Secondary,
+  closeButton,
```
sshanzel, line 74: Just so it won't break the app when the prop is null or undefined.
```suggestion
  closeButton = {},
```

sshanzel (approved):
Great work! Looking sleek with the smaller popup :shipit:

packages/shared/src/components/tooltips/InteractivePopup.tsx
rebelchris: I wanted to pass as is to set the defaults in below load, let me know what you think?
(Idea was that you might want to override only 1 for instance) 

```js
const {
    size: buttonSize = ButtonSize.Small,
    variant: buttonVariant = ButtonVariant.Secondary,
    position: buttonPosition = 'right-2 top-2',
  } = closeButton;
```

sshanzel: Yes, that snippet would still work as intended, the issue is, JS would throw an error if you destructure the `closeButton` when it is null or undefined.

Since it's an empty object, the `size` prop for example, would be `undefined` as well, in that case, the value would be as what you defined which is `ButtonSize.Small`.

nensidosari (approved):
Looks good!

omBratteng (approved):

capJavert (approved):
Awesome

PR #2700, opened 2024-02-28
chore: remove feed layout feature - set default on mobile

nensidosari:

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done


packages/extension/src/newtab/CustomLinks.tsx
```diff
@@ -25,15 +25,15 @@ export function CustomLinks({
   className,
   onLinkClick = noop,
 }: CustomLinksProps): ReactElement {
-  const { shouldUseFeedLayoutV1 } = useFeedLayout();
+  const { shouldUseListFeedLayout } = useFeedLayout();
```
nensidosari, line 28: Please suggest any other name for this, I'm really not thinking any other, but this might confuse us with our other list view (which hopefully we'll convert into the same one...)

Or maybe I can make as shouldUseMobileFeedLayout, since this is only in mobile at the moment, and change to list only if the list view is also set to use these cards.

rebelchris: Yeah think your suggestion of mobile makes sense as it's only targeted on mobile for now.

rebelchris (approved):
Nothing to add :)

packages/extension/src/newtab/CustomLinks.tsx
capJavert: Agreed with mobile suggestion. Also with these things in the future we should just name them something without using `V1` so we don't have these massive renames when experiment succeeds.

capJavert (approved):

packages/extension/src/newtab/CustomLinks.tsx
nensidosari: Cool then, changing the names

rebelchris (approved):

capJavert (approved):

PR #2702, opened 2024-02-28
fix: flip the targetId for impression

omBratteng:
## Changes

Noticed that when in control group, the impression event sent `v4dot5` as target_id


rebelchris (approved):
Oh woops, so tracking was reversed?

omBratteng:
> Oh woops, so tracking was reversed?

Yeah 😅 Only for the first page impression

PR #2703, opened 2024-02-28
refactor: post page to track referral

sshanzel:
## Changes
- It seems as though we only have the post page that needs the hook for referral since we don't share links pointing to other places.
- Keeping this as draft until confirmed by the changes we will have on copy buttons.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

MI-189 #done


sshanzel:
Seems like it is the only place that needs the hook as discussed with Milan. cc @dailydotdev/web-team

nensidosari (approved):
Looks good

packages/shared/src/hooks/referral/useJoinReferral.ts
```diff
@@ -47,6 +47,8 @@ export const useJoinReferral = (): void => {
         }
       }
     }
+
+    return null;
```
capJavert, line 51: Why is this return `null` needed?

sshanzel: Getting an error for some reason. I think `useQuery` needs to have a return value always, probably running til the end of time if not.

https://app.circleci.com/jobs/github/dailydotdev/apps/27458

sshanzel:
> Looks good, but we also can also share user profile links, squad links, etc. Not sure if those pages don't need the same treatment?

How do we share user profile links? Do we have a button for it?

> squad links

What kind of Squad links? If invitation, that should be taken into account already.

PR #2705, opened 2024-02-29
feat: post card copy/share after upvoting

sshanzel:
## Changes
- Refactored the implementation of whether to use an Image or the VideoImage component.
  - Due to the fact that they have different props. It was a bit hard to follow which one goes to which since all the props were just spread.
  - At least, now it is more straightforward on what goes to where.
  - Also abstracts the part of how we should wrap the component itself to display the CTAs.
- Also fixed the border-radius from the image below:
Screenshot 2024-02-29 at 7 49 05 AM

Currently deployed at preview: https://preview.app.daily.dev/

Preview:
Screenshot 2024-02-29 at 9 22 17 AM
Screenshot 2024-02-29 at 7 42 30 AM

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done


packages/shared/src/hooks/post/useTrackedCopyPostLink.ts
```diff
@@ -0,0 +1,23 @@
...
+import { Post } from '../../graphql/posts';
+import { useCopyPostLink } from '../useCopyPostLink';
+import AnalyticsContext from '../../contexts/AnalyticsContext';
+import { ShareProvider } from '../../lib/share';
+import { postAnalyticsEvent } from '../../lib/feed';
+import { Origin } from '../../lib/analytics';
+
+type UseTrackedCopyPostLink = (provider: ShareProvider) => void;
+
+export const useTrackedCopyPostLink = (post: Post): UseTrackedCopyPostLink => {
```
rebelchris, line 11: I could have sworn we already had this hook 😓 
(Can't find it, but guess we at sometime need to clean up our hooks a bit)

packages/shared/src/styles/base.css
```diff
@@ -87,6 +87,7 @@ body {
   --theme-shadow2: theme('boxShadow.2-black');
   --theme-shadow3: theme('boxShadow.3-black');
 
+  --theme-blur-highlight: rgba(27, 31, 38, 0.88);
```
rebelchris, line 90: I think we already have this blur style somewhere.
Is it actually not the same as the video blur?

If not @capJavert also recently introduced one, might be the same hit.

If not the same, what color/setup is this from our guideline, might have an alternative idea.

rebelchris (changes requested):

packages/shared/src/styles/base.css
sshanzel: > Is it actually not the same as the video blur?

No, it is much darker. This feels like something new from the guideline they call blur. I also remember introducing some blur in the past, can't exactly find where it is now 😅

sshanzel: image

rebelchris: Ah yes I think this is exactly the one Ante added already then.
Let's see if we can reuse that rather else we need to check the actual values. We should not add raw hex colors for now. (its probably just a figma issue you can't see the named value)

packages/shared/src/hooks/post/useTrackedCopyPostLink.ts
sshanzel: Yep. We also need to start organizing files into folders 😅

packages/shared/src/styles/base.css
sshanzel: The reason we had to go for the raw hex colors is, that the color used here is not found in the guideline. It

rebelchris: Yes, but did you check it with design team? 🙏
It's often like 1 number off because of a glitch.

I think he introduced it on search merge?
https://github.com/dailydotdev/apps/pull/2510/files#diff-501943f96c58a477247d24220eb9c8fca482b85b5e60de358ad8e1500026ee8eR124

Not a 100% sure if it's the same, but let's see.

nensidosari: Agree here, let's try to not introduce new things unless design specifically mentions. If we can't find the colors/blur, let's pick the most similar ones we have. We can leave a note for this to design, so that they are aware and they can let us know if we need to change for a reason.

packages/shared/src/hooks/post/useTrackedCopyPostLink.ts
nensidosari: We are using this hook only here now, do we intend to replace other usages with this hook?

packages/shared/src/components/FeedItemComponent.tsx
```diff
@@ -154,6 +154,7 @@ export default function FeedItemComponent({
   onAdClick,
   onReadArticleClick,
 }: FeedItemComponentProps): ReactElement {
+  const [justUpvoted, setJustUpvoted] = useState(false);
```
nensidosari, line 157: With this being just a local state we will only be showing here if the user just upvoted, if we move away and come back this will go away, but it was mentioned we want to keep this until the user clicks.
I think it's fine this way as well as the post might not ever show again, but just checking

nensidosari (commented):
Looks great, one main concern, let's see if we can clear, and then good to merge.

packages/shared/src/styles/base.css
sshanzel: > Agree here, let's try to not introduce new things unless design specifically mentions.

When it is labeled as shared on the screenshot, that usually means it is part of our design system.

> Not a 100% sure if it's the same, but let's see.

After checking, the value used was related to `backdrop-blur` - which is a totally different setting.

packages/shared/src/hooks/post/useTrackedCopyPostLink.ts
sshanzel: I think so, but we should be careful when we do refactoring of events as it is a very delicate area.

packages/shared/src/components/cards/common/CardCover.tsx
```diff
@@ -0,0 +1,59 @@
...
+interface CardCoverProps extends CommonCardCoverProps {
+  imageProps: ImageProps;
+  videoProps?: Omit<VideoImageProps, 'imageProps'>;
+  isVideoType?: boolean;
+}
+
+export function CardCover({
+  imageProps,
+  videoProps,
+  justUpvoted,
```
capJavert, line 18: I see lots of props drilling to pull `justUpvoted` down to this component, maybe you can hook up into upvote mutatation with `useMutationSubscription` from the `CardCover` and keep the state there.

sshanzel: Updated here https://github.com/dailydotdev/apps/pull/2705/commits/249608c50c4bd49604d5bbf4944c6b21018ee376

sshanzel:
@rebelchris @nensidosari have removed the color and used opacity instead, as discussed in Slack.

nensidosari (changes requested):
Looks good in my opinion, we just need to add it behind a feature flag for experiment. 
Please use key `share_loops`. 🙏 Checking with product here if the value should be boolean or 'control' & 'v1'

packages/shared/src/components/cards/common/CardCover.tsx
```diff
@@ -0,0 +1,106 @@
...
+  imageProps,
+  videoProps,
+  isVideoType,
+  onShare,
+  post,
+}: CardCoverProps): ReactElement {
+  const { feedName } = useActiveFeedNameContext();
+  const [justUpvoted, setJustUpvoted] = useState(false);
+  const [hasInteracted, setHasInteracted] = useState(false);
+  const shareLoopsEnabled = useFeatureIsOn(feature.shareLoops);
```
nensidosari, line 34: I had never used this `useFeatureIsOn`, is this what we use for boolean flags?

packages/shared/src/components/cards/common/CardCover.tsx
```diff
@@ -0,0 +1,106 @@
...
+      }
+
+      setJustUpvoted(!!vars.vote);
+    },
+  });
+
+  if (!shareLoopsEnabled) {
+    return null;
+  }
+
```
nensidosari, line 59: Are we sure we want to do this here? We're now using card covers also for video images, in this case the video image wouldn't show no?

packages/shared/src/components/cards/common/CardCover.tsx
sshanzel: Yes. Either enabled or disabled.

packages/shared/src/components/cards/common/CardCover.tsx
sshanzel: Yes, that's the reason we made this component, so either video or image, they will utilize the same process and conditions.

nensidosari: Yup, but in this case they are set dependant on `shareLoopsEnabled`, this should only be a condition for the share overlay not normal video or image.

sshanzel: Ohhh, right right. Wrong placement. Good catch right there 😂

nensidosari (approved):
Looks good to me

packages/shared/src/components/cards/common/CardCover.tsx
capJavert: It would also return `true` even if `shareLoops` is set to `false` for the user I think.

capJavert (approved):

PR #2706, opened 2024-02-29
feat: add JSON-LD schema for post pages

DragosIorgulescu:
## Changes

- adding a script tag to attach a schema for post pages

TODO:
- [x] determine if we need to add additional attributes here
- [x] check if we need to sanitize the object since it's inserted in a script tag

## Preview

A few examples of the output schema for different articles:
- article post: https://validator.schema.org/?hl=en#url=https%3A%2F%2Fpreview.app.daily.dev%2Fposts%2FrJWMQPqO5
- video post: https://validator.schema.org/?hl=en#url=https%3A%2F%2Fpreview.app.daily.dev%2Fposts%2FBXXywqEO5

## Events

n / a

- [x] MobileL (420px)
- [x] Tablet (656px)
- [x] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

MI-163 #done


packages/webapp/pages/posts/[id]/index.tsx
```diff
@@ -132,6 +132,21 @@ const PostPage = ({ id, initialData }: Props): ReactElement => {
...
+      dangerouslySetInnerHTML={{
+        __html: JSON.stringify({
+          '@context': 'https://schema.org',
+          '@type': 'Article',
+          headline: post?.title,
+          datePublished: post?.createdAt,
+          dateModified: post?.updatedAt,
+          description: getSeoDescription(post),
+        }),
+      }}
```
Comment (DragosIorgulescu) [inline packages/webapp/pages/posts/[id]/index.tsx:147]: should we sanitize these? -- since this is user data, it can expose us to attacks

this might be a non-issue since the JSON-LD schema is just json, not executable JS, but still worth double checking

Comment (DragosIorgulescu) [reply packages/webapp/pages/posts/[id]/index.tsx:147]: @kopancek worth checking what happens if stringify raises an error

Comment (kopancek) [reply packages/webapp/pages/posts/[id]/index.tsx]: I just checked. If `JSON.stringify()` fails and throws an error in a script tag in the head ofthe HTML document, it will not "break" the page in the sense of causing a blank screen or a browser crash. It will just stop execution of the current script that has the SEO JSON. So in the worst case, we are not sending that piece to the search engine.

packages/webapp/pages/posts/[id]/index.tsx
```diff
@@ -132,6 +132,64 @@ const PostPage = ({ id, initialData }: Props): ReactElement => {
...
+    <script
+      type="application/ld+json"
+      dangerouslySetInnerHTML={{
+        __html: JSON.stringify({
+          '@context': 'https://schema.org',
+          '@type': 'Article',
+          headline: post?.title,
+          dateCreated: post?.createdAt,
+          datePublished: post?.createdAt,
+          dateModified: post?.updatedAt,
```
Comment (kopancek) [inline packages/webapp/pages/posts/[id]/index.tsx:145]: This code looks ugly. But in sake of being scrappy, I'd rather merge what we have now. We can always improve the code quality in a followup PR.

packages/webapp/pages/posts/[id]/index.tsx
```diff
@@ -132,6 +132,64 @@ const PostPage = ({ id, initialData }: Props): ReactElement => {
   };
 
   const seoComponent = <NextSeo {...seo} />;
+  const schemaComponent = (
+    <script
+      type="application/ld+json"
+      dangerouslySetInnerHTML={{
+        __html: JSON.stringify({
```
Comment (omBratteng) [inline packages/webapp/pages/posts/[id]/index.tsx:139]: Could we extract this as a function and store it in a separate file, that way we don't "overload" this file with lots of code?

Comment (capJavert) [reply packages/webapp/pages/posts/[id]/index.tsx:139]: Just wanted to write, I would even pull it in a component 👍

packages/webapp/pages/posts/[id]/index.tsx
```diff
@@ -196,6 +254,7 @@ const PostPage = ({ id, initialData }: Props): ReactElement => {
     <>
       <Head>
         <link rel="preload" as="image" href={post?.image} />
+        {schemaComponent}
```
Comment (capJavert) [inline packages/webapp/pages/posts/[id]/index.tsx:257]: We can also maybe utilize https://github.com/garmeeh/next-seo?tab=readme-ov-file#article-1

packages/webapp/pages/posts/[id]/index.tsx
Comment (kopancek) [reply packages/webapp/pages/posts/[id]/index.tsx:139]: Yup, that was my idea, but I'm running out of working time today 😅 

Hence why I commented: https://github.com/dailydotdev/apps/pull/2706#discussion_r1507598648

Comment (capJavert) [reply packages/webapp/pages/posts/[id]/index.tsx:139]: Or use `next-seo` https://github.com/dailydotdev/apps/pull/2706#discussion_r1507612410

packages/webapp/pages/posts/[id]/index.tsx
Comment (kopancek) [reply packages/webapp/pages/posts/[id]/index.tsx:257]: It is unclear to me if that library supports all of the parameters we exposed here. But definitely worth a look.

Comment (rebelchris) [reply packages/webapp/pages/posts/[id]/index.tsx:201]: Came here to comment the same on using this 😅

Comment (omBratteng) [reply packages/webapp/pages/posts/[id]/index.tsx:200]: From looking at the source, they have predefined some JSON-LD keys for article, and then just accepts `...rest`.

So I think it takes all

https://github.com/garmeeh/next-seo/blob/96b49b63caae776d3f459b7bb1b49d6cd416a241/src/jsonld/article.tsx#L50

https://github.com/garmeeh/next-seo/blob/96b49b63caae776d3f459b7bb1b49d6cd416a241/src/jsonld/jsonld.tsx#L30

Comment (DragosIorgulescu) [reply packages/webapp/pages/posts/[id]/index.tsx:200]: I can change this, but I am just curious to understand why we are pushing for using a library for something relatively simple for which we already have a pretty clear idea of how to implement. I guess I don't see the advantage of using `next-seo` for this, almost feels like an `isOdd` library, but maybe it's just me 😅

nensidosari (approved):
Looks good to me, let's wait for others approval as well just in case.

packages/webapp/pages/posts/[id]/index.tsx
Comment (rebelchris) [reply packages/webapp/pages/posts/[id]/index.tsx:200]: We already have next-seo library :) 
Kind of neater implementation atleast this is my opinion

Comment (nensidosari) [reply packages/webapp/pages/posts/[id]/index.tsx:200]: I don't personally have any strong preference, looks fine to me the way it is, just maybe that library makes it more beautiful to the eye

Comment (DragosIorgulescu) [reply packages/webapp/pages/posts/[id]/index.tsx:200]: my concern with these things is they obfuscate the underlying API, and you end up needing to lookup 2 documentations for the same thing, the official schema, and then the docs of next-seo to see how they are expecting that schema to be attached to their components. 

I'll follow the majority here, but in general my 2 cents on third-party libraries is to only use them if you really need to

packages/webapp/pages/posts/[id]/seo.ts
```diff
@@ -0,0 +1,65 @@
...
+    return post?.description;
+  }
+  return `Join us to the discussion about "${post?.title}" on daily.dev ✌️`;
+};
+
+export const getSEOJsonLd = (post: Post): string => {
+  return JSON.stringify({
+    '@context': 'https://schema.org',
+    '@type': 'Article',
+    headline: post?.title,
```
Comment (capJavert) [inline packages/webapp/pages/posts/[id]/seo.ts:17]: From all of these I think this one is the only one we need to sanitize because it comes from user.

packages/webapp/pages/posts/[id]/index.tsx
Comment (capJavert) [reply packages/webapp/pages/posts/[id]/index.tsx:200]: Agree on being basic but what I think this should end up as a our component either way so then it does not matter what we use under the hood (next-seo or our implementation) it should just accept the `post` and we can change implementation under the hood any time.

packages/webapp/pages/posts/[id]/seo.ts
Comment (capJavert) [reply packages/webapp/pages/posts/[id]/seo.ts]: I was thinking about someone adding `</script><script>alert("xss");</script>` but looks like `element.innerHtml` can not be attacked by that since the content is always a child node of `element.`

packages/webapp/pages/posts/[id]/seo.tsx
```diff
@@ -0,0 +1,86 @@
+import React, { ReactElement } from 'react';
```
Comment (capJavert) [inline packages/webapp/pages/posts/[id]/seo.tsx:1]: Is this file meant to be in `pages` since it will try to create `posts/:id/seo` page handler. Probably move it to utils?

packages/webapp/pages/posts/[id]/seo.tsx
```diff
@@ -0,0 +1,86 @@
...
+        thumbnailUrl: post?.image,
+        uploadDate: post?.createdAt,
+        duration: `PT${post?.readTime}M`,
+        url: post?.permalink,
+      },
+    }),
+  });
+};
+
+export const SEOSchema = ({ post }: { post: Post }): ReactElement => {
```
Comment (capJavert) [inline packages/webapp/pages/posts/[id]/seo.tsx:69]: Maybe `PostSEOSchema`?

capJavert (commented):
Nice, much cleaner!

packages/webapp/pages/posts/[id]/seo.tsx
Comment (capJavert) [reply packages/webapp/pages/posts/[id]/seo.tsx:1]: Maybe that is why the build is failing as well.

packages/webapp/components/PostSEOSchema.tsx
```diff
@@ -0,0 +1,86 @@
...
+        duration: `PT${post?.readTime}M`,
+        url: post?.permalink,
+      },
+    }),
+  });
+};
+
+export const PostSEOSchema = ({ post }: { post: Post }): ReactElement => {
+  return (
+    <ArticleJsonLd
```
capJavert, line 71: One last question, would it make sense to return `null` early if `post` is undefined here? Not sure if the schema would have anything useful without basic post info? 

This would happen in `isFallback` mode when pre-rendering in the background.

capJavert (approved):
Awesome, one last question! 🚀

packages/webapp/components/PostSEOSchema.tsx
```diff
@@ -0,0 +1,86 @@
...
+      '@type': 'InteractionCounter',
+      interactionType: 'https://schema.org/LikeAction',
+      userInteractionCount: post?.numUpvotes,
+    },
+    keywords: post?.tags?.join(','),
+    timeRequired: `PT${post?.readTime}M`,
+    thumbnailUrl: post?.image,
+    ...(post?.type === PostType.VideoYouTube && {
+      video: {
+        '@type': 'VideoObject',
```
rebelchris, line 57: Won't this battle with the article type above?

rebelchris (approved):
One minor question, but not blocking.
Great work 👏

packages/webapp/components/PostSEOSchema.tsx
DragosIorgulescu: this should be fine I think, `video` is an accepted property of the `Article` type -- see this https://schema.org/Article and this https://schema.org/video

nensidosari (approved):

capJavert (approved):

PR #2707, opened 2024-02-29
Chore update text

nensidosari:
- Nimrod's [request](https://dailydotdev.slack.com/archives/D05UA50PW6T/p1709111182164489) 

> I need your help with changing two minor copy things in the app (no need to experiment on that)
Anywhere we have the sign up button (see #1) it should say Sign up - Free forever
Anywhere we have the log in option (see #2) it should say Already using [daily.dev](http://daily.dev/)? Log in

SCR-20240228-kcin
SCR-20240228-kbyo

### DIFF
--- a/packages/shared/src/components/auth/AuthDefault.tsx
+++ b/packages/shared/src/components/auth/AuthDefault.tsx
@@ -175,7 +175,7 @@ const AuthDefault = ({
             text={{
               body: shouldLogin
                 ? 'Not a member yet?'
-                : 'Already a daily.dev member?',
+                : 'Already using daily.dev?',
               button: shouldLogin ? 'Sign up' : 'Log in',
             }}
             onClick={() => {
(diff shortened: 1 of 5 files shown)


packages/shared/src/components/auth/AuthDefault.tsx
```diff
@@ -175,7 +175,7 @@ const AuthDefault = ({
             text={{
               body: shouldLogin
                 ? 'Not a member yet?'
-                : 'Already a daily.dev member?',
+                : 'Already using daily.dev?',
               button: shouldLogin ? 'Sign up' : 'Log in',
```
kopancek, line 179: Should we say free forever for sign up here as well?

kopancek (approved):
Have a small question, otherwise looks good to me 👍

packages/shared/src/components/auth/AuthDefault.tsx
nensidosari: No it's only where it also says it's free

PR #2710, opened 2024-02-29
fix: change condition to show streak record modal

kopancek:
## Changes

Screenshot 2024-02-29 at 13 15 26

New streak record modal is shown whenever the currentStreak >= maxStreak

- [X] MobileL (420px)
- [X] Tablet (656px)
- [X] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

MI-192 #done

### DIFF
--- a/packages/shared/src/components/modals/streaks/NewStreakModal.tsx
+++ b/packages/shared/src/components/modals/streaks/NewStreakModal.tsx
@@ -12,11 +12,12 @@ const Paragraph = classed('p', 'text-center text-theme-label-tertiary');
 
 export default function NewStreakModal({
   currentStreak,
+  maxStreak,
   onRequestClose,
   ...props
 }: StreakModalProps): ReactElement {
   const { toggleOptOutWeeklyGoal, optOutWeeklyGoal } = useSettingsContext();
-  const shouldShowSplash = currentStreak > 20;
+  const shouldShowSplash = currentStreak >= maxStreak;
 
   return (
     <Modal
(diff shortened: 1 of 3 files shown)


packages/shared/src/components/modals/streaks/NewStreakModal.tsx
```diff
@@ -12,11 +12,12 @@ const Paragraph = classed('p', 'text-center text-theme-label-tertiary');
 
 export default function NewStreakModal({
   currentStreak,
+  maxStreak,
   onRequestClose,
   ...props
 }: StreakModalProps): ReactElement {
   const { toggleOptOutWeeklyGoal, optOutWeeklyGoal } = useSettingsContext();
-  const shouldShowSplash = currentStreak > 20;
+  const shouldShowSplash = currentStreak >= maxStreak;
```
omBratteng, line 20: Shouldn't it be `>` instead of `>=`? If your max streak was 20, and your new current streak is 20, you haven't really beaten a record?

kopancek: We generally update both values at the same time. So maxStreak is usually the same as currentStreak if you are beating it. We could use `===` as well here, but `>=` is one less character to write 😅

kopancek: See related PR
- https://github.com/dailydotdev/daily-api/pull/1726

rebelchris (approved):

nensidosari (approved):
Looks good

PR #2713, opened 2024-02-29
fix: use loading state instead of fetched to handle error state

capJavert:

On logout (or if actions failed to fetch for any reason) the `isFetched` would be `true`. Which would mean logic that relies on actions being fetched so it can use `checkHasCompleted` would get falsy value which would then trigger any kind of logic as if action was not completed.

Example on reputation popup:
- user gets logged out
- after logout redirect actions still get to be fetched once (they fail because user is no longer authenticated)
- query is in error state but isFetched is true (because query did fetch once, no matter if it failed or not)
- logic checks `ack_rep_250` was completed, gets falsy and shows the modal

Fix uses `isLoading` instead of `isFetched` because `state.loading` for `useQuery` makes sure that query was once fetched succesfully or has cached data. 
image

I think generally in 99% of the cases for `useQuery` we want to use `isLoading` instead of `isFetched`. So might be something to consider when adding new stuff. 

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done

### DIFF
--- a/packages/shared/src/hooks/useActions.ts
+++ b/packages/shared/src/hooks/useActions.ts
@@ -21,7 +21,7 @@ export const useActions = (): UseActions => {
   const client = useQueryClient();
   const { user } = useAuthContext();
   const actionsKey = generateQueryKey(RequestKey.Actions, user);
-  const { data: actions, isFetched: isActionsFetched } = useQuery(
+  const { data: actions, isLoading } = useQuery(
     actionsKey,
     async () => {
       const data = await getUserActions();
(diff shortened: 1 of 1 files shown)


nensidosari (approved):
Looks good

rebelchris (approved):
Ah this might actually be a legacy bug from the react query upgrade now that I see it 😅

PR #2722, opened 2024-03-04
fix: use a `+` icon on the new post btn when streaks are enabled

DragosIorgulescu:
## Changes

Screenshot 2024-03-04 at 14 04 15

## Events

N / A

- [x] MobileL (420px)
- [x] Tablet (656px)
- [x] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done


packages/shared/src/components/post/write/CreatePostButton.tsx
```diff
@@ -43,20 +44,34 @@ export function CreatePostButton({
     return !allowedToPost;
   };
 
-  return (
+  const href =
+    link.post.create + (squad && allowedToPost ? `?sid=${squad.handle}` : '');
+
+  const renderBtn = () => (
```
capJavert, line 50: You can make it `RenderBtn` and then `<RenderBtn />` below, more react-ish.

rebelchris (approved):
Besides Ante's comment nothing blocking from my side.

packages/shared/src/components/post/write/CreatePostButton.tsx
```diff
@@ -43,20 +44,34 @@ export function CreatePostButton({
...
+        href,
+      })}
     >
-      {!compact || isLaptop ? 'New post' : null}
+      {!compact ? 'New post' : null}
     </Button>
   );
+
+  return compact ? (
+    <LinkWithTooltip
```
nensidosari, line 68: Wouldn't it be better with conditional wrapper component? This way we can avoid the renderBtn() at all

nensidosari (commented):
Suggested different solution to the render button created 🙏

packages/shared/src/components/post/write/CreatePostButton.tsx
nensidosari: See https://github.com/dailydotdev/apps/pull/2722#discussion_r1511162601

packages/shared/src/components/post/write/CreatePostButton.tsx
rebelchris: Actually we can just render and have the tooltip disabled when it's compact :) (as another alternative)
@omBratteng Just added a feature for this.

omBratteng: See [here](https://github.com/dailydotdev/apps/pull/2714/files#diff-6eafb5899cb23318abfef2ff71f5bf4f40e4bfd3c2b00de80adf904539ae59e9) for how I did it. (add `show` prop)

```diff
@@ -13,13 +13,14 @@ const TippyTooltip = dynamicParent<TooltipProps>(
 );
 
 export function SimpleTooltip({
+  show = true,
   children,
   content,
   onTrigger,
   onShow,
   forceLoad,
   ...props
-}: TooltipProps): ReactElement {
+}: TooltipProps & { show?: boolean }): ReactElement {
   /**
    * We introduced the `shouldShow` variable to manage a re-focus issue
    * The old implementation would re-focus the tooltip whenever a modal would close
@@ -64,6 +65,10 @@ export function SimpleTooltip({
     return false;
   };
 
+  if (!show) {
+    return component;
+  }
+
   return (
     <TippyTooltip
       {...props}
@@ -78,13 +83,4 @@ export function SimpleTooltip({
   );
 }
```

packages/shared/src/components/post/write/CreatePostButton.tsx
DragosIorgulescu: updated based on @nensidosari suggestion, looking cleaner now 🙏

packages/shared/src/components/post/write/CreatePostButton.tsx
DragosIorgulescu: Since we are not using `SimpleTooltip` here, I went ahead with the wrapper approach, I kind of prefer it better too

rebelchris: Wrapper is fine in this case 👍
Just note it can cause some re render issues if you toggle evaluations during render. (Shouldnt apply here i think)

omBratteng: Could add `show` prop to `LinkWithTooltip`, and if that is false, you just return the `component`

rebelchris: I dont see any commits that change this is that correct?

nensidosari: I don't see any commits either? 🤔

DragosIorgulescu: wnot sure what happened, this should be updated now

packages/shared/src/components/post/write/CreatePostButton.tsx
```diff
@@ -43,20 +44,32 @@ export function CreatePostButton({
     return !allowedToPost;
   };
 
+  const href =
+    link.post.create + (squad && allowedToPost ? `?sid=${squad.handle}` : '');
+
+  const Wrapper = compact ? LinkWithTooltip : React.Fragment;
```
rebelchris, line 50: We have ConditionalWrapprr for this actually.
It has some side effects on rerenders but i think with your evaluation its ok

DragosIorgulescu: thx for the pointer 👍

rebelchris (approved):

nensidosari (approved):

PR #2727, opened 2024-03-05
fix: create my feed button in extension

omBratteng:
- send string for new url in extension


packages/shared/src/components/onboarding/OnboardingFeedHeader.tsx
```diff
@@ -62,12 +65,16 @@ export const OnboardingFeedHeader = ({
...
+    return router.replace(
+      isExtension
+        ? `${window.location.href}?welcome=true`
+        : {
+            pathname: router.route,
+            query: {
+              welcome: 'true',
+            },
+          },
+    );
```
omBratteng, line 77: Alternatively we could update `CustomRouter` to accept objects and deconstruct them when a object is passed.
https://github.com/dailydotdev/apps/blob/50192c8ce187519e01144e506a17cfbcce8f91af/packages/extension/src/lib/CustomRouter.ts#L56-L59

omBratteng: Will re-do it and use the same util that next.js uses (and exports)
https://github.com/vercel/next.js/blob/473ef817a77c1103e034cb3dcb1c8b10c1bc9e9d/packages/next/src/shared/lib/router/utils/format-url.ts#L29-L73

nensidosari: Just checking, there is another place where we set welcome=true, would that need the same fix for the extension? If yes, does it make sense to do it in this PR?

omBratteng: The other place is on the onboarding page, which does not exist in the extension.
But if I implement formatUrl from next into our CustomRouter, it wouldn't be an issue anyways

nensidosari (approved):

omBratteng:
For some reason I am unable to get the extension to work locally, so this change has not been tested.

packages/shared/src/components/onboarding/OnboardingFeedHeader.tsx
omBratteng: I implemented it, but it didn't really have any good effect. It worked fine as intended, it set the `?welcome=true` when I asked it to. But it doesn't persist in `router.query`.

Had a [chat with Sab](https://dailydotdev.slack.com/archives/C068F77M8Q7/p1709639575360589?thread_ts=1709636486.310309&cid=C068F77M8Q7), and the easiest way is just to send them to webapp and not spend too much time, as we might remove this welcome banner after a future experiment anyways.
And it's a edge case flow too.

packages/shared/src/components/onboarding/OnboardingFeedHeader.tsx
```diff
@@ -62,13 +68,25 @@ export const OnboardingFeedHeader = ({
...
-      pathname: router.route,
-      query: {
-        welcome: 'true',
-      },
-    });
-  }, [registerLocalFilters, router, trackEvent, updateAlerts]);
+    if (!onboardingOptimizations) {
+      router.replace(
+        isExtension
+          ? 'https://app.daily.dev/?welcome=true'
```
rebelchris, line 74:
```suggestion
          ? `${webappUrl}/?welcome=true`
```

rebelchris (changes requested):
One minor request

packages/shared/src/components/onboarding/OnboardingFeedHeader.tsx
omBratteng: Done. I knew there was something like that, but I didn't find it when I looked 😬😓

rebelchris (approved):

packages/extension/src/lib/CustomRouter.ts
```diff
@@ -54,7 +55,7 @@ export default class CustomRouter implements NextRouter {
   }
 
   async replace(url: Url): Promise<boolean> {
-    window.location.href = url as string;
+    window.location.href = typeof url === 'object' ? formatUrl(url) : url;
     return true;
```
capJavert, line 59: We don't need this changes anymore since we redirect to app?

capJavert (approved):

packages/extension/src/lib/CustomRouter.ts
omBratteng: Not really

omBratteng: Should I revert it?

capJavert: I would say yes, just to minimize code changes

omBratteng: Removed.

nensidosari (approved):

PR #2729, opened 2024-03-05
fix: right blue color for link

rebelchris:
- We need to overwrite to link color as there's a default color
- We already check manually if tag == A, but in this case it's not.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done

### DIFF
--- a/packages/shared/src/components/cards/ShowMoreContent.tsx
+++ b/packages/shared/src/components/cards/ShowMoreContent.tsx
@@ -42,7 +42,7 @@ export default function ShowMoreContent({
         {getContent()}{' '}
         {displayShowMoreLink() && (
           <ClickableText
-            className="inline-flex text-theme-label-link"
+            className="inline-flex !text-theme-label-link"
             onClick={toggleTextExpanded}
           >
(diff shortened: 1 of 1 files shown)


capJavert (approved):

nensidosari (approved):

packages/shared/src/components/cards/ShowMoreContent.tsx
```diff
@@ -42,7 +42,7 @@ export default function ShowMoreContent({
         {getContent()}{' '}
         {displayShowMoreLink() && (
           <ClickableText
-            className="inline-flex text-theme-label-link"
+            className="inline-flex !text-theme-label-link"
```
kopancek, line 45: Do we need to do anything with hover and focus states? Since this now trumps all...

rebelchris: Think they always keep being this
the logic was already there.

PR #2734, opened 2024-03-05
feat: post modal share loops

sshanzel:
## Changes
- Refactored the logic to be inside a hook to determine share loops visibility.
- Add the same process on post modal/page.

Preview (note: color of the border is a [separate issue entirely](https://dailydotdev.slack.com/archives/C01L0QXQJNB/p1709648963630299)):
Screenshot 2024-03-05 at 10 23 40 PM

## Events

| Change/New | share post  | extra: { origin: 'post content' } |

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

MI-185 #done


nensidosari (commented):
Looks great, just the concern in analytics event. Let's make sure that is correct, then ready to approve.

packages/shared/src/components/post/common/PostContentShare.tsx
```diff
@@ -0,0 +1,33 @@
...
+  post,
+}: PostContentShareProps): ReactElement {
+  const { shouldShowOverlay, onInteract } = usePostShareLoop(post);
+
+  if (!shouldShowOverlay) {
+    return null;
+  }
+
+  return (
+    <div className="border-theme-subtlest-tertiary mt-6 flex flex-row items-center rounded-16 border px-4 py-2">
```
rebelchris, line 21:
```suggestion
    <div className="mt-6 flex flex-row items-center rounded-16 border border-border-subtlest-tertiary px-4 py-2">
```

packages/shared/src/hooks/post/usePostShareLoop.ts
```diff
@@ -0,0 +1,42 @@
...
+interface UsePostShareLoop {
+  shouldShowOverlay: boolean;
+  onInteract: () => void;
+}
+
+export const usePostShareLoop = (post: Post): UsePostShareLoop => {
+  const { feedName } = useActiveFeedNameContext();
+  const [justUpvoted, setJustUpvoted] = useState(false);
+  const [hasInteracted, setHasInteracted] = useState(false);
+  const shareLoopsEnabled = useFeatureIsOn(feature.shareLoops);
```
rebelchris, line 18:
```suggestion
  const shareLoopsEnabled = useFeature(feature.shareLoops);
```

rebelchris (changes requested):
Minor changes 🙏

packages/shared/src/lib/analytics.ts
```diff
@@ -169,6 +169,7 @@ export enum TargetId {
   GenericReferralPopup = 'generic referral popup',
   ProfilePage = 'profile page',
   InviteFriendsPage = 'invite friends page',
+  PostModal = 'post modal',
```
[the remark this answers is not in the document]
sshanzel: This is not an additional event, this is a target id to let them know where did the click of the copy.

sshanzel:
Screenshot 2024-03-05 at 11 28 46 PM

rebelchris (approved):
Looks good

nensidosari (approved):
Looks great!

packages/shared/src/hooks/post/usePostShareLoop.ts
```diff
@@ -0,0 +1,42 @@
...
+export const usePostShareLoop = (post: Post): UsePostShareLoop => {
+  const { feedName } = useActiveFeedNameContext();
+  const [justUpvoted, setJustUpvoted] = useState(false);
+  const [hasInteracted, setHasInteracted] = useState(false);
+  const shareLoopsEnabled = useFeature(feature.shareLoops);
+  const shouldShowOverlay = justUpvoted && !hasInteracted && shareLoopsEnabled;
+
+  useMutationSubscription({
+    matcher: ({ status, mutation }) => {
+      const key = [...upvoteMutationKey, { feedName }];
```
kopancek, line 23: Should we define the `key` variable out of the matcher function? Otherwise we kind of keep redefining this every time the `matcher` function is called.

packages/shared/src/hooks/post/usePostShareLoop.ts
```diff
@@ -0,0 +1,42 @@
...
+
+      return (
+        status === 'success' &&
+        mutation?.options?.mutationKey?.toString() === key.toString()
```
kopancek, line 27: `key` is not used anywhere else and here we conver it with `toString()`. Seems like it would be more efficient to assign `key` with the `.toString()` already executed. That way we only do the conversion once on the assignment.

kopancek (commented):
The new code is much easier to read 👍 🙇 

Do we need to make the same changes to the post page or do we already have share loops implemented there?

Added a few non-blocking suggestions on how we can make it even better.

packages/shared/src/hooks/post/usePostShareLoop.ts
sshanzel: It is not working, so I had to fix it.

sshanzel: Yeah, we can introduce a memoized variable outside.

packages/shared/src/hooks/post/usePostShareLoop.ts
sshanzel: I think this is probably to keep it easier to understand that both of them should be converted and not be missed. Wdyt? Should we add the `toString` on the memoized variable?

sshanzel:
> Do we need to make the same changes to the post page or do we already have share loops implemented there?

The `PostEngagements` component is used on every modal/page, so we should be good to go here 🙌

nensidosari (approved):
Looks good on my side. If u don't ming please add the additional event with origin in the PR description

PR #2735, opened 2024-03-05
fix: overflow on public squad post page and upvote modal

omBratteng:
- fix overflow issue in upvoted modal
- fix overflow issue on public squad post page and modal

<table>
<tr>
 <td>Before
 <td>After
<tr>
 <td>
  
  
  
  
 <td>
  
  
  
  
</table>

### DIFF
--- a/packages/shared/src/components/post/SquadPostContent.tsx
+++ b/packages/shared/src/components/post/SquadPostContent.tsx
@@ -95,7 +95,7 @@ function SquadPostContent({
       >
         <div
           className={classNames(
-            'relative max-w-[42.5rem] px-4 tablet:px-8',
+            'relative min-w-0 px-4 tablet:px-8',
             className?.content,
             isPublicSquad && 'flex flex-1 flex-col',
           )}
--- a/packages/shared/src/components/post/SquadPostContent.tsx
+++ b/packages/shared/src/components/post/SquadPostContent.tsx
@@ -95,7 +95,7 @@ function SquadPostContent({
       >
         <div
           className={classNames(
-            'relative px-4 tablet:px-8',
+            'relative max-w-[42.5rem] px-4 tablet:px-8',
             className?.content,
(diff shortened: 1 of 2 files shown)


rebelchris (approved):
Perhaps we should add some user with super long handles to our seeds 👀👀👀 will help identify early

omBratteng:
Good idea, I'll note it down to add it.
Already found another edge case with really long usernames in public squad posts 😬

packages/shared/src/components/post/SquadPostContent.tsx
```diff
@@ -95,7 +95,7 @@ function SquadPostContent({
       >
         <div
           className={classNames(
-            'relative px-4 tablet:px-8',
+            'relative max-w-[42.5rem] px-4 tablet:px-8',
```
omBratteng, line 98: I took `42.5rem` as that is the width of `pageMaxWidth`

rebelchris: Hmm why do we need it?
Cant it just set to full?

omBratteng: No, full doesn't work 😞

omBratteng: The parent div already have a `!w-[42.5rem]` applied to it, but when it is a squad post, it also gets `flex-1` added, which then disregards the width of the element.

omBratteng: Playing with it some more, setting `min-w-0` helped too, and it fixed another overflow issue on smaller screens.

rebelchris: Lol we might have to start applying that mind w fix on *

rebelchris (approved):

nensidosari (approved):

PR #2736, opened 2024-03-06
feat: onboarding loading

sshanzel:
## Changes
- Animation before redirecting the user.
- Preview on Slack: https://dailydotdev.slack.com/archives/C06K5KAHM0X/p1709693669923399

Currently deployed at preview3.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

MI-205 #done

### DIFF
--- a/packages/shared/src/components/auth/PreparingYourFeed.tsx
+++ b/packages/shared/src/components/auth/PreparingYourFeed.tsx
@@ -5,7 +5,6 @@ import LogoText from '../../svg/LogoText';
 import classed from '../../lib/classed';
 import { useFeature } from '../GrowthBookProvider';
 import { feature } from '../../lib/featureManagement';
-import { OnboardingAnimation } from '../../lib/featureValues';
 import { Loader } from '../Loader';
 
 interface PreparingYourFeedProps {
@@ -21,10 +20,8 @@ export function PreparingYourFeed({
   isAnimating,
 }: PreparingYourFeedProps): ReactElement {
   const onboardingAnimation = useFeature(feature.onboardingAnimation);
-  const isAnimationControl =
-    onboardingAnimation === OnboardingAnimation.Control;
 
-  if (isAnimationControl) {
+  if (!onboardingAnimation) {
     return (
       <OnboardingContainer className="justify-center typo-title2">
         <Loader innerClassName="before:border-t-theme-color-cabbage after:border-theme-color-cabbage typo-title2" />
--- a/packages/shared/src/components/auth/PreparingYourFeed.tsx
+++ b/packages/shared/src/components/auth/PreparingYourFeed.tsx
@@ -10,7 +10,7 @@ interface PreparingYourFeedProps {
 
 export const OnboardingContainer = classed(
   'div',
-  'flex flex-col overflow-x-hidden items-center min-h-[100vh] w-full h-full max-h-[100vh] flex-1 z-max',
+  'flex flex-col overflow-x-hidden items-center min-h-screen w-full h-full max-h-screen flex-1 z-max',
 );
 
 export function PreparingYourFeed({
@@ -22,7 +22,7 @@ export function PreparingYourFeed({
         <LogoIcon className="w-16" />
         <LogoText className="w-32" />
       </span>
-      <span className="flex h-1.5 w-full max-w-[19.125rem] flex-row items-center rounded-12 bg-theme-divider-tertiary px-0.5">
+      <span className="flex h-1.5 w-full max-w-[19.125rem] flex-row items-center rounded-12 bg-border-subtlest-tertiary px-0.5">
         <span
           className={classNames(
             'relative flex h-0.5 w-0 flex-row items-center bg-theme-color-cabbage transition-[width] duration-[2000ms] ease-in-out',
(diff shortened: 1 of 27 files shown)


packages/shared/src/components/auth/PreparingYourFeed.tsx
```diff
@@ -0,0 +1,40 @@
...
+import LogoText from '../../svg/LogoText';
+import classed from '../../lib/classed';
+
+interface PreparingYourFeedProps {
+  isAnimating: boolean;
+}
+
+export const OnboardingContainer = classed(
+  'div',
+  'flex flex-col overflow-x-hidden items-center min-h-[100vh] w-full h-full max-h-[100vh] flex-1 z-max',
```
rebelchris, line 13: Just wondering did we not fix these 100vh to be able to use tailwinds ones?

packages/shared/src/components/auth/PreparingYourFeed.tsx
```diff
@@ -0,0 +1,40 @@
...
+export function PreparingYourFeed({
+  isAnimating,
+}: PreparingYourFeedProps): ReactElement {
+  return (
+    <OnboardingContainer className="justify-center">
+      <span className="mb-10 flex flex-row gap-3">
+        <LogoIcon className="w-16" />
+        <LogoText className="w-32" />
+      </span>
+      <span className="flex h-1.5 w-full max-w-[19.125rem] flex-row items-center rounded-12 bg-theme-divider-tertiary px-0.5">
```
rebelchris, line 25: Do we want to maybe start using the new named colors here already?

rebelchris (commented):
Main question, should this not be an experiment?

nensidosari (changes requested):
Looks great, please take in consideration Chris comments 🙏 

Last thing from me is, please move all changes behind `onboarding_fake_loading` feature flag - boolean value.

packages/shared/src/components/auth/PreparingYourFeed.tsx
sshanzel: Yes, it was just not working when I was working with this, so I applied the correct working one.

packages/shared/src/components/auth/PreparingYourFeed.tsx
sshanzel: I just copy-pasted this from the other file, but sure I will check it.

sshanzel:
> Main question, should this not be an experiment?

Yes, I just received a reply from Sab.

packages/webapp/pages/onboarding.tsx
```diff
@@ -394,12 +397,7 @@ export function OnboardPage(): ReactElement {
   }
 
   if (finishedOnboarding) {
-    return (
```
nensidosari, line 397: Did we ever use this return?

packages/shared/src/lib/featureManagement.ts
```diff
@@ -21,6 +22,10 @@ export class Feature<T extends JSONValue> {
 
 const feature = {
   feedVersion: new Feature('feed_version', 15),
+  onboardingAnimation: new Feature(
+    'onboarding_animation',
+    OnboardingAnimation.Control,
+  ),
```
nensidosari, line 28: I think we missed this comment of mine 😬 https://github.com/dailydotdev/apps/pull/2736#pullrequestreview-1918945730. Would u mind changing the naming and values, else I can delete and create a new one in growthbook as u've created this.

packages/webapp/pages/onboarding.tsx
nensidosari: Resolved now

sshanzel: The old one? It should be part of the new component already.

packages/shared/src/lib/featureManagement.ts
nensidosari: I believe versions are mostly valuable only on a/b testings, when we use them usually as follow up experiments doesn't make much sense as we set the winning version as default (not using the feature at all) and then the other as the next variation - which can be fine just as a true boolean value.
Plus this is a micro experiment so I don't think we have anything we need to relate with versions

nensidosari: Checking with product just in case, anyways we should be ok either way.

sshanzel: I mean this is an a/b testing, animations can easily be made, which means there can be follow-up experiments to this if they suddenly think of something new. But sure, let's go for boolean.

nensidosari: I don't know, don't think we use versions for most of them in their full potential. But yeah this doubt I think is also coming from not having any real guidance on how we do these. 🙌 Thanks for resolving

nensidosari (approved):

packages/shared/src/lib/featureManagement.ts
sshanzel: Agree with that, we can maybe write a guideline in the future on how we define variables based on certain assessments.

PR #2737, opened 2024-03-06
fix: add tracking event for 404 page

rebelchris:
- Requested by Dave

Screenshot 2024-03-06 at 10 55 30

## Events

|New | 404 page  |  |

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done

### DIFF
--- a/packages/shared/src/lib/analytics.ts
+++ b/packages/shared/src/lib/analytics.ts
@@ -132,6 +132,8 @@ export enum AnalyticsEvent {
   GenerateDevcard = 'generate devcard',
   DownloadDevcard = 'download devcard',
   CopyDevcardCode = 'copy devcard code',
+  // 404 page
+  View404Page = '404 page',
 }
 
 export enum FeedItemTitle {
--- a/packages/webapp/pages/404.tsx
+++ b/packages/webapp/pages/404.tsx
@@ -1,4 +1,4 @@
-import React, { ReactElement, useContext, useEffect, useState } from 'react';
+import React, { ReactElement, useContext, useEffect, useRef } from 'react';
 import Custom404 from '@dailydotdev/shared/src/components/Custom404';
 import { NextSeo } from 'next-seo';
 import { AnalyticsEvent } from '@dailydotdev/shared/src/lib/analytics';
@@ -6,17 +6,17 @@ import AnalyticsContext from '@dailydotdev/shared/src/contexts/AnalyticsContext'
 
 export default function Custom404Seo(): ReactElement {
   const { trackEvent } = useContext(AnalyticsContext);
-  const [trackedImpression, setTrackedImpression] = useState(false);
+  const trackedImpression = useRef(false);
 
   useEffect(() => {
-    if (trackedImpression) {
+    if (trackedImpression.current) {
       return;
     }
 
     trackEvent({
       event_name: AnalyticsEvent.View404Page,
     });
-    setTrackedImpression(true);
+    trackedImpression.current = true;
   }, [trackEvent, trackedImpression]);
 
   return (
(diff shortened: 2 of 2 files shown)


packages/webapp/pages/404.tsx
```diff
@@ -1,8 +1,24 @@
...
+  useEffect(() => {
+    if (trackedImpression) {
+      return;
+    }
+
+    trackEvent({
+      event_name: AnalyticsEvent.View404Page,
+    });
+    setTrackedImpression(true);
+  }, [trackEvent, trackedImpression]);
```
nensidosari, line 20: Can't we do this just on mount? Or trackEvent changes somehow?

nensidosari (approved):
Looks good, just one small check.

packages/webapp/pages/404.tsx
```diff
@@ -1,8 +1,24 @@
-import React, { ReactElement } from 'react';
+import React, { ReactElement, useContext, useEffect, useState } from 'react';
 import Custom404 from '@dailydotdev/shared/src/components/Custom404';
 import { NextSeo } from 'next-seo';
+import { AnalyticsEvent } from '@dailydotdev/shared/src/lib/analytics';
+import AnalyticsContext from '@dailydotdev/shared/src/contexts/AnalyticsContext';
 
 export default function Custom404Seo(): ReactElement {
+  const { trackEvent } = useContext(AnalyticsContext);
+  const [trackedImpression, setTrackedImpression] = useState(false);
```
capJavert, line 9: You can use ref for this. Less re-renders.

packages/shared/src/lib/analytics.ts
```diff
@@ -132,6 +132,8 @@ export enum AnalyticsEvent {
   GenerateDevcard = 'generate devcard',
   DownloadDevcard = 'download devcard',
   CopyDevcardCode = 'copy devcard code',
+  // 404 page
+  View404Page = '404 page',
```
capJavert, line 136: Are we fixing 404 on profile page that we mentioned or is that out of scope for now?

packages/webapp/pages/404.tsx
rebelchris: Mainly to avoid re-render triggers we always do the page view ones like this :)

packages/shared/src/lib/analytics.ts
rebelchris: The change is out of scope, this will track all implementations of 404 showing though.
Regardless of post 404 or profile one :)

packages/webapp/pages/404.tsx
nensidosari: 🤷‍♀️ yeah I've seen it a lot, not 100% ok with it personally in the idea, don't see why it would re-render, feels like either some extra unnecessary layer/code added, or we just haven't written our hook well.

But yeah ok with it for this, we can recheck the whole usage of these in a separate refactor if/when time comes.

capJavert (approved):

packages/webapp/pages/404.tsx
rebelchris: We don't have a hook is the problem :) 
It's a backlog ticket somewhere to have a custom once off tracking hook, but think we waiting for React 18 hooks :")

capJavert: > don't see why it would re-render

it is because `trackEvent` does not have stable reference but we need it as dependency when in `useEffect`, so to make sure we are really only sending one event per component lifecycle we have this hack. The new hook [useEventEffect](https://react.dev/learn/separating-events-from-effects#declaring-an-effect-event) in future React allows to create handlers that have stable reference and are not required as dependencies.

PR #2742, opened 2024-03-06
fix: don't show pinned posts if not a member

omBratteng:
- hide pinned posts button on squad page if they are not a member

<table>
<tr>
 <td>Before
 <td>After
<tr>
 <td>
 <td>
</table>

AS-159


nensidosari (approved):

rebelchris (approved):
Can't recall but it doesn't work for non members or only non users?

omBratteng:
> Can't recall but it doesn't work for non members or only non users?

Both. For non-users you just get errors in the console. For non-members, the button sometimes changes, but nothing else happens

PR #2744, opened 2024-03-06
feat: add events tracking for streaks related actions

DragosIorgulescu:
**Open streaks event**
---
Screenshot 2024-03-06 at 13 32 08

**Milestone impression event**
---
Screenshot 2024-03-06 at 13 33 23

**Milestone dismiss event**
---
Screenshot 2024-03-06 at 13 33 44

## Events

Events added as per taxonomy sheet - streaks tab: https://docs.google.com/spreadsheets/d/18Lv7zXges9QfVX5VYL1a-Hyl0e1sQ3sLr0OK8YZWKXI/edit#gid=1148653887

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

MI-206 #done


packages/shared/src/components/streak/ReadingStreakButton.tsx
```diff
@@ -47,11 +49,20 @@ function CustomStreaksTooltip({
...
   const hasReadToday =
     new Date(streak.lastViewAt).getDate() === new Date().getDate();
 
+  useEffect(() => {
+    if (shouldShowStreaks) {
+      trackEvent({
+        event_name: AnalyticsEvent.OpenStreaks,
+      });
+    }
+  }, [shouldShowStreaks, trackEvent]);
```
nensidosari, line 64: We usually do some checks on these cases when tracking events to make sure this function doesn't cause any re-render and resend of track event. Just making sure we're fine like this. 
This is what I would usually expect, but commenting since I've seen it being checked more.

packages/shared/src/hooks/streaks/useStreakMilestone.ts
```diff
@@ -42,5 +45,11 @@ export const useStreakMilestone = (): void => {
         },
       },
     });
-  }, [modalType, openModal, shouldHideModal, streak, updateAlerts]);
+
+    trackEvent({
```
nensidosari, line 49: When is this supposed to be tracked. I see this in analytics but doesn't make much sense to me `When the shortcut component renders`. Also left a comment for Dave there

packages/shared/src/components/streak/ReadingStreakButton.tsx
rebelchris: Yeah think in this case on click makes most sense actually 🙏

packages/shared/src/hooks/streaks/useStreakMilestone.ts
rebelchris: Probably, when the streaks visual element renders.
But it's pretty much the same as the click then? 👀

nensidosari: I was thinking when the button renders, then it wouldn't be the same as we click one. But not sure, maybe it's for any other modal from streaks showing. Let's confirm with Dave 🙏

rebelchris: Yeah after looking I think it's the impression for the modal.
As he handles the clicks/dismissal seperate.

packages/shared/src/components/streak/ReadingStreakButton.tsx
DragosIorgulescu: Fair point, thought about using a click handler initially but since `trackEvent` is attached to a context, use effect would not trigger re-renders based on it here. I'll change it to your suggestion 👍

packages/shared/src/hooks/streaks/useStreakMilestone.ts
nensidosari: Makes sense from the descriptions as well. Just here concerned about what I mentioned above then since this is in a useEffect.

DragosIorgulescu: @nensidosari I'm not sure we have an alternative to a side effect here, since these modals are displayed on layout mount and not on a click handler -- see this: https://github.com/dailydotdev/apps/blob/083c4ce047accfd19645978bc0ffc4b47b941bd1/packages/shared/src/components/MainLayout.tsx#L91

An alternative would be to track the event on the `FirstStreakModal` and `NewStreakModal` component, I chose to do it here to avoid duplicating code.

rebelchris: You can probably add a ref just to see if it already fired? 🙏

See [here](https://github.com/dailydotdev/apps/commit/cb61c985acf3a93b2ef9877eb3f65ec4e3320ccb)

nensidosari (approved):
Looks good to me

rebelchris (approved):
Cool, thanks!

PR #2751, opened 2024-03-07
fix: stale time by updating cache

sshanzel:
## Changes
- Introduced a hook to organize the logic when viewing a post.
- Fix the un-updated icon and count by manually adding the counts.
- Note: I already tried invalidating the cache to force a refetch and didn't work. I also tried to manually run the `refetch` function from the query that handles the counts, also failed. We have no other choice but to either remove the stale time or manually update the count.

https://github.com/dailydotdev/apps/assets/13744167/31cf6999-6e61-42b3-bf9c-487b3eb1ba10

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done


nensidosari (commented):
Looks good, just one check on my side.

DragosIorgulescu (commented):
Great work, other than the timezone consideration, looks good

packages/shared/src/hooks/post/useViewPost.ts
```diff
@@ -0,0 +1,60 @@
...
+
+        streak.lastViewAt = new Date();
+
+        return streak;
+      });
+
+      const reading = client.getQueryData<ReadingDay[]>(readKey);
+
+      if (reading) {
+        const date = new Date().toISOString().split('T')[0];
```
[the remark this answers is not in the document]
sshanzel: Good catch! Should be fixed now 🚀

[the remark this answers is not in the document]
nensidosari: Do we have anything against using any date formatter like datefns rather then doing this ourselves?

packages/shared/src/hooks/post/useViewPost.ts
```diff
@@ -0,0 +1,62 @@
...
+        streak.lastViewAt = new Date();
+
+        return streak;
+      });
+
+      const reading = client.getQueryData<ReadingDay[]>(readKey);
+
+      if (reading) {
+        const timezoned = getTodayTz(user?.timezone || 'UTC');
+        const date = new Date(timezoned).toISOString().split('T')[0];
```
nensidosari, line 45: Can't we directly do `timezoned.toISOString().split('T')[0];` ? Seems like what the above function return is already a `new Date(timeZonedToday)`

As mentioned above, do we have anything against using the date libraries (datefns)? Definitely think it can be safer to use

sshanzel: This is what we use in our API, so I think it is better to match it.

> Can't we directly do timezoned.toISOString().split('T')[0]; ? Seems like what the above function return is already a new Date(timeZonedToday)

Oh, right. Yeah, we can do that.

sshanzel, line 45:
```suggestion
        const date = timezoned.toISOString().split('T')[0];
```

nensidosari (approved):
All good on my side

DragosIorgulescu (approved):
🙏

PR #2753, opened 2024-03-07
fix: placeholder for streaks button

sshanzel:
## Changes
- Placeholder for reading streaks button.

Sample size side-by-side:
Screenshot 2024-03-07 at 12 28 40 PM

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

MI-209 #done

### DIFF
--- a/packages/shared/src/components/layout/MainLayoutHeader.tsx
+++ b/packages/shared/src/components/layout/MainLayoutHeader.tsx
@@ -85,10 +85,22 @@ function MainLayoutHeader({
     });
   };
 
+  const getStreakButton = () => {
+    if (isLoading) {
+      return <div className="h-10 w-20 rounded-12 bg-surface-float" />;
+    }
+
+    if (!isStreaksEnabled || !streak) {
+      return null;
+    }
+
+    return <ReadingStreakButton streak={streak} />;
+  };
+
   const RenderButtons = () => {
     return (
       <div className="flex gap-3">
-        {streak && <ReadingStreakButton streak={streak} />}
+        {getStreakButton()}
         <CreatePostButton compact={isStreaksEnabled} />
         {!hideButton && user && (
           <>


rebelchris (approved):

packages/shared/src/components/layout/MainLayoutHeader.tsx
```diff
@@ -85,10 +85,22 @@ function MainLayoutHeader({
...
+    }
+
+    return <ReadingStreakButton streak={streak} />;
+  };
+
   const RenderButtons = () => {
     return (
       <div className="flex gap-3">
-        {streak && <ReadingStreakButton streak={streak} />}
+        {getStreakButton()}
```
rebelchris, line 103: As @capJavert would say, we can make it more react like by doing `StreakButton` as a inline component 😅

rebelchris (approved):

nensidosari (approved):
Looks good, but agree with the above comment ⬆️ 😄

packages/shared/src/components/layout/MainLayoutHeader.tsx
```diff
@@ -41,6 +42,24 @@ const SearchPanel = dynamic(
...
+}
+
+const StreakButton = ({ streak, isEnabled, isLoading }: StreakButtonProps) => {
+  if (isLoading) {
+    return <div className="h-10 w-20 rounded-12 bg-surface-float" />;
+  }
+
+  if (!isEnabled || !streak) {
+    return null;
+  }
```
omBratteng, line 58: Just wondering if this should be the first one? So could technically see a loader but have no streak?

sshanzel: Yeah, thought of that too, but then it will never be `isLoading` since in the hook itself, we check whether the query should be enabled or not. Hence, will not get the loading if it is disabled.

omBratteng (approved):

PR #2758, opened 2024-03-07
feat: add share tracking URL query parameters and shorten the links

kopancek:
## Changes

Adds `userId` and `cid` query params to all shared URLs.
Minifies the URL before sharing.
Modifies `copyLink` to send the same URL as well, in case `copyLink` feature flag is on. 

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [X] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

MI-188 #done

### DIFF
--- a/packages/shared/src/components/ShareBar.spec.tsx
+++ b/packages/shared/src/components/ShareBar.spec.tsx
@@ -1,4 +1,10 @@
-import { render, RenderResult, screen, waitFor } from '@testing-library/react';
+import {
+  fireEvent,
+  render,
+  RenderResult,
+  screen,
+  waitFor,
+} from '@testing-library/react';
 import { QueryClient, QueryClientProvider } from '@tanstack/react-query';
 import React from 'react';
 import nock from 'nock';
@@ -116,7 +122,7 @@ describe('ShareBar Test Suite:', () => {
     renderComponent();
     const btn = await screen.findByTestId(`social-share-Facebook`);
 
-    btn.click();
+    fireEvent.click(btn);
 
     await waitFor(() => {
       expect(mockWindowOpen).toHaveBeenCalledWith(
--- a/packages/shared/src/components/ShareBar.spec.tsx
+++ b/packages/shared/src/components/ShareBar.spec.tsx
@@ -63,6 +63,19 @@ const renderComponent = (loggedIn = true, hasSquads = true): RenderResult => {
 };
 
 describe('ShareBar Test Suite:', () => {
+  const mockWindowOpen = jest.fn();
+  let origWindowOpen = null;
+
+  beforeEach(() => {
+    origWindowOpen = window.open;
+    window.open = mockWindowOpen;
+  });
+
+  afterEach(() => {
+    mockWindowOpen.mockClear();
+    window.open = origWindowOpen;
+  });
+
   it('should render the component for anonymous users', async () => {
     renderComponent(false, false);
     expect(
@@ -99,14 +112,18 @@ describe('ShareBar Test Suite:', () => {
     });
   });
 
-  it('should render the Facebook button and have the correct link', async () => {
+  it('should render the Facebook button and navigate to correct link', async () => {
     renderComponent();
-    const link = await screen.findByTestId(`social-share-Facebook`);
+    const btn = await screen.findByTestId(`social-share-Facebook`);
 
-    expect(link).toHaveAttribute(
-      'href',
-      getFacebookShareLink(defaultPost.commentsPermalink),
-    );
+    btn.click();
+
+    await waitFor(() => {
+      expect(mockWindowOpen).toHaveBeenCalledWith(
+        getFacebookShareLink(defaultPost.commentsPermalink),
+        '_blank',
+      );
+    });
   });
 
   it('should render the copy link button and copy link to clipboard', async () => {
--- a/packages/shared/src/components/ShareBar.tsx
+++ b/packages/shared/src/components/ShareBar.tsx
@@ -13,7 +13,8 @@ import { Squad } from '../graphql/sources';
 import { SocialShareButton } from './widgets/SocialShareButton';
 import { SquadsToShare } from './squads/SquadsToShare';
 import { ButtonSize, ButtonVariant } from './buttons/common';
-import { ReferralCampaignKey, useGetShortUrl } from '../hooks';
+import { useGetShortUrl } from '../hooks';
+import { ReferralCampaignKey } from '../lib/referral';
 
 interface ShareBarProps {
   post: Post;
--- a/packages/shared/src/components/ShareBar.tsx
+++ b/packages/shared/src/components/ShareBar.tsx
@@ -3,7 +3,7 @@ import { CopyIcon, WhatsappIcon, TwitterIcon, FacebookIcon } from './icons';
 import { Post } from '../graphql/posts';
 import { useCopyPostLink } from '../hooks/useCopyPostLink';
 import {
-  addLinkShareTrackingParams,
+  addLinkShareTrackingQuery,
   getShareLink,
   ShareProvider,
 } from '../lib/share';
@@ -29,7 +29,9 @@ export default function ShareBar({ post }: ShareBarProps): ReactElement {
   const { user, isAuthReady } = useAuthContext();
   const cid = 'share_post';
   const link =
-    isAuthReady && user ? addLinkShareTrackingParams(href, user.id, cid) : href;
+    isAuthReady && user
+      ? addLinkShareTrackingQuery({ link: href, userId: user.id, cid })
+      : href;
   const { getShortUrl } = useGetShortUrl();
   const [copying, copyLink] = useCopyPostLink(link);
   const { trackEvent } = useContext(AnalyticsContext);
@@ -46,7 +48,11 @@ export default function ShareBar({ post }: ShareBarProps): ReactElement {
     trackClick(provider);
 
     const shortLink = await getShortUrl(link);
-    const shareLink = getShareLink(provider, shortLink, post?.title);
+    const shareLink = getShareLink({
+      provider,
+      link: shortLink,
+      text: post?.title,
+    });
     window.open(shareLink, '_blank');
   };
 
--- a/packages/shared/src/components/ShareBar.tsx
+++ b/packages/shared/src/components/ShareBar.tsx
@@ -3,9 +3,8 @@ import { CopyIcon, WhatsappIcon, TwitterIcon, FacebookIcon } from './icons';
 import { Post } from '../graphql/posts';
 import { useCopyPostLink } from '../hooks/useCopyPostLink';
 import {
-  getFacebookShareLink,
-  getTwitterShareLink,
-  getWhatsappShareLink,
+  addLinkShareTrackingParams,
+  getShareLink,
   ShareProvider,
 } from '../lib/share';
 import AnalyticsContext from '../contexts/AnalyticsContext';
@@ -18,6 +17,8 @@ import { Squad } from '../graphql/sources';
 import { SocialShareButton } from './widgets/SocialShareButton';
 import { SquadsToShare } from './squads/SquadsToShare';
 import { ButtonSize, ButtonVariant } from './buttons/common';
+import { useAuthContext } from '../contexts/AuthContext';
+import { useGetShortUrl } from '../hooks';
 
 interface ShareBarProps {
   post: Post;
@@ -25,20 +26,34 @@ interface ShareBarProps {
 
 export default function ShareBar({ post }: ShareBarProps): ReactElement {
   const href = post.commentsPermalink;
-  const [copying, copyLink] = useCopyPostLink(href);
+  const { user, isAuthReady } = useAuthContext();
+  const cid = 'share_post';
+  const link =
+    isAuthReady && user ? addLinkShareTrackingParams(href, user.id, cid) : href;
+  const { getShortUrl } = useGetShortUrl();
+  const [copying, copyLink] = useCopyPostLink(link);
   const { trackEvent } = useContext(AnalyticsContext);
   const { openModal } = useLazyModal();
 
-  const onClick = (provider: ShareProvider) =>
+  const trackClick = (provider: ShareProvider) =>
     trackEvent(
       postAnalyticsEvent('share post', post, {
         extra: { provider, origin: Origin.ShareBar },
       }),
     );
 
-  const trackAndCopyLink = () => {
-    copyLink();
-    onClick(ShareProvider.CopyLink);
+  const onClick = async (provider: ShareProvider) => {
+    trackClick(provider);
+
+    const shortLink = await getShortUrl(link);
+    const shareLink = getShareLink(provider, shortLink, post?.title);
+    window.open(shareLink, '_blank');
+  };
+
+  const trackAndCopyLink = async () => {
+    const shortLink = await getShortUrl(link);
+    copyLink({ link: shortLink });
+    trackClick(ShareProvider.CopyLink);
   };
 
   const onShareToSquad = (squad: Squad) => {
@@ -76,7 +91,6 @@ export default function ShareBar({ post }: ShareBarProps): ReactElement {
         <SocialShareButton
           size={ButtonSize.Medium}
           variant={ButtonVariant.Tertiary}
-          href={getWhatsappShareLink(href)}
           icon={<WhatsappIcon secondary />}
           className="text-white"
           onClick={() => onClick(ShareProvider.WhatsApp)}
@@ -85,7 +99,6 @@ export default function ShareBar({ post }: ShareBarProps): ReactElement {
         <SocialShareButton
           size={ButtonSize.Medium}
           variant={ButtonVariant.Tertiary}
-          href={getFacebookShareLink(href)}
           icon={<FacebookIcon secondary />}
           onClick={() => onClick(ShareProvider.Facebook)}
           label="Facebook"
@@ -93,7 +106,6 @@ export default function ShareBar({ post }: ShareBarProps): ReactElement {
         <SocialShareButton
           size={ButtonSize.Medium}
           variant={ButtonVariant.Tertiary}
-          href={getTwitterShareLink(href, post?.title)}
           icon={<TwitterIcon />}
           onClick={() => onClick(ShareProvider.Twitter)}
           label="X"
--- a/packages/shared/src/components/ShareBar.tsx
+++ b/packages/shared/src/components/ShareBar.tsx
@@ -2,11 +2,7 @@ import React, { ReactElement, useContext } from 'react';
 import { CopyIcon, WhatsappIcon, TwitterIcon, FacebookIcon } from './icons';
 import { Post } from '../graphql/posts';
 import { useCopyPostLink } from '../hooks/useCopyPostLink';
-import {
-  addTrackingQueryParams,
-  getShareLink,
-  ShareProvider,
-} from '../lib/share';
+import { getShareLink, ShareCID, ShareProvider } from '../lib/share';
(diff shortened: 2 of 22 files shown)


packages/shared/src/lib/share.ts
```diff
@@ -30,3 +30,47 @@ export const getTelegramShareLink = (link: string, text: string): string =>
   `https://t.me/share/url?url=${encodeURIComponent(link)}&text=${text}`;
 export const getEmailShareLink = (link: string, subject: string): string =>
   `mailto:?subject=${subject}&body=${encodeURIComponent(link)}`;
+
+export const getShareLink = (
+  provider: ShareProvider,
+  link: string,
+  text?: string,
```
capJavert, line 37: We usually use object arguments, easier to add new ones. Same below for `addLinkShareTrackingParams`

packages/shared/src/lib/share.ts
```diff
@@ -30,3 +30,47 @@ export const getTelegramShareLink = (link: string, text: string): string =>
...
+    case ShareProvider.LinkedIn:
+      return getLinkedInShareLink(link);
+    case ShareProvider.Telegram:
+      return getTelegramShareLink(link, text ?? '');
+    case ShareProvider.Email:
+      return getEmailShareLink(link, text ?? '');
+    default:
+      return link;
+  }
+};
```
capJavert, line 57: Non blocking, but this looks like a great example of where unit tests would help just so we don't screw up the logic at any point now that we have it per provider.

packages/shared/src/lib/share.ts
```diff
@@ -30,3 +30,47 @@ export const getTelegramShareLink = (link: string, text: string): string =>
...
+    return null;
+  }
+
+  if (!userId) {
+    return link;
+  }
+
+  const url = new URL(link);
+  url.searchParams.set('userId', userId);
+  url.searchParams.set('cid', cid);
```
capJavert, line 74: I know `cid` is not optional but it does not block someone to call it or being missed by some miss in types, so probably better to check `if (userid) searchParams.set`, or something similar.

packages/shared/src/hooks/utils/useGetShortUrl.ts
```diff
@@ -0,0 +1,49 @@
...
+  if (!isAuthReady || !user) {
+    return {
+      getShortUrl: async (url: string) => {
+        return url;
+      },
+    };
+  }
+
+  return {
+    getShortUrl: async (url: string, cid?: string) => {
```
capJavert, line 25: add `useCallback` here

packages/shared/src/hooks/useSquadInvitation.ts
```diff
@@ -22,8 +24,13 @@ export const useSquadInvitation = ({
 }: UseSquadInvitationProps): UseSquadInvitation => {
   const { trackEvent } = useAnalyticsContext();
   const { completeAction } = useActions();
+  const { user } = useAuthContext();
 
-  const invitation = squad.referralUrl;
+  const invitation = addLinkShareTrackingParams(
+    squad.referralUrl,
```
capJavert, line 30: In production `squad.referralUrl` is already shortened on API side and short url returned to frontend so we either have to remove that on API side or add the tracking there already.

packages/shared/src/components/profile/Header.tsx
```diff
@@ -24,7 +25,7 @@ export function Header({
 }: HeaderProps): ReactElement {
   const [, onShareOrCopyLink] = useShareOrCopyLink({
     text: `Check out ${user.name}'s profile on daily.dev`,
-    link: user.permalink,
+    link: addLinkShareTrackingParams(user.permalink, user.id, 'share_profile'),
```
capJavert, line 28: We don't shorten here?

capJavert (changes requested):
Overall I like it, some minor concerns because quite a lot of moving parts ⚒️

packages/shared/src/hooks/useSquadInvitation.ts
kopancek: good point, this is an oversight on my side. We have the same situation with invite links, which are already passed in as minified.

packages/shared/src/lib/share.ts
```diff
@@ -30,3 +30,62 @@ export const getTelegramShareLink = (link: string, text: string): string =>
...
+}: AddLinkShareTrackingQueryParams): string => {
+  // return link as is if not provided
+  if (!link) {
+    return link;
+  }
+
+  // return link as is if params are not provided
+  if (!userId || !cid) {
+    return link;
+  }
```
nensidosari, line 84: can't we merge them into the same if condition?

packages/shared/src/lib/share.ts
```diff
@@ -30,3 +30,62 @@ export const getTelegramShareLink = (link: string, text: string): string =>
...
+
+  const url = new URL(link);
+  url.searchParams.set('userId', userId);
+  url.searchParams.set('cid', cid);
```
nensidosari, line 88: Should we check before setting if any of them is missing?

packages/webapp/pages/account/invite.tsx
```diff
@@ -103,6 +104,7 @@ const AccountInvitePage = (): ReactElement => {
           description={labels.referral.generic.inviteText}
           onNativeShare={onShareOrCopyLink}
           onClickSocial={onTrackShare}
+          shortenUrl={false}
```
nensidosari, line 107: Why false?

packages/shared/src/components/widgets/SocialShareList.tsx
```diff
@@ -42,7 +35,25 @@ export function SocialShareList({
   onCopy,
   onNativeShare,
   onClickSocial,
```
nensidosari, line 37: So we don't use this anymore? can we remove?

nensidosari (commented):
Looks good, have just few comments, mostly to understand for myself as well

packages/shared/src/components/ShareBar.tsx
```diff
@@ -18,27 +17,49 @@ import { Squad } from '../graphql/sources';
...
+    isAuthReady && user
+      ? addLinkShareTrackingQuery({ link: href, userId: user.id, cid })
+      : href;
+  const { getShortUrl } = useGetShortUrl();
+  const [copying, copyLink] = useCopyPostLink(link);
   const { trackEvent } = useContext(AnalyticsContext);
   const { openModal } = useLazyModal();
 
-  const onClick = (provider: ShareProvider) =>
+  const trackClick = (provider: ShareProvider) =>
```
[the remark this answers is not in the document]
kopancek: Well, `trackClick` was already an improvement over `onClick` from my point of view :) But sure, let's make it even better 👍

packages/shared/src/components/ShareBar.tsx
```diff
@@ -76,7 +97,6 @@ export default function ShareBar({ post }: ShareBarProps): ReactElement {
         <SocialShareButton
           size={ButtonSize.Medium}
           variant={ButtonVariant.Tertiary}
-          href={getWhatsappShareLink(href)}
```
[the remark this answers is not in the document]
kopancek: `href` param is not used anymore. We instead render the button as a button and do `window.location.open` in the `onClick` handler. The `getWhatsappShareLink` handler is called inside the `getShareLink` function based on the supplied `provider`.

packages/shared/src/lib/share.ts
kopancek: We already check above. If `userId` or `cid` is not defined, we return the URL as is.

packages/webapp/pages/account/invite.tsx
kopancek: Because the invite URL is already shortened

packages/shared/src/components/widgets/SocialShareList.tsx
kopancek: We use it on line 43 now.

packages/shared/src/components/ShareBar.spec.tsx
```diff
@@ -99,14 +112,18 @@ describe('ShareBar Test Suite:', () => {
...
+  it('should render the Facebook button and navigate to correct link', async () => {
     renderComponent();
-    const link = await screen.findByTestId(`social-share-Facebook`);
+    const btn = await screen.findByTestId(`social-share-Facebook`);
 
-    expect(link).toHaveAttribute(
-      'href',
-      getFacebookShareLink(defaultPost.commentsPermalink),
-    );
+    btn.click();
```
rebelchris, line 119: Non blocking but for future reference, prefer these events they wrap in act and just less prone to flakiness
```suggestion
    fireEvent.click(btn);
```

packages/shared/src/components/ShareOptionsMenu.tsx
```diff
@@ -41,18 +41,20 @@ export default function ShareOptionsMenu({
...
     trackEvent(
       postAnalyticsEvent('share post', post, {
         extra: { provider, origin: Origin.ShareBar },
       }),
     );
 
-  const trackAndCopyLink = () => {
-    copyLink();
+  const trackAndCopyLink = async () => {
+    const shortLink = await getShortUrl(link, 'share_post');
```
rebelchris, line 56: This is the CID right `share_post` can we maybe store is in some const?
As i've seen it used twice already could use with abstraction layer so it's easier to maintain.

packages/shared/src/components/modals/ShareModal.spec.tsx
```diff
@@ -98,26 +110,36 @@ describe('ShareModal Test Suite:', () => {
...
+  it('should render the Facebook button and navigate to the correct link', async () => {
     renderComponent();
-    const link = await screen.findByTestId(`social-share-Facebook`);
+    const btn = await screen.findByTestId(`social-share-Facebook`);
 
-    expect(link).toHaveAttribute(
-      'href',
-      getFacebookShareLink(defaultPost.commentsPermalink),
-    );
+    btn.click();
```
rebelchris, line 117:
```suggestion
    fireEvent.click(btn);
```

packages/shared/src/components/modals/ShareModal.spec.tsx
```diff
@@ -98,26 +110,36 @@ describe('ShareModal Test Suite:', () => {
...
-    const link = await screen.findByTestId(`social-share-Facebook`);
+    const btn = await screen.findByTestId(`social-share-Facebook`);
 
-    expect(link).toHaveAttribute(
-      'href',
-      getFacebookShareLink(
-        `${defaultPost.commentsPermalink}${getCommentHash(defaultComment.id)}`,
-      ),
-    );
+    btn.click();
```
rebelchris, line 131:
```suggestion
    fireEvent.click(btn);
```

packages/shared/src/components/profile/Header.tsx
```diff
@@ -24,7 +25,11 @@ export function Header({
 }: HeaderProps): ReactElement {
   const [, onShareOrCopyLink] = useShareOrCopyLink({
     text: `Check out ${user.name}'s profile on daily.dev`,
-    link: user.permalink,
+    link: addTrackingQueryParams({
+      link: user.permalink,
+      userId: user.id,
+      cid: 'share_profile',
```
rebelchris, line 31: This one I guess we could also abstract.
Maybe something like we do for AuthEvents?

packages/shared/src/components/widgets/SocialShare.tsx
```diff
@@ -32,13 +34,17 @@ export const SocialShare = ({
...
   const isComment = !!comment;
+  const { user, isAuthReady } = useAuthContext();
   const href = isComment
     ? `${post?.commentsPermalink}${getCommentHash(comment.id)}`
     : post?.commentsPermalink;
-  const [copying, copyLink] = useCopyLink(() => href);
-  const link = isComment
-    ? `${post?.commentsPermalink}${getCommentHash(comment.id)}`
-    : post?.commentsPermalink;
+  const cid = isComment ? 'share_comment' : 'share_post';
```
rebelchris, line 41: Here as well, the CID abstraction layer

packages/shared/src/components/widgets/SocialShareList.tsx
```diff
@@ -54,59 +65,52 @@ export function SocialShareList({
         />
       )}
       <SocialShareButton
-        href={getTwitterShareLink(link, description)}
         icon={<TwitterIcon />}
         variant={ButtonVariant.Primary}
         color={ButtonColor.Twitter}
-        onClick={() => onClickSocial(ShareProvider.Twitter)}
+        onClick={() => onClick(ShareProvider.Twitter)}
```
rebelchris, line 71: The naming is giving me headaches, but non blocking 😅

packages/shared/src/hooks/useShareComment.ts
```diff
@@ -38,10 +40,12 @@ export function useShareComment(
       openShareComment: async (comment, post) => {
         if ('share' in navigator) {
           try {
+            const shortUrl = await getShortUrl(
+              `${post.commentsPermalink}${getCommentHash(comment.id)}`,
+              'share_comment',
```
rebelchris, line 45: CID i guess, alternative comment, maybe we should make this object based? (just easier for everyone to understand what you are passing here)

packages/shared/src/hooks/useSharePost.ts
```diff
@@ -51,7 +53,11 @@ export function useSharePost(origin: Origin): {
             extra: { provider: ShareProvider.CopyLink, origin },
           }),
         );
-        copyLink({ link: post.commentsPermalink });
+        const shortLink = await getShortUrl(
+          post.commentsPermalink,
+          'share_post',
```
rebelchris, line 58: CID

packages/shared/src/hooks/utils/useGetShortUrl.ts
```diff
@@ -0,0 +1,45 @@
+import { useQueryClient } from '@tanstack/react-query';
+import request from 'graphql-request';
+import { useCallback } from 'react';
+import { graphqlUrl } from '../../lib/config';
+import { useAuthContext } from '../../contexts/AuthContext';
+import { GET_SHORT_URL_QUERY } from '../../graphql/urlShortener';
+import { addTrackingQueryParams } from '../../lib/share';
+
+interface UseGetShortUrlResult {
+  getShortUrl: (url: string, cid?: string) => Promise<string>;
```
rebelchris, line 10: I would make the props object based here.
Just easier for other devs to know what you pass/add/change order etc.

packages/shared/src/hooks/utils/useGetShortUrl.ts
```diff
@@ -0,0 +1,45 @@
...
+  const getShortUrl = useCallback(
+    async (url: string, cid?: string) => {
+      if (!url || !isAuthReady || !user) {
+        return url;
+      }
+
+      const trackingUrl = cid
+        ? addTrackingQueryParams({ link: url, userId: user?.id, cid })
+        : url;
+      const shortUrlKey = ['short_url', trackingUrl];
```
rebelchris, line 26: For query keys you should use `generateQueryKey` it's a stronger typed object and does handling for you.

packages/shared/src/hooks/utils/useGetShortUrl.ts
```diff
@@ -0,0 +1,45 @@
...
+  const queryClient = useQueryClient();
+
+  const getShortUrl = useCallback(
+    async (url: string, cid?: string) => {
+      if (!url || !isAuthReady || !user) {
+        return url;
+      }
+
+      const trackingUrl = cid
+        ? addTrackingQueryParams({ link: url, userId: user?.id, cid })
```
rebelchris, line 24: More of a question I see this function used here as well as on all components, wasn't there a better abstraction we could do?
Non blocking, and didn't really follow all steps right, so might be invalid.

rebelchris (changes requested):
Nice work, some minor comments and questions, nothing super blocking.
Just would appreciate a enum for CID at least 🙏

packages/shared/src/components/ShareOptionsMenu.tsx
kopancek: Yup, good idea. We will still send it as a string over to the API though, so that we can easily extend in the future in one place only.

packages/shared/src/components/widgets/SocialShareList.tsx
kopancek: I don't want to intentionally hurt anyone, so I'll consider renaming 😄

packages/shared/src/hooks/utils/useGetShortUrl.ts
kopancek: I think if it's 2 parameters that we pass, one being optional, is it really that hard to figure out? Especially if you look at it in an IDE, you pretty much see immediately which parameter means what.

packages/shared/src/hooks/utils/useGetShortUrl.ts
kopancek: I kept 2 separate concerns:
1. Adding the query params to the URL
2. Sending the URL to a URL shortener

It's because sometimes we need to use the unminified URL for other purposes or when we can easily cache it. So the `addTrackingQueryParams` is definitely not used on all of the components. 

The `getShortUrl` function accepts 2 parameters, if you supply the CID as a second param, it will do both of these steps. I think I can rewrite more of the components to use this notation.

packages/shared/src/components/widgets/SocialShareList.tsx
kopancek: I renamed to `openShareLink`, let me know if that's still awful to you 🙏

packages/shared/src/hooks/utils/useGetShortUrl.ts
kopancek: I kept one place where we sitll need it because of the quirkiness of how we implemented the `SocialShare` component, that wraps the `SocialShareList`. Unless we want to refactor that as well 🤔

packages/shared/src/components/ShareOptionsMenu.tsx
rebelchris: Yeah 100% 🙏

packages/shared/src/hooks/utils/useGetShortUrl.ts
rebelchris: Not with 2, but next week Lee will have to add a third one, then a fourth you know how these things go.
Especially if all become optional it's a pain to pass only the ones you want.

kopancek: Well, let's cross that bridge once we get there :) 

I think we can stick to the rule of - if it's 3 or more parameters, convert to object params, otherwise, feel free to use normal function params. 

Most of the projects I've worked in the past had similar logic applied to the linter, maybe we can do the same?

rebelchris: I prefer to just always go object and add linting for that.
Don't see any negatives on that approach 🙏

It's non blocking for now, but will raise it for web guild.

rebelchris (approved):
Nothing blocking from my side :)

packages/shared/src/lib/share.ts
```diff
@@ -30,3 +30,62 @@ export const getTelegramShareLink = (link: string, text: string): string =>
...
+    default:
+      return link;
+  }
+};
+
+export enum ShareCID {
+  Post = 'share_post',
+  Comment = 'share_comment',
+  Profile = 'share_profile',
+}
```
capJavert, line 69: We already have `ReferralCampaignKey` which was mapped to `cid` for referral so I think we can reuse it.

kopancek: I switched to that, can you take a look again? 🙏

capJavert (approved):
🚀

PR #2760, opened 2024-03-08
feat: new list drawer component

sshanzel:
## Changes
- Drawer component for lists. Mostly utilized on mobile only (for now).
- Feel free to check the storybook component.
  - For some reason, their array control is not working, had to make it comma-separated.

Preview:

https://github.com/dailydotdev/apps/assets/13744167/16e11181-e8d2-439c-830d-1e35ef7efd74

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

MI-125 #done

### DIFF
--- a/packages/shared/src/components/drawers/ListDrawer.tsx
+++ b/packages/shared/src/components/drawers/ListDrawer.tsx
@@ -3,18 +3,13 @@ import classNames from 'classnames';
 import { Drawer, DrawerRef, DrawerWrapperProps } from './Drawer';
 import { VIcon } from '../icons';
 import { IconSize } from '../Icon';
-
-export interface SelectProps {
-  value: string;
-  index: number;
-  event: React.MouseEvent<HTMLButtonElement>;
-}
+import { SelectParams } from './common';
 
 interface ListDrawerProps {
   drawerProps: Omit<DrawerWrapperProps, 'children'>;
   options: string[];
   selected: number; // index
-  onSelectedChange(props: SelectProps): void;
+  onSelectedChange(props: SelectParams): void;
 }
 
 export function ListDrawer({
--- a/packages/shared/src/components/drawers/ListDrawer.tsx
+++ b/packages/shared/src/components/drawers/ListDrawer.tsx
@@ -21,13 +21,14 @@ export function ListDrawer({
   const ref = React.useRef<DrawerRef>();
 
   return (
-    <Drawer {...drawerProps} ref={ref}>
+    <Drawer {...drawerProps} ref={ref} role="menu">
       {options.map((value, index) => {
         const isSelected = index === selected;
 
         return (
           <button
             key={value}
+            role="menuitem"
             type="button"
             className={classNames(
               'flex h-10 flex-row items-center overflow-hidden text-ellipsis whitespace-nowrap px-2 typo-callout',
--- a/packages/shared/src/components/drawers/ListDrawer.tsx
+++ b/packages/shared/src/components/drawers/ListDrawer.tsx
@@ -1,0 +1,55 @@
+import React, { ReactElement } from 'react';
+import classNames from 'classnames';
+import { Drawer, DrawerRef, DrawerWrapperProps } from './Drawer';
+import { VIcon } from '../icons';
+import { IconSize } from '../Icon';
+
+export interface SelectProps {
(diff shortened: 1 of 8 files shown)


packages/shared/src/components/drawers/ListDrawer.tsx
```diff
@@ -0,0 +1,51 @@
...
+  options,
+}: ListDrawerProps): ReactElement {
+  const ref = React.useRef<DrawerRef>();
+
+  return (
+    <Drawer {...drawerProps} ref={ref} role="menu">
+      {options.map((value, index) => {
+        const isSelected = index === selected;
+
+        return (
```
rebelchris, line 28: Isn't this basically the same as ContextMenuDrawer?
Or atleast many common parts right?

Maybe we can fully abstract, but atleast abstract to common classed component?

packages/shared/src/components/fields/Dropdown.tsx
```diff
@@ -8,13 +8,15 @@ import React, {
...
-  ItemParams,
   Menu,
   TriggerEvent,
   useContextMenu,
 } from '@dailydotdev/react-contexify';
 import { ArrowIcon, VIcon } from '../icons';
 import styles from './Dropdown.module.css';
+import { useViewSize, ViewSize } from '../../hooks';
+import { ListDrawer } from '../drawers/ListDrawer';
+import { SelectParams } from '../drawers/common';
```
rebelchris, line 19: Seeing this is only type remember we can use `import type { bla }` (not urgent, just incase you wanted to)

rebelchris (approved):
Generic question:
End result they both look very similar right?

What makes the biggest difference?

packages/shared/src/components/fields/Dropdown.tsx
sshanzel: Ah, yes, of course. I've been trying to use the most of it lately. Let me apply it here.

packages/shared/src/components/drawers/ListDrawer.tsx
sshanzel: For the ContextMenuDrawer, I think the primary difference is the abstracted check icon for selected items.

rebelchris: Hmm, wondering if it's worth combining them?
It will just make our lives easier down the line maybe.. (but open to both to be honest)

sshanzel:
> End result they both look very similar right?
> 
> What makes the biggest difference?

Yes, there's quite no difference, but you might notice at first when I scrolled down, you will see the item being Checked, in that case, at the case if I had scrolled down, you will see it on the selected item. Were you expecting to see the first item at the top?

packages/shared/src/components/drawers/ListDrawer.tsx
sshanzel: Yeah, the context menu components feel like a bit all over the place since we implemented a generic one but then never checked old implementations to be migrated.

rebelchris:
@sshanzel The last question is more for product/design I think.
> Were you expecting to see the first item at the top? Or at least scrolled down to it?

As for my concern was more around having 2 almost identical componetns, but got you let's leave it for now.

packages/storybook/stories/drawers/ListDrawer.stories.tsx
```diff
@@ -21,23 +19,15 @@ const meta: Meta<typeof ListDrawer> = {
...
-    options: {
-      description: 'Comma separated list of options (only in storybook)',
-      control: { type: 'text', separator: ',' },
-      defaultValue: {
-        summary: 'Option 1, Option 2, Option 3',
-      },
-    },
-  },
+  args: {
+    options: ['Option 1', 'Option 2', 'Option 3'],
```
sshanzel, line 23: @rebelchris no way it is this simple 😭

sshanzel:
> As for my concern was more around having 2 almost identical componetns, but got you let's leave it for now.

Yeah, it was the intention since the requirements of the two components are a bit different when it comes to the options. This is so to make it easier to use either of them and not have to add some sort of condition to identify if this is for a dropdown or not.

PR #2761, opened 2024-03-08
refactor: post options menu to utilize drawer on mobile

sshanzel:
## Changes
- Refactor the post options menu to utilize the drawer component on mobile.

Preview:

https://github.com/dailydotdev/apps/assets/13744167/280070e8-9661-4401-b682-569076901b93

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

MI-127 #done

### DIFF
--- a/packages/shared/src/components/PostOptionsMenu.tsx
+++ b/packages/shared/src/components/PostOptionsMenu.tsx
@@ -28,7 +28,13 @@ import useTagAndSource from '../hooks/useTagAndSource';
 import AnalyticsContext from '../contexts/AnalyticsContext';
 import { postAnalyticsEvent } from '../lib/feed';
 import { MenuIcon } from './MenuIcon';
-import { ToastSubject, useFeedLayout, useToastNotification } from '../hooks';
+import {
+  ToastSubject,
+  useFeedLayout,
+  useToastNotification,
+  useViewSize,
+  ViewSize,
+} from '../hooks';
 import {
   AllFeedPages,
   generateQueryKey,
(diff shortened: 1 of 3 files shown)


packages/shared/src/components/PostOptionsMenu.tsx
```diff
@@ -426,23 +434,34 @@ export default function PostOptionsMenu({
     });
   }
 
+  const isMobile = useViewSize(ViewSize.MobileL);
```
rebelchris, line 437: I feel we are going to need this logic quite often right?
Should we simply wrap it into it's own abstraction?

Maybe even PortalMenu? (but not sure if other occasions use it)
Otherwise it could be `ContextMenu` maybe? 
inside it will evalaute whether to show drawer or portal

WDYT?

rebelchris (commented):
Just one discussion point

packages/shared/src/components/PostOptionsMenu.tsx
sshanzel: Yeah, I was hoping to do that too, but it is a bit more complicated with the PortalMenu that we have than the dropdown. Though we have a separate ticket for that, to migrate all context menus. We can maybe do that there. Wdyt?

sshanzel: We just wanted to push this first as it is a bit higher priority than the others.

rebelchris: Ok, maybe note this comment down on that ticket so we have reference.
Happy to not block, but eventually would love one generic component for this 🙏

rebelchris (approved):

packages/shared/src/components/PostOptionsMenu.tsx
sshanzel: Yeah, just our context menus have different implementations, old ones not using the generic component, so I didn't want to spend the time for now. I will add a comment here.

sshanzel: Done 🥳  https://dailydotdev.atlassian.net/browse/MI-128?focusedCommentId=11464

PR #2767, opened 2024-03-08
fix: add login trigger to subscribe button

rebelchris:
### Describe what this PR does
- Just waiting for confirmation from analytics team [here](https://dailydotdev.slack.com/archives/C015T1Y237Z/p1709901184845719)

## Events

Did you introduce any new tracking events?
Don't forget to update the [Analytics Taxonomy sheet](https://docs.google.com/spreadsheets/d/18Lv7zXges9QfVX5VYL1a-Hyl0e1sQ3sLr0OK8YZWKXI/edit#gid=0)

Log the new/changed events below:

| Type   | event_name  | value |
|--------|-------------|-------|
| Change | open signup  | extra: { trigger: 'collection subscribe' } |

### On those affected packages:
- [ ] Have you done sanity checks in the webapp?
- [ ] Have you done sanity checks in the extension?
- [ ] Does this not break anything in companion?

### Did you test the modified components media queries?
- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

#### Did you test on actual mobile devices?
- [ ] iOS (Chrome and Safari)
- [ ] Android

AS-173 #done

PR #2769, opened 2024-03-08
fix: don't render item if you not a member

rebelchris:
- We shouldn't render this item if you're not a user

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

AS-174 #done

### DIFF
--- a/packages/shared/src/components/squads/SquadHeaderMenu.tsx
+++ b/packages/shared/src/components/squads/SquadHeaderMenu.tsx
@@ -1,4 +1,4 @@
-import React, { ReactElement, useMemo } from 'react';
+import React, { ReactElement, useContext, useMemo } from 'react';
 import dynamic from 'next/dynamic';
 import { useRouter } from 'next/router';
 import { useMutation } from '@tanstack/react-query';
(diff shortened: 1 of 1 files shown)


packages/shared/src/components/squads/SquadHeaderMenu.tsx
```diff
@@ -40,6 +41,7 @@ interface SquadHeaderMenuProps {
 export default function SquadHeaderMenu({
   squad,
 }: SquadHeaderMenuProps): ReactElement {
+  const { isLoggedIn } = useContext(AuthContext);
```
omBratteng, line 44: I think we also have `useAuthContext` as a hook 😬 
Never knew which one was the norm to use

omBratteng (approved):

packages/shared/src/components/squads/SquadHeaderMenu.tsx
omBratteng: Hehe, I see you suggested that change in #2767 😂 
image

rebelchris: Ah I don't even look at these most of the time, just whatever autocompleted.
Probably should use the named hooks it's easier.

Let me make a suggestion here 😂

omBratteng (approved):

kopancek (approved):

PR #2771, opened 2024-03-08
fix: use user instead of anonymous to generate short_url query key

kopancek:
## Changes

This is my own doing, used the `generateQueryKey` function, but did not provide the user parameter. Which resulted in query keys being generated for `"anonymous"`, even though we only generate it for logged in users. Using the user ID there actually makes more sense, especially when debugging react query.

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done


capJavert (approved):
Yep, missed it in review, it is behind `!user` check but better this way 👌

PR #2775, opened 2024-03-11
fix: active users count format

sshanzel:
## Changes
- Convert to local string to add either a comma or a period on numbers.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done

### DIFF
--- a/packages/shared/src/components/auth/ActiveUsersCounter.tsx
+++ b/packages/shared/src/components/auth/ActiveUsersCounter.tsx
@@ -11,11 +11,15 @@ export function ActiveUsersCounter(): ReactElement {
     getRandomUsersCount,
   );
 
+  if (!data) {
+    return null;
+  }
+
   return (
     <div className="relative mb-2 flex w-fit flex-row items-center rounded-8 bg-surface-float p-1 pr-3">
       <UserIcon secondary size={IconSize.Small} />
       <span className="ml-3 typo-footnote">
-        <strong>{data} </strong>
+        <strong>{data.toLocaleString()} </strong>
         devs online
       </span>
       <span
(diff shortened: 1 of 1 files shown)


packages/shared/src/components/auth/ActiveUsersCounter.tsx
```diff
@@ -11,11 +11,15 @@ export function ActiveUsersCounter(): ReactElement {
...
+  if (!data) {
+    return null;
+  }
+
   return (
     <div className="relative mb-2 flex w-fit flex-row items-center rounded-8 bg-surface-float p-1 pr-3">
       <UserIcon secondary size={IconSize.Small} />
       <span className="ml-3 typo-footnote">
-        <strong>{data} </strong>
+        <strong>{data.toLocaleString()} </strong>
```
capJavert, line 22: Keep in mind this will also do some locale [specific conversions which are more ](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toLocaleString#using_locales)then just fraction separators.
image

Maybe specifying `locales` in order would be better to avoid it?

sshanzel: Oh, very cool. Didn't know about this. Let me spend more time on how we can manipulate better.

sshanzel: Applied `en-US` as locale as part of the [requirement](https://dailydotdev.slack.com/archives/C06K5KAHM0X/p1710100529461399?thread_ts=1709785127.865909&cid=C06K5KAHM0X) to use comma. We currently don't support different languages, so using the current local wouldn't make sense.

capJavert (approved):

PR #2790, opened 2024-03-12
fix: mobile image

sshanzel:
## Changes
- The primary issue was, that no matter what kind of changes we made for the mobile side, it will never get reflected since we are still using the same component for desktop. Hence, the desktop is getting most of the changes.
- There should be no changes here for desktop.
- On mobile, we are setting the max height to be larger.

Currently deployed on preview3. Quick links:
https://preview3.app.daily.dev/sources/awscommunity
https://preview3.app.daily.dev/sources/fireship
https://preview3.app.daily.dev/squads/ai
https://preview3.app.daily.dev

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done

### DIFF
--- a/packages/shared/src/components/cards/v1/Card.tsx
+++ b/packages/shared/src/components/cards/v1/Card.tsx
@@ -44,7 +44,7 @@ export const CardContent = classed('div', 'flex flex-col mobileXL:flex-row');
 
 export const CardImage = classed(
   Image,
-  'rounded-12 max-h-[12.5rem] object-cover w-full h-auto min-h-[10rem] mobileXL:aspect-auto mobileXL:max-h-40 mobileXL:w-40 mobileXXL:max-h-56 mobileXXL:w-56',
+  'rounded-12 max-h-[12.5rem] object-cover w-full h-auto mobileXL:aspect-auto mobileXL:max-h-40 mobileXL:w-40 mobileXXL:max-h-56 mobileXXL:w-56',
 );
 
 export const CardSpace = classed('div', 'flex-1');
--- a/packages/shared/src/components/cards/v1/Card.tsx
+++ b/packages/shared/src/components/cards/v1/Card.tsx
@@ -44,7 +44,7 @@ export const CardContent = classed('div', 'flex flex-col mobileXL:flex-row');
 
 export const CardImage = classed(
   Image,
-  'rounded-12 max-h-[12.5rem] object-cover w-full h-auto mobileXL:aspect-auto mobileXL:max-h-40 mobileXL:w-40 mobileXXL:max-h-56 mobileXXL:w-56',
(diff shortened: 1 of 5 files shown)


nensidosari:
I still see issues on preview3 for mobile new feed and old feed
Screenshot 2024-03-12 at 11 56 02 AM
Screenshot 2024-03-12 at 11 55 53 AM
Screenshot 2024-03-12 at 11 54 48 AM

packages/shared/src/components/cards/v1/Card.tsx
```diff
@@ -44,7 +44,7 @@ export const CardContent = classed('div', 'flex flex-col mobileXL:flex-row');
 
 export const CardImage = classed(
   Image,
-  'rounded-12 max-h-[calc(100vw-34px)] mobileXL:aspect-auto mobileXL:max-h-40 mobileXL:w-40 mobileXXL:max-h-56 mobileXXL:w-56',
+  'rounded-12 max-h-[12.5rem] object-cover w-full h-auto mobileXL:aspect-auto mobileXL:max-h-40 mobileXL:w-40 mobileXXL:max-h-56 mobileXXL:w-56',
```
nensidosari, line 47: Just to understand, is this what was causing the main issue the PR was raised for?

packages/shared/src/components/cards/v1/SharedPostCardFooter.tsx
```diff
@@ -20,32 +20,26 @@ export const SharedPostCardFooter = ({
   return (
     <div
       className={classNames(
-        'mb-2 flex h-auto min-h-0 w-full flex-auto flex-col gap-3 rounded-12 border border-theme-divider-tertiary p-3 mobileL:flex-row',
+        'mb-2 flex h-auto min-h-0 w-full flex-auto flex-col gap-3 rounded-12 border border-theme-divider-tertiary p-3',
```
nensidosari, line 23: What issue was this causing?

packages/shared/src/components/cards/v1/Card.tsx
sshanzel: Yes, first the `w-full` ensure the image is rendered end to end. Then sets a max height which I fixed since it the calculation is based on width, which doesn't make sense.

packages/shared/src/components/cards/v1/SharedPostCardFooter.tsx
sshanzel: This is not necessary, we don't render anything greater than mobileL.

nensidosari (approved):
Please do check my comments to understand better what fixes each, just in case some other thing shows out of nowhere 😅 

From my tests everything looked good.

sshanzel:
I will merge tomorrow, so I can have more time doing one last end-to-end testing.

PR #2791, opened 2024-03-12
feat: update tabs components to work with URLs as well

DragosIorgulescu:
## Changes

- adding a `url` prop to tabs
- trigger router navigation if tabs have urls

## Events

N / A

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

MI-136 #done

### DIFF
--- a/packages/shared/src/components/tabs/TabContainer.spec.tsx
+++ b/packages/shared/src/components/tabs/TabContainer.spec.tsx
@@ -6,6 +6,8 @@ import {
   screen,
 } from '@testing-library/react';
 import nock from 'nock';
+import { NextRouter, useRouter } from 'next/router';
+import { mocked } from 'ts-jest/utils';
 import { Tab, TabContainer, TabContainerProps } from './TabContainer';
 
 beforeEach(() => {
@@ -20,6 +22,22 @@ const renderComponent = (props: TabContainerProps = {}): RenderResult => {
       <Tab label="First">Sample</Tab>
       <Tab label="Second">Test</Tab>
       <Tab label="Third">Try</Tab>
+    </TabContainer>,
+  );
+};
+
+const renderUrlComponent = (props: TabContainerProps = {}): RenderResult => {
+  return render(
+    <TabContainer {...props} onActiveChange={onActiveClick}>
+      <Tab label="First" url="/first">
+        Sample
+      </Tab>
+      <Tab label="Second" url="/second">
+        Test
+      </Tab>
+      <Tab label="Third" url="/third">
+        Try
+      </Tab>
     </TabContainer>,
   );
 };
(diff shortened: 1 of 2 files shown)


packages/shared/src/components/tabs/TabContainer.spec.tsx
```diff
@@ -54,4 +72,46 @@ describe('tab container component', () => {
...
+      expect(inactive2).not.toHaveClass('bg-theme-active');
+    });
+
+    it('should redirect to the given URL on tab click', async () => {
+      const router = useRouter();
+      renderUrlComponent({ shouldMountInactive: true });
+
+      const first = await screen.findByText('First');
+      fireEvent.click(first);
+      expect(router.push).toHaveBeenCalledWith('/first');
```
DragosIorgulescu, line 114: for some reason this assertion fails, it looks like the mock is not registering the function call? Maybe someone else has some thoughts here, in the meantime I'll try to fix it as well 🙏

packages/shared/src/components/tabs/TabContainer.tsx
```diff
@@ -58,15 +60,30 @@ export function TabContainer<T extends string = string>({
...
+
+    if (child?.props?.url) {
+      router.push(child.props.url);
+    }
   };
 
-  const isTabActive = (child: ReactElement<TabProps<T>>) =>
-    child.props.label === currentActive;
+  const isTabActive = (child: ReactElement<TabProps<T>>) => {
+    if (child.props.url) {
```
nensidosari, line 81: Maybe we can destruct the child object before this

packages/shared/src/components/tabs/TabContainer.spec.tsx
nensidosari: See example here on how to mock useRouter https://github.com/dailydotdev/apps/blob/dfa97065f51a017db5babf2f8bda03570c6f5dda/packages/shared/src/components/SidebarRankProgress.spec.tsx#L34

nensidosari (commented):
Looks good, happy to approve once test is fixed

packages/shared/src/components/tabs/TabContainer.spec.tsx
DragosIorgulescu: right, of course, thanks @capJavert 🙏

packages/shared/src/components/tabs/TabContainer.tsx
```diff
@@ -15,6 +16,7 @@ export interface TabProps<T extends string> {
   className?: string;
   style?: CSSProperties;
   showHeader?: boolean;
+  url?: string;
```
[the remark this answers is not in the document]
DragosIorgulescu: in its current form, it will only work with pathname, since we are using `router.push`, it won't work with an absolute path. would you prefer this prop be named `pathname` instead?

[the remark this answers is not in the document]
capJavert: `router.push` can work with absolute urls as well, I was just curious of the usage and if we want it, up to you on the naming 👍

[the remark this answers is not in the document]
DragosIorgulescu: I guess I need to keep up to date with things more, last time I checked, using `router.push` with a URL that is outside the app domain was not an option 😬 let's leave it as `url` for now then if no one else has a strong preference

nensidosari (approved):
Looks good to me

PR #2797, opened 2024-03-12
feat: add navigation drawer

kopancek:
## Changes

The `NavDrawer` is needed for implementing [Profile settings on mobile](https://www.figma.com/file/nmfWPS7x3kzLUvYMkBx2kW/daily.dev---Dev-Mode?node-id=999%3A2766&mode=dev) and probably some other parts I'm not aware off yet. Splitting it into it's own PR to make it easier to reuse in case others need it.

- Added `NavDrawer` component that reuses `Drawer` and adds some styling on top of it
- Added `NavDrawer` storybook

The component is not used anywhere, just in storybook for now.

### Screenshot

Screenshot 2024-03-12 at 17 51 26

- [X] MobileL (420px)
- [X] Tablet (656px)
- [X] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

MI-131 #done

### DIFF
--- a/packages/shared/src/components/drawers/Drawer.tsx
+++ b/packages/shared/src/components/drawers/Drawer.tsx
@@ -17,6 +17,8 @@ import { Button } from '../buttons/Button';
 export enum DrawerPosition {
   Bottom = 'bottom',
   Top = 'top',
+  Left = 'left',
+  Right = 'right',
 }
 
 interface ClassName {
@@ -39,11 +41,15 @@ interface DrawerProps
 const drawerPositionToClassName: Record<DrawerPosition, string> = {
   [DrawerPosition.Bottom]: 'bottom-0 rounded-t-16',
   [DrawerPosition.Top]: 'top-0 rounded-b-16',
+  [DrawerPosition.Left]: 'left-0 rounded-r-16',
+  [DrawerPosition.Right]: 'right-0 rounded-l-16',
 };
 
 const animatePositionClassName: Record<DrawerPosition, string> = {
   [DrawerPosition.Bottom]: 'translate-y-full',
   [DrawerPosition.Top]: '-translate-y-full',
+  [DrawerPosition.Left]: '-translate-x-full',
+  [DrawerPosition.Right]: 'translate-x-full',
 };
 
 function BaseDrawer({
@@ -75,11 +81,10 @@ function BaseDrawer({
       <div
         {...props}
         className={classNames(
-          'absolute flex flex-col overflow-y-auto bg-background-default transition-transform duration-300 ease-in-out',
+          'absolute flex max-h-[calc(100%-5rem)] w-full flex-col overflow-y-auto bg-background-default transition-transform duration-300 ease-in-out',
           isAnimating && animatePositionClassName[position],
           drawerPositionToClassName[position],
           !title && classes,
-          'max-h-[calc(100%-5rem)] w-full',
         )}
         ref={(node) => {
           container.current = node;
--- a/packages/shared/src/components/drawers/Drawer.tsx
+++ b/packages/shared/src/components/drawers/Drawer.tsx
@@ -32,6 +32,7 @@ interface DrawerProps
   className?: ClassName;
   position?: DrawerPosition;
   closeOnOutsideClick?: boolean;
+  isFullScreen?: boolean;
   isClosing?: boolean;
   title?: string;
   onClose(): void;
@@ -57,6 +58,7 @@ function BaseDrawer({
   className = {},
   position = DrawerPosition.Bottom,
   closeOnOutsideClick = true,
+  isFullScreen = false,
   isClosing = false,
   title,
   onClose,
@@ -73,7 +75,8 @@ function BaseDrawer({
   return (
     <div
       className={classNames(
-        'fixed inset-0 z-max bg-overlay-quaternary-onion transition-opacity duration-300 ease-in-out',
+        'fixed inset-0 z-max transition-opacity duration-300 ease-in-out',
+        !isFullScreen && 'bg-overlay-quaternary-onion',
         className?.overlay,
         isAnimating && 'opacity-0',
       )}
@@ -81,9 +84,10 @@ function BaseDrawer({
       <div
         {...props}
         className={classNames(
-          'absolute flex max-h-[calc(100%-5rem)] w-full flex-col overflow-y-auto bg-background-default transition-transform duration-300 ease-in-out',
+          'absolute flex w-full flex-col overflow-y-auto bg-background-default transition-transform duration-300 ease-in-out',
+          isFullScreen ? 'inset-0' : 'max-h-[calc(100%-5rem)]',
+          !isFullScreen && drawerPositionToClassName[position],
           isAnimating && animatePositionClassName[position],
-          drawerPositionToClassName[position],
           !title && classes,
         )}
         ref={(node) => {
--- a/packages/shared/src/components/drawers/NavDrawer.tsx
+++ b/packages/shared/src/components/drawers/NavDrawer.tsx
@@ -43,7 +43,7 @@ export function NavDrawer({
       {...otherDrawerProps}
       closeOnOutsideClick={false}
       displayCloseButton={displayCloseButton}
-      isFullScreen={true}
+      isFullScreen
       position={position ?? DrawerPosition.Left}
       ref={ref}
       role="menu"
--- a/packages/shared/src/components/drawers/NavDrawer.tsx
+++ b/packages/shared/src/components/drawers/NavDrawer.tsx
@@ -18,7 +18,7 @@ interface NavDrawerProps {
 
 const NavDrawerHeader = classed(
   'div',
-  'flex items-center gap-3 px-4 border-b border-theme-divider-tertiary h-12',
+  'flex items-center gap-3 px-4 border-b border-border-subtlest-tertiary h-12',
 );
 
 const NavDrawerContent = classed('div', 'flex flex-col px-4 py-5');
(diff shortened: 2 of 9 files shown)


packages/shared/src/components/drawers/NavDrawer.tsx
```diff
@@ -0,0 +1,72 @@
...
+  return (
+    <Drawer
+      {...otherDrawerProps}
+      closeOnOutsideClick={false}
+      displayCloseButton={displayCloseButton}
+      position={position ?? DrawerPosition.Left}
+      ref={ref}
+      role="menu"
+      className={{
+        drawer: 'bottom-0 top-0 !max-h-none !rounded-none',
```
sshanzel, line 50: For drawers coming from the sides, these classes should be coming from the Drawer component itself.

We can modify the `drawerPositionToClassName` to include these classes, along with the removal of max height and only applied for top or bottom drawer.

packages/shared/src/components/drawers/NavDrawer.tsx
```diff
@@ -0,0 +1,72 @@
...
+        overlay: 'bg-transparent',
```
sshanzel, line 51: Is it required to have a transparent bg? Usually when we open popups, etc. we add overlay so the users can focus on the shown display.

sshanzel (commented):
Looking great! Just have a few minor comments, but I really like the idea of having a navigation drawer.

packages/shared/src/components/drawers/NavDrawer.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+interface NavDrawerProps {
+  drawerProps: Omit<DrawerWrapperProps, 'children'>;
+  items: NavItemProps[];
+  header?: string;
+}
+
+const NavDrawerHeader = classed(
+  'div',
+  'flex items-center gap-3 px-4 border-b border-theme-divider-tertiary h-12',
```
rebelchris, line 21:
```suggestion
  'flex items-center gap-3 px-4 border-b border-border-subtlest-tertiary h-12',
```

packages/shared/src/components/drawers/NavDrawer.tsx
```diff
@@ -0,0 +1,72 @@
...
+    >
+      {header && (
+        <NavDrawerHeader>
+          <Button
+            variant={ButtonVariant.Tertiary}
+            size={ButtonSize.Small}
+            onClick={() => ref.current?.onClose()}
+            icon={<ArrowIcon className="-rotate-90" />}
+          />
+          <NavHeading className="font-bold typo-body">{header}</NavHeading>
```
rebelchris, line 62:
```suggestion
          <NavHeading>{header}</NavHeading>
```
The classes are already on NavHeading

packages/shared/src/components/drawers/NavDrawerItem.tsx
```diff
@@ -0,0 +1,41 @@
...
+  return (
+    <Button
+      role="menuitem"
+      type="button"
+      icon={icon}
+      className="!justify-start !font-normal"
+      onClick={onClick}
+      href={href}
+      tag={href ? 'a' : 'button'}
+      variant={ButtonVariant.Tertiary}
```
rebelchris, line 36: Should these render as button variants?
Seems like it's not actually a button, more a styled element? (doesn't match our button guidelines basically)

rebelchris: Maybe double check what we do in sidebar currently, but think it should not abuse the buttons basically looking at figma.

rebelchris (dismissed):
So cool to quickly try the storybook 👏

packages/shared/src/components/drawers/NavDrawer.tsx
kopancek: This is more of a `full-screen` drawer edge case. Maybe I should add it as a prop to the drawer itself, which will add all of the positioning classes for us.

We still might want a drawer to animate from the side, but not cover the whole screen...

packages/shared/src/components/drawers/Drawer.tsx
```diff
@@ -75,11 +81,10 @@ function BaseDrawer({
       <div
         {...props}
         className={classNames(
-          'absolute flex flex-col overflow-y-auto bg-background-default transition-transform duration-300 ease-in-out',
+          'absolute flex max-h-[calc(100%-5rem)] w-full flex-col overflow-y-auto bg-background-default transition-transform duration-300 ease-in-out',
```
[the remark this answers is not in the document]
kopancek: No, by default Drawer is not full screen. This is more a question for @sshanzel. I just rearranged the classnames here.

packages/shared/src/components/drawers/NavDrawer.tsx
kopancek: Ooo, nice, didn't see that :)

packages/shared/src/components/drawers/NavDrawer.tsx
kopancek: Refactoring leftover 👍

packages/shared/src/components/drawers/NavDrawerItem.tsx
kopancek: I did a preview of the component in Figma, and it looks exactly like a button to me.

Screenshot 2024-03-13 at 08 45 41

packages/shared/src/components/drawers/NavDrawer.tsx
sshanzel: Yeah, that is true. We may want full screen but we can still decide the origin of the animation.

packages/shared/src/components/drawers/Drawer.tsx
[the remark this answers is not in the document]
sshanzel: Yes, drawers, were not intended to be full screen.

Screenshot 2024-03-13 at 3 47 19 PM

packages/shared/src/components/drawers/NavDrawer.tsx
kopancek: Since in this case the drawer is full-screen, it does not make sense to show the overlay. This is just poor mans (me) implementation to hide the overlay without killing all the existing styling, e.g. the animation. 

On top of that, the animation looks rather ugly, if we put a blue overlay there and then in 300ms completely cover it.

kopancek: I switched to adding an `isFullScreen` prop to the `Drawer` itself. Now we only set the background if the drawer is not fullscreen, otherwise it's whatever is the default background.

packages/shared/src/components/drawers/NavDrawer.tsx
kopancek: Added an `isFullScreen` prop to the `Drawer` and refactored the styling accordingly. Now we don't need to supply these anymore.

packages/shared/src/components/drawers/NavDrawerItem.tsx
kopancek: In sidebar we use `<li` tags with an `<a href` inside, but in this case, the buttons already have most of the styling we need. Out button component also supports rendering an `<a href`, so not sure why you think this would be abuse?

packages/shared/src/components/drawers/Drawer.tsx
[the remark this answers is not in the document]
rebelchris: Hmm interesting, the figma for this settings one shows full screen AFAIK?

packages/shared/src/components/drawers/Drawer.tsx
```diff
@@ -67,19 +75,20 @@ function BaseDrawer({
   return (
     <div
       className={classNames(
-        'fixed inset-0 z-max bg-overlay-quaternary-onion transition-opacity duration-300 ease-in-out',
+        'fixed inset-0 z-max transition-opacity duration-300 ease-in-out',
```
sshanzel, line 78: We can even make this overlay wrapper be used in `ConditionalWrapper`. Though it might break the styles and unnecessarily add more dev time here. We can settle with what we have now. Your call.

packages/shared/src/components/drawers/Drawer.tsx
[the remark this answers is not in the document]
sshanzel: That's the Figma guideline we have for drawers 😅 

https://www.figma.com/file/nmfWPS7x3kzLUvYMkBx2kW/daily.dev---Dev-Mode?node-id=2384%3A21195&mode=dev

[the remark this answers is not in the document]
sshanzel: @rebelchris you mean the feed filter settings? We won't be using Drawer for that. We will stick with Modals but on full screen.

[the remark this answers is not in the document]
rebelchris: Screenshot 2024-03-13 at 10 33 55
But how sure are you that this was intended to be a drawer?
I would argue you need to check with Design as it clearly is intended for full screen no?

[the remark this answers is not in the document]
rebelchris: @sshanzel Yes, but also don't see any top space here.
I think these full height elements should exist.
Screenshot 2024-03-13 at 10 36 10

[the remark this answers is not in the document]
sshanzel: > @sshanzel Yes, but also don't see any top space here.
> I think these full height elements should exist.

Yes, then that is not the common drawer that we have, though Milan and I have talked about it, and we will instead introduce a prop to make the Drawer full screen so we won't have to introduce another component with the same animation.

packages/shared/src/components/drawers/NavDrawerItem.tsx
kopancek: Created a new button variant `Option` for now. Also created a followup task to clean this up: https://dailydotdev.atlassian.net/browse/MI-231

nensidosari (approved):
Looks good to me

PR #2798, opened 2024-03-12
fix: enable switching of search via gb

rebelchris:
- Option to experiment with different search versions

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done

### DIFF
--- a/packages/shared/src/components/MainFeedLayout.tsx
+++ b/packages/shared/src/components/MainFeedLayout.tsx
@@ -126,6 +126,7 @@ export default function MainFeedLayout({
   });
   const feedVersion = useFeature(feature.feedVersion);
   const onboardingV4dot5 = useFeature(feature.onboardingV4dot5);
+  const searchVersion = useFeature(feature.searchVersion);
   const isOnboardingV4dot5 = onboardingV4dot5 === OnboardingV4dot5.V4dot5;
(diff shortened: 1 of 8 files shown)


packages/shared/src/components/MainFeedLayout.tsx
```diff
@@ -126,6 +126,7 @@ export default function MainFeedLayout({
   });
   const feedVersion = useFeature(feature.feedVersion);
   const onboardingV4dot5 = useFeature(feature.onboardingV4dot5);
+  const searchVersion = useFeature(feature.searchVersion);
```
idoshamun, line 129: When we start experimenting with search, we would have to use it in a better place. Only once someone actually queries something. Otherwise we will enroll everyone. For now it's fine, I just want to clarify that

idoshamun (approved):
Please see my remark. It's ok for poc

packages/shared/src/components/MainFeedLayout.tsx
rebelchris: Got it, should keep that in mind.
Probably can refactor a bit anyway to only use the hook I think 🤔

But let's leave it as is for POC

PR #2805, opened 2024-03-13
feat: add logout reason to logout request

omBratteng:
## Changes

Adds an optional `reason=something` as a query parameter to the logout request, this is used in API for logout metrics

### DIFF
--- a/packages/shared/src/components/ProfileMenu.tsx
+++ b/packages/shared/src/components/ProfileMenu.tsx
@@ -84,7 +84,7 @@ export default function ProfileMenu({
         title: 'Logout',
         buttonProps: {
           icon: <PowerIcon />,
-          onClick: logout,
+          onClick: () => logout('manual logout'),
         },
       },
     ];
--- a/packages/shared/src/components/ProfileMenu.tsx
+++ b/packages/shared/src/components/ProfileMenu.tsx
@@ -22,6 +22,7 @@ import { reputation, webappUrl } from '../lib/constants';
 import { UserMetadata } from './profile/UserMetadata';
 import { HeroImage } from './profile/HeroImage';
 import { anchorDefaultRel } from '../lib/strings';
+import { LogoutReason } from '../lib/user';
 
 interface ListItem {
   title: string;
@@ -84,7 +85,7 @@ export default function ProfileMenu({
         title: 'Logout',
         buttonProps: {
           icon: <PowerIcon />,
-          onClick: () => logout('manual logout'),
+          onClick: () => logout(LogoutReason.ManualLogout),
         },
       },
     ];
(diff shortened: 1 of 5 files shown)


packages/shared/src/components/ProfileMenu.tsx
```diff
@@ -84,7 +84,7 @@ export default function ProfileMenu({
         title: 'Logout',
         buttonProps: {
           icon: <PowerIcon />,
-          onClick: logout,
+          onClick: () => logout('manual logout'),
```
rebelchris, line 87: Let's make this some enum somewhere, just to have it in case we do new logout methods or whatever

rebelchris (approved):
non blocking

capJavert (approved):
Also agreed what Chris said about enum.

packages/shared/src/contexts/AuthContext.tsx
```diff
@@ -78,8 +78,8 @@ export const getQueryParams = (): Record<string, string> => {
 
 export const REGISTRATION_PATH = '/register';
 
-const logout = async (): Promise<void> => {
-  await dispatchLogout();
+const logout = async (reason?: string): Promise<void> => {
```
kopancek, line 81: Should we make reason a required parameter? That way, the typescript type system will enforce having the parameter everywhere the logout function is used. Otherwise we might potentially miss places where we should have added it and didn't.

kopancek (approved):
Looks good, agree with Chris on enum. Added non-blocking comment.

packages/shared/src/contexts/AuthContext.tsx
omBratteng: Probably could make it required, it'll fallback on the API side anyways. There are two other places it is used currently

This one in AuthModal, not really sure if it's even in use anymore. Don't think you ever see the AuthModal anymore if you're logged in.
https://github.com/dailydotdev/apps/blob/345db939e85fcff112b29da2c67c9f62e34f7297/packages/shared/src/components/auth/AuthModal.tsx#L37-L39

And this one in AuthContext. Don't fully understand when this would be triggered, haven't looked too much into it
https://github.com/dailydotdev/apps/blob/345db939e85fcff112b29da2c67c9f62e34f7297/packages/shared/src/contexts/AuthContext.tsx#L137-L139

Maybe @rebelchris has some insight to those two.

packages/shared/src/components/ProfileMenu.tsx
omBratteng: Made it into an enum

packages/shared/src/contexts/AuthContext.tsx
rebelchris: Both are basically for users who start onboarding but never finish.
The last one specifically for when they reload and didn't finish signing up. 

I'm ok with making it required, not many implementations anyway.

omBratteng: Made it required and added a new reason `IncomleteOnboarding = 'incomplete onboarding',`

rebelchris (approved):

PR #2806, opened 2024-03-13
fix: x feedback loop

nensidosari:
Fix x on feedback loop https://dailydotdev.slack.com/archives/C01L0QXQJNB/p1709751916071319

### DIFF
--- a/packages/shared/src/components/cards/FeedbackCard.tsx
+++ b/packages/shared/src/components/cards/FeedbackCard.tsx
@@ -26,7 +26,7 @@ export const FeedbackCard = ({
   return (
     <div className="flex-1 space-y-4 p-6 pb-5">
       <div className="relative block">
-        <p className="font-bold typo-callout mr-0 tablet:mr-4">
+        <p className="mr-0 font-bold typo-callout tablet:mr-4">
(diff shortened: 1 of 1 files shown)


packages/shared/src/components/cards/FeedbackCard.tsx
```diff
@@ -25,8 +25,8 @@ export const FeedbackCard = ({
 
   return (
     <div className="flex-1 space-y-4 p-6 pb-5">
-      <div className="relative flex justify-between">
-        <p className="font-bold typo-callout">
+      <div className="relative block">
+        <p className="font-bold typo-callout mr-0 tablet:mr-4">
```
sshanzel, line 29: You mentioned that the range for the issue is at `656px to 690px` - should we be removing the margin for sizes bigger than `tablet`?

nensidosari: Figured out it happens also on some bigger px but small ranges, so since it doesn't ruin anything left from tablet and above

sshanzel (approved):
Nice. In that case, this should be good 🚀

rebelchris (approved):
Looks good to me

PR #2815, opened 2024-03-14
refactor: modals to utilize the drawer or be on full-screen

sshanzel:
## Changes
- Convert all modals on mobile to be either on full screen or drawer.
- Initially, we thought of abstracting the drawer inside the Modal component, but as the development went on. I realized that we will have to pass another object to the Modal which is `drawerProps`. By doing that, when we render the Drawer inside of the Modal, a lot of things are happening that are unrelated to the Drawer itself, and by passing the drawer props, it makes no sense to not just render the Drawer on demand inside the file rather than be "abstracted".
- There are only a few Modals that require rendering the Drawer, and some of them require customizations of the content which you won't be able to do easily if the Drawer is abstracted.

**To view all the previews, I prepared a confluence page to make it all easier**. The changes here are based on discussion with Tomer.

https://dailydotdev.atlassian.net/wiki/spaces/HAN/pages/837517352/Modals+migration+to+either+full+screen+or+drawer

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

MI-129 #done


packages/shared/src/components/drawers/Drawer.tsx
```diff
@@ -35,7 +40,7 @@ interface DrawerProps
   isFullScreen?: boolean;
   isClosing?: boolean;
   title?: string;
-  onClose(): void;
+  onClose(e: MouseEvent | KeyboardEvent | MessageEvent): void;
```
sshanzel, line 43: You might see a lot of these changes, this is due to Drawer's requirement to fulfill outside click, it uses the native event handler.

packages/shared/src/components/modals/SubmitArticleModal.tsx
```diff
@@ -202,18 +202,14 @@ export default function SubmitArticleModal({
...
         >
-          <div>
-            <p className="mb-2 text-theme-label-tertiary typo-callout">
+          <div className="flex flex-col gap-4">
+            <p className="text-theme-label-tertiary typo-callout">
               Found an interesting post? Do you want to share it with the
               community? Enter the post&apos;s URL / link below to add it to the
               feed.
             </p>
-            {!isEnabled && (
```
sshanzel, line 212: The design doesn't have this anymore.

packages/shared/src/components/modals/SubmitArticleModal.tsx
```diff
@@ -222,52 +218,50 @@ export default function SubmitArticleModal({
...
-              fieldType="tertiary"
-              name="articleUrl"
-              inputId="article_url"
-              label="Paste post url"
-              disabled={!isEnabled}
-              hint={urlHint}
-              valid={!urlHint}
-              valueChanged={onUrlChanged}
+          </div>
+          <p className="typo-callout">
```
sshanzel, line 222: The diff looks crazy again, just a wrapper removal.

packages/shared/src/components/checklist/SquadChecklistCard.tsx
```diff
@@ -5,19 +5,18 @@ import { Squad } from '../../graphql/sources';
...
+import { useViewSize, ViewSize } from '../../hooks';
+import { Drawer } from '../drawers';
 
 const SquadChecklistCard = ({ squad }: { squad: Squad }): ReactElement => {
   const { steps, isChecklistVisible, setChecklistVisible, isChecklistReady } =
     useSquadChecklist({
       squad,
     });
-  const { sidebarRendered } = useSidebarRendered();
+  const isMobile = useViewSize(ViewSize.MobileL);
```
kopancek, line 19: Instead of relying on javascript, can we render both and rely on CSS media queries and hide the element we don't want to see?

packages/shared/src/components/drawers/Drawer.tsx
```diff
@@ -116,11 +121,16 @@ function BaseDrawer({
...
+            className={classNames(
+              'sticky -bottom-3 bg-background-default',
+              className?.close,
+            )}
+          >
             <Button
               variant={ButtonVariant.Float}
               className="mt-3 w-full"
-              onClick={onClose}
+              onClick={(e) => onClose(e.nativeEvent)}
```
kopancek, line 133: Why was this change needed?

packages/shared/src/components/modals/PushNotificationModal.tsx
```diff
@@ -1,13 +1,18 @@
...
 import { Modal, ModalProps } from './common/Modal';
 import { NotificationPromptSource } from '../../lib/analytics';
 import { usePushNotificationMutation } from '../../hooks/notifications';
+import { useViewSize, ViewSize } from '../../hooks';
+import { Drawer } from '../drawers';
+import ConditionalWrapper from '../ConditionalWrapper';
 
 function PushNotificationModal(modalProps: ModalProps): ReactElement {
   const { onRequestClose } = modalProps;
+  const isMobile = useViewSize(ViewSize.MobileL);
```
kopancek, line 15: Same as above, do we need to use the hook to display the drawer instead of CSS media query directly?

packages/shared/src/components/modals/common/Modal.tsx
```diff
@@ -101,16 +101,18 @@ export function Modal({
...
     modalKindToOverlayClassName[kind],
     overlayClassName,
   );
   const modalClassName = classNames(
-    'modal relative flex max-w-full flex-col items-center rounded-16 border border-theme-divider-secondary bg-accent-pepper-subtlest antialiased shadow-2 focus:outline-none',
+    'modal relative flex max-w-full flex-col items-center border border-theme-divider-secondary bg-accent-pepper-subtlest antialiased shadow-2 focus:outline-none',
+    'h-full tablet:h-auto tablet:rounded-16',
     modalKindToClassName[kind],
-    modalSizeToClassName[size],
+    isMobile ? 'w-full' : modalSizeToClassName[size],
```
kopancek, line 115: We could do this with css media queries, no?

packages/shared/src/components/checklist/SquadChecklistCard.tsx
DragosIorgulescu: The only potential drawback to that is if we trigger any events or things like that, if both are rendered, we would be duplicating them? maybe something to consider on a case by case basis

packages/shared/src/components/modals/common/Modal.tsx
DragosIorgulescu: in this case, I'm echoing @kopancek on using media queries

DragosIorgulescu (commented):
Looking great overall! I just have one curiosity on a higher level:

Would it make sense to update the `Modal` component itself to behave as a drawer on mobile by default, and maybe have some prop to force it to a modal on mobile in the edge cases where we need it? 

It feels like in 90% of cases we have the same content, it's just wrapped in either `Modal` or `Drawer`, maybe there's an opportunity to make things a little DRY-er here. Curious about your thoughts.

packages/shared/src/components/drawers/Drawer.tsx
sshanzel: Previously, the type was not needing to pass an event.

packages/shared/src/components/modals/common/Modal.tsx
sshanzel: Right. Yeah, we can convert this.

sshanzel:
> Maybe we could refactor that into a separate hook or component?

This is what I mentioned in the description, at first what we wanted was to make this condition be part of the `Modal` component and contain a `isDrawerOnMobile` prop. But then I realized there are some needed customizations that made the implementation "ugly" on some components.

packages/shared/src/components/checklist/SquadChecklistCard.tsx
sshanzel: Yes, one thing that easily comes to mind is closing on outside click. Either one of them will think you are always clicking outside its elements, hence closing it. There's probably some workaround, but let's see what are the pros and cons on both ends.

DragosIorgulescu:
> > Maybe we could refactor that into a separate hook or component?
> 
> This is what I mentioned in the description, at first what we wanted was to make this condition be part of the `Modal` component and contain a `isDrawerOnMobile` prop. But then I realized there are some needed customizations that made the implementation "ugly" on some components.

Ah, no problem then, ignore my comment 🙏  thank you for the clarification

sshanzel:
> Ah, no problem then, ignore my comment 🙏 thank you for the clarification

All cool. I will push another commit doing that, then we can discuss what we would prefer better. Thank you! 🚀

packages/shared/src/components/checklist/SquadChecklistCard.tsx
rebelchris: Good point Milan.
The idea is that we should use CSS media queries wherever possible as the default.

There some edge cases where we need it in JS evals, then we prefer the useViewSize hook.

packages/shared/src/components/modals/PushNotificationModal.tsx
```diff
@@ -27,19 +28,37 @@ function PushNotificationModal(modalProps: ModalProps): ReactElement {
...
       kind={Modal.Kind.FlexibleTop}
       size={Modal.Size.Medium}
+      isDrawerOnMobile
+      drawerProps={{ className: { drawer: 'pb-4', close: 'mx-4' } }}
     >
       <Modal.Header />
-      <Modal.Body>
-        <Modal.Title>Enable Push Notifications</Modal.Title>
-        <Modal.Text className="text-center">
+      <img
```
sshanzel, line 35: This is one of the parts where I thought it didn't seem right. Where we will be rendering a lot of hidden elements and display based on screen size.

packages/shared/src/components/modals/common/ModalClose.tsx
```diff
@@ -44,7 +45,9 @@ function ModalCloseComponent(
   }: ModalCloseProps,
   ref: Ref<HTMLButtonElement>,
 ): ReactElement {
-  if (!onClick) {
+  const { isDrawer } = useModalContext();
```
sshanzel, line 48: Modal close button is not necessary in Drawers since it has its own close button.

packages/shared/src/components/modals/common/ModalFooter.tsx
```diff
@@ -16,8 +16,8 @@ export function ModalFooter({
   justify = Justify.End,
   view,
 }: ModalFooterProps): ReactElement {
-  const { activeView } = useContext(ModalPropsContext);
-  if (view && view !== activeView) {
+  const { activeView, isDrawer } = useContext(ModalPropsContext);
```
sshanzel, line 19: Footer is not needed on Drawers.

packages/shared/src/components/modals/common/ModalHeader.tsx
```diff
@@ -37,8 +37,14 @@ export function ModalHeader({
   title,
   showCloseButton = true,
 }: ModalHeaderProps): ReactElement {
-  const { activeView, onRequestClose, tabs } = useContext(ModalPropsContext);
+  const { activeView, onRequestClose, tabs, isDrawer } =
+    useContext(ModalPropsContext);
   const modalTitle = title ?? (tabs ? activeView : undefined);
+
+  if (isDrawer) {
```
sshanzel, line 44: Header is not needed on drawers (this component contains title and close button).

sshanzel:
@dailydotdev/web-team let me know what you guys think about having the Drawer abstracted from the Modal. As mentioned before, the key difference is that we will be rendering elements that are sometimes not necessary to be displayed on different screen sizes, but of course, those elements are hidden.

The `isModal` is now at its minimum, which is at the major components only. Modal and InteractivePopup.

packages/shared/src/components/modals/PushNotificationModal.tsx
capJavert: We can make a quick drawer wrapper component which will internally check Modal context and then render or not render depending on `isDrawer`. 

```js
const DrawerImg = (props) => {
    const modalCtx = useModalContext()

    if (modalCtx.isDrawer) {
        return null
    }

    return (
        <img {...props} />
    )
}
```

```js
<Modal>
    <DrawerImg />
</Modal>
```

packages/shared/src/components/checklist/SquadChecklistCard.tsx
```diff
@@ -49,27 +45,17 @@ const SquadChecklistCard = ({ squad }: { squad: Squad }): ReactElement => {
...
-    );
-  }
-
   return (
     <InteractivePopup
+      isDrawerOnMobile
+      drawerProps={{ className: { drawer: 'pb-4' } }}
       position={InteractivePopupPosition.RightEnd}
-      className="rounded-none !bg-transparent"
+      className="flex w-full max-w-[21.5rem] justify-center rounded-none"
```
kopancek, line 57: Why do we need a custom max width here?

capJavert (commented):
I like the direction, some questions/ideas.

packages/webapp/pages/posts/[id]/index.tsx
```diff
@@ -222,17 +225,19 @@ const PostPage = ({ id, initialData }: Props): ReactElement => {
...
-              authGradientBg,
-              'sticky left-0 top-0 z-max w-full justify-center gap-2 border-b border-theme-color-cabbage px-4 py-2',
-            ),
-            button: 'flex-1 tablet:max-w-[9rem]',
-          }}
-        />
-      ),
+      customBanner: shouldShowAuthBanner &&
+        !isLaptop &&
+        (isTablet || !shouldShowLogin) && (
```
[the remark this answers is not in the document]
Comment (sshanzel) [reply packages/webapp/pages/posts/[id]/index.tsx:230]: When the popup (shouldShowLogin) is opened, this banner should not beseen.

packages/shared/src/components/modals/common/ModalFooter.tsx
```diff
@@ -16,8 +16,8 @@ export function ModalFooter({
...
+  if ((view && view !== activeView) || isDrawer) {
```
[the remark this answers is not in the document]
sshanzel: Personally, I prefer what you are suggesting. I used to do that way back (separate null returns), but from all the feedback, it was preferred to have the conditions merged. We should maybe define some guidelines for this so it won't have to be mentioned again.

kopancek (commented):
Maybe I'm weird, but I like this approach much more and it seems to be easier to use than the previous one.

We could maybe do a better job with how we display the footer in those exceptional cases where we still show footer content in the drawer, e.g. by providing another prop to the footer itself.

packages/webapp/pages/posts/[id]/index.tsx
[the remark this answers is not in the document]
Comment (capJavert) [reply packages/webapp/pages/posts/[id]/index.tsx:230]: Aha, just tricky naming, ok.

packages/shared/src/components/drawers/Drawer.tsx
```diff
@@ -14,6 +14,15 @@ import { useOutsideClick } from '../../hooks/utils/useOutsideClick';
...
 import { Button } from '../buttons/Button';
 
+export type PopupEventType =
+  | MouseEvent
+  | KeyboardEvent
+  | MessageEvent
+  | React.MouseEvent
+  | React.KeyboardEvent;
+
+export type PopupCloseEvent = (e: PopupEventType) => void;
```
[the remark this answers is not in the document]
sshanzel: Ah, good catch. Yes, that should be named callback for func.

packages/shared/src/components/modals/RanksModal/DevCardFooter.tsx
```diff
@@ -17,7 +17,7 @@ export default function DevCardFooter({
 }: DevCardFooterProps): ReactElement {
   return (
     <div className="pt-2">
-      <Modal.Subtitle>Generate your DevCard</Modal.Subtitle>
+      <Modal.Subtitle className="mb-2">Generate your DevCard</Modal.Subtitle>
```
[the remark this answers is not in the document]
sshanzel: Yes, in the component where we declared this, we removed these margins. This is because it should be defined on the consuming end. Unless it is a padding, it should be from what was expelained.

packages/webapp/pages/squads/[handle]/index.tsx
```diff
@@ -194,7 +194,6 @@ const SquadPage = ({
           )}
         />
         <SquadPageHeader squad={squad} members={squadMembers} />
-        <SquadChecklistCard squad={squad} />
```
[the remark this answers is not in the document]
Comment (sshanzel) [reply packages/webapp/pages/squads/[handle]/index.tsx:197]: Moved it somewhere we actually manage the state to show this or not.

packages/webapp/pages/posts/[id]/index.tsx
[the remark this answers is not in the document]
Comment (sshanzel) [reply packages/webapp/pages/posts/[id]/index.tsx:230]: Same here https://github.com/dailydotdev/apps/pull/2815#discussion_r1528329639

packages/shared/src/components/checklist/SquadChecklistCard.tsx
sshanzel: We added the `w-full` - on laptop, this would unnecessarily consume the full page.

packages/shared/src/components/modals/common/Modal.tsx
```diff
@@ -88,10 +89,15 @@ export function Modal({
...
   steps,
+  isDrawerOnMobile,
+  drawerProps,
+  shouldCloseOnOverlayClick,
   ...props
 }: ModalProps): ReactElement {
   const stepTitle = steps ? steps?.[0].key : undefined;
   const tabTitle = tabs ? modalTabTitle(tabs[0]) : undefined;
+  const isMobile = useViewSize(ViewSize.MobileL);
+  const isDrawerOpen = isDrawerOnMobile && isMobile;
```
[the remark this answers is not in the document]
sshanzel: I think it is more preferred to use `is` or `has` since `showDrawer` sounds like it can a function.

packages/shared/src/components/modals/report/ReportModal.tsx
```diff
@@ -55,9 +57,12 @@ export function ReportModal({
...
           data-testid="report_comment"
         />
+        {isMobile ? footer : null}
       </Modal.Body>
-      <Modal.Footer justify={footer ? Justify.Between : Justify.End}>
-        {footer}
+      <Modal.Footer
+        justify={footer && !isMobile ? Justify.Between : Justify.End}
+      >
+        {isMobile ? null : footer}
```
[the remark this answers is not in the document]
sshanzel: That's intentional, the conditions are the opposite.

packages/shared/src/components/modals/common/ModalFooter.tsx
sshanzel: It is highly likely that it won't happen. Since the footer has a top border and is at the bottom-most part. It would clash with the built-in "Close" button.

capJavert (approved):

packages/shared/src/components/checklist/ChecklistCard.spec.tsx
```diff
@@ -67,35 +67,4 @@ describe('ChecklistCard component', () => {
     const progress = await screen.findAllByTestId('checklist-card-progress');
     expect(progress).toHaveLength(2);
   });
-
-  it('should not render close button if onRequestClose is provided', async () => {
```
sshanzel, line 71: Unused now since the recent changes.

packages/shared/src/components/modals/common/Modal.tsx
```diff
@@ -44,35 +47,33 @@ const modalKindAndSizeToOverlayClassName: Partial<
   Record<ModalKind, Partial<Record<ModalSize, string>>>
 > = {
   [ModalKind.FlexibleTop]: {
-    [ModalSize.Small]: 'pt-10 mobileL:pt-20',
-    [ModalSize.Medium]: 'pt-10 mobileL:pt-20',
+    [ModalSize.Small]: 'tablet:pt-20',
+    [ModalSize.Medium]: 'tablet:pt-20',
```
rebelchris, line 51: Are these just generic modal changes in regards to mobileUX?

packages/shared/src/components/modals/common/ModalFooter.tsx
rebelchris: Also wondered if we could just make it dynamic based on high level component or perhpas just not render modalFooter at all.
But might miss some design edge-cases?

packages/shared/src/components/modals/PushNotificationModal.tsx
```diff
@@ -1,10 +1,43 @@
...
+  }
+
+  return (
+    <img
+      className="mx-auto my-14"
+      src={cloudinary.notifications.big}
+      alt="A sample browser notification"
+    />
+  );
+};
```
rebelchris, line 40: Wait this is exactly the same, but one has classes?
Can't we just uniform them? *(or is it a placement issue?)

rebelchris (commented):
Looks a lot better, just wondering if there's no way we could do it reverse?

Have something like ModalOrDrawer

```
if (isDrawer) <Drawer />

<Modal />
```

Because technically they just 2 separate components right, that shouldn't really care about each other.
Just share the same content basically.

packages/shared/src/components/modals/PushNotificationModal.tsx
sshanzel: I did before, but we render two elements at the same time. Then Ante suggested this approach to avoid the double elements. https://github.com/dailydotdev/apps/pull/2815#discussion_r1528325281

packages/shared/src/components/modals/common/Modal.tsx
sshanzel: Yes, all of our Modal is going to be either full screen or drawer on mobile. No more specific width sizes.

DragosIorgulescu (approved):
I don't have much to add other than what has already been discussed, great effort here 👌  to echo what has been said, if we want to further improve on this, we might want to add a tech debt item to avoid delaying the release of mobile UX.

Of course, at some point we will need to take care of these tech debt items 😅

nensidosari (approved):
Nothing much to add here on my end as well, especially after our agreement on [thread](https://dailydotdev.slack.com/archives/C0650M2KX8E/p1710839348107749?thread_ts=1710491202.765389&cid=C0650M2KX8E). 

All the discussions here super valid

sshanzel:
Just working on switching the close button to arrow icon.

rebelchris (approved):

packages/shared/src/components/modals/common/ModalFooter.tsx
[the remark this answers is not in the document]
sshanzel: @dailydotdev/web-team would love to get some thoughts here, just so if no one objects, we can probably start doing it.

PR #2817, opened 2024-03-14
feat: source subscribe push notifications

capJavert:
- Offer to subscribe to push notifications after source subscribe
- Available to test on [preview](https://preview2.app.daily.dev/)

Offer:
image

Subscribed:
image

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

AS-154 #done

### DIFF
--- a/packages/shared/src/components/notifications/EnableNotification.tsx
+++ b/packages/shared/src/components/notifications/EnableNotification.tsx
@@ -48,6 +48,11 @@ const sourceRenderTextCloseButton: Record<NotificationPromptSource, boolean> = {
   [NotificationPromptSource.SourceSubscribe]: true,
 };
 
+const sourceToButtonText: Partial<Record<NotificationPromptSource, string>> = {
+  [NotificationPromptSource.SquadPostModal]: 'Subscribe',
+  [NotificationPromptSource.SourceSubscribe]: 'Enable',
+};
+
 function EnableNotification({
   source = NotificationPromptSource.NotificationsPage,
   contentName,
@@ -80,6 +85,7 @@ function EnableNotification({
   const message = sourceToMessage[source];
   const classes = containerClassName[source];
   const showTextCloseButton = sourceRenderTextCloseButton[source];
+  const buttonText = sourceToButtonText[source] ?? 'Enable notifications';
 
   if (source === NotificationPromptSource.SquadPostModal) {
     return (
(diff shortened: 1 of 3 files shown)


packages/shared/src/components/notifications/EnableNotification.tsx
```diff
@@ -155,18 +178,20 @@ function EnableNotification({
...
+            acceptedJustNow ? 'right-14' : 'right-4',
+          )}
+          src={
+            acceptedJustNow
+              ? cloudinary.notifications.browser_enabled
+              : cloudinary.notifications.browser
+          }
+          alt="A sample browser notification"
+        />
+      )}
```
capJavert, line 194: Only modification I had to do is move the image since it was hidden on mobile usually. But I think its ok compromise.

packages/shared/src/components/notifications/EnableNotification.tsx
```diff
@@ -117,22 +120,40 @@ function EnableNotification({
...
+              >
+                notification settings
+              </a>{' '}
+              can be done anytime through your account details
+            </>
+          ) : (
+            message
+          )}
+        </p>
+        {source === NotificationPromptSource.SourceSubscribe && (
```
rebelchris, line 145: Not urgent, but can we not just have one generic image placement?
(I mean would it be acceptable to have the other types also render like this?

Feels kinda dumb to have different image position only for 1 of the types.

rebelchris (approved):
Cool non blocking, but would be great if somehow we could uniform this stuff :) 
Hate having exceptions just feels like design didn't re-use what we have.

packages/shared/src/components/notifications/EnableNotification.tsx
capJavert: Yeah, agree, but wanted to get it out of the way. Let me take another stab.

capJavert: Not sure how it looks on desktop, if ok I'll see with others if we can go with it.

rebelchris: Yeah non blocking really, also fine if we add a cleanup task somewhere :)

capJavert: @rebelchris since default is absolute position I managed to move it around and have both positions. Still have a bit condition for source but its just on `className`. Let me know what you think.

capJavert: I checked all usages of component and it works.

rebelchris (approved):
Yeah works for now thanks a lot!

PR #2822, opened 2024-03-15
feat: use a Drawer on mobile for the share post menu

DragosIorgulescu:
## Changes

_Note_: I will be pushing separate PRs for each item in the ticket for easier review 🙏 

See slack thread as well: https://dailydotdev.slack.com/archives/C02E2C3C13R/p1710496818220929

- slight refactor of the `useFeedContextMenu`
- ~~TODO: refactor / remove the usage of `react-contextify` -- we'd idealy have access to the state of the menu as well (opened, closed) instead of just the show/hide functions.~~
  - this is handled by the `useContextMenu` hook

### Comment Actions

https://github.com/dailydotdev/apps/assets/1225100/445ad1f1-0571-4cb9-af2c-1120abba2595

### Squad Actions Menu

https://github.com/dailydotdev/apps/assets/1225100/15d48b07-c705-400e-a353-0758ecc68093

## Events

N / A

- [x] MobileL (420px)
- [x] Tablet (656px)
- [x] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

ref MI-128


packages/shared/src/components/Feed.tsx
```diff
@@ -250,8 +250,10 @@ export default function Feed<T>({
     onMenuClick,
     onShareMenuClick,
     postMenuIndex,
-    postMenuLocation,
+    postShareMenuIndex,
+    postShareMenuLocation,
     setPostMenuIndex,
+    setPostShareMenuIndex,
```
rebelchris, line 256: Ideally we want to get rid of more props here.
Rather think some state hook would be better as discussed on slack.

Maybe also take a quick look at `Lean feed` think I had this issue as well and solved it like that.
Basically each component can rather have it's own logic vs prop drilling.

Ref:
https://github.com/dailydotdev/apps/pull/2279

packages/shared/src/components/ShareOptionsMenu.tsx
```diff
@@ -90,6 +96,19 @@ export default function ShareOptionsMenu({
     action: trackAndCopyLink,
   });
 
+  const isMobile = useViewSize(ViewSize.MobileL);
+
+  if (isMobile) {
```
rebelchris, line 101: Is there anyway we can put such logic rather in contextMenuDrawer or some shared component?
(Same note as on the Modal one)

rebelchris (changes requested):
Let's see if we can rather solve it differently and avoid having to prop drill more.
Also subnote on the context menu usage

packages/shared/src/components/ShareOptionsMenu.tsx
DragosIorgulescu: The only option we have is to move this up the chain inside the `PortalMenu` component. I updated this PR to reflect that, let me know what you think

packages/shared/src/components/fields/PortalMenu.tsx
```diff
@@ -3,19 +3,42 @@ import { Item, Menu, MenuProps } from '@dailydotdev/react-contexify';
...
+import {
+  ContextMenuDrawer,
+  ContextMenuDrawerItem,
+} from '../drawers/ContextMenuDrawer';
 
 export default function PortalMenu({
   id,
   onHidden: onHiddenProps,
+  isOpen,
+  drawerOptions,
```
DragosIorgulescu, line 16: Added this as a separate item from `options` because we have slight inconsistencies between the two interfaces, but I am open to consolidating these if everyone is happy with this approach

rebelchris: I'm quite happy with this yes.
Basically if I remember correct the PortalMenu is only used as `ContextMenu` right? (Maybe we should also rename it at one stage)

But yes let's see what others think and we can clean it up a bit.

packages/shared/src/components/PostOptionsMenu.tsx
```diff
@@ -458,34 +460,16 @@ export default function PostOptionsMenu({
...
-      {postOptions.map(({ icon, label, action }) => (
-        <Item key={label} className="typo-callout" onClick={action}>
-          <span className="flex w-full items-center typo-callout">
-            {icon} {label}
-          </span>
-        </Item>
-      ))}
-    </PortalMenu>
+      drawerOptions={postOptions}
+      options={postOptions}
```
DragosIorgulescu, line 471: I'll consolidate these into using just `options`, `drawerOptions` was a temporary solution until I got everything working

packages/shared/src/components/PostOptionsReadingHistoryMenu.tsx
```diff
@@ -22,9 +22,14 @@ import {
...
 import { Origin } from '../lib/analytics';
+import useContextMenu from '../hooks/useContextMenu';
+import { ContextMenu as ContextMenuIds } from '../hooks/constants';
 
-const PortalMenu = dynamic(
-  () => import(/* webpackChunkName: "portalMenu" */ './fields/PortalMenu'),
+const ContextMenu = dynamic(
+  () =>
+    import(/* webpackChunkName: "portalMenu" */ './fields/PortalMenu').then(
+      (module) => module.ContextMenu,
```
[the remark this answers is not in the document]
rebelchris: Also what are we loading a ContextMenu or a PortalMenu?

packages/shared/src/hooks/useReadingHistoryContextMenu.ts
```diff
@@ -25,10 +26,7 @@ export default function useReadingHistoryContextMenu(): {
   ): void => {
     setReadingHistoryContextItem(readingHistory);
     setQueryIndexes(indexes);
-    const { right, bottom } = event.currentTarget.getBoundingClientRect();
-    showTagOptionsMenu(event, {
-      position: { x: right, y: bottom + 4 },
```
rebelchris, line 30: What about the position removal here?

packages/shared/src/hooks/feed/useFeedContextMenu.ts
```diff
@@ -62,21 +65,21 @@ export default function useFeedContextMenu(): FeedContextMenu {
...
   ) => {
     if (postMenuIndex === index) {
-      setPostMenuLocation(null);
+      setPostShareMenuLocation(null);
       return;
     }
-    setPostMenuLocation({ index, row, column });
-    const { right, bottom } = e.currentTarget.getBoundingClientRect();
-    showShareMenu(e, {
-      position: { x: right, y: bottom + 4 },
```
rebelchris, line 71: Here as well, I think we'll need this right?

rebelchris (commented):
Looks a lot cleaner, let's indeed see if we can make the props shared between 2 interfaces, otherwise some minor questions.
But looks good overall.

packages/shared/src/hooks/useReadingHistoryContextMenu.ts
DragosIorgulescu: the position arguments are being handled inside the `onMenuClick` function definition in the `useContextMenu` hook -- so this is simply removing duplicated code: https://github.com/dailydotdev/apps/blob/b3d1c8d32112083cdc6fd8c0a3220f8dd4a8e2fe/packages/shared/src/hooks/useContextMenu.ts#L41-L53

packages/shared/src/hooks/feed/useFeedContextMenu.ts
DragosIorgulescu: same as above, this is already handled by `onMenuClick`: https://github.com/dailydotdev/apps/blob/b3d1c8d32112083cdc6fd8c0a3220f8dd4a8e2fe/packages/shared/src/hooks/useContextMenu.ts#L41-L53

packages/shared/src/components/PostOptionsReadingHistoryMenu.tsx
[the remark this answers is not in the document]
DragosIorgulescu: I'm using `ContextMenu` here for consistency, under the hood it wraps `PortalMenu`  and adds a bit more functionality: https://github.com/dailydotdev/apps/blob/91bd6478d5fab8b2dc0f89a4c2e2767f3d68e882/packages/shared/src/components/fields/PortalMenu.tsx#L72-L82

Moving from `PortalMenu` to `ContextMenu` here is aimed to bring consistency everywhere we use menus.

packages/shared/src/components/filters/TagOptionsMenu.tsx
```diff
@@ -29,30 +33,46 @@ export default function TagOptionsMenu({
...
+    tagOptions.push({ label: 'Unfollow', action: onUnfollow });
+  }
+
+  if (onBlock) {
+    tagOptions.push({ label: 'Block', action: onBlock });
+  }
+
+  if (onUnblock) {
+    tagOptions.push({ label: 'Unblock', action: onUnblock });
+  }
```
nensidosari, line 64: We could extract this in a separate func, a helper or sth to gives the tag options

packages/shared/src/components/tooltips/notifications/NotificationPreferenceMenu.tsx
```diff
@@ -67,36 +68,31 @@ export const NotificationPreferenceMenu = ({
...
       disableBoundariesCheck
       id={contextId}
       className="menu-primary"
       animation="fade"
       onHidden={onClose}
-    >
-      <Item className="w-64 py-1 typo-callout" onClick={onItemClick}>
-        <span className="flex w-full flex-row items-center gap-1">
-          <Icon />
-          <Copy />
```
nensidosari, line 97: This seem to have 2 icons, why only one now?

packages/webapp/pages/notifications.tsx
```diff
@@ -119,8 +119,7 @@ const Notifications = (): ReactElement => {
     }
 
     setNotification(item);
-    const { right, bottom } = e.currentTarget.getBoundingClientRect();
-    show(e, { position: { x: right, y: bottom + 4 } });
+    onMenuClick(e);
```
nensidosari, line 122: Noticed u rename this in other files, and not here.. any reason? Let's be consistent maybe and rename also here based on the menu it opens..

packages/shared/src/components/PostOptionsReadingHistoryMenu.tsx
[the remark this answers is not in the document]
nensidosari: So in this case we are keeping PortalMenu as default export since it's usage is still not replaced everywhere with Context menu?

nensidosari (approved):
Looks really good to me, minor comments/questions that are non blocking or anything.

packages/shared/src/components/PostOptionsReadingHistoryMenu.tsx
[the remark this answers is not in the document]
rebelchris: Yeah either Nensi's suggestion or can we split them in 2 for now?
(just want to avoid confusion down the line)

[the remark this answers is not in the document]
DragosIorgulescu: At this stage with this PR the original PortalMenu is no longer used anywhere, so we could remove it from exports and set the ContextMenu as the default export -- I was just waiting to see everyones thoughts on the review stage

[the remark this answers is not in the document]
rebelchris: Cool so let's remove/rename it so it becomes ContextMenu and revamp the module use :)

packages/shared/src/components/tooltips/notifications/NotificationPreferenceMenu.tsx
DragosIorgulescu: Copy was not an icon here, but rather a variable for the label, the naming was a bit confusing, although 'copy' is a valid name, I initially confused it for an icon too, so I renamed to 'label' as seen above 🙏

packages/shared/src/components/PostOptionsReadingHistoryMenu.tsx
```diff
@@ -114,34 +114,35 @@ export default function PostOptionsReadingHistoryMenu({
...
+    },
+    {
+      icon: <MenuIcon Icon={XIcon} />,
+      label: 'Remove post',
+      action: () => onHideHistoryPost(post?.id),
+    },
+  ];
+
   return (
     <>
```
sshanzel, line 136: Non-blocking. Can you maybe remove the extra Fragment? I understand this was a legacy thing, just for cleanup 😄

packages/shared/src/components/filters/TagOptionsMenu.tsx
sshanzel: Probably a useMemo can help?

sshanzel (approved):
Love the refactor of legacy implementations on how this turned out 🚀 

I have nothing blocking. There should be a checklist in the ticket for all the found context menus. Once checked they are all working as intended, we should be good (after working on other's comments too).

packages/shared/src/components/ShareOptionsMenu.tsx
```diff
@@ -15,8 +14,10 @@ import { useFeature } from './GrowthBookProvider';
 import { feature } from '../lib/featureManagement';
 import { useFeedLayout, useGetShortUrl } from '../hooks';
 import { ReferralCampaignKey } from '../lib/referral';
+import { ContextMenu as ContextMenuIds } from '../hooks/constants';
+import useContextMenu from '../hooks/useContextMenu';
 
-const PortalMenu = dynamic(
+const ContextMenu = dynamic(
   () => import(/* webpackChunkName: "portalMenu" */ './fields/PortalMenu'),
```
rebelchris, line 21: Let's also rename these please 🙏

packages/shared/src/components/PostOptionsReadingHistoryMenu.tsx
```diff
@@ -22,8 +21,10 @@ import {
   updateInfiniteCache,
 } from '../lib/query';
 import { Origin } from '../lib/analytics';
+import useContextMenu from '../hooks/useContextMenu';
+import { ContextMenu as ContextMenuIds } from '../hooks/constants';
 
-const PortalMenu = dynamic(
+const ContextMenu = dynamic(
   () => import(/* webpackChunkName: "portalMenu" */ './fields/PortalMenu'),
```
rebelchris, line 28:
```suggestion
  () => import(/* webpackChunkName: "contextMenu" */ './fields/ContextMenu'),
```

packages/shared/src/components/PostOptionsMenu.tsx
```diff
@@ -60,13 +57,13 @@ import { ActiveFeedContext } from '../contexts';
...
-import { RootPortal } from './tooltips/Portal';
 import { SourceSubscribeExperiment } from '../lib/featureValues';
+import { ContextMenu as ContextMenuTypes } from '../hooks/constants';
+import useContextMenu from '../hooks/useContextMenu';
 import { SourceType } from '../graphql/sources';
 import { withExperiment } from './withExperiment';
 
-const PortalMenu = dynamic(
+const ContextMenu = dynamic(
   () => import(/* webpackChunkName: "portalMenu" */ './fields/PortalMenu'),
```
rebelchris, line 67:
```suggestion
  () => import(/* webpackChunkName: "contextMenu" */ './fields/ContextMenu'),
```

packages/shared/src/components/filters/TagOptionsMenu.tsx
```diff
@@ -1,11 +1,12 @@
...
-import Link from 'next/link';
 import { Tag } from '../../graphql/feedSettings';
 import { getTagPageLink } from '../../lib/links';
+import { MenuItemProps } from '../fields/PortalMenu';
+import useContextMenu from '../../hooks/useContextMenu';
+import { ContextMenu as ContextMenuIds } from '../../hooks/constants';
 
-const PortalMenu = dynamic(
+const ContextMenu = dynamic(
   () => import(/* webpackChunkName: "portalMenu" */ '../fields/PortalMenu'),
```
rebelchris, line 10:
```suggestion
  () => import(/* webpackChunkName: "contextMenu" */ '../fields/ContextMenu'),
```

rebelchris (changes requested):

packages/shared/src/components/PostOptionsMenu.tsx
```diff
@@ -59,14 +56,14 @@ import { ActiveFeedContext } from '../contexts';
...
 import { SourceSubscribeExperiment } from '../lib/featureValues';
+import { ContextMenu as ContextMenuTypes } from '../hooks/constants';
+import useContextMenu from '../hooks/useContextMenu';
 import { SourceType } from '../graphql/sources';
 import { withExperiment } from './withExperiment';
 
-const PortalMenu = dynamic(
-  () => import(/* webpackChunkName: "portalMenu" */ './fields/PortalMenu'),
+const ContextMenu = dynamic(
+  () => import(/* webpackChunkName: "ContextMenu" */ './fields/PortalMenu'),
```
rebelchris, line 66: Very minor, but it's a common pattern we keep
Second part ofcourse when we change filename :) 
```suggestion
  () => import(/* webpackChunkName: "contextMenu" */ './fields/ContextMenu'),
```

packages/shared/src/components/fields/PortalMenu.tsx
```diff
@@ -33,44 +39,72 @@ export interface MenuItemProps<
...
 }
 
 interface ContextMenuProps extends Omit<MenuProps, 'children'> {
-  options: MenuItemProps[];
+  options: MenuItemProps[] | ContextMenuDrawerItem[];
+  isOpen: boolean;
 }
 
-export const ContextMenu = ({
+export default function ContextMenu({
```
rebelchris, line 54: Can we also change the filename please to reflect `ContextMenu`?

rebelchris (changes requested):
Just one remaining point from my side 🙏

rebelchris (approved):
Awesome!

PR #2824, opened 2024-03-15
fix: share tracking on share loops

sshanzel:
## Changes
- At first, I thought of using mutation to only get the short link after clicking the button. But then the component we are using to copy the link is used in other places as well. Might affect the other implementations.
- Introduced a hook to manage the link.

Deployed at preview3.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done


packages/shared/src/hooks/post/useTrackedCopyPostLink.ts
```diff
@@ -1,23 +1,40 @@
...
+  origin = Origin.FeedbackCard,
+): UseTrackedCopyPostLink => {
+  const { shareLink, isLoading } = useTrackedLink({
+    link: post.commentsPermalink,
+    cid: ReferralCampaignKey.SharePost,
+  });
+  const [, copyLink] = useCopyPostLink(shareLink);
   const { trackEvent } = useContext(AnalyticsContext);
 
-  return (provider: ShareProvider) => {
```
sshanzel, line 15: The only thing that changed here was the `origin` and had to put it in a `useCallback`.

packages/shared/src/hooks/utils/useTrackedLink.ts
```diff
@@ -0,0 +1,36 @@
...
+interface UseTrackedLink {
+  shareLink: string;
+  isLoading: boolean;
+}
+
+export const useTrackedLink = ({
+  enabled = true,
+  link,
+  cid,
+}: UseTrackedLinkProps): UseTrackedLink => {
```
capJavert, line 23: This is another hook we introduce for links, feels like it is getting out of hand 😆 

I think we should be able to somehow manage it from `useGetShortUrl`and implement the logic specific to this one component.

Also that hook already stores short url under `RequestKey.ShortUrl` key so we for sure don't need `TrackingLink`. 

Personally I think just having logic

sshanzel: Yep, I don't even know how much hook we have that processes copying links 😅

sshanzel: Makes sense. There is actually quite a bit to refactor here. For example, this function could have been in a `useMutation` hook than `queryClient.fetchQuery` so we can track loading phase.

I would love to refactor everything here now, but can we maybe create another ticket instead just so we can merge this as they are considering this issue as something critical that we should fix asap. 

Screenshot 2024-03-15 at 9 36 28 PM

sshanzel: I can even do an immediate follow up PR after this gets merged.

capJavert: I agree we should keep the footprint small for this, so I think we can just avoid adding any hooks in the first place and implement logic you added in that one component?

capJavert: Just because I am sure by adding yet another hook we will confuse the usage even more 😆

rebelchris: Yeah I wouldn't rush to push this out.
Rather fix it properly I think 🤔

sshanzel: Yeah, sure, I'd be glad to do the refactoring here, I was just considering the product's urgency on the matter.

capJavert: No need to refactor here, just move the hook logic to that single component you are using it in? To not block you and we can add/talk about refactoring as tech debt. Would that work?

sshanzel: Feels like it is the same thing with the number of code changes with the existing hook, which is basically another thing to clean up since the hook is now doing two things and we don't even need the original callback.

packages/shared/src/lib/query.ts
```diff
@@ -93,6 +93,7 @@ export enum RequestKey {
   Source = 'source',
   OneSignal = 'onesignal',
   ActiveUsers = 'active_users',
+  TrackingLink = 'tracking_link',
```
capJavert, line 96: Leftover

packages/shared/src/hooks/utils/useGetShortUrl.ts
```diff
@@ -7,45 +7,80 @@ import { GET_SHORT_URL_QUERY } from '../../graphql/urlShortener';
...
       } catch (err) {
-        return trackingUrl;
+        return trackedUrl;
       }
     },
-    [queryClient, isAuthReady, user],
+    [isAuthReady, user, getProps, queryClient],
+  );
+
+  const isEnabled = query.enabled ?? true;
```
capJavert, line 72: Should be `false` by default?

packages/shared/src/hooks/utils/useGetShortUrl.ts
```diff
@@ -7,45 +7,80 @@ import { GET_SHORT_URL_QUERY } from '../../graphql/urlShortener';
...
+  const { queryKey, trackedUrl } = query
+    ? getProps(query.url, query.cid)
+    : { queryKey: [], trackedUrl: '' };
+  const { data: shareLink, isLoading } = useQuery(
+    queryKey,
+    () => queryShortUrl(trackedUrl),
+    {
+      ...disabledRefetch,
+      enabled: !!query?.url && isEnabled && !!user,
```
capJavert, line 81: We can also add `staleTime: Infinity` as in `fetchQuery`.

capJavert (commented):
Nice direction 👌

packages/shared/src/hooks/utils/useGetShortUrl.ts
sshanzel: Should be true by default, we check whether the query object anyway whether it has value or not.

capJavert (approved):

packages/shared/src/hooks/utils/useGetShortUrl.ts
sshanzel: Fixed 🚢

rebelchris (approved):
Nice thanks for the retro.
We can still add backlog ticket to look at all share/copy/link stuff and see if we can optimize it better. (non urgent)

PR #2826, opened 2024-03-18
fix: remove stale time on streaks if we have an alert

DragosIorgulescu:
## Changes

Self-explanatory

## Events

N / A

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

MI-232 #done


packages/shared/src/hooks/streaks/useReadingStreak.ts
```diff
@@ -10,13 +10,16 @@ interface UserReadingStreak {
...
   const { user } = useAuthContext();
   const { shouldShowStreak } = useStreakExperiment();
   const { data: streak, isLoading } = useQuery(
     generateQueryKey(RequestKey.UserStreak, user),
     getReadingStreak,
-    { enabled: shouldShowStreak, staleTime: StaleTime.Default },
+    {
+      enabled: shouldShowStreak,
+      staleTime: skipCache ? null : StaleTime.Default,
+    },
```
capJavert, line 22: I think it will be better to move the alerts logic inside the hook. Like this if any of the hooks without `skipCache` passed executes first it will still have stale time. That way also all hooks will behave the same based on alerts.

DragosIorgulescu: @capJavert I changed this to just invalidating the query when the milestone modal needs to be displayed, wanted to avoid polluting this hook with alert related business logic. WDYT?

capJavert (approved):

packages/shared/src/hooks/streaks/useStreakMilestone.ts
```diff
@@ -36,6 +41,12 @@ export const useStreakMilestone = (): void => {
       return;
     }
 
+    // since the streaks query is cached and has a staleTime set
+    // this is a good time to invalidate that query
+    queryClient.invalidateQueries({
+      queryKey: generateQueryKey(RequestKey.UserStreak, user),
+    });
```
rebelchris, line 48: I think if we know the action we prefer to mutate rather.
Which I think we do know here, but it's non blocking. (Do note if you not going to change it, maybe we can log it for backlog optimization)

rebelchris (approved):
One non blocking comment

packages/shared/src/hooks/streaks/useStreakMilestone.ts
DragosIorgulescu: As discussed in slack, this is not applicable here unfortunately since we are not relying on a mutation triggered by the client to update streaks, instead these are updated async via the `viewPost` mutation.

PR #2831, opened 2024-03-18
fix: prevent injection of scripts over and over

omBratteng:
😢

### DIFF
--- a/packages/extension/src/lib/extensionScripts.ts
+++ b/packages/extension/src/lib/extensionScripts.ts
@@ -53,6 +53,7 @@ export const registerBrowserContentScripts = async (): Promise<void> => {
       },
     ]);
   }
+  hasInjectedScripts = true;
 
   return null;
 };
--- a/packages/extension/src/lib/extensionScripts.ts
+++ b/packages/extension/src/lib/extensionScripts.ts
@@ -20,7 +20,7 @@ export const HOST_PERMISSIONS = [
   'https://*.daily.dev/*',
 ];
 
-const hasInjectedScripts = false;
+let hasInjectedScripts = false;
 const companionScriptId = 'daily-companion-app';
 
 export const registerBrowserContentScripts = async (): Promise<void> => {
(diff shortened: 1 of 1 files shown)


packages/extension/src/lib/extensionScripts.ts
```diff
@@ -20,7 +20,7 @@ export const HOST_PERMISSIONS = [
   'https://*.daily.dev/*',
 ];
 
-const hasInjectedScripts = false;
+let hasInjectedScripts = false;
```
rebelchris, line 23: This even looks like it was meant to do something like this? 🤔

rebelchris (approved):
Worth a shot by now.

packages/extension/src/lib/extensionScripts.ts
omBratteng: Yeah. It did it originally
https://github.com/dailydotdev/apps/blob/111feb347c5cd312a6cc6d75a54043005dbd5a3c/packages/extension/src/lib/extensionScripts.ts#L26-L35

omBratteng: Probably lint auto-fix that saw a let that wasn't re-written and converted it to a const. That's what happens locally with me if I remove the `hasInjectedScripts = true;`, then it's convert to `const hasInjectedScripts = false;`.

omBratteng: And then no-one caught the change

DragosIorgulescu (approved):
Great work on this investigation, and sorry for introducing this issue 🤦‍♂️

packages/extension/src/lib/extensionScripts.ts
```diff
@@ -53,6 +53,7 @@ export const registerBrowserContentScripts = async (): Promise<void> => {
       },
     ]);
   }
+  hasInjectedScripts = true;
```
omBratteng, line 56: Could probably remove this one.

omBratteng: Removed it, didn't do anything really.

kopancek (approved):
Great catch!

PR #2832, opened 2024-03-19
fix: split if name defined

nensidosari:
fix: split if name defined

https://dailydotdev.slack.com/archives/C01L0QXQJNB/p1710834548807209


idoshamun (approved):

DragosIorgulescu (approved):

rebelchris (approved):
Interesting there are cases where name is empty? 🤯

PR #2835, opened 2024-03-20
fix: scroll to top button size on mobile

sshanzel:
## Changes
Preview:

Screenshot 2024-03-20 at 4 12 08 PM

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done

### DIFF
--- a/packages/shared/src/components/ScrollToTopButton.tsx
+++ b/packages/shared/src/components/ScrollToTopButton.tsx
@@ -17,7 +17,13 @@ export default function ScrollToTopButton(): ReactElement {
   const [show, setShow] = useState(false);
   const isLaptop = useViewSize(ViewSize.Laptop);
   const isTablet = useViewSize(ViewSize.Tablet);
-  const touchDeviceSize = isTablet ? ButtonSize.Large : ButtonSize.Small;
+  const size = (() => {
+    if (isLaptop) {
+      return ButtonSize.XLarge;
+    }
+
+    return isTablet ? ButtonSize.Large : ButtonSize.Small;
+  })();
 
   useEffect(() => {
     const callback = () => {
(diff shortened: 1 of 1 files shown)


packages/shared/src/components/ScrollToTopButton.tsx
```diff
@@ -16,6 +16,8 @@ const baseStyle: CSSProperties = {
 export default function ScrollToTopButton(): ReactElement {
   const [show, setShow] = useState(false);
   const isLaptop = useViewSize(ViewSize.Laptop);
+  const isTablet = useViewSize(ViewSize.Tablet);
+  const toucheDeviceSize = isTablet ? ButtonSize.Large : ButtonSize.Small;
```
nensidosari, line 20: Pardon my english, what does touche mean in this case? 😅 
Personally I would do all conditions in one also for the laptop, and use just the variable below.

nensidosari (approved):
All good, one suggestion, and some naming question from my side.

packages/shared/src/components/ScrollToTopButton.tsx
sshanzel: Ah, should be `touch`. Lol, I misspelled touch device.

> Personally I would do all conditions in one also for the laptop, and use just the variable below.

Do you mean in one line? I think it can become confusing, and our lint doesn't accept multiple ternaries in a single line. Or do you mean something else?

nensidosari: Yeah, that's what I ment, but ok like this as well. Just feels weird we have a variable for mobile tablet size but check condition separately for laptop

sshanzel: Yeah, same sentiments. I even thought, we can maybe use a function.

```
function getSize() {
  if (laptop) return xl;
  if (tablet) return l;

  return s;
}
```

Wdyt?

nensidosari: Yeah looks good

PR #2840, opened 2024-03-20
refactor: squad directory on tablet and mobile

sshanzel:
## Changes
- Should be ready for initial review but I still have some concerns I will relay to the design team. Regarding width size - so on that side, feel free to ignore it.
- Deployed on preview. Quick link: https://preview.app.daily.dev/squads

Previews:
Screenshot 2024-03-20 at 11 51 18 PM
Screenshot 2024-03-20 at 11 51 15 PM
Screenshot 2024-03-20 at 11 51 05 PM
Screenshot 2024-03-20 at 11 51 00 PM
Screenshot 2024-03-20 at 11 50 53 PM

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

MI-132 #done

### DIFF
--- a/packages/shared/src/components/squads/SquadsDirectoryHeader.tsx
+++ b/packages/shared/src/components/squads/SquadsDirectoryHeader.tsx
@@ -1,52 +1,59 @@
 import React, { ReactElement } from 'react';
+import classNames from 'classnames';
 import { cloudinary } from '../../lib/image';
 import { Button, ButtonVariant } from '../buttons/Button';
 import SourceBetaIcon from '../../../icons/source_beta.svg';
 import { Origin } from '../../lib/analytics';
 import { useSquadNavigation } from '../../hooks';
 
-export const SquadsDirectoryHeader = (): ReactElement => {
+interface SquadsDirectoryHeaderProps {
+  className?: string;
+}
+
+export const SquadsDirectoryHeader = ({
+  className,
+}: SquadsDirectoryHeaderProps): ReactElement => {
   const { openNewSquad } = useSquadNavigation();
 
   return (
-    <div className="mb-4">
-      <div
-        className="relative mb-16 flex flex-col items-center rounded-24 bg-background-default bg-cover bg-center p-6 pb-1 text-center"
-        style={{
-          backgroundImage: `url(${cloudinary.squads.directory.banner})`,
-        }}
-      >
-        <div className="relative z-1 flex flex-col items-center">
-          <SourceBetaIcon
-            width="auto"
-            height="3rem"
-            className="mb-4 translate-x-[1.125rem] text-white"
-          />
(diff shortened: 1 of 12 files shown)


packages/shared/src/lib/query.ts
```diff
@@ -15,7 +15,7 @@ import { SharedFeedPage } from '../components/utilities';
 
 export enum OtherFeedPage {
   Tag = 'tag',
-  Squad = 'squad',
```
sshanzel, line 18: Was non-existent. After checking, the value in the `feedName` context is `squads`.

packages/shared/src/components/squads/SquadsDirectoryHeader.tsx
```diff
@@ -1,52 +1,59 @@
...
+  className?: string;
+}
+
+export const SquadsDirectoryHeader = ({
+  className,
+}: SquadsDirectoryHeaderProps): ReactElement => {
   const { openNewSquad } = useSquadNavigation();
 
   return (
-    <div className="mb-4">
```
sshanzel, line 12: Diff looks crazy, but the major difference is just the removal of this unnecessary wrapper and allowing to pass `className`.

nensidosari:
Screenshot 2024-03-21 at 9 12 22 AM
Maybe something u're already checking with design, but this was the only thing that caught my eye that doesn't look 100% right. 

The border surrounding the cards, and then no space between the cards itself, I know this is how mobile feed looks like, but here doesn't look that goood on my opinion.

packages/webapp/pages/squads/index.tsx
```diff
@@ -15,15 +15,33 @@ import {
   SourceCard,
```
nensidosari, line 15: Feel like our components always end up with lots of lines of code making it harder to read....

nensidosari (approved):
Looks good to me

rebelchris (changes requested):
Agree with Nensi, spacing on mobile for public squads looks very odd.
I'll block it until we resolve that 🙏

packages/webapp/pages/squads/index.tsx
sshanzel: We can probably break this file down in a separate PR. Would be glad to do that.

sshanzel:
After checking with the design team, it seems we should not use full-width on single columns.
[Thread](https://dailydotdev.slack.com/archives/C0650M2KX8E/p1711012731064069?thread_ts=1710996225.906289&cid=C0650M2KX8E)

Screenshot 2024-03-21 at 5 22 18 PM

sshanzel:
Screenshot 2024-03-21 at 5 24 34 PM

rebelchris (approved):
Now it looks fine indeed!

nensidosari (approved):
Looks good to me now as well

PR #2845, opened 2024-03-21
refactor: post header navigation

sshanzel:
## Changes
- Unify the header navigation on post pages (note: post modal is not included).
- The custom navigation is solely for post pages navigation which made the refactor easier.
- Introduce a new component.
- Will deploy in preview once one becomes available.

Screenshot 2024-03-22 at 12 50 56 AM
Screenshot 2024-03-22 at 12 50 51 AM
Screenshot 2024-03-22 at 12 50 31 AM
Screenshot 2024-03-22 at 12 50 24 AM
Screenshot 2024-03-22 at 12 50 19 AM

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

MI-143 #done


packages/shared/src/components/post/PostContentHeader.tsx
```diff
@@ -0,0 +1,50 @@
...
+      <Button
+        icon={<ArrowIcon className="-rotate-90" />}
+        size={ButtonSize.Small}
+        variant={ButtonVariant.Tertiary}
+        onClick={canGoBack ? router.back : onBackClick}
+      />
+      <PostHeaderActions
+        post={post}
+        className="ml-auto"
+        contextMenuId="post-page-header-actions"
```
rebelchris, line 45: Just to note this might interfear with @DragosIorgulescu his PR on uniforming/enuming these context menu ids? (not sure, just raising it)

packages/shared/src/components/post/PostContentHeader.tsx
```diff
@@ -0,0 +1,50 @@
...
+import { PostNavigationProps } from './common';
+import { SourceType } from '../../graphql/sources';
+
+export function PostContentHeader({
+  post,
+  onReadArticle,
+}: Pick<PostNavigationProps, 'onReadArticle' | 'post'>): ReactElement {
+  const router = useRouter();
+  const canGoBack = !!globalThis?.window?.history?.length;
+  const routedFromSquad = router?.query?.squad;
```
rebelchris, line 15: Before we also had from notification right? (not sure we need it, just raising it)

rebelchris: And button will no longer have copy? (eg back or back to squad?)

packages/shared/src/components/post/PostContentHeader.tsx
```diff
@@ -0,0 +1,50 @@
...
+  const squadLink = `/squads/${router.query.squad}`;
+
+  if (!canGoBack && !routedFromSquad && post.source.type !== SourceType.Squad) {
```
rebelchris, line 18: Should it not be `||` on the sourcetype?
I think the logic is something like even if you end from google on a squad post we still show back to squad there right?

rebelchris (commented):
Mostly some questions,. code itself looks good.

packages/shared/src/components/post/PostContentHeader.tsx
sshanzel: Going back to the notifications should be automatic since it will be a part of the history.

Yeah, no more copy, as per design. I think the reason is that "back to squad" is a bit long and might interfere with the action buttons.

packages/shared/src/components/post/PostContentHeader.tsx
sshanzel: > Should it not be || on the sourcetype?

The issue would be the same for any type, the reason Squad is allowed is we can go back to its Squad itself.

> if you end from google on a squad post we still show back to squad there right?

Yeah, was thinking of that too, I tried to check whether the back history is within the app, but didn't find any quick way.

packages/shared/src/components/post/BasePostContent.tsx
```diff
@@ -35,7 +36,9 @@ export function BasePostContent({
 
   return (
     <>
-      {customNavigation ?? (
+      {isPostPage ? (
+        <PostContentHeader post={post} onReadArticle={onSharePost} />
```
nensidosari: How does onReadArticle link to onSharePost, a bit confusing..

packages/shared/src/components/post/PostContentHeader.tsx
```diff
@@ -0,0 +1,50 @@
...
+    return null;
+  }
+
+  const onBackClick = () => {
```
nensidosari, line 22: the conditions here seem to be related only to squad post types or routed from squads else it returns null, should it just go back to history if there is history?

packages/shared/src/components/post/PostContentHeader.tsx
```diff
@@ -0,0 +1,50 @@
...
+    return null;
+  };
+
+  return (
+    <span className="-mx-4 flex flex-row items-center border-b border-border-subtlest-tertiary px-4 py-2 tablet:hidden">
+      <Button
+        icon={<ArrowIcon className="-rotate-90" />}
+        size={ButtonSize.Small}
+        variant={ButtonVariant.Tertiary}
+        onClick={canGoBack ? router.back : onBackClick}
```
nensidosari, line 40: Ah I see the condition I mentioned above it here. I think is best to have all conditions on the onBackClick function else it's confusing. 

Also in this case we are sure that history takes precedence over squad links if it's a squad post?

packages/shared/src/components/post/PostContentHeader.tsx
nensidosari: Nvm saw the other condition below, though think is best if all conditions are here

nensidosari (commented):
Looks good overall. Few questions and concerns in my side

Ps.
Also on your description, when you say post modal navigation is not included, what does that mean? Will the navigation on post modal still continue to work as before? Do we plan to do something different to it.

packages/shared/src/components/post/BasePostContent.tsx
nensidosari: Seems like u already fixed this while I was reviewing

packages/shared/src/components/post/PostContentHeader.tsx
sshanzel: Yeah, that makes sense. Let's do that.

packages/shared/src/components/post/PostContentHeader.tsx
sshanzel: > That's what we do at the moment, so I believe that should be the case here as well. Just not in this condition

It is still the same condition (PostBackButton) we have in the past, just mixed all together now rather than being on different components.

packages/shared/src/components/post/BasePostContent.tsx
```diff
@@ -35,7 +36,12 @@ export function BasePostContent({
 
   return (
     <>
-      {customNavigation ?? (
+      {isPostPage ? (
+        <PostContentHeaderMobile
```
rebelchris, line 40: As the component is called mobile, shouldn't we maybe make it dynamic and then only show if isMobile?
(Optional idea)

rebelchris (approved):
Just one idea, not blocking :)

nensidosari (approved):
Looks good 🎉 

Had only one question not answered above @sshanzel 

> Also on your description, when you say post modal navigation is not included, what does that mean? Will the navigation on post modal still continue to work as before? Do we plan to do something different to it.

packages/shared/src/components/post/BasePostContent.tsx
sshanzel: Yeah, that makes sense, though, in the component itself, we have `tablet:hidden` - so it shouldn't show. I tried to use CSS over JS but we can make an exception here. Wdyt?

sshanzel:
> Looks good 🎉
>
> Had only one question not answered above @sshanzel

Oh, I missed it. My bad. It means there are no changes required for the post modal (on mobile we redirect so they never see the modal) since this is for post pages.

packages/shared/src/components/post/BasePostContent.tsx
rebelchris: Not sure how many other things it loads, so either way works for me 🙏
EG if that component would do a lot of computing i would say move it up, if it's mostly rendering don't worry about it.

sshanzel: Ah, that is a good point about it. The component is fairly simple and doesn't do much, so I think it is safe to say we should be good with just having it hidden.

PR #2847, opened 2024-03-22
feat: comment feed

rebelchris:
- Added additional comment feed for users enrolled in this experiment
- It's the quick-n-dirty implementation but tried to keep the mess to a minimum. (Ideally lean feed would make this a bit easier to organize, took some learnings from here)
- Sidebar menu has a `new` badge until you click it (uses actions)
- Refactored profile replies screen to re-use this feed

Screenshot 2024-03-22 at 13 18 25

Placeholder vs loaded
Screenshot 2024-03-22 at 13 00 21

Try it out at: (Please wait till API is merged https://github.com/dailydotdev/daily-api/pull/1767)
https://preview3.app.daily.dev/

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

AS-202 #done


packages/shared/src/components/MainFeedLayout.tsx
```diff
@@ -240,24 +256,43 @@ export default function MainFeedLayout({
...
         />
+      ) : (
+        feedProps && (
+          <Feed
+            {...feedProps}
+            className={classNames(
+              shouldUseMobileFeedLayout && !isFinder && 'laptop:px-6',
+            )}
+          />
+        )
```
DragosIorgulescu, line 295: More for my understanding, I am not sure I understand why this change is needed if the comment feed is just added to the user replies page?

packages/shared/src/components/sidebar/DiscoverSection.tsx
```diff
@@ -51,6 +50,40 @@ export function DiscoverSection({
...
+  } else {
+    discoverMenuItems.splice(2, 0, {
+      icon: (active: boolean) => (
+        <ListIcon Icon={() => <DiscussIcon secondary={active} />} />
+      ),
+      title: 'Best discussions',
+      path: '/discussed',
+      action: () => onNavTabClick?.('discussed'),
+    });
+  }
```
DragosIorgulescu, line 85: lacking a bit of context here, seems we are marking the `/discussed` page as new? I don't see any modifications to the page though, or is this just to register the action?

packages/shared/src/hooks/useFeedLayout.ts
```diff
@@ -26,18 +29,24 @@ export const useFeedLayout = ({
...
-  const shouldUseMobileFeedLayout = !feedRelated
-    ? isMobile
-    : isMobile && isIncludedFeed;
+
+  const shouldUseCommentFeedLayout =
+    hasCommentFeed && feedName === SharedFeedPage.Discussed;
+  const isMobileAndIncludedFeed = isMobile && isIncludedFeed;
+  const shouldUseMobileFeedLayout = feedRelated
+    ? isMobileAndIncludedFeed
+    : isMobile;
```
DragosIorgulescu, line 46: ah, I see now why the `MainFeedLayout` was modified, you're switching to the comments feed on the `/discussed` page if the feature flag is on 👌

DragosIorgulescu (approved):
Nice, not seeing any issues here 👌

packages/shared/src/hooks/useFeedLayout.ts
```diff
@@ -26,18 +29,24 @@ export const useFeedLayout = ({
   feedName: feedNameProp,
   feedRelated = true,
 }: UseFeedLayoutProps = {}): UseFeedLayout => {
+  const hasCommentFeed = useFeature(feature.commentFeed);
```
capJavert, line 32: I think this `useFeature` should be replace with `withExperiment` on `CommentFeed` and then only allocate if users is on `discussions` page.

capJavert: The hook should only return true if user is on discussions. This will prevent user being allocated on any other place where `useFeedLayout` is used.

capJavert (commented):
Good work 👌 , approved API so we can test on preview.

packages/shared/src/components/MainFeedLayout.tsx
rebelchris: It also replaces `Most discussed feed` for users in the experiment basically.

packages/shared/src/components/sidebar/DiscoverSection.tsx
rebelchris: Correct yes, so the change is the one above.
If you're in the experiment you'll see `CommentFeed` (feed of comments) else you'll see `Feed` with most commented posts.

DragosIorgulescu: nevermind, reading the PR description more carefully, I understand why this is here now 👍

packages/shared/src/hooks/useFeedLayout.ts
rebelchris: Hmm not sure, as it's also a sidebar condition?
Allocation would happen on mainLayout pretty much guaranteed anyway.

capJavert: Yeah, since it is sidebar it is less relevant but I think cleaner vs adding dependencies to hook that can technically be used anywhere.

rebelchris: Yeah agreed.
Think technically probably best to just have separate hook even.
Let's see how this experiment evolves

capJavert: Solved it on slack. There were some other dependencies.

capJavert (approved):

packages/shared/src/components/sidebar/DiscoverSection.tsx
```diff
@@ -15,6 +19,9 @@ export function DiscoverSection({
   enableSearch,
   ...defaultRenderSectionProps
 }: DiscoverSectionProps): ReactElement {
+  const hasCommentFeed = useFeature(feature.commentFeed);
+  const { checkHasCompleted, completeAction } = useActions();
+  const hasCompletedCommentFeed = checkHasCompleted(ActionType.CommentFeed);
```
capJavert, line 24: @rebelchris you are missing `isActionsFetched, now the "New" label flashes on each page load since by default `checkHasCompleted` returns `false` until actions are fetched.

PR #2849, opened 2024-03-22
feat: update sidebar on resolutions below tablet

DragosIorgulescu:
## Notes

- this is the first part of the changes associated with ticket MI-130
  - the rest of the changes will be broken down in additional PRs
- introducing a feature flag for easier control over the release of these changes

## Changes

https://github.com/dailydotdev/apps/assets/1225100/04021c60-6825-4f44-a455-9ebc19d96bfc

## Events

N / A

- [x] MobileL (420px)
- [x] Tablet (656px)
- [x] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android


packages/shared/src/components/buttons/Button.tsx
```diff
@@ -95,13 +95,16 @@ function ButtonComponent<TagName extends AllowedTags>(
...
         !color && VariantToClassName[variant],
         VariantColorToClassName[variant]?.[color],
         className,
       )}
     >
       {icon &&
-        iconPosition === ButtonIconPosition.Left &&
+        new RegExp(
+          `(${ButtonIconPosition.Left}|${ButtonIconPosition.Top})`,
+        ).test(iconPosition) &&
```
rebelchris, line 107: Why do we need this exactly?
Do you mean something like `[Left,Top].includes(iconPosition)`?

packages/shared/src/components/sidebar/Sidebar.tsx
```diff
@@ -75,6 +92,77 @@ export default function Sidebar({
     return <></>;
   }
 
+  if (useNewMobileLayout && !isLaptop && !isMobile) {
+    const buttonProps: ButtonProps<'a' | 'button'> = {
+      variant: ButtonVariant.Tertiary,
+      style: { width: '100%' },
```
rebelchris, line 98: Why don't we just use tailwind?

packages/shared/src/hooks/useMobileUxExperiment.ts
```diff
@@ -0,0 +1,17 @@
...
+import { feature } from '../lib/featureManagement';
+import { MobileUXLayout } from '../lib/featureValues';
+import { useViewSize, ViewSize } from './useViewSize';
+
+interface UseStreakExperiment {
+  useNewMobileLayout: boolean;
+}
+
+export const useMobileUxExperiment = (): UseStreakExperiment => {
+  const layout = useFeature(feature.mobileUxLayout);
```
rebelchris, line 11: Not sure the evaluation is ok here.
This will already assign people to the experiment even if they not mobile.

packages/shared/src/components/buttons/Button.tsx
DragosIorgulescu: I think regex is faster than `includes`, it also avoid having to create a new array -- this was just me being a little fancy 😅

packages/shared/src/components/sidebar/Sidebar.tsx
DragosIorgulescu: copy/pasta from here: https://github.com/dailydotdev/apps/blob/34863591dc2dcb5066107ea19957e5fe98d9a885/packages/webapp/components/FooterNavBar.tsx#L106-L110

I'll fix in both places

packages/shared/src/hooks/useMobileUxExperiment.ts
DragosIorgulescu: That's why there is also a check below for tablet view size.

Also, this is just for us to quickly rollback this change in case we need to, mobile UX will not actually be an experiment. See this discussion: https://dailydotdev.slack.com/archives/C0650M2KX8E/p1710934118481569

rebelchris: Ah completely missed this.
Will add the flag :)

packages/shared/src/components/layout/MainLayoutHeader.tsx
```diff
@@ -150,6 +153,10 @@ function MainLayoutHeader({
     );
   };
 
+  if ((isMobile || isTablet) && useNewMobileLayout) {
```
nensidosari, line 156: I think instead of `isMobile || isTablet` we can use !isLaptop

packages/shared/src/components/sidebar/Sidebar.tsx
```diff
@@ -75,6 +93,79 @@ export default function Sidebar({
     return <></>;
   }
 
+  if (useNewMobileLayout && !isLaptop && !isMobile) {
```
nensidosari, line 96: Would this mean we want this for mobile? `!isLaptop && !isMobile` Can we check just isTablet instead?

nensidosari: Also I see the useNewMobileLayout already excludes isLaptop, meaning we don't have to check for it is we have useNewMobileLayout

packages/shared/src/components/layout/MainLayoutHeader.tsx
nensidosari: Also I see the useNewMobileLayout already excludes isLaptop, and in this case looks like we don't need this part of the condition at all `(isMobile || isTablet)`

nensidosari (commented):
Looks good overall, few comments.

Also something I noticed in the video in the description, in one point the sidebar overrides the feed settings button

Screenshot 2024-03-25 at 11 30 51 AM

DragosIorgulescu:
> Also something I noticed in the video in the description, in one point the sidebar overrides the feed settings button

I tackled this in the PR for the layout changes: https://github.com/dailydotdev/apps/pull/2856 🙏

packages/shared/src/components/sidebar/Sidebar.tsx
DragosIorgulescu: > Would this mean we want this for mobile? !isLaptop && !isMobile Can we check just isTablet instead?

We only want this on resolutions above mobile but below tablet, passing just `isTablet` will make it show up on laptop resolutions as well

packages/shared/src/components/buttons/Button.tsx
capJavert: Inline regex can be really slow depending on browser. For example V8 has some good optimizations but Safari (iOS) and Firefox are much slower. If we are afraid of inline arrays we can just define const of `[Left,Top]` above component.

capJavert: Quick bench https://www.measurethat.net/Benchmarks/ShowResult/514328

rebelchris: I also think personally just doing includes on array is a bit more readable.
(this specific regex is ok, but it can get very complicated ofcourse) 😅

packages/shared/src/components/sidebar/Sidebar.tsx
nensidosari: I really don't like that we're using only min values...in that case I think we need to name variables like isTabletAndAbove 😢 else it's confusing.

Anyways in this case u can check for isTablet only since u're already checking for `!isLaptop` in `useNewMobileLayout`

packages/shared/src/components/buttons/Button.tsx
DragosIorgulescu: oh, nevermind then 👀

packages/shared/src/components/sidebar/Sidebar.tsx
DragosIorgulescu: > I really don't like that we're using only min values...in that case I think we need to name variables like isTabletAndAbove 😢 else it's confusing.

I agree, at the same time other options can become verbose. I think that as long as we know how these function, it's OK?

packages/shared/src/components/sidebar/Sidebar.tsx
```diff
@@ -75,6 +92,79 @@ export default function Sidebar({
...
+    const buttonProps: ButtonProps<'a' | 'button'> = {
+      variant: ButtonVariant.Tertiary,
+      size: ButtonSize.Large,
+      className: 'w-full',
+    };
+
+    return (
+      <SidebarAside
+        data-testid="sidebar-aside"
+        className={classNames('max-w-16 basis-10 items-center gap-4')}
```
rebelchris, line 105: Just checking should this not be just a static width? (instead of using max-w)

rebelchris (approved):
Alright nothing blocking from my side.

nensidosari (approved):
Looks good

packages/shared/src/components/layout/MainLayoutHeader.tsx
```diff
@@ -150,6 +152,10 @@ function MainLayoutHeader({
     );
   };
 
+  if (isNewMobileLayout) {
```
sshanzel, line 155: Tagging just to make sure this was intended.

packages/shared/src/components/sidebar/Sidebar.tsx
sshanzel: Also, not sure `max-w-` works with numbers ootb.

sshanzel (approved):
Have nothing to add from what was already pointed out. Great work! 🥳

packages/shared/src/components/buttons/Button.tsx
```diff
@@ -95,13 +95,16 @@ function ButtonComponent<TagName extends AllowedTags>(
...
         iconOnly ? IconOnlySizeToClassName[size] : SizeToClassName[size],
+        iconPosition === ButtonIconPosition.Top && `flex-col`,
         !color && VariantToClassName[variant],
         VariantColorToClassName[variant]?.[color],
         className,
       )}
     >
       {icon &&
-        iconPosition === ButtonIconPosition.Left &&
+        [ButtonIconPosition.Left, ButtonIconPosition.Top].includes(
```
capJavert, line 105: If you don't want to instantiate each time youcan pull array as const above component.

capJavert (approved):

DragosIorgulescu:
closing as we will be using https://github.com/dailydotdev/apps/pull/2855 for final review

PR #2853, opened 2024-03-23
fix: prevent flash of new

rebelchris:
- Since we don't wait for isActionsFetched it would always flash the new badge quickly.
- Decided to add this as raised [here](https://dailydotdev.slack.com/archives/C06LMML6A8N/p1711120684667839?thread_ts=1711120611.841409&cid=C06LMML6A8N). 
- Maybe we should actually add it to the default of `checkHasCompleted`? (not really sure though)

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done

### Preview domain
https://fix-new-flash.preview.app.daily.dev

### DIFF
--- a/packages/shared/src/components/sidebar/DiscoverSection.tsx
+++ b/packages/shared/src/components/sidebar/DiscoverSection.tsx
@@ -20,8 +20,9 @@ export function DiscoverSection({
   ...defaultRenderSectionProps
 }: DiscoverSectionProps): ReactElement {
   const hasCommentFeed = useFeature(feature.commentFeed);
-  const { checkHasCompleted, completeAction } = useActions();
-  const hasCompletedCommentFeed = checkHasCompleted(ActionType.CommentFeed);
+  const { checkHasCompleted, completeAction, isActionsFetched } = useActions();
+  const hasCompletedCommentFeed =
+    !isActionsFetched || checkHasCompleted(ActionType.CommentFeed);
   const discoverMenuItems: SidebarMenuItem[] = [
     {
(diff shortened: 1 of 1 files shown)


packages/shared/src/components/sidebar/DiscoverSection.tsx
```diff
@@ -20,8 +20,9 @@ export function DiscoverSection({
   ...defaultRenderSectionProps
 }: DiscoverSectionProps): ReactElement {
   const hasCommentFeed = useFeature(feature.commentFeed);
-  const { checkHasCompleted, completeAction } = useActions();
-  const hasCompletedCommentFeed = checkHasCompleted(ActionType.CommentFeed);
+  const { checkHasCompleted, completeAction, isActionsFetched } = useActions();
+  const hasCompletedCommentFeed =
```
rebelchris, line 24: This now basically means it flashes the other way around (from not -> showing) but you don't notice really as it's on page load.

capJavert: I think issue is some conditions are meant to be true by default and some false, we can maybe add default value for `checkHasCompleted` so that in cases where we expect `true` by default we can say `checkHasCompleted(action, true)`?

rebelchris: Yeah probably, lets log it for future 👍

capJavert (approved):

PR #2862, opened 2024-03-26
fix: go back navigation

sshanzel:
## Changes
- The back button should purely act as the history back button.
- Thread: https://dailydotdev.slack.com/archives/C0650M2KX8E/p1711444583148789?thread_ts=1711372461.259189&cid=C0650M2KX8E

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done

### Preview domain
https://fix-navigation.preview.app.daily.dev

### DIFF
--- a/packages/shared/src/components/post/PostContentHeaderMobile.tsx
+++ b/packages/shared/src/components/post/PostContentHeaderMobile.tsx
@@ -12,10 +12,10 @@ export function PostContentHeaderMobile({
 }: Pick<PostNavigationProps, 'onReadArticle' | 'post'>): ReactElement {
   const router = useRouter();
   const canGoBack =
-    globalThis?.history &&
-    !!globalThis.history.length &&
+    !!globalThis?.history?.length &&
+    globalThis?.history?.state &&
     isNullOrUndefined(globalThis.history.state.options.shallow) &&
-    !globalThis.document.referrer; // empty referrer means you are from the same site
+    !globalThis?.document?.referrer; // empty referrer means you are from the same site
 
   return (
     <span className="-mx-4 flex flex-row items-center border-b border-border-subtlest-tertiary px-4 py-2 tablet:hidden">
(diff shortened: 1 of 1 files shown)


packages/shared/src/components/post/PostContentHeaderMobile.tsx
```diff
@@ -4,46 +4,27 @@ import { Button, ButtonSize, ButtonVariant } from '../buttons/Button';
...
         variant={ButtonVariant.Tertiary}
-        onClick={onBackClick}
-        className={
-          !canGoBack &&
-          !routedFromSquad &&
-          post.source.type !== SourceType.Squad &&
-          'invisible'
-        }
+        onClick={router.back}
+        className={!canGoBack && 'invisible'}
```
nensidosari, line 27: Should we just do hidden rather then invisible?

packages/shared/src/components/post/PostContentHeaderMobile.tsx
```diff
@@ -4,46 +4,27 @@ import { Button, ButtonSize, ButtonVariant } from '../buttons/Button';
...
-
-    if (post.source.type === SourceType.Squad) {
-      return router.push(post.source.permalink);
-    }
-
-    return null;
-  };
+  const canGoBack =
+    globalThis?.history &&
+    !!globalThis.history.length &&
```
nensidosari, line 16: Why didn't u reuse `!!globalThis?.window?.history?.length` for these?

nensidosari (approved):
Looks good, non blocking comments

packages/shared/src/components/post/PostContentHeaderMobile.tsx
sshanzel: We want to keep the height consistent, so we used invisible. We can use either of the two, just saw some benefits here.

packages/shared/src/components/post/PostContentHeaderMobile.tsx
sshanzel: The condition `isNullOrUndefined(globalThis.history.state.options.shallow)` would be true if the `globalThis?.history?.state` is used.

sshanzel: Changed the condition to make it easier to understand.

PR #2864, opened 2024-03-26
fix: add condition render hook and forced tag selection

rebelchris:
- Added a new hook to conditionally evaluate experiments (to avoid allocaiton)
- Re-added onboarding 4.5 as `forced_tag_selection`

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done

### Preview domain
https://as-209-forced-tag-selection-exp.preview.app.daily.dev


packages/shared/src/hooks/useConditionalFeature.ts
```diff
@@ -0,0 +1,37 @@
...
+      return growthbook
+        ? growthbook.getFeatureValue(feature.id, feature.defaultValue)
+        : (feature.defaultValue as WidenPrimitives<T>);
+    },
+    {
+      enabled: shouldEvaluate,
+      ...disabledRefetch,
+    },
+  );
+
```
capJavert, line 35: We can probably return it from `initialData` instead of this `??` check?

packages/shared/src/hooks/useConditionalFeature.ts
```diff
@@ -0,0 +1,37 @@
...
+import { disabledRefetch } from '../lib/func';
+
+export const useConditionalFeature = <T extends JSONValue>({
+  feature,
+  shouldEvaluate,
+}: {
+  feature: Feature<T>;
+  shouldEvaluate: boolean;
+}): WidenPrimitives<T> => {
+  const growthbook = useGrowthBook();
```
capJavert, line 15: I think it should come from `useGrowthBookContext`, not sure `useGrowthBook` gets the same instance?

packages/shared/src/hooks/useConditionalFeature.ts
```diff
@@ -0,0 +1,37 @@
...
+  const queryKey = generateQueryKey(
+    RequestKey.Feature,
+    null,
+    feature.id,
+    shouldEvaluate,
+  );
```
capJavert, line 21: Should it not be per user eg. for logouts? Also why is `shouldEvaluate`  needed, it is not used in queryFn so it just makes additional query key while disabled from what I can see.

capJavert (approved):
Looking good to me 👌

packages/shared/src/hooks/useConditionalFeature.ts
```diff
@@ -0,0 +1,36 @@
...
+
+  const { data: featureValue } = useQuery(
+    queryKey,
+    () => {
+      return growthbook
+        ? growthbook.getFeatureValue(feature.id, feature.defaultValue)
+        : (feature.defaultValue as WidenPrimitives<T>);
+    },
+    {
+      initialData: feature.defaultValue as WidenPrimitives<T>,
```
idoshamun, line 29: Shouldn't this be optional? Some cases would want to wait for the evaluation I guess

packages/shared/src/hooks/useConditionalFeature.ts
```diff
@@ -0,0 +1,36 @@
...
+  shouldEvaluate: boolean;
+}): WidenPrimitives<T> => {
+  const { user } = useContext(AuthContext);
+  const { growthbook } = useGrowthBookContext();
+  const queryKey = generateQueryKey(RequestKey.Feature, user, feature.id);
+
+  const { data: featureValue } = useQuery(
+    queryKey,
+    () => {
+      return growthbook
```
idoshamun, line 24: In case the gb object updates, shouldn't we retrigger this?

nensidosari (approved):
Looks good to me, glad to have an option to conditionally enroll

nensidosari:
I'm thinking if there are going to be use cases where still this wouldn't be enough, but can't find any at the moment, so we should be good to go with this and then see if there is still any limitations.

packages/shared/src/hooks/useConditionalFeature.ts
rebelchris: I guess yes, not sure the current useFeatureValue does it.
But can add it in querykey somehow

packages/shared/src/hooks/useConditionalFeature.spec.tsx
```diff
@@ -0,0 +1,61 @@
...
+  it('should return feature value if evaluation is true', async () => {
+    const { result } = renderHook(
+      () =>
+        useConditionalFeature({ feature: testFeature, shouldEvaluate: true }),
+      {
+        wrapper: Wrapper,
+      },
+    );
+
+    expect(result.current).toBe('default_value');
```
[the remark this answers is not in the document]
rebelchris: Yeah kind of missing the option to change enrollment in tests atm 🙈

packages/shared/src/hooks/useConditionalFeature.spec.tsx
```diff
@@ -0,0 +1,87 @@
...
+
+  it('should return feature value if evaluation is true', async () => {
+    const { result } = renderHook(
+      () =>
+        useConditionalFeature({
+          feature: testFeature,
+          shouldEvaluate: true,
+        }),
+      {
+        wrapper: createWrapper({ value: 'new_value' }),
```
rebelchris, line 80: @nensidosari added different test evaluation now :)

packages/shared/src/hooks/useConditionalFeature.ts
capJavert: User is defaultValue (control) until proven different right? But loading state makes sense. Maybe we can also add an argument to hook for defaultValue so we can fallback it when needed?

packages/shared/src/hooks/useConditionalFeature.ts
```diff
@@ -0,0 +1,44 @@
...
+  feature: Feature<T>;
+  shouldEvaluate: boolean;
+}): UseConditionalFeature<T> => {
+  const { user } = useContext(AuthContext);
+  const { growthbook } = useGrowthBookContext();
+  const queryKey = generateQueryKey(
+    RequestKey.Feature,
+    user,
+    feature.id,
+    growthbook,
```
capJavert, line 27: growthbook is a complex object I am not sure it would even do anything in query key since query key is stringified in react-query. Why is this needed?

capJavert: image

this is how query key looks 😆

rebelchris: Basically for [this](https://github.com/dailydotdev/apps/pull/2864#discussion_r1539387861), but not sure what alternatives we have?
There's not really a clear update/retrigger exposure I think.

packages/shared/src/hooks/useConditionalFeature.ts
rebelchris: I'm actually not sure we need to retrigger?
Wouldn't it just reflect a next evaluation on it's own?

I think pretty much how everything works now (and why sometimes you need to hard refresh on apps)

idoshamun: ok so let's leave it as it is. but just double check that it works this way

packages/shared/src/hooks/useConditionalFeature.ts
capJavert: Ah, our growthbook instance never changes currently during single page visit, and we don't add new features once initiated https://github.com/dailydotdev/apps/blob/2f31655db1810b3122ca7ca20c83b0350ca4aebc/packages/shared/src/components/GrowthBookProvider.tsx#L63-L87 so it should be good to go without looking at change.

That said if it was to ever change and we change logic in `GrowthBookProvider` better way would be to just invalidate `queryClient.invalidateQueries(RequestKey.Feature)` I think.

packages/shared/src/hooks/useConditionalFeature.ts
idoshamun: ok, again we just have to watch out from flickering and layout shifts

packages/shared/src/hooks/useConditionalFeature.ts
idoshamun: Good idea!

capJavert: Thats we have "you need to refresh" for experiment to become active thing on preview as well.

rebelchris: Moved we'll need to check on preview as locally doesn't really work to override.

idoshamun (approved):

rebelchris:
@capJavert // @idoshamun let me know what you think.
Looks ok to me?

https://github.com/dailydotdev/apps/assets/554874/121dd715-9871-4439-ac32-da962623e76e

idoshamun:
@rebelchris perfect!

capJavert (approved):
Looking good, if this works we should also add it to experimentation docs in storybook at some point since it is good utility 👌

rebelchris:
@capJavert // @idoshamun See my test failed with not having the GB eval.
There's an edge case where it would be too fast, (gb not loaded)
So added the features ready state as enabled state + in the key.
Let me know if that works?

PR #2865, opened 2024-03-26
ci: optimize ci time with some perks 😈

idoshamun:
image

* I managed to reduce CI time by 73% to a ridicules time of 2.5min! ⚡ 
* Thanks to @rebelchris I learned about the ["Rerun failed tests"](https://circleci.com/docs/rerun-failed-tests/) feature. So you can rerun only failed tests now if needed 🤩 
* Parallel support in case we get more tests per project, we can easily spool up more workers per project

The essence of the change is that each package's CI now runs independently. In addition since the shared package lint takes time, I moved it to its own process to run it in parallel with the tests.

It's about time we get a better feedback loop. Enjoy!

eslint timing breakdown:
image

### Preview domain
https://split-tests.preview.app.daily.dev


.circleci/config.yml
```diff
@@ -27,45 +26,113 @@ jobs:
             - ./packages/shared/node_modules
             - ./packages/webapp/node_modules
             - ./packages/extension/node_modules
+  build:
+    docker:
+      - image: cimg/node:18.16.0
+    resource_class: medium+
```
rebelchris, line 32: Did you try with a smaller one now?

rebelchris: Actually seeing they fully split, maybe even small can do?

rebelchris (approved):
Nothing blocking, but guess we can go lower resource right?

packages/eslint-config/index.js
```diff
@@ -33,6 +33,7 @@ module.exports = {
   },
   rules: {
     curly: ['error', 'all'],
+    'import/no-cycle': 0,
```
nensidosari, line 36: In the end we didn't want to do any pre-push check for this? Anyways we can think about it and do later

nensidosari (approved):
Looks great

packages/eslint-config/index.js
omBratteng: I think it could be worth having this enabled, so that we get this lint rule during development.
But instead disabling it during CI (might be possible by checking if env `CI=true`)

capJavert: I am not a fan of spliting and different lint for dev/ci since whatever you don't do in CI is subject to local dev env which leads to inconsistencies. Per this issue here https://github.com/import-js/eslint-plugin-import/issues/2348 they are recommending `{ ignoreExternal: true }`, since we really want to check only our own code. Worth a try if that is a culprit. https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-cycle.md#ignoreexternal

capJavert: Also might be worth a try to play with `maxDepth`, since I think it is Infinity by default.

idoshamun: Yes I agree, that's what I wanted to try today before merging

.circleci/config.yml
idoshamun: I did try to optimize resources but i also changed couple things after that, so I can try again.
Specifically about build it fails with anything less than medium+. But maybe I can try again other flows

packages/eslint-config/index.js
idoshamun: So adding this settings adds something like 1 minute to the linting process, do we still want it? tagging also @rebelchris

idoshamun: With large instance I managed to keep the same timing. So we'll keep this new setting

rebelchris: Let's see if that one can we large and the others maybe less?
(that should even out cost I guess)

capJavert (approved):
Awesome! 🚀

packages/eslint-config/index.js
idoshamun: already fixed the other ones 💪

PR #2867, opened 2024-03-27
fix: sidebar navigation on mobile

sshanzel:
## Changes
- We should always open the new sidebar on mobile.

https://github.com/dailydotdev/apps/assets/13744167/fb30eb79-6c4e-426c-8963-28d76b0069b4

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

MI-241 #done

### Preview domain
https://mi-241.preview.app.daily.dev


packages/webapp/components/layouts/AccountLayout/index.tsx
```diff
@@ -44,10 +66,18 @@ export default function AccountLayout({
...
-          basePath="account"
-        />
+        {isMobile ? (
+          <ProfileSettingsMenu
+            isOpen={isOpen}
+            logout={logout}
+            onClose={closeSideNav}
+          />
+        ) : (
+          <SidebarNav
```
rebelchris, line 76: Ah was hoping we could remove it or uniform it.
Maybe check with design for a follow up task to uniform it? (non urgent)

packages/shared/src/components/drawers/NavDrawer.tsx
```diff
@@ -64,7 +64,16 @@ export function NavDrawer({
...
+          <NavDrawerItem
+            key={item.label}
+            {...item}
+            onClick={(e) => {
+              if (item.onClick) {
+                item.onClick(e);
+              }
+              ref.current?.onClose();
+            }}
+          />
```
rebelchris, line 76: Why is this needed?
Can't we uniform it inside the NavDrawerItem maybe?

packages/webapp/components/layouts/AccountLayout/index.tsx
```diff
@@ -1,22 +1,44 @@
...
-import SidebarNav from './SidebarNav';
 import { getTemplatedTitle } from '../utils';
+import SidebarNav from './SidebarNav';
 
 export interface AccountLayoutProps {
   profile: PublicProfile;
   children?: ReactNode;
 }
 
+export const navigationKey = generateQueryKey(
```
rebelchris, line 24: Not urgent, but seeing it used in 2 places now would maybe make for a good hook?
I'm missing a bit of context here, why do we need react query for this?
Wouldn't simply state work?

rebelchris (commented):
Minor comments, let me know what you think

packages/webapp/components/layouts/AccountLayout/index.tsx
sshanzel: Yeah, at first I actually deleted this completely thinking we won't need it anymore, but then on laptop/tablet, they have a completely different design/components.

> Maybe check with design for a follow up task to uniform it? (non urgent)

Sure, will send a message.

packages/shared/src/components/drawers/NavDrawer.tsx
sshanzel: The ref is not found in the item itself. Unless we create a context or something to access this close function, we will have to make the call in this function.

packages/webapp/components/layouts/AccountLayout/index.tsx
sshanzel: Because the back button is found not on where we initialized the sidebar.

packages/shared/src/components/drawers/NavDrawer.tsx
rebelchris: Hmm couldn't it just accept a ref as prop?
(anyway not breaking, more of a thought)

packages/webapp/components/layouts/AccountLayout/index.tsx
rebelchris: Hmm seems like usePersistentContext almost?
Anyway up to you but dedicated hook might be needed just to maintain it better? (also optional)

sshanzel: Yeah, could've used the persistent context, though we don't need the extra usage of the idb storage. Maybe we should introduce a hook without it called `useQueryState`.

rebelchris: Hmm ah right, yes think in current flows we indeed just use dedicated hooks right to solve this.
We have quite a few that do something like this.

sshanzel: Let me know what you think of this https://github.com/dailydotdev/apps/pull/2867/commits/588741f91b43500a62ddb819c573be0f7b0707a1

packages/shared/src/components/drawers/NavDrawer.tsx
sshanzel: Yeah, we can do that. Clicking a nav item should automatically render a close action.

packages/shared/src/hooks/utils/useQueryState.ts
```diff
@@ -0,0 +1,32 @@
...
+type UseQueryState<T> = [
+  T,
+  (value: T) => void,
+  Omit<UseQueryResult<T>, 'data'>,
+];
+
+export const useQueryState = <T>(
+  key: QueryKey,
+  defaultValue: T,
+): UseQueryState<T> => {
```
rebelchris, line 19: Let's maybe make it object based (just in case we add more later)

rebelchris (approved):
Looks good to me.
Good hook to have in the belt.

@capJavert / @nensidosari thoughts?

packages/shared/src/hooks/utils/useQueryState.ts
sshanzel: Yeah, this has no restriction, you can pass an object, boolean, string, or anything. Also we did not set a specific type.

rebelchris: I mean instead of plain props make it object props :) 

```
export const useQueryState = <T>({
  key
  defaultValue
}: PropsDef): UseQueryState<T> => {
```

sshanzel: Ahh, cool. You meant the props. I just thought of making it very similar to `useState`, but yeah, we can do that too.

sshanzel: I was just thinking we can have the optional parameters on the third when we need it. But yeah, updated the props now.

capJavert (approved):
I like the hook 👌

PR #2878, opened 2024-03-28
refactor: move feed settings from MainFeedLayout to OnboardingFeedHeader

omBratteng:
# Changes

Moves `useFeedSettings` from `MainFeedLayout` to `OnboardingFeedHeader`, as it does a query `FeedPreferences` query in every feed right now when it is not needed.


rebelchris (approved):
Tested, and seems to work fine 
Why did we have it like this before, could have done like this right?

nensidosari (approved):

PR #2882, opened 2024-03-28
fix: second evaluation would have cache

rebelchris:
- We had a small issue with onboarding if you where in the experiment, then complete onboarding the `shouldEvaluate` becomes false, but it would not retrigger unless you did a hard refresh on the page.
- This introduces a second trigger to always return false when already evaluated and this value changes

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done

### Preview domain
https://fix-second-eval-fix.preview.app.daily.dev

### DIFF
--- a/packages/shared/src/hooks/useConditionalFeature.ts
+++ b/packages/shared/src/hooks/useConditionalFeature.ts
@@ -35,7 +35,7 @@ export const useConditionalFeature = <T extends JSONValue>({
     queryKey,
     () => {
       if (!shouldEvaluate) {
-        return false as WidenPrimitives<T>;
+        return feature.defaultValue as WidenPrimitives<T>;
       }
       return growthbook
         ? growthbook.getFeatureValue(feature.id, feature.defaultValue)
--- a/packages/shared/src/hooks/useConditionalFeature.ts
+++ b/packages/shared/src/hooks/useConditionalFeature.ts
@@ -24,11 +24,19 @@ export const useConditionalFeature = <T extends JSONValue>({
   const { user } = useContext(AuthContext);
   const { growthbook } = useGrowthBookContext();
   const { ready } = useFeaturesReadyContext();
-  const queryKey = generateQueryKey(RequestKey.Feature, user, feature.id);
+  const queryKey = generateQueryKey(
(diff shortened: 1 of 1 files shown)


packages/shared/src/hooks/useConditionalFeature.ts
```diff
@@ -24,11 +24,19 @@ export const useConditionalFeature = <T extends JSONValue>({
...
+    feature.id,
+    shouldEvaluate,
+  );
 
   const { data: featureValue, isLoading } = useQuery(
     queryKey,
     () => {
+      if (!shouldEvaluate) {
+        return false as WidenPrimitives<T>;
+      }
```
capJavert, line 39: What about features that are not bool?

rebelchris: I can change it to the default value, would that make more sense I guess?

capJavert: Yes, just to support other feature flags in this hook.

capJavert (approved):

packages/shared/src/hooks/useConditionalFeature.ts
rebelchris: @capJavert Although, at some point we might change default to true, and then you get this experiment forever.
But guess once we hit that we can add something like `valueWhenNotEvaluated` or something.

capJavert: Hmm, don't know at top of my head, but sounds complicated and yes something to evaluate for future usage of the hook.

capJavert: Basically just returning something that GB does not return feels wrong right? We should hide the added UI in regards to something else and feature value should always be true if user is part of experiment.

rebelchris: Yeah correct, for this flow it's fine, but can imagine there might be something where it won't work ilke this.
There we could potentially use the initial evaluation for rendering determination.

PR #2886, opened 2024-03-29
fix: mobile ux layout fixes

DragosIorgulescu:
## Changes

Fixes the following:
- [x] The notification button is not having the correct background when active
- [x] Feed actions are not in a single row (sorting enabled)
- [x] Make the tabs header scrollable to access the items in the far right
- [x] Unnecessary margin on the right of the notification bell
- [x] Reduce top spacing on feeds

---

Screenshot 2024-03-29 at 10 09 05

---

Screenshot 2024-03-29 at 10 08 55

---

Screenshot 2024-03-29 at 10 02 56

---

Screenshot 2024-03-29 at 10 15 06

## Events

N / A

- [x] MobileL (420px)
- [x] Tablet (656px)
- [x] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

MI-252

### Preview domain
https://mi-252-mobile-layout-fixes.preview.app.daily.dev

### DIFF
--- a/packages/shared/src/components/feeds/FeedContainer.tsx
+++ b/packages/shared/src/components/feeds/FeedContainer.tsx
@@ -173,7 +173,7 @@ export const FeedContainer = ({
                 className={classNames(
                   'flex flex-col rounded-16 border border-theme-divider-tertiary tablet:mt-6',
                   isSearch && 'mt-6',
-                  isNewMobileLayout && 'border-0 mobileL:mt-2 tablet:mt-2',
+                  isNewMobileLayout && '!mt-2 border-0',
                 )}
               >
                 <span className="flex w-full flex-row items-center justify-between px-6 py-4">
--- a/packages/shared/src/components/feeds/FeedContainer.tsx
+++ b/packages/shared/src/components/feeds/FeedContainer.tsx
@@ -173,6 +173,7 @@ export const FeedContainer = ({
                 className={classNames(
                   'flex flex-col rounded-16 border border-theme-divider-tertiary tablet:mt-6',
                   isSearch && 'mt-6',
+                  isNewMobileLayout && 'border-0 mobileL:mt-2 tablet:mt-2',
                 )}
               >
                 <span className="flex w-full flex-row items-center justify-between px-6 py-4">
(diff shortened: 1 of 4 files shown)


nensidosari:
🙏 add a description on what points does this fix so we can test 🙏

DragosIorgulescu:
> 🙏 add a description on what points does this fix so we can test 🙏

@nensidosari good point, updated the description

nensidosari:
On this I see one issue:
`Make the tabs header scrollable to access the items in the far right`

there is a div wrapping the notification icon which if u try to start scrolling on top of that it won't allow it, so messes with the scroll. While if u try to scroll outside that div, so more on the left it works ok.

nensidosari:
Verified:

✅ The notification button is not having the correct background when active
✅ Feed actions are not in a single row (sorting enabled)
✅ Unnecessary margin on the right of the notification bell

nensidosari:
What does this mean? reduce top spacing on feeds

Not sure if related, but I see this inconsistency in top spacing for dropdown - once is bigger, then gets smaller, then gets bigger again:

https://github.com/dailydotdev/apps/assets/36197304/b838710a-a111-48e8-b4a1-fcf99b7c2949

DragosIorgulescu:
> Not sure if related, but I see this inconsistency in top spacing for dropdown - once is bigger, then gets smaller, then gets bigger again:

On tablet it is OK I think, but the reason it goes back again on smaller resolutions is because of the fact that we have `420px` as the minimum threshold set for mobile, not sure if something we should be changing in this PR, as it may affect other areas if we do: https://github.com/dailydotdev/apps/blob/c8458906c9b0b6fc9a64d37c13dacda14af42019/packages/shared/tailwind.config.ts#L153

DragosIorgulescu:
> there is a div wrapping the notification icon which if u try to start scrolling on top of that it won't allow it, so messes with the scroll. While if u try to scroll outside that div, so more on the left it works ok.

adjusted the wrapper setting the gradient a little bit to make the UX here better

nensidosari (approved):
Looks good now 🎉

PR #2891, opened 2024-04-01
fix: set right event_page for page view events on extension

idoshamun:
Page views on the extension were tracked with the wrong event page for a long time.

This is now fixed, hopefully 😬

### Preview domain
https://page-view-fix.preview.app.daily.dev

### DIFF
--- a/packages/extension/src/newtab/App.tsx
+++ b/packages/extension/src/newtab/App.tsx
@@ -62,7 +62,7 @@ function InternalApp(): ReactElement {
   useLazyModal();
   usePrompt();
   useWebVitals();
-  const { setCurrentPage } = useExtensionContext();
+  const { setCurrentPage, currentPage } = useExtensionContext();
   const { unreadCount } = useNotificationContext();
   const { closeLogin, shouldShowLogin, loginState } = useContext(AuthContext);
   const { contentScriptGranted } = useContentScriptStatus();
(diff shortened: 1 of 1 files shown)


DragosIorgulescu (approved):
👀

packages/extension/src/newtab/App.tsx
```diff
@@ -81,13 +81,12 @@ function InternalApp(): ReactElement {
...
       routeChangedCallbackRef.current();
     }
-  }, [isPageReady, routeChangedCallbackRef]);
+  }, [isPageReady, routeChangedCallbackRef, currentPage]);
 
   const { dismissToast } = useToastNotification();
 
   const onPageChanged = (page: string): void => {
     setCurrentPage(page);
-    routeChangedCallbackRef.current();
```
DragosIorgulescu, line 90: will removing this break anything?

idoshamun: it has but i added it to other the use effect so it should compensate that.
the issue was that calling this function right where we mutate the state causes it to use an invalid state because of how react works. So it used the wrong page for the event, with the current implementation I call this function only once the state was applied

DragosIorgulescu: Gotcha, makes sense, thanks for clarifying 🙏

rebelchris (approved):

PR #2894, opened 2024-04-01
fix: adjust discussed feed styles to match new mobile UX

DragosIorgulescu:
### New UI

Screenshot 2024-04-01 at 14 15 40

### Previous UI

Screenshot 2024-04-02 at 12 24 32

---

### New UI

Screenshot 2024-04-01 at 14 14 12

### Previous UI

Screenshot 2024-04-02 at 12 24 53

## Events

N / A

- [x] MobileL (420px)
- [x] Tablet (656px)
- [x] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

MI-252

### DIFF
--- a/packages/shared/src/components/CommentFeed.tsx
+++ b/packages/shared/src/components/CommentFeed.tsx
@@ -103,10 +103,7 @@ export default function CommentFeed<T>({
         isFetchingNextPage={queryResult.isFetchingNextPage}
         canFetchMore={checkFetchMore(queryResult)}
         fetchNextPage={queryResult.fetchNextPage}
-        className={classNames(
-          'w-full',
-          isNewMobileLayout && isMainFeed && 'px-4',
-        )}
+        className="w-full"
       >
         {queryResult.data.pages.flatMap((page) =>
           page.page.edges.map(({ node }, index) => (
--- a/packages/shared/src/components/MainFeedLayout.tsx
+++ b/packages/shared/src/components/MainFeedLayout.tsx
@@ -117,7 +117,7 @@ const getQueryBasedOnLogin = (
 const DEFAULT_ALGORITHM_KEY = 'feed:algorithm';
 
 const commentClassName = {
-  container: 'rounded-none border-0 border-b border-x',
+  container: 'rounded-none border-0 border-b tablet:border-x',
   commentBox: {
     container: 'relative border-0 rounded-none',
   },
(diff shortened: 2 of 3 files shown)


packages/shared/src/components/CommentFeed.tsx
```diff
@@ -103,10 +103,7 @@ export default function CommentFeed<T>({
         isFetchingNextPage={queryResult.isFetchingNextPage}
         canFetchMore={checkFetchMore(queryResult)}
         fetchNextPage={queryResult.fetchNextPage}
-        className={classNames(
-          'w-full',
-          isNewMobileLayout && isMainFeed && 'px-4',
```
sshanzel, line 108: What should this look like in control values? Can you also attach screenshot of the old UI? Just to make sure everything still looks normal.

sshanzel (approved):
Looks great. Kindly verify that the control still looks normal, we should be good here :shipit:

sshanzel (changes requested):
Sorry for the late request, I think the labels on the navigation for tablet is a bit bigger than expected.

DragosIorgulescu:
> Sorry for the late request, I think the labels on the navigation for tablet is a bit bigger than expected.

@sshanzel not sure what you are referring to, could you elaborate? Which labels?

sshanzel:
@DragosIorgulescu hi, sorry for the lack of clarity. These ones. In the Figma they're at `typo-caption2`, after inspecting the element, I think they're on `typo-callout`.

Screenshot 2024-04-02 at 5 28 59 PM

DragosIorgulescu:
> @DragosIorgulescu hi, sorry for the lack of clarity. These ones. In the Figma they're at typo-caption2, after inspecting the element, I think they're on typo-callout.

@sshanzel thank you for the clarification 🙏  I updated this, however to me these look way tinier than designs now (even though I agree these are specified with `typo-caption2` there. See below

Screenshot 2024-04-02 at 13 14 25

This feels like we should use `typo-caption1` rather 🤔

sshanzel:
> This feels like we should use typo-caption1 rather

Yep, totally agree with this. Sometimes it really looks different from Figma, I think what you suggested it the sweet spot.

sshanzel (approved):
Feel free to change it to `typo-caption1` - got nothing more on this, looks amazing! :shipit:

PR #2909, opened 2024-04-02
feat: use new vote mutation and add useVote hook

capJavert:
- adapting current vote hooks to new mutation
- new `useVote` hook
  - exposes `onMutate` for optimistic UI
  - uses new `vote` mutation with entity
- current useVotePost uses `useVote` under the hood (can be future refactored) and serves as implementation example for other entity hooks
- comment vote and downvote will be implemented in separate PR

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

AS-213 #done

### Preview domain
https://as-213-vote-mutation.preview.app.daily.dev


capJavert:
Have to test once over but should be ready for review.

sshanzel (approved):
Looks good to me! 🚀

packages/shared/src/hooks/vote/types.ts
```diff
@@ -1,81 +1,139 @@
...
   id: string;
+  entity: UserVoteEntity;
+};
+
+export type UseVote = {
+  upvote: (props: VoteProps) => Promise<void>;
+  downvote: (props: VoteProps) => Promise<void>;
+  cancelVote: (props: VoteProps) => Promise<void>;
+  toggleUpvote: (
+    props: ToggleVoteProps<PostEntity | ReadHistoryPost>,
```
rebelchris, line 35: Do we need the explicit entities here? (is the idea to add more as we add them?)

packages/shared/src/components/post/PostActions.tsx
```diff
@@ -52,7 +52,13 @@ export function PostActions({
       ? ({ id, vote }) => {
           const updatePost = updateCachedPagePost(feedQueryKey, queryClient);
 
-          return mutateVoteFeedPost({ id, vote, items, updatePost });
+          return mutateVoteFeedPost({
+            id,
+            vote,
+            items,
+            updatePost,
+            entity: UserVoteEntity.Post,
```
rebelchris, line 60: Do we need to send the entity here?
Does the votePost hook not handle it already

rebelchris (approved):
Nothing blocking :)

packages/shared/src/components/post/PostActions.tsx
capJavert: Its not part of the hook, it is the mutation handler. I adjusted coming type as alternative since it was not used really.

packages/shared/src/hooks/vote/types.ts
capJavert: Currently it is like this since we split post types. This one is specific to `useVotePost`. So at this point just legacy but can't clean it up fast. `ToggleVoteProps` can accept any future entity that supports voting.

nensidosari (approved):
Looks good to me as well

packages/shared/src/components/post/PostActions.tsx
rebelchris: Ah ok, hard to follow from the half visible PR interface :)

PR #2911, opened 2024-04-02
fix: making a comment through header actions

sshanzel:
## Changes
- The reference was missing. Tested locally.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done

### Preview domain
https://fix-making-comment.preview.app.daily.dev

### DIFF
--- a/packages/shared/src/contexts/WriteCommentContext.ts
+++ b/packages/shared/src/contexts/WriteCommentContext.ts
@@ -2,11 +2,11 @@ import { createContext, useContext } from 'react';
 import { UseMutateCommentResult } from '../hooks/post/useMutateComment';
 
 interface WriteCommentContextProp {
-  mutateCommentResult: UseMutateCommentResult;
+  mutateComment: UseMutateCommentResult;
 }
 
 export const WriteCommentContext = createContext<WriteCommentContextProp>({
-  mutateCommentResult: null,
+  mutateComment: null,
 });
 
 export const useWriteCommentContext = (): WriteCommentContextProp =>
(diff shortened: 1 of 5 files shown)


capJavert (approved):

sshanzel:
May look small and unnecessary at first, but it is because we haven't refactored the internals of the comment input.

packages/shared/src/hooks/post/useMutateComment.ts
```diff
@@ -152,10 +152,15 @@ export const useMutateComment = ({
     isLoading: isCommenting,
     isSuccess,
   } = useMutation<MutateCommentResult, unknown, SubmitComment>(
-    (variables) =>
-      requestMethod(graphqlUrl, mutation, variables, {
+    (variables) => {
+      if (isCommenting || isSuccess) {
+        return Promise.resolve(null);
+      }
```
capJavert, line 158: Why this change? It makes sense to add isCommenting but returning empty promise after one success feels like user would not be able to do two comments in a row?

packages/shared/src/contexts/WriteCommentContext.ts
```diff
@@ -0,0 +1,13 @@
+import { createContext, useContext } from 'react';
+import { UseMutateCommentResult } from '../hooks/post/useMutateComment';
+
+interface WriteCommentContextProp {
+  mutateCommentResult: UseMutateCommentResult;
```
capJavert, line 5: I think `mutateComment` is better and shorter?

packages/shared/src/hooks/post/useMutateComment.ts
sshanzel: They should be able to. This is a safeguard to avoid multiple same comments. However we can maybe add this in a different place to be safer.

capJavert (approved):
Looks good 👍

PR #2912, opened 2024-04-03
mobile ux: use a modal again to display comment inputs

kopancek:
## Changes

1. Reverts to rendering a modal with comment input instead. This is faster to render, since we don't need to go to the nextjs webserver for server side rendering and don't need to load any data - we already have everything we need.
2. Fixes the context menu drawer staying open after you click on Edit Comment button (or any other context menu action).
3. Reintroduces the reply preview when replying to a comment. See TODO below.

## Video

https://github.com/dailydotdev/apps/assets/9974711/891bb4fd-9636-43dc-9f7f-ce0034002bb7

### TODO

We still have a problem with mobile safari - it breaks the scrolling experience in the modal on the replies, in case the reply is long - the textarea is hidden and you cannot scroll to it. Ideas welcome.

https://github.com/dailydotdev/apps/assets/9974711/407fb73f-3180-47dd-ba0f-d47f04fe7df5

## Events

No new events

- [X] MobileL (420px)
- [X] Tablet (656px)
- [X] Laptop (1020px)

- [X] iOS (Chrome and Safari)
- [X] Android

MI-240 #done

### Preview domain
https://mi-258-comments-here-we-go-again.preview.app.daily.dev


packages/shared/src/components/comments/CommentInputOrModal.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+export default function CommentInputOrModal({
+  onClose,
+  className,
+  ...props
+}: CommentInputOrModalProps): ReactElement {
+  const isModal = !useViewSize(ViewSize.Laptop);
+
+  if (isModal) {
+    return <CommentModal {...props} isOpen onRequestClose={onClose} />;
```
kopancek, line 29: Need to render modal this way. If we use the `useLazyModal` hook, it does not open the virtual keyboard on Safari on mobile.

packages/shared/src/components/drawers/ContextMenuDrawer.tsx
```diff
@@ -55,7 +53,7 @@ export function ContextMenuDrawer({
                 className={classes}
                 onClick={(event) => {
                   action({ value: label, index, event });
-                  ref.current.onClose();
+                  drawerProps.onClose?.(event);
```
kopancek, line 56: We already have `onClose` in the `drawerProps` that we can use to close the context menu. No need for having a `ref` anymore.

packages/shared/src/components/comments/CommentInputOrModal.tsx
sshanzel: Good find.

packages/shared/src/components/modals/common.tsx
```diff
@@ -139,6 +143,7 @@ export const modals = {
   [LazyModal.ReputationPrivileges]: ReputationPrivilegesModal,
   [LazyModal.MarketingCta]: MarketingCtaModal,
   [LazyModal.UserSettings]: UserSettingsModal,
+  [LazyModal.Comment]: CommentModal,
```
kopancek, line 146: We could theoretically not include the comment modal in the lazy modal infra, since we are rendering it directly. I decided to keep it here in case we want to trigger it from other places manually. Let me know if you would instead prefer to remove it...

packages/shared/src/components/modals/post/CommentModal.tsx
```diff
@@ -0,0 +1,254 @@
+import React, {
```
kopancek, line 1: Code in this file is mostly copied over from the `CommentInputPage.tsx`. And adapted to work in a modal.

packages/shared/src/components/comments/CommentInputOrModal.tsx
sshanzel: Would it be possible to lazy load this through `dynamic` and open the modal with the keyboard on?

packages/shared/src/components/drawers/ContextMenuDrawer.tsx
sshanzel: Can you do some sanity checks on other context menus just to make sure they aren't broken after this?

packages/shared/src/components/modals/common.tsx
sshanzel: We might have to not add it here completely since if we are going to need it, most likely we want the keyboard opened and this is only most likely used on mobile. But I'm open to keeping it here. Shouldn't hurt to be more prepared.

packages/shared/src/components/modals/post/CommentModal.tsx
sshanzel: Can you apply the recent changes on the `CommentInputPage` to utilize the new Context for writing a comment. This is to fix the state of the buttons on the header along with keyboard shortcuts.

packages/shared/src/components/fields/MarkdownInput/CommentMarkdownInput.tsx
```diff
@@ -4,8 +4,10 @@ import React, {
   FormHTMLAttributes,
   ReactElement,
   useEffect,
+  useLayoutEffect,
   useRef,
 } from 'react';
+import { FocusOn } from 'react-focus-on';
```
sshanzel, line 10: Is this a new package? Seems like not used and not found in our `package.json`.

sshanzel (commented):
Looks great overall! Thank you for the effort! 🚀 

Just had some minor comments and merge concerns.

packages/shared/src/components/fields/MarkdownInput/CommentMarkdownInput.tsx
kopancek: that's just something I was testing locally, removed 👍

packages/shared/src/components/drawers/ContextMenuDrawer.tsx
kopancek: I did sanity check post context menu and squad context menu, seems to be working as expected.

sshanzel: Nice! Thank you! 🚀

sshanzel: I think the only thing we will lose here is the animation if we are not using the `ref`.

packages/shared/src/components/comments/CommentInputOrModal.tsx
kopancek: With `dynamic` it does not open the keyboard reliably. So I think we need to stick with static import. It's just one modal in the end, should not impact the load times too much 🤞

packages/shared/src/components/modals/common.tsx
kopancek: OK, I'll remove it for now, since it's not needed to be lazily loaded.

packages/shared/src/components/modals/post/CommentModal.tsx
```diff
@@ -0,0 +1,262 @@
...
+
+  return undefined;
+};
+
+// statically defined heights of various containers that
+// affect final height of the input container
+// when styles change, these should be updated accordingly
+const headerSize = 57;
+const replySize = 40;
+const footerSize = 48;
```
rebelchris, line 71: This looks highly bug prone.
Can't we get them from calculations in JS rather if you need them there?

Maybe by using: `globalThis?.window?.getComputedStyle`

packages/shared/src/components/modals/post/CommentModal.tsx
```diff
@@ -0,0 +1,262 @@
...
+  onRequestClose,
+  onAfterClose,
+  onCommented,
+  parentCommentId,
+  replyToCommentId,
+  editCommentId,
+  post,
+}: CommentModalProps): ReactElement {
+  const [value, setValue] = useState('');
+  const [styleHeight, setStyleHeight] = useState<number | 'auto'>('auto');
```
rebelchris, line 90: Feels like with a ref you better off here actually.

packages/shared/src/components/modals/post/CommentModal.tsx
```diff
@@ -0,0 +1,262 @@
...
+
+  const { height } = useVisualViewport();
+  const replyHeight = height > 0 ? replySize : 0;
+  const footerHeight = shouldShowCta ? footerSize : 0;
+  const totalHeight = height - headerSize - replyHeight - footerHeight;
+  const inputHeight = totalHeight > 0 ? Math.max(totalHeight, 300) : 'auto';
+
+  useEffect(() => {
+    setStyleHeight(inputHeight);
+  }, [inputHeight, setStyleHeight]);
```
rebelchris, line 156: Don't think you need this actually?
Can just be outside the effect no?

packages/shared/src/components/modals/post/CommentModal.tsx
```diff
@@ -0,0 +1,262 @@
...
+    }
+    return { submitCopy: 'Comment' };
+  }, [isEdit, isReply, comment, user?.id]);
+
+  return (
+    <Modal
+      isOpen={isOpen}
+      onRequestClose={onRequestClose}
+      onAfterClose={onAfterClose}
+      className="!h-full !max-h-none !w-full !border-none"
```
rebelchris, line 182: Why all the overwrites?
Don't we just have this type of modal already?

rebelchris (changes requested):
Minor concerns about the computations in the modal.
And agree with Lee, don't add it to lazy modals if we don't use it 👍

packages/shared/src/components/modals/post/CommentModal.tsx
sshanzel: Aalso, modals should now be default full-screen on mobile.

packages/shared/src/components/drawers/ContextMenuDrawer.tsx
kopancek: In that case, it might be better to revert to previous behavior and manually close the drawer when we trigger the edit.

packages/shared/src/components/modals/post/CommentModal.tsx
kopancek: I think this is just copy & paste from the page, which was not using a modal. Removing 👍

packages/shared/src/components/modals/post/CommentModal.tsx
kopancek: Refactored to get the proper sizes from element refs instead.

packages/shared/src/components/modals/post/CommentModal.tsx
kopancek: Refactored, the new implementation is setting the style prop directly on the current ref instead.

packages/shared/src/components/modals/post/CommentModal.tsx
kopancek: Refactored.

packages/shared/src/components/comments/CommentInputOrModal.tsx
```diff
@@ -18,38 +20,22 @@ interface CommentInputOrPageProps
...
-export default function CommentInputOrPage({
+export default function CommentInputOrModal({
   onClose,
   className,
   ...props
-}: CommentInputOrPageProps): ReactElement {
-  const isMobile = useViewSize(ViewSize.MobileL);
-  const router = useRouter();
+}: CommentInputOrModalProps): ReactElement {
+  const isModal = !useViewSize(ViewSize.Laptop);
```
sshanzel, line 28: I believe we should keep the same behavior on tablet.

image

kopancek: True, it should work as today for tablet 👍 Changing...

kopancek: Fixed

sshanzel (approved):
I don't have anything to add. Everything looked great. Let's just test all the cases before merging.

- Submit through header on mobile.
- Submit through button on tablet/laptop.
- Submit through keyboard shortcut on laptop.

These cases both on new comment and replying to a comment.

packages/shared/src/graphql/fragments.ts
```diff
@@ -117,11 +117,15 @@ export const COMMENT_FRAGMENT = gql`
   fragment CommentFragment on Comment {
     id
     contentHtml
+    content
```
rebelchris, line 120: Why is this addition needed, mainly asking becuase comments are already heavy to load and adding more fields will just make them slower.

If it's just for the edit we actually talked about just doing it from a seperate call on the edit itself.
(but let me know what other placement you needed it for)

packages/shared/src/graphql/fragments.ts
```diff
@@ -117,11 +117,15 @@ export const COMMENT_FRAGMENT = gql`
...
     createdAt
     lastUpdatedAt
     permalink
     upvoted
     numUpvotes
+    parent {
+      id
+    }
```
rebelchris, line 128: Same here, what did you need it for?
I think only comment feed uses this altogether right? and that has it's own type of fragment for this case.

rebelchris (approved):
Not blocking but two questions about the load of graphQL fields 🙏

DragosIorgulescu (approved):
Nothing to add other than what's been mentioned. Nice work 🫡

packages/shared/src/graphql/fragments.ts
kopancek: Yes, it's just for the edit.

packages/shared/src/graphql/fragments.ts
kopancek: In this case, we need to know the parent comment id for both replying and edit. I think it's more a result of refactoring to a page and then back to modal. We could supply this as a prop as well if you're concerned about it.

PR #2913, opened 2024-04-03
fix: prevent default event handling when opening context menu

DragosIorgulescu:
## Changes

- this is already applied in most places, was not being handled on the squad members modal users list

https://github.com/dailydotdev/apps/assets/1225100/ef2aa2c0-1a6e-42f8-b9cb-7ecab3ae9f03

## Events

N / A

- [x] MobileL (420px)
- [x] Tablet (656px)
- [x] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

MI-260 #done


sshanzel (approved):
LGTM! 🚀

sshanzel:
Would it be not possible to handle this generally? Like the component we use for the list?

DragosIorgulescu:
> Would it be not possible to handle this generally? Like the component we use for the list?

We can apply it generally [here](https://github.com/dailydotdev/apps/blob/a8c0039bf5a807eaa5ede1fecce42b3b9f8e01f0/packages/shared/src/hooks/useContextMenu.ts#L44-L60) but I'm hesitant to do that as I am not sure if would break or mess up with anything wherever this is used. 

When I initially refactored this I did try to go through all menus and make sure this is handled wherever it is needed, but I must have missed this one.

sshanzel:
> @sshanzel good point, we can apply it generally [here](https://github.com/dailydotdev/apps/blob/a8c0039bf5a807eaa5ede1fecce42b3b9f8e01f0/packages/shared/src/hooks/useContextMenu.ts#L44-L60) but I'm hesitant to do that as I am not sure if would break or mess up with anything wherever this is used.

Yeah, if ever we decide to change it there, we will fix all the other cases. Up to you, we can merge or if you want to make changes.

rebelchris (approved):

PR #2915, opened 2024-04-03
fix: add all nav items to the footer even for anon users

DragosIorgulescu:
## Changes

Screenshot 2024-04-03 at 18 53 59

## Events

N / A

- [x] MobileL (420px)
- [x] Tablet (656px)
- [x] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

MI-265 #done

### Preview domain
https://mi-265-footer-nav-anon-user.preview.app.daily.dev

### DIFF
--- a/packages/shared/src/components/post/write/CreatePostButton.tsx
+++ b/packages/shared/src/components/post/write/CreatePostButton.tsx
@@ -19,6 +19,7 @@ interface CreatePostButtonProps {
   className?: string;
   compact?: boolean;
   sidebar?: boolean;
+  footer?: boolean;
   onClick?: () => void;
 }
 
@@ -26,6 +27,7 @@ export function CreatePostButton({
   className,
   compact,
   sidebar,
+  footer,
   onClick,
 }: CreatePostButtonProps): ReactElement {
   const { user, squads } = useAuthContext();
@@ -38,7 +40,7 @@ export function CreatePostButton({
     verifyPermission(item, SourcePermissions.Post),
   );
 
-  if (!user || !squads?.length) {
+  if (!footer && (!user || !squads?.length)) {
     return null;
   }
 
@@ -74,11 +76,15 @@ export function CreatePostButton({
     >
       <Button
         {...buttonProps}
-        variant={sidebar ? ButtonVariant.Float : ButtonVariant.Secondary}
+        variant={
+          sidebar || footer ? ButtonVariant.Float : ButtonVariant.Secondary
+        }
         className={className}
         disabled={getIsDisabled()}
         icon={compact && <PlusIcon />}
-        size={isLaptop || sidebar ? ButtonSize.Medium : ButtonSize.Small}
+        size={
+          isLaptop || sidebar || footer ? ButtonSize.Medium : ButtonSize.Small
+        }
       >
         {!compact ? 'New post' : null}
       </Button>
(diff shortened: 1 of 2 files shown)


sshanzel (approved):
Looks good to me! 🚀

packages/shared/src/components/post/write/CreatePostButton.tsx
```diff
@@ -38,7 +40,7 @@ export function CreatePostButton({
     verifyPermission(item, SourcePermissions.Post),
   );
 
-  if (!user || !squads?.length) {
+  if (!footer && (!user || !squads?.length)) {
```
rebelchris, line 43: So footer is mantadory?
A bit unsure why this return is needed

rebelchris (approved):
Just one question I'm not sure about, but non blocking.

packages/shared/src/components/post/write/CreatePostButton.tsx
DragosIorgulescu: This check ensures the button is rendered if the footer prop is present. Otherwise, the following checks would kick in and the btn would not be rendered at all in the case where there is no user. 

Essentially we want to make sure that if this button is rendered in the footer, we don't care if there is no user, so it's always rendered

packages/shared/src/components/post/write/CreatePostButton.tsx
```diff
@@ -74,11 +76,15 @@ export function CreatePostButton({
     >
       <Button
         {...buttonProps}
-        variant={sidebar ? ButtonVariant.Float : ButtonVariant.Secondary}
+        variant={
+          sidebar || footer ? ButtonVariant.Float : ButtonVariant.Secondary
```
nensidosari, line 80: With the above check where we return null if !footer, wouldn't this always be true?

packages/shared/src/components/post/write/CreatePostButton.tsx
```diff
@@ -74,11 +76,15 @@ export function CreatePostButton({
...
+        }
         className={className}
         disabled={getIsDisabled()}
         icon={compact && <PlusIcon />}
-        size={isLaptop || sidebar ? ButtonSize.Medium : ButtonSize.Small}
+        size={
+          isLaptop || sidebar || footer ? ButtonSize.Medium : ButtonSize.Small
```
nensidosari, line 86: Same as above here `With the above check where we return null if !footer, wouldn't this always be true?`

packages/webapp/components/FooterNavBar.tsx
```diff
@@ -126,12 +126,11 @@ export const mobileUxTabs: Tab[] = [
   {
     component: (onClick: () => void) => (
       <div key="new-post" className="text-center">
-        <CreatePostButton onClick={onClick} compact sidebar className="h-8" />
+        <CreatePostButton onClick={onClick} compact footer className="h-8" />
```
nensidosari, line 129: Why are we not sending sidebar anymore?

packages/shared/src/components/post/write/CreatePostButton.tsx
DragosIorgulescu: The above check is not returning `null` if `!footer`, it just makes sure that the subsequent `(!user || !squads?.length)` check is ignored if `footer` is `true`, otherwise, the button is rendered if there is a `user` and there are postable `squads`.

nensidosari (approved):

PR #2917, opened 2024-04-03
fix: footer nav bar

sshanzel:
## Changes
- A lot of refactor on the footer nav bar for maintainability. It was difficult to follow the ternary conditions which sometimes are only for the legacy and sometimes for the new one.
- We will proceed here only one or the other, so I instead introduced separate components to manage each variant.
- Key fixes here are:
  - Backdrop on the floating input.
  - Issue with indicator not working properly for control.
  - Support both v1 and control for indicators.

Previews:

https://github.com/dailydotdev/apps/assets/13744167/81a4d5f8-9233-4363-9897-30f0281a5c9a

Feel free to check the old behavior from production. Fixed preview:
https://github.com/dailydotdev/apps/assets/13744167/6cfc1ee9-8d1f-4c6d-b234-0a2df3cdb2cc

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

MI-256 #done

### Preview domain
https://mi-256.preview.app.daily.dev


packages/shared/src/styles/base.css
```diff
@@ -36,6 +36,15 @@ body {
   }
 
   overflow-x: hidden;
+
+  @screen mobileL {
```
sshanzel, line 40: This helps fixing the footer navbar. One of the issues is that on all pages, you must add extra padding due to the fact that we are using `fixed` property. With the properties below, this will make using `sticky` work.

packages/webapp/components/FooterNavBar.tsx
```diff
@@ -1,326 +1,104 @@
-import React, { ReactElement, useContext, useMemo, useState } from 'react';
```
sshanzel, line 1: Separated into two components.

packages/webapp/components/footer/FooterNavBarControl.tsx
```diff
@@ -0,0 +1,163 @@
+import React, { ReactElement } from 'react';
```
sshanzel, line 1: Basically the old FooterNavBar, just broken down into multiple components.

packages/webapp/components/FooterNavBar.tsx
```diff
@@ -1,326 +1,104 @@
...
+  home: 'Home',
+  search: 'Search',
+  bookmarks: 'Bookmarks',
+  notifications: 'Notifications',
+  squads: 'Squads',
+  profile: 'Profile',
+};
+
+const combinedTabs = tabs.concat(
+  ...(mobileUxTabs.filter((tab) => !isValidElement(tab)) as FooterTab[]),
```
DragosIorgulescu, line 39: why is this needed here?

DragosIorgulescu (commented):
Great work on the refactoring, I don't see anything obvious, overall looks great, just have one concern regarding the preview, it looks like the active `flipper` is delayed? 

It seems that it's always one step behind the current active tab? Eg. when navigating from search to squads, the active state moves from home to search, etc.

packages/webapp/components/FooterNavBar.tsx
sshanzel: This is if in case there is no active page returned by the `useActiveNav`, we will check the paths here if any matches the current route.

DragosIorgulescu: no, I mean why do you need to combine tabs if we are splitting the components rendering them?

packages/shared/src/components/MainLayout.tsx
```diff
@@ -225,7 +225,7 @@ function MainLayoutComponent({
       />
       <main
         className={classNames(
-          'flex flex-row',
+          'flex flex-col tablet:flex-row',
```
rebelchris, line 228: This could have massive impact on some pages as many use this.
We have to make tripple sure none break, think about devcard, sources, tags, whatever hidden page we have.

Can you just make sure before we merge (and really try to find all pages 🙏 )

packages/shared/src/styles/base.css
rebelchris: This feels very heavy to add to generic body tag 🙈 
I'm a bit worried it will have some implication on somethings we haven't tested.

packages/shared/src/styles/base.css
```diff
@@ -661,6 +670,14 @@ html.light .invert .invert,
   position: relative;
   flex-direction: column;
   align-items: stretch;
+
+  @screen mobileL {
+    flex: 1;
+  }
+
+  @screen tablet {
+    flex: unset;
```
rebelchris, line 679: This as well

rebelchris (commented):
The split is very good, love it.
Just have some concerns on the super global CSS implications, but not sure how to address them 😅

CC: @nensidosari maybe you have some idea how we can validate impact across the board?

packages/shared/src/styles/base.css
sshanzel: Yeah, same sentiments. However, I think it is best to use `sticky` on the footer navbar as it is problematic across all pages which this fixes. Also why I am immediately setting the values back to its original state after mobile.

packages/webapp/components/FooterNavBar.tsx
sshanzel: We have to support it the way we define which tab should be active for indicator using the `flipper`. Otherwise, we could've done it directly from either component.

packages/shared/src/components/MainLayout.tsx
sshanzel: Yeah, but I realized no pages are using row layout for mobile since it wouldn't fit. With that, definitely going to triple-check all pages.

packages/shared/src/styles/base.css
rebelchris: There is no way we can only target that element vs the whole body tag?

sshanzel: Let me check, though as far as I tested yesterday, those were the requirements. Let me check if I move the elements 1 level deeper.

packages/shared/src/styles/base.css
nensidosari: This just nulls out the one set above so not much concerns on it

nensidosari (approved):
Love the refactoring. 

As about the css changes, a bit concerned as well, in idea they look ok since it's only changing in mobile.

Since we don't have tests 😬 , no better idea 🤷‍♀️ I think the only way to test is manually. I can help test a bunch of pages, we can split if we want to make this faster and see if we detect anything obvious...

packages/shared/src/styles/base.css
sshanzel: I checked having it inside `MainLayout` component but still needed the CSS properties defined here. With that, I've created a list to check on all 3 screen sizes.

Maybe we should store this list somewhere to make it easier next time we have something similar to this in the future.
Screenshot 2024-04-04 at 5 35 04 PM

sshanzel:
> Since we don't have tests 😬 , no better idea 🤷‍♀️ I think the only way to test is manually. I can help test a bunch of pages, we can split if we want to make this faster and see if we detect anything obvious...

Thank you for the offer, I think I still have the time to check them myself, though I will share a list if you want to run through the same things I will be running through.

rebelchris (approved):
Won't block it, but will manually check your screenshots again :)

PR #2919, opened 2024-04-04
feat: feed nav horizontal scroll

DragosIorgulescu:
## Changes

- make spacing tighter between feed nav tabs
- auto-scroll the active tab into view on the feed nav

_screenshots pending preview deploy_

## Events

N / A

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

MI-266 #done

### DIFF
--- a/packages/shared/src/components/tabs/TabList.tsx
+++ b/packages/shared/src/components/tabs/TabList.tsx
@@ -1,5 +1,5 @@
 import classNames from 'classnames';
-import React, { ReactElement, useState } from 'react';
+import React, { ReactElement, useEffect, useRef, useState } from 'react';
 
 interface ClassName {
   indicator?: string;
@@ -11,6 +11,7 @@ export interface TabListProps {
   active: string;
   onClick?: (label: string) => unknown;
   className?: ClassName;
+  autoScrollActive?: boolean;
 }
 
 function TabList({
@@ -18,8 +19,20 @@ function TabList({
   active,
   onClick,
   className = {},
+  autoScrollActive,
 }: TabListProps): ReactElement {
-  const [offset, setOffset] = useState<number>(undefined);
+  const [offset, setOffset] = useState<number>(0);
+  const [indicatorOffset, setIndicatorOffset] = useState<number>(undefined);
+  const currentActiveTab = useRef<HTMLButtonElement>(null);
+
+  useEffect(() => {
+    if (autoScrollActive && currentActiveTab.current) {
+      currentActiveTab.current.parentElement.parentElement.scrollTo({
+        left: offset,
+        behavior: 'smooth',
+      });
+    }
+  }, [offset, autoScrollActive]);
 
   return (
     <ul className="relative flex flex-row">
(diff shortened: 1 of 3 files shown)


sshanzel (approved):
Tested on preview and worked really great! 🚀

nensidosari (commented):
I think is better then just reseting the scroll position.

The best idea would be just to keep the user scrolling, so if user scrolled to somepoint and clicked history, it is moved at that scroll position when history page is opened, if user doesn't scroll but just clicks on the tabs currently shown, then we keep that scrolling position.

Or to only scroll if the active element is not into view.

packages/shared/src/components/tabs/TabList.tsx
```diff
@@ -1,24 +1,49 @@
...
+      if (
+        activeTabRect.left < scrollableParentRect.left ||
+        activeTabRect.right > scrollableParentRect.right
+      ) {
+        currentActiveTab.current.parentElement.parentElement.scrollTo({
+          left: offset,
+          behavior: 'smooth',
+        });
+      }
+    }
```
nensidosari, line 45: if u extract this in a func called scrollIfActiveTabNotInView there is no need for the comment anymore :D

nensidosari (approved):
Love the new approach, better on UX i think. :D 

One small comment

packages/shared/src/components/tabs/TabList.tsx
```diff
@@ -30,17 +55,20 @@ function TabList({
...
               return;
             }
 
-            const size = rect.width / 2;
-            const value = size + el.offsetLeft;
-            setOffset(value);
+            const size = rect.width;
+            const leftOffset = el.offsetLeft;
+            setOffset(leftOffset);
+            setIndicatorOffset(size / 2 + leftOffset);
```
capJavert, line 68: Could `indicatorOffset` just always be computed as this? To avoid one `useState`?

packages/shared/src/components/tabs/TabList.tsx
```diff
@@ -1,24 +1,49 @@
...
+      const scrollableParentRect = scrollableParent.getBoundingClientRect();
+
+      // only try to scroll if the active tab is not in view
+      if (
+        activeTabRect.left < scrollableParentRect.left ||
+        activeTabRect.right > scrollableParentRect.right
+      ) {
+        currentActiveTab.current.parentElement.parentElement.scrollTo({
+          left: offset,
+          behavior: 'smooth',
```
capJavert, line 42: Not 100% sure but I think `scrollTo` or some of its options do not work on safari iOS or something so just check compatibility 🙏

DragosIorgulescu: found it hard to test because it keeps redirecting me to the main app.daily.dev website, but we'll merge and double check if we need to do smth differently for safari (or if there is anything we can do)

packages/shared/src/components/tabs/TabList.tsx
DragosIorgulescu: sure thing, that works too 😅

packages/shared/src/components/tabs/TabList.tsx
sshanzel: Removed the states completely since we already have a hold of which element is currently active. https://github.com/dailydotdev/apps/pull/2928/files#diff-a6e56221ccd7c5f4fe9c333d7082e8439f4b34772c3555880f158ffba9fd66dd

PR #2920, opened 2024-04-04
feat: use a drawer for the streaks tooltip on mobile

DragosIorgulescu:
## Changes

Screenshot 2024-04-04 at 11 01 04

---

Screenshot 2024-04-04 at 11 01 20

---

Screenshot 2024-04-04 at 11 01 33

## Events

N / A

- [x] MobileL (420px)
- [x] Tablet (656px)
- [x] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

MI-267 #done

### Preview domain
https://mi-267-streaks-drawer-mobile.preview.app.daily.dev

### DIFF
--- a/packages/shared/src/components/streak/ReadingStreakButton.tsx
+++ b/packages/shared/src/components/streak/ReadingStreakButton.tsx
@@ -10,6 +10,9 @@ import { isTesting } from '../../lib/constants';
 import { useAnalyticsContext } from '../../contexts/AnalyticsContext';
 import { AnalyticsEvent } from '../../lib/analytics';
 import { useMobileUxExperiment } from '../../hooks/useMobileUxExperiment';
+import { RootPortal } from '../tooltips/Portal';
+import { Drawer } from '../drawers';
+import ConditionalWrapper from '../ConditionalWrapper';
 
 interface ReadingStreakButtonProps {
   streak: UserStreak;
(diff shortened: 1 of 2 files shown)


nensidosari (approved):
Looks good

packages/shared/src/components/streak/ReadingStreakButton.tsx
```diff
@@ -86,31 +89,52 @@ export function ReadingStreakButton({
...
-        onClick={handleToggle}
-        className={classnames('gap-1', compact && 'text-theme-color-bacon')}
-        size={
-          (isLaptop || !compact) && !isMobile
-            ? ButtonSize.Medium
-            : ButtonSize.Small
-        }
+    <>
+      <ConditionalWrapper
+        condition={!isMobile}
```
sshanzel, line 94: Since there is content when it is not compact, I think we should also include the condition.
```suggestion
        condition={!isMobile && compact}
```

packages/shared/src/components/streak/popup/ReadingStreakPopup.tsx
```diff
@@ -83,7 +86,14 @@ export function ReadingStreakPopup({
         <StreakSection streak={streak.current} label="Current streak" />
         <StreakSection streak={streak.max} label="Longest streak 🏆" />
       </div>
-      <div className="mt-6 flex flex-row gap-2">{streaks}</div>
+      <div
+        className={classNames(
+          'mt-6 flex flex-row gap-2',
+          fullWidth && 'justify-between',
```
sshanzel, line 92: We could've maybe used a className object prop here, but I really don't have a strong opinion here. Just makes it more flexible. Not necessary, this is optional.

sshanzel (approved):
Looks great! Just had one optional comment, not really required. Up to you on that end 🚀

packages/shared/src/components/streak/ReadingStreakButton.tsx
DragosIorgulescu: Not sure I follow your comment here, the conditional wrapper is only for the `Tooltip`, which dictates wether it's wrapped around the button on non-mobile resolutions. It should have nothing to do with the `compact` property 🤔

sshanzel: Yep, but the thing is, we only display tooltip for buttons with text. Which when compact is applied, there is no text, and only when should be applied.

DragosIorgulescu: but tooltips are not a thing on mobile, right? so including `compact` here would not really matter. 

Also, since the `Tooltip` or `CustomStreaksTooltip` is not dictated by `compact`, I feel like we could mistakenly create a situation where someone uses the `compact` version and ends up displaying both the main streaks tooltip and the drawer. I would suggest we tinker with this only if the situation arises that requires it, WDYT?

packages/shared/src/components/streak/popup/ReadingStreakPopup.tsx
DragosIorgulescu: since it's a classname specific to the streak days list, we would have to create a `className` prop which allows passing classnames to each section of the component internals. My suggestion would be to only do that if we need to control this component's UI further in the future 🙏

packages/shared/src/components/streak/ReadingStreakButton.tsx
sshanzel: Ah, okay, I misunderstood, I thought this is only for tooltip to display a text. In that case, feel free to ignore and move on.

PR #2929, opened 2024-04-04
fix: post page mobile navigation should be sticky

sshanzel:
## Changes
- Make the mobile navigation sticky for easy-access.
- Refactored the `FixedNavigation` to be coming from the same container.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

MI-271 #done

### Preview domain
https://mi-271.preview.app.daily.dev


packages/shared/src/components/post/SquadPostContent.tsx
```diff
@@ -73,71 +73,72 @@ function SquadPostContent({
...
 
   return (
-    <>
-      {position === 'fixed' && (
-        <FixedPostNavigation
-          {...navigationProps}
-          isBannerVisible={isBannerVisible}
-          onReadArticle={onReadArticle}
-          className={className?.fixedNavigation}
-        />
```
sshanzel, line 83: Diff looks crazy due to the removed unnecessary Fragment, but the only thing that changed here was this block passing it to as props.

packages/shared/src/components/post/PostContentContainer.tsx
```diff
@@ -21,10 +27,29 @@ function PostContentContainer({
...
+      condition={isNavigationOutside}
+      wrapper={(component) => (
+        <>
+          {props && <FixedPostNavigation {...props} />}
+          {component}
+        </>
+      )}
+    >
+      <Wrapper className={className}>
+        {props && !isNavigationOutside && <FixedPostNavigation {...props} />}
```
nensidosari, line 48: I don't think this double check is necessary since the wrapper is looking out for it `!isNavigationOutside`

nensidosari (approved):
Looks good

rebelchris (approved):

packages/shared/src/components/post/PostContentContainer.tsx
sshanzel: Yeah, but the wrapper renders this part too.

PR #2936, opened 2024-04-05
chore: reduce number of fields we fetch for post comments

kopancek:
We originally added more fields to facilitate comment modal dialog. However we can get the same data by querying comment by Id when the modal itself is opened instead. That way, we do not unnecessary query fields, that we don't need to have most of the time.

See comments on the original PR:
- https://github.com/dailydotdev/apps/pull/2912#discussion_r1550978383
- https://github.com/dailydotdev/apps/pull/2912#discussion_r1550979025

- [X] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

### Preview domain
https://chore-graphql-comment-fields.preview.app.daily.dev

### DIFF
--- a/packages/shared/src/components/modals/post/CommentModal.tsx
+++ b/packages/shared/src/components/modals/post/CommentModal.tsx
@@ -94,8 +94,8 @@ export default function CommentModal({
   const isReply = !isEdit && !!replyToCommentId;
 
   const { comment: commentById } = useCommentById({
-    id: isEdit ? editCommentId : parentCommentId,
-    options: { enabled: !!editCommentId || !!parentCommentId },
+    id: editCommentId,
+    options: { enabled: !!editCommentId },
   });
 
   const refCallback = useCallback(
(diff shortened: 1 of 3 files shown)


packages/shared/src/components/modals/post/CommentModal.tsx
```diff
@@ -92,6 +93,11 @@ export default function CommentModal({
   const isEdit = !!editCommentId;
   const isReply = !isEdit && !!replyToCommentId;
 
+  const { comment: commentById } = useCommentById({
+    id: isEdit ? editCommentId : parentCommentId,
+    options: { enabled: !!editCommentId || !!parentCommentId },
```
capJavert, line 98: When is the modal triggered with `parentCommentId` id only?

packages/shared/src/graphql/fragments.ts
```diff
@@ -117,15 +117,11 @@ export const COMMENT_FRAGMENT = gql`
   fragment CommentFragment on Comment {
     id
     contentHtml
-    content
```
capJavert, line 120: Just double check we really do not need content, this fragment is used all over the place 🙏

packages/shared/src/components/modals/post/CommentModal.tsx
kopancek: In case of replies. But in that case, I guess we don't need to fetch, since we should have everything already. Let me modify the code, so that it's clearer.

packages/shared/src/graphql/fragments.ts
kopancek: Yup, I only added §content§ field in the previous PR and this was the only place it was used AFAIK.

DragosIorgulescu (approved):
👍

PR #2952, opened 2024-04-08
fix: add new experiment to show tags on cards

rebelchris:
- New experiment to show tags on the cards
- Also fixed a minor issue with articles cards having wrong margings
- And collections missing the tags on the modal/page

Screenshot 2024-04-08 at 16 18 12

Screenshot 2024-04-08 at 16 18 17

Screenshot 2024-04-08 at 16 08 19

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

AS-236 #done

### Preview domain
https://as-236-micro-show-tags-on-card.preview.app.daily.dev

### DIFF
--- a/packages/shared/src/components/cards/PostTags.tsx
+++ b/packages/shared/src/components/cards/PostTags.tsx
@@ -1,4 +1,4 @@
-import React, { ReactElement, useMemo, useRef } from 'react';
+import React, { ReactElement, useMemo } from 'react';
 import { Post } from '../../graphql/posts';
 import Classed from '../../lib/classed';
 
--- a/packages/shared/src/components/cards/PostTags.tsx
+++ b/packages/shared/src/components/cards/PostTags.tsx
@@ -1,0 +1,39 @@
+import React, { ReactElement, useMemo, useRef } from 'react';
+import classNames from 'classnames';
+import { Post } from '../../graphql/posts';
+import Classed from '../../lib/classed';
+
+type PostTagsProps = Pick<Post, 'tags'>;
+
+const Chip = Classed(
+  'div',
+  'rounded-8 border border-border-subtlest-tertiary px-2 h-6 flex items-center justify-center typo-footnote text-text-quaternary my-2',
+);
+
+export default function PostTags({ tags }: PostTagsProps): ReactElement {
+  const totalLength = useRef(0);
+  const tagList = useMemo(() => {
+    return tags.reduce((acc, tag) => {
+      if (totalLength.current >= 15) {
+        return acc;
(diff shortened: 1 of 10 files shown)


packages/shared/src/components/post/collection/CollectionPostContent.tsx
```diff
@@ -139,6 +140,7 @@ export const CollectionPostContent = ({
             >
               {post.title}
             </h1>
+            <TagLinks tags={post.tags || []} />
```
capJavert, line 143: Should this be under feature flag?

packages/shared/src/components/cards/PostTags.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+type PostTagsProps = Pick<Post, 'tags'>;
+
+const Chip = Classed(
+  'div',
+  'rounded-8 border border-border-subtlest-tertiary px-2 h-6 flex items-center justify-center typo-footnote text-text-quaternary my-2',
+);
+
+export default function PostTags({ tags }: PostTagsProps): ReactElement {
+  const totalLength = useRef(0);
```
capJavert, line 13: Why do you need ref when you are not setting it anywhere else? It can be just local var in `useMemo` right?

capJavert: Also can we maybe just do `slice(0, 15)` and then `map` it would save iterations and make it simpler I think.

capJavert (commented):
Looks good, some questions on implementation.

packages/shared/src/components/cards/PostTags.tsx
rebelchris: The slice is dynamic it's not one piece, so it can either be 2 or 3 or 4 tags.
I had the ref as let first, but eslint was complaining ref would be better

packages/shared/src/components/post/collection/CollectionPostContent.tsx
rebelchris: These are not for the experiment, just something we never added on the modal but should be there always regardless :)

packages/shared/src/components/cards/PostTags.tsx
capJavert: > he slice is dynamic it's not one piece

Sorry, I don't understand? 

> I had the ref as let first, but eslint was complaining ref would be better

Probably because you had it outside of `useMemo`

PR #2953, opened 2024-04-08
fix: post share loop showing on downvote

kopancek:
## Changes

Fixed the `usePostShareLoop` hook so that it only returns `shouldShowOverlay === true` when it should - on user upvote action. Downvote or None should not cause the share loop to be shown.

Added unit test for the `usePostShareLoop` hook as well.

- [X] MobileL (420px)
- [ ] Tablet (656px)
- [X] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

MI-281 #done

### Preview domain
https://mi-281-fix-downvote-share-loop.preview.app.daily.dev


DragosIorgulescu (approved):
bonus points for the test 💪

PR #2955, opened 2024-04-08
fix: glitchy icon on scroll

sshanzel:
## Changes
- Fix the glitchy icon when scrolling the screen. See ticket for more details on how to reproduce it. UI-wise, no change.
- `fixed` and `absolute` doesn't work since `sticky` behaves differently (which is the parent element).

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

MI-268 #done

### Preview domain
https://mi-268.preview.app.daily.dev

### DIFF
--- a/packages/shared/src/components/feeds/FeedNav.tsx
+++ b/packages/shared/src/components/feeds/FeedNav.tsx
@@ -75,7 +75,7 @@ function FeedNav(): ReactElement {
         <Tab label={FeedNavTab.History} url="/history" />
       </TabContainer>
       {isMobile && sortingEnabled && isSortableFeed && (
-        <div className="fixed right-0 top-12 my-1 flex h-12 w-20 items-center justify-end bg-gradient-to-r from-transparent via-background-default via-40% to-background-default pr-4">
+        <div className="sticky flex h-12 w-20 -translate-y-12 translate-x-[calc(100vw-100%)] items-center justify-end bg-gradient-to-r from-transparent via-background-default via-40% to-background-default pr-4">
           <Dropdown
             className={{ label: 'hidden', chevron: 'hidden', button: '!px-1' }}
             dynamicMenuWidth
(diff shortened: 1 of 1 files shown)


packages/shared/src/components/feeds/FeedNav.tsx
```diff
@@ -75,7 +75,7 @@ function FeedNav(): ReactElement {
         <Tab label={FeedNavTab.History} url="/history" />
       </TabContainer>
       {isMobile && sortingEnabled && isSortableFeed && (
-        <div className="fixed right-0 top-12 my-1 flex h-12 w-20 items-center justify-end bg-gradient-to-r from-transparent via-background-default via-40% to-background-default pr-4">
+        <div className="sticky flex h-12 w-20 -translate-y-12 translate-x-[calc(100vw-100%)] items-center justify-end bg-gradient-to-r from-transparent via-background-default via-40% to-background-default pr-4">
```
rebelchris, line 78: Don't we get issues now like we had with the bottom bar? (think it was the sticky there right on ios that was the issue)

nensidosari: This looks ok from my test on ios as well.

nensidosari: My only concern is that design wise now it's different, the bottom line of nav doesn't show anymore where sort icon is located

IMG_8033

sshanzel: The thing is, sticky is easy to work with when on top of the layout. Shouldn't be an issue in this case and the header has been sticky since then.

sshanzel: @nensidosari should be fixable by applying a negative margin or reduce the y-translate.

kopancek (approved):

packages/shared/src/components/feeds/FeedNav.tsx
nensidosari: Fixed what I mentioned above on this commit [21fd003](https://github.com/dailydotdev/apps/pull/2955/commits/21fd0030b7a05cd2a55f9553bf1d89b8ea8ce10f)

nensidosari (approved):
Looks ok, I think we can move on and rollback if we see any issues reported again from this change

PR #2960, opened 2024-04-09
fix: feed nav reacts to route changes

kopancek:
## Changes

- Added new hook `useRouterOnChange` which handles subscribing to nextjs router path changes
- Added new prop `shouldMatchUrl` to `TabContainer` that uses this hook to match the active tab to the current route
- Use the `shouldMatchUrl` prop in the `FeedNav` component, so that selected tab always reflects current route
- Unit test for the new hook

## Related issue
- https://dailydotdev.atlassian.net/browse/MI-282

- [X] MobileL (420px)
- [X] Tablet (656px)
- [X] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

MI-282 #done

### Preview domain
https://mi-282-feed-nav-fix.preview.app.daily.dev

### DIFF
--- a/packages/shared/src/components/feeds/FeedNav.tsx
+++ b/packages/shared/src/components/feeds/FeedNav.tsx
@@ -33,7 +33,8 @@ function FeedNav(): ReactElement {
   const { feedName } = useActiveFeedNameContext();
   const { sortingEnabled } = useContext(SettingsContext);
   const { isSortableFeed } = useFeedName({ feedName });
-  const { home: shouldRenderNav, notifications } = useActiveNav(feedName);
+  const { home: shouldRenderNav } = useActiveNav(feedName);
+
   const isMobile = useViewSize(ViewSize.MobileL);
   const [selectedAlgo, setSelectedAlgo] = usePersistentContext(
     DEFAULT_ALGORITHM_KEY,
@@ -54,7 +55,7 @@ function FeedNav(): ReactElement {
     >
       {isMobile && <MobileFeedActions />}
       <TabContainer
-        controlledActive={notifications ? 'notifications' : undefined}
+        shouldMatchUrl
         shouldMountInactive
         className={{
           header: classNames(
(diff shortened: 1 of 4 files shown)


packages/shared/src/components/feeds/FeedNav.tsx
```diff
@@ -54,7 +55,7 @@ function FeedNav(): ReactElement {
     >
       {isMobile && <MobileFeedActions />}
       <TabContainer
-        controlledActive={notifications ? 'notifications' : undefined}
+        shouldMatchUrl
```
capJavert, line 58: I think we can listen to `useRouter.pathname` and pass that as `controlledActive`? We can make a map `tabUrlToLabel` and then use it below for `controlledActive` and also `Tab`.

This would add no new code to `TabContainer` and also avoid adding new hook.

nensidosari: +1 was trying point to how we do these kind of checks currently.

nensidosari (commented):
One thing I'm not entirely sure about is, what is the expected behaviour. 

In steps Chris reported:

- open app
- Click discussions
- Click home icon in footer bar
- You go to my feed, but discussions stays active

Should we actually go to my feed when clicking home, or we should stay at discussions where we are since that is a tab within home navigation

packages/shared/src/components/feeds/FeedNav.tsx
kopancek: I did it like that before, but there was a visible lag when the route change was taking more time

capJavert: I would like to see it on preview, where pages are prerender? Because I think it will not be as near as bad as in dev.

capJavert: Either way I don't think we need to add internal state to `TabContainer`, it will just cause problems down the line when you have too props that cancel each other.

kopancek: There already was internal state that controlled active tab. Let me create another PR with controlledActive and we can see how it behaves in preview :)

kopancek: Here's the other approach: 
- https://github.com/dailydotdev/apps/pull/2963

PR #2962, opened 2024-04-09
fix: add shortcut experiment version

rebelchris:
- Added feature for new shortcut system
- Personally find it a little bit messy because of the 3 different placements depending on 3 other experiments 😢  (but for the sake of shipping fast this was the cleanest way)

Screenshot 2024-04-09 at 15 46 40
Screenshot 2024-04-09 at 15 46 02
Screenshot 2024-04-09 at 15 45 55
Screenshot 2024-04-09 at 15 46 07
Screenshot 2024-04-09 at 15 46 11

## Events

|--------|------------|-------|
| Change| impression (shortcuts)  | added sources `browser, custom, placeholder, button` |
(Basically it can now trigger 2 impressions (one for the button/placeholder and one for the actual rendered element)

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

AS-200 #done

### Preview domain
https://as-200-shortcut-experiment.preview.app.daily.dev

### DIFF
--- a/packages/extension/src/newtab/ShortcutLinks.tsx
+++ b/packages/extension/src/newtab/ShortcutLinks.tsx
@@ -143,8 +143,6 @@ export default function ShortcutLinks({
     shouldEvaluate: showTopSites,
     returnDefault: true,
   });
-  console.log('isShortcutsV1', isShortcutsV1);
-
   const { onMenuClick, isOpen } = useContextMenu({
     id: ContextMenu.ShortcutContext,
   });
--- a/packages/extension/src/newtab/ShortcutLinks.tsx
+++ b/packages/extension/src/newtab/ShortcutLinks.tsx
@@ -141,7 +141,6 @@ export default function ShortcutLinks({
   const { value: isShortcutsV1 } = useConditionalFeature({
     feature: feature.onboardingMostVisited,
     shouldEvaluate: showTopSites,
-    returnDefault: true,
   });
   const { onMenuClick, isOpen } = useContextMenu({
     id: ContextMenu.ShortcutContext,
@@ -223,31 +222,6 @@ export default function ShortcutLinks({
     });
   };
 
-  const ShortcutControl = () => {
-    return (
-      <>
-        {shortcutLinks?.length ? (
-          <CustomLinks
-            links={shortcutLinks}
-            className={className}
-            onOptions={onOptionsOpen}
-            onLinkClick={onLinkClick}
-          />
-        ) : (
-          <Button
-            className={className}
-            variant={ButtonVariant.Tertiary}
-            icon={<PlusIcon />}
-            iconPosition={ButtonIconPosition.Right}
-            onClick={onOptionsOpen}
-          >
-            Add shortcuts
-          </Button>
-        )}
-      </>
-    );
-  };
-
   const ShortcutV1 = () => {
     return (
       <div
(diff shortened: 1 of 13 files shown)


packages/shared/src/components/feeds/FeedContainer.tsx
```diff
@@ -100,6 +106,11 @@ export const FeedContainer = ({
   shortcuts,
   actionButtons,
 }: FeedContainerProps): ReactElement => {
+  const { value: isShortcutsV1 } = useConditionalFeature({
+    feature: feature.onboardingMostVisited,
+    shouldEvaluate: !!shortcuts,
+    returnDefault: true,
```
nensidosari, line 112: Not sure yet why we needed, but would suggest a naming of `allowReturnDefault` rather then the current one since it can be mistaken as if the func will always return default

packages/extension/src/newtab/ShortcutLinks.tsx
```diff
@@ -101,31 +215,74 @@ export default function ShortcutLinks({
...
+              className={className}
+              variant={ButtonVariant.Tertiary}
+              icon={<PlusIcon />}
+              iconPosition={ButtonIconPosition.Right}
+              onClick={onOptionsOpen}
+            >
+              Add shortcuts
+            </Button>
+          )}
+        </>
```
nensidosari, line 285: 😢 too many ? : ? : ...

packages/extension/src/newtab/ShortcutLinks.tsx
```diff
@@ -101,31 +215,74 @@ export default function ShortcutLinks({
...
+                variant={ButtonVariant.Tertiary}
+                onClick={toggleShowTopSites}
+                size={ButtonSize.Small}
+                icon={<ClearIcon secondary />}
+                className="mt-2"
+              />
+            </>
+          )}
+        </div>
+      ) : (
```
nensidosari, line 265: Can we move these inner components + conditions in an other named component that maybe it's fine to be in this file but makes it easier to read?

Valid for conditions below as well

nensidosari (commented):
Only one main concern on the return default, if u can elaborate why we need it?

Happy to approve after, all other comments of mine are improvements that I'll leave to your judgment if it's best to do now or later

packages/shared/src/hooks/useConditionalFeature.ts
```diff
@@ -47,5 +49,11 @@ export const useConditionalFeature = <T extends JSONValue>({
...
   );
 
-  return { value: featureValue, isLoading };
+  return {
+    value:
+      returnDefault && !featureValue
+        ? (feature.defaultValue as WidenPrimitives<T>)
+        : featureValue,
+    isLoading,
+  };
```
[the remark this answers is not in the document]
rebelchris: Basically wanted a way to indicate to return defaultValue if that option is set.
We can't "always" return default as for some we want to wait for the loading state.

packages/shared/src/components/feeds/FeedContainer.tsx
rebelchris: It won't always return default though.
In the default way it won't even get evaluated, thus return undefined forever.

packages/extension/src/newtab/ShortcutLinks.tsx
rebelchris: I had that, but it failed the tests because of too many re-renders 😅

capJavert: Maybe `withExperiment` can help to just split out the components into two and we don't need to test the experimented ones?

packages/shared/src/hooks/useConditionalFeature.ts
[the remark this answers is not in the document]
rebelchris: We can do that, just not sure at this point if it will introduce many side effects for the existing use-cases 
Let me try and see quickly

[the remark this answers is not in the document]
capJavert: Its not used in many places so we can add isLoading on those if it creates issues?

packages/extension/src/newtab/ShortcutLinks.tsx
rebelchris: @capJavert // @nensidosari Jest really doesn't like the existing one as inline. component.
I did the v1 as one, hopefully that's ok for now?

Also removed the default value in favor of abstracting logic to component itself.

capJavert (approved):

nensidosari (approved):

PR #2969, opened 2024-04-10
feat: add seo meta to tag pages if it exists for keyword

capJavert:
- fetch keyword and add meta to SEO if exists
- API PR https://github.com/dailydotdev/daily-api/pull/1818

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

AS-241 #done

### Preview domain
https://as-241-tag-page-description.preview.app.daily.dev


packages/webapp/pages/tags/[tag].tsx
```diff
@@ -185,12 +192,34 @@ interface TagPageParams extends ParsedUrlQuery {
...
+  } catch (error) {
+    // keyword not found, ignoring for now
+  }
+
   return {
     props: {
       tag: params.tag,
+      initialData,
     },
+    revalidate: 3600,
```
Comment (capJavert) [inline packages/webapp/pages/tags/[tag].tsx:223]: Since feed is client side fetched I think 1 hour is good time to cache this. We can potentially do even more.

Comment (capJavert) [reply packages/webapp/pages/tags/[tag].tsx:223]: If we want we can also prerender some good keywords on each build with getStaticPaths.

rebelchris (approved):
Nice looks great

rebelchris (approved):

packages/shared/src/graphql/keywords.ts
```diff
@@ -2,11 +2,17 @@ import { gql } from 'graphql-request';
...
+  title?: string;
+  description?: string;
+};
+
 export interface Keyword {
   __typename?: string;
   value: string;
   occurrences: number;
   status: KeywordStatus;
+  flags?: KeywordFlags;
```
nensidosari, line 15: is flags an array of flags or just object? if just one, maybe better name it flag 🙏

capJavert: We call them `flags` in all entities, its is set of properties.

PR #2971, opened 2024-04-11
fix: ascii logo with "colours"

omBratteng:
Console logo is now bigger and looks to have "shading" and more similar to actual logo

image

### Preview domain
https://prettier-console-logo.preview.app.daily.dev


rebelchris (approved):
It kind of was smaller for people with smaller debug windows (on the side)
So it wouldn't break to quick.

But not even sure how many people see this 😅

omBratteng:
Enough people to make [this post](https://dly.to/zQxqZZeKQSu)

PR #2974, opened 2024-04-11
fix: show dailydev logo instead of back button if goback is not possible

kopancek:
## Changes

When going back in browser history is not possible, the back button would lead to no action.

This fix replaces the back button with daily.dev logo, leading to the default feed if that's the case.

Based on slack discussion here: https://dailydotdev.slack.com/archives/C0650M2KX8E/p1712770830284389

The fix only applies to mobile UX, should not apply to companion or extension.

#### Back action possible
Screenshot 2024-04-11 at 11 06 59

#### No back action possible

Screenshot 2024-04-11 at 11 06 37

- [X] MobileL (420px)
- [X] Tablet (656px)
- [X] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

MI-289 #done

### Preview domain
https://mi-289-fix-back-button.preview.app.daily.dev


rebelchris (approved):
This still works for squad posts though?
Meaning if you come directly on them, they should show a back which links to squad page afaik.

nensidosari (approved):
Looks good, nothing apart from @rebelchris question above, let's confirm that 🙏

packages/shared/src/components/post/GoBackHeaderMobile.tsx
```diff
@@ -24,11 +25,16 @@ export function GoBackHeaderMobile({
...
 }: PropsWithChildren<WithClassNameProps>): ReactElement {
   const router = useRouter();
+  const goHome = useCallback(() => router.push('/'), [router]);
+  const logoButton = (
+    <Logo onLogoClick={goHome} position={LogoPosition.Initial} />
+  );
+
   const canGoBack =
-    !!globalThis?.history?.length && (checkSameSite() || isDevelopment);
+    globalThis?.history?.length > 1 && (checkSameSite() || isDevelopment);
```
nensidosari, line 34: One thing actually, why are we checking here for >1 ?

kopancek: Because when you land on a page, the history is already length 1 -> the page you have just opened in the browser. So previous check did not make sense, since it would be always true.

kopancek:
> This still works for squad posts though? Meaning if you come directly on them, they should show a back which links to squad page afaik.

No, this button does not work like that today.

This is preview: https://mi-289-fix-back-button.preview.app.daily.dev/posts/good-article-on-lego-marketing-developers-love-lego-so-much-there-s-inspiration-to-be-borrowed-the-7vo5mpanw

This is production: https://app.daily.dev/posts/good-article-on-lego-marketing-developers-love-lego-so-much-there-s-inspiration-to-be-borrowed-the-7vo5mpanw

The back button in production on the squad post does nothing, though you can click on it.

We could change it so that it goes back to source, as well, but I'd rather do that in a different PR if that's the request here...

nensidosari:
Actually I think somewhere in mobile ux it was agreed there wouldn't be any different behaviour for squad posts now that u mentioned.

rebelchris:
Oh ok, it used to work like that, but I probably miss some decision somewhere from mobile ux then.

PR #2977, opened 2024-04-11
feat: cleanup source header

rebelchris:
- Minor cleanup on the tag page
- Started on the header for source pages (need API first to add tags to it, but we can ship like this so long)
- @capJavert Set StaleTime to 1 hour, since I don't expect it to change a lot/make any impact.

Screenshot 2024-04-11 at 14 54 43

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done

### DIFF
--- a/packages/shared/src/components/RecommendedTags.tsx
+++ b/packages/shared/src/components/RecommendedTags.tsx
@@ -1,0 +1,41 @@
+import React, { ReactElement } from 'react';
+import { ElementPlaceholder } from './ElementPlaceholder';
+import { TagLink } from './TagLinks';
+import type { Tag } from '../graphql/feedSettings';
+
+interface RecommendedTagsProps {
+  isLoading: boolean;
+  tags: Tag[];
+}
+export const RecommendedTags = ({
+  isLoading,
+  tags,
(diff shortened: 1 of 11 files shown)


packages/shared/src/components/RecommendedTags.tsx
```diff
@@ -0,0 +1,41 @@
...
+          <ElementPlaceholder className="h-6 w-12 rounded-8" />
+          <ElementPlaceholder className="h-6 w-12 rounded-8" />
+        </div>
+      </div>
+    );
+  }
+
+  return (
+    <>
+      {tags.length > 0 && (
```
capJavert, line 29: We can probably early return `null` in this case, will be cleaner.

capJavert (approved):
Looks good 👌

packages/shared/src/components/RecommendedTags.tsx
rebelchris: Good point, let's do it :)

PR #2983, opened 2024-04-12
hotfix: close post modal on navigation change

rebelchris:
- Looked at injecting where we close the lazyModals, but it's missing too many references to the hook, so this was the cleanest implementation and keeps it close to where it needs to be.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done

### DIFF
--- a/packages/shared/src/hooks/usePostModalNavigation.ts
+++ b/packages/shared/src/hooks/usePostModalNavigation.ts
@@ -1,4 +1,5 @@
 import { useContext, useEffect, useMemo, useRef, useState } from 'react';
+import { useRouter } from 'next/router';
 import AnalyticsContext from '../contexts/AnalyticsContext';
 import { Post, PostType } from '../graphql/posts';
 import { postAnalyticsEvent } from '../lib/feed';
@@ -31,6 +32,7 @@ export const usePostModalNavigation = (
   updatePost: UpdateFeedPost,
   canFetchMore: boolean,
 ): UsePostModalNavigation => {
+  const router = useRouter();
   const [currentPage, setCurrentPage] = useState<string>();
   const isExtension = checkIsExtension();
   const [openedPostIndex, setOpenedPostIndex] = useState<number>(null);
@@ -87,11 +89,20 @@ export const usePostModalNavigation = (
   };
 
   useEffect(() => {
+    router.events.on('routeChangeStart', onCloseModal);
+
+    return () => {
+      router.events.off('routeChangeStart', onCloseModal);
+    };
+  }, [onCloseModal, router.events]);
+
+  useEffect(() => {
     if (isExtension) {
       return null;
     }
 
     const onPopState = () => {
+      console.log('popstate trigger');
       const url = new URL(window.location.href);
       if (url.pathname.indexOf('/posts/') !== 0) {
         onCloseModal(true);


packages/shared/src/hooks/usePostModalNavigation.ts
```diff
@@ -86,12 +88,21 @@ export const usePostModalNavigation = (
     scrollPositionOnFeed.current = 0;
   };
 
+  useEffect(() => {
+    router.events.on('routeChangeStart', onCloseModal);
```
sshanzel, line 92: This might close the modal when you navigate through its header (left and right arrows). Can you check if it does not?

omBratteng: Probably testable on https://hotfix-close-post-modal-nav.preview.app.daily.dev/ 😉

capJavert: Yeah, we have to check if it is in `/posts/` right?

packages/shared/src/hooks/usePostModalNavigation.ts
```diff
@@ -86,12 +88,21 @@ export const usePostModalNavigation = (
...
+    };
+  }, [onCloseModal, router.events]);
+
   useEffect(() => {
     if (isExtension) {
       return null;
     }
 
     const onPopState = () => {
+      console.log('popstate trigger');
```
capJavert, line 105: cleanup 🙏

packages/shared/src/hooks/usePostModalNavigation.ts
rebelchris: @sshanzel valid concern, see here seems to be working as expected:

https://github.com/dailydotdev/apps/assets/554874/ef4a7438-5027-47ae-b682-d0e445383549

capJavert: Yeah, looks to be working due to using shallow history routing for back/forward arrows. Looks ok from my end 👍

capJavert (approved):

packages/shared/src/hooks/usePostModalNavigation.ts
rebelchris: Ah yes, let's add that to be safe, see small edge-case on first click

rebelchris: @sshanzel // @capJavert added the failsafe anyway :)

packages/shared/src/hooks/usePostModalNavigation.ts
```diff
@@ -86,6 +88,20 @@ export const usePostModalNavigation = (
...
+        return;
+      }
+      onCloseModal();
+    };
+    router.events.on('routeChangeStart', routeHandler);
+
+    return () => {
+      router.events.off('routeChangeStart', routeHandler);
+    };
+  }, [router.events]);
```
sshanzel, line 103: Lint might need to have the `onCloseModal` as a dependency. In that case, we might have to convert it to use `useCallback`.

rebelchris: Yep, done now :)

nensidosari (approved):

PR #2984, opened 2024-04-12
fix: tab list component not showing indicator in correct spot

kopancek:
## Changes

Fixed tab indicator not moving to the correct active tab on first click.

- changed from `useEffect` to `useLayoutEffect`, since that's more appropriate to dealing with component refs (e.g. we need to do it after render, not before)
- changed the calculation of the variables based on the ref
  - put it in useLayoutEffect as well
  - result is set into state - which makes sure the component is rerendered when it happens

### Video

https://www.loom.com/share/245f46980c1e42d5ab079b037fb1b006?sid=9458744e-9e4c-4cc0-8e2f-e86926a46f45

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [X] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

MI-298 #done

### Preview domain
https://mi-298-fix-comment-active-tabs.preview.app.daily.dev

### DIFF
--- a/packages/shared/src/components/tabs/TabList.tsx
+++ b/packages/shared/src/components/tabs/TabList.tsx
@@ -1,11 +1,22 @@
 import classNames from 'classnames';
-import React, { ReactElement, useCallback, useEffect, useRef } from 'react';
+import React, {
+  ReactElement,
+  useCallback,
+  useLayoutEffect,
+  useRef,
+  useState,
+} from 'react';
 
 interface ClassName {
   indicator?: string;
   item?: string;
 }
 
+interface DimensionProps {
+  activeTabRect?: DOMRect;
+  offset: number;
+  indicatorOffset: number;
+}
 export interface TabListProps {
   items: string[];
   active: string;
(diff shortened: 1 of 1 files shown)


packages/shared/src/components/tabs/TabList.tsx
```diff
@@ -47,12 +60,40 @@ function TabList({
...
+      : 0;
+
+    setDimensions((current) => {
+      if (
+        current.activeTabRect === activeTabRect &&
+        current.offset === offset &&
+        current.indicatorOffset === indicatorOffset
+      ) {
+        return current;
+      }
```
nensidosari, line 80: Why do we need to do this within set state, when actually we're setting the same state? We should better do this check before

kopancek: I was contemplating on this, and here's my reasoning:
1. The `dimension` state itself is an object. So to actually set the state, we need to send it a new reference, which we are doing below
3. To avoid unnecessarily setting a reference with the same information, we need to check if the current state is different than what we want to set. We have 2 options:
  a. Check the state before calling the state setter and if it's the same, do not call it. For this we would need to add the `dimension` to the `useLayoutEffect` dependencies, which would mean that this function would be called on each update of state. Which seems unnecessary.
  b. Send a callback to state setter and make the check there.
  
I decided for option b, since it seems cleaner to me from the execution point of view.

I agree, the code looks a bit unweildy this way, but it works. We could theoretically extract some of the logic to an external function to make it look nicer, but since it's just a few lines, I decided not to.

Let me know your thoughts :)

nensidosari: don't really have a strong opinion on it, so we should be good either way

nensidosari (approved):

rebelchris (approved):

PR #2978, opened 2024-04-11
feat: add feed layout without filters

nensidosari:
## Changes
- removal all feed related filters from feed layout hook
- added profile feeds on feedlayout hook to render the new cards
- adjusted some paddings and margins for the new cards on new feeds
- adjusted tests for desktop view

### Describe what this PR does
- Set the new cards in all feeds on mobile (bookmarks/squad feed/tags/sources)
- Set the new cards in profile posts & upvotes for both mobile and desktop
- We no longer open post modals in mobile/tablet in any feed

## Events

Did you introduce any new tracking events? No
Don't forget to update the [Analytics Taxonomy sheet](https://docs.google.com/spreadsheets/d/18Lv7zXges9QfVX5VYL1a-Hyl0e1sQ3sLr0OK8YZWKXI/edit#gid=0)

Log the new/changed events below:

| Type   | event_name  | value |
|--------|-------------|-------|
| Change/New | event name  | extra: { ... } |

### On those affected packages:
- [x] Have you done sanity checks in the webapp?
- [x] Have you done sanity checks in the extension?
- [ ] Does this not break anything in companion?

### Did you test the modified components media queries?
- [x] MobileL (420px)
- [x] Tablet (656px)
- [x] Laptop (1020px)

#### Did you test on actual mobile devices?
- [ ] iOS (Chrome and Safari)
- [ ] Android

MI-254 #done
MI-255 #done
MI-119 #done

### Preview domain
https://mi-254-feed-layout.preview.app.daily.dev

PR #2990, opened 2024-04-13
Mi 254 feed layout

nensidosari:
## Changes
**- Reintroduced feed filters from first [PR](https://github.com/dailydotdev/apps/pull/2978) that broke some feeds (public squads/notifications) & added the required feeds**
- added profile feeds on feedlayout hook to render the new cards
- adjusted some paddings and margins for the new cards on new feeds
- adjusted tests for desktop view

- Set the new cards in all feeds on mobile (bookmarks/squad feed/tags/sources)
- Set the new cards in profile posts & upvotes for both mobile and desktop
- We no longer open post modals in mobile/tablet in any feed

## Events

- [x] MobileL (420px)
- [x] Tablet (656px)
- [x] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

MI-254 #done
MI-255 #done
MI-119 #done

### Preview domain
https://mi-254-feed-layout.preview.app.daily.dev/


packages/shared/src/hooks/useFeedLayout.ts
```diff
@@ -1,45 +1,67 @@
...
-  FeedLayoutMobileFeedPages.has(feedName) && feedName !== SharedFeedPage.Search;
+export const UserProfileFeedPages = new Set(Object.values(UserProfileFeed));
+
+const checkShouldUseMobileFeedLayout = (
+  isLaptop: boolean,
+  feedName: AllFeedPages | UserProfileFeed | FeedPagesWithMobileLayout,
+): boolean =>
+  (!isLaptop &&
+    FeedLayoutMobileFeedPages.has(feedName as FeedPagesWithMobileLayout)) ||
+  UserProfileFeedPages.has(feedName as UserProfileFeed);
```
nensidosari, line 46: Change from last PR that broke some feeds where this was totally removed, have reintroduced this, so now it works as it was working before where everything was fine, and have added the new feeds we need in the enums

rebelchris: So this will exclude public squads and notifications basically as they are not described?

Not for now, but maybe we should have a all feeds enum and abstract portions of it to make it more clear what types of feed exist.

rebelchris (approved):
Not sure what the last one was, but it added the new layout everywhere?
Vs this one checking specific feeds?

Not urgent, but see my comment which might make it clearer down the line.

packages/shared/src/hooks/useFeedLayout.ts
nensidosari: > So this will exclude public squads and notifications basically as they are not described?

Exactly, before I removed this check totally since I didn't consider the squads public page and notifications as feed since they had different components, but they were using the same layout so that's where the issue happened.

> Not for now, but maybe we should have a all feeds enum and abstract portions of it to make it more clear what types of feed exist.

We have this at the moment [OtherFeedPages](https://github.com/dailydotdev/apps/blob/9e70f98d53203b0da37e1025cb3f853aad35b4c9/packages/shared/src/lib/query.ts#L16-L31) + [Shared feed pages ](https://github.com/dailydotdev/apps/blob/62c027157da0e8fc778044bbcda3149c4babfe31/packages/shared/src/components/utilities/common.tsx#L162-L168) which we usually use everywhere. 

And in this PR I created this for [mobileLayout](https://github.com/dailydotdev/apps/blob/7aeff0aa7442b6699a2c20733e85f2842ac61951/packages/shared/src/hooks/useFeedLayout.ts#L15-L27) pages. I couldn't extract portion of the enums unless I convert them into an array or sth, so went with this way. Let me know if it's best to extract what I need from the enums to an array to use.

nensidosari:
```
Not sure what the last one was, but it added the new layout everywhere?
Vs this one checking specific feeds?
```

Yup exactly

packages/shared/src/components/Feed.spec.tsx
```diff
@@ -223,6 +224,9 @@ it('should add one placeholder when loading', async () => {
 });
 
 it('should replace placeholders with posts and ad', async () => {
+  // is desktop
+  jest.spyOn(hooks, 'useViewSize').mockImplementation(() => true);
```
kopancek, line 228: Can we add this mock to a beforeEach or similar? Seems wasteful to repeat in every test.

packages/shared/src/hooks/useFeedLayout.ts
```diff
@@ -1,45 +1,66 @@
...
+  Search = 'search',
+  Upvoted = 'upvoted',
+  Discussed = 'discussed',
+  Bookmarks = 'bookmarks',
+  UserPosts = 'user-posts',
+  UserUpvoted = 'user-upvoted',
+  SquadPage = 'squads[handle]',
+  TagPage = 'tags[tag]',
+  SourcePage = 'sources[source]',
+}
```
kopancek, line 27: I'll repeat my concern from previous PR - we already have these constants defined elsewhere, e.g. in SharedFeedPage and OtherFeedPage enums. Can we reuse them here instead of redefining?

nensidosari: See this [comment](https://github.com/dailydotdev/apps/pull/2990#discussion_r1565441374), let me know what u think 🙏

kopancek: Technically, do we even need a separate enum type? I see it only used in `checkShouldUseMobileFeedLayout` and then in `feedNameToHeading`. Could we rely on `AllFeedPages | UserProfileFeed` type instead? 

Also, consider this non-blocking :)

kopancek (approved):

packages/shared/src/hooks/useFeedLayout.ts
nensidosari: Something like this could replace my enum:
export const SharedFeedPageArray = new Set(
SharedFeedPage
);
export const OtherFeedPageArray = new Set(
  Object.values(OtherFeedPage).filter(
    (feedPage) => feedPage !== OtherFeedPage.Author && feedPage !== OtherFeedPage.Notifications........,
  ),
);

FeedLayoutMobileFeedPages = [...SharedFeedPageArray, ...OtherFeedPageArray]

rebelchris: You could re-use the enum, but pick what you need right? (atleast there is no confusion about what enum is the main one then)
I think this last nuance is the issue here.

nensidosari: Think I can find another way, give me a sec

nensidosari: Yeah, no whatever I think of I still end up with the above filtering.

If u think that is best I can change, else I think we should be good with the new enum

rebelchris: I think it's mainly about not adding a new enum, but rather a filter that uses the existing one.
It will make it clear there's only one source (enum) of feed pages

nensidosari: Have changed this to reuse the enums :D

PR #2991, opened 2024-04-15
fix: spacing issue when sorting action is enabled on mobile

DragosIorgulescu:
## Changes

- using `sticky` makes it so the element still takes space in the dom

### Before

Screenshot 2024-04-15 at 10 23 14

### After

Screenshot 2024-04-15 at 10 22 49

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done

### DIFF
--- a/packages/shared/src/components/feeds/FeedNav.tsx
+++ b/packages/shared/src/components/feeds/FeedNav.tsx
@@ -83,7 +83,7 @@ function FeedNav(): ReactElement {
         ))}
       </TabContainer>
       {isMobile && sortingEnabled && isSortableFeed && (
-        <div className="sticky flex h-11 w-20 -translate-y-12 translate-x-[calc(100vw-100%)] items-center justify-end bg-gradient-to-r from-transparent via-background-default via-40% to-background-default pr-4">
+        <div className="fixed flex h-11 w-20 -translate-y-12 translate-x-[calc(100vw-100%)] items-center justify-end bg-gradient-to-r from-transparent via-background-default via-40% to-background-default pr-4">
           <Dropdown
             className={{ label: 'hidden', chevron: 'hidden', button: '!px-1' }}
             dynamicMenuWidth


packages/shared/src/components/feeds/FeedNav.tsx
```diff
@@ -83,7 +83,7 @@ function FeedNav(): ReactElement {
         ))}
       </TabContainer>
       {isMobile && sortingEnabled && isSortableFeed && (
-        <div className="sticky flex h-11 w-20 -translate-y-12 translate-x-[calc(100vw-100%)] items-center justify-end bg-gradient-to-r from-transparent via-background-default via-40% to-background-default pr-4">
+        <div className="fixed flex h-11 w-20 -translate-y-12 translate-x-[calc(100vw-100%)] items-center justify-end bg-gradient-to-r from-transparent via-background-default via-40% to-background-default pr-4">
```
kopancek, line 86: I don't think we even need fixed here. Maybe `absolute` is enough?

kopancek (approved):
Approving, but I think fixed is overkill for this element, since parent is already fixed.

sshanzel (changes requested):
This will make the glitchy icon return.

packages/shared/src/components/feeds/FeedNav.tsx
sshanzel: `fixed` and `absolute` won't work.

sshanzel: Can you elaborate on the primary difference, UI-wise?

sshanzel: Yeah, I've tried numerous times as well. Though from your screenshots, gradient seems to be present on both.

sshanzel (approved):
Looks great! Just not really a fan of additional wrapper since we can set it from the container. Though probably just my personal preference so approving here to unblock :shipit:

DragosIorgulescu:
> Looks great! Just not really a fan of additional wrapper since we can set it from the container. Though probably just my personal preference so approving here to unblock :shipit:

I agree, but in this case we don't have a fixed height for the header section above the tabs, we only have it for tabs, so it would require hard coding some height values which comes with it's own drawbacks

sshanzel:
> we don't have a fixed height for the header section above the tabs

I don't think its size should change in the near future, but yeah, probably safer to have the set height within the tabs only.

PR #2996, opened 2024-04-15
feat: randomise the values in acquisition cards

nensidosari:
## Changes
- installed lodash.shuffle
- shuffle acquisition options on initial state

There was a [request](https://dailydotdev.slack.com/archives/C02E2C3C13R/p1713176002777429) from product to make the values appear in a random positions in user acquisition card

## Events

- [x] MobileL (420px)
- [x] Tablet (656px)
- [x] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

MI-303 #done

### Preview domain
https://mi-303-user-acquisition.preview.app.daily.dev


packages/shared/src/lib/func.spec.ts
```diff
@@ -0,0 +1,10 @@
+import { shuffle } from './func';
+
+describe('shuffle function', () => {
+  it('should shuffle the array', () => {
+    const array = [1, 2, 3, 4, 5];
+    const arrayCopy = [...array];
+    shuffle(array);
+    expect(array).not.toEqual(arrayCopy);
```
rebelchris, line 8: Technically it could return the same as array right? (making it potentially flaky)

packages/shared/src/lib/func.ts
```diff
@@ -63,3 +63,11 @@ export enum ArrowKeyEnum {
   Down = 'ArrowDown',
   Left = 'ArrowLeft',
 }
+
+export const shuffle = (array: Array<unknown>): void => {
```
rebelchris, line 67: We could actually use `_.shuffle` to avoid building logic ourselves?

capJavert: In any case the method should return new array without mutating in place.

packages/shared/src/lib/func.spec.ts
nensidosari: Yes but what are the odds 🤔 ... any ideas to not have a flaky test? or should we remove it?

packages/shared/src/lib/func.ts
nensidosari: Yeah might as well

packages/shared/src/lib/func.spec.ts
kopancek: We could use a random seed in the `Math.random` function. That way, the test is always repeatable.

rebelchris (approved):

capJavert (approved):

PR #3005, opened 2024-04-16
fix: share post/comment on mobile

sshanzel:
## Changes
- This was primarily for fixing the post modal to allow users to see the Share modal back again.
- Fell into the refactor rabbit hole. But this was to make it easier for us to maintain the current component structure relating to ShareModal.
  - The `ShareModal` is instantiated in a lot of places that makes it hard to understand where the initialization was coming from.
  - With that, we converted it to become a `LazyModal`. The only issue was, there is a dep-cycle error after doing so. To avoid it, instead of using the `useLazyModal` for the `CreateSharePostModal`, we now initialize it here rather than initializing numerous `ShareModal`.
- Fixed a lot of type issues 😣 
- Removal of eslint-disable for dependencies.

TODO:
- ~~Test Share function on Discussion's feed.~~
- ~~Test the ShareModal on Companion (might need to be initialized somewhere).~~
- ~~Fix tests.~~
- ~~Hide behind an experiment flag.~~

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

MI-297 #done


packages/shared/src/components/Feed.tsx
```diff
@@ -302,7 +293,7 @@ export default function Feed<T>({
     row: number,
     column: number,
   ) => {
-    copyLink(post, index, row, column);
+    copyLink({ post, columns: index, row, column });
```
sshanzel, line 296: One of the benefits of using an object is this: the row was on the 3rd parameter, but it is supposed to be the column.

packages/shared/src/components/Feed.tsx
```diff
@@ -357,17 +348,17 @@ export default function Feed<T>({
...
     onShare: () =>
-      openSharePost(
+      openSharePost({
         post,
-        virtualizedNumCards,
-        postMenuLocation.row,
-        postMenuLocation.column,
-      ),
+        columns: virtualizedNumCards,
+        row: postMenuLocation.row,
```
sshanzel, line 359: One of the benefits of using an object is this: the row was on the 3rd parameter, but it is supposed to be the column.

packages/shared/src/components/history/ReadingHistoryList.tsx
```diff
@@ -87,7 +81,7 @@ export default function ReadHistoryList({
       <InfiniteScrollScreenOffset ref={infiniteScrollRef} />
       <PostOptionsReadingHistoryMenu
         post={readingHistoryContextItem?.post}
-        onShare={() => openSharePost(readingHistoryContextItem.post)}
```
sshanzel, line 90: Had to fix a lot of types from the hooks related to this as the object is being defined as `any`.

packages/shared/src/components/modals/ShareModal.tsx
```diff
@@ -67,10 +56,11 @@ export default function ShareModal({
   return (
     <Modal
       size={Modal.Size.Small}
-      kind={isMobile ? Modal.Kind.FixedBottom : Modal.Kind.FlexibleCenter}
```
sshanzel, line 70: As per Mobile UX initiative, this should now be a Drawer.

packages/shared/src/components/modals/ShareModal.tsx
```diff
@@ -81,7 +71,7 @@ export default function ShareModal({
           columns={columns}
           column={column}
           row={row}
-          onSquadShare={() => onRequestClose(null)}
+          onClose={() => onRequestClose(null)}
```
sshanzel, line 74: To be more accurate and avoid confusion since there is already an `onSquadsShare` prop inside.

packages/shared/src/components/widgets/SocialShare.tsx
```diff
@@ -106,6 +94,14 @@ export const SocialShare = ({
           emailTitle="I found this amazing post"
         />
       </SocialShareContainer>
+      {squadToShare && (
+        <CreateSharedPostModal
```
sshanzel, line 98: There is a dep-cycle issue that arises if we use `LazyModal` for this. imo, it is better to have a single instance of not using the LazyModal flow rather than numerous instances of `ShareModal`.

packages/shared/src/hooks/useSharePost.ts
```diff
@@ -1,101 +1,80 @@
...
+}
+
+export function useSharePost(origin: Origin): UseSharePost {
   const { trackEvent } = useContext(AnalyticsContext);
-  const [shareModal, setShareModal] = useState<Post>();
   const [, copyLink] = useCopyPostLink();
-  const [sharePostFeedLocation, setSharePostFeedLocation] =
-    useState<FeedItemPosition>({});
   const { getShortUrl, getTrackedUrl } = useGetShortUrl();
+  const { openModal } = useLazyModal();
```
sshanzel, line 26: Most changes here are related to migrating to `useLazyModal` over using two states to manage the modal.

packages/shared/src/components/modals/ShareModal.spec.tsx
```diff
@@ -86,9 +82,10 @@ describe('ShareModal Test Suite:', () => {
     mockWindowOpen.mockClear();
     window.open = origWindowOpen;
   });
+
   it('should render the component without logged in user', async () => {
     renderComponent(false, false);
-    expect(screen.getByText('Share post')).toBeInTheDocument();
+    expect(screen.getByText('Share with your squad')).toBeInTheDocument();
```
nensidosari, line 88: Why did this change?

packages/shared/src/components/modals/ShareModal.spec.tsx
```diff
@@ -112,8 +109,7 @@ describe('ShareModal Test Suite:', () => {
     renderComponent(true, true);
     const btn = await screen.findByTestId(`social-share-@${squads[0].handle}`);
     btn.click();
-    const modal = client.getQueryData(MODAL_KEY);
-    expect(modal.type).toEqual(LazyModal.CreateSharedPost);
```
nensidosari, line 116: Why we're not checking about these anymore?

packages/shared/src/components/post/PostActions.tsx
```diff
@@ -166,14 +169,25 @@ export function PostActions({
...
+            className="btn-tertiary-cabbage hidden tablet:flex"
           >
             Copy
           </QuaternaryButton>
+          <QuaternaryButton
+            id="share-post-btn"
+            onClick={() =>
+              globalThis?.navigator?.share
+                ? openNativeSharePost(post)
+                : openSharePost({ post })
```
nensidosari, line 185: Would it be better if we did this comparison in `useSharePost` and export `openSharePost` as either one or the other?

nensidosari (commented):
love the refactoring, a bit concerned since it's quite big so we have to test this carefully 🙏 
Left just a few comments for check, and btw, do we have a preview domain for this? would be good to test.

packages/shared/src/components/modals/ShareModal.spec.tsx
sshanzel: Some things related to mocking `useViewSize` which is not relevant here anymore.

packages/shared/src/components/modals/ShareModal.spec.tsx
sshanzel: We are mocking it differently as shown directly below this since as mentioned, the CreateSharedPost is now rendered directly.

packages/shared/src/components/post/PostActions.tsx
sshanzel: Not all cases have to be like that. Only in this implementation.

sshanzel:
> do we have a preview domain for this? would be good to test.

Let me check, the automatic setting preview domain didn't seem to apply. Will set it manually.

sshanzel:
Deployed at https://preview2.app.daily.dev/discussed

sshanzel:
> @sshanzel Just going off the screenshots you shared, in other places where we use this horizontal scroll we hide the 
scrollbar (no-scrollbar) just check with design/product if they want that here as well 🙏

Right, let me confirm this.

> The PR is still in draft, do you want me to review or wait?

Overall, should be reviewable, the only thing left is to hide this behind a feature flag.

sshanzel:
Applied the `no-scrollbar` and the feature flag. Also verified the track events are sent correctly. @rebelchris let me know what you think.

Screenshot 2024-04-17 at 7 22 31 PM

packages/shared/src/components/Feed.tsx
rebelchris: Hehe I know that's one of the reasons I prefer Object notation (think I had a discussion with @kopancek about this a while back?)

kopancek: Yes, we discussed this on a different PR, where IIRC we only had 2 params on a function. Completely agree that object notation is the way to go here, since it's more than 2 parameters :)

packages/shared/src/components/drawers/Drawer.tsx
```diff
@@ -92,7 +92,7 @@ function BaseDrawer({
     <div
       className={classNames(
         'fixed inset-0 transition-opacity duration-300 ease-in-out',
-        isFullScreen ? 'z-modal' : 'z-max bg-overlay-quaternary-onion',
+        isFullScreen ? 'z-modal' : 'z-modal bg-overlay-quaternary-onion',
```
kopancek, line 95: `z-modal` is used in both cases. I think we can add `z-modal` to the line above, and change this one to achieve the same functionality. Seems shorter to me:
```suggestion
        !isFullScreen && 'bg-overlay-quaternary-onion',
```

nensidosari (approved):
Looks good to me

packages/shared/src/lib/featureManagement.ts
```diff
@@ -45,6 +45,7 @@ const feature = {
   readingReminder: new Feature('reading_reminder', false),
   tagsOnCard: new Feature('tags_on_card', false),
   onboardingMostVisited: new Feature('onboarding_most_visited', false),
+  shareExperience: new Feature('share_experience', false),
```
nensidosari, line 48: https://app.growthbook.io/features/share_experience

packages/shared/src/components/modals/ShareModal.tsx
```diff
@@ -26,10 +16,9 @@ export default function ShareModal({
   row,
   onRequestClose,
   ...props
-}: ShareModalProps): ReactElement {
+}: ShareModalProps & ModalProps): ReactElement {
```
kopancek, line 20: Can we instead change the `ShareModalProps` interface to also extend `ModalProps`?

packages/shared/src/components/post/PostActions.tsx
```diff
@@ -166,14 +172,30 @@ export function PostActions({
             Bookmark
           </QuaternaryButton>
           <QuaternaryButton
-            id="share-post-btn"
+            id="copy-post-btn"
```
kopancek, line 175: Do we even need to specify the `id` param here?

packages/shared/src/components/post/PostActions.tsx
```diff
@@ -166,14 +172,30 @@ export function PostActions({
...
             onClick={() => onCopyLinkClick(post)}
             icon={<LinkIcon />}
             responsiveLabelClass={actionsClassName}
-            className="btn-tertiary-cabbage"
+            className={classNames(
+              'btn-tertiary-cabbage',
```
kopancek, line 180: IIRC we should be able to send `variant` and `color` props instead. Feel free to ignore if my memory fails me 😊

packages/shared/src/components/post/PostActions.tsx
```diff
@@ -166,14 +172,30 @@ export function PostActions({
...
           >
             Copy
           </QuaternaryButton>
+          {shareExperience && (
+            <QuaternaryButton
+              id="share-post-btn"
+              onClick={() =>
+                globalThis?.navigator?.share
+                  ? openNativeSharePost(post)
+                  : openSharePost({ post })
```
kopancek, line 192: Looking at this code, my refactoring urge is growing. `openSharePost` accepts an object, but `openNativeSharePost` accepts an argument instead.

Also, since both `openSharePost` and `openNativeSharePost` come from the `useSharePost` hook, maybe the hook can provide a single function instead and move the `globalThis?.navigator?.share` logic into the hook itself?

packages/shared/src/components/widgets/SocialShare.tsx
```diff
@@ -72,27 +71,16 @@ export const SocialShare = ({
   const onSharedSuccessfully = () => {
     trackEvent(postAnalyticsEvent(AnalyticsEvent.ShareToSquad, post));
 
-    if (onSquadShare) {
-      onSquadShare();
+    if (onClose) {
+      onClose();
     }
```
kopancek, line 78: We could do `onClose?.();` in typescript instead to convert this to a 1-liner.

packages/shared/src/components/widgets/SocialShare.tsx
kopancek: Maybe add a comment into the code as well, so that we know why it's done like this? :)

kopancek (commented):
Looks great. Added a few small comments, let me know what you think.

packages/shared/src/components/post/PostActions.tsx
sshanzel: Required by the component itself (legacy) 😅

packages/shared/src/components/post/PostActions.tsx
sshanzel: Oh, right. Didn't notice this. Lol. Thank you.

packages/shared/src/components/drawers/Drawer.tsx
sshanzel: Right. Yes, this is definitely better.

packages/shared/src/components/modals/ShareModal.tsx
sshanzel: The `ShareModalProps` is used somewhere and that part doesn't need the `ModalProps`. Maybe I should change the name to something else as it implies Modal and must have Modal properties 🤔

packages/shared/src/components/widgets/SocialShare.tsx
sshanzel: Used to do that in the past, but I received comments about having it like this (even though what you suggest is basically this and also what I prefer). Let me write these cases down in a note to have it discussed on a bigger forum in the future.

packages/shared/src/components/post/PostActions.tsx
sshanzel: Good point. Yeah, we can probably abstract this.

packages/shared/src/components/widgets/SocialShare.tsx
sshanzel: Makes sense. Let me add that.

kopancek (approved):
Nothing to add :shipit: 🔥

PR #3012, opened 2024-04-17
feat: promote apps after bookmarking

sshanzel:
## Changes
- Created a generic modal for doing promotions similar to this since they are all the same in terms of layout.
- Refactored the streaks migration popup to utilize the generic popup.

Preview:
Screenshot 2024-04-17 at 10 06 10 PM

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

MI-312 #done

### Preview domain
https://mi-312.preview.app.daily.dev


packages/shared/src/components/Feed.tsx
```diff
@@ -368,9 +383,6 @@ export default function Feed<T>({
         postMenuLocation.row,
         postMenuLocation.column,
       ),
-    onBookmark: () => {
```
sshanzel, line 371: Not used anymore (post context menu bookmark).

rebelchris:
Where did this come from? Its basically marketing cta we just build from the looks of it 👀 @omBratteng what do you think?

omBratteng:
> Where did this come from? Its basically marketing cta we just build from the looks of it 👀 @omBratteng what do you think?

Yeah 😬

packages/shared/src/components/modals/generic/common.ts
```diff
@@ -0,0 +1,25 @@
...
+    },
+    image: cloudinary.streak.migrate,
+    cta: { copy: 'Tell me more', link: migrateUserToStreaks },
+  },
+  bookmarkPromoteMobile: {
+    pill: { copy: 'Mobile version' },
+    title: 'Get back to your bookmarks on the go',
+    description:
+      'Your saved posts are waiting for you on daily.dev mobile. Perfect for reading anytime, anywhere.',
+    image: '',
```
sshanzel, line 22: Waiting for the image to get uploaded to cloudinary.

sshanzel:
> Where did this come from? Its basically marketing cta we just build from the looks of it 👀 @omBratteng what do you think?

Not sure I understand what's going on 😅 

Ticket are:
- https://dailydotdev.atlassian.net/jira/software/projects/MI/boards/31?selectedIssue=MI-312
- https://dailydotdev.atlassian.net/jira/software/projects/MI/boards/31?selectedIssue=MI-296

rebelchris:
@sshanzel Ah maybe it needs to launch pre the integration, but this is typically what we build the marketing CTA for.
I do think you can also re-use that modal instead of bringing another one that's exactly the same.

sshanzel:
> @sshanzel Ah maybe it needs to launch pre the integration, but this is typically what we build the marketing CTA for.
I do think you can also re-use that modal instead of bringing another one that's exactly the same.

I see. Yeah, probably a task that was made before the marketing CTA. Let me check how we can use that here.

rebelchris (commented):
@sshanzel Another thing missing here is that bookmark loops is an experiment and should have specific evaluation for enrollment.

sshanzel:
> @sshanzel Another thing missing here is that bookmark loops is an experiment and should have specific evaluation for enrollment.

Should now be added. Only thing missing is uploading the [asset to cloudinary](https://dailydotdev.slack.com/archives/C06Q7GYTBHN/p1713358467149559).

packages/shared/src/lib/constants.ts
```diff
@@ -40,6 +40,8 @@ export const searchFeedback = 'https://r.daily.dev/search-feedback';
 export const searchDocs = 'https://r.daily.dev/search-docs';
 export const migrateUserToStreaks = 'https://r.daily.dev/streaks';
 
+export const dailyDevApps = 'https://daily.dev/apps';
```
rebelchris, line 43: This is already a rebrandly link afaik.
If unclear check with nimrod but we decided to only use them in case marketing ever wants to change 👍

packages/shared/src/components/MainFeedLayout.tsx
```diff
@@ -302,7 +303,10 @@ export default function MainFeedLayout({
       return;
     }
 
-    openModal({ type: LazyModal.MigrateUserStreak });
+    openModal({
```
rebelchris, line 306: This one is not a experiment?

packages/shared/src/components/Feed.tsx
```diff
@@ -207,7 +213,7 @@ export default function Feed<T>({
 
   const useList = insaneMode && numCards > 1;
   const virtualizedNumCards = useList ? 1 : numCards;
-
+  const bookmarkLoops = useFeature(feature.bookmarkLoops);
```
rebelchris, line 216: This will enroll users that dont bookmark, can you check with Nensi/ido what the rule is?

rebelchris (commented):
Wow love the approach 👏👏👏
Have some minor points we need to validate but dont keep me as blocker 👍

packages/shared/src/components/MainFeedLayout.tsx
sshanzel: It is, but the condition above considers the check for that. Should be safe.

packages/shared/src/components/Feed.tsx
sshanzel: Right. Very good point. Let me check.

nensidosari: As Ido mentioned in thread, only users who bookmark, for this Lee u can use this hook @rebelchris created `useConditionalFeature`, u might need a way to know though that keeps track if a user bookmarked after this is merged, maybe actions?

packages/shared/src/components/MainFeedLayout.tsx
nensidosari: Why do we have it on Feed and MainFeedLayout as well?

packages/shared/src/components/modals/generic/common.ts
```diff
@@ -0,0 +1,39 @@
+import { cloudinary } from '../../../lib/image';
+import { dailyDevApps } from '../../../lib/constants';
+import {
+  MarketingCta,
+  MarketingCtaVariant,
+} from '../../cards/MarketingCta/common';
+
+export const promotion: Record<string, MarketingCta> = {
+  migrateStreaks: {
```
nensidosari, line 9: nice

packages/shared/src/components/MainFeedLayout.tsx
sshanzel: They are two different promotions. The one on Feed is the Bookmark loops, this one is `promotion.migrateStreaks`.

packages/shared/src/hooks/useConditionalFeature.ts
```diff
@@ -33,13 +41,20 @@ export const useConditionalFeature = <T extends JSONValue>({
...
-      return growthbook
+
+      const value = await (growthbook
         ? growthbook.getFeatureValue(feature.id, feature.defaultValue)
-        : (feature.defaultValue as WidenPrimitives<T>);
+        : Promise.resolve(feature.defaultValue as WidenPrimitives<T>));
+
+      if (onAfterEvaluation) {
+        onAfterEvaluation(value);
+      }
```
sshanzel, line 55: Introduced this so we won't have to introduce a `useEffect` whenever we need to do something after the evaluation. For example, the modal would have to be called in a `useEffect` if we didn't have this.

packages/shared/src/components/cards/MarketingCta/common.tsx
```diff
@@ -22,8 +22,8 @@ export enum MarketingCtaVariant {
 }
 
 export interface MarketingCta {
-  campaignId: string;
-  createdAt: Date;
+  campaignId?: string;
+  createdAt?: Date;
```
sshanzel, line 26: Made this optional to understand whether the promotion came from the marketing team so we can apply the necessary tracking only if applicable.

sshanzel: Thoughts? cc @rebelchris @omBratteng @nensidosari

packages/shared/src/components/modals/MarketingCtaModal.tsx
```diff
@@ -23,19 +23,23 @@ export const MarketingCtaModal = ({
...
     param: React.MouseEvent<Element, MouseEvent> | React.KeyboardEvent<Element>,
     eventName: AnalyticsEvent,
   ) => {
-    trackEvent({
-      event_name: eventName,
-      target_type: TargetType.MarketingCtaPopover,
-      target_id: marketingCta.campaignId,
-    });
-    clearMarketingCta(marketingCta.campaignId);
+    if (campaignId) {
```
capJavert, line 34: since previously we were sending no matter `camaignId` can you confirm it is expected to only send in this case?

packages/shared/src/hooks/useConditionalFeature.ts
```diff
@@ -33,13 +41,20 @@ export const useConditionalFeature = <T extends JSONValue>({
...
+    async () => {
       if (!shouldEvaluate) {
         return feature.defaultValue as WidenPrimitives<T>;
       }
-      return growthbook
+
+      const value = await (growthbook
         ? growthbook.getFeatureValue(feature.id, feature.defaultValue)
-        : (feature.defaultValue as WidenPrimitives<T>);
+        : Promise.resolve(feature.defaultValue as WidenPrimitives<T>));
```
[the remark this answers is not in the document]
sshanzel: Let me check.

packages/shared/src/components/Feed.tsx
```diff
@@ -149,6 +160,27 @@ export default function Feed<T>({
   const { getMarketingCta } = useBoot();
   const marketingCta = getMarketingCta(MarketingCtaVariant.Card);
   const showMarketingCta = !!marketingCta;
+  const { openModal } = useLazyModal();
+  const { completeAction, checkHasCompleted } = useActions();
+  const seenBookmarkPromotion = useMemo(
+    () => checkHasCompleted(ActionType.BookmarkPromoteMobile),
+    [checkHasCompleted],
```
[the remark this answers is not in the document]
sshanzel: Yeah makes sense. Better to be 100% accurate.

packages/shared/src/hooks/useConditionalFeature.ts
capJavert: Example:
- hook A mounts to check the value
- query resolves but onAfterEvaluation is not called since hook A does not need it
- hook B mounts to check the value with onAfterEvaluation
- query was already resolved so no onAfterEvaluation is called
- hook B logic never executes

sshanzel: I think it would still execute since hook B won't have the correct value unless it goes through that code. Are we probably missing an early return of the value (client.getData) if it has been fetched already? Which would then make the scenario you mentioned, valid?

capJavert: > I think it would still execute since hook B won't have the correct value unless it goes through that code

I think not, `useQuery` will just return cached value per query key which would be the same for both hooks.

capJavert: > Are we probably missing an early return of the value (client.getData) if it has been fetched already? Which would then make the scenario you mentioned, valid?

Again, if this happens inside `queryFn` it would still not execute because useQuery would just use cached value without calling a function.

capJavert: With your comment of useEffect, when would you theoretically want `onAfterEvaluation` to execute? Any time the feature value changes or only once per hook call?

sshanzel: > I think not, useQuery will just return cached value per query key which would be the same for both hooks.

Ah, yes. That makes sense.

> With your comment of useEffect, when would you theoretically want onAfterEvaluation to execute? Any time the feature value changes or only once per hook call?

In this case, it is probably better to do a useEffect for now since it is more predictable to say whenever the value pulled is correct, we should run the function. Open to suggestions, but for now, I will have to resort to `useEffect`, since I believe the `onSuccess` for `useQuery` will soon be obsolete.

capJavert: Think so yes, only alternative that we can look is that if we can toggle modal from `onCardBookmark`?

capJavert: We can, so instead of calling `setJustBookmarked(true);` in `onCardBookmark` we call `growthbook.getFeatureValue` there and then we can toggle modal depending on returned value then and there. 

So in `onCardBookmark`:
```js
if (!post.bookmarked && !seenBookmarkPromotion && !justBookmarked) {
  const featureValue = growthbook.getFeatureValue(feature.bookmarkLoops)

  if (featureValue) {
    completeAction(ActionType.BookmarkPromoteMobile);
    openModal({
      type: LazyModal.MarketingCta,
      props: { marketingCta: promotion.bookmarkPromoteMobile },
    });
  }
}
```

packages/shared/src/components/cards/MarketingCta/common.tsx
nensidosari: Yeah I think that's good

sshanzel: Checked with product/ido. They want to have a campaign id for each.

packages/shared/src/hooks/useConditionalFeature.ts
sshanzel: Yeah, that's true, we can do this. My only concern is the usage of `growthbook.getFeatureValue` is not often how we enroll to features. It might have some confusion on which to use and when for others in the future, but yeah, for me, that works alright. If everyone agrees, then I'll be cool with it.

rebelchris: I think either you should use Antes component here or do the eval top level.
This hook should be as lean as possible (just my 2 cents without context )

Ante should your component not do with 2 buttons one with the exp one not?

capJavert: We can do it in multiple ways yes, but I think enrolment wise it would be most accurate to enrol user when he actually bookmarks and not anytime he sees bookmark button?

sshanzel: > We can do it in multiple ways yes, but I think enrolment wise it would be most accurate to enrol user when he actually bookmarks and not anytime he sees bookmark button?

Yes, that's why we are trying to only decide to enroll when bookmark is executed. In that case, I will change the implementation to use `growthbook.getFeatureValue`.

rebelchris: Hmm yeah, should then be as simple as direct eval.
Although maybe youll need another wrapper to avoid too many deps

omBratteng (commented):
Does this fire the `clearUserMarketingCta` mutation to API if a user closes the modal?

sshanzel:
> Does this fire the clearUserMarketingCta mutation to API if a user closes the modal?

Yes, it would. Should we avoid executing that?

packages/shared/src/lib/constants.ts
sshanzel: Updated 🚀

packages/shared/src/hooks/useConditionalFeature.ts
sshanzel: Deps are looking good.

Should now be updated as to what was discussed :shipit:

packages/shared/src/hooks/useConditionalFeature.ts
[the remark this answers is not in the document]
sshanzel: Reverted to the hook's changes.

packages/shared/src/components/Feed.tsx
```diff
@@ -347,6 +358,22 @@ export default function Feed<T>({
       },
     });
 
+    if (!!user && !post.bookmarked && !seenBookmarkPromotion) {
+      const bookmarkLoops = await growthbook.getFeatureValue(
```
rebelchris, line 362: Not super blocking, but is there anyway we can move this as a function inside the GB provider or lib?
It would make it easier to not have to import multiple dependencies for each case.

rebelchris (approved):
Non blocking comment :)

packages/shared/src/components/Feed.tsx
rebelchris: @sshanzel If we can easily avoid calling the clear function (maybe something like if it's a app defined CTA, so part of the object you made) don't fire it?

sshanzel: Yeah, I've thought of that too, also that we already have the default value at the same time in the featureManagement. Would at least avoid mistakes of putting a different default value.

> @sshanzel If we can easily avoid calling the clear function (maybe something like if it's a app defined CTA, so part of the object you made) don't fire it?

Yeah we can do that, should be a simple find and we should be good.

packages/shared/src/components/modals/MarketingCtaModal.tsx
sshanzel: Confirmed, we still need it here. Reverted this change.

packages/shared/src/lib/featureManagement.ts
```diff
@@ -16,11 +16,8 @@ export class Feature<T extends JSONValue> {
 const feature = {
   feedVersion: new Feature('feed_version', 15),
   lowImps: new Feature('feed_low_imps'),
-  bookmarkOnCard: new Feature('bookmark_on_card', false),
-  readingStreaks: new Feature(
-    'reading_streaks',
-    ReadingStreaksExperiment.Control,
-  ),
+  bookmarkOnCard: new Feature('bookmark_on_card', true),
+  readingStreaks: new Feature('reading_streaks', ReadingStreaksExperiment.V2),
```
nensidosari, line 20: Let's not change this from control, we need it for existing streaks user experiment

packages/shared/src/lib/featureManagement.ts
```diff
@@ -16,11 +16,8 @@ export class Feature<T extends JSONValue> {
 const feature = {
   feedVersion: new Feature('feed_version', 15),
   lowImps: new Feature('feed_low_imps'),
-  bookmarkOnCard: new Feature('bookmark_on_card', false),
-  readingStreaks: new Feature(
-    'reading_streaks',
-    ReadingStreaksExperiment.Control,
-  ),
+  bookmarkOnCard: new Feature('bookmark_on_card', true),
```
nensidosari, line 19: I'd rather we keep this as is until we cleanup

packages/shared/src/lib/featureManagement.ts
```diff
@@ -35,6 +32,7 @@ const feature = {
   readingReminder: new Feature('reading_reminder', false),
   onboardingMostVisited: new Feature('onboarding_most_visited', false),
   shareExperience: new Feature('share_experience', false),
+  bookmarkLoops: new Feature('bookmark_loops', true),
```
nensidosari, line 35: Why is default value true here?

nensidosari (changes requested):
All looks good except some experiment values 🙏  please check

packages/shared/src/lib/featureManagement.ts
sshanzel: Ah, my bad. Was testing. Accidentally pushed.

packages/shared/src/lib/featureManagement.ts
sshanzel: Ah, my bad. Was testing before merging. Accidentally pushed.

nensidosari (approved):
Looks good

capJavert (commented):
Some minor questions on feature implementation, overall almost ready 💪

packages/shared/src/lib/constants.ts
```diff
@@ -64,3 +62,8 @@ export const authUrl =
 export const heimdallUrl = isDevelopment
   ? process.env.NEXT_PUBLIC_HEIMDALL_URL || 'http://127.0.0.1:3000'
   : authUrl;
+
+export const rebrandly = {
+  bookmarkLoops: 'https://r.daily.dev/bookmarkloops',
+  migrateUserToStreaks: 'https://r.daily.dev/streaks',
+};
```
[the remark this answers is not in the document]
sshanzel: This was discussed before, and we should probably do it similarly to image links, just so we will have to import one variable from the file. Either way works for me though.

packages/shared/src/components/GrowthBookProvider.tsx
```diff
@@ -26,12 +26,18 @@ import { useRequestProtocol } from '../hooks/useRequestProtocol';
...
+) => WidenPrimitives<T>;
+
 export type FeaturesReadyContextValue = {
   ready: boolean;
+  getFeatureValue: GetFeatureValue;
 };
 
 export const FeaturesReadyContext = createContext<FeaturesReadyContextValue>({
   ready: false,
+  getFeatureValue: () => null,
```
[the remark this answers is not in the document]
sshanzel: Good point. Let's do that.

rebelchris (approved):
Love it!
Thanks for the addition in GB

packages/shared/src/lib/constants.ts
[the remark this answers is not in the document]
capJavert: Code splitting wise it would be worse since the whole object would then be pulled all over the place. Non blocking I just wanted to make it as it is currently.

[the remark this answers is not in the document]
sshanzel: I see. In that case, it is maybe better to keep them plain as you mentioned.

sshanzel:
@capJavert fixed the comments. Will be merging now as I am planning to use what we introduced here to another PR.

PR #3013, opened 2024-04-17
feat: add h1 elements for seo

kopancek:
## Changes

- Converted some HTML elements into `h1`
- Search displays as h1 in sidebar if it's active
- See https://dailydotdev.atlassian.net/browse/MI-299 for more details

- [X] MobileL (420px)
- [X] Tablet (656px)
- [X] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

MI-299 #done

### Preview domain
https://mi-299-seo-h1.preview.app.daily.dev


sshanzel (approved):
Looks good to me 🚀

packages/shared/src/components/search/SearchHistory.tsx
```diff
@@ -40,6 +40,7 @@ export function SearchHistory({
       <span className="font-bold text-text-quaternary typo-footnote">
         {title}
       </span>
+      <h1 className="hidden">Search History</h1>
```
rebelchris, line 43: Dont bots ignore hidden text? Pretty sure google bot does from what I remember.

rebelchris (approved):
Non blocking comment, but since its for SEO purposes maybe good to relay back 👍

packages/shared/src/components/search/SearchHistory.tsx
kopancek: Yeah, you're right :/ back to the drawing board...

kopancek: I decided to implement it in a different way. For feeds, the active item in sidebar is shown as h1. For some other components, the active tab is displayed as h1.

This should get rid of all of the `<h1 className="hidden"` cases.

rebelchris: Not the biggest fan tbh.
since these are only behind auth maybe we should just omit the h1 there?

It won't really do anything anyway.

kopancek: That's a valid point. The search itself is not behind auth at the moment though. Should we treat it the same? 🤔

kopancek: OK, removed most of my changes. Kept the one for search, until we decide what to do with it, since now you can go to `/search` even if not authenticated.

kopancek: I'm going to merge this PR, we can decide what to do with the search path in a followup.

PR #3019, opened 2024-04-18
fix: spamming voting actions

sshanzel:
## Changes
- Spamming becomes an issue because we do optimistic update.
- Disable the buttons when upvoting to avoid spam.
- I kinda hate that we have to update at least 8 files to update a single prop. An option would probably subscribe to the mutation and check the status from there? Thoughts?

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

MI-228 #done

### Preview domain
https://mi-228.preview.app.daily.dev


rebelchris (approved):
Think mainly the lean feed iniative deals a lot with this.
It abstract logic to it's own component vs prop drilling so much, so for now this is fine I think.

packages/shared/src/components/Feed.tsx
```diff
@@ -417,6 +417,7 @@ export default function Feed<T>({
             onCommentClick={onCommentClick}
             onAdClick={onAdClick}
             onReadArticleClick={onReadArticleClick}
+            isVoting={isVoting}
```
capJavert, line 420: Do we have to do this and expose `isVoting`? Maybe as alternative approach we can just early return from mutation inside `useVote` without any user interactio. I would say spamming is not a common action so it is fine that it has no UI feedback. Also having disabled feedback kinda beats the purpose of optimistic UI?

This would also avoid a bunch of changes all over codebase.

sshanzel: That makes sense. Will check this in a moment.

rebelchris (approved):

PR #3020, opened 2024-04-18
feat: add multiple feeds to source page

rebelchris:
- Added multiple feeds and a new horizontal wrapper
- After some discussions decided to abuse the current feeds as a lot of logic for actions/modals etc already lives here
- Ended up doing a wrapper just to abstract some logic
- Lean feed technically would make a lot of this work redundant, but for the time being this would do

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done

### Preview domain
https://as-247-source-multiple-feeds.preview.app.daily.dev


packages/shared/src/components/feeds/HorizontalFeed.tsx
```diff
@@ -0,0 +1,107 @@
...
+              if (feedContainerRef.current) {
+                const currentPosition = feedContainerRef.current.scrollLeft;
+                const newPosition = Math.max(
+                  0,
+                  currentPosition - numCards * 320,
+                );
+                feedContainerRef.current.scrollLeft =
+                  newPosition < 150 ? 0 : newPosition;
+              }
+            }}
```
nensidosari, line 69: Probably best in a separate func

packages/shared/src/components/feeds/HorizontalFeed.tsx
```diff
@@ -0,0 +1,107 @@
...
+            disabled={feedScrolledPosition > (10 - numCards) * 320}
+            onClick={() => {
+              if (feedContainerRef.current) {
+                const currentPosition = feedContainerRef.current.scrollLeft;
+                feedContainerRef.current.scrollLeft = Math.max(
+                  0,
+                  currentPosition + numCards * 320,
+                );
+              }
+            }}
```
nensidosari, line 84: same here

packages/shared/src/hooks/useFeedLayout.ts
```diff
@@ -59,12 +62,16 @@ const checkShouldUseMobileFeedLayout = (
...
 export const useFeedLayout = ({
   feedRelated = true,
+  originFeedName,
 }: UseFeedLayoutProps = {}): UseFeedLayoutReturn => {
   const isLaptop = useViewSize(ViewSize.Laptop);
   const { feedName } = useActiveFeedNameContext();
-
+  const evaluateFeedName =
+    originFeedName === OtherFeedPage.SourceMostUpvoted
+      ? originFeedName
```
nensidosari, line 71: Hmm, how we were using this before is that when we would pass the feedname it would override the feedName from context, no matter what feed it was.

Personally think it makes more sense like that, any chance we can pass this prop only when feed is SourceMostUpvoted?

nensidosari: This way and also from the changes above looks like we are passing this prop almost everywhere, which gives also the impression this is a required prop. 

Since we also have a route for this from what I see below `sources[source]/most-upvoted` why can't we evaluate like we do for others?

nensidosari (commented):
All other code seem good, happy to approve, just concerned on the new feed layout additions, feel like we can do better

packages/shared/src/hooks/useFeedLayout.ts
rebelchris: Ah yes can do like that indeed 👍

nensidosari: Maybe it could be a combination of `sources[source]/most-upvoted` + any additional prop we can take for the card ones which are not going to change to new layout

packages/shared/src/hooks/useOnPostClick.ts
```diff
@@ -98,7 +98,10 @@ export default function useOnPostClick({
   const { trackEvent } = useContext(AnalyticsContext);
   const { incrementReadingRank } = useIncrementReadingRank();
   const { queryKey: feedQueryKey, items } = useContext(ActiveFeedContext);
-  const { shouldUseMobileFeedLayout } = useFeedLayout({ feedRelated: false });
+  const { shouldUseMobileFeedLayout } = useFeedLayout({
+    feedRelated: false,
+    originFeedName: feedName as AllFeedPages,
```
nensidosari, line 103: feedRelated overrules originFeedName, so originFeedName is not needed here

packages/shared/src/hooks/useFeedLayout.ts
nensidosari: Btw if this way is staying, more additional feednames need to be added on other places like `CardCover` ex.
Screenshot 2024-04-19 at 3 48 01 PM
Causes this shift in cover

nensidosari (commented):
Like it better with the new changes for components, much clear. 🙌 
Few more thoughts added, non blocking, just what I think is better.

packages/shared/src/components/Feed.tsx
```diff
@@ -139,7 +147,9 @@ export default function Feed<T>({
   const insaneMode = !forceCardMode && listMode;
   const numCards = currentSettings.numCards[spaciness ?? 'eco'];
   const isSquadFeed = feedName === OtherFeedPage.Squad;
-  const { shouldUseMobileFeedLayout } = useFeedLayout();
+  const { shouldUseMobileFeedLayout } = useFeedLayout({
+    originFeedName: feedName,
```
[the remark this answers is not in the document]
rebelchris: I decided not to do it like that as it would require a lot of logic in many places.
Now you only see 1, but the actual is 4 types of feeds which require quite a complex thing, and easy to miss one, vs the hook handling the right case always.

packages/shared/src/hooks/useFeedLayout.ts
```diff
@@ -59,12 +64,17 @@ const checkShouldUseMobileFeedLayout = (
 
 export const useFeedLayout = ({
   feedRelated = true,
+  originFeedName,
 }: UseFeedLayoutProps = {}): UseFeedLayoutReturn => {
   const isLaptop = useViewSize(ViewSize.Laptop);
   const { feedName } = useActiveFeedNameContext();
```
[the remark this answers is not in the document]
rebelchris: Makes sense, can do that :)

packages/shared/src/components/Feed.tsx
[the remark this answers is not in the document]
nensidosari: Cool, we keep it as is then 🙌 
If I can also suggest we call this feedName and rename in the hook as we do for contextFeedName?

packages/shared/src/hooks/useFeedLayout.ts
```diff
@@ -59,12 +64,17 @@ const checkShouldUseMobileFeedLayout = (
 
 export const useFeedLayout = ({
   feedRelated = true,
+  originFeedName,
```
capJavert, line 67: I am not sure about this, I would much rather have it overwritten by mounting active feed context provider.

packages/shared/src/components/Feed.tsx
```diff
@@ -176,6 +186,7 @@ export default function Feed<T>({
       },
     },
   );
+  const canFetchMore = allowFetchMore ?? queryCanFetchMore;
```
capJavert, line 189: Might be missing something, but from usage, if query `canFetchMore` is false why would we overwrite it since there should be no more pages to fetch?

packages/shared/src/components/Feed.tsx
```diff
@@ -69,6 +69,10 @@ export interface FeedProps<T>
   showSearch?: boolean;
   actionButtons?: ReactNode;
   disableAds?: boolean;
+  allowFetchMore?: boolean;
+  forcedLimit?: number;
```
capJavert, line 73: Just name it `pageSize`?

capJavert (commented):
Overall nothing blocking, good work 🙌 , maybe we can get away with less additional variables for Feed.

packages/shared/src/components/Feed.tsx
rebelchris: Technically there are (eg api will say yes more pages)

packages/shared/src/hooks/useFeedLayout.ts
rebelchris: The issue is the context doesn't exist everywhere and doesn't wrap everything.
Think it's half the issue with the router thing going on now.

I didn't want to make this a full refactor which I think someone even tried before.

rebelchris: Or do you mean by using feedContext inside this hook to get this value?

capJavert: Yes, basically can we get away without adding this parameter?

rebelchris: Yeah good point, let me try and refactor that part :)

rebelchris: Need to check if we have access to all context from everywhere

packages/shared/src/components/Feed.tsx
capJavert: Does `hasNextPage` not mean anything then? Maybe we can instead expose `getNextPageParam` and `getPreviousPageParam`, it will make it more versatile in the future.

packages/shared/src/hooks/useFeedLayout.ts
nensidosari: Not sure if I'm getting this right, but don't think it would work since the context would need to be different in the same page for different feeds.

packages/shared/src/components/Feed.tsx
rebelchris: They different things basically since we didn't want to actually load 2 posts.
So the FE shows only 2 posts with it's own pagination, but internally loads all 10 pages at once (As a optimization)

Keep in mind ideally this whole thing would have been it's own component, but we decided to re-use feed since it has all the card logic

Hope that makes sense?

capJavert: > So the FE shows only 2 posts with it's own pagination, but internally loads all 10 pages at once (As a optimization)

You mean on feed/api side cache, frontend loads page per page as far as I know?

rebelchris: I'll show you in our 1:1 quickly it's a bit difficult to explain :)

packages/shared/src/hooks/useFeedLayout.ts
rebelchris: I've pushed these changes.
Just a bit afraid naming might be a bit complicated as we have to deal with 2 feedNames 🥲 (it's because in some places the context wouldn't exist)

@capJavert // @nensidosari let me know what you think (I do agree it's cleaner)

nensidosari: Yeah definitely cleaner, but still complicated and can easily be misinterpreted. I think we moved to the feedName context for feed layout to unify them. 

What if we add a new feedName - named differently for this case to differentiate, but would still be within ActiveFeedNameContext. We would just need to add a modifier there as well for this feedName, but it would just keep us within the same context.

capJavert: Was gonna say the same thing for `ActiveFeedNameContext`, sounds better there.

nensidosari: Yeah I think it makes sense and might be the best solution of all!

rebelchris: Oh wait the issue lies in `ActiveFeedNameContext` always being higher then `ActiveFeedContext` so it will result in being empty always 🤣

Gosh I really hate this feed system setup we have.

packages/shared/src/components/Feed.tsx
```diff
@@ -176,12 +184,14 @@ export default function Feed<T>({
...
   );
+  const canFetchMore = allowFetchMore ?? queryCanFetchMore;
   const feedContextValue = useMemo(() => {
     return {
+      feedName,
       queryKey: feedQueryKey,
       items,
     };
-  }, [feedQueryKey, items]);
+  }, [feedName, feedQueryKey, items]);
```
nensidosari, line 194: We don't seem to need this change anymore

packages/shared/src/components/feeds/HorizontalFeed.tsx
```diff
@@ -0,0 +1,112 @@
...
+            icon={<ArrowIcon className="rotate-90" />}
+            disabled={feedScrolledPosition > (10 - numCards) * 320}
+            onClick={onClickNext}
+            aria-label="Scroll right"
+          />
+        </div>
+      </div>
+      <ActiveFeedNameContext.Provider
+        value={{ feedName: OtherFeedPage.SourceMostUpvoted }}
+      >
```
capJavert, line 92: I would keep it outside of `HorizontalFeed` since then it always has value of `SourceMostUpvoted` but component seems generic. But I guess you still need to cleanup the queries?

rebelchris: Ah yes can do it inside the source page that wraps the horizontal feed :) 
I'll start working on the actual queries next and finalize this PR

packages/shared/src/contexts/ActiveFeedContext.tsx
```diff
@@ -1,7 +1,9 @@
 import { createContext } from 'react';
 import { FeedReturnType } from '../hooks/useFeed';
+import { AllFeedPages } from '../lib/query';
 
 export type ActiveFeedContextValue = {
+  feedName?: AllFeedPages;
```
capJavert, line 6: We don't need this one since we are using `ActiveFeedNameContext`?

capJavert (approved):
Nice work!

packages/shared/src/components/feeds/FeedContainer.tsx
```diff
@@ -33,6 +34,9 @@ export interface FeedContainerProps {
   showSearch?: boolean;
   shortcuts?: ReactNode;
   actionButtons?: ReactNode;
+  isHorizontal?: boolean;
+  feedContainerRef?: React.RefObject<HTMLDivElement>;
+  feedName?: AllFeedPages;
```
nensidosari, line 39: Do we still need this?

nensidosari (approved):
Looks great!

packages/shared/src/contexts/ActiveFeedContext.tsx
rebelchris: I think you looked at non refreshed one as it was removed :)

PR #3022, opened 2024-04-18
feat: sidebar default state

sshanzel:
## Changes
- Experiment with the default state of the sidebar.
- The default value of the experiment flag is `true` since that is the default value of the settings to make it easier.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

MI-317 #done

### Preview domain
https://mi-317.preview.app.daily.dev


packages/shared/src/contexts/SettingsContext.tsx
```diff
@@ -197,6 +198,8 @@ export const SettingsContextProvider = ({
           ...settings,
           sidebarExpanded: !settings.sidebarExpanded,
         }),
+      onSidebarExpanded: (value: boolean) =>
+        setSettings({ ...settings, sidebarExpanded: value }),
```
sshanzel, line 202: We have to introduce a function to set it on what value we really expect it to be rather than toggling. The issue is that, the default value might not match what is the default value defined in GB, so toggling would just reset the value to the same thing.

kopancek: What does GB stands for here? I guess it's not Great Britain, neither Group Buy nor Giga Byte 😅

packages/shared/src/lib/featureManagement.ts
```diff
@@ -46,6 +46,7 @@ const feature = {
   tagsOnCard: new Feature('tags_on_card', false),
   onboardingMostVisited: new Feature('onboarding_most_visited', false),
   shareExperience: new Feature('share_experience', false),
+  sidebarToggle: new Feature('sidebar_toggle', true),
```
kopancek, line 49: I think a better name would be `sidebarOpen`, since that's what it actually does no? 🤔

packages/shared/src/components/sidebar/Sidebar.tsx
```diff
@@ -67,6 +71,19 @@ export default function Sidebar({
...
+  const sidebarToggleState = hasChangedToggle ? sidebarExpanded : sidebarToggle;
+
+  const onToggleSidebar = () => {
+    if (!sidebarToggle && !hasChangedToggle) {
+      completeAction(ActionType.SidebarToggle);
+      return onSidebarExpanded(true);
+    }
+
+    return toggleSidebarExpanded();
+  };
```
kopancek, line 86: Why do we need such a complicated logic? Can we instead change the `SettingsContext` to take the `feature.sidebarToggle` into account?

In this code, I don't really know what is happening :)

kopancek (commented):
I think the current logic is quite complex. Maybe we can simplify it a bit?

packages/shared/src/components/sidebar/Sidebar.tsx
sshanzel: Yeah, totally agree with that. Hence, I raised the discussion about making it [through the API](https://dailydotdev.slack.com/archives/C02E2C3C13R/p1713449074470439).

From the SettingsContext alone, it won't still know whether the value is coming from the hard-coded defaults or the actual settings. If we go through API, it will all be easier.

sshanzel:
> I think the current logic is quite complex. Maybe we can simplify it a bit?

Our only other option, imo, is going through API. Let's see what will the sync meeting tomorrow hold up.

packages/shared/src/lib/featureManagement.ts
nensidosari: I agree with Milan here on the naming toggle not being ideal - I would actually suggest `sidebarClosed`, this way we can keep default as we always do to false and have it true for experiment.

I can change the growthbook experiment if we agree on this

packages/shared/src/contexts/SettingsContext.tsx
nensidosari: Hahah Growthbook @kopancek

packages/shared/src/components/sidebar/Sidebar.tsx
```diff
@@ -67,6 +71,19 @@ export default function Sidebar({
   const showSettings = modal?.type === LazyModal.UserSettings;
   const { isNewMobileLayout } = useMobileUxExperiment();
   const isTablet = useViewSize(ViewSize.Tablet);
+  const sidebarToggle = useFeature(feature.sidebarToggle);
+  const { completeAction, checkHasCompleted } = useActions();
```
nensidosari, line 75: Why exactly do we have an action for this?

packages/shared/src/lib/featureManagement.ts
sshanzel: I just followed the convention to do camelCasing on the actual value. If we want we should probably change its feature name too.

packages/shared/src/components/sidebar/Sidebar.tsx
sshanzel: Because we don't want to persist the toggle being closed (if they end up on that value for the experiment) if they decide to manually toggle it open.

packages/shared/src/components/sidebar/Sidebar.tsx
capJavert: @sshanzel since this is new users only, what do you think that we just do it in onboarding after user registration. We can call `UPDATE_USER_SETTINGS_MUTATION` there and set the sidebar value depending if user is in experiment. All other logic then stays the same?

Basically it will simplify almost as if we did it on the API side (as you said) but it will be easier to implement I think.

packages/shared/src/lib/featureManagement.ts
nensidosari: Yeah I was referring to both, I'll change the growthbook feature to sidebar_closed, let's change the naming here as well for both 🙌

packages/shared/src/components/sidebar/Sidebar.tsx
nensidosari: That's a good idea, I'm just thinking it would need other implementation if product decides after this they want to experiment with current users, but maybe we don't need to care much at this point as it depends..

packages/shared/src/components/sidebar/Sidebar.tsx
nensidosari: I think @capJavert idea above solves this as well, let's see if we can consider going on that direction

packages/shared/src/components/sidebar/Sidebar.tsx
sshanzel: Good point. We can do that yes. Will try it quickly.

PR #3023, opened 2024-04-18
fix: layout shift on mobile safari when upvote cta is shown

kopancek:
## Changes

Details in thread: https://dailydotdev.slack.com/archives/C01L0QXQJNB/p1712925738062049

Layout shift is caused by the `ConditionalWrapper` component, which does not render fast enough on mobile safari. When condition is toggled to `true`, it causes Safari to remove the children, before rendering the wrapper. This causes the layout shift.

The only workaround I found was to render the wrapper all the time, which adds another `div`. I'm not a big fan, but cannot see a different solution, unless we render the whole upvote CTA differently.

- [X] MobileL (420px)
- [X] Tablet (656px)
- [X] Laptop (1020px)

- [X] iOS simulator (Safari)
- [ ] iOS (Chrome and Safari)
- [ ] Android

MI-301 #done

### Preview domain
https://mi-301-fix-upvote-cta-layout-shift.preview.app.daily.dev


packages/shared/src/components/cards/common/CardCover.tsx
```diff
@@ -26,15 +25,18 @@ export function CardCover({
...
+      <CardCoverShare
+        post={post}
+        onShare={() => {
+          onInteract();
+          onShare(post);
+        }}
+        onCopy={onInteract}
+      />
+    ),
+    [onInteract, onShare, post],
```
capJavert, line 39: No point in memoizing this. React has no use of this memozation and it is super fast in parsing virtual dom. You can move it below inline. As you seen in this issue, actually writting to DOM is the heaviest operation.

packages/shared/src/components/cards/common/CardCover.tsx
```diff
@@ -56,20 +58,13 @@ export function CardCover({
...
-    <ConditionalWrapper
-      condition={shouldShowOverlay}
-      wrapper={(component) => (
-        <div className="relative flex flex-1">
-          {coverShare}
-          {component}
-        </div>
-      )}
-    >
+    <div className="relative flex flex-1">
```
capJavert, line 61: This `flex` now used all the time does not break anything else layout wise?

capJavert (commented):
We had issues with conditional wrapper before. I am fine with this kind of fix. 👍

packages/shared/src/components/cards/common/CardCover.tsx
kopancek: Not that I can see.

packages/shared/src/components/cards/common/CardCover.tsx
kopancek: Yes, this was an oversight of mine, I tried a different approach before and forgot to revert this part.

sshanzel:
Looks fine to me too. Another option is probably adding a min-height?

nensidosari (approved):
See no issues with this fix

PR #3024, opened 2024-04-18
fix: usage of stale user data

sshanzel:
## Changes
- After logging back in, it still says you are logged out.
- I noticed the data and sometimes the user is getting overwritten.
- ~~After further investigation, there was a lot of missing dependencies on the useMemo.~~
- The memoized callback for updating the experiments was using stale data which makes the state gets overwritten with old data again.
- Tested production build locally.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

MI-315 #done

### Preview domain
https://mi-315.preview.app.daily.dev


packages/shared/src/contexts/AuthContext.tsx
```diff
@@ -1,8 +1,8 @@
 import React, {
   ReactElement,
   ReactNode,
+  useCallback,
```
sshanzel, line 4: Changes here are unnecessary, but we at least removed one `eslint-disable-next-line react-hooks/exhaustive-deps`

packages/shared/src/contexts/BootProvider.tsx
```diff
@@ -179,13 +179,9 @@ export const BootDataProvider = ({
     [setBootData],
   );
 
-  const updateExperimentation = useCallback(
-    (exp: BootCacheData['exp']) => {
-      const updated = updateLocalBootData(cachedBootData, { exp });
-      setCachedBootData(updated);
-    },
-    [cachedBootData],
```
sshanzel, line 187: Root cause of the issue, even though the `cachedBootData` is part of the dependencies, the callback execution was too fast after the data changed that the callback was not yet updated. Thankfully we don't need it since we can utilize using arrow function when setting state.

packages/shared/src/contexts/AuthContext.tsx
```diff
@@ -143,58 +143,44 @@ export const AuthContextProvider = ({
...
+        loadedUserFromCache,
+        getRedirectUri,
+        anonymous: user,
+        visit,
+        isFetched,
+        refetchBoot,
+        deleteAccount,
+        accessToken,
+        squads,
+      }}
```
nensidosari, line 181: If changes were unnecessary why do it? Couldn't we have removed the lint disable by having this in a variable still?

nensidosari (commented):
Just one check, looks good otherwise, good to have this resolved

packages/shared/src/contexts/AuthContext.tsx
sshanzel: The useMemo isn't really useful to put everything in it. We prefer using memoization only on places that is necessary.

nensidosari: Yeah not about memoization, just about this being in a separate variable

nensidosari (approved):
Just the comment above, looks good on my side otherwise

packages/shared/src/contexts/AuthContext.tsx
sshanzel: Right. That part, I just thought we wouldn't use it anywhere from here, so I just applied it directly. I don't have any preference though, so either way works for me.

capJavert (approved):

packages/shared/src/contexts/AuthContext.tsx
sshanzel: Merging now to unblock so we can immediately build an extension as this is pretty important.

If we ever decide otherwise, I can open another PR.

PR #3029, opened 2024-04-21
fix: companion popup close on outside click

shachargiladi:
Add onClickOutside prop to SimpleTooltip component in CompanionPopupButton component, and pass new closeCompoanionPopupButton as callback
Add hook usage of useKeyboardNavigation hook from the shared folder to handle Esc key press

- Now, with the dailydev extension when you press the compoanionPopupButton you can close it by clicking outside the popup or by clicking the esc key.

## Manual Testing
I checked that the bug was fixed (relevant only to extension), and did sanity tests to the extension.


shachargiladi:
[https://github.com/dailydotdev/daily/issues/1094](url)

packages/extension/src/companion/CompanionPopupButton.tsx
```diff
@@ -29,6 +30,12 @@ export const CompanionPopupButton = (): ReactElement => {
     setShowCompanionPermission(!showCompanionPermission);
   };
 
+  const closeCompanionPopupButton = () => {
+    setShowCompanionPermission(false);
+  };
+
+  useKeyboardNavigation(window, [['Escape', closeCompanionPopupButton]]);
```
sshanzel, line 37: While it is great to have this (also impressed you found a way how to utilize it), we might have to remove it for now since this will always have an event listener even when the popup is closed. Either we remove it as mentioned or update this hook to include another property on its 3rd parameter to indicate that no listener should be attached (an `enabled` boolean property for us to tell to return early).

shachargiladi: So would you prefer to remove it and find another way to achieve the close popup effect when pressing the 'esc' key? Or ignore this for now and focus only on close the popup when click outside?

sshanzel: I would prefer if we focus now on fixing the outside click since that was the primary issue.

> So would you prefer to remove it and find another way to achieve the close popup effect when pressing the 'esc' key?

I actually prefer what you have found, we just need to refactor the hook to support what I explained above. If you feel like it, feel free to open another PR once we have this merged if you are up for the challenge.

shachargiladi: great, so I fixed it and removed what you asked for

sshanzel (approved):
Thank you for the fix! 🚀

sshanzel:
@shachargiladi thank you so much for the effort 🙏 can you maybe manually update your branch, apparently GH is having some issues updating it through the PR. After that, we will be able to merge your PR 💪 

Screenshot 2024-04-24 at 7 20 28 PM

PR #3036, opened 2024-04-23
chore: cleanup mobile ux feature flag

nensidosari:
## Changes
- cleanup mobile ux flag since it's rolled out to all users and it was a success

MI-280 #done

### Preview domain
https://mi-280-mobileux-cleanup.preview.app.daily.dev


packages/shared/src/components/layout/MainLayoutHeader.tsx
```diff
@@ -50,7 +49,7 @@ function MainLayoutHeader({
   const isStreakLarge = streak?.current > 99; // if we exceed 100, we need to display it differently in the UI
   const router = useRouter();
   const isSearchPage = !!router.pathname?.startsWith('/search');
-  const { isNewMobileLayout } = useMobileUxExperiment();
+  const isMobileLayout = !useViewSize(ViewSize.Laptop);
```
rebelchris, line 52: Can't we solve these things with CSS in this case?

rebelchris (commented):
There just seem to be a couple cases where we can avoid JS evaluation I think and just go for CSS

packages/shared/src/components/layout/MainLayoutHeader.tsx
nensidosari: Yes, many of them actually 🤦‍♀️ thank u

nensidosari: I'll go through them 1 by 1 and see what makes sense. Just heads up, since most of mobile layout starts from tablet not just mobile, in some places it may become a mess on css when we try to define something from tablet and below (since at the moment tablet: is used for tablet and above).

packages/shared/src/components/notifications/NotificationsBell.tsx
```diff
@@ -37,12 +37,12 @@ function NotificationsBell({ compact }: { compact?: boolean }): ReactElement {
       <div
         className={classNames(
           'relative laptop:flex',
-          !isNewMobileLayout && 'hidden',
+          !isMobileLayout && 'hidden',
```
sshanzel, line 40: Now this becomes very confusing. We negated the laptop then we negated the `isMobileLayout`, then we have a `laptop` CSS class.

We will need to merge the line directly above this.

packages/shared/src/components/sidebar/Sidebar.tsx
```diff
@@ -95,7 +94,7 @@ export default function Sidebar({
     return <></>;
   }
 
-  if (isNewMobileLayout && isTablet) {
+  if (!isLaptop && isTablet) {
```
sshanzel, line 97: We don't even need the laptop condition here.
```suggestion
  if (isTablet) {
```

packages/webapp/__tests__/MainLayout.tsx
```diff
@@ -44,6 +46,6 @@ describe('MainLayout', () => {
     });
     const [el] = await screen.findAllByAltText(`idoshamun's profile`);
     expect(el).toHaveAttribute('src', 'https://daily.dev/ido.png');
-    expect(await screen.findAllByText('5')).toHaveLength(2);
```
sshanzel, line 47: Why did we end up changing the evaluation here.

packages/webapp/components/layouts/ProfileLayout/NavBar.tsx
```diff
@@ -52,7 +52,7 @@ export default function NavBar({
       className={classNames(
         'sticky top-12 z-3 -mt-px flex justify-around bg-background-default tablet:top-14 tablet:justify-start',
         styles.nav,
-        isNewMobileLayout && 'tablet:!top-0',
+        isMobileLayout && 'tablet:!top-0',
```
sshanzel, line 55: Then we can remove the `tablet:top-14` from above.
```suggestion
        'tablet:top-0',
```

packages/webapp/components/search/SearchPostFinderPage.tsx
```diff
@@ -57,7 +57,7 @@ const Search = (): ReactElement => {
           className={classNames(
             'flex w-full max-w-[32rem] flex-col items-center gap-4  self-center px-6',
             sidebarExpanded ? '!-left-32' : '!-left-6',
-            isNewMobileLayout && 'mt-12 laptop:mt-0',
+            isMobileLayout && 'mt-12 laptop:mt-0',
```
sshanzel, line 60:
```suggestion
            'mt-12 laptop:mt-0',
```

packages/webapp/pages/history.tsx
```diff
@@ -40,7 +39,7 @@ const History = (): ReactElement => {
   const router = useRouter();
   const tabQuery = router.query?.t?.toString() as HistoryType;
   const [page, setPage] = useState(HistoryType.Reading);
-  const { isNewMobileLayout } = useMobileUxExperiment();
+  const isMobileLayout = !isLaptop;
```
sshanzel, line 42: We don't even need this, we can just use the `isLaptop` alone from the conditionals since we negate them from time to time anyway.

packages/webapp/pages/squads/create.tsx
```diff
@@ -137,18 +136,17 @@ function CreatePost(): ReactElement {
           onActiveChange={(active) => setDisplay(active)}
           controlledActive={display}
           shouldMountInactive={false}
-          className={{ header: 'px-1' }}
-          showHeader={!isNewMobileLayout || isTablet}
+          showHeader={isTablet}
```
sshanzel, line 139: I think the condition here is not accurate.
```suggestion
          showHeader={!isMobile}
```

packages/webapp/pages/squads/create.tsx
```diff
@@ -137,18 +136,17 @@ function CreatePost(): ReactElement {
           onActiveChange={(active) => setDisplay(active)}
           controlledActive={display}
           shouldMountInactive={false}
-          className={{ header: 'px-1' }}
```
sshanzel, line 140: Why did we end up removing this header padding?

sshanzel (changes requested):
Just a few minor things and some conditionals that needs update, but finally we are removing all of this.

packages/shared/src/components/MainLayout.tsx
```diff
@@ -230,7 +228,7 @@ function MainLayoutComponent({
           !isScreenCentered && sidebarExpanded
             ? 'laptop:pl-60'
             : 'laptop:pl-11',
-          isNewMobileLayout && 'tablet:pl-16',
+          isMobileLayout && 'tablet:pl-16',
```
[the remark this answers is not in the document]
nensidosari: But from evaluation seems we want this only on tablet not laptop, if we remove isMobileLayout means it will be on tablet and laptop as well

packages/shared/src/components/layout/common.tsx
```diff
@@ -60,11 +59,11 @@ export const SearchControlHeader = ({
   const { sortingEnabled } = useContext(SettingsContext);
   const { isUpvoted, isSortableFeed } = useFeedName({ feedName });
   const { shouldUseMobileFeedLayout } = useFeedLayout();
-  const { isNewMobileLayout } = useMobileUxExperiment();
+  const isMobileLayout = !useViewSize(ViewSize.Laptop);
```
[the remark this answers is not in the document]
nensidosari: I'm converting all to isLaptop, was a bit in between myself when writing this

packages/shared/src/components/sidebar/Sidebar.tsx
nensidosari: Why not, isTablet stays true even in laptop sizes... 😞

packages/shared/src/components/MainLayout.tsx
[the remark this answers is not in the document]
sshanzel: Right, but the line directly above this will apply its needed `pl` values for laptop.

packages/shared/src/components/sidebar/Sidebar.tsx
sshanzel: Apparently, in our JS it is not the case 😓 

Yeah, they kinda work differently from what we have in CSS. Hopefully we would be able to make them the same in the future.

packages/webapp/pages/squads/create.tsx
nensidosari: This should be by mistake

packages/webapp/pages/squads/create.tsx
nensidosari: It is correct with the current isTablet implementation, since isTablet stays true even in laptop sizes

sshanzel: Yes, and with the condition of`! isNewMobileLayout`, this should be displayed on laptop.

sshanzel: Oh, right. This is `true` on `isTablet`. Nevermind 🤦

packages/shared/src/components/MainLayout.tsx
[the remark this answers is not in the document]
nensidosari: That will apply those values only if `!isScreenCentered && sidebarExpanded` is true, in cases it isn't laptop values wouldn't be applied.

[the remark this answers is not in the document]
sshanzel: It is a ternary operator, meaning if that condition fails, it will fallback to the `laptop:pl-11`.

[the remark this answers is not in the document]
nensidosari: That will only work correctly if I add any laptop:pl-0 on top and then add ! to the ones adding values to override it.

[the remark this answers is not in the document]
nensidosari: Ah thank u, my eyes were not working properly 😓

packages/shared/src/components/sidebar/Sidebar.tsx
nensidosari: I think they work exactly like what we have in css, when we say `tablet:...` we mean everything tablet and above, unless it's overriten by any `laptop:...`, which I still don't like, I think it's not bad to maintain even breakpoints with min & max values so u don't have to write so many conditions when u want something only on a specific breakpoint.

packages/webapp/components/layouts/ProfileLayout/NavBar.tsx
nensidosari: this ended up to be laptop:top-14 from what I understood, just adding comment for u to check as well 🙏

packages/webapp/__tests__/MainLayout.tsx
nensidosari: I'll check this again, it started failing, not sure why it finds 2 in the first place so will dig a bit more.

packages/webapp/components/layouts/ProfileLayout/NavBar.tsx
sshanzel: Ah, yes, was a mistake with the suggestion, should be `laptop:top-14`.

packages/shared/src/components/filters/MyFeedHeading.tsx
```diff
@@ -55,16 +54,8 @@ function MyFeedHeading({
...
-              ? ButtonSize.Small
-              : ButtonSize.Medium
-          }
-          variant={
-            isMobile && isNewMobileLayout
-              ? ButtonVariant.Tertiary
-              : ButtonVariant.Float
-          }
+          size={ButtonSize.Medium}
+          variant={isMobile ? ButtonVariant.Tertiary : ButtonVariant.Float}
```
sshanzel, line 58: I understand this is not part of the supposed changes, but should we maybe store these values in a const, so whenever we need some changes in the future, we won't miss a button that would be misaligned?

packages/webapp/components/FooterNavBar.tsx
```diff
@@ -65,15 +65,15 @@ export default function FooterNavBar({
     'shadow-[0_4px_30px_rgba(0,0,0.1)]',
   );
 
-  const Component = isNewMobileLayout ? FooterNavBarV1 : FooterNavBarControl;
+  const Component = isLaptop ? FooterNavBarControl : FooterNavBarV1;
```
sshanzel, line 68: We can completely remove the `FooterNavBarControl` and the component itself. This component is only available for tablet and mobile.

packages/webapp/components/FooterNavBar.tsx
```diff
@@ -65,15 +65,15 @@ export default function FooterNavBar({
...
 
   return (
     <div
       className={classNames(
         'fixed !bottom-0 left-0 z-2 w-full',
-        isNewMobileLayout
-          ? 'footer-navbar bg-gradient-to-t from-background-subtle from-70% to-transparent px-2 pt-2'
-          : post && 'bg-blur-bg backdrop-blur-20',
+        post && 'laptop:bg-blur-bg laptop:backdrop-blur-20',
+        !isLaptop &&
```
sshanzel, line 75: This is going to be true all the time since as mentioned this file is not available for laptop. This means we don't need the condition anymore.

packages/webapp/components/FooterNavBar.tsx
```diff
@@ -96,11 +96,9 @@ export default function FooterNavBar({
         spring="veryGentle"
         element="nav"
         className={classNames(
-          'grid w-full grid-flow-col items-center justify-between px-3',
+          'laptop:footer-navbar grid w-full grid-flow-col items-center justify-between rounded-16 px-3 laptop:h-14 laptop:rounded-none laptop:rounded-t-24 laptop:bg-background-default',
```
sshanzel, line 99: No need to add laptop classes around here. Should make our lives easier.

packages/webapp/components/FooterNavBar.tsx
```diff
@@ -96,11 +96,9 @@ export default function FooterNavBar({
...
           !showNav && 'hidden',
-          isNewMobileLayout
-            ? classNames('rounded-16', !post && activeClasses)
-            : 'footer-navbar h-14 rounded-t-24 bg-background-default',
+          !isLaptop && !post && activeClasses,
```
sshanzel, line 101:
```suggestion
          !post && activeClasses,
```

packages/webapp/components/FooterNavBar.tsx
```diff
@@ -65,15 +65,15 @@ export default function FooterNavBar({
...
+  const Component = isLaptop ? FooterNavBarControl : FooterNavBarV1;
 
   return (
     <div
       className={classNames(
         'fixed !bottom-0 left-0 z-2 w-full',
-        isNewMobileLayout
-          ? 'footer-navbar bg-gradient-to-t from-background-subtle from-70% to-transparent px-2 pt-2'
-          : post && 'bg-blur-bg backdrop-blur-20',
+        post && 'laptop:bg-blur-bg laptop:backdrop-blur-20',
```
sshanzel, line 74: We can completely remove this line. This is for the old UI.

packages/webapp/components/layouts/FooterNavBarLayout.tsx
```diff
@@ -21,20 +20,20 @@ export default function FooterNavBarLayout({
...
   const isTablet = useViewSize(ViewSize.Tablet);
+  const isLaptop = useViewSize(ViewSize.Laptop);
 
   const showNav = useMemo(() => {
     if (!windowLoaded) {
       return false;
     }
 
-    if (isNewMobileLayout) {
+    if (!isLaptop) {
```
sshanzel, line 31: We can change this line 31 to line 36 with `return isMobile`. The code at line 36 basically checks whether it is a touch device or not.

packages/webapp/components/search/SearchPostFinderPage.tsx
```diff
@@ -55,9 +53,8 @@ const Search = (): ReactElement => {
       {!searchQuery && (
         <div
           className={classNames(
-            'flex w-full max-w-[32rem] flex-col items-center gap-4  self-center px-6',
+            'mt-12 flex w-full max-w-[32rem] flex-col items-center  gap-4 self-center px-6 laptop:mt-0',
```
sshanzel, line 56:
```suggestion
            'mt-12 flex w-full max-w-[32rem] flex-col items-center gap-4 self-center px-6 laptop:mt-0',
```

packages/shared/src/components/MainLayout.tsx
```diff
@@ -225,12 +222,9 @@ function MainLayoutComponent({
...
       <main
         className={classNames(
-          'flex flex-col',
+          'flex flex-col tablet:pl-16 laptop:pl-11',
           className,
-          !isScreenCentered && sidebarExpanded
-            ? 'laptop:pl-60'
-            : 'laptop:pl-11',
-          isNewMobileLayout && 'tablet:pl-16',
+          !isScreenCentered && sidebarExpanded && 'laptop:!pl-60',
```
sshanzel, line 227: Have no strong opinion here since it looks more understandable with the inline changes you made above, but just not a fan of `!`. Either way should be good 🚀

sshanzel (commented):
We are almost there! 🚀

packages/shared/src/components/filters/MyFeedHeading.tsx
nensidosari: Hmm not sure I understand here, this is only used in this line so what other button could be missaligned?

packages/webapp/components/FooterNavBar.tsx
nensidosari: 👍 but why did we have even control in this case?

packages/shared/src/components/MainLayout.tsx
nensidosari: No strong opinion either, just thought this way it was more clear

packages/webapp/components/FooterNavBar.tsx
sshanzel: If the user was under the old or new layout experiment.

packages/shared/src/components/filters/MyFeedHeading.tsx
sshanzel: There is another button a bit below that has the same `size` and `variant` values/conditions.

nensidosari: Hm, it has the same size and variant but the condition there is isLaptop, while here isMobile

sshanzel: Ah, okay. Cool, feel free to ignore this one. Thank you!

packages/webapp/components/FooterNavBar.tsx
nensidosari: This file is actually available on laptop, as with showNav we only use it to conditionally render only parts of this file, I'll move showNav to actually conditionally render this whole file so this works correctly

packages/webapp/components/footer/FooterNavBar.tsx
```diff
@@ -82,7 +82,7 @@ const Tab = ({ tab, isActive }: TabProps) => {
   );
 };
 
-export function FooterNavBarV1({
+export function FooterNavBar({
```
sshanzel, line 85: There are now two `FooterNavBar` component. Maybe we can rename the other one to put `Container` at the end or `Tabs` in here? Anything should work as long as they will have different names.

sshanzel (approved):
Nice! Thank you so much for the effort, and the additional cleanup on the `showNav`. Makes maintaining the code so much better 💪 

Two last things, but everything looks amazing already, so approving and to unblock.

packages/webapp/components/FooterNavBar.tsx
sshanzel: Yeah, it seems that some parts were rendered conditionally only. Now that we completely do not render this footer, we should maybe take the Scroll to top component out so it would still be available on Tablet and Laptop.

packages/webapp/__tests__/MainLayout.tsx
nensidosari: So the test without my changes runs on mobile, and because we have different things on mobile main layout, this starts failing in the new layout, it looks for an src where we don't show any. So test at line 46 fails, reason in this in mainLayoutHeader:
Screenshot 2024-04-25 at 12 29 33 PM

If we force the test to run on laptop, like I've done now, the test fails even in production (that length 2 doesn't exist), so for a safer test until we cover both mobile and laptop, it's better to leave this run only on laptop, and fix the test for both production and here.

sshanzel: Yeah, agreed. Since we removed legacy code, this would definitely be affected. Sounds good to me 🚀

packages/webapp/components/footer/FooterNavBar.tsx
nensidosari: Going with FooterNavBarTabs

packages/webapp/components/FooterNavBar.tsx
nensidosari: Brought back show nav on the FooterNavBarTabs, but fixed classes to show the scroll to top properly

packages/shared/src/components/filters/MyFeedHeading.tsx
```diff
@@ -73,18 +64,12 @@ function MyFeedHeading({
           }
           loading={isRefreshing}
         >
-          {!isMobile && !isNewMobileLayout ? 'Refresh feed' : null}
+          {isLaptop ? 'Refresh feed' : null}
```
rebelchris, line 67: Just wondering but should this not match the variant above?
(so !isMobile)

rebelchris (approved):
One non blocking question, otherwise good to go.

packages/shared/src/components/filters/MyFeedHeading.tsx
nensidosari: The expression would be `!isMobile && isLaptop` which would only be true if `isLaptop` is true, thus changed to `isLaptop`. If we would say only `!isMobile` that would include tablet as well, which the above expression is not including.

sshanzel:
It is always a good day when you see more red lines than green 🤩 

Definitely a job well done with the cleanup, it would greatly help everyone maintaining the code 🚀

PR #3037, opened 2024-04-23
feat: add user experience dropdown to signup and account/profile

kopancek:
## Changes

Related issue: https://dailydotdev.atlassian.net/browse/MI-314

Adds experience level dropdown to signup pages and to the edit account profile page.

Depends on API changes PR: 
- https://github.com/dailydotdev/daily-api/pull/1851

- new feature `experienceLevel`, default `false`
- adds experience level dropdown (required param) to the signup page if feature is on
- adds experience level dropdown to the edit account profile page, if feature is on

#### Profile page

Screenshot 2024-04-23 at 14 08 36
Screenshot 2024-04-23 at 14 08 44
Screenshot 2024-04-23 at 14 08 53

#### Social signup

Screenshot 2024-04-23 at 14 12 23
Screenshot 2024-04-23 at 14 13 22
Screenshot 2024-04-23 at 14 13 33

#### Email signup

Screenshot 2024-04-23 at 14 09 33
Screenshot 2024-04-23 at 14 10 50
Screenshot 2024-04-23 at 14 11 10

- [X] MobileL (420px)
- [ ] Tablet (656px)
- [X] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

MI-314 #done

### DIFF
--- a/packages/shared/src/components/auth/RegistrationForm.tsx
+++ b/packages/shared/src/components/auth/RegistrationForm.tsx
@@ -30,6 +30,9 @@ import { AuthFormProps } from './common';
 import ConditionalWrapper from '../ConditionalWrapper';
 import AuthContainer from './AuthContainer';
 import { onValidateHandles } from '../../hooks/useProfileForm';
+import { useFeature } from '../GrowthBookProvider';
+import { feature } from '../../lib/featureManagement';
+import ExperienceLevelDropdown from '../profile/ExperienceLevelDropdown';
 
 export interface RegistrationFormProps extends AuthFormProps {
   email: string;
@@ -63,6 +66,7 @@ export const RegistrationForm = ({
   const [name, setName] = useState('');
   const isAuthorOnboarding = trigger === AuthTriggers.Author;
   const { username, setUsername } = useGenerateUsername(name);
+  const showExperienceLevel = useFeature(feature.experienceLevel);
 
   useEffect(() => {
     trackEvent({
@@ -96,7 +100,11 @@ export const RegistrationForm = ({
       formRef?.current ?? form,
     );
 
-    if (!values['traits.name']?.length || !values['traits.username']?.length) {
+    if (
+      !values['traits.name']?.length ||
+      !values['traits.username']?.length ||
+      (showExperienceLevel && !values['traits.experienceLevel']?.length)
+    ) {
       const setHints = { ...hints };
 
       if (!values['traits.name']?.length) {
@@ -104,6 +112,9 @@ export const RegistrationForm = ({
       }
       if (!values['traits.username']?.length) {
         setHints['traits.username'] = 'Please provide username.';
+      }
+      if (!values['traits.experienceLevel']?.length) {
+        setHints['traits.experienceLevel'] = 'Please provide experience level.';
       }
 
       onUpdateHints(setHints);
@@ -141,6 +152,8 @@ export const RegistrationForm = ({
 
   const isNameValid = !hints?.['traits.name'] && isSubmitted;
   const isUsernameValid = !hints?.['traits.username'] && isSubmitted;
+  const isExperienceLevelValid =
+    !isSubmitted || !hints?.['traits.experienceLevel'];
 
   return (
     <>
@@ -224,6 +237,19 @@ export const RegistrationForm = ({
             label="X"
             type="text"
             required
+          />
+        )}
+        {showExperienceLevel && (
+          <ExperienceLevelDropdown
+            className={{ container: 'w-full' }}
+            name="traits.experienceLevel"
+            valid={isExperienceLevelValid}
+            hint={hints?.['traits.experienceLevel']}
+            onChange={() =>
+              hints?.['traits.experienceLevel'] &&
+              onUpdateHints({ ...hints, 'traits.experienceLevel': '' })
+            }
+            saveHintSpace
           />
         )}
         <span className="border-b border-border-subtlest-tertiary pb-4 text-text-secondary typo-subhead">
--- a/packages/shared/src/components/fields/Dropdown.tsx
+++ b/packages/shared/src/components/fields/Dropdown.tsx
@@ -2,6 +2,7 @@ import React, {
   CSSProperties,
   ReactElement,
   ReactNode,
+  useEffect,
   useRef,
   useState,
 } from 'react';
@@ -22,7 +23,7 @@ import { DrawerProps } from '../drawers';
 import { Button, ButtonSize, ButtonVariant } from '../buttons/Button';
 import { IconProps } from '../Icon';
 
-interface ClassName {
+export interface DropdownClassName {
   container?: string;
   menu?: string;
   label?: string;
@@ -36,7 +37,7 @@ export interface DropdownProps {
   icon?: ReactNode;
   shouldIndicateSelected?: boolean;
   dynamicMenuWidth?: boolean;
-  className?: ClassName;
+  className?: DropdownClassName;
   style?: CSSProperties;
   selectedIndex: number;
   options: string[];
@@ -49,6 +50,7 @@ export interface DropdownProps {
   iconOnly?: boolean;
   drawerProps?: Omit<DrawerProps, 'children' | 'onClose'>;
   openFullScreen?: boolean;
+  onOpenChange?: (open: boolean) => void;
 }
 
 export function Dropdown({
@@ -57,6 +59,7 @@ export function Dropdown({
   selectedIndex,
   options,
   onChange,
+  onOpenChange,
   dynamicMenuWidth,
   shouldIndicateSelected,
   buttonSize = ButtonSize.Large,
@@ -75,6 +78,10 @@ export function Dropdown({
   const [menuWidth, setMenuWidth] = useState<number>();
   const triggerRef = useRef<HTMLButtonElement>();
   const { show, hideAll } = useContextMenu({ id });
+
+  useEffect(() => {
+    onOpenChange?.(isVisible);
+  }, [isVisible, onOpenChange]);
 
   const showMenu = (event: TriggerEvent): void => {
     const { right, bottom, width } = triggerRef.current.getBoundingClientRect();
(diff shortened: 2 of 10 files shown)


packages/shared/src/components/auth/RegistrationForm.tsx
```diff
@@ -96,7 +100,11 @@ export const RegistrationForm = ({
       formRef?.current ?? form,
     );
 
-    if (!values['traits.name']?.length || !values['traits.username']?.length) {
+    if (
+      !values['traits.name']?.length ||
+      !values['traits.username']?.length ||
+      (showExperienceLevel && !values['traits.experienceLevel']?.length)
```
rebelchris, line 106: Did you communicate this with Viktor to allow it from Heimdall side?

packages/webapp/pages/account/profile.tsx
```diff
@@ -54,6 +67,18 @@ const AccountProfilePage = (): ReactElement => {
...
+  // fetch user experience level
+  const { data } = useQuery<UserExperienceLevelData>(
+    generateQueryKey(RequestKey.UserExperienceLevel, user),
+    () =>
+      request(graphqlUrl, GET_USER_EXPERIENCE_LEVEL, {
+        id: user.id,
+      }),
+    { enabled: !!user && showExperienceLevel },
+  );
+  const experienceLevel = data?.user?.experienceLevel;
```
rebelchris, line 81: Why do we do this as a seperate query, can't we just add this field to the profile query?

packages/webapp/pages/account/profile.tsx
```diff
@@ -183,6 +209,17 @@ const AccountProfilePage = (): ReactElement => {
...
           value={user.username}
         />
+        {showExperienceLevel && (
+          <ExperienceLevelDropdown
+            defaultValue={experienceLevel as UserExperienceLevelKey}
+            name="experienceLevel"
+            className={{
+              container: 'mt-6 max-w-sm',
+              button:
+                '!text-text-primary hover:shadow-[inset_0.125rem_0_0_var(--theme-text-primary)]',
```
rebelchris, line 219: Why is the override needed here? (it doesn't match guidelines?)

rebelchris (commented):
Since you opt for onboaring + traits we need to ensure heimdall can pass it through.
Meaning on API I think you'll also need to accept it on the new user endpoint heimdall uses.

packages/webapp/pages/account/profile.tsx
kopancek: We get all the other fields directly from the user object, as part of the `AuthContext`. Since this is the only place we use the field, it seems cleaner to me to add a specific query for this page, rather than query this field for AuthContext unnecessarily.

packages/shared/src/components/auth/RegistrationForm.tsx
kopancek: As far as I understand, Heimdall has nothing to do with this, since we just set this data in the database with a graphql mutation after heimdall verifies that the email is eligible for signup. But correct me if I'm wrong :)

packages/webapp/pages/account/profile.tsx
kopancek: You're right, not needed, removing...

packages/shared/src/components/auth/RegistrationForm.tsx
capJavert: New `traits` need to be added on Heimdall side to be passed over https://github.com/dailydotdev/heimdall/pull/71

packages/webapp/pages/account/profile.tsx
kopancek: That means adding it to boot, just FYI...

capJavert: We have `useProfile` which is used to get more complete version of user profile, but ti avoid refactoring of that page adding to boot for now is ok for me.

packages/shared/src/components/auth/RegistrationForm.tsx
rebelchris: Yeah it's something Viktor can help with, he owns the system there.

packages/shared/src/components/fields/Dropdown.tsx
```diff
@@ -76,6 +79,10 @@ export function Dropdown({
   const triggerRef = useRef<HTMLButtonElement>();
   const { show, hideAll } = useContextMenu({ id });
 
+  useEffect(() => {
+    onOpenChange?.(isVisible);
```
nensidosari, line 83: Why did our dropdown need this exactly?

packages/shared/src/lib/featureManagement.ts
```diff
@@ -35,6 +35,7 @@ const feature = {
   readingReminder: new Feature('reading_reminder', false),
   onboardingMostVisited: new Feature('onboarding_most_visited', false),
   shareExperience: new Feature('share_experience', false),
+  experienceLevel: new Feature('experience_level', false),
```
nensidosari, line 38: Please use v1 & control as mentioned in the ticket in order to allow any follow up on this.

packages/shared/src/components/auth/RegistrationForm.tsx
kopancek: Heimdall PR has been merged now:
- https://github.com/dailydotdev/heimdall/pull/94

packages/shared/src/components/fields/Dropdown.tsx
kopancek: The experience level dropdown needs to know if the underlying dropdown is open or not - we change the styles depending on that.

packages/shared/src/lib/featureManagement.ts
kopancek: Missed that, good catch!

packages/shared/src/components/fields/Dropdown.tsx
nensidosari: 👍 Hmm, just thinking, not to block this PR, but maybe we can check with design if this is intentional, if we want these styles changed in every dropdown we have when it's open or not. As it seems like a new addition on their side

nensidosari: 👍 Hmm, just thinking, not to block this PR, but maybe we can check with design if this is intentional, if we want these styles changed in every dropdown we have when it's open or not. As it seems like a new addition on their side

packages/shared/src/components/auth/RegistrationForm.tsx
```diff
@@ -141,6 +152,8 @@ export const RegistrationForm = ({
 
   const isNameValid = !hints?.['traits.name'] && isSubmitted;
   const isUsernameValid = !hints?.['traits.username'] && isSubmitted;
+  const isExperienceLevelValid =
+    !isSubmitted || !hints?.['traits.experienceLevel'];
```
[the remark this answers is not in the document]
kopancek: If the form has not been submitted, the isExperienceLevelValid would always be false. I'm not sure how the other fields handle that, but this was causing the experience level warning to be shown as red (screenshot below).

Screenshot 2024-04-24 at 08 34 24

If you think about it, this is actually correct logic. E.g. if the form is not submitted, we do not mark the field as invalid.

packages/webapp/pages/account/profile.tsx
kopancek: Changed accordingly, we now get the experienceLevel field from boot.

packages/shared/src/components/fields/Dropdown.tsx
kopancek: I looked at the code again and now understand why I did that. The error hint implementation is different than in Figma, to make it aligned to all the other fields (instead of showing a rounded rectangle with an exclamation mark triangle icon, we just show red text below and add red box-shadow). But it does not make sense to add the red box-shadow when the dropdown is open, hence the following code:
```
              !open &&
              !valid &&
              '!shadow-[inset_0.125rem_0_0_var(--status-error)]',
```

I'll verify with design if this is OK.

kopancek: https://dailydotdev.slack.com/archives/C06UBS0N5FU/p1713941141715529

nensidosari: Do u think this should be a behaviour for only this dropdown or all of them if the field is required and not filled? I feel like this should be part of the component itself no?

packages/shared/src/components/auth/RegistrationForm.tsx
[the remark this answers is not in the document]
nensidosari: > if the form is not submitted, we do not mark the field as invalid.

Agree with that, just because the others were handled differently it caught my eye. Still don't understand 100% how come they are handled differently.

packages/shared/src/components/fields/Dropdown.tsx
kopancek: @nensidosari design is OK with the changes and will apply it to the guidelines: https://dailydotdev.slack.com/archives/C06UBS0N5FU/p1713943722338799?thread_ts=1713941141.715529&cid=C06UBS0N5FU

nensidosari: > Do u think this should be a behaviour for only this dropdown or all of them if the field is required and not filled? I feel like this should be part of the component itself no?

please read this one more time, not much to do with design in general but component itself on how we implement, do we keep error states (css) within the component or outside as u're doing atm?

nensidosari: As discussed I'm good if we keep this implementation and work on a refactor later on a different PR

packages/shared/src/svg/StarWarsLogoIcon.tsx
```diff
@@ -0,0 +1,24 @@
+import React, { ReactElement } from 'react';
+
+export default function StarWarsLogoIcon(
```
nensidosari, line 3: @kopancek I think u've confused branches

kopancek: you're right 😢

nensidosari (approved):
Looks good on my side

packages/shared/src/components/fields/Dropdown.tsx
kopancek: Created https://dailydotdev.atlassian.net/browse/MI-331 as a followup for the refactor. If I have time in the sprint, I'll try to squeeze it in :)

PR #3039, opened 2024-04-23
feat: experiment on sidebar state

sshanzel:
## Changes
- Whenever registration happens, we sync the local settings of our user, for example, when they switched to light mode, etc.
- In this case, we can be sure we will only apply the experiment on new users to set their default sidebar state value.

Kudos to @capJavert for the great idea.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

MI-317 #done

### Preview domain
https://mi-317-post-registration.preview.app.daily.dev

### DIFF
--- a/packages/shared/src/contexts/SettingsContext.tsx
+++ b/packages/shared/src/contexts/SettingsContext.tsx
@@ -18,6 +18,8 @@ import AuthContext from './AuthContext';
 import { graphqlUrl } from '../lib/config';
 import { capitalize } from '../lib/strings';
 import { storageWrapper } from '../lib/storageWrapper';
+import { useFeaturesReadyContext } from '../components/GrowthBookProvider';
+import { feature } from '../lib/featureManagement';
 
 export enum ThemeMode {
   Dark = 'dark',
@@ -120,6 +122,7 @@ export const SettingsContextProvider = ({
 }: SettingsContextProviderProps): ReactElement => {
   const { user } = useContext(AuthContext);
   const userId = user?.id;
+  const { getFeatureValue } = useFeaturesReadyContext();
 
   useEffect(() => {
     if (!loadedSettings) {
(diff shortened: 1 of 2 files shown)


packages/shared/src/contexts/SettingsContext.tsx
```diff
@@ -174,7 +177,17 @@ export const SettingsContextProvider = ({
   };
 
   const syncSettings = async (bootUserId?: string) => {
-    await updateRemoteSettingsFn(settings, bootUserId);
+    const updatedSettings = settings;
+
+    if (settings.sidebarExpanded) {
+      const sidebarClosed = getFeatureValue(feature.sidebarClosed);
```
rebelchris, line 183: Here we could have actually used the conditionalFeature but also happy with this. 👍👍

packages/shared/src/contexts/SettingsContext.tsx
```diff
@@ -174,7 +177,17 @@ export const SettingsContextProvider = ({
...
+
+      if (sidebarClosed) {
+        updatedSettings.sidebarExpanded = false;
+      }
+    }
+
+    await updateRemoteSettingsFn(updatedSettings, bootUserId);
```
rebelchris, line 190: Oh wow we already had the exact thing 😂

rebelchris (approved):
Awesome

packages/shared/src/contexts/SettingsContext.tsx
sshanzel: My concern was that the user might have not fully registered yet but might get enrolled. So I thought of using this.

rebelchris: The timing would be the same though 👀 but this is fine for me 👍

capJavert (approved):

packages/shared/src/contexts/SettingsContext.tsx
sshanzel: I think the reason we pass `bootUserId` on this function is that the user object (from the auth context hook) might not be updated yet which could also be the case if we call the feature flag from the hook. But yeah, they both might work similarly.

rebelchris: Oh but then does the useFeatureValue use the right user id?

This might be crucial for the analytics parts so rather double check/inform the analytics team if its some nuance 🙏

PR #3042, opened 2024-04-24
 fix: copy comment link unnecessary characters

sshanzel:
## Changes
- Remove the unwanted extra characters when sharing a comment.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

MI-327 #done

### Preview domain
https://mi-327.preview.app.daily.dev

### DIFF
--- a/packages/shared/src/hooks/useShareComment.ts
+++ b/packages/shared/src/hooks/useShareComment.ts
@@ -27,7 +27,7 @@ export function useShareComment(origin: Origin): UseShareComment {
             ReferralCampaignKey.ShareComment,
           );
           await navigator.share({
-            text: `${post.title}\n${shortUrl}$}`,
+            text: `${post.title}\n${shortUrl}`,
           });
           trackEvent(
             postAnalyticsEvent('share comment', post, {


packages/shared/src/hooks/useShareComment.ts
```diff
@@ -27,7 +27,7 @@ export function useShareComment(origin: Origin): UseShareComment {
             ReferralCampaignKey.ShareComment,
           );
           await navigator.share({
-            text: `${post.title}\n${shortUrl}$}`,
+            text: `${post.title}\n${shortUrl}`,
```
nensidosari, line 30: Should we even leave post title when they just copy the link not try to share it in squad or social media?

rebelchris (approved):

packages/shared/src/hooks/useShareComment.ts
sshanzel: Good point, we should keep it, but to do that, we should make the implementation similar to when sharing a post. Let me update it a little bit.

sshanzel: With the applied changes, when you just hit copy on the native share, it won't add any link, then if you share it through something else, the title will be present.

sshanzel: I tried to make it similar to `useSharePost` by utilizing the `title` property, but when you share, for example, on Twitter, the title is present on post share but on comment share, there is no preview. Not sure if we should make changes here now.

sshanzel: I think we should stick to the original process until product says otherwise since we've had this flow since then.

PR #3043, opened 2024-04-24
refactor: squads creation page

sshanzel:
## Changes
- Put the actions on the header even on laptop, then remove the footer.
- Removed a lot of avoidable extra wrappers.
- Cleaned up some unused components after the changes.

Previews:

Screenshot 2024-04-25 at 12 04 04 AM
Screenshot 2024-04-25 at 12 03 59 AM
Screenshot 2024-04-25 at 12 03 55 AM
Screenshot 2024-04-24 at 11 36 03 PM
Screenshot 2024-04-24 at 11 35 59 PM
Screenshot 2024-04-24 at 11 35 50 PM

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

MI-318 #done

### Preview domain
https://mi-318.preview.app.daily.dev

### DIFF
--- a/packages/shared/src/components/fields/form/FormWrapper.tsx
+++ b/packages/shared/src/components/fields/form/FormWrapper.tsx
@@ -2,7 +2,6 @@ import React, { MutableRefObject, ReactElement, ReactNode } from 'react';
 import classNames from 'classnames';
 import { Button, ButtonProps, ButtonVariant } from '../../buttons/Button';
 import classed from '../../../lib/classed';
-import { ArrowIcon } from '../../icons';
 
 interface Copy {
   left?: string;
--- a/packages/shared/src/components/fields/form/FormWrapper.tsx
+++ b/packages/shared/src/components/fields/form/FormWrapper.tsx
@@ -21,7 +21,7 @@ export interface FormWrapperProps {
   leftButtonProps?: ButtonProps<'button'>;
   rightButtonProps?: ButtonProps<'button'>;
   title?: string;
-  isCancelArrow?: boolean;
+  isHeaderTitle?: boolean;
   headerRef?: MutableRefObject<HTMLDivElement>;
 }
 
@@ -35,14 +35,14 @@ export function FormWrapper({
   leftButtonProps = {},
   rightButtonProps = {},
   title,
-  isCancelArrow,
+  isHeaderTitle,
   headerRef,
 }: FormWrapperProps): ReactElement {
   const { left = 'Cancel', right = 'Submit' } = copy;
   const titleElement = (
     <Title
       className={classNames(
-        !isCancelArrow && 'mt-5',
+        !isHeaderTitle && 'mt-5',
         className?.title ?? 'typo-body',
       )}
     >
@@ -59,14 +59,10 @@ export function FormWrapper({
         )}
         ref={headerRef}
       >
-        <Button
-          {...leftButtonProps}
-          icon={isCancelArrow ? <ArrowIcon className="-rotate-90" /> : null}
-          variant={ButtonVariant.Tertiary}
-        >
-          {isCancelArrow ? null : left}
+        <Button {...leftButtonProps} variant={ButtonVariant.Tertiary}>
+          {isHeaderTitle ? null : left}
         </Button>
-        {isCancelArrow && title && titleElement}
+        {isHeaderTitle && title && titleElement}
         <Button
           {...rightButtonProps}
           variant={ButtonVariant.Primary}
@@ -76,7 +72,7 @@ export function FormWrapper({
           {right}
         </Button>
       </div>
-      {!isCancelArrow && title && titleElement}
+      {!isHeaderTitle && title && titleElement}
       {children}
     </div>
   );
--- a/packages/shared/src/components/fields/form/FormWrapper.tsx
+++ b/packages/shared/src/components/fields/form/FormWrapper.tsx
@@ -1,6 +1,7 @@
 import React, { MutableRefObject, ReactElement, ReactNode } from 'react';
 import classNames from 'classnames';
 import { Button, ButtonProps, ButtonVariant } from '../../buttons/Button';
+import classed from '../../../lib/classed';
 
 interface Copy {
   left?: string;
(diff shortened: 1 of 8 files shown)


packages/shared/src/components/fields/form/FormWrapper.tsx
```diff
@@ -9,18 +10,22 @@ interface Copy {
 interface ClassName {
   container?: string;
   header?: string;
+  title?: string;
 }
 export interface FormWrapperProps {
   children: ReactNode;
-  className?: string | ClassName;
```
sshanzel, line 15: Had to remove the union type. Not really a fan in the first place, but for each and every new property to the `ClassName` object, we will add another line of ternary operator.

packages/shared/src/components/squads/Details.tsx
```diff
@@ -206,150 +164,109 @@ export function SquadDetails({
...
-              label={createMode ? 'Name your Squad' : 'Squad name'}
-              inputId="name"
-              name="name"
-              valid={!!name}
-              leftIcon={<SquadIcon />}
-              value={name ?? ''}
-              className={{
-                container: 'w-full',
-              }}
+      <form
```
sshanzel, line 167: Most of the changes here is the wrapper. Maybe I should put them in a separate commit next time.

packages/shared/src/components/fields/form/FormWrapper.tsx
```diff
@@ -29,34 +34,44 @@ export function FormWrapper({
   leftButtonProps = {},
   rightButtonProps = {},
   title,
+  isTitleBesideCancel,
```
nensidosari, line 37: is the title beside on placed instead of cancel?

packages/shared/src/components/fields/form/FormWrapper.tsx
```diff
@@ -29,34 +34,44 @@ export function FormWrapper({
...
+          'flex flex-row items-center border-b border-border-subtlest-tertiary px-4 py-2',
+          className?.header,
         )}
         ref={headerRef}
       >
         <Button {...leftButtonProps} variant={ButtonVariant.Tertiary}>
-          {copy?.left ?? 'Cancel'}
+          {left}
         </Button>
+        {isTitleBesideCancel && title && titleElement}
```
nensidosari, line 64: Can u attach an image of these 2 variations of title, just not picturing them right in my head

packages/shared/src/components/squads/Details.tsx
nensidosari: just to check that this is tested on all screens with all the changes and it still works ok?

nensidosari (commented):
Quite a lot of refactoring changes, love that we're removing things, just to make sure we don't miss anything will check the preview domain and see if everything is alright. 
Happy to approve after

nensidosari:
In between some screens the squad details post permissions and below,  info stay centered - maybe we should fix to stay left. Not a big issue though.
Screenshot 2024-04-25 at 10 11 50 AM


Other then that I think it looks all good, tried to check all the affected pages

packages/shared/src/components/fields/form/FormWrapper.tsx
sshanzel: If we make it part of the button, it looks weird when hovered and it will have a different font size if you are referring to make it as one with the Cancel button.

packages/shared/src/components/fields/form/FormWrapper.tsx
sshanzel: The other one is the existing layout we have. For example:

Screenshot 2024-04-25 at 6 20 14 PM

sshanzel:
> In between some screens the squad details post permissions and below, info stay centered - maybe we should fix to stay left. Not a big issue though.

They look okay to me, are you maybe looking at the production environment instead of the preview? They kinda change URLs when visiting a Squad.

Screenshot 2024-04-25 at 6 22 21 PM


Screenshot 2024-04-25 at 6 23 08 PM

nensidosari:
> They look okay to me, are you maybe looking at the production environment instead of the preview? They kinda change URLs when visiting a Squad.

Nope I use the url redirector that helps me with this 😄 , did u try it in the screen size I had above 646px...something like this, it is just in specific part, so that's why I said not much important anyways

packages/shared/src/components/fields/form/FormWrapper.tsx
nensidosari: No not what I ment, what I'm seeing is, we either show 
- left arrow
- left arrow and title
- cancel button

so `isTitleBesideCancel` naming think gives another sense as we show cancel button and title, or do we call the back arrow cancel?

packages/shared/src/components/fields/form/FormWrapper.tsx
nensidosari: Cool thank u, commented above about what I ment https://github.com/dailydotdev/apps/pull/3043#discussion_r1579357147

packages/shared/src/components/fields/form/FormWrapper.tsx
sshanzel: Right good point. It would probably be weird if we have a "Cancel" and a title. But yes, we do call the back arrow as "Cancel". I will update it to make this more straightforward.

packages/shared/src/components/squads/Details.tsx
sshanzel: Yep. The screenshots I attached were taken after the changes.

sshanzel:
> Nope I use the url redirector that helps me with this 😄 , did u try it in the screen size I had above 646px...something like this, it is just in specific part, so that's why I said not much important anyways

I see. After checking the config, it seems that falls under mobile now, not sure if we should still consume full-width on that size.

packages/shared/src/components/fields/form/FormWrapper.tsx
sshanzel: Let me know if the changes here make it a bit more understandable https://github.com/dailydotdev/apps/pull/3043/commits/a34311396b8460a0c369c170bb9efb6140bdeeb7

nensidosari: I think still the naming is a bit confusing, since its more related to title, what about `hasHeaderTitle` or `isHeaderTitle`?

nensidosari (approved):
Looks good, left one last comment on the title thing naming but non blocking

packages/shared/src/components/fields/form/FormWrapper.tsx
sshanzel: Updated. It is indeed much clearer now with `isHeaderTitle`.

PR #3048, opened 2024-04-25
feat: add feature to only have one boot popup per day

rebelchris:
- This integrates all existing boot popups into one hook so we can better maintain it
- It also ensures only one is shown per day, and we keep track of it in one place

Related API:
https://github.com/dailydotdev/daily-api/pull/1855

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done

### Preview domain
https://feat-boot-popups.preview.app.daily.dev


packages/shared/src/components/modals/useBootPopups.tsx
```diff
@@ -0,0 +1,179 @@
...
+  // hide modal if settings are not loaded
+  let shouldHideStreaksModal = !loadedSettings;
+  // hide modal if feature is not enabled or user opted out
+  shouldHideStreaksModal =
+    shouldHideStreaksModal || !isEnabled || optOutWeeklyGoal;
+  // hide modal if user already closed it
+  shouldHideStreaksModal =
+    shouldHideStreaksModal || alerts?.showStreakMilestone !== true;
+  // hide modal if there's no streak
+  shouldHideStreaksModal = shouldHideStreaksModal || !streak?.current;
```
nensidosari, line 46: I prefer better variable naming rather then comments as they can easily become outdated. I think naming only few of them differently and leaving the other ones as they are seems enough for the conditions to be fully understood.

```suggestion
  const streakModalClosed = alerts?.showStreakMilestone !== true
  const noCurrentStreak = !streak?.current

  let shouldHideStreaksModal = !loadedSettings || !isEnabled || optOutWeeklyGoal || streakModalClosed || noCurrentStreak;
```

packages/shared/src/graphql/alerts.ts
```diff
@@ -12,6 +12,8 @@ export type Alerts = {
   lastBanner?: string;
   banner?: boolean;
   showStreakMilestone?: boolean;
+  lastBootPopup?: Date;
+  bootPopup?: boolean;
```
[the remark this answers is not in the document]
rebelchris: Because API uses the same date structure.
If we rely on apps it can use a different date comparison to what's stored.

It's a common practice we follow for the other dates as well basically.

packages/shared/src/components/modals/useBootPopups.tsx
rebelchris: This was already made like this from the previous thing/
If I remember someone back then requested @kopancek to do it like this.
And want to avoid changing too much logic in favor of ensuring same evaluations as much as possible.

nensidosari (approved):
Looks good

packages/shared/src/components/MainLayout.tsx
```diff
@@ -101,28 +90,7 @@ function MainLayoutComponent({
...
-              event_name: AnalyticsEvent.Impression,
-              target_type: TargetType.MarketingCtaPopover,
-              target_id: marketingCta.campaignId,
-            });
-          },
-        },
-      });
-    }
-  }, [marketingCta, openModal, trackEvent]);
+  useBootPopups();
```
capJavert, line 93: As we talked this should be a component we render to avoid re-renders due to `useBootPopups` dependencies, for example in place where `ReputationPrivilegesModalTrigger` was.

packages/shared/src/components/modals/referral/GenericReferralModal.tsx
```diff
@@ -10,27 +16,31 @@ import ReferralSocialShareButtons from '../../widgets/ReferralSocialShareButtons
...
-    trackEvent({
-      event_name: AnalyticsEvent.Impression,
-      target_type: TargetType.ReferralPopup,
-    });
-
-    // @NOTE see https://dailydotdev.atlassian.net/l/cp/dK9h1zoM
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, []);
+    if (!isTracked.current) {
+      updateLastReferralReminder();
```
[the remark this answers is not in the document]
rebelchris: Previously it was in the useEffect below evaluation (so only once)
Just wanted to ensure it only happened once in this version.

packages/shared/src/components/MainLayout.tsx
rebelchris: What would we render as there is no UI?

capJavert: Just return null from component.

packages/shared/src/contexts/AlertContext.tsx
```diff
@@ -93,20 +97,39 @@ export const AlertContextProvider = ({
     },
   );
 
+  const { mutateAsync: updateLastBootPopup } = useMutation(
+    () => request(graphqlUrl, UPDATE_LAST_BOOT_POPUP),
+    {
+      onMutate: () =>
+        updateAlerts({
+          ...alerts,
+          lastBootPopup: new Date(),
```
nensidosari, line 106: Do we still need this new Date() here?

nensidosari (approved):
Looks good, only one comment

packages/shared/src/contexts/AlertContext.tsx
rebelchris: Its the optimistic update, lets leave it for now.
Oh we should actually add the state itself then this doesnt matter even 👍

capJavert: It is just optimistic update I think.

packages/shared/src/components/modals/BootPopups.tsx
```diff
@@ -0,0 +1,178 @@
...
+      !loadedAlerts ||
+      !alerts?.bootPopup ||
+      bootPopups.size === 0 ||
+      shownModal
+    ) {
+      return;
+    }
+    openModal(bootPopups.values().next().value);
+    updateLastBootPopup();
+    setShownModal(true);
```
capJavert, line 167: Do you need this state, seems extra with all other conditions?

capJavert (approved):

packages/shared/src/hooks/useActions.ts
```diff
@@ -39,7 +40,8 @@ export const useActions = (): UseActions => {
     },
     { enabled: !!user, ...disabledRefetch },
   );
-  const isActionsFetched = !isLoading;
+
+  const isActionsFetched = !!user && isSuccess;
```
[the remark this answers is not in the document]
capJavert: I still see it triggered as it was with `isLoading`. I think the fix with out own state in this case is the quickest we can do.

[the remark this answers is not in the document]
rebelchris: Sorry what do you mean should be the fix?

packages/shared/src/components/modals/BootPopups.tsx
rebelchris: Its more a extra for when no refresh is triggered to avoid showing a next one.
Felt more safe just in case, you reckon its overkill?

capJavert: More of an extra state. We can just do `setBootPopups(new Map())` to clear everything.

packages/shared/src/hooks/useActions.ts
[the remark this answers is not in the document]
capJavert: The one you did this morning. Just `useState` and then set it when fetch actually passed inside queryFn.

PR #3050, opened 2024-04-25
feat: support easter egg themes

kopancek:
## Changes

A draft PR, so that we can verify in preview if the behavior is as we expect.

Added support for easter egg themes. This requires a feature flag in growthbook `easter_egg_theme` to be defined. Example below for the starwars theme:
```json
{
  "version": 1,
  "light": {
    "logo": "https://daily-now-res.cloudinary.com/image/upload/s--7RocpMdi--/q_auto/v1714401089/feature-theme/sw-logo-light",
    "logoText": "https://daily-now-res.cloudinary.com/image/upload/s--KEtXuZmE--/q_auto/v1714401089/feature-theme/sw-logo-text-light",
    "body": {
      "background": "url('https://res.cloudinary.com/daily-now/image/upload/s--AyMCeVaQ--/f_auto/v1713647013/public/Star%20wars%20bg') 50% / contain, var(--theme-background-default)",
      "background-blend-mode": "difference, normal"
    },
    "navbar": {
      "border-color": "var(--theme-surface-primary)"
    }
  },
  "dark": {
    "logo": "https://daily-now-res.cloudinary.com/image/upload/s--M8Oh5ojp--/q_auto/v1714401089/feature-theme/sw-logo-dark",
    "logoText": "https://daily-now-res.cloudinary.com/image/upload/s--ovb1rR2S--/q_auto/v1714401089/feature-theme/sw-logo-text-dark",
    "body": {
      "background": "url('https://res.cloudinary.com/daily-now/image/upload/s--AyMCeVaQ--/f_auto/v1713647013/public/Star%20wars%20bg') 50% / contain, var(--theme-background-default)"
    },
    "navbar": {
      "border-color": "#fae610"
    }
  },
  "cursor": "https://daily-now-res.cloudinary.com/image/upload/s--HrncIFqV--/q_auto/v1714401157/feature-theme/cursor_purple"
}
```

So we basically have these fields:
```typescript
// defines light or dark version of the theme
interface ThemeDef {
  logo?: string; // URL that points to the SVG of the daily.dev logo
  logoText?: string; //  URL that points to the SVG of the daily.dev logo text
  body?: Object; // list of style properties and their values, this will be applied directly to the `id="__next"` element
  navbar?: string; // string className in tailwind, that will be applied to the navbar, in the example above, border color
}

interface EasterEggThemeSchema {
  version: number; // version of the feature flag value
  light: ThemeDef;
  dark: ThemeDef;
  cursor?: string;  // URL or Base64 encoded cursor image
}
```

To support loading SVGs from URL, I needed to add a new dependency on `react-inlinesvg` library:
- https://github.com/gilbarbara/react-inlinesvg

However, I probably did not add it the right way, because the extension refuses to build now 😭 

### TODO
- [x] Make sure it works in extension
- [ ] Test the 💩 out of it

- [X] MobileL (420px)
- [X] Tablet (656px)
- [X] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

MI-329 #done

### Preview domain
https://mi-329-star-wars-theme.preview.app.daily.dev


kopancek:
> Just wondering this would still mean we would need to add this specific color to webapp/extension first right?
> 
> ```
>     navbarBorder: 'border-[var(--theme-starwars-primary)]',
> ```

No, we define the CSS variable in the `body` definition and attach it to the HTML element in runtime. The `body` config  accepts any object with style properties. So it is pretty powerful and will work in extension as well as webapp.

packages/extension/webpack.config.js
```diff
@@ -53,6 +53,7 @@ const baseConfig = {
     extensions: ['.svg', '.ts', '.tsx', '.js', '.json'],
     alias: {
       'react-onesignal': false,
+      'react/jsx-runtime': 'react/jsx-runtime.js',
```
rebelchris, line 56: Why do you need the alias here?

packages/shared/src/components/feeds/FeedNav.tsx
```diff
@@ -50,6 +51,10 @@ function FeedNav(): ReactElement {
     [0, 1],
     DEFAULT_ALGORITHM_INDEX,
   );
+  const scrollClassName = useScrollTopClassName({
+    scrolledClassName: 'bg-transparent',
```
rebelchris, line 55: This should affect everything already?

packages/shared/src/components/layout/MainLayoutHeader.tsx
```diff
@@ -99,7 +106,7 @@ function MainLayoutHeader({
       <SearchPanel
         className={{
           container: classNames(
-            'left-0 top-0 z-header mx-auto bg-background-default py-3 tablet:left-16 laptop:left-0 laptop:bg-transparent',
+            'left-0 top-0 z-header mx-auto flex-1 items-center bg-background-default py-3 tablet:left-16 laptop:left-0 laptop:bg-transparent',
```
rebelchris, line 109: We'll just have to make sure this items-center doesn't add edge case for any of the renders inside. (mobile/desktop)

packages/shared/src/components/search/SearchPanel/SearchPanel.tsx
```diff
@@ -144,7 +144,7 @@ export const SearchPanel = ({ className }: SearchPanelProps): ReactElement => {
       >
         <SearchPanelInput
           className={{
-            container: classNames('w-full laptop:w-[35rem]'),
+            container: classNames('w-full laptop:max-w-[35rem]'),
```
rebelchris, line 147: Isn't this supposed to be a fixed width?

packages/shared/src/hooks/utils/useEasterEggTheme.ts
```diff
@@ -0,0 +1,67 @@
...
+    return themeMode === ThemeMode.Light;
+  }, [themeMode]);
+  const theme = isLight ? ThemeMode.Light : ThemeMode.Dark;
+
+  useLayoutEffect(() => {
+    const elem = globalThis?.document?.body;
+
+    // reset body styles
+    if (elem?.style?.cssText) {
+      elem.style.cssText = '';
```
rebelchris, line 38: Why is this needed?

packages/shared/src/lib/featureManagement.ts
```diff
@@ -17,6 +17,34 @@ export class Feature<T extends JSONValue> {
   }
 }
 
+// TODO: keeping this here temporarily, so that we can see how it works in preview
```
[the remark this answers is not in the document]
rebelchris: Also we can just set it in GB for team members :)

rebelchris (changes requested):
Nice work!
Mainly some questions and the one blocker basically.

packages/shared/src/components/Logo.tsx
```diff
@@ -19,6 +20,22 @@ const logoPositionToClassName: Record<LogoPosition, string> = {
...
+  fallback: typeof LogoText | typeof LogoIcon;
+}
+
+const LogoSvgElem = ({
+  className,
+  src,
+  fallback: FallbackElem,
+}: LogoSvgElemProps): ReactElement => {
+  if (src) {
+    return <SVG src={src} className={className} />;
```
capJavert, line 35: Do we need a lib, we can just use `` to load it up?

packages/shared/src/components/search/SearchPanel/SearchPanel.tsx
kopancek: If it's fixed width, the navbar overflows on smallest laptop screens with the starwars logo. Because the starwars logo is wider.

packages/shared/src/hooks/utils/useEasterEggTheme.ts
```diff
@@ -0,0 +1,67 @@
...
+
+    if (!useTheme) {
+      return;
+    }
+
+    if (easterEggTheme?.[theme]?.body) {
+      Object.entries(easterEggTheme[theme]?.body).forEach(([key, value]) => {
+        elem?.style?.setProperty(key, value);
+      });
+    }
```
capJavert: Maybe more efficient solution would be to compute style tag and then set it to `elem.style` in one go. Like this you get a global repaint on each key iteration which can be quite taxing.

capJavert (commented):
Interesting concept 👍 Some potential adjustments.

packages/shared/src/components/feeds/FeedNav.tsx
kopancek: What do you mean by affect everything already? :)

packages/shared/src/components/Logo.tsx
kopancek: Not the way we do it today. `<img` element does not allow us to change the fill or stroke color with CSS, as we do now.

packages/shared/src/hooks/utils/useEasterEggTheme.ts
kopancek: Good idea, let me give it a try.

packages/shared/src/components/Logo.tsx
capJavert: For what do we use that?

packages/shared/src/components/feeds/FeedNav.tsx
rebelchris: I mean should this affect everything even when your not in experiment?

packages/shared/src/components/search/SearchPanel/SearchPanel.tsx
rebelchris: I'm just afraid max-w will introduce some issues, pretty sure we had a issue with this exact line not long ago 😅

kopancek:
We just had a quick sync with Chris and Nensi. The result was, that I'll refactor a few bits:
- we will change the CSS variable name
- we will upload 4 SVG files to cloudinary with fixed colors
  - we can get rid of the external dependency that was introduced if we do this
  - in the future, if the need comes, we can still reintroduce this mechanism

packages/extension/webpack.config.js
kopancek: Since we are going to remove the dependency, I'm removing this change as well.

packages/shared/src/components/feeds/FeedNav.tsx
kopancek: It does not have any effect, since there's nothing behind the navbar. Once you scroll, the default bg will kick in. But if we want to be super safe, I can add it behind the experiment.

packages/shared/src/components/search/SearchPanel/SearchPanel.tsx
kopancek: Not sure what else can be done here, other than changing the size to be smaller?

packages/shared/src/components/Logo.tsx
kopancek: Replacing with an `<img` element, based on latest sync with Chris and Nensi

packages/shared/src/hooks/useScrollTopClassName.ts
```diff
@@ -0,0 +1,27 @@
...
+import { useScrollTopOffset } from './useScrollTopOffset';
+
+interface UseScrollTopStyleProps {
+  scrolledClassName?: string;
+  defaultClassName?: string;
+  scrollProperty?: 'scrollTop' | 'scrollY';
+  offset?: number;
+}
+
+export const useScrollTopClassName = ({
```
nensidosari, line 11: We seem to depend on scroll for setting the background to either default or transparent, but curious why do we need this?

packages/shared/src/components/search/SearchPanel/SearchPanel.tsx
nensidosari: Maybe we can change the size to something smaller if the feature is on on smaller screens?

packages/shared/src/hooks/utils/useFeatureTheme.ts
```diff
@@ -0,0 +1,64 @@
...
+  navbar?: string;
+  cursor?: string;
+}
+
+export const useFeatureTheme = (): UseFeatureThemeResult | undefined => {
+  const { themeMode } = useSettingsContext();
+  const featureTheme = useFeature(feature.featureTheme);
+  const router = useRouter();
+  const isOnboarding = router?.pathname?.startsWith('/onboarding') ?? false;
+  const useTheme = !isOnboarding && !!featureTheme;
```
nensidosari, line 21: Are u expecting for the featureTheme to be null/undefined as default with this evaluation?

packages/shared/src/lib/featureManagement.ts
```diff
@@ -48,6 +74,7 @@ const feature = {
     'experience_level',
     ExperienceLevelExperiment.Control,
   ),
+  featureTheme: new Feature('feature_theme', eTheme),
```
nensidosari, line 77: Default value should be `{}`.

packages/shared/src/hooks/utils/useFeatureTheme.ts
nensidosari: Growthbook does not allow that for JSON feature values, so you should adopt for a default value of `{}`

packages/shared/src/lib/featureManagement.ts
nensidosari: https://app.growthbook.io/features/feature_theme growthbook created 

default value: {}
for team it is set to the star wars config

nensidosari (changes requested):
GB values need to be adopted for a default value of {}

packages/shared/src/components/layout/MainLayoutHeader.tsx
kopancek: Switch to make the search panel smaller on laptop screens and added another breakpoint for laptopL. That way, we do not need to make changes in the layout here.

packages/shared/src/components/search/SearchPanel/SearchPanel.tsx
kopancek: Did just that, on laptop breakpoint new width is set to 29.5 rem (the max width until the layout starts overflowing) and on laptopL I kept the original width of 35 rem.

packages/shared/src/hooks/utils/useEasterEggTheme.ts
kopancek: Switched to adding a stylesheet element to the page head instead. Also, the content of the stylesheet is compared to the existing stylesheet added previously, so we do not replace the existing stylesheet with the same one, avoiding repaint of the same styles.

packages/shared/src/hooks/useScrollTopClassName.ts
kopancek: The designs call for a transparent navbar. But when you start scrolling, the content goes below the navbar. Which looks ugly and makes the navbar less readable IMO. 

Try setting the navbar background to transparent !important on preview to see how it behaves if we do not have this code.

packages/shared/src/hooks/utils/useFeatureTheme.ts
kopancek: Didn't know that, good catch. Will rewrite to make it work with the default growthbook forces on us.

packages/shared/src/lib/featureValues.ts
```diff
@@ -21,3 +21,18 @@ export enum ExperienceLevelExperiment {
...
+  body?: Record<string, string>;
+  navbar?: string;
+}
+
+export interface FeatureTheme {
+  version?: number;
+  light?: FeatureThemeVariant;
+  dark?: FeatureThemeVariant;
+  cursor?: string;
+  [key: string]: number | FeatureThemeVariant | string | undefined;
```
nensidosari, line 37: Why is this here?

rebelchris (approved):
Nothing blocking code wise, just the change we'll need in navbar :)

nensidosari (approved):
Looks good on my side, one small last comment

packages/shared/src/lib/featureValues.ts
kopancek: typescript would not build, claiming that the type is not string keyed and therefore not compatible with whatever growthbook returns as JSONValue.

packages/shared/src/hooks/utils/useFeatureTheme.ts
```diff
@@ -0,0 +1,72 @@
...
+    if (featureTheme?.cursor) {
+      bodyStyles.push(`cursor: url('${featureTheme.cursor}'), auto; }`);
+    }
+    const body = `body { ${bodyStyles.join('; ')} }`;
+
+    const oldStyleElem = document.getElementById(id);
+    if (oldStyleElem?.innerText === body) {
+      // nothing to do, styles were alreay applied
+      return;
+    }
```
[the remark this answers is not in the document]
kopancek: If we add alreadyInjected, the only thing we know is, that some stylesheet was already injected. If there was a change in the growthbook theme definition and resulting stylesheet.

So IMO, we need to do a string comparison anyway. Adding it to state would be an unneeded extra step.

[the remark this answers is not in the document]
rebelchris: Got it, let's leave it for now :)

packages/webapp/components/layouts/AccountLayout/index.tsx
```diff
@@ -81,7 +84,10 @@ export default function AccountLayout({
           />
         ) : (
           <SidebarNav
-            className="absolute z-3 ml-auto h-full w-full border-l border-border-subtlest-tertiary bg-background-default tablet:relative tablet:w-[unset]"
+            className={classNames(
+              'absolute z-3 ml-auto h-full w-full border-l border-border-subtlest-tertiary bg-background-default tablet:relative tablet:w-[unset]',
```
nensidosari, line 88:
```suggestion
              'absolute z-3 ml-auto h-full w-full border-l border-border-subtlest-tertiary tablet:relative tablet:w-[unset]',
```

nensidosari (approved):
Just one fix. Looks great!

PR #3056, opened 2024-04-26
fix: small boot prompt rules

rebelchris:
- Fixes two things:
- 1. We use the same rule per prompt on when it should set the lastBootPopup (for some it's on show, some on close)
- 2. There was a issue with useActions hook where potentially it could return `isLoading: false` while no data was loaded, it's caused by react query mutations making isLoading to be false and data to exist.
- To solve 2 I introduced a internal state return that indicates whether data is loaded from the server, it should fix some edge-cases but not impact the existing hasAction check that happens for mutated data.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done

### Preview domain
https://fix-boot-popup-rules.preview.app.daily.dev

### DIFF
--- a/packages/shared/src/hooks/useActions.ts
+++ b/packages/shared/src/hooks/useActions.ts
@@ -17,49 +17,62 @@ interface UseActions {
   isActionsFetched: boolean;
 }
 
+interface ActionQueryData {
+  actions: Action[];
+  serverLoaded: boolean;
+}
+
 export const useActions = (): UseActions => {
   const client = useQueryClient();
   const { user } = useAuthContext();
   const actionsKey = generateQueryKey(RequestKey.Actions, user);
 
-  const { data: actions, isLoading } = useQuery(
+  const { data, isLoading } = useQuery(
     actionsKey,
     async () => {
-      const data = await getUserActions();
-      const current = client.getQueryData<Action[]>(actionsKey);
+      const serverData = await getUserActions();
+      const current = client.getQueryData<ActionQueryData>(actionsKey);
 
       if (!current || !Array.isArray(current)) {
-        return data;
+        return { actions: serverData, serverLoaded: true };
       }
 
-      const filtered = data.filter(({ type }) =>
+      const filtered = serverData.filter(({ type }) =>
         current.every((action) => action.type !== type),
       );
 
-      return [...current, ...filtered];
+      return { actions: [...current, ...filtered], serverLoaded: true };
     },
     { enabled: !!user, ...disabledRefetch },
   );
 
-  const isActionsFetched = !isLoading;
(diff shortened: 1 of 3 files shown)


packages/shared/src/hooks/useActions.ts
```diff
@@ -17,49 +17,62 @@ interface UseActions {
...
 
-      return [...current, ...filtered];
+      return { actions: [...current, ...filtered], serverLoaded: true };
     },
     { enabled: !!user, ...disabledRefetch },
   );
 
-  const isActionsFetched = !isLoading;
+  const actions = data?.actions;
+  const isActionsFetched = !isLoading && data?.serverLoaded;
```
capJavert, line 50: Cast to `boolean` as `!!data?.serverLoaded` because it would return `undefined` from `isActionsFetched`?

packages/shared/src/hooks/useActions.ts
```diff
@@ -17,49 +17,62 @@ interface UseActions {
...
+          serverLoaded: old?.serverLoaded,
+        };
       });
 
-      return () => client.setQueryData<Action[]>(actionsKey, actions);
+      return () =>
+        client.setQueryData<ActionQueryData>(actionsKey, {
+          actions,
+          serverLoaded: data?.serverLoaded,
+        });
```
capJavert, line 75: same here https://github.com/dailydotdev/apps/pull/3056/files#r1580852940

packages/shared/src/hooks/useActions.ts
rebelchris: Yep good spot!

packages/shared/src/hooks/useActions.ts
```diff
@@ -17,49 +17,62 @@ interface UseActions {
...
-        if (!Array.isArray(data)) {
-          return [optimisticAction];
+        if (!Array.isArray(old?.actions)) {
+          return { actions: [optimisticAction], serverLoaded: false };
         }
 
-        return [...data, optimisticAction];
+        return {
+          actions: [...old?.actions, optimisticAction],
+          serverLoaded: old?.serverLoaded,
```
capJavert, line 67: @rebelchris here as well 🙏

capJavert (approved):
Just one more comment, nice work!

packages/shared/src/hooks/useActions.ts
rebelchris: Ah for some reason thought this would just return something from the main query.
We don't really use this direct, but happy to apply it here.

PR #3069, opened 2024-05-02
fix: do not update streak cache on posts that were read already

kopancek:
## Changes

Do not update streak query cache when post is viewed, but was already read before.

Related issue:
https://dailydotdev.atlassian.net/browse/MI-230

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

MI-230 #done

### Preview domain
https://mi-230-fix-streak-cache-read-posts.preview.app.daily.dev

### DIFF
--- a/packages/shared/src/hooks/post/useViewPost.ts
+++ b/packages/shared/src/hooks/post/useViewPost.ts
@@ -3,17 +3,15 @@ import {
   useMutation,
   useQueryClient,
 } from '@tanstack/react-query';
-import { sendViewPost } from '../../graphql/posts';
+import { Post, sendViewPost } from '../../graphql/posts';
 import { generateQueryKey, RequestKey } from '../../lib/query';
 import { useAuthContext } from '../../contexts/AuthContext';
 import { ReadingDay, UserStreak } from '../../graphql/users';
 import { getTodayTz } from '../../lib/dateFormat';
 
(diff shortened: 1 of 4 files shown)


packages/shared/src/hooks/post/useViewPost.ts
```diff
@@ -3,40 +3,40 @@ import {
...
-          streak.max += 1;
-          streak.total += 1;
-          streak.current += 1;
-        }
-
-        streak.lastViewAt = new Date();
-
-        return streak;
-      });
+      if (!post.read) {
```
nensidosari, line 22: Just checking that this does not become read once user clicks on the post to read, thus disabling the cache to update in any case (if the post is being read now, or was read before)

nensidosari (approved):
Looks good, just one check. Make sure to test with a read post and non read post.

sshanzel (approved):
Looks good to me 🚀

packages/shared/src/hooks/post/useViewPost.ts
kopancek: I tested this manually and it works as expected. E.g. if the post was not read before, it will update the cache correctly.

PR #3075, opened 2024-05-03
fix: streak popup updating with a delay

kopancek:
## Changes

Related slack thread: https://dailydotdev.slack.com/archives/C01L0QXQJNB/p1714142180431059

Several issues found with streaks popup have been fixed:
- if user had browser tab opened after date change (midnight), popup would never update correctly, because the streakDays were calculated on the module
- memoization of streaks was not done correctly, because it used the consts defined on the module (as mentioned above)
- `getTodayTz` was giving incorrect dates, e.g. '2024-05-03T13:12:11+02:00' would result in `2024-05-02T00:00:00` on my machine, resulting in -1 day skew
- switched to invalidating of the query for streak history and rely on API call being done instead of calculating and filling the cache on the UI side

## Events

No new events

- [X] MobileL (420px)
- [X] Tablet (656px)
- [X] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

MI-333 #done

### Preview domain
https://mi-333-fix-streaks-delay.preview.app.daily.dev

### DIFF
--- a/packages/shared/src/components/streak/popup/ReadingStreakPopup.tsx
+++ b/packages/shared/src/components/streak/popup/ReadingStreakPopup.tsx
@@ -1,23 +1,43 @@
-import React, { ReactElement, useMemo, useState } from 'react';
+import React, { ReactElement, useMemo } from 'react';
 import { addDays, isSameDay, subDays } from 'date-fns';
 import { useQuery } from '@tanstack/react-query';
 import classNames from 'classnames';
 import { StreakSection } from './StreakSection';
 import { DayStreak, Streak } from './DayStreak';
 import { generateQueryKey, RequestKey, StaleTime } from '../../../lib/query';
-import { getReadingStreak30Days, UserStreak } from '../../../graphql/users';
+import {
+  getReadingStreak30Days,
+  ReadingDay,
+  UserStreak,
+} from '../../../graphql/users';
 import { useAuthContext } from '../../../contexts/AuthContext';
 import { Weekends } from '../../../lib/dateFormat';
 
 const getStreak = ({
-  isCompleted,
-  isToday,
-  isFreezeDay,
+  value,
+  today,
+  dateToday,
+  history,
 }: {
-  isCompleted: boolean;
-  isToday: boolean;
-  isFreezeDay: boolean;
-}) => {
+  value: Date;
+  today: Date;
+  dateToday: number;
+  history?: ReadingDay[];
+}): Streak => {
+  const day = value.getDay();
+  const date = value.getDate();
+  const isFreezeDay = Weekends.includes(day);
+  const isToday = date === dateToday;
+  const isFuture = value > today;
+  const isCompleted =
+    !isFuture &&
+    history?.some(({ date: historyDate, reads }) => {
+      const dateToCompare = new Date(historyDate);
+      const sameDate = isSameDay(dateToCompare, value);
+
+      return sameDate && reads > 0;
+    });
+
   if (isCompleted) {
     return Streak.Completed;
   }
@@ -34,8 +54,7 @@ const getStreak = ({
 };
 
 const getStreakDays = (today: Date) => {
-  const dateToday = today.getDate();
-  const streakDays = [
+  return [
     subDays(today, 4),
     subDays(today, 3),
     subDays(today, 2),
@@ -46,8 +65,6 @@ const getStreakDays = (today: Date) => {
     addDays(today, 3),
     addDays(today, 4),
   ]; // these dates will then be compared to the user's post views
-
-  return { dateToday, streakDays };
 };
 
 interface ReadingStreakPopupProps {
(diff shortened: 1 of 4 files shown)


packages/shared/src/components/streak/popup/ReadingStreakPopup.tsx
```diff
@@ -9,19 +9,46 @@ import { getReadingStreak30Days, UserStreak } from '../../../graphql/users';
...
+  if (isFreezeDay) {
+    return Streak.Freeze;
+  }
+
+  if (isToday) {
+    return Streak.Pending;
+  }
+
+  return Streak.Upcoming;
+};
```
capJavert, line 34: Just a note, might be wrong, but this looks like a additional state transformation that does not do a lot. Maybe getStreak could actually contain [below logic](https://github.com/dailydotdev/apps/pull/3075/files#diff-75b6b950d2d429c84bce12f44b6b3980637e4c95ff04ce893869b73ea551950aR82-R94) and then return `Streak` enum.

packages/shared/src/hooks/post/useViewPost.ts
```diff
@@ -41,17 +40,8 @@ export const useViewPost = (
...
-
-        if (index === -1) {
-          return client.setQueryData(readKey, [...reading, { date, reads: 1 }]);
-        }
-
-        const updatedReading = [...reading];
-        updatedReading[index].reads += 1;
-        return client.setQueryData(readKey, updatedReading);
+        // just mark the query as stale
+        await client.invalidateQueries(readKey);
```
capJavert, line 44: I like this approach 👍

capJavert (commented):
Agree with this direction.

packages/shared/src/components/streak/popup/ReadingStreakPopup.tsx
```diff
@@ -33,52 +60,52 @@ export function ReadingStreakPopup({
   fullWidth,
 }: ReadingStreakPopupProps): ReactElement {
   const { user } = useAuthContext();
+  const [today, setToday] = useState(new Date());
```
capJavert, line 63: Does this need to be state? I think you can just always instantiate it as `const today = new Date()` inside `useMemo`.

packages/shared/src/components/streak/popup/ReadingStreakPopup.tsx
```diff
@@ -33,52 +60,52 @@ export function ReadingStreakPopup({
...
-          return Streak.Completed;
-        }
+  // only change today and streakDays if the current day changed to the next one
+  if (!isSameDay(new Date(), today)) {
+    setToday(new Date());
+  }
+  const { dateToday, streakDays } = useMemo(
+    () => getStreakDays(today),
+    [today],
+  );
```
capJavert, line 77: Also in addition to https://github.com/dailydotdev/apps/pull/3075/files#r1589621107, we can also move this whole evaluation to below `useMemo`, no need to keep it separate since both memoizations have the same dependency `today.`

packages/shared/src/hooks/post/useViewPost.ts
rebelchris: Generally we always try to do optimistic manipulation historically to avoid extra query invocation when we know the result already.
Just a note on why we do manual cache setting, not against this 👍

rebelchris (commented):
Ill do another check on Monday bit deeper

packages/shared/src/hooks/post/useViewPost.ts
kopancek: In this case, we only make the query if data is stale and when the reading streak dropdown is open. I think it should be fine. If we don't open the reading streak popup in 5 minutes, the data will be marked as stale anyway. Which I think would do the optimistic manipulation needlessly in most cases.

Because the logic was so convoluted, it was actually returning the wrong data :) I'd rather have a single point of truth in this case - the backend.

packages/shared/src/components/streak/popup/ReadingStreakPopup.tsx
kopancek: The `useMemo` called is memoizing based on the value of `today`. I could change it to memoize based on just the date portion of today, but not sure if it would actually work. Let me try :)

kopancek: OK, refactored into single `useMemo`, removed the `setState` hook on `today` and hopefully now the code is more readable.

Added a comment though to the `useMemo`, since now we need to explicitly omit the `today` dependency from it, otherwise it would recalculate on every render.

packages/shared/src/components/streak/popup/ReadingStreakPopup.tsx
```diff
@@ -33,52 +77,35 @@ export function ReadingStreakPopup({
...
-
-          return sameDate && reads > 0;
-        });
-
-      const getStreak = () => {
-        if (isCompleted) {
-          return Streak.Completed;
-        }
+  const today = new Date();
+  const dateToday = today.getDate();
```
capJavert, line 87:
```suggestion
  const dateToday = new Date().getDate();
```

packages/shared/src/components/streak/popup/ReadingStreakPopup.tsx
```diff
@@ -33,52 +77,35 @@ export function ReadingStreakPopup({
...
 
-        if (Weekends.includes(day)) {
-          return Streak.Freeze;
-        }
-
-        if (date === dateToday) {
-          return Streak.Pending;
-        }
+  const streaks = useMemo(() => {
+    const streakDays = getStreakDays(today);
```
capJavert, line 90:
```suggestion
    const today = new Date();
    const streakDays = getStreakDays(today);
```

capJavert (approved):
Looks good from my end, already read an article today, but eager to test tomorrow if it works better 👌

packages/shared/src/hooks/post/useViewPost.ts
rebelchris: Yeah makes total sense indeed :)

rebelchris (approved):
Nice cleanup :)

PR #3078, opened 2024-05-03
fix: back button now correctly refreshes route

capJavert:
- looks like close handler for modal introduced issue with back button not working because `onCloseModal` changed history to interfere with Next.js router
- back button now works when you are navigating between multiple tag pages by constantly selecting them eg. go to /tags/php (select tag) -> /tags/javascript -> back -> /tags/php (before URL would change but the page would stay the same as if you are on javascript page)

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done

### Preview domain
https://fix-back-button-on-feed-pages.preview.app.daily.dev

### DIFF
--- a/packages/shared/src/hooks/usePostModalNavigation.ts
+++ b/packages/shared/src/hooks/usePostModalNavigation.ts
@@ -107,11 +107,8 @@ export const usePostModalNavigation = (
       return null;
     }
 
-    const routeHandler = (newRoute: string) => {
-      if (newRoute.includes('/posts/')) {
-        return;
-      }
-      onCloseModal();
+    const routeHandler = () => {
+      onCloseModal(true);
     };
     router.events.on('routeChangeStart', routeHandler);
 


packages/shared/src/hooks/usePostModalNavigation.ts
```diff
@@ -107,11 +107,8 @@ export const usePostModalNavigation = (
       return null;
     }
 
-    const routeHandler = (newRoute: string) => {
-      if (newRoute.includes('/posts/')) {
-        return;
-      }
-      onCloseModal();
+    const routeHandler = () => {
+      onCloseModal(true);
```
capJavert, line 111: Even though I put true for `fromPopState` basically what it does:
- skips calling `changeHistory`
  - when this `routerHandler` is called Next.js always takes over routing to no need to worry about clearing history
- this is never called on our custom modal history changes so no need for `newRoute.includes` from my testing 
- back button now works when you are navigating between multiple tag pages by constantly selecting them eg. /tags/php -> /tags/javascript -> back -> /tags/php (before URL would change but the page would stay the same as if you are on javascript)

capJavert:
Think this fixes it with minimal changes but please test since there is a lot of edge cases.

rebelchris (approved):
Lets see definitely cleaner 😅

rebelchris:
Oh didnt we add the posts route check for the soft stuff on webapp push?
Where it would flicker sometimes?

capJavert:
> Oh didnt we add the posts route check for the soft stuff on webapp push?
Where it would flicker sometimes?

@rebelchris, actually we discussed here https://github.com/dailydotdev/apps/pull/2983#discussion_r1562425189, it was added as failsafe but it was not needed in the first place from what I see.

kopancek (approved):
Looks good to me, like the simplification of the code as well 👍

omBratteng (approved):

PR #3079, opened 2024-05-06
refactor: bookmark loops enrollment process

sshanzel:
## Changes
- Introduce a new feature flag named `bookmarkPopup`.
- When it is true, the bookmark on card should be visible as well, then once the action is performed, we should show the popup.

Discussion: https://dailydotdev.slack.com/archives/C02JAUF8HJL/p1714974297586559

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done


packages/shared/src/hooks/useBookmarkPost.ts
```diff
@@ -174,23 +174,19 @@ const useBookmarkPost = ({
...
-        const bookmarkLoops = getFeatureValue(feature.bookmarkLoops);
-
-        if (bookmarkLoops) {
-          completeAction(ActionType.BookmarkPromoteMobile);
-          openModal({
-            type: LazyModal.MarketingCta,
-            props: { marketingCta: promotion.bookmarkPromoteMobile },
-          });
-        }
+      if (!seenBookmarkPromotion && bookmarkPopup) {
```
rebelchris, line 177: Should it enroll everyone or only is !seenBookmarkPromotion?

rebelchris (commented):
Looks ok, is the enrollment supposed to happen on click only?

packages/shared/src/hooks/useBookmarkPost.ts
sshanzel: Right. We can do that too.

sshanzel: Ah, just realized, it wouldn't matter anyway since we load enroll this on the feed to show bookmark on card (yes we now have two flags for that, but this one includes popup).

packages/shared/src/lib/featureManagement.ts
```diff
@@ -39,6 +39,7 @@ const feature = {
   onboardingMostVisited: new Feature('onboarding_most_visited', false),
   shareExperience: new Feature('share_experience', false),
   bookmarkLoops: new Feature('bookmark_loops', false),
```
nensidosari, line 41: So are we using bookmarkLoops anymore? Looks like we aren't.. in this case, shall we remove or just not add another flag `bookmarkPopup` but keep `bookmarkLoops` and just adjust it's usage?

packages/shared/src/hooks/useBookmarkPost.ts
nensidosari: why did we have the need to change the way we get/use feature value here, seems the end result would be the same. This way we don't even doubt anything about enrolment anyway since we get the value after !seenBookmarkPromotion at least in this case even if it might not matter.

rebelchris: Yeah makes sense since they already got enrolled :)

sshanzel: @nensidosari we won't be using the old flag now. We should probably clean it up too.

nensidosari (approved):

rebelchris (approved):
Works for me!

rebelchris (approved):

PR #3081, opened 2024-05-06
fix: search enrollment event

rebelchris:
- Since this feature was created pre the new hooks and rules we didn't wait for enrollment
- This adjustment ensures we only enroll on query/search

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done

### Preview domain
https://as-296-search-enrollment.preview.app.daily.dev

### DIFF
--- a/packages/shared/src/components/MainFeedLayout.tsx
+++ b/packages/shared/src/components/MainFeedLayout.tsx
@@ -140,7 +140,10 @@ export default function MainFeedLayout({
     hasUser: !!user,
   });
   const feedVersion = useFeature(feature.feedVersion);
-  const searchVersion = useFeature(feature.searchVersion);
+  const { value: searchVersion } = useConditionalFeature({
+    feature: feature.searchVersion,
+    shouldEvaluate: isSearchOn && !!searchQuery,
+  });
   const { isUpvoted, isPopular, isSortableFeed } = useFeedName({ feedName });
   const {
     shouldUseMobileFeedLayout,
--- a/packages/shared/src/components/PostsSearch.tsx
+++ b/packages/shared/src/components/PostsSearch.tsx
@@ -67,7 +67,7 @@ export default function PostsSearch({
   const [items, setItems] = useState<string[]>([]);
   const { value: searchVersion } = useConditionalFeature({
     feature: feature.searchVersion,
-    shouldEvaluate: !!query,
+    shouldEvaluate: !!query && suggestionType === 'searchPostSuggestions',
   });
   const SEARCH_URL = SEARCH_TYPES[suggestionType];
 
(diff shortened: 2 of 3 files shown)


idoshamun (commented):
Why there are three different places of enrollment? Can you explain the need for each one please

packages/shared/src/components/MainFeedLayout.tsx
```diff
@@ -140,7 +140,10 @@ export default function MainFeedLayout({
     hasUser: !!user,
   });
   const feedVersion = useFeature(feature.feedVersion);
-  const searchVersion = useFeature(feature.searchVersion);
+  const { value: searchVersion } = useConditionalFeature({
+    feature: feature.searchVersion,
+    shouldEvaluate: isSearchOn && !!searchQuery,
+  });
```
rebelchris, line 146: Used for the actual search post page feed

packages/shared/src/hooks/search/useSearchProvider.ts
```diff
@@ -66,6 +66,8 @@ export const useSearchProvider = (): UseSearchProvider => {
           };
         }
 
+        const searchVersion = getFeatureValue(feature.searchVersion);
```
rebelchris, line 69: The suggestions query

packages/shared/src/components/PostsSearch.tsx
```diff
@@ -66,6 +65,10 @@ export default function PostsSearch({
     width: number;
   }>(null);
   const [items, setItems] = useState<string[]>([]);
+  const { value: searchVersion } = useConditionalFeature({
+    feature: feature.searchVersion,
+    shouldEvaluate: !!query,
+  });
```
rebelchris, line 71: The post search query inside the dropdown

packages/shared/src/components/MainFeedLayout.tsx
rebelchris: @idoshamun Added the 3 types as comments

idoshamun:
@rebelchris thanks for the info. can we add it only in a single place where we actually send the query? (so it's two places actually one for suggestion and one for the search itself). Similar to how you implemented suggestions

nensidosari (approved):
Looks good on my side if we want to keep the implementation as is

rebelchris:
@idoshamun Was busy refactoring, but the one is actually dedicated to posts search, we support multiple variants there in a dynamic way (reading history/bookmarks and posts) so it would become a big refactor if we want to combine with the other one.
My suggestion to leave it for now?

idoshamun:
@rebelchris ok. we'll be more careful with the analysis then to make sure the enrollment is right

idoshamun (approved):

rebelchris:
@idoshamun I gave the post only one stricter evaluation criteria so we don't enroll for bookmarks search and reading history search as they don't support version at the moment.
Does that work?

idoshamun:
yes thank you!

PR #3084, opened 2024-05-07
feat: fetch squads via graphql instead of from boot

kopancek:
## Changes

- added new useSquads hook
- removed all squad references from boot and AuthContext
- using the new useSquads hook everywhere where we used AuthContext squads before
- no visual changes

## Important

Without the API PR being merged and deployed, it will not work in preview or extension:
- https://github.com/dailydotdev/daily-api/pull/1879

## Events

No new events

- [X] MobileL (420px)
- [X] Tablet (656px)
- [X] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

MI-324 #done

### Preview domain
https://mi-324-do-not-use-squads-in-boot.preview.app.daily.dev


packages/shared/src/graphql/squads.ts
```diff
@@ -205,6 +210,30 @@ export const SQUAD_QUERY = gql`
   ${SOURCE_BASE_FRAGMENT}
 `;
 
+export const MY_SQUADS_QUERY = gql`
+  query MySquads {
+    mySourceMemberships(first: 100, type: squad) {
```
rebelchris, line 215: I think there's also a option to pass it as infinite maybe more failsafe (although don't think anyone but daphna will ever hit 100+ 🤣 )

packages/shared/src/hooks/squads/useSquads.ts
```diff
@@ -0,0 +1,88 @@
...
+    data: squads,
+    isLoading,
+    isFetched,
+    error,
+  } = useQuery<Squad[], ClientError>(
+    queryKey,
+    getCurrentUserSquads(requestMethod),
+    {
+      enabled: !!user?.id,
+      retry: false,
```
rebelchris, line 47: Why did you disable retry specifically?
We also have a wrapper for `disabledRefetch` which maybe you want?

packages/shared/src/hooks/useDeleteSquad.ts
```diff
@@ -22,7 +22,7 @@ export const useDeleteSquad = ({
 }: UseDeleteSquadProps): UseDeleteSquadModal => {
```
rebelchris, line 22: This whole part feels it could be part of useSquad, how did you feel about it looking at code?
(Don't need to refactor, just looking for your opinion as someone who touched the code)

rebelchris (approved):
Nice work!

Can we maybe do a bit longer internal test on this one?
I think we have to be careful with mobile sharing from various places/extension/companion adoption etc.

It's easy to have forgotten a place where squads where used potentially.
Code wise nothing much to add though 🙏

nensidosari (changes requested):
Looks great! Nothing to add 🚀 
Just blocking the PR until the api PR is merged so we can test before merging or to avoid any accidental merge.

packages/shared/src/graphql/squads.ts
kopancek: AFAIK, we limit to 100 by default on the API side anyway: https://github.com/dailydotdev/daily-api/blob/main/src/common/datePageGenerator.ts#L39
https://github.com/dailydotdev/daily-api/blob/main/src/schema/sources.ts#L878-L883

IMO, we cannot even show 100 items in the sidebar anyway, so having this paginated probably makes sense.

packages/shared/src/hooks/squads/useSquads.ts
kopancek: Bad copy paste from the `useSquad` hook. I think we can retry here.

kopancek:
> Can we maybe do a bit longer internal test on this one? I think we have to be careful with mobile sharing from various places/extension/companion adoption etc.
> 
> It's easy to have forgotten a place where squads where used potentially. Code wise nothing much to add though 🙏

Yup, for sure. I tried to test in extension/companion locally, but since it does not use local API endpoint I could only see that the requests are firing. (Hence why I kept the checkboxes on the PR descriptions unchecked).

Still waiting on the [CI to deploy the API changes](https://app.circleci.com/pipelines/github/dailydotdev/daily-api?branch=main).

packages/shared/src/hooks/useDeleteSquad.ts
kopancek: For me, the `useDeleteSquad` hook is more about showing a prompt confirmation when deleting a squad. So maybe we can rename it to `useDeleteSquadPrompt` instead. That way it should be clearer.

I'd rather refrain from refactoring the hook in this PR, since it's already touching a lot of code. But if you feel like we should have the functionality in useSquad hook, let's add a tech debt issue for that :)

nensidosari (approved):

PR #3085, opened 2024-05-07
fix: use useConditionalFeature to only enroll mobile users

nensidosari:
## Changes
- this experiment seemed to be affecting for bad the non mobile users, as we didn't find any reason why, we are making sure this experiment only enrols mobile users so we can see it's effect only on them. See thread for more info: https://dailydotdev.slack.com/archives/C02JAUF8HJL/p1714967983589999
- decision to do the change in thread [here](https://dailydotdev.slack.com/archives/C02JAUF8HJL/p1715067685797829)
- used conditional feature hook to only evaluate if it's mobile

### Preview domain
https://fix-enrollment-share-experience.preview.app.daily.dev

### DIFF
--- a/packages/shared/src/components/post/PostActions.tsx
+++ b/packages/shared/src/components/post/PostActions.tsx
@@ -12,7 +12,12 @@ import {
 import { Post, UserVote } from '../../graphql/posts';
 import { QuaternaryButton } from '../buttons/QuaternaryButton';
 import { PostOrigin } from '../../hooks/analytics/useAnalyticsContextData';
-import { useVotePost } from '../../hooks';
+import {
+  useConditionalFeature,
+  useViewSize,
+  useVotePost,
+  ViewSize,
+} from '../../hooks';
 import { Origin } from '../../lib/analytics';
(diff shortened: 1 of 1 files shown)


rebelchris (approved):
I could have sworn @idoshamun mentioned in his presentation mobile is one of the few things they CAN do on GB side.
Maybe it's good to align this as I'm pretty sure we expected this in the past for other experiments?

Otherwise the code seems good though.

nensidosari:
> I could have sworn @idoshamun mentioned in his presentation mobile is one of the few things they CAN do on GB side.
> Maybe it's good to align this as I'm pretty sure we expected this in the past for other experiments?
> 
> Otherwise the code seems good though.


Well yes I keep mentioning it, because we did add it for them to have available on GB side, hence I've mentioned many times in the conversation there with Dave...I'm fine doing like this for the moment but after this we'll need to align again, if this is the new default they want us to follow for mobile or any condition really (registrationDate that they also have there), cuz whenever it's this 2 I always assume it's going to be GB filtering.

PR #3091, opened 2024-05-10
chore: cleanup reading streak experiemnt

nensidosari:
Ready for review - in the meantime I'll test and tick the checkboxes when I've done all neccessary ones.

## Changes
- cleans up streak experiment since it won - see [thread](https://dailydotdev.slack.com/archives/C05TQL6HWTH/p1714540757518679)

- [x] MobileL (420px)
- [x] Tablet (656px)
- [x] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

### Preview domain
https://chore-cleanup-streaks.preview.app.daily.dev


packages/shared/src/components/modals/BootPopups.tsx
```diff
@@ -30,15 +30,14 @@ export const BootPopups = (): null => {
   const [bootPopups, setBootPopups] = useState(() => new Map());
   const { getMarketingCta } = useBoot();
   const marketingCta = getMarketingCta(MarketingCtaVariant.Popover);
-  const { shouldShowPopup: shouldShowStreaksPopup } = useStreakExperiment();
-  const { streak, isEnabled } = useReadingStreak();
+  const { streak, shouldShowPopup: shouldShowStreaksPopup } =
+    useReadingStreak();
```
capJavert, line 34: Do we need it called here if it is not used?

packages/shared/src/hooks/streaks/useReadingStreak.ts
```diff
@@ -2,26 +2,29 @@ import { useQuery } from '@tanstack/react-query';
...
+    { enabled: true, staleTime: StaleTime.Default },
   );
+  const { checkHasCompleted } = useActions();
 
   return {
     streak,
     isLoading,
-    isEnabled: shouldShowStreak,
+    shouldShowPopup:
+      user?.createdAt < '2024-03-14T00:00:00.000Z' &&
```
capJavert, line 27: What are we doing here?

capJavert (approved):
Nice cleanup, just a few questions. Approving either way. 👍

packages/shared/src/components/modals/BootPopups.tsx
nensidosari: Which in particular are you referring to? the ones we have now are all used

capJavert: My bad, it was just the formatting. Looked like `useReadingStreak` is called alone on new line.

packages/shared/src/hooks/streaks/useReadingStreak.ts
capJavert: Tnx for explanation!

omBratteng (approved):
More red than green. Me likey

packages/shared/src/hooks/streaks/useReadingStreak.ts
rebelchris: Are there even people still that where from before this date and haven't seen this yet? 🔱  (In which case we can also remove the modal?)

rebelchris (approved):
What a cleanup!
Can we also remove the modal that was assigned to the old ranks?
And maybe the settings interface for it? (might have missed this part now)

packages/shared/src/hooks/streaks/useReadingStreak.ts
nensidosari: Hmm will need to check, would be nice to remove

packages/shared/src/components/modals/BootPopups.tsx
```diff
@@ -30,15 +30,14 @@ export const BootPopups = (): null => {
...
   const { loadedSettings, optOutWeeklyGoal } = useSettingsContext();
```
kopancek, line 35: I think this PR actually removes `optOutWeeklyGoal` from the settings context. Should this have been removed or I just do not understand things in the morning? 😊

packages/shared/src/components/modals/BootPopups.tsx
```diff
@@ -30,15 +30,14 @@ export const BootPopups = (): null => {
...
 
   // hide modal if settings are not loaded
   let shouldHideStreaksModal = !loadedSettings;
   // hide modal if feature is not enabled or user opted out
-  shouldHideStreaksModal =
-    shouldHideStreaksModal || !isEnabled || optOutWeeklyGoal;
+  shouldHideStreaksModal = shouldHideStreaksModal || optOutWeeklyGoal;
```
kopancek, line 40: Same as above

kopancek (changes requested):
Looks good to me, with one exception, where I think we should not use the `optOutWeeklyGoal` from settings context anymore.

packages/shared/src/components/feeds/MobileFeedActions.tsx
```diff
@@ -26,17 +26,15 @@ export function MobileFeedActions(): ReactElement {
...
-        )}
+        <>
+          <ReadingStreakButton
+            isLoading={isLoading}
+            streak={streak}
+            compact
+            iconPosition={ButtonIconPosition.Right}
+          />
+          <Divider className="bg-border-subtlest-tertiary" vertical />
+        </>
```
sshanzel, line 37:
```suggestion
        <ReadingStreakButton
          isLoading={isLoading}
          streak={streak}
          compact
          iconPosition={ButtonIconPosition.Right}
        />
        <Divider className="bg-border-subtlest-tertiary" vertical />
```

packages/shared/src/hooks/streaks/useReadingStreak.ts
```diff
@@ -2,26 +2,29 @@ import { useQuery } from '@tanstack/react-query';
...
 }
 
 export const useReadingStreak = (): UserReadingStreak => {
   const { user } = useAuthContext();
-  const { shouldShowStreak } = useStreakExperiment();
   const { data: streak, isLoading } = useQuery(
     generateQueryKey(RequestKey.UserStreak, user),
     getReadingStreak,
-    { enabled: shouldShowStreak, staleTime: StaleTime.Default },
+    { enabled: true, staleTime: StaleTime.Default },
```
sshanzel, line 19: We should maybe check if the user opted out of weekly goal. If I remember correctly, we will use it for users to opt out. Probably better to check again with the product team.

packages/webapp/pages/[userId]/index.tsx
```diff
@@ -71,14 +59,6 @@ const ProfilePage = ({
       <Readme user={user} />
       {readingHistory?.userReadingRankHistory && (
         <>
```
Comment (sshanzel) [inline packages/webapp/pages/[userId]/index.tsx:61]: We should not need the fragments here anymore.

sshanzel (commented):
Looks great overall. So happy we finally get to remove this. Lots of legacy code that gave us some trouble back then. Just had a few comments but looks great overall.

sshanzel:
A few more files we can probably delete:
- Most of the files inside the `RanksModal` folder and the modal itself.
- `rank.ts` file.
- `useIncrementReadingRank.ts`.
- `useReadingRank.ts`.

sshanzel (changes requested):
Just the files listed to be removed, other than that, everything looks great.

packages/shared/src/components/modals/BootPopups.tsx
nensidosari: Yes true true, this can be removed

nensidosari: Hmm, based on Lee's comment looks like we might want to keep it. Will check with product about this https://github.com/dailydotdev/apps/pull/3091#discussion_r1599502582

packages/shared/src/hooks/streaks/useReadingStreak.ts
nensidosari: Will check with product about this, since here at least we were not taking it in consideration anyways

packages/shared/src/lib/rank.ts
```diff
@@ -1,22 +1,3 @@
-export const NO_RANK = 'No rank';
```
sshanzel, line 1: We should completely delete this file once the relevant hooks are deleted as well.

packages/shared/src/hooks/streaks/useReadingStreak.ts
nensidosari: product thread here: https://dailydotdev.slack.com/archives/C05TQL6HWTH/p1715694222530479

Anyways, I don't think this should be blocking since even if we want a similar toggle for reading streaks, keeping the same one with the weekly goal naming could bring complications. I suggest we remove and then product can decide if they want to introduce something like this as a follow up.

packages/shared/src/lib/rank.ts
nensidosari: Yup yup working on it :D

packages/shared/src/hooks/streaks/useReadingStreak.ts
nensidosari: Also yes we seem to use this for the moment, maybe the best thing to do is to actually work on renaming this into something reading streaks related

packages/shared/src/hooks/streaks/useReadingStreak.ts
nensidosari: I checked an not all existing users have seen this

nensidosari:
@sshanzel @kopancek have adjusted all feedback given, please check again 🙏

packages/shared/src/hooks/streaks/useReadingStreak.ts
sshanzel: Yep, sure thing. Thank you for checking.

sshanzel (commented):
Last few things! We are almost there. This is why I was really looking forward to seeing this cleanup, it involves a lot of legacy things 😂 

In the `/graphql/users.ts` file, there are a lot of queries and interfaces that were used for the reading rank. We can remove those as well as it can save us some bandwidth and reduce the calculation needed for some queries such as on tooltip, etc.

In our `base.css` file, there are also a lot of reading rank CSS variables we can cleanup.

packages/shared/src/lib/rank.ts
nensidosari: Very soon 😅  I still forgot to remove some other stuff u mentioned below

packages/webapp/pages/[userId]/index.tsx
Comment (sshanzel) [reply packages/webapp/pages/[userId]/index.tsx:61]: Right. I was under the impression that we would not need the `readingHistory?.userReadingRankHistory` anymore. Though after checking, it seems to be used to check whether there is data already to decide whether to render the components.

While the statement about conditionally rendering is true, we might need to use a different variable since the `readingHistory?.userReadingRankHistory` should not exist anymore after the cleanup of the queries.

Comment (nensidosari) [reply packages/webapp/pages/[userId]/index.tsx:61]: Yup, will spend a bit more on this to possibly get rid of everything that mentions rank

nensidosari:
Just thoughts, since it's going to be more removals and adjustments, do we want to do this in 2 steps ? Merge this and then I can continue with the rest of the cleanup for everything rank related?

sshanzel:
> Just thoughts, since it's going to be more removals and adjustments, do we want to do this in 2 steps ? Merge this and then I can continue with the rest of the cleanup for everything rank related?
This to not make the PR bigger then it is, and review easier as well

Yeah, sounds good to me. The removal of the queries might involve more careful testing to avoid any breakage, so I think doing it in a separate PR is a great idea. We've already removed a lot of lines here which is nice. Feel free to merge once you are okay with the changes.

sshanzel (approved):
I have to say, this is my favorite cleanup of all time 🥳 :shipit:

packages/shared/src/hooks/streaks/useReadingStreak.ts
kopancek: enabled property is true by default, so I think we can just omit it now 
```suggestion
    { staleTime: StaleTime.Default },
```

packages/shared/src/hooks/streaks/useReadingStreak.ts
```diff
@@ -1,27 +1,49 @@
...
   );
+  const { checkHasCompleted } = useActions();
+  const hasReadToday =
+    new Date(streak?.lastViewAt).getDate() === new Date().getDate();
```
kopancek, line 26: I think for this comparison, we are using date-fns is same-day function. But maybe only on the backend side? 🤔

packages/shared/src/hooks/streaks/useReadingStreak.ts
```diff
@@ -1,27 +1,49 @@
...
+  const userStreakQueryKeyRef = useRef<unknown[]>();
+  const queryClient = useQueryClient();
+
+  const [clearQueries] = useDebounce(async () => {
```
kopancek, line 30: Why do we debounce here?

packages/shared/src/hooks/streaks/useReadingStreak.ts
```diff
@@ -1,27 +1,49 @@
...
+    if (!hasReadToday && userStreakQueryKeyRef.current?.length > 0) {
+      await queryClient.invalidateQueries(userStreakQueryKeyRef.current);
+    }
+  }, 100);
 
   return {
     streak,
     isLoading,
-    isEnabled: shouldShowStreak,
+    checkReadingStreak: async () => {
```
kopancek, line 39: Why was this hook changed so much? Seems like new functionality to me vs what we had :)

kopancek (approved):
Mostly non blocking comments, looks good.

packages/shared/src/hooks/streaks/useReadingStreak.ts
nensidosari: I found this exact comparison used somewhere else for streaks, that's why took the same implementation to not have 2 different evaluations.

nensidosari: https://github.com/dailydotdev/apps/blob/21de818eb6e2b94bf673dbf14b86bcd7c94cf3a4/packages/shared/src/components/streak/ReadingStreakButton.tsx#L71-#L72

packages/shared/src/hooks/streaks/useReadingStreak.ts
nensidosari: Kept this as it was in useIncrementReadingRank since it was invalidating user streaks query as well, and totally removed useIncrementReadingRank https://github.com/dailydotdev/apps/pull/3091/files#diff-2ca4b5627b297d8adac0ec5fa73626249c395ceb53225a067d128ed3fad2a6d5

packages/shared/src/hooks/streaks/useReadingStreak.ts
nensidosari: Same answer as the above https://github.com/dailydotdev/apps/pull/3091#discussion_r1601369413

One more functionality was added to it that was in useIncrementReadingRank which was deleted

PR #3093, opened 2024-05-13
feat: post creation with new squad

sshanzel:
## Changes
- Image component fallback for Squad type images.
- Refactor SquadsDropdown to make the list come from props.
- New hook to centralize creation of Squad functions.
- Create Squad when the selected item is "Create new squad".

Screenshot 2024-05-13 at 1 37 34 PM

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

MI-319 #done

### Preview domain
https://mi-319.preview.app.daily.dev


packages/shared/src/components/post/write/CreatePostButton.tsx
```diff
@@ -36,11 +36,11 @@ export function CreatePostButton({
   const handle = route === '/squads/[handle]' ? (query.handle as string) : '';
   const { squad } = useSquad({ handle });
   const allowedToPost = verifyPermission(squad, SourcePermissions.Post);
-  const hasAccess = squads?.some((item) =>
-    verifyPermission(item, SourcePermissions.Post),
-  );
+  const hasAccess = handle
+    ? squads?.some((item) => verifyPermission(item, SourcePermissions.Post))
+    : true;
```
sshanzel, line 41: When there is a handle, it indicates the user is at the actual post-writing page, which we need validation. But if it is `undefined` it means we are at any other page in which the header "Post" button must be enabled.

packages/shared/src/hooks/squads/useSquadCreate.ts
```diff
@@ -0,0 +1,62 @@
...
+    unknown,
+    Omit<SquadForm, 'commentary'>
+  >;
+}
+
+type CustomHook = (props?: UseSquadCreateProps) => UseSquadCreate;
+
+const DEFAULT_ERROR = "Oops! That didn't seem to work. Let's try again!";
+
+export const useSquadCreate: CustomHook = ({ onSuccess } = {}) => {
```
sshanzel, line 32: Basically copy-pasted from Squads creation page.

packages/shared/src/lib/strings.ts
```diff
@@ -13,3 +13,11 @@ export const checkLowercaseEquality = (
   value1: string,
   value2: string,
 ): boolean => value1?.toLowerCase() === value2?.toLowerCase();
+
+export const sortAlphabetically = (a: string, b: string): number => {
```
sshanzel, line 17: Found out that the `localCompare` is quite worse when used for sorting.

capJavert: Nice, today I learned 👍

capJavert (approved):
Looks good code wise 👌

packages/shared/src/components/post/write/SquadsDropdown.tsx
```diff
@@ -2,30 +2,42 @@ import React, { ReactElement } from 'react';
...
 
 interface SquadsDropdownProps {
   onSelect: (index: number) => void;
   selected: number;
+  list: Squad[];
 }
 
+export const generateDefaultSquad = (username: string): Squad => ({
+  id: username,
+  handle: username,
```
rebelchris, line 18: More wondering what happens frontend wise when you accidentally create a duplicate id/handle squad?
You get the handle toast message, but then what?

packages/webapp/pages/squads/create.tsx
```diff
@@ -40,9 +55,28 @@ function CreatePost(): ReactElement {
...
+
+    if (!user) {
+      return filtered;
+    }
+
+    return [
+      {
+        ...generateDefaultSquad(user.username),
+        name: 'Create new Squad',
+        handle: null,
```
rebelchris, line 75: Why do we do this here?

packages/webapp/pages/squads/create.tsx
```diff
@@ -75,6 +109,13 @@ function CreatePost(): ReactElement {
       onAskConfirmation(true);
     },
   });
+  const { onCreateSquad, isLoading } = useSquadCreate({
+    onSuccess: (newSquad) => {
+      const form = formToJson<CreatePostProps>(formRef.current);
+
+      return onCreatePost({ ...form, sourceId: newSquad.id });
```
rebelchris, line 116: Another follow up, so let's say handle was failed, this will never fire.
What happens if user updates handle re-submits, this will just retrigger?

packages/shared/src/lib/strings.ts
kopancek: Also, this is not doing the sorting, just comparing, so not a good name.

And MDN itself suggests to use built in [`compare` for this same purpose](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Collator/compare).

packages/shared/src/hooks/squads/useSquadCreate.ts
```diff
@@ -0,0 +1,62 @@
...
+  const { addSquad } = useBoot();
```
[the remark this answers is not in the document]
kopancek: Nevermind, the PR with those changes has been reverted (twice now 😭 )

packages/shared/src/components/post/write/SquadsDropdown.tsx
sshanzel: If I remember correctly, from the RFC, it was decided to not cover the edge case as it is a bit unique. However, I will verify and get back to this.

packages/shared/src/lib/strings.ts
sshanzel: I doubt there would be extreme difference in terms of performance, but if we can do it without spending a lot of effort, I guess why not?

Reference: https://stackoverflow.com/a/52369951/5532217

The difference here between localCompare and non-locale is quite significant. Shouldn't hurt to be performant on little time spent.
Screenshot 2024-05-14 at 10 35 26 AM

packages/webapp/pages/squads/create.tsx
sshanzel: Just so the dropdown would not have any handle displayed, as per Figma.

packages/webapp/pages/squads/create.tsx
sshanzel: Yes.

packages/shared/src/components/post/write/CreatePostButton.tsx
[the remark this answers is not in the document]
sshanzel: Yeah, we can definitely do that.

packages/shared/src/components/post/write/CreatePostButton.tsx
```diff
@@ -36,11 +36,11 @@ export function CreatePostButton({
...
 
-  if (!footer && (!user || !squads?.length)) {
+  if (!footer && !user) {
```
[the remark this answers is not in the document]
sshanzel: Yes.

packages/shared/src/components/post/write/SquadsDropdown.tsx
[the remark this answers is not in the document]
sshanzel: Rechecked the RFC, we should add some numbers when the handle is already taken. User's handle is also sanitized and checked on user creation/edit. At the same time, we check the validity of the squad's handle on the API side. So we should be safe from unwanted characters.

packages/webapp/pages/squads/create.tsx
```diff
@@ -40,9 +55,28 @@ function CreatePost(): ReactElement {
...
   const [selected, setSelected] = useState(-1);
-  const activeSquads = squads?.filter(
-    (squad) => squad?.active && verifyPermission(squad, SourcePermissions.Post),
-  );
+  const activeSquads = useMemo(() => {
+    const filtered = squads
+      ?.filter(
+        (squaitem) =>
+          squaitem?.active &&
+          verifyPermission(squaitem, SourcePermissions.Post),
```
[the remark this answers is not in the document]
sshanzel: Thanks!

packages/shared/src/components/post/write/SquadsDropdown.tsx
```diff
@@ -2,30 +2,42 @@ import React, { ReactElement } from 'react';
...
+  image: cloudinary.squads.imageFallback,
+  permalink: null,
+  active: true,
+  public: false,
+  type: SourceType.Squad,
+  membersCount: 1,
+  description: null,
+  memberPostingRole: SourceMemberRole.Admin,
+  memberInviteRole: SourceMemberRole.Admin,
+});
```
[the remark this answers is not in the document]
sshanzel: Yeah, we can di this as well.

packages/shared/src/lib/strings.ts
kopancek: In my experience, it's usually better to use standard way. Since more people will be familiar with it. But if you feel strongly about this, I'll not stand in the way, consider my comments in this thread non-blocking :)

kopancek: But can we at least try to rename the function to `compareAlphabetically` or `fastCompare` or similar? 🙏

kopancek (approved):
Approving to unblock

packages/shared/src/lib/strings.ts
sshanzel: Updated here: https://github.com/dailydotdev/apps/pull/3093/commits/1a182beb51e8cc92bcf158ccaa7c818ae44ed49f - think it is much cleaner this way 🚀

PR #3103, opened 2024-05-15
fix: call search version on search moment

rebelchris:
- Need to verify on preview deployment 

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done

### Preview domain
https://fix-search-version.preview.app.daily.dev


packages/shared/src/components/MainFeedLayout.tsx
```diff
@@ -218,6 +215,7 @@ export default function MainFeedLayout({
     }
 
     if (isSearchOn && searchQuery) {
+      const searchVersion = getFeatureValue(feature.searchVersion);
```
nensidosari, line 218: Just curiosity why are only getting the version here? or we only need it here?

rebelchris: Yeah only needed here,
Was a edge case where the other hook would still be loading and return the default value and then never retrigger.

nensidosari (approved):

idoshamun (commented):
Can we remove the other evaluations of this feature outside this query?

rebelchris:
@idoshamun not sure what you mean, the other ones are different queries (there's only useConditionalFeature left in this regard)
But it's the suggestions and haven't seen that one fail getting the right version.

Not sure if that's what you meant.

idoshamun (approved):
I thought there was an additional one. But if not we're good to merge!

PR #3105, opened 2024-05-15
chore: remove `post-scraper-one-ai` from Tiltfile

omBratteng:
Deprecated, so not needed locally

### Preview domain
https://ombratteng-patch-1.preview.app.daily.dev


capJavert (changes requested):
This will break sharing links to squad locally.

omBratteng:
💩

omBratteng:
Guess we'll need to keep it then. So post-scraper-one-ai is also still used in production

PR #3107, opened 2024-05-15
fix: squad graphql query

kopancek:
## Changes

[More details on slack](https://dailydotdev.slack.com/archives/C06CUKX1P3N/p1715760110448429)

TLDR: our graphorm implementation cannot deal with children having timestamps without timezone. The `createdAt` field on `UserShortInfo` fragment was added to serve devcard requests. Easy fix was to remove the `createdAt` from the fragment and add the field directly to the devcard request. This is the easy fix.

Root cause fix would be to fix the graphorm implementation, but that's out of scope of public squad work.

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

### Preview domain
https://fix-squad-graphql-query.preview.app.daily.dev

### DIFF
--- a/packages/shared/src/graphql/fragments.ts
+++ b/packages/shared/src/graphql/fragments.ts
@@ -23,7 +23,6 @@ export const USER_SHORT_INFO_FRAGMENT = gql`
     permalink
     username
     bio
-    createdAt
     reputation
   }
 `;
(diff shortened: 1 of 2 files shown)


sshanzel (approved):
Thank you for spending the time on investigating this issue 🙏

PR #3110, opened 2024-05-16
feat: hook up mutation with submit public squad for review modal

kopancek:
## Changes

- Added new mutation to the frontend
- Added a hook that encapsulates the mutation
- Using the hook in the submit squad for review modal
- Rudimentary error handling

Related issue:
- https://dailydotdev.atlassian.net/browse/MI-345

### Screenshots

Screenshot 2024-05-16 at 14 54 34
Screenshot 2024-05-16 at 14 54 43

## Events

No new events, do we want some? 🤔 

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [X] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

MI-345 #done


packages/shared/src/graphql/sources.ts
```diff
@@ -142,3 +142,18 @@ export const SIMILAR_SOURCES_QUERY = gql`
     }
   }
 `;
+
+export enum PublicSquadRequestStatus {
+  Pending = 'pending',
```
nensidosari, line 147: Didn't we type these somewhere else? maybe not, I remember seeing something like this, but not sure so just checking

nensidosari (approved):
Think this looks good

packages/shared/src/graphql/sources.ts
kopancek: I'm not sure anymore. I think we only did it on the API side :)

PR #3112, opened 2024-05-17
feat: public squads

sshanzel:
Intermediate branch for public Squads (collection of approved PRs).

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done

### Preview domain
https://mi-217-public-squads.preview.app.daily.dev


packages/shared/src/components/squads/SquadPageHeader.tsx
```diff
@@ -72,14 +114,44 @@ export function SquadPageHeader({
           >
             {squad.name}
           </h1>
-          <h2
-            className={classNames(
-              'mt-1 text-center text-text-tertiary typo-body tablet:mt-2',
-              !shouldUseListModeV1 && 'laptopL:text-left',
+          <div className="mt-1 flex flex-row items-center justify-center text-text-quaternary tablet:mt-2 laptopL:justify-start">
```
nensidosari, line 117: We should add a condition `!shouldUseListModeV1 && 'laptopL:justify-start'` since we want the laptop styles only if it's not list mode or it's mobile/tablet

packages/shared/src/components/squads/SquadPageHeader.tsx
```diff
@@ -72,14 +114,44 @@ export function SquadPageHeader({
...
+            <h2
+              className={classNames(
+                'text-center text-text-tertiary typo-footnote laptopL:text-left',
```
nensidosari, line 120:
```suggestion
                'text-center text-text-tertiary typo-footnote',
```
this is added below with the list layout condition

packages/shared/src/components/squads/SquadPageHeader.tsx
```diff
@@ -93,18 +165,25 @@ export function SquadPageHeader({
...
+        )}
+      </div>
       <SquadHeaderBar
         squad={squad}
         members={members}
-        className={classNames(
-          !shouldUseListModeV1 &&
-            'laptopL:absolute laptopL:right-18 laptopL:top-0',
-        )}
+        className="mt-8 laptopL:absolute laptopL:right-18 laptopL:top-0 laptopL:mt-0"
```
nensidosari, line 186: All laptopL styles should be behind `!shouldUseListModeV1` flag

sshanzel:
@nensidosari  should look like this now 🥳 

Screenshot 2024-05-24 at 5 29 47 PM

nensidosari (approved):
looks great. Did just an overall review not in details since it has already been approved

omBratteng (approved):

packages/shared/src/components/checklist/EditSquadStep.tsx
```diff
@@ -0,0 +1,25 @@
...
+  ...props
+}: ChecklistStepPropsWithSquad): ReactElement => {
+  return (
+    <ChecklistStep {...props}>
+      <div className="flex">
+        <Button
+          tag="a"
+          variant={ButtonVariant.Primary}
+          icon={<EditIcon />}
+          href={`/squads/${squad.handle}/edit`}
```
rebelchris, line 18: Is there anyway this could show in extension?
Let's maybe just put `webappUrl` in front to be safe?

packages/shared/src/components/checklist/SquadChecklistCard.tsx
```diff
@@ -60,7 +51,14 @@ const SquadChecklistCard = ({ squad }: { squad: Squad }): ReactElement => {
...
       onClose={onRequestClose}
     >
-      {checklistElement}
+      <ChecklistCard
+        title={
+          isDone ? 'Good job! you nailed it. 🥳' : 'Get started with squads'
+        }
+        description={`${totalStepsCount} simple steps to Squad greatness!`}
+        steps={steps}
+        className="max-w-full border-0"
```
rebelchris, line 60: Don't you need `!border-0` here?
It might work without, but that's just luck of tailwind ordering then?

packages/shared/src/components/modals/squads/SquadNotificationsModal.tsx
```diff
@@ -143,12 +144,9 @@ export function SquadNotificationsModal({
         )}
         <p className="text-text-tertiary typo-callout">
           For more options, go to your
-          <a
-            className="ml-1 text-text-link underline hover:no-underline"
-            href="/account/notifications"
-          >
+          <Anchor className="ml-1" href="/account/notifications">
```
rebelchris, line 147: Probably also safer to use full URL here?

packages/shared/src/components/modals/squads/SubmitSquadForReviewModal.tsx
```diff
@@ -79,14 +87,13 @@ const SubmitSquadForReviewModal = ({
         <Checklist>
           <ChecklistItem iconTag={MegaphoneIcon}>
             The posts I shared meet the{' '}
-            <a
-              className="text-text-link underline hover:no-underline"
+            <Anchor
               href="https://docs.daily.dev/docs/for-content-creators/content-guidelines"
```
rebelchris, line 91: I know this is old, but don't we have a rebrandly here?
Can we add it to backlog?

packages/shared/src/components/squads/Members/PrivilegedMemberItem.tsx
```diff
@@ -0,0 +1,31 @@
...
+}: PrivilegedMemberItemProps): ReactElement {
+  return (
+    <ProfileTooltip user={user} tooltip={{ placement: 'bottom' }}>
+      <ProfileLink
+        href={user.permalink}
+        className="flex flex-row items-center rounded-10 border border-border-subtlest-tertiary p-2"
+      >
+        <ProfilePicture user={user} size="large" />
+        <div className="flex-col">
+          <span className="ml-2.5 flex overflow-hidden text-ellipsis whitespace-nowrap text-text-tertiary typo-subhead">
```
rebelchris, line 23: Just a check, @omBratteng didn't you add a shorthand className for this? *(the truncating)

packages/shared/src/components/squads/PublicSquadEligibilityCard.tsx
```diff
@@ -1,59 +1,65 @@
...
+  }
 
+  const isEligible = postsCount >= MIN_SQUAD_POSTS;
   const CardComponent = shouldUseListFeedLayout ? CardV1 : Card;
 
   return (
     <CardComponent
       data-testid="publicSquadEligibilityCard"
-      className="justify-between !bg-transparent p-4"
+      className="min-h-[21.75rem] justify-between !bg-transparent p-4"
```
rebelchris, line 62: Is this a different min-h from the main card?
won't it force others differently?

packages/shared/src/components/squads/SquadHeaderBar.tsx
```diff
@@ -51,48 +48,44 @@ export function SquadHeaderBar({
...
+        <Button
+          variant={ButtonVariant.Secondary}
+          className={classNames(
+            firstItemClasses,
+            tourIndex === TourScreenIndex.CopyInvitation && 'highlight-pulse',
           )}
-      </div>
+          onClick={() => {
+            trackAndCopyLink();
+          }}
```
rebelchris, line 70:
```suggestion
          onClick={trackAndCopyLink}
```

packages/shared/src/components/squads/SquadPageHeader.tsx
```diff
@@ -28,19 +44,55 @@ interface SquadPageHeaderProps {
...
+const SquadStat = ({ count, label }: SquadStatProps) => (
+  <span className="flex flex-row text-text-tertiary typo-footnote">
+    <strong className="mr-1 text-text-primary typo-subhead">
+      {largeNumberFormat(count)}
+    </strong>
+    {label}
+  </span>
+);
+
 export function SquadPageHeader({
```
rebelchris, line 61: This file is massive, think we could have split it into some components.

packages/shared/src/components/squads/SquadPageHeader.tsx
```diff
@@ -93,19 +187,36 @@ export function SquadPageHeader({
...
-      <SquadMemberShortList
-        squad={squad}
-        members={members}
-        className={classNames('my-6', !shouldUseListModeV1 && 'laptopL:hidden')}
-      />
+      <span className="mt-6 text-text-quaternary typo-footnote">
+        Moderated by
+      </span>
+      <div className="mt-2 flex flex-row items-center gap-3">
+        {squad.privilegedMembers?.slice(0, listMax).map((member) => (
```
rebelchris, line 194: Why does this happen inline, can we abstract it outside the rendering?

packages/shared/src/components/squads/SquadPageHeader.tsx
```diff
@@ -134,7 +245,7 @@ export function SquadPageHeader({
               </FlexCentered>
               <Button
                 tag="a"
-                href={`${link.post.create}?sid=${squad?.handle}`}
+                href={`${link.post.create}?sid=${squad.handle}`}
```
rebelchris, line 248: Guess this is always webapp right?

rebelchris (approved):
Did a quick look, some minor comments nothing blocking I don't think.

packages/shared/src/components/squads/Members/PrivilegedMemberItem.tsx
nensidosari: oh yes, I think I remember that as well

packages/shared/src/components/checklist/EditSquadStep.tsx
sshanzel: This is never shown from the extension, but yeah we can add that.

packages/shared/src/components/squads/SquadPageHeader.tsx
sshanzel: Yes.

packages/shared/src/components/squads/SquadHeaderBar.tsx
sshanzel: This function accepts an optional parameter that doesn't match native `onClick`.

packages/shared/src/components/squads/SquadPageHeader.tsx
sshanzel: Not sure why not. Seems like a pretty common thing on the render function (similar to rendering infinite query results).

packages/shared/src/components/squads/SquadPageHeader.tsx
sshanzel: Let's maybe create a ticket to revisit this. We've done quite some testing already and don't want to change too much with the components as we'll be forced to run through it again.

packages/shared/src/components/squads/SquadHeaderBar.tsx
rebelchris: Ah ok no problem

packages/shared/src/components/squads/SquadPageHeader.tsx
rebelchris: Guess it was mostly annoying because this file is already so big and thought it would be nicer outside rendering, but really not urgent/needed

sshanzel: Ah, I see. Yeah, maybe once we break it down, we might be able to find some way to make more bearable.

packages/shared/src/components/modals/squads/SquadNotificationsModal.tsx
sshanzel: Yeah, let's update this one too.

packages/shared/src/components/checklist/SquadChecklistCard.tsx
sshanzel: Let me check whether we have to override to be sure.

packages/shared/src/components/squads/PublicSquadEligibilityCard.tsx
sshanzel: The min-h for the card is normally much bigger. Should we probably just use the same?

packages/shared/src/components/squads/Members/PrivilegedMemberItem.tsx
sshanzel: Ah yes, there is also a native truncate class iirc.

packages/shared/src/components/modals/squads/SubmitSquadForReviewModal.tsx
sshanzel: Found one at `constants` file. Let me update this now.

packages/shared/src/components/squads/PublicSquadEligibilityCard.tsx
nensidosari: well in idea it should be just like the other cards on the feed no need to be different

PR #3113, opened 2024-05-17
Mi 146 list view - retry

nensidosari:
The initial PR was reverted because of an [issue](https://dailydotdev.slack.com/archives/C01L0QXQJNB/p1715866920025429) in tags and sources page - see commits [here](https://github.com/dailydotdev/apps/pull/3113/files/532ea760ac81703939dabaf727eb4dd4329962e6..2823f6ba63b60bb6800e17662991ec67bfd6fdfd) for the last changes to fix the issue. 

Issue was related to isLaptop not being defined in SSR and resulting to false statement that the env was mobile when it was not. Resolved by not considering isLaptop view evaluation if it is null or undefined

Decision to move on with this PR [here](https://dailydotdev.slack.com/archives/C01L0QXQJNB/p1716278512251279)

Preview domain
https://mi-146-list-view.preview.app.daily.dev/

### Preview domain
https://mi-146-list-view.preview.app.daily.dev


packages/shared/src/hooks/useFeedLayout.ts
```diff
@@ -84,22 +96,51 @@ export const useFeedLayout = ({
...
+  const shouldUseListFeedLayoutOnProfilePages = UserProfileFeedPages.has(
+    feedName as UserProfileFeedType,
+  );
+
+  const isFeedIncludedInListLayout = FeedLayoutMobileFeedPages.has(
+    feedName as FeedPagesWithMobileLayoutType,
+  );
+
+  const shouldUseListFeedLayoutOnMobileTablet =
+    !isNullOrUndefined(isLaptop) && !isLaptop && isFeedIncludedInListLayout;
```
rebelchris, line 117: Weird feels as the `!isLaptop` would already be enough here.

rebelchris (approved):
Nice find, couldn't reproduce.
Happy we found the issue here 😅

packages/shared/src/hooks/useFeedLayout.ts
nensidosari: Not enough because it's taking the contrary value of what it should take.

PR #3118, opened 2024-05-20
fix: public squads integration

sshanzel:
## Changes
- Utilize queries and mutations related to requesting squads to go public.
- Introduced a new type of Feed item to render the eligibility card.
- Refactored the feed item type from strings to enums.
- Make the edit settings page reflect the actual request status.
- Display eligibility card based on status.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

MI-336 #done
MI-344 #done

### Preview domain
https://mi-336.preview.app.daily.dev

### DIFF
--- a/packages/shared/src/components/Feed.tsx
+++ b/packages/shared/src/components/Feed.tsx
@@ -53,7 +53,10 @@ import { acquisitionKey } from './cards/AcquisitionFormCard';
 import { MarketingCtaVariant } from './cards/MarketingCta/common';
 
 export interface FeedProps<T>
-  extends Pick<UseFeedOptionalParams<T>, 'options'>,
+  extends Pick<
+      UseFeedOptionalParams<T>,
+      'options' | 'showPublicSquadsEligibility'
+    >,
     Pick<FeedContainerProps, 'shortcuts'> {
   feedName: AllFeedPages;
   feedQueryKey: unknown[];
@@ -131,6 +134,7 @@ export default function Feed<T>({
   pageSize,
   isHorizontal = false,
   feedContainerRef,
+  showPublicSquadsEligibility,
 }: FeedProps<T>): ReactElement {
   const origin = Origin.Feed;
   const { trackEvent } = useContext(AnalyticsContext);
@@ -176,6 +180,7 @@ export default function Feed<T>({
       query,
       variables,
       options,
+      showPublicSquadsEligibility,
       settings: {
         disableAds,
         adPostLength: isSquadFeed ? 2 : undefined,
--- a/packages/shared/src/components/FeedItemComponent.tsx
+++ b/packages/shared/src/components/FeedItemComponent.tsx
@@ -180,7 +180,7 @@ export default function FeedItemComponent({
   );
 
   switch (item.type) {
-    case FeedItemType.RegularPost: {
+    case FeedItemType.Post: {
       if (
         !!item.post.pinnedAt &&
         item.post.source?.currentMember?.flags?.collapsePinnedPosts
@@ -251,7 +251,7 @@ export default function FeedItemComponent({
         </PostTag>
       );
     }
-    case FeedItemType.RegularAd:
+    case FeedItemType.Ad:
       return (
         <AdTag
           ref={inViewRef}
--- a/packages/shared/src/components/FeedItemComponent.tsx
+++ b/packages/shared/src/components/FeedItemComponent.tsx
@@ -20,12 +20,14 @@ import { SharePostCard as SharePostCardV1 } from './cards/v1/SharePostCard';
 import { WelcomePostCard } from './cards/WelcomePostCard';
 import { WelcomePostCard as WelcomePostCardV1 } from './cards/v1/WelcomePostCard';
 import { Origin } from '../lib/analytics';
-import { UseVotePost, useFeedLayout } from '../hooks';
+import { useFeedLayout, UseVotePost } from '../hooks';
 import { CollectionCard } from './cards/CollectionCard';
 import { CollectionCard as CollectionCardV1 } from './cards/v1/CollectionCard';
 import { AcquisitionFormCard } from './cards/AcquisitionFormCard';
 import { MarketingCtaCard, MarketingCtaList } from './cards';
 import { MarketingCtaCardV1 } from './cards/v1/MarketingCtaCard';
+import { FeedItemType } from './cards/common';
+import { PublicSquadEligibilityCard } from './squads/PublicSquadEligibilityCard';
 
 const CommentPopup = dynamic(
   () => import(/* webpackChunkName: "commentPopup" */ './cards/CommentPopup'),
@@ -178,7 +180,7 @@ export default function FeedItemComponent({
   );
 
   switch (item.type) {
-    case 'post': {
+    case FeedItemType.RegularPost: {
       if (
         !!item.post.pinnedAt &&
         item.post.source?.currentMember?.flags?.collapsePinnedPosts
@@ -249,7 +251,7 @@ export default function FeedItemComponent({
         </PostTag>
       );
     }
-    case 'ad':
+    case FeedItemType.RegularAd:
       return (
         <AdTag
           ref={inViewRef}
@@ -258,17 +260,21 @@ export default function FeedItemComponent({
           showImage={!insaneMode}
         />
       );
-    case 'userAcquisition': {
+    case FeedItemType.UserAcquisition:
       return <AcquisitionFormCard key="user-acquisition-card" />;
-    }
-    case 'marketingCta': {
+    case FeedItemType.PublicSquadEligibility:
+      return (
+        <PublicSquadEligibilityCard
+          placeholder={<PlaceholderTag showImage={!insaneMode} />}
+        />
+      );
+    case FeedItemType.MarketingCta:
       return (
         <MarketingCtaTag
           key="marketing-cta-card"
           marketingCta={item.marketingCta}
         />
       );
-    }
     default:
       return <PlaceholderTag showImage={!insaneMode} />;
   }
--- a/packages/shared/src/components/FeedItemComponent.tsx
+++ b/packages/shared/src/components/FeedItemComponent.tsx
@@ -263,11 +263,7 @@ export default function FeedItemComponent({
     case FeedItemType.UserAcquisition:
       return <AcquisitionFormCard key="user-acquisition-card" />;
     case FeedItemType.PublicSquadEligibility:
-      return (
-        <PublicSquadEligibilityCard
-          placeholder={<PlaceholderTag showImage={!insaneMode} />}
-        />
-      );
+      return <PublicSquadEligibilityCard />;
     case FeedItemType.MarketingCta:
       return (
         <MarketingCtaTag
--- a/packages/shared/src/components/cards/common.tsx
+++ b/packages/shared/src/components/cards/common.tsx
@@ -71,8 +71,8 @@ export const generateTitleClamp = ({
 };
 
 export enum FeedItemType {
-  RegularPost = 'post',
-  RegularAd = 'ad',
+  'Post' = 'post',
+  'Ad' = 'ad',
   Placeholder = 'placeholder',
   UserAcquisition = 'userAcquisition',
   PublicSquadEligibility = 'publicSquadEligibility',
--- a/packages/shared/src/components/cards/common.tsx
+++ b/packages/shared/src/components/cards/common.tsx
@@ -69,3 +69,12 @@ export const generateTitleClamp = ({
 
   return hasHtmlContent ? 'line-clamp-4' : 'line-clamp-9';
 };
+
+export enum FeedItemType {
+  RegularPost = 'post',
+  RegularAd = 'ad',
+  Placeholder = 'placeholder',
+  UserAcquisition = 'userAcquisition',
+  PublicSquadEligibility = 'publicSquadEligibility',
+  MarketingCta = 'marketingCta',
+}
--- a/packages/shared/src/components/squads/Details.tsx
+++ b/packages/shared/src/components/squads/Details.tsx
@@ -1,7 +1,7 @@
 import React, { FormEvent, ReactElement, ReactNode, useState } from 'react';
 import classNames from 'classnames';
 import { ClientError } from 'graphql-request';
-import { useMutation, useQuery } from '@tanstack/react-query';
+import { useMutation } from '@tanstack/react-query';
 import { useRouter } from 'next/router';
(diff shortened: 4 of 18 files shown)


packages/shared/src/components/squads/PublicSquadEligibilityCard.tsx
```diff
@@ -10,42 +16,46 @@ import { useLazyModal } from '../../hooks/useLazyModal';
...
   }, [completeAction]);
 
   const onSubmit = useCallback(() => {
-    openModal({ type: LazyModal.SubmitSquadForReview, props: { squadId } });
-  }, [openModal, squadId]);
+    if (!squad.id) {
+      return;
+    }
 
-  if (hideCard) {
```
sshanzel, line 47: We can't return `null` from the card as it would affect the actual placement vs the virtual placement of the card (posts after this would be pushed 1 item backwards).

packages/shared/src/components/squads/SquadPostsProgressBar.tsx
```diff
@@ -42,7 +42,7 @@ export const SquadPostsProgressBar = ({
           Posts
         </span>
         <span className={classNames(className?.labelContent ?? 'font-bold')}>
-          {effectivePostsCount} / {goal}
```
sshanzel, line 45: From Figma, there should be no spaces in between. However, what do you think looks better?

Screenshot 2024-05-19 at 9 09 19 PM
Screenshot 2024-05-19 at 8 05 45 PM

packages/shared/src/components/cards/common.tsx
```diff
@@ -69,3 +69,12 @@ export const generateTitleClamp = ({
 
   return hasHtmlContent ? 'line-clamp-4' : 'line-clamp-9';
 };
+
+export enum FeedItemType {
+  RegularPost = 'post',
+  RegularAd = 'ad',
```
nensidosari, line 75: do we need Regular on the naming? I think it's not necessary since other items have other names

packages/shared/src/components/squads/Details.tsx
```diff
@@ -70,32 +68,19 @@ export function SquadDetails({
...
-      return SquadStatus.InProgress;
-    },
-    {
-      ...disabledRefetch,
-      cacheTime: Infinity,
-      enabled: !isNullOrUndefined(form?.flags?.totalPosts) && !createMode,
-    },
-  );
+  const isRequestsEnabled =
+    !createMode && !form.public && !form.flags?.totalPosts && !!form?.id;
```
nensidosari, line 79: we were using `!isNullOrUndefined` above, don't we need it here?

nensidosari: Since u're integrating also MI-344 should the above be here?

`TODO: MI-344 add another 2 if statement here to check if the squad is rejected/pending`

packages/shared/src/components/squads/PublicSquadEligibilityCard.tsx
nensidosari: Have you checked what we do with other cards like user acquisition?

packages/shared/src/components/squads/SquadPostsProgressBar.tsx
nensidosari: I think I'm fine with the no space

packages/shared/src/hooks/squads/usePublicSquadRequests.ts
```diff
@@ -1,22 +1,61 @@
...
 
-export const usePublicSquadRequests = (): UsePublicSquadRequestsResult => {
+interface UsePublicSquadRequestsProps {
+  isQueryEnabled?: boolean;
+  sourceId: string;
+  status?: string;
+}
+
+const remoteStatusMap: Record<PublicSquadRequestStatus, SquadStatus> = {
+  [PublicSquadRequestStatus.Approved]: SquadStatus.Approved,
```
nensidosari, line 40: why do we have 2 of these?

packages/shared/src/components/Feed.tsx
```diff
@@ -53,7 +53,10 @@ import { acquisitionKey } from './cards/AcquisitionFormCard';
 import { MarketingCtaVariant } from './cards/MarketingCta/common';
 
 export interface FeedProps<T>
-  extends Pick<UseFeedOptionalParams<T>, 'options'>,
+  extends Pick<
+      UseFeedOptionalParams<T>,
+      'options' | 'showPublicSquadsEligibility'
```
rebelchris, line 58: Wouldn't this be one of the options better?

packages/shared/src/components/FeedItemComponent.tsx
```diff
@@ -178,7 +180,7 @@ export default function FeedItemComponent({
   );
 
   switch (item.type) {
-    case 'post': {
+    case FeedItemType.RegularPost: {
```
rebelchris, line 183: Why the `regular` here and for `ad` do we have other types?

packages/shared/src/components/cards/common.tsx
rebelchris: Same feelings :)

packages/shared/src/components/squads/SquadPostsProgressBar.tsx
```diff
@@ -42,7 +42,7 @@ export const SquadPostsProgressBar = ({
...
+          {effectivePostsCount}/{goal}
```
rebelchris, line 45: This looks better to me :)

packages/shared/src/hooks/squads/usePublicSquadRequests.ts
```diff
@@ -34,5 +73,29 @@ export const usePublicSquadRequests = (): UsePublicSquadRequestsResult => {
   return {
```
rebelchris, line 73: Should we maybe memoize some of these?

rebelchris (changes requested):
Minor comments from my side :)

packages/shared/src/components/squads/PublicSquadEligibilityCard.tsx
sshanzel: User acquisition is also generated from the same hook, but not sure if we return null from there.

packages/shared/src/components/cards/common.tsx
sshanzel: It just interferes with the imported `Post` type, I should maybe find another file to put this enum with where the `Post` is not imported, but this is already the `common.tsx`. I thought of adding `types.ts` file but this is not even just a type.

packages/shared/src/components/squads/Details.tsx
sshanzel: Shouldn't matter which one we would use. If the value is `0`, you don't have to make any request still since that would mean you have never requested.

packages/shared/src/hooks/squads/usePublicSquadRequests.ts
sshanzel: One is remote values (API), one is from the frontend. The API doesn't have `InProgress`.

packages/shared/src/components/Feed.tsx
sshanzel: I tried that at first, but it was a type from `useQuery` itself. `options?: UseInfiniteQueryOptions<FeedData>`

Thought it would be simpler just to add it.

packages/shared/src/components/FeedItemComponent.tsx
sshanzel: Lint issues: https://github.com/dailydotdev/apps/pull/3118#discussion_r1609656253

packages/shared/src/components/squads/SquadPostsProgressBar.tsx
sshanzel: Awesome. Let's move forward without spaces.

packages/shared/src/hooks/squads/usePublicSquadRequests.ts
sshanzel: Good point. Most of the values here are stable. I think we can memoize the status so we won't have to instantiate and calculate days all the time.

packages/shared/src/components/cards/common.tsx
sshanzel: I can disable the link here since this is something that doesn't affect negatively. Thoughts?
Screenshot 2024-05-22 at 6 00 48 PM

nensidosari: Maybe if u save it as a string 'Post', no need to disable eslint

sshanzel:
Would love to hear you guys think 👁️ cc @rebelchris @nensidosari

packages/shared/src/components/FeedItemComponent.tsx
```diff
@@ -258,17 +260,21 @@ export default function FeedItemComponent({
...
-    case 'userAcquisition': {
+    case FeedItemType.UserAcquisition:
       return <AcquisitionFormCard key="user-acquisition-card" />;
-    }
-    case 'marketingCta': {
+    case FeedItemType.PublicSquadEligibility:
+      return (
+        <PublicSquadEligibilityCard
+          placeholder={<PlaceholderTag showImage={!insaneMode} />}
+        />
```
rebelchris, line 269: Just wondering why did you decide to pass the placeholder, wouldn't it be better to do it inside the component?

packages/shared/src/components/squads/PublicSquadEligibilityCard.tsx
```diff
@@ -10,42 +16,46 @@ import { useLazyModal } from '../../hooks/useLazyModal';
...
 } from './SquadPostsProgressBar';
 import { squadsPublicGuide } from '../../lib/constants';
+import { SquadStatus } from './settings';
 
-interface Props {
-  postsCount: number;
-  inReview: boolean;
-  squadId: string;
+interface PublicSquadEligibilityCardProps {
+  placeholder: ReactNode;
```
rebelchris, line 26: Not sure this should be passible?
Also now that I think about it, you tested support for new list view as well?

packages/shared/src/components/squads/PublicSquadEligibilityCard.tsx
```diff
@@ -10,42 +16,46 @@ import { useLazyModal } from '../../hooks/useLazyModal';
...
-}: Props): ReactElement => {
+export function PublicSquadEligibilityCard({
+  placeholder,
+}: PublicSquadEligibilityCardProps): ReactElement {
+  const router = useRouter();
+  const { squad } = useSquad({ handle: router.query.handle as string });
+  const { isFetched, status } = usePublicSquadRequests({
+    sourceId: squad?.id,
+  });
+  const postsCount = squad?.flags.totalPosts;
```
rebelchris, line 37: Is there a possibility where server side props have squad, but not fetched flags yet?
Maybe make it `?` as well just to be safe?

packages/shared/src/hooks/squads/usePublicSquadRequests.ts
```diff
@@ -31,8 +71,31 @@ export const usePublicSquadRequests = (): UsePublicSquadRequestsResult => {
...
+      return SquadStatus.InProgress;
+    }
+
+    if (latestRequest.status === PublicSquadRequestStatus.Rejected) {
+      const fourteenDaysAgo = subDays(new Date(), 14);
+      const requestDate = new Date(latestRequest.createdAt);
+      return fourteenDaysAgo < requestDate
+        ? SquadStatus.Rejected
+        : SquadStatus.InProgress;
+    }
```
rebelchris, line 88: Not sure what was discussed on RFC, but why do we do this logic on the frontend, feels like this should be handled on API side?
CC: @capJavert

rebelchris (changes requested):
Some minor questions, think we can almost wrap this up :)

packages/shared/src/components/FeedItemComponent.tsx
sshanzel: The placeholder depends on the feed, I didn't want to include all that logic again hence we passed it.

packages/shared/src/components/squads/PublicSquadEligibilityCard.tsx
sshanzel: Yeah, sure, we can add that.

packages/shared/src/hooks/squads/usePublicSquadRequests.ts
sshanzel: This is not going to be used for validation but something we need to display some alert.

packages/shared/src/components/FeedItemComponent.tsx
rebelchris: How does it depend on the feed?
this card is only in squad feed no?

sshanzel: There are certain conditions when fetching the placeholder to use, and iirc, we have plans to move the squad feed to use the new layout in the future. I could be wrong though.

Screenshot 2024-05-23 at 3 58 26 PM

rebelchris: Ah yeah so for those cases you should actually create 2 different cards with placeholders inside.
It's the better approach, but some system didn't take it into account already.

I see not for all new cards we do use this system.

sshanzel: Cool. Yeah, we can fix the problem when we get there. Have update the implementation.

rebelchris (approved):
Cool no blockers code wise from my side.

packages/shared/src/components/FeedItemComponent.tsx
nensidosari: squad feed already uses the new layout 😬

sshanzel: I will check this before merging.

omBratteng (approved):

packages/shared/src/components/squads/PublicSquadEligibilityCard.tsx
nensidosari: with that we actually replace the first ad with the card. I think at some point we decided to do the same here to not deal with post shift. Let me know if I'm wrong though about the decisions for it

nensidosari: we seem to be doing that as well here just saw the last file so all good

nensidosari (approved):
This looks good from my side as well! 🚀

PR #3121, opened 2024-05-20
Feat new fb event

nensidosari:
## Changes
- only call trackAnalyticsSignUp on onboarding
- pass new user fetched data to the tracking function
- create event with experience level fb  

- Requested by Ido to have tracking on fb for experience

### Preview domain
https://feat-new-fb-event.preview.app.daily.dev

### DIFF
--- a/packages/shared/src/components/auth/OnboardingAnalytics.tsx
+++ b/packages/shared/src/components/auth/OnboardingAnalytics.tsx
@@ -135,7 +135,9 @@ export const trackAnalyticsSignUp = ({
 
   if (typeof globalThis.fbq === 'function') {
     globalThis.fbq('track', 'signup');
-    globalThis.fbq('track', 'signup_experience_level', experienceLevel);
+    if (experienceLevel) {
+      globalThis.fbq('track', 'signup_experience_level', experienceLevel);
+    }
   }
 
   if (typeof globalThis.twq === 'function') {
--- a/packages/shared/src/components/auth/OnboardingAnalytics.tsx
+++ b/packages/shared/src/components/auth/OnboardingAnalytics.tsx
@@ -127,11 +127,6 @@ export const trackAnalyticsSignUp = ({ experienceLevel }): void => {
   }
 
   if (typeof globalThis.fbq === 'function') {
-    console.log(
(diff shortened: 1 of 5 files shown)


packages/shared/src/components/auth/OnboardingAnalytics.tsx
```diff
@@ -121,13 +122,20 @@ export const TiktokTracking = (): ReactElement => {
...
+export const trackAnalyticsSignUp = ({
+  experienceLevel,
+}: TrackAnalyticsSignUpProps): void => {
   if (typeof globalThis.gtag === 'function') {
     globalThis.gtag('event', 'signup');
   }
 
   if (typeof globalThis.fbq === 'function') {
     globalThis.fbq('track', 'signup');
+    globalThis.fbq('track', 'signup_experience_level', experienceLevel);
```
capJavert, line 138: Probably only call this if `experienceLevel` is not `undefined`?

capJavert (approved):

packages/shared/src/components/auth/OnboardingAnalytics.tsx
nensidosari: Just thought we wouldn't have a case like that hopefully since everyone sets experience when they signup... 🤔 but yeah can add that

packages/shared/src/components/auth/OnboardingAnalytics.tsx
```diff
@@ -121,13 +122,22 @@ export const TiktokTracking = (): ReactElement => {
...
+  experienceLevel,
+}: TrackAnalyticsSignUpProps): void => {
   if (typeof globalThis.gtag === 'function') {
     globalThis.gtag('event', 'signup');
   }
 
   if (typeof globalThis.fbq === 'function') {
     globalThis.fbq('track', 'signup');
+    if (experienceLevel) {
+      globalThis.fbq('track', 'signup_experience_level', experienceLevel);
```
idoshamun, line 139: Unfortunately FB doesn't support sending parameters so we need this:
```
globalThis.fbq('track', `signup_${experienceLevel}`)
```

PR #3128, opened 2024-05-23
fix: squad page header

sshanzel:
## Changes
- Fix the privileged member list orientation.
- Fix the dummy stats.

Before:
Screenshot 2024-05-23 at 4 56 58 PM

After:
Screenshot 2024-05-23 at 4 58 44 PM

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done

### Preview domain
https://fix-squad-header.preview.app.daily.dev

### DIFF
--- a/packages/shared/src/components/squads/SquadPageHeader.tsx
+++ b/packages/shared/src/components/squads/SquadPageHeader.tsx
@@ -64,7 +64,7 @@ export function SquadPageHeader({
     tourIndex === TourScreenIndex.Post ||
     (isChecklistVisible && openStep === ActionType.SquadFirstPost);
   const isSquadMember = !!squad.currentMember;
-  const isFeatured = squad.flags.featured;
+  const isFeatured = squad.flags?.featured;
 
   const props = (() => {
     if (isFeatured) {
@@ -125,9 +125,9 @@ export function SquadPageHeader({
                 </>
               )}
             </Button>
-            <SquadStat count={squad.flags.totalPosts} label="Posts" />
-            <SquadStat count={squad.flags.totalViews} label="Views" />
-            <SquadStat count={squad.flags.totalUpvotes} label="Upvotes" />
+            <SquadStat count={squad.flags?.totalPosts} label="Posts" />
+            <SquadStat count={squad.flags?.totalViews} label="Views" />
+            <SquadStat count={squad.flags?.totalUpvotes} label="Upvotes" />
           </div>
         </FlexCol>
       </div>
--- a/packages/shared/src/components/squads/SquadPageHeader.tsx
+++ b/packages/shared/src/components/squads/SquadPageHeader.tsx
@@ -24,7 +24,10 @@ import { link } from '../../lib/links';
 import SquadChecklistCard from '../checklist/SquadChecklistCard';
 import { Separator } from '../cards/common';
 import { EarthIcon, LockIcon, SourceIcon, SparkleIcon } from '../icons';
-import { PrivilegedMemberItem } from './Members/PrivilegedMemberItem';
+import {
+  PrivilegedMemberContainer,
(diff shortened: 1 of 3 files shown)


packages/shared/src/components/squads/SquadPageHeader.tsx
```diff
@@ -47,6 +50,8 @@ const SquadStat = ({ count, label }: SquadStatProps) => (
   </span>
 );
 
+const MAX_PRIVILEGED_MEMBERS = 3;
```
omBratteng, line 53: Could this be moved up to config.ts or constants.ts?

packages/shared/src/components/squads/SquadPageHeader.tsx
```diff
@@ -139,10 +144,17 @@ export function SquadPageHeader({
...
-        {squad?.privilegedMembers?.map((member) => (
-          <PrivilegedMemberItem key={member.user.id} member={member} />
-        ))}
+      <div className="mt-2 flex flex-row items-center gap-3">
+        {squad.privilegedMembers
+          ?.slice(0, MAX_PRIVILEGED_MEMBERS)
+          .map((member) => (
+            <PrivilegedMemberItem key={member.user.id} member={member} />
+          ))}
+        {privilegedLength > 3 && (
```
omBratteng, line 153:
```suggestion
        {privilegedLength > MAX_PRIVILEGED_MEMBERS && (
```
?

sshanzel: Oh, right. Forgot to use it 😂

packages/shared/src/components/squads/SquadPageHeader.tsx
sshanzel: I was thinking that since this is used only here, we can maybe keep it. Do you know if we would be using this somewhere else?

packages/shared/src/components/squads/SquadPageHeader.tsx
omBratteng: 😂

packages/shared/src/components/squads/SquadPageHeader.tsx
sshanzel: Yeah, totally makes sense. Also to avoid duplication and issues in the future. Let me find a place to store this.

omBratteng: https://github.com/dailydotdev/apps/blob/main/packages/shared/src/lib/config.ts or https://github.com/dailydotdev/apps/blob/main/packages/shared/src/lib/constants.ts?

omBratteng: Maybe config is a better one.

sshanzel: Yeah, also a bit more empty. Lots of room for new constants. Thank you!

packages/shared/src/lib/config.ts
```diff
@@ -11,3 +11,5 @@ export const fallbackImages = {
   avatar:
     'https://daily-now-res.cloudinary.com/image/upload/f_auto/v1664367305/placeholders/placeholder3',
 };
+
+export const MAX_PRIVILEGED_MEMBERS = 3;
```
omBratteng, line 15: Perhaps add a prefix like `SQUAD_` or `SQUAD_HEADER_`? Or rename to `MAX_VISIBLE_PRIVILEGED_MEMBERS`, so it can be re-used?

sshanzel: I like the last one. Let's go with it 🚀

PR #3131, opened 2024-05-23
fix: share to squad on mobile

sshanzel:
## Changes
- There are two instances of Drawers, and since it closes on an outside click, the first Drawer thinks the user is clicking outside when in fact they were clicking on the second Drawer, which in fact is true.
- In order to understand when should we disable the close on outside click for the parent Drawer, we should track the state.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done

### Preview domain
https://fix-share-squad.preview.app.daily.dev

### DIFF
--- a/packages/shared/src/components/widgets/SocialShare.tsx
+++ b/packages/shared/src/components/widgets/SocialShare.tsx
@@ -1,4 +1,4 @@
-import React, { ReactElement, useContext, useState } from 'react';
+import React, { Dispatch, ReactElement, useContext } from 'react';
 import { ShareProvider, addTrackingQueryParams } from '../../lib/share';
 import { Post } from '../../graphql/posts';
 import { FeedItemPosition, postAnalyticsEvent } from '../../lib/feed';
@@ -23,6 +23,7 @@ interface SocialShareProps extends Pick<ModalProps, 'parentSelector'> {
   comment?: Comment;
   onClose?: () => void;
   commentary?: string;
+  shareToSquadState: [Squad, Dispatch<Squad>];
 }
 
 export const SocialShare = ({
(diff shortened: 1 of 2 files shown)


packages/shared/src/components/widgets/SocialShare.tsx
```diff
@@ -34,6 +35,7 @@ export const SocialShare = ({
   row,
   onClose,
   parentSelector,
+  shareToSquadState,
```
nensidosari, line 38: I'd actually think it would be better to send squadToShare and onRequestClose then shareToSquadState. Since state is managed in the higher up component it's better if it's left there, and the component here just uses the value and calls a function no matter what function.

nensidosari (approved):
I'll approve as it looks good. Please consider my comment before merging. If u think it makes sense let's change 🙏

packages/shared/src/components/widgets/SocialShare.tsx
sshanzel: We still need to set the value from this component, meaning we would still need the dispatch to be passed along.

nensidosari: we can pass it as `onRequestClose` and we define in the other component `onRequestClose={() => setSquadToShare(null)}` so in this component we can just say `onRequestClose={onRequestClose}`

nensidosari: We don't need the dispatch for anything else here right? or we do? Maybe I'm just not seeing it in the changed code, so if it's needed just never mind my comment.

sshanzel: We need it at line 87:

```
<SquadsToShare onClick={(_, squad) => setSquadToShare(squad)} />
```

PR #3132, opened 2024-05-24
fix: close modal on submission and hiding of card

sshanzel:
## Changes
- Fix the submission modal to automatically close and invalidate the query to reflect the correct status for the card.
- Fix the dismissal of card not reflecting immediately. After investigation, it was due to missing dependencies. 
- Fixed the condition for displaying the eligibility card.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

MI-359 #done

### Preview domain
https://mi-359.preview.app.daily.dev

### DIFF
--- a/packages/shared/src/hooks/useFeed.ts
+++ b/packages/shared/src/hooks/useFeed.ts
@@ -213,15 +213,16 @@ export default function useFeed<T>(
       );
     }
     return newItems;
-    // @NOTE see https://dailydotdev.atlassian.net/l/cp/dK9h1zoM
-    // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [
     feedQuery.data,
     feedQuery.isFetching,
-    adsQuery.data,
-    adsQuery.isFetching,
-    settings.showAcquisitionForm,
     settings.marketingCta,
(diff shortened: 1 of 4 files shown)


packages/shared/src/hooks/useFeed.ts
```diff
@@ -213,15 +213,16 @@ export default function useFeed<T>(
       );
     }
     return newItems;
-    // @NOTE see https://dailydotdev.atlassian.net/l/cp/dK9h1zoM
-    // eslint-disable-next-line react-hooks/exhaustive-deps
```
nensidosari, line 217: Just to make sure that for some reason the right dependencies were not set, so let's just keep a good eye when testing

nensidosari (approved):

packages/shared/src/hooks/useFeed.ts
sshanzel: Yes, thankfully this will go through internal testing so we should have ample amount of time for testing it.

PR #3136, opened 2024-05-25
fix: show notification icon on tablet

nensidosari:
## Changes
- show notification icon in tablet (issue reported here: https://dailydotdev.slack.com/archives/C01L0QXQJNB/p1715498458516419)

Screenshot 2024-05-25 at 7 35 53 AM
Screenshot 2024-05-25 at 7 35 45 AM

### Preview domain
https://fix-notification-tablet.preview.app.daily.dev

### DIFF
--- a/packages/shared/src/components/feeds/FeedNav.tsx
+++ b/packages/shared/src/components/feeds/FeedNav.tsx
@@ -21,6 +21,8 @@ import { ButtonSize, ButtonVariant } from '../buttons/common';
 import { useScrollTopClassName } from '../../hooks/useScrollTopClassName';
 import { useFeatureTheme } from '../../hooks/utils/useFeatureTheme';
 import { webappUrl } from '../../lib/constants';
+import NotificationsBell from '../notifications/NotificationsBell';
+import classed from '../../lib/classed';
 
 enum FeedNavTab {
   ForYou = 'For you',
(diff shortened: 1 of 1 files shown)


packages/shared/src/components/feeds/FeedNav.tsx
```diff
@@ -91,7 +99,7 @@ function FeedNav(): ReactElement {
         </TabContainer>
 
         {isMobile && sortingEnabled && isSortableFeed && (
-          <div className="sticky flex h-11 w-20 -translate-y-12 translate-x-[calc(100vw-100%)] items-center justify-end bg-gradient-to-r from-transparent via-background-default via-40% to-background-default pr-4">
+          <StickyNavIconWrapper className="translate-x-[calc(100vw-100%)]">
```
rebelchris, line 102: Can't we do the className with `translate-x-[calc(100vw-100%)] tablet:translate-x-[calc(100vw-180%)]`

packages/shared/src/components/feeds/FeedNav.tsx
```diff
@@ -108,7 +116,12 @@ function FeedNav(): ReactElement {
               onChange={(_, index) => setSelectedAlgo(index)}
               drawerProps={{ displayCloseButton: true }}
             />
-          </div>
+          </StickyNavIconWrapper>
+        )}
+        {!isMobile && !isLaptop && (
+          <StickyNavIconWrapper className="translate-x-[calc(100vw-180%)]">
+            <NotificationsBell compact />
+          </StickyNavIconWrapper>
```
rebelchris, line 124: This is the only icon we need to add?
Just wondering as there are quite a few additional items right

rebelchris (approved):
Non blocking comments

packages/shared/src/components/feeds/FeedNav.tsx
nensidosari: It's 2 different icons, this is sorting, the one we're fixing is notifications which show on different places.

rebelchris: But they both the same component? I mean the StickyNavIconWrapper

nensidosari: It's the wrapper but it should show on different places for different icons, not on different places for tablet/mobile so this way we can decide where to show the wrapper +component based on the component

packages/shared/src/components/feeds/FeedNav.tsx
nensidosari: Though I can remove the conditional render via js and will hide in laptop and mobile via css

PR #3141, opened 2024-05-27
fix: moderated by section to allow redirection

sshanzel:
## Changes
- Introduce a new modal to display privileged members.
- Fix the moderated by section to be clickable and redirect the user to the profile.
- Tooltip should also work.

Screenshot 2024-05-27 at 8 20 05 PM
Screenshot 2024-05-27 at 8 17 53 PM

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

MI-362 #done

### Preview domain
https://mi-362.preview.app.daily.dev

### DIFF
--- a/packages/shared/src/components/squads/SquadPageHeader.tsx
+++ b/packages/shared/src/components/squads/SquadPageHeader.tsx
@@ -24,10 +24,7 @@ import { link } from '../../lib/links';
 import SquadChecklistCard from '../checklist/SquadChecklistCard';
 import { Separator } from '../cards/common';
 import { EarthIcon, LockIcon, SourceIcon, SparkleIcon } from '../icons';
-import {
-  memberContainerClasses,
-  PrivilegedMemberItem,
-} from './Members/PrivilegedMemberItem';
+import { PrivilegedMemberItem } from './Members/PrivilegedMemberItem';
 import { formatMonthYearOnly } from '../../lib/dateFormat';
 import {
   MAX_VISIBLE_PRIVILEGED_MEMBERS_LAPTOP,
@@ -198,12 +195,9 @@ export function SquadPageHeader({
           <PrivilegedMemberItem key={member.user.id} member={member} />
         ))}
         {privilegedLength > listMax && (
-          <button
-            type="button"
-            className={classNames(
-              'h-fit font-bold text-text-tertiary typo-callout',
-              memberContainerClasses,
-            )}
+          <Button
+            variant={ButtonVariant.Tertiary}
+            className="aspect-square border border-border-subtlest-tertiary"
             onClick={() =>
               openModal({
                 type: LazyModal.PrivilegedMembers,
@@ -212,7 +206,7 @@ export function SquadPageHeader({
             }
           >
             +{privilegedLength - listMax}
-          </button>
+          </Button>
         )}
       </div>
       <SquadHeaderBar
--- a/packages/shared/src/components/squads/Members/PrivilegedMemberItem.tsx
+++ b/packages/shared/src/components/squads/Members/PrivilegedMemberItem.tsx
@@ -9,15 +9,15 @@ interface PrivilegedMemberItemProps {
   member: SourceMember;
 }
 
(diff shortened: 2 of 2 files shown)


packages/shared/src/components/modals/squads/PrivilegedMembersModal.tsx
```diff
@@ -0,0 +1,33 @@
...
+}
+
+function PrivilegedMembersModal({
+  members,
+  ...props
+}: PrivilegedMembersModalProps): ReactElement {
+  return (
+    <Modal kind={Modal.Kind.FlexibleCenter} size={Modal.Size.Medium} {...props}>
+      <Modal.Header title="Moderated by" />
+      <Modal.Body className="!p-0">
```
rebelchris, line 20: Just checking can't we re-use `UserListModal` is also has search for instance.

packages/shared/src/components/squads/SquadPageHeader.tsx
```diff
@@ -196,9 +198,21 @@ export function SquadPageHeader({
           <PrivilegedMemberItem key={member.user.id} member={member} />
         ))}
         {privilegedLength > listMax && (
-          <PrivilegedMemberContainer className="h-fit font-bold text-text-tertiary typo-callout">
+          <button
+            type="button"
+            className={classNames(
+              'h-fit font-bold text-text-tertiary typo-callout',
+              memberContainerClasses,
+            )}
```
rebelchris, line 206: This is supposed to be a custom button? 
Why is it not a guidelined one?

rebelchris (commented):
Just one question if we can't re-use one
And why its not a `Button`

packages/shared/src/components/modals/squads/PrivilegedMembersModal.tsx
sshanzel: Tried it at first, but it was an infinite modal. Thought of adding a more generic modal for this new implementation but the object is `SourceMember` which is not very common (but this one can still be reused). I tried the existing SquadMembersModal and add a filter on the prop but it can only accept one Role at a time which we need to show both admin and moderators.

packages/shared/src/components/squads/SquadPageHeader.tsx
sshanzel: From Figma, there is an override that happened. Do we proceed with the `Button` component while using `!` on its classes or do we go with a plain button?

Screenshot 2024-05-27 at 8 34 38 PM

packages/shared/src/components/modals/squads/PrivilegedMembersModal.tsx
rebelchris: Ah ok, this doesn't return infinite query?
Assumed it was a callable like that.

But let's not over complicate for now, just thought it would be nice to have one base.

packages/shared/src/components/squads/SquadPageHeader.tsx
rebelchris: Hmm maybe double check with design if they happy with this temp or they made some mistake?
(Because it feels like tertiary almost from the classes)

packages/shared/src/components/modals/squads/PrivilegedMembersModal.tsx
sshanzel: Yeah, totally agree with that. Was really hoping to introduce a generic user list modal. For this one, we are just pulling the data from the `squad.privilegedMembers` object which is not from an infinite query.

packages/shared/src/components/squads/SquadPageHeader.tsx
sshanzel: Right. Let me send a message.

sshanzel:
Updated the implementation. Preview:

Screenshot 2024-05-27 at 9 11 27 PM

rebelchris (approved):

PR #3151, opened 2024-05-29
fix: minor issues with notification page

rebelchris:
- We should not auto unsub/sub with the main push toggle as push is per device and it could cause side effects
- We didn't check which field was actually set so optimistic update could reflect empty data

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

AS-346 #done

### DIFF
--- a/packages/webapp/pages/account/notifications.tsx
+++ b/packages/webapp/pages/account/notifications.tsx
@@ -150,22 +150,13 @@ const AccountNotificationsPage = (): ReactElement => {
       NotificationCategory.ReadingReminder,
     );
 
-    if (value) {
-      trackEvent({
-        event_name: AnalyticsEvent.ScheduleReadingReminder,
-        extra: JSON.stringify({
-          hour: readingTimeIndex,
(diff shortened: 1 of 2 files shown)


omBratteng (approved):

packages/webapp/pages/account/notifications.tsx
```diff
@@ -150,22 +150,13 @@ const AccountNotificationsPage = (): ReactElement => {
...
-          timezone: user?.timezone,
-        }),
-      });
-      subscribePersonalizedDigest({
-        type: UserPersonalizedDigestType.ReadingReminder,
-      });
-    } else {
-      unsubscribePersonalizedDigest({
-        type: UserPersonalizedDigestType.ReadingReminder,
-      });
```
omBratteng, line 167: I think we probably shouldn't have disabled these, as it's now impossible to subscribe to reading reminders (no mutations are triggered)

PR #3154, opened 2024-05-29
As 345 sources directory

nensidosari:

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done

### Preview domain
https://as-345-sources-directory.preview.app.daily.dev


packages/shared/src/components/widgets/PostUsersHighlights.tsx
```diff
@@ -88,7 +103,7 @@ const Image = (props: SourceAuthorProps) => {
   );
 };
 
-const UserHighlight = (props: SourceAuthorProps) => {
+export const UserHighlight = (props: UserHighlightProps): ReactElement => {
```
rebelchris, line 106: I was confused by this file changes (based on name)
Maybe we should abstract some of these to different components?

(not blocking ofcourse)

packages/shared/src/graphql/sources.ts
```diff
@@ -82,6 +82,7 @@ export interface Source {
   headerImage?: string;
   color?: string;
   description?: string;
+  createdAt?: Date;
```
rebelchris, line 85: Probably not exposed for this one right?
We can just query based on it and not return it.

rebelchris (approved):
Awesome, think we can about merge it and solve from there?

packages/shared/src/components/widgets/PostUsersHighlights.tsx
nensidosari: I thought of it as well as this component is too big and does to many stuff.

packages/shared/src/graphql/sources.ts
nensidosari: yeah true we won't need to expose it, just query by it would be enough

PR #3156, opened 2024-05-29
feat: improved shared post cards

omBratteng:
## Changes

Makes the shared posts cards look like the article posts cards.
https://dailydotdev.slack.com/archives/C015T1Y237Z/p1715776043294589

AS-349

### Preview domain
https://as-349-micro-improve-shared-post-card.preview.app.daily.dev

### DIFF
--- a/packages/shared/src/components/cards/SharePostCard.tsx
+++ b/packages/shared/src/components/cards/SharePostCard.tsx
@@ -51,6 +51,9 @@ export const SharePostCard = forwardRef(function SharePostCard(
   const isVideoType = isVideoPost(post);
   const improvedSharedPostCard = useFeature(feature.improvedSharedPostCard);
 
+  // Just for this experiment, we are modifying the post object to use set the
+  // image from the shared post, this is just to avoid modifying the `PostCardFooter`
+  // component to handle this case just for this experiment.
   if (improvedSharedPostCard) {
     // eslint-disable-next-line no-param-reassign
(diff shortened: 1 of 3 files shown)


rebelchris (approved):
I guess for experiment run this is fine for now.
Kind of expected we would just render the different component at first, but to be fast/test it's fine for me.

CC: Thoughts from @nensidosari and @sshanzel ?

omBratteng:
I was thinking about rendering the article card component instead, but I would've needed to start messing with the enums and stuff. Just found it quicker to copy the relevant parts from the article component into the shared post component.

@idoshamun I forgot about the tablet/mobile card, you want to update them as well?

sshanzel (approved):
Yeah, since this is going to be a quick experiment, I think we can go for it, but probably we can mention on the cleanup ticket the other option as it will be easier without all the conditions for the `enums` like Ole-Martin has mentioned.

idoshamun:
@omBratteng yes please

nensidosari (approved):
I'm fine with this for fast validation. Let's place something to clean up and reuse our existing cards for this if it wins. 

Personally - I liked the old UI better 😢 

I see u mentioned tablet and mobile - just make sure to check list mode since changes to tablet and mobile will affect that too.

omBratteng:
If the experiment wins, we can cleanup by using the articles card for shared posts, and just add the required changes to that.

PR #3162, opened 2024-05-30
fix: don't override ← or → keys in search panel

omBratteng:
Gives back control of `←` or `→` keys when inside input element so it is possible to navigate in input text.

AS-330

### Preview domain
https://as-330-fix-arrow-keys-in-search-bar.preview.app.daily.dev


nensidosari (approved):

rebelchris (approved):
Validated it works :)

PR #3166, opened 2024-05-31
feat: full background image on onboarding page

omBratteng:
## AS-353 Changes

Extends `onboarding_visual` to accept full background images

<table>
<tr>
 <td>Mobile
 <td>Desktop
<tr>
 <td>
 <td>
</table>

## Testing

This can be tested by setting the `ONBOARDING_VISUAL` value in Growthbook devtools to 
```json
{
  "fullBackground": {
    "mobile": "https://daily-now-res.cloudinary.com/image/upload/s--EwsBTBt6--/f_auto/v1716969841/dailydev_where_developers_suffer_together_mobile_shkn1w",
    "desktop": "https://daily-now-res.cloudinary.com/image/upload/s--r2ffZPB4--/f_auto/v1716969841/dailydev_where_developers_suffer_together_sfvfog"
  }
}
```

### Preview domain
https://as-353-onboarding-background.preview.app.daily.dev


packages/webapp/pages/onboarding.tsx
```diff
@@ -73,6 +73,17 @@ import { LoggedUser } from '@dailydotdev/shared/src/lib/user';
...
+type OnboardingVisual = {
+  showCompanies?: boolean;
+  fullBackground?:
+    | {
+        mobile?: string;
+        desktop?: string;
+      }
+    | false;
+  image?: string;
+};
```
nensidosari, line 85: Totally fine like this as well, just another suggestion that makes more sense in my opinion, but no strong opinion. Only change if it makes sense for u as well.
```suggestion
type OnboardingVisual = {
  showCompanies?: boolean;
  fullBackground?: boolean;
  image?: string;
  imageMobile?:string
};
```

packages/shared/src/lib/featureManagement.ts
```diff
@@ -24,6 +24,10 @@ const feature = {
   bookmarkOnCard: new Feature('bookmark_on_card', false),
   onboardingVisual: new Feature('onboarding_visual', {
     showCompanies: true,
+    fullBackground: {
+      mobile: cloudinary.onboarding.fullBackground.mobile,
+      desktop: cloudinary.onboarding.fullBackground.desktop,
+    },
```
nensidosari, line 30: Here we're setting it as a default value, so let's better leave the last value at the moment if this is going to be an experiment

nensidosari (approved):
Looks good, one suggestion that for me looks better, but up to u.

Also, from this issue Ante raised here, I believe it's best to check with product for a light mode image?

packages/shared/src/lib/featureManagement.ts
omBratteng: As long as they don't get sent from growthbook, it'll be undefined. But I'll set `fullBackground` to false

packages/webapp/pages/onboarding.tsx
omBratteng: I think it's fine to keep it in `fullBackground`. It defaults to false (or undefined if not specified)

PR #3173, opened 2024-06-02
fix: reduce flickering of images of profile/source images

omBratteng:
- Reduces flickering of profile and source images in sidebar and header
- Removes flickering of sources on source page

AS-355

### Preview domain
https://as-355-fix-image-flash.preview.app.daily.dev

### DIFF
--- a/packages/shared/src/components/MainLayout.tsx
+++ b/packages/shared/src/components/MainLayout.tsx
@@ -116,29 +116,6 @@ function MainLayoutComponent({
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [isNotificationsReady, unreadCount, hasTrackedImpression]);
 
-  const RenderSidebar = () => {
-    if (sidebarRendered === null || (sidebarRendered && !showSidebar)) {
-      return null;
-    }
-
-    return (
-      <Sidebar
-        promotionalBannerActive={isBannerAvailable}
-        sidebarRendered={sidebarRendered}
-        openMobileSidebar={openMobileSidebar}
-        onNavTabClick={onNavTabClick}
-        enableSearch={enableSearch}
-        activePage={activePage}
-        showDnd={showDnd}
-        dndActive={dndActive}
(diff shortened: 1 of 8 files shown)


packages/shared/src/components/MainLayout.tsx
```diff
@@ -200,7 +177,22 @@ function MainLayoutComponent({
...
+            enableSearch={enableSearch}
+            activePage={activePage}
+            showDnd={showDnd}
+            dndActive={dndActive}
+            isNavButtons={isNavItemsButton}
+            onShowDndClick={onShowDndClick}
+            onLogoClick={onLogoClick}
+            setOpenMobileSidebar={() => onMobileSidebarToggle(false)}
+          />
+        )}
```
omBratteng, line 195: I removed the `sidebarRendered === null` check, as `useSidebarRendered` either returns `true` or `false`. And I moved it down here, as something with it being in a const made it re-render often

packages/webapp/components/common.tsx
```diff
@@ -70,15 +76,9 @@ export const TopList = ({
...
   const isMobile = useViewSize(ViewSize.MobileL);
-  const MobileDiv = classed(
-    'div',
-    'flex flex-col border-b border-b-border-subtlest-tertiary p-4',
-    className,
-  );
-  const CardElement = classed(Card, '!p-4 !max-h-none', className);
-  const Wrapper = isMobile ? MobileDiv : CardElement;
+  const Wrapper = isMobile ? TopListMobileDiv : TopListCardElement;
   return (
```
omBratteng, line 80: Option was either memoize these or move outside, moving outside was easiest

rebelchris (approved):
Maybe that sidebar issue was the main one.
Hoping it will solve the shortcuts as well, so let's give it a try

nensidosari (approved):

sshanzel (approved):
Looks great! 🚀

PR #3181, opened 2024-06-04
fix: streaks popup to not show

nensidosari:
## Changes
- missed one part on not showing the modal

### Preview domain
https://fix-streak-popup.preview.app.daily.dev


omBratteng (approved):

packages/shared/src/components/modals/streaks/NewStreakModal.tsx
```diff
@@ -58,9 +58,8 @@ export default function NewStreakModal({
       trackEvent({
         event_name: AnalyticsEvent.DismissStreaksMilestone,
       });
+      completeAction(ActionType.DisableReadingStreakMilestone);
```
rebelchris, line 61: Wait we use action for this, does it work with disabling it again?
Not sure we ever clear actions?

rebelchris (approved):
The code is ok, just wondering about using actions for this purpose.
Why didn't we use settings for it?

packages/shared/src/components/modals/streaks/NewStreakModal.tsx
nensidosari: This we can never disable, this is only about the modals. The streaks in general we're using settings - u can enable disable again. Discussion here https://dailydotdev.slack.com/archives/C05TQL6HWTH/p1717146432887299

rebelchris: Oh ok this one is a once you remove it never get it back thing :)

nensidosari: There's also no way ui wise even if it was with settings to get this back

PR #3187, opened 2024-06-04
fix: re-enable reading reminder toggle

omBratteng:
## Changes

Re-adds the possibility to actually toggle reading reminder that got removed in #3151

### Preview domain
https://fix-reading-reminder-toggle.preview.app.daily.dev

### DIFF
--- a/packages/webapp/pages/account/notifications.tsx
+++ b/packages/webapp/pages/account/notifications.tsx
@@ -150,13 +150,22 @@ const AccountNotificationsPage = (): ReactElement => {
       NotificationCategory.ReadingReminder,
     );
 
-    trackEvent({
-      event_name: AnalyticsEvent.ScheduleReadingReminder,
-      extra: JSON.stringify({
-        hour: readingTimeIndex,
-        timezone: user?.timezone,
-      }),
-    });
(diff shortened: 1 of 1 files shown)


packages/webapp/pages/account/notifications.tsx
```diff
@@ -166,8 +175,6 @@ const AccountNotificationsPage = (): ReactElement => {
       NotificationCategory.Product,
     );
 
-    onToggleReadingReminder(!isSubscribed);
-
```
omBratteng, line 170: Removed it from onTogglePush, so it doesn't get automagically enabled on notification toggle.

rebelchris (changes requested):
Oh somehow missed this indeed.
But now it will sub/unsub on the global push agian so we'll need to make some changes there so it doesn't mutate on those

rebelchris (approved):

packages/webapp/pages/account/notifications.tsx
rebelchris: Yeah somehow totally missed this one 👍

sshanzel (approved):
🚀

PR #3188, opened 2024-06-04
feat: reading streak reminder toggle

omBratteng:
## Changes

AS-359

Needs https://github.com/dailydotdev/daily-api/pull/1968 to be able to verify the toggle

### Preview domain
https://preview.app.daily.dev

### DIFF
--- a/packages/shared/src/lib/analytics.ts
+++ b/packages/shared/src/lib/analytics.ts
@@ -145,6 +145,7 @@ export enum AnalyticsEvent {
   // Reading Streaks
   OpenStreaks = 'open streaks',
   DismissStreaksMilestone = 'dismiss streaks milestone',
+  ScheduleReadingStreakReminder = 'schedule reading streak reminder',
   // 404 page
   View404Page = '404 page',
   // Refresh feed
@@ -218,6 +219,7 @@ export enum NotificationCategory {
   Product = 'product',
   Digest = 'digest',
   ReadingReminder = 'reading reminder',
+  ReadingStreakReminder = 'reading streak reminder',
 }
 
 export enum NotificationTarget {
--- a/packages/shared/src/lib/analytics.ts
+++ b/packages/shared/src/lib/analytics.ts
@@ -145,7 +145,7 @@ export enum AnalyticsEvent {
   // Reading Streaks
   OpenStreaks = 'open streaks',
   DismissStreaksMilestone = 'dismiss streaks milestone',
-  ScheduleReadingStreakReminder = 'schedule reading streak reminder',
+  ScheduleStreakReminder = 'schedule streak reminder',
   // 404 page
   View404Page = '404 page',
   // Refresh feed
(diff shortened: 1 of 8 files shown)


packages/shared/src/lib/analytics.ts
```diff
@@ -144,6 +144,7 @@ export enum AnalyticsEvent {
   // Reading Streaks
   OpenStreaks = 'open streaks',
   DismissStreaksMilestone = 'dismiss streaks milestone',
+  ScheduleReadingStreakReminder = 'schedule reading streak reminder',
```
rebelchris, line 147: @idoshamun It's now a engineering decision, this works for you, basically copy of reading reminder, but for streaks specific

rebelchris (approved):
Looks good :)

packages/shared/src/lib/analytics.ts
omBratteng: There wasn't anything in the spec, so I was going to ask about it, But I see you beat me to it.

idoshamun: Good call my suggestion would be:

```suggestion
  ScheduleReadingStreakReminder = 'schedule streak reminder',
```

packages/webapp/pages/account/notifications.tsx
```diff
@@ -398,6 +430,24 @@ const AccountNotificationsPage = (): ReactElement => {
             <Checkbox name="newsletter" checked disabled>
               Activity (mentions, replies, upvotes, etc.)
             </Checkbox>
+            <Checkbox
```
nensidosari, line 433: Should we hide this behind the isStreaksEnabled? and also unsubscribe if user has enabled the notification but disables the reading streaks at all after

nensidosari (changes requested):
Just one comment on reading streaks that we now have behind a flag. Otherwise looks good.

packages/webapp/pages/account/notifications.tsx
omBratteng: 👍

packages/shared/src/lib/analytics.ts
omBratteng: Changed it

packages/webapp/pages/account/notifications.tsx
omBratteng: Instead of hiding it, we could show it but as unchecked and disabled?
What do you think @rebelchris @nensidosari?

nensidosari: If we show it it might be good to also show any info when u hover over that says they need to enable reading streaks if they want this setting. I think it's best to check with product/design for this. 

On my opinion to unblock and move forward I'd say let's hide, and then add as u suggested if product/design confirms as a later step - no need to do in this PR.

PS.
do we have any other checkbox here that stays disabled for some reason?

omBratteng: I added the mutation for unsubscribing to the mutation for opt-out of streaks dde7ce7
image

omBratteng: But I will also add a check in API just to be sure, you know, edge cases and all

rebelchris (approved):

nensidosari (approved):

PR #3199, opened 2024-06-07
feat: explore tab experiment

sshanzel:
## Changes
- Display a tabbed component on laptop for switching to popular/by date.
- Dropdown option on mobile to switch popular/by date.

Note:
- No need for any experiment flags since this is not going to be accessible anywhere unless you type the URL through webapp.

Previews:

https://github.com/dailydotdev/apps/assets/13744167/33616eb9-be99-4ccb-86f1-6d16e021f99a

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

MI-375 #done

### Preview domain
https://mi-375.preview.app.daily.dev


packages/webapp/components/common.tsx
```diff
@@ -33,33 +27,6 @@ export const ListItem = ({
   );
 };
 
-const BreadCrumbsWrapper = classed(
```
sshanzel, line 36: Just moved to its separate file.

packages/shared/src/components/feeds/FeedContainer.tsx
```diff
@@ -280,7 +284,6 @@ export const FeedContainer = ({
             <div
               className={classNames(
                 'grid',
-                isSearch && !shouldUseListFeedLayout && 'mt-8',
```
sshanzel, line 283: Not really used anymore. The `isSearch` is never true with our current search.

packages/shared/src/components/utilities/common.tsx
```diff
@@ -164,6 +164,10 @@ export type HTMLElementComponent<T = HTMLElement> = React.FC<HTMLAttributes<T>>;
 export enum SharedFeedPage {
   MyFeed = 'my-feed',
   Popular = 'popular',
+  Explore = 'explore',
+  ExploreLatest = 'explorelatest',
+  ExploreDiscussed = 'explorediscussed',
+  ExploreUpvoted = 'exploreupvoted',
```
sshanzel, line 170: Not a fan, but due to the way we store feed names, we don't allow special characters.

packages/shared/src/components/fields/Dropdown.tsx
```diff
@@ -143,25 +153,34 @@ export function Dropdown({
         tabIndex={0}
         aria-haspopup="true"
         aria-expanded={isVisible}
-      >
-        {icon &&
+        icon={
```
sshanzel, line 156: Properly use the `icon` property.

packages/shared/src/components/fields/Dropdown.tsx
```diff
@@ -143,25 +153,34 @@ export function Dropdown({
...
-            'ml-auto text-xl transition-transform group-hover:text-text-tertiary',
-            isVisible ? 'rotate-0' : 'rotate-180',
-            styles.chevron,
-            className.chevron,
-          )}
-        />
+          })
+        }
+      >
+        {iconOnly ? null : (
```
sshanzel, line 164: The `iconOnly` isn't really doing what you would expect to do, fixed it here.

packages/shared/src/components/fields/Dropdown.tsx
```diff
@@ -124,9 +127,16 @@ export function Dropdown({
   const fullScreen = openFullScreen ?? isMobile;
 
   return (
-    <div
-      className={classNames(styles.dropdown, className.container)}
-      {...props}
+    <ConditionalWrapper
```
sshanzel, line 130: Believe it or not, without this, there is a crazy bug going on, the classes used here are magically passed to some other elements 😭

packages/shared/src/components/header/FeedExploreHeader.tsx
```diff
@@ -0,0 +1,97 @@
...
+        ) : (
+          <TabContainer
+            controlledActive={tabsToFeedMap[path]}
+            className={{ header: 'border-b-0' }}
+            shouldMountInactive
+          >
+            {Object.entries(urlToTab).map(([url, label]) => (
+              <Tab key={label} label={label} url={url} />
+            ))}
+          </TabContainer>
```
nensidosari, line 79: This isn't something we do with all tabs we have extension/web app? would it be worth this was part of tab components to differentiate between both depending on where it's used?

packages/shared/src/components/utilities/common.tsx
nensidosari: What about like`my-feed` in kebab case?

nensidosari (commented):
Few things to check, but looks very good overall. Great work!! 🚀

packages/shared/src/components/MainFeedLayout.tsx
```diff
@@ -224,26 +262,41 @@ export default function MainFeedLayout({
...
 
-      return query.variables;
+      return config.variables;
     };
 
     const variables = getVariables();
+    const classes = isExplore && {
+      wrapper: '!min-h-0',
+      emptyScreen: 'mt-8 h-auto',
+    };
```
rebelchris, line 299: Why does this one only have specific needs?
Is there no other way we can maybe fix it?

packages/shared/src/components/MainFeedLayout.tsx
```diff
@@ -252,25 +305,43 @@ export default function MainFeedLayout({
         user,
         ...Object.values(variables ?? {}),
       ),
-      query: query.query,
+      query: config.query,
       variables,
-      emptyScreen: <FeedEmptyScreen />,
-      header: null,
-      actionButtons: feedWithActions && (
+      header: !isLaptop && (isExplorePopular || isExploreLatest) && (
```
rebelchris, line 310: Should this not be under the actionButtons already?

packages/shared/src/components/feeds/FeedContainer.tsx
```diff
@@ -280,7 +288,8 @@ export const FeedContainer = ({
...
+                !isLaptop && (isExplorePopular || isExploreLatest) && 'mt-4',
+                isSearch && !shouldUseListFeedLayout && !isExplore && 'mt-8',
```
rebelchris, line 292: Here as well wondering if we can uniform it somehow, not sure what the issue is design wise but we can suggest things to keep it neater.

packages/shared/src/components/fields/Dropdown.tsx
rebelchris: Also love to know the issue, feels like it's not needed 😅

packages/shared/src/components/header/BreadCrumbs.tsx
```diff
@@ -0,0 +1,31 @@
+import React, { PropsWithChildren, ReactElement } from 'react';
+import classed from '../../lib/classed';
+import { Button, ButtonSize, ButtonVariant } from '../buttons/Button';
+import { HomeIcon } from '../icons';
+import { WithClassNameProps } from '../utilities';
+
+const BreadCrumbsWrapper = classed(
+  'div',
+  'hidden h-10 gap-0.5 items-center p-1.5 text-border-subtlest-tertiary laptop:flex',
```
rebelchris, line 9: Wait that's not a text color actually? 👀

rebelchris (commented):
I'm mainly wondering if we can't uniform the dropdown/feed behaviour it will make future things easier/quicker and cleaner.

packages/shared/src/components/header/BreadCrumbs.tsx
sshanzel: Oh, looks like it should probably be `text-text-tertiary` 👀 

Will verify this.

packages/shared/src/components/utilities/common.tsx
sshanzel: Each page is under the explore directory. Feels like it is more appropriate instead of making each page under the root with hyphens, but that is just my opinion. I could be wrong though.

packages/shared/src/components/fields/Dropdown.tsx
sshanzel: I will do a loom recording to explain it easier 😭

packages/shared/src/components/MainFeedLayout.tsx
```diff
@@ -137,23 +163,36 @@ export default function MainFeedLayout({
...
   });
+  const isLaptop = useViewSize(ViewSize.Laptop);
   const feedVersion = useFeature(feature.feedVersion);
-  const { isUpvoted, isPopular, isSortableFeed, isCustomFeed } = useFeedName({
+  const {
+    isUpvoted,
+    isPopular,
+    isExplore,
+    isExplorePopular,
+    isExploreLatest,
```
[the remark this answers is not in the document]
sshanzel: The `isExplorePopular`. Maybe I should rename the `isExplore` to `isAnyExplore`.

packages/shared/src/components/header/FeedExploreHeader.tsx
sshanzel: The extension and webapp would behave differently when the tabs are clicked. The `TabContainer` has a built-in function to do redirection. I would've used `TabList` on both of them if it had supported the redirection as mentioned.

packages/shared/src/components/MainFeedLayout.tsx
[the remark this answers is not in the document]
nensidosari: Yeah maybe

packages/shared/src/components/utilities/common.tsx
nensidosari: nah no, it's better the way u've done it

packages/shared/src/components/MainFeedLayout.tsx
sshanzel: The placement of the empty screen is not the absolute center of the screen anymore, and with the existing classes, it produces a scrollbar unnecessarily. Arguably not needed, but UX-wise, I think it is better to not have the scrollbar.

packages/shared/src/components/feeds/FeedContainer.tsx
sshanzel: The thing here is that it should not be part of the `FeedLayout` itself, might sound counter-intuitive, but when the screen is empty, you lose the header tabs to navigate between the options. At first, I was rendering both of them at the `header` property, but then the explained case above surfaced.

packages/shared/src/components/MainFeedLayout.tsx
sshanzel: We can actually do it there, I don't have any strong preference on where to put this. One too many options 😂

packages/shared/src/components/MainFeedLayout.tsx
rebelchris: Hmm and we can't put it on "all" empty states?

packages/shared/src/components/MainFeedLayout.tsx
rebelchris: Ah if we can uniform it in one way of the other I'm fine with that :)

sshanzel: Yep. Updated it to use the same property like suggested to make it more consistent.

packages/shared/src/components/MainFeedLayout.tsx
sshanzel: Let me check, though as I remember it looks fine on the current setup. Just with the addition of the header tabs that broke it.

sshanzel: What do you think of the screenshots below? The CTA became more centered on the screen than being more at the bottom side. Not sure on mobile if it is better to be near the bottom or the center. However, on laptop, we also removed the scrollbar.

Before:
Screenshot 2024-06-11 at 5 16 02 PM
Screenshot 2024-06-11 at 5 15 56 PM
Screenshot 2024-06-11 at 5 15 46 PM

After adding the same classes:
Screenshot 2024-06-11 at 5 14 03 PM
Screenshot 2024-06-11 at 5 14 19 PM
Screenshot 2024-06-11 at 5 13 56 PM

packages/shared/src/components/header/FeedExploreHeader.tsx
nensidosari: Well we're seeing them in one file or another 😅 Anyways I think it's ok, we can refactor later if another thing like this comes our way

sshanzel: Yeah, I think we can clean this up once we decide to replace the Popular page with the new Explore page eventually. I mean as this is technically an experiment.

packages/shared/src/components/MainFeedLayout.tsx
rebelchris: Such a edge-case anyway I'm fine with it.

packages/shared/src/components/MainFeedLayout.tsx
```diff
@@ -137,23 +162,36 @@ export default function MainFeedLayout({
...
     shouldUseCommentFeedLayout,
     FeedPageLayoutComponent,
   } = useFeedLayout();
-  let query: { query: string; variables?: Record<string, unknown> };
 
-  if (feedName) {
+  const config = useMemo(() => {
+    if (!feedName) {
+      return { query: null };
+    }
```
rebelchris, line 193: OMG! Love the early return optimization!

packages/shared/src/components/MainFeedLayout.tsx
```diff
@@ -224,26 +257,56 @@ export default function MainFeedLayout({
...
 
-      return query.variables;
+      return config.variables;
     };
 
     const variables = getVariables();
+    const classes = isAnyExplore && {
+      wrapper: '!min-h-0',
+      emptyScreen: 'mt-8 h-auto',
+    };
```
rebelchris, line 294: Ah was kind of hoping we could avoid sending this at all.
But if it's too much hassle never mind :)

packages/shared/src/components/MainFeedLayout.tsx
```diff
@@ -224,26 +257,56 @@ export default function MainFeedLayout({
...
+          options={algorithmsList}
+          onChange={(_, index) => setSelectedAlgo(index)}
+        />
+      );
+    const controlFeedActions = feedWithActions && (
+      <SearchControlHeader
+        algoState={[selectedAlgo, setSelectedAlgo]}
+        feedName={feedName}
+      />
+    );
```
rebelchris, line 309: Still isn't this basically the exact same component?

rebelchris (approved):
Minor query, but non blocking from my side.

packages/shared/src/components/MainFeedLayout.tsx
sshanzel: Yeah, we can do that. Less code to maintain. Was just worried the scrollbar and placement of the CTA was quite a turn-off.

https://github.com/dailydotdev/apps/pull/3199#discussion_r1634502509

packages/shared/src/components/MainFeedLayout.tsx
sshanzel: Unfortunately, we don't render this component on mobile which is exactly the opposite of what we want - which is to render on mobile. Also, there are certain elements from there that we don't to render, like the streak-related ones.

packages/shared/src/components/fields/Dropdown.tsx
nensidosari: 🤣

nensidosari (approved):
Looks great! 🚀

packages/shared/src/components/MainFeedLayout.tsx
rebelchris: Ah ok :)

PR #3204, opened 2024-06-07
fix: check is action fetched

nensidosari:
## Changes

- check isActionFetched

### Preview domain
https://fix-action-streaks-modal.preview.app.daily.dev

### DIFF
--- a/packages/shared/src/components/modals/BootPopups.tsx
+++ b/packages/shared/src/components/modals/BootPopups.tsx
@@ -152,7 +152,7 @@ export const BootPopups = (): null => {
    * Boot popup for streaks milestone
    */
   useEffect(() => {
-    if (shouldHideStreaksModal) {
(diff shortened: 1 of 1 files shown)


packages/shared/src/components/modals/BootPopups.tsx
```diff
@@ -152,7 +152,7 @@ export const BootPopups = (): null => {
    * Boot popup for streaks milestone
    */
   useEffect(() => {
-    if (shouldHideStreaksModal) {
+    if (shouldHideStreaksModal || !isActionsFetched) {
```
capJavert, line 155: Should it not be move above to just make `shouldHideStreaksModal` true when actions are still not fetched?

nensidosari: done

capJavert (approved):

rebelchris (approved):

PR #3208, opened 2024-06-07
chore: remove sidebar closed experiment

nensidosari:
## Changes
- remove sidebar closed experiment since it won

- see thread https://dailydotdev.slack.com/archives/C02JAUF8HJL/p1717669215632889

## Events

- [x] MobileL (420px)
- [x] Tablet (656px)
- [x] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

MI-383 #done

### DIFF
--- a/packages/shared/src/contexts/SettingsContext.tsx
+++ b/packages/shared/src/contexts/SettingsContext.tsx
@@ -183,11 +183,7 @@ export const SettingsContextProvider = ({
     const updatedSettings = settings;
 
     if (settings.sidebarExpanded) {
-      const sidebarClosed = getFeatureValue(feature.sidebarClosed);
-
-      if (sidebarClosed) {
-        updatedSettings.sidebarExpanded = false;
-      }
+      updatedSettings.sidebarExpanded = false;
     }
 
     await updateRemoteSettingsFn(updatedSettings, bootUserId);
--- a/packages/shared/src/lib/featureManagement.ts
+++ b/packages/shared/src/lib/featureManagement.ts
@@ -35,7 +35,6 @@ const feature = {
   onboardingMostVisited: new Feature('onboarding_most_visited', false),
   shareExperience: new Feature('share_experience', false),
   bookmarkLoops: new Feature('bookmark_loops', false),
-  sidebarClosed: new Feature('sidebar_closed', false),
   tagSourceSocialProof: new Feature(
     'tag_source_social_proof',
     TagSourceSocialProof.Control,


rebelchris (approved):
Guess since it only happens on auth we can omit the date check 🙏

sshanzel (commented):
I think it would be better to make the changes in the backend to ensure 100% this will always be false on creation. Then we will cleanup this if statement as it is no longer needed once done.

When cleaning on this side, we should remove the business logic here and update the default values from FE as well.

https://dailydotdev.slack.com/archives/C02JAUF8HJL/p1717423135710919?thread_ts=1716515900.040149&cid=C02JAUF8HJL

nensidosari:
@sshanzel yeah, will clean this whole thing up then and change the default config in backend.

nensidosari:
https://github.com/dailydotdev/daily-api/pull/1980

packages/shared/src/contexts/SettingsContext.tsx
```diff
@@ -180,17 +177,7 @@ export const SettingsContextProvider = ({
...
-    if (settings.sidebarExpanded) {
-      const sidebarClosed = getFeatureValue(feature.sidebarClosed);
-
-      if (sidebarClosed) {
-        updatedSettings.sidebarExpanded = false;
-      }
-    }
-
-    await updateRemoteSettingsFn(updatedSettings, bootUserId);
+    await updateRemoteSettingsFn(settings, bootUserId);
```
capJavert, line 180: With `false` being new default we should probably also look at the layout shift on post pages and decide which state we want supported there. Since now the padding during SSR expects menu to be there. cc: @rebelchris you will remember this as well when you optimized SSR there

nensidosari: Yeah true, I remember that discussion. Let me take a look

nensidosari: Previous thread where this was discussed https://dailydotdev.slack.com/archives/C01L0QXQJNB/p1714294470698459

nensidosari:
@capJavert with [this](https://dailydotdev.slack.com/archives/C072B341DC6/p1718025007351989) being discussed I don't think we can yet do anything about layout shifts until we figure out where we're at with this.

I still think we can cleanup here, and then on onboarding checklist PR we can do the same we were doing here with sidebar default set to true depending on that experiment.

packages/shared/src/contexts/SettingsContext.tsx
capJavert: As long as we make sure everyone knows what will happen now for new users I am fine with it 👍

capJavert (approved):

sshanzel (approved):
Looks great! :shipit: 

Before merging, there is one last place we need to update, at the `SettingsContext`, we should update the default value there.

PR #3210, opened 2024-06-12
fix:layout density issues

nensidosari:
## Changes
- disabling layout mode for mobile/tablet & adding a tooltip saying why it's disabled
- disabling density on list mode and mobile/tablet & adding a tooltip saying why it's disabled
- since the density settings where somehow affecting the layouts, we're now using settings only from useFeedLayout hook, which checks if the layout is list mode or mobile/tablet and sets the settings to 'eco' which is the default value.
- we're not directly modifying settings with the above, this way if user gets back to card layout they can continue having the density they choose.

- both list mode or mobile/tablet only have one density setting, when users play with these density settings weird layout issues may appear
- issues raised in these threads:
- https://dailydotdev.slack.com/archives/C01473UNK9C/p1718030989392199
- https://dailydotdev.slack.com/archives/C01473UNK9C/p1717996969816239
- https://dailydotdev.slack.com/archives/C01473UNK9C/p1717753320852489
- https://dailydotdev.slack.com/archives/C02AA18RD1D/p1717655445750759

I personally haven't been able to replicate the exact issue mentioned, but I believe this fix should do it.

## Events

- [x] MobileL (420px)
- [x] Tablet (656px)
- [x] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

MI-385 #done


packages/shared/src/components/fields/CustomSwitch.tsx
```diff
@@ -68,60 +77,69 @@ export function CustomSwitch({
   const difference = Math.abs(rightWidth - leftWidth);
 
   return (
-    <label
```
nensidosari, line 71: The only change here is wrapping this whole thing with the tooltip

packages/shared/src/components/fields/CustomSwitch.tsx
```diff
@@ -68,60 +77,69 @@ export function CustomSwitch({
...
+        htmlFor={inputId}
+      >
+        <input
+          id={inputId}
+          name={name}
+          type="checkbox"
+          className="absolute h-0 w-0 opacity-0"
+          checked={checked}
+          onChange={onToggle}
+          disabled={disabled}
```
nensidosari, line 95: This value is also now added to the input

packages/shared/src/components/fields/CustomSwitch.tsx
```diff
@@ -68,60 +77,69 @@ export function CustomSwitch({
...
+                className: classNames(
+                  (rightContent as ReactElement).props.className,
+                  rightClasses,
+                ),
+              })}
+        </ContentContainer>
+        <span
+          className={classNames(
+            'absolute inset-0 my-auto h-7 rounded-10 opacity-24 group-hover:opacity-32',
+            disabled ? 'bg-surface-disabled' : 'bg-raw-cabbage-50',
```
nensidosari, line 126: And also the disabled colors

packages/shared/src/components/fields/CustomSwitch.tsx
```diff
@@ -68,60 +77,69 @@ export function CustomSwitch({
...
+          )}
+        />
+        <span
+          className={classNames(
+            'absolute  top-0 z-1 h-full rounded-12',
+            styles.knob,
+            disabled && '!bg-surface-disabled',
```
nensidosari, line 133: And also the disabled colors

packages/shared/src/components/Settings.tsx
```diff
@@ -117,6 +131,13 @@ export default function Settings({
           checked={insaneMode}
           className="mx-1.5"
           onToggle={toggleInsaneMode}
+          disabled={!isLaptop}
+          tooltip={
+            !isLaptop && {
+              content: 'Layout will be fixed for mobile and tablet',
```
rebelchris, line 137: Did you check this copy with Sab, sounds a bit funky.
Would expect: Layout is fixed for mobile and tablet view

packages/shared/src/components/Settings.tsx
```diff
@@ -135,6 +156,13 @@ export default function Settings({
           options={densities}
           value={spaciness}
           onChange={setSpaciness}
+          tooltip={
+            showDensityTooltip && {
+              content: getDensityTooltipContent(),
```
rebelchris, line 161: This might cause some flickers because of evaluation? 
(not 100% sure, also not urgent) and same copy question

packages/shared/src/components/fields/CustomSwitch.tsx
```diff
@@ -68,60 +77,69 @@ export function CustomSwitch({
...
-        }}
-      />
-    </label>
+        htmlFor={inputId}
+      >
+        <input
+          id={inputId}
+          name={name}
+          type="checkbox"
+          className="absolute h-0 w-0 opacity-0"
```
rebelchris, line 92:
```suggestion
          className="absolute size-0 opacity-0"
```

packages/shared/src/components/fields/CustomSwitch.tsx
```diff
@@ -68,60 +77,69 @@ export function CustomSwitch({
...
-      </ContentContainer>
-      <span className="absolute inset-0 my-auto h-7 rounded-10 bg-raw-cabbage-50 opacity-24 group-hover:opacity-32" />
-      <span
+    <SimpleTooltip content={tooltip?.content} placement={tooltip?.placement}>
+      <label
         className={classNames(
-          'absolute  top-0 z-1 h-full rounded-12',
-          styles.knob,
+          'group relative flex h-9 w-fit cursor-pointer font-bold typo-callout',
+          className,
```
rebelchris, line 84: I see we accept `className` here and below `props.className` can you check which one is wrong?

packages/shared/src/components/fields/CustomSwitch.tsx
```diff
@@ -68,60 +77,69 @@ export function CustomSwitch({
...
+        />
+        <ContentContainer
+          ref={leftRef}
+          className={checked && 'text-text-tertiary'}
+        >
+          {typeof leftContent === 'string'
+            ? leftContent
+            : React.cloneElement(leftContent as ReactElement, {
+                className: classNames(
+                  (leftContent as ReactElement).props.className,
```
rebelchris, line 105: Ah maybe it's on a specific element :)

packages/shared/src/components/fields/CustomSwitch.tsx
rebelchris, line 126:
```suggestion
            disabled ? 'bg-surface-disabled' : 'bg-accent-cabbage-bolder',
```
We should avoid using raw colors at all cost :)

packages/shared/src/components/fields/RadioItem.tsx
```diff
@@ -62,7 +62,9 @@ export function RadioItem<T extends string>({
           <span
             className={classNames(
               'flex h-full w-full rounded-full border-2 ',
-              !disabled && 'border-text-tertiary',
+              !disabled
+                ? 'border-text-tertiary'
+                : '!border-surface-disabled !bg-surface-disabled',
```
rebelchris, line 67: Since you do ternary on the border you don't need important on the second one right?

packages/shared/src/hooks/useFeedLayout.ts
```diff
@@ -125,11 +128,14 @@ export const useFeedLayout = ({
     shouldUseCommentFeedLayout,
   });
 
+  const spaciness = isListMode || !isLaptop ? 'eco' : settingsSpaciness;
```
rebelchris, line 131: Here we should use the `Spaciness` enum mabye instead of `eco` (if it exists)

rebelchris: I'm actually wondering if it's the right place here.
It might be confusing that we override this somewhere down the line.

EG someone might use the spaciness from the original side and expect it to work.
Now we need to tell people to only use this exported spaciness.

(*thinking maybe something like squad directory for instance uses it as well)

rebelchris (changes requested):
Nice, have a minor concern about clarity for the spaciness override as it might not be super clear at this point?

packages/shared/src/components/Settings.tsx
nensidosari: I got this copy from Sab and just adjusted for mobile/tablet and layout with my thoughts based on this, but will confirm with him.
Didn't notice any flickering, just wanted to split the disabled info since it could be for different reasons

packages/shared/src/components/Settings.tsx
nensidosari: Just adjusted this copy from the other copy given, let me confirm.
Checking here: https://dailydotdev.slack.com/archives/C02AA18RD1D/p1717655445750759

packages/shared/src/components/fields/CustomSwitch.tsx
nensidosari: yup, this was already there and just didn't think of changing, will try to keep this in mind for next things 🙌

packages/shared/src/hooks/useFeedLayout.ts
nensidosari: I think it just extends the issue from the list mode to list mode and spaciness, since they both make sense on this hook somehow thought this could be the best place. 
I've searched already and I've changed all places where it was used from settings, the only place where it's used raw is in Settings.tsx, SettingsContext.tsx and in this analytics hook useLogSharedProps. 

I thought of doing this change in settingsContext or settings, but don't think it is the right place, as that gives us raw value from db, and we don't really want to change this value in db since it makes no sense, we just don't want it to be considered on list view, mobile and tablet.

I do definitely agree though that it is confusing if we need to use this variable somewhere else, just like with insane mode, but I'm not really thinking a better simpler solution at the moment 🤷‍♀️ .

Any ideas?

nensidosari: Spaciness enum does not exist, but can create

packages/shared/src/components/Settings.tsx
nensidosari: ok so will totally remove this for mobile and tablet as confirmed with Sab

packages/shared/src/components/Settings.tsx
nensidosari: ok so will totally remove this for mobile and tablet as confirmed with Sab, this way we won't have the function and the copy is already confirmed for density in list view

packages/shared/src/hooks/useFeedLayout.ts
rebelchris: @capJavert // @sshanzel how do you guys feel about it?
Main concern is new people/someone looking at it 2 months from now and just using the raw context directly.

Personally feel maybe the SettingsContext make more sense? (Just so there's only 1 source of truth for what's exported?)

nensidosari: But then in SettingsContext I believe we would need to export a separate value, not the one stored in db, which leaves us with a similar issue of actually the new joiner not sure which value to use since in settings the raw value is being used and other places this modified one.

Or, would we want to actually change the value if user moves to list mode? It's actually not that bad to think about now that we're just removing the options for mobile and tablet (that was the thing that was more troublesome for me to change the value, as u wouldn't want to change the value if user just resizes screen.)

nensidosari: Cool then, I'm moving on with actually changing the value when user goes for list mode since we might be good with that now. @capJavert @sshanzel if u have any better idea, we can definitely evaluate again 🙌

packages/shared/src/components/fields/RadioItem.tsx
nensidosari: That's because there are some injected styles in the bottom that this is overriding, I'm not even sure the border-text-tertiary is even applied

nensidosari: removing it

packages/shared/src/contexts/SettingsContext.tsx
```diff
@@ -192,7 +192,11 @@ export const SettingsContextProvider = ({
       setSpaciness: (density) =>
         setSettings({ ...settings, spaciness: density }),
       toggleInsaneMode: () =>
-        setSettings({ ...settings, insaneMode: !settings.insaneMode }),
+        setSettings({
+          ...settings,
+          insaneMode: !settings.insaneMode,
+          spaciness: !settings.insaneMode ? 'eco' : settings.spaciness,
```
rebelchris, line 198: How about existing users on list mode now?

rebelchris (approved):
We'll need to check existing users.
Mainly raising because you said it makes a impact for some rendering right? (or do we tackle it somehow already, then it's fine)

packages/shared/src/contexts/SettingsContext.tsx
nensidosari: nah we don't tackle it for existing users unless they switch. As discussed in chat, let's not do this change first just disable the options for users when in list mode assuming the issues have been a one off. And then we can consider adding this after if users complain again.

PR #3216, opened 2024-06-13
feat: add reading streaks to devcard

nensidosari:
## Changes
- adding longest streak to the devcard

- related PR https://github.com/dailydotdev/daily-api/pull/1986

Screenshot 2024-06-13 at 6 11 13 AM
Screenshot 2024-06-13 at 6 11 05 AM

Still not 100% sure if we need to do sth else for image download, checking

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done

### Preview domain
https://mi-366-devcard-streaks.preview.app.daily.dev


packages/shared/src/components/profile/devcard/DevCardStats.tsx
```diff
@@ -28,6 +30,12 @@ export function DevCardStats({
         iconClassName="text-raw-onion-40"
         Icon={ReputationIcon}
       />
+      <DevCardStatsSection
+        amount={longestStreak}
+        label="Longest streak"
+        iconClassName="p-[0.19rem]"
```
nensidosari, line 36: This is added because this icon is different from others, while others do seem to have some padding incorporated in the SVG this one doesn't making it bigger in the same sizes.
I tried changing the icon, but that would mean changing a lot into where this icon is used as well, like the streaks button on header, the modals, the completed streak on the popup, and still the sizes wouldn't fit the same as they do now in the current usage, so decided to just move forward for now.

It's an issue though I've seen sometime with icons we have where some are built differently with this padding or not, and in designs this padding is sometimes considered and sometimes not.

rebelchris: It's better to check with design team, generally speaking all the outline of icons should be the same size.
It happened before where they can easily fix this.

packages/shared/src/components/profile/devcard/DevCardStats.tsx
```diff
@@ -1,24 +1,26 @@
...
   user,
   className,
   articlesRead,
+  longestStreak,
 }: DevCardStatsProps): ReactElement {
   return (
     <span
       className={classNames(
-        'flex w-full flex-row gap-3 rounded-16 bg-raw-pepper-90 p-4 shadow-2',
+        'flex w-full flex-row gap-3 rounded-16 bg-raw-pepper-90 px-4 py-2 shadow-2',
```
rebelchris, line 23: Just checking this should be pepper always?

packages/shared/src/components/profile/devcard/DevCardStatsSection.tsx
```diff
@@ -17,12 +17,14 @@ export function DevCardStatsSection({
...
 }: StatsSectionProps): ReactElement {
   return (
-    <span className="flex flex-row items-center gap-1">
+    <span className="flex flex-col">
       <strong>
         <h2 className="text-white typo-title3">{largeNumberFormat(amount)}</h2>
       </strong>
-      <Icon size={IconSize.XSmall} secondary className={iconClassName} />
-      <span className="text-raw-salt-90 typo-caption2">{label}</span>
+      <span className="flex items-center text-raw-salt-90 typo-caption2">
```
rebelchris, line 24: This should be raw salt color always?

rebelchris (approved):
Not sure, but we didn't need it on API level for the image export for the automation? (Can't fully recall)

packages/shared/src/components/profile/devcard/DevCardStats.tsx
nensidosari: They can fix it true, we can even fix it without them if we export the svg they are using now for this, but that would mean we need to change all other usages of it and make it align.

packages/shared/src/components/profile/devcard/DevCardStats.tsx
nensidosari: I believe yes, their background doesn't changed based on the theme

packages/shared/src/components/profile/devcard/DevCardStatsSection.tsx
nensidosari: I'm guessing still yes since it's background doesn't change based on the theme.

sshanzel:
> @rebelchris can't recall properly either, we tried at that time to do it in another way which required API but the existing one that was there and that we're still using I'm thinking no - have checked on API side all possible places but couldn't find anything I missed. Will check with @sshanzel once, maybe he remembers

We should be good afair, since the image downloading will just call our FE to generate what should be captured.

PR #3219, opened 2024-06-13
fix: rendering of shared post date for exp

rebelchris:
- We need to render the share date not the actual post date

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

AS-372 #done

### Preview domain
https://as-372-fix-shared-post-date.preview.app.daily.dev

### DIFF
--- a/packages/shared/src/components/cards/SharePostCard.tsx
+++ b/packages/shared/src/components/cards/SharePostCard.tsx
@@ -57,6 +57,8 @@ export const SharePostCard = forwardRef(function SharePostCard(
   if (improvedSharedPostCard) {
     // eslint-disable-next-line no-param-reassign
     post.image = post.sharedPost.image;
+    // eslint-disable-next-line no-param-reassign
+    post.sharedPost.createdAt = post.createdAt;
   }
 
   return (


packages/shared/src/components/cards/SharePostCard.tsx
```diff
@@ -57,6 +57,8 @@ export const SharePostCard = forwardRef(function SharePostCard(
   if (improvedSharedPostCard) {
     // eslint-disable-next-line no-param-reassign
     post.image = post.sharedPost.image;
+    // eslint-disable-next-line no-param-reassign
+    post.sharedPost.createdAt = post.createdAt;
```
capJavert, line 61: I know you did not do the change but this can have a bunch of edge cases for any component that is using `post` before this component mounts. Better way would be to instantiate new shallow copy and assign the image and createdAt there and then pass it to child components.

rebelchris: I think the main issue is it's used a lot down the line in components making it harder to switch.
But can have a quick peek if it's easy to avoid edge-cases

PR #3221, opened 2024-06-13
fix: pass props for shared post card

rebelchris:
### Describe what this PR does
- @capJavert Added a temporary workaround for this override.
- It's not ideal as the full implementation would take many, many changes but think for now it's safer atleast, let me know what you think.

## Events

Did you introduce any new tracking events?
Don't forget to update the [Analytics Taxonomy sheet](https://docs.google.com/spreadsheets/d/18Lv7zXges9QfVX5VYL1a-Hyl0e1sQ3sLr0OK8YZWKXI/edit#gid=0)

Log the new/changed events below:

| Type   | event_name  | value |
|--------|-------------|-------|
| Change/New | event name  | extra: { ... } |

### On those affected packages:
- [ ] Have you done sanity checks in the webapp?
- [ ] Have you done sanity checks in the extension?
- [ ] Does this not break anything in companion?

### Did you test the modified components media queries?
- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

#### Did you test on actual mobile devices?
- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done

### Preview domain
https://fix-pass-props.preview.app.daily.dev

PR #3224, opened 2024-06-13
fix: remove a param-reassign

omBratteng:
Removed in #3221 but got reverted in #3223.
This one we can keep, instead of re-assigning the variable when it is only used inside a ternary

### Preview domain
https://ombratteng-patch-1.preview.app.daily.dev

### DIFF
--- a/packages/shared/src/components/cards/SharePostCard.tsx
+++ b/packages/shared/src/components/cards/SharePostCard.tsx
@@ -57,8 +57,6 @@ export const SharePostCard = forwardRef(function SharePostCard(
   if (improvedSharedPostCard) {
     // eslint-disable-next-line no-param-reassign
     post.image = post.sharedPost.image;
-    // eslint-disable-next-line no-param-reassign
-    post.sharedPost.createdAt = post.createdAt;
   }
 
   return (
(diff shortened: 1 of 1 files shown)


packages/shared/src/components/cards/SharePostCard.tsx
```diff
@@ -93,7 +91,7 @@ export const SharePostCard = forwardRef(function SharePostCard(
             <CardSpace />
             <PostTags tags={post.sharedPost.tags} />
             <PostMetadata
-              createdAt={post.sharedPost.createdAt}
+              createdAt={post.createdAt}
```
omBratteng, line 94: This is inside the `improvedSharedPostCard` ternary, so we don't need to re-assign the variable.
https://github.com/dailydotdev/apps/pull/3221#discussion_r1637934759

capJavert:
We can merge it in https://github.com/dailydotdev/apps/pull/3225 since it also remove mutating post in place in the first place

packages/shared/src/components/cards/SharePostCard.tsx
capJavert: I am not even sure why that was there but it was fix from this morning to use sharedPost.createdAt https://github.com/dailydotdev/apps/commit/b765437c961094d5c5c1625400e6d5f96de113d9 @rebelchris just verify this is ok

capJavert (approved):
Just wait for Chris to confirm.

packages/shared/src/components/cards/SharePostCard.tsx
omBratteng: There was an issue that the cards in the experiment showed the created date of the shared post, and not of the share itself

rebelchris: Yeah exactly :)

rebelchris (approved):

PR #3242, opened 2024-06-19
Mi 384 bookmark provider

nensidosari:
## Changes
- added just bookmarked context in order to keep what bookmarks are posted at the moment
- show bookmark provider ui when a bookmarked post come through the api on my feed
- added behind feature value
- changed header and container for all cards based on designs

- add bookmark provider design for all types of cards

Screenshot 2024-06-19 at 5 17 09 PM
Screenshot 2024-06-19 at 5 17 43 PM

## Events

- [x] MobileL (420px)
- [x] Tablet (656px)
- [x] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

MI-384 #done

[mi-384-bookmark-provider.preview.app.daily.dev](http://mi-384-bookmark-provider.preview.app.daily.dev/)

### Preview domain
https://mi-384-bookmark-provider.preview.app.daily.dev


packages/shared/src/components/cards/FeedItemContainer.tsx
```diff
@@ -24,14 +25,19 @@ interface FeedItemContainerProps {
...
-  { flagProps, children, domProps }: FeedItemContainerProps,
+  { flagProps, children, domProps, bookmarked, postId }: FeedItemContainerProps,
   ref?: Ref<HTMLElement>,
 ): ReactElement {
+  const { highlightBookmarkedPost } = useBookmarkProvider({
+    bookmarked,
+    postId,
+  });
   const { listMode, pinnedAt, trending } = flagProps;
-  const Component = listMode ? ListCard : Card;
```
nensidosari, line 34: this is never used in list mode

sshanzel (commented):
Looks great on facilitating the bookmark provider. Though imho, I think the context is a bit overkill. I think a simple state similar to what we did on feedback loop when the user just upvoted would have been sufficient. When the `post.bookmarked` is `true` and the boolean state is `false` then we show this UI (the state only turns true when you click the bookmark button, which then we will not show this UI). However, I don't have any strong opinion here as there are pros and cons, just stating some existing implementations.

nensidosari:
@sshanzel agree, 1 min

packages/shared/src/hooks/useBookmarkProvider.ts
```diff
@@ -0,0 +1,43 @@
...
+}: UseBookmarkProviderProps): UseBookmarkProviderReturn => {
+  const [isBookmarked] = useState<boolean>(bookmarked);
+  const [justBookmarked, setJustBookmarked] = useState<boolean>(false);
+  const { feedName } = useActiveFeedNameContext();
+
+  useEffect(() => {
+    if (bookmarked !== isBookmarked) {
+      setJustBookmarked(bookmarked);
+    }
+  }, [bookmarked, justBookmarked, setJustBookmarked, isBookmarked]);
```
sshanzel, line 26: We tend to avoid use effects whenever possible, however, this has made the process much simpler by not injecting the state update on bookmark click. Would love to hear more thoughts here.

packages/shared/src/hooks/useBookmarkProvider.ts
```diff
@@ -0,0 +1,43 @@
...
+}
+
+interface UseBookmarkProviderProps {
+  bookmarked?: boolean;
+}
+
+const useBookmarkProvider = ({
+  bookmarked = false,
+}: UseBookmarkProviderProps): UseBookmarkProviderReturn => {
+  const [isBookmarked] = useState<boolean>(bookmarked);
```
sshanzel, line 18: We can probably turn this into a `useRef` as we don't expect state updates and we just want to hold the value here.

sshanzel (commented):
Looks great overall! Just had some comments that are nothing major, but I like how simple it became.

packages/shared/src/hooks/useBookmarkProvider.ts
nensidosari: I don't think we actually need useEffect and the state justBookmarked at all, is useless, we can just compare the values directly. 
was a bit tired yesterday not sure what was I thinking 😓

sshanzel (approved):
Nice! No use effects nor use states. Looks great on my end! 🥳

packages/shared/src/components/cards/PostCardHeader.tsx
```diff
@@ -33,31 +40,45 @@ export const PostCardHeader = ({
   source,
   postLink,
   openNewTab,
+  showFeedback,
 }: CardHeaderProps): ReactElement => {
   const isFeedPreview = useFeedPreviewMode();
+  const { highlightBookmarkedPost } = useBookmarkProvider({
+    bookmarked: post.bookmarked && !showFeedback,
```
rebelchris, line 47: What is the showFeedback based on?

rebelchris (approved):
Nice, like to solution just wondering what the showFeedback is for?

packages/shared/src/components/cards/PostCardHeader.tsx
nensidosari: Is about the feedback loop, so if user reads any article that was bookmarked and feeback loop shows, these colors to now show as they look weird. Once feedback loop is removed these get back there if the post is still there

PR #3245, opened 2024-06-20
fix: cleanup `search_google` experiment

omBratteng:
AS-392

<table>
<tr>
<td>Before
<td>After
<tr>
<td>
<td>
</table>

### Preview domain
https://as-392-search-google-cleanup.preview.app.daily.dev

### DIFF
--- a/packages/shared/src/components/search/SearchPanel/SearchPanel.tsx
+++ b/packages/shared/src/components/search/SearchPanel/SearchPanel.tsx
@@ -22,7 +22,7 @@ import { SearchPanelPostSuggestions } from './SearchPanelPostSuggestions';
 import SettingsContext from '../../../contexts/SettingsContext';
 import { useConditionalFeature, useEventListener } from '../../../hooks';
 import { defaultSearchProvider, providerToLabelTextMap } from './common';
-import { ArrowKeyEnum, checkIsExtension } from '../../../lib/func';
+import { ArrowKeyEnum, isExtension } from '../../../lib/func';
 import { ArrowIcon } from '../../icons';
 import { useSearchProvider } from '../../../hooks/search';
 import { SearchPanelCustomAction } from './SearchPanelCustomAction';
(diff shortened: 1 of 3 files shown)


packages/shared/src/components/search/SearchPanel/SearchPanel.tsx
```diff
@@ -203,7 +198,7 @@ export const SearchPanel = ({ className }: SearchPanelProps): ReactElement => {
               <div className="flex flex-1 flex-col">
                 <SearchPanelAction provider={SearchProviderEnum.Posts} />
                 <SearchPanelAction provider={SearchProviderEnum.Chat} />
-                {isGoogleSearchEnabled && (
+                {checkIsExtension() && (
```
rebelchris, line 201: Should we consider something like this? (Not sure but would like to avoid re-render because func)
```
const isExtension = checkIsExtension();
```

rebelchris (approved):
Non blocking comment

packages/shared/src/components/search/SearchPanel/SearchPanel.tsx
omBratteng: Add this?
```ts
const isExtension = !!process.env.TARGET_BROWSER;
```
As `checkIsExtension` is this
```ts
export const checkIsExtension = (): boolean => !!process.env.TARGET_BROWSER;
```

rebelchris: Also works :) 
Not really sure why it's a func tbh

Just noticed we do the first one quite often already.

omBratteng: Done. Going to test that it works in ext with this

omBratteng: Works.

PR #3264, opened 2024-06-24
fix: comment input header bg

sshanzel:
## Changes
- After further investigation, the issue seems to be not the border but the bg.
- After the mobile UX project, we made the bg to use the `default` but that should only be on the popup which is available on mobile only.
- For tablet and up, it should be `float`.
- For reference: https://www.figma.com/design/nmfWPS7x3kzLUvYMkBx2kW/daily.dev---Dev-Mode?node-id=6718-24966&m=dev
- Also fixed the minimum height issue when toggling left and right. 

Preview from Figma:
Screenshot 2024-06-24 at 9 57 07 PM

Fix recording:
https://github.com/dailydotdev/apps/assets/13744167/749e4782-9692-4820-a1bd-22c1fe675828

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

MI-395 #done

### DIFF
--- a/packages/shared/src/components/fields/MarkdownInput/index.tsx
+++ b/packages/shared/src/components/fields/MarkdownInput/index.tsx
@@ -192,7 +192,7 @@ function MarkdownInput(
             tabListProps={{ className: { indicator: '!w-6' } }}
           >
             <Tab label={CommentTab.Write}>{children}</Tab>
-            <Tab label={CommentTab.Preview} className="p-4">
+            <Tab label={CommentTab.Preview} className="min-h-[11.125rem] p-4">
               <MarkdownPreview
                 input={input}
                 sourceId={sourceId}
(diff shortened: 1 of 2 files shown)


packages/shared/src/components/fields/MarkdownInput/index.tsx
```diff
@@ -192,7 +192,7 @@ function MarkdownInput(
             tabListProps={{ className: { indicator: '!w-6' } }}
           >
             <Tab label={CommentTab.Write}>{children}</Tab>
-            <Tab label={CommentTab.Preview} className="p-4">
+            <Tab label={CommentTab.Preview} className="min-h-[11.125rem] p-4">
```
sshanzel, line 195: The actual box already has its natural `min-h` value so this won't break any existing implementation.

nensidosari (approved):

rebelchris (approved):

PR #3269, opened 2024-06-25
feat: add notifications on footer on mobile

omBratteng:
# Changes

Moves notifications to bottom

image

AS-398

### Preview domain
https://as-398-notification-micro.preview.app.daily.dev

### DIFF
--- a/packages/shared/src/components/profile/Header.tsx
+++ b/packages/shared/src/components/profile/Header.tsx
@@ -9,9 +9,8 @@ import { largeNumberFormat, ReferralCampaignKey } from '../../lib';
 import { ProfileSettingsMenu } from './ProfileSettingsMenu';
 import { RootPortal } from '../tooltips/Portal';
 import { GoBackButton } from '../post/GoBackHeaderMobile';
-import { ViewSize, useViewSize } from '../../hooks';
+import { ViewSize, useConditionalFeature, useViewSize } from '../../hooks';
 import { feature } from '../../lib/featureManagement';
-import { useFeature } from '../GrowthBookProvider';
 
 export interface HeaderProps {
   user: PublicProfile;
@@ -35,7 +34,10 @@ export function Header({
     logObject: () => ({ event_name: 'share profile', target_id: user.id }),
   });
   const isMobile = useViewSize(ViewSize.MobileL);
-  const notificationsNavBar = useFeature(feature.notificationsNavBar);
+  const { value: notificationsNavBar } = useConditionalFeature({
+    feature: feature.notificationsNavBar,
+    shouldEvaluate: isMobile,
+  });
   const [isMenuOpen, setIsMenuOpen] = useState(false);
 
   return (
--- a/packages/shared/src/components/profile/Header.tsx
+++ b/packages/shared/src/components/profile/Header.tsx
@@ -8,6 +8,10 @@ import { ProfileImageSize, ProfilePicture } from '../ProfilePicture';
 import { largeNumberFormat, ReferralCampaignKey } from '../../lib';
 import { ProfileSettingsMenu } from './ProfileSettingsMenu';
 import { RootPortal } from '../tooltips/Portal';
+import { GoBackButton } from '../post/GoBackHeaderMobile';
+import { ViewSize, useViewSize } from '../../hooks';
+import { feature } from '../../lib/featureManagement';
+import { useFeature } from '../GrowthBookProvider';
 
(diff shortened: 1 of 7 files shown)


packages/webapp/components/footer/FooterNavBarTabs.tsx
```diff
@@ -82,12 +88,45 @@ const Tab = ({ tab, isActive }: TabProps) => {
...
+      {!!unreadCount && (
+        <Bubble
+          className={classNames('-right-1.5 -top-1.5 cursor-pointer px-1')}
+        >
+          {getUnreadText(unreadCount)}
+        </Bubble>
+      )}
+    </div>
+  );
+};
```
omBratteng, line 106: I wanted to re-use the `NotificationsBell` component, but it adds a tooltip and an extra button that doesn't fit in the bottom navbar. So for experiment sake, I thought of just doing a local component here.

packages/webapp/components/footer/FooterNavBarTabs.tsx
```diff
@@ -82,12 +88,45 @@ const Tab = ({ tab, isActive }: TabProps) => {
...
+      mobileUxTabs.splice(-1, 0, {
+        requiresLogin: true,
+        path: '/notifications',
+        title: 'Notifications',
+        icon: (active: boolean) => <Notifications active={active} />,
+      });
+    }
+
+    return mobileUxTabs;
+  }, [notificationsNavBar]);
```
omBratteng, line 125: Had to memoize it, unless you'd end up with lots of notification buttons
image

packages/shared/src/components/profile/Header.tsx
```diff
@@ -30,30 +34,37 @@ export function Header({
     cid: ReferralCampaignKey.ShareProfile,
     logObject: () => ({ event_name: 'share profile', target_id: user.id }),
   });
+  const isMobile = useViewSize(ViewSize.MobileL);
+  const notificationsNavBar = useFeature(feature.notificationsNavBar);
```
rebelchris, line 38: We probably want to limit enrollment to only mobile users, maybe use `useConditionalFeature` here to be safe :)

packages/shared/src/hooks/useActiveNav.tsx
```diff
@@ -16,6 +18,7 @@ export interface UseActiveNav {
 export default function useActiveNav(activeFeed: AllFeedPages): UseActiveNav {
   const router = useRouter();
   const isLaptop = useViewSize(ViewSize.Laptop);
+  const notificationsNavBar = useFeature(feature.notificationsNavBar);
```
rebelchris, line 21: Here we probably also want to hard enroll only to mobile?
(also note the current one is `!laptop` which might be wrong as we want hard `isMobile`)

packages/webapp/components/footer/FooterNavBarTabs.tsx
```diff
@@ -82,12 +88,45 @@ const Tab = ({ tab, isActive }: TabProps) => {
...
+        </Bubble>
+      )}
+    </div>
+  );
+};
+
 export function FooterNavBarTabs({
   activeTab,
 }: FooterNavBarContainerProps): ReactElement {
+  const notificationsNavBar = useFeature(feature.notificationsNavBar);
```
rebelchris, line 111: Can you check if this component only get's rendered on mobile?
Else we'll need to do enrollment check here as well.

rebelchris (changes requested):
Looks good, just some enrollment check we need to add.

packages/webapp/components/footer/FooterNavBarTabs.tsx
omBratteng: FooterNavBarTabs is rendered by FooterNavBar if `showNav` is set to true, which is set to true in FooterNavBarLayout if isMobile.

So, only on mobile.

packages/shared/src/components/profile/Header.tsx
omBratteng: Used useConditionalFeature

packages/shared/src/hooks/useActiveNav.tsx
omBratteng: Added useConditionalFeature and an extra check for !isMobile where it appends the notifications to array.

nensidosari (approved):
Looks good

rebelchris (approved):

PR #3270, opened 2024-06-26
chore: react package upgrade to v18

sshanzel:
## Changes
- Most of the changes were explained within the RFC. Those that were not, encountered on the actual upgrade, should be mentioned below.
- When a `useEffect` returns a function, it should be consistent across all returns.
- The callback for empty feed is now inline rather than having a `useEffect`.
- Fixed some minor type issues.
  - Notable issue is that, usages of `FunctionComponent` or `FC` would now require to explicitly define the`children` property. 
- There have been numerous errors as well about:
  - Missing key on mapped render.
  - Some props that have been passed that are not necessary.
- Migrated some usages of icon to be inline with how we normally do it.
- The `act` function should now be coming from `react` package.

## Events

- [x] MobileL (420px)
- [x] Tablet (656px)
- [x] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

WT-{number} #done

### Preview domain
https://mi-391.preview.app.daily.dev


packages/shared/src/components/layout/MainLayoutHeader.tsx
```diff
@@ -61,12 +59,7 @@ function MainLayoutHeader({
             {additionalButtons}
             {isLoggedIn ? (
               <>
-                <LinkWithTooltip
```
sshanzel, line 64: The `NotificationBell` itself is already wrapped with this tooltip component.

packages/shared/src/components/Feed.tsx
```diff
@@ -241,12 +242,6 @@ export default function Feed<T>({
     selectedPostIndex,
   } = usePostModalNavigation(items, fetchPage, updatePost, canFetchMore);
 
-  useEffect(() => {
```
sshanzel, line 244: One less useEffect. Makes test easier as well.

packages/shared/src/components/icons/TimeZone/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from './DaytimeIcon';
+export * from './NighttimeIcon';
```
sshanzel, line 2: These files were just moved and turned to how we normally use SVGs. Was getting an error due to some existing implementations utilizing the icon props.

packages/shared/src/components/profile/ExperienceLevelDropdown.tsx
```diff
@@ -87,6 +87,7 @@ const ExperienceLevelDropdown = ({
           className="hidden"
           name={name}
           value={Object.keys(UserExperienceLevel)[selectedIndex]}
+          readOnly
```
sshanzel, line 90: Inputs without `onChange` handler should have `readOnly` props.

packages/shared/src/components/squads/SquadHeaderMenu.tsx
```diff
@@ -146,7 +144,6 @@ export default function SquadHeaderMenu({
       className="menu-primary"
       animation="fade"
       options={items}
-      isOpen={isOpen}
```
sshanzel, line 149: Unnecessary props. This is not used from the component itself.

packages/shared/src/components/squads/SquadsDirectoryHeader.tsx
```diff
@@ -27,9 +27,8 @@ export const SquadsDirectoryHeader = ({
     >
       <div className="relative z-1 flex flex-col items-center">
         <SourceBetaIcon
-          width="auto"
```
sshanzel, line 30: On SVGs, `width="auto"` is not allowed. However, adding a `w-auto` class is rendering a different result. With the `w-full`, it goes back to normal.

packages/shared/src/hooks/useFeed.ts
```diff
@@ -112,21 +113,33 @@ export default function useFeed<T>(
...
-      }),
+      });
+
+      if (
+        !feedQuery?.data?.pages[0]?.page.edges.length &&
+        !res?.page?.edges?.length &&
+        onEmptyFeed
+      ) {
+        onEmptyFeed();
+      }
```
sshanzel, line 139: Avoids usage of `useEffect` and makes test easier to manage.

packages/webapp/components/FooterNavBar.tsx
```diff
@@ -83,19 +83,20 @@ export default function FooterNavBar({
...
-        className={classNames(
-          'grid w-full grid-flow-col items-center justify-between rounded-16 px-3',
-          !post && activeClasses,
-          !showNav && 'hidden',
-          !post && 'border-t border-border-subtlest-tertiary',
-        )}
-      >
-        <FooterNavBarTabs activeTab={activeTab} />
-      </Flipper>
+      {showNav && (
```
sshanzel, line 86: We don't need to render the footer navbar on tablet and desktop which is evaluated at `showNav`.

packages/webapp/pages/account/profile.tsx
```diff
@@ -135,7 +135,7 @@ const AccountProfilePage = (): ReactElement => {
     </span>
   );
 
-  const RenderForm = () => (
```
sshanzel, line 138: Producing an infinite loop. Either we probably wrap it with `useMemo` or the changes made here.

packages/shared/src/lib/dateFormat.spec.ts
```diff
@@ -163,7 +163,7 @@ describe('getTodayTz', () => {
 
   it('should return same date as today if second argument is not supplied', () => {
     const dateNow = new Date();
-    const date = getTodayTz('UTC');
```
sshanzel, line 166: Would fail at a certain time on local which could happen at remote test env too.

packages/shared/src/components/modals/report/ReportComentModal.spec.tsx
```diff
@@ -125,14 +125,16 @@ it('submit the report with text', async () => {
...
+    await expect(otherRadio).toBeChecked();
 
-  const input = screen.getByTestId('report_comment') as HTMLTextAreaElement;
-  fireEvent.change(input, { target: { value: 'test note' } });
-  input.dispatchEvent(new Event('input', { bubbles: true }));
+    const input = screen.getByTestId('report_comment') as HTMLTextAreaElement;
+    fireEvent.change(input, { target: { value: 'test note' } });
+    input.dispatchEvent(new Event('input', { bubbles: true }));
+    expect(input.value).toBe('test note');
+  });
```
sshanzel, line 136: Without the `act`, I get this error below.

Screenshot 2024-06-29 at 11 13 51 AM

packages/webapp/pages/squads/[handle]/index.tsx
```diff
@@ -71,6 +70,14 @@ const SquadEmptyScreen = dynamic(
     ),
 );
 
+const SquadLoading = dynamic(
+  () =>
+    import(
+      /* webpackChunkName: "squadLoading" */ '@dailydotdev/shared/src/components/errors/SquadLoading'
+    ),
+  { ssr: false },
+);
```
Comment (sshanzel) [inline packages/webapp/pages/squads/[handle]/index.tsx:79]: There are certain components that we don't need to be loaded from the server side. Especially those components that are dynamic and change on load. One example is this, the layout changes based on the screen size which React throws an error if we don't disable it here.

packages/shared/src/hooks/profile/useProfile.ts
```diff
@@ -12,14 +13,11 @@ export function useProfile(initialUser?: PublicProfile): PublicProfile {
     generateQueryKey(RequestKey.Profile, initialUser),
     () => getProfile(initialUser?.id),
     {
-      placeholderData: initialUser,
```
sshanzel, line 15: For some reason using `placeholderData` produces an infinite loop. Used `initialData` instead and to imitate placeholder data, when the `user` value is null, we will fallback to the `initialUser`.

Screenshot 2024-07-01 at 7 59 44 PM

packages/webapp/components/layouts/MainFeedPage.tsx
```diff
@@ -7,14 +7,21 @@ import React, {
...
 import { getFeedName as getFeedNameLib } from '@dailydotdev/shared/src/lib/feed';
+import dynamic from 'next/dynamic';
 import { getLayout } from './FeedLayout';
 
+const MainFeedLayout = dynamic(
+  () =>
+    import(
+      /* webpackChunkName: "goBackHeaderMobile" */ '@dailydotdev/shared/src/components/MainFeedLayout'
+    ),
+  { ssr: false },
```
sshanzel, line 22: Producing an error about a server-client mismatch. Also, I think no point in doing SSR here since the whole content of this file is purely based on dynamic content.

Screenshot 2024-07-01 at 8 33 34 PM

sshanzel:
@dailydotdev/web-team should be ready for review. I've tested both webapp and extension. I've also tested the mobile and tablet sizes.

However, there is one issue I have encountered. Since the upgrade (without any other changes), the context menu has started to behave abnormally on close. Either we remove the animation on exit (close) of the context menu, or we should probably replace it completely like we have always planned (which I highly doubt we should, this would immensely increase the time we need for this upgrade).

https://github.com/dailydotdev/apps/assets/13744167/db1055ca-d15c-4d8e-89e5-11ea2f54e7a1

omBratteng:
YOLO and approve without looking at code?

sshanzel:
> YOLO and approve without looking at code?

Let's do it. LOL. Nensi and Chris would choke us to death 😂

packages/extension/src/companion/index.tsx
```diff
@@ -9,18 +9,29 @@ import { setOnError } from '@dailydotdev/shared/src/components/ProfilePicture';
 import { getCompanionWrapper } from '@dailydotdev/shared/src/lib/extension';
 import App, { CompanionData } from './App';
 
+let root: ReturnType<typeof createRoot>;
```
rebelchris, line 12: Is there no other way to solve this then with let?
EG just declaring as root const?

packages/shared/src/components/BookmarkFeedLayout.tsx
```diff
@@ -100,7 +100,7 @@ export default function BookmarkFeedLayout({
       </FeedPageHeader>
       <CustomFeedHeader
         className={classNames(
-          'mb-6 flex',
+          'mb-6',
```
rebelchris, line 103: Why was this related? or just a bug you caught?

packages/shared/src/components/Feed.tsx
rebelchris: Was no longer used anywhere?
(missing context why it was needed/no longer needed)

packages/shared/src/components/MainLayout.tsx
```diff
@@ -34,10 +35,17 @@ import {
   useActiveFeedNameContext,
 } from '../contexts';
 import { useFeedLayout, useViewSize, ViewSize } from '../hooks';
-import { GoBackHeaderMobile } from './post/GoBackHeaderMobile';
 import { BootPopups } from './modals/BootPopups';
 import { useFeedName } from '../hooks/feed/useFeedName';
 
+const GoBackHeaderMobile = dynamic(
```
rebelchris, line 41: I think you can revert this with the latest change again as it's always rendered?

packages/shared/src/components/feeds/HorizontalFeed.tsx
```diff
@@ -33,7 +33,7 @@ export default function HorizontalFeed<T>({
   useEffect(() => {
     const element = feedContainerRef;
     if (!element) {
-      return null;
+      return () => {};
```
rebelchris, line 36: But this is the cleanup effect right?
Shouldn't it be something like `return false` or `return undefined` now? *(CC @capJavert you might know)

rebelchris: I think it might work, but it's a quick/dirty fix.
We probably want to revert this to a opposite solution saying "if (element)" do something instead?

packages/shared/src/components/fields/ContextMenu.tsx
```diff
@@ -48,12 +48,13 @@ export interface MenuItemProps<
...
 interface ContextMenuProps extends Omit<MenuProps, 'children'> {
   options: MenuItemProps[] | ContextMenuDrawerItem[];
-  isOpen: boolean;
+  isOpen?: boolean;
 }
 
 export default function ContextMenu({
   options,
   onHidden,
+  isOpen,
```
rebelchris, line 57: This makes it so you miss `isOpen` in the `PortalMenu` is that ok?

packages/shared/src/hooks/companion/useRawBackgroundRequest.ts
```diff
@@ -19,7 +19,7 @@ export const useRawBackgroundRequest = (
 
   useEffect(() => {
     if (!browser) {
-      return null;
+      return () => {};
```
rebelchris, line 22: Depending on the above answer from Ante, maybe we should revert the eval here

packages/shared/src/hooks/profile/useProfile.ts
rebelchris: Probably good to debug a bit more then?
Feels like a re-render twice related thing?

packages/shared/src/hooks/useEventListener.ts
```diff
@@ -137,7 +137,9 @@ const useEventListener = <
 
   useEffect(() => {
     const eventListener = (...args) => {
-      return handlerRef.current(...args);
+      return () => {
+        handlerRef.current(...args);
+      };
```
rebelchris, line 142: This is now inside a cleanup?

packages/shared/src/hooks/useFeed.ts
rebelchris: Oh I see now :) 
Was this one of the few places?

rebelchris (changes requested):

packages/shared/src/components/feeds/HorizontalFeed.tsx
nensidosari: > We probably want to revert this to a opposite solution saying "if (element)" do something instead

I also think this would be best, and the return function would be consistent on all returns this way as well.

packages/shared/src/components/fields/ContextMenu.tsx
nensidosari: and why chosen to make it optional now?

packages/extension/src/companion/index.tsx
sshanzel: We can't unfortunately. At initial load, the companion wrapper is not there yet for the root to reference. So when the time we need to render, the root is empty.

packages/webapp/pages/account/profile.tsx
capJavert: This seems weird, we should be very careful on why this happens.

packages/shared/src/components/BookmarkFeedLayout.tsx
sshanzel: Oh, sorry forgot to mention. There was some bug tracing to this element and noticed the flex was applied twice so i removed this excess one.

packages/webapp/pages/squads/[handle]/index.tsx
Comment (capJavert) [reply packages/webapp/pages/squads/[handle]/index.tsx:79]: Also weird, it should only be different in loading in client and/or server bundle, would not 100% brush it off before going to prod.

packages/shared/src/components/Feed.tsx
sshanzel: When we return an empty feed, edge case when users tags are too narrow and literally no content is shown.

packages/shared/src/lib/dateFormat.spec.ts
capJavert: Nice catch

packages/shared/src/components/MainLayout.tsx
sshanzel: But it's content is dynamic and not static on build time.

packages/shared/src/hooks/useEventListener.ts
capJavert: Also interested why this change, it should be called since it is a callback that needs to execute for event.

packages/shared/src/components/feeds/HorizontalFeed.tsx
sshanzel: We would still have to return an empty function but on the other side of the condition, which is at the last part of the function

packages/shared/src/components/fields/ContextMenu.tsx
sshanzel: Yes. React was screaming that it is not needed to the other element.

packages/shared/src/hooks/profile/useProfile.ts
```diff
@@ -12,14 +13,11 @@ export function useProfile(initialUser?: PublicProfile): PublicProfile {
...
-      refetchOnWindowFocus: false,
-      // Make sure the logged in user's profile is always up to date
+      ...disabledRefetch,
+      initialData: initialUser,
       enabled: !!initialUser && isSameUser,
     },
   );
 
-  return user;
+  return user ?? initialUser;
```
capJavert, line 22: This might break places where we expect `undefined|null`, just to sanity check since we do have `initialData`

packages/shared/src/hooks/profile/useProfile.ts
sshanzel: I tried making the parameters to the query to be stable, but none worked. Do you maybe have some ideas on where to look further?

packages/shared/src/hooks/useFeed.ts
sshanzel: Yes, should ensure executes only once and avoids the use effect 🥳

packages/shared/src/hooks/useEventListener.ts
sshanzel: This has been inside, was just without curly braces to enclose it. React does not like directly returning values on cleanup

packages/shared/src/components/feeds/HorizontalFeed.tsx
capJavert: This is only linter complaining for `consistent-return` rule, React does not care really, I would go with `return undefined` yes 👍

packages/webapp/package.json
```diff
@@ -28,8 +28,8 @@
     "next": "^12.2.2",
     "next-pwa": "^5.5.4",
     "next-seo": "^5.4.0",
-    "react": "17.0.2",
-    "react-dom": "17.0.2",
```
capJavert, line 32: `next@12.x` is not compatible with React 18 I think, probably works because react is backwards compatible but due to core `react-dom` changes it might break in some weird cases (as you had below), we should maybe leave on `17.x` for webapp until Next upgrade?

capJavert (commented):
Nice work 💪 , one note on React version in Next@12, maybe server side on lazy load and weird rendering issues were due to that? Mainly following comments https://github.com/dailydotdev/apps/pull/3270#discussion_r1661132116 https://github.com/dailydotdev/apps/pull/3270#discussion_r1661113429 https://github.com/dailydotdev/apps/pull/3270#discussion_r1661111657

packages/shared/src/hooks/useEventListener.ts
capJavert: React should not care, was it linter bugging you?

packages/shared/src/components/feeds/HorizontalFeed.tsx
sshanzel: It was actually React that throws the error 😓 

But yeah, after testing to return `undefined` it work normally. Let me update everything quickly 🥳 

Screenshot 2024-07-02 at 3 48 31 AM

packages/shared/src/hooks/useEventListener.ts
sshanzel: Right, I might have mistaken this to be the consistent return above and should not really be a part of it. Let me verify.

packages/webapp/pages/squads/[handle]/index.tsx
Comment (sshanzel) [reply packages/webapp/pages/squads/[handle]/index.tsx:79]: Yeah, I was also surprised to see these kinds of issues.

Screenshot 2024-07-02 at 3 54 56 AM

packages/webapp/pages/account/profile.tsx
sshanzel: I think the components within are getting re-mounted since it is a function and gets recreated at every state update, then some `useEffect` or some that do state change executes which loops it back.

This actually fixes the current issue about changing the timezone would reset everything.

packages/webapp/package.json
sshanzel: Interesting. Yeah, we could maybe have avoided some of the weird things introduced here if we did that. I will try it tomorrow and see how it turns out.

packages/shared/src/contexts/PushNotificationContext.tsx
```diff
@@ -120,7 +120,7 @@ export function PushNotificationContextProvider({
 
   useEffect(() => {
     if (!OneSignalCache) {
-      return undefined;
+      return () => {};
```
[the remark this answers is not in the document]
sshanzel: Have updated all of the changes related to this 🥳

packages/webapp/package.json
sshanzel: After testing to install react-dom v17.x and react on 18.x - there seem to be some package issues after clearing up all the modules and then run the local env.

I think we should be safe having updated react-dom. Like you mentioned, React 18 is backwards compatible, and on the NextJS side, in order to use the React 18 features, we will have to add some experimental flags within the config. That means the default setting is without the React 18 features integrated.

Screenshot 2024-07-02 at 2 12 59 PM

omBratteng: I checked the package.json for next.js v12.3.4, and they used react 18. But they also had tests for react 17 for backwards compatibility, so I think we're safe to upgrade to v18.

capJavert: I know 13.x and above has the internal check to only use 18 features when in app router, but was not sure if they backported to 12.x. Luckly we don't plan on staying on 12.x for long so if we don't find anything in testing its good to go. Tnx for checking!

packages/shared/src/components/feeds/HorizontalFeed.tsx
capJavert: This happens if you return anything but a function, if you `return undefined` its like doing just `return` but our linter does not scream.

packages/shared/src/hooks/profile/useProfile.ts
sshanzel: I think we can just return `user` since the object would basically return the `initialData`.

nensidosari (approved):
Looks good on my side! Great work Lee! 🎉

packages/shared/src/hooks/profile/useProfile.ts
capJavert: How to trigger this infinite loop? What are the repro steps?

rebelchris (approved):
Code wise nothing blocking.
Also testing seems ok

packages/shared/src/hooks/profile/useProfile.ts
sshanzel: Just reload the page and it would do infinite loop if the `placeholderData` is used while viewing the page.

capJavert (approved):

PR #3271, opened 2024-06-26
fix: minor fixes for notifications on bottom

omBratteng:
# Changes
- Fixes the copy of the notifications button
- Fixes the post button to be centered
- Fixes multiple copies of notifications button

### Preview domain
https://as-398-fix.preview.app.daily.dev

### DIFF
--- a/packages/webapp/components/FooterNavBar.tsx
+++ b/packages/webapp/components/FooterNavBar.tsx
@@ -88,7 +88,7 @@ export default function FooterNavBar({
         spring="veryGentle"
         element="nav"
         className={classNames(
-          'grid w-full auto-cols-fr grid-flow-col items-center justify-between rounded-16 px-3',
+          'grid w-full auto-cols-fr grid-flow-col items-center justify-between rounded-16',
           !post && activeClasses,
           !showNav && 'hidden',
           !post && 'border-t border-border-subtlest-tertiary',
--- a/packages/webapp/components/FooterNavBar.tsx
+++ b/packages/webapp/components/FooterNavBar.tsx
@@ -27,7 +27,7 @@ const selectedMapToTitle: Record<keyof UseActiveNav, string> = {
   home: 'Home',
   search: 'Search',
   bookmarks: 'Bookmarks',
-  notifications: 'Notifications',
+  notifications: 'Activity',
   squads: 'Squads',
   profile: 'Profile',
 };
--- a/packages/webapp/components/FooterNavBar.tsx
+++ b/packages/webapp/components/FooterNavBar.tsx
@@ -88,7 +88,7 @@ export default function FooterNavBar({
         spring="veryGentle"
         element="nav"
         className={classNames(
-          'grid w-full grid-flow-col items-center justify-between rounded-16 px-3',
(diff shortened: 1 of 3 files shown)


packages/webapp/components/FooterNavBar.tsx
```diff
@@ -88,7 +88,7 @@ export default function FooterNavBar({
         spring="veryGentle"
         element="nav"
         className={classNames(
-          'grid w-full grid-flow-col items-center justify-between rounded-16 px-3',
+          'grid w-full auto-cols-fr grid-flow-col items-center justify-between rounded-16 px-3',
```
omBratteng, line 91:
```suggestion
          'grid w-full auto-cols-fr grid-flow-col items-center justify-between rounded-16',
```
Since using `auto-cols-fr`, we could also probably remove `px-3`, as it just makes the buttons tighter.
image

omBratteng: It was never properly centered before
image
image

packages/webapp/components/footer/FooterNavBarTabs.tsx
```diff
@@ -112,13 +112,15 @@ export function FooterNavBarTabs({
...
   const tabs = useMemo(() => {
     if (notificationsNavBar) {
-      mobileUxTabs.pop();
-      mobileUxTabs.splice(-1, 0, {
-        requiresLogin: true,
-        path: '/notifications',
-        title: 'Notifications',
-        icon: (active: boolean) => <Notifications active={active} />,
-      });
+      if (!mobileUxTabs.some((tab: FooterTab) => tab?.title === 'Activity')) {
```
nensidosari, line 115: I still don't understand why would this appear multiple times? I know with the change now it wouldn't, but why in the first place?

omBratteng: React... 🤷‍♂️

omBratteng: I think it has to do with feature evaluation, since the `useMemo` depends on the feature, it gets re-triggered because of a "re-render" of useFeature somewhere somehow.

nensidosari (approved):

packages/webapp/components/FooterNavBar.tsx
```diff
@@ -27,7 +27,7 @@ const selectedMapToTitle: Record<keyof UseActiveNav, string> = {
   home: 'Home',
   search: 'Search',
   bookmarks: 'Bookmarks',
-  notifications: 'Notifications',
+  notifications: 'Activity',
```
rebelchris, line 30: Where did this decision come from, missing the context

rebelchris (approved):
Not against it, but where does the title change come from?

packages/webapp/components/FooterNavBar.tsx
omBratteng: image
From the design

rebelchris: Ah ok :) missed it before then

omBratteng: Me too 🙈

PR #3272, opened 2024-06-26
fix: onboarding on light mode

nensidosari:
## Changes
- don't keep any boot info on onboarding since there is not authenticated user after account deletion.

- this is replicated if u've just deleted your account which was in light mode

## Events

- [x] MobileL (420px)
- [x] Tablet (656px)
- [x] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

MI-400 #done

### Preview domain
https://mi-400-onboarding-not-themed.preview.app.daily.dev


packages/shared/src/contexts/AuthContext.tsx
```diff
@@ -146,6 +148,11 @@ export const AuthContextProvider = ({
     setLoginState({ trigger: AuthTriggers.LegacyLogout });
   }
 
+  const deleteAccount = async () => {
+    await dispatchDeleteAccount();
+    updateLocalBootThemeData?.();
```
sshanzel, line 153: I think it is too specific. We can probably just pass the whole update settings function so we won't have to introduce another function when we need to implement something similar in the future.

omBratteng:
Wouldn't it be easier just to hardcode the onboarding page to use the black background?

nensidosari:
@omBratteng 

> Wouldn't it be easier just to hardcode the onboarding page to use the black background?

Nope there would be a lot of overrides, since it's not just the background but also the Auth options component with all the fields and stuff, plus the logo.

packages/shared/src/contexts/AuthContext.tsx
nensidosari: @sshanzel actually after a discussion with Chris, we don't need to keep any boot data when an account is deleted, so I'll move on removing the whole thing rather then updating.

rebelchris:
The main idea is that onboarding like all other pages SHOULD support themes :) 
It's just something we sometimes forget.

And indeed on deletion (not logout) we can simply remove ALL relevant data

packages/webapp/components/layouts/AccountLayout/Security/index.tsx
```diff
@@ -142,6 +143,7 @@ function AccountSecurityDefault({
       await deleteAccount();
       await onUpdateSignBack(null, null);
       window.location.replace('/');
+      globalThis?.localStorage.removeItem(BOOT_LOCAL_KEY);
```
rebelchris, line 146: I think you might need to call it before the replace (it might become a race condition otherwise)

rebelchris (approved):
Minor comment, but otherwise good.

PR #3281, opened 2024-06-28
fix: profile update fixes

rebelchris:
- Solves two issues
- 1. Before we just to always redirect to the profile you where changing too, but on fails this would be invalid and show you someone else profile. So now we only redirect if mutation succeeds
- 2. The error wasn't shown in red (app wide issue) which this PR solves due to memo issue.

Disclaimer:
There is still a separate issue where it will reset what you filled out on error but tackling it as separate ticket.

## Events

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

AS-411  #done

### Preview domain
https://as-411-username-update-fail.preview.app.daily.dev

### DIFF
--- a/packages/shared/src/hooks/useInputFieldFunctions.ts
+++ b/packages/shared/src/hooks/useInputFieldFunctions.ts
@@ -82,9 +82,7 @@ function useInputFieldFunctions<
     if (validInput !== undefined && valid !== undefined) {
       setValidInput(valid);
     }
-    // @NOTE see https://dailydotdev.atlassian.net/l/cp/dK9h1zoM
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [valid]);
+  }, [valid, validInput]);
 
   const onBlur = () => {
     clearIdleTimeout();
--- a/packages/shared/src/hooks/useProfileForm.ts
+++ b/packages/shared/src/hooks/useProfileForm.ts
@@ -23,6 +23,7 @@ export interface ProfileFormHint {
 
 export interface UpdateProfileParameters extends Partial<UserProfile> {
   image?: File;
+  onUpdateSuccess?: () => void;
 }
 
 interface UseProfileForm {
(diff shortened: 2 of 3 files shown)


packages/shared/src/hooks/useProfileForm.ts
```diff
@@ -89,15 +90,16 @@ const useProfileForm = ({
...
         data,
         upload: image,
       }),
     {
-      onSuccess: async (_, { image, ...vars }) => {
+      onSuccess: async (_, { image, onUpdateSuccess, ...vars }) => {
         setHint({});
         await updateUser({ ...user, ...vars });
+        onUpdateSuccess?.();
         onSuccess?.();
```
nensidosari, line 103: Hard to understand where these 2 differ

nensidosari (commented):
> Before we just to always redirect to the profile you where changing too, but on fails this would be invalid and show you someone else profile. So now we only redirect if mutation succeeds

I really don't understand what the above means 😅 

> The error wasn't shown in red (app wide issue) which this PR solves due to memo issue.

What error?

rebelchris:
@nensidosari Before this fix we would simply always redirect to the "new" value of username you filled out.
So if the query failed (*due to existing username for instance) you end up on someone else profile, but your username didn't actually get updated.

Hope this makes more sense?

packages/shared/src/hooks/useProfileForm.ts
rebelchris: It's mainly bad design on the `onSuccess` but didn't want to change ALL use-cases so just added a new one that can be optionally passed to the mutation directly.

If can change it to something as this if you prefer?:
```
if(onUpdateSuccess) {
onUpdateSucces();
} else {
onSuccess?.();
}
```

nensidosari:
> Before this fix we would simply always redirect to the "new" value of username you filled out.
So if the query failed (*due to existing username for instance) you end up on someone else profile, but your username didn't actually get updated.


Got it now, so if user would be updating their username. 👍

packages/shared/src/hooks/useInputFieldFunctions.ts
```diff
@@ -82,9 +82,7 @@ function useInputFieldFunctions<
     if (validInput !== undefined && valid !== undefined) {
       setValidInput(valid);
     }
-    // @NOTE see https://dailydotdev.atlassian.net/l/cp/dK9h1zoM
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [valid]);
+  }, [valid, validInput]);
```
nensidosari, line 85: Glad one more of these is removed.

nensidosari (approved):

PR #3288, opened 2024-07-01
fix: cleanup notifications_nav_bar

rebelchris:
## Events

## Experiment

## Manual Testing

> [!CAUTION]

### Preview domain
https://as-417-cleanup-notifications-nav-bar.preview.app.daily.dev

### DIFF
--- a/packages/shared/src/hooks/useActiveNav.tsx
+++ b/packages/shared/src/hooks/useActiveNav.tsx
@@ -3,8 +3,6 @@ import { useMemo } from 'react';
 import { AllFeedPages, OtherFeedPage } from '../lib/query';
 import { SharedFeedPage } from '../components/utilities';
 import { useViewSize, ViewSize } from './useViewSize';
-import { useConditionalFeature } from './useConditionalFeature';
-import { feature } from '../lib/featureManagement';
 
 export interface UseActiveNav {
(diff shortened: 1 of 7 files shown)


omBratteng (approved):
Thanks for cleaning up my mess 😈

packages/shared/src/hooks/useActiveNav.tsx
```diff
@@ -44,8 +38,7 @@ export default function useActiveNav(activeFeed: AllFeedPages): UseActiveNav {
 
     if (!isLaptop) {
       homePages.push(OtherFeedPage.Bookmarks);
-
-      if (!notificationsNavBar && !isMobile) {
+      if (!isMobile) {
```
nensidosari, line 41: Is this the correct eval? this would always be false with the other condition now that notificationsNavBar is true since it won.

nensidosari (changes requested):
1 thing only

packages/shared/src/hooks/useActiveNav.tsx
rebelchris: It needs to be for tablet only basically 😅
Just easier this way to not break too many things.

nensidosari (approved):

PR #3294, opened 2024-07-02
feat: updated Storybook & added shortcuts component to it

ilasw:
## Changes
- **Upgrade Storybook to v8**;
- Soft refactored ShortcutLinks component for enhance readability;
- **Added  ShortcutLinks** component **to Storybook**;
- Added SB **mocks for mandatory business logic functions/libraries** (`node-fetch`, `webextension-polyfill`, `useRouter`, `browser` from extensions context);
- Added Storybook **support for API mock** using `msw`;
- Mocker `/api/boot` fetch in ShortcutLinks component;

## Manual Testing

> [!CAUTION]

### Preview domain
https://storybook-git-mi-405-shortcuts-component-storybook-dailydotdev.vercel.app/?path=/story/extension-shortcutlinks--list-feed-layout


packages/extension/src/newtab/ShortcutLinks.tsx
```diff
@@ -315,73 +247,30 @@ export default function ShortcutLinks({
...
-        )}
-      </div>
-    );
-  };
-
   return (
     <>
       {isShortcutsV1 ? (
-        <ShortcutV1 />
+        <ShortcutLinksV1
```
sshanzel, line 253: Maybe we should create a separate file for each version, just to make it easier to maintain and review.

packages/extension/src/newtab/ShortcutLinks.tsx
```diff
@@ -52,84 +52,16 @@ import ShortcutOptionsMenu from './ShortcutOptionsMenu';
...
+interface ShortcutLinksProps {
+  shouldUseListFeedLayout: boolean;
+}
 
-const ShortcutV1Item = ({
-  url,
-  onLinkClick,
-}: {
-  url: string;
+type ActualShortcutLinksProps = {
```
rebelchris, line 59: I'm a bit confused by the `Actual` vs the one above.
Why is there a difference?

packages/extension/src/newtab/ShortcutLinks.tsx
rebelchris: Yeah think it might be time.
During experiment we normally don't but seeing it's a stale experiment (will live for a while) we probably should abstract into two files.

packages/extension/src/newtab/ShortcutOptionsMenu.tsx
```diff
@@ -39,6 +39,10 @@ export default function ShortcutOptionsMenu({
     },
   ];
 
+  if (!React.isValidElement(ContextMenu)) {
+    return null;
+  }
```
rebelchris, line 44: This is just a fix for SB purposes?

packages/storybook/package.json
```diff
@@ -4,33 +4,47 @@
...
-    "build": "storybook build --docs",
+    "build": "storybook build",
     "preview-storybook": "serve storybook-static",
     "clean": "rm -rf .turbo && rm -rf node_modules",
     "lint": "eslint ./stories/*.stories.tsx"
   },
   "prettier": "@dailydotdev/prettier-config",
   "dependencies": {
     "@dailydotdev/shared": "*",
+    "@tanstack/react-query": "^4.36.1",
```
rebelchris, line 15: Wait do you need this here for sb?

packages/storybook/package.json
```diff
@@ -4,33 +4,47 @@
...
+    "@storybook/addon-essentials": "^8.1.11",
+    "@storybook/addon-interactions": "^8.1.11",
+    "@storybook/addon-links": "^8.1.11",
+    "@storybook/addon-themes": "^8.1.11",
+    "@storybook/react": "^8.1.11",
+    "@storybook/react-vite": "^8.1.11",
+    "@storybook/theming": "^8.1.11",
+    "@svgr/plugin-jsx": "^8.1.0",
+    "@svgr/plugin-svgo": "^8.1.0",
+    "@types/node": "^16.11.41",
```
rebelchris, line 40: This also doesn't match what we have, not sure it was intended?

packages/storybook/package.json
```diff
@@ -41,9 +55,9 @@
     "postcss-rem-to-responsive-pixel": "^5.1.1",
     "postcss-simple-vars": "^6.0.3",
     "serve": "^13.0.4",
-    "storybook": "^7.6.1",
-    "typescript": "4.9.5",
-    "vite": "^4.4.9",
+    "storybook": "^8.1.11",
+    "typescript": "5.3.3",
```
rebelchris, line 59: Not sure if this will give issues down the line as it doesn't match the rest of monorepo? (I prefer all packages to have the same version)

packages/storybook/stories/extension/ShortcutLinks.stories.tsx
```diff
@@ -0,0 +1,91 @@
...
+    return (
+      <ExtensionProviders><ShortcutLinks {...args} /></ExtensionProviders>
+    );
+  },
+};
+
+export default meta;
+type Story = StoryObj<typeof ShortcutLinks>;
+
+export const ListFeedLayout: Story = {
```
rebelchris, line 27: Not full against it, but somehow feel we should probably aim to make components agnostic to all our implementation details (eg they shouldn't really care about boot/settings/etc)

Would love to hear what @sshanzel / @nensidosari think about it?
(it's ok for the sake of this project perhaps to avoid scope creeping)

rebelchris (changes requested):
Not major, but prefer if we could keep logic out of storybook for now it will avoid us having to import React/tanstack/context etc.

Ideally Storybook only cares about components with data coming from elsewhere.
I know our components are now not always build in this way, but maybe something we should strive for in the long term?

packages/extension/src/newtab/ShortcutLinks.tsx
ilasw: Hi @sshanzel , thanks! Totally agree with you, was very hard time to understand how this component works.
I'm making a new directory with all component-related files and split v1 and control

packages/extension/src/newtab/ShortcutOptionsMenu.tsx
ilasw: Yes, unfortunately `dynamic` imports seems not working with the current sb setup and this is a fix for this (just don't load in Storybook)

packages/extension/src/newtab/ShortcutLinks.tsx
ilasw: Hi @rebelchris , it is the internal type for the two versions, maybe is not the best name! `Inner` prefix should be better?

packages/storybook/package.json
ilasw: Yes, extension component are consuming a lot of business logic inside and `QueryClientProvider` is mandatory in order to show the component :(

packages/storybook/package.json
ilasw: Okay, reverting ts and svgr!

packages/storybook/package.json
ilasw: Not intended, just moving to 20.14. Thanks!

packages/storybook/stories/extension/ShortcutLinks.stories.tsx
sshanzel: Yes, I think we can actually refactor it to make the component as dumb as possible. We will make it such that the component would only accept an array of objects with link and icon properties. Then that is what we will only add in our Storybook. Completely detached from all the hooks/contexts.

ilasw:
> Ideally Storybook only cares about components with data coming from elsewhere. I know our components are now not always build in this way, but maybe something we should strive for in the long term?

I totally agree. Actually there is a bond too strong between UI and business logic and it is hard to separate them, but if we do it,  it will pay in the long run!

packages/extension/src/newtab/ShortcutLinks.tsx
rebelchris: I think maybe if you abstract into 2 files it will already solve it, right?

packages/storybook/stories/extension/ShortcutLinks.stories.tsx
nensidosari: Couldn't agree more, that's why it is currently hard for us to actually just add all components we have in storybook, they are more complex then necessary.

.idea/prettier.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
rebelchris, line 1: Was this intended?

packages/extension/src/newtab/ShortcutLinks/experiments/ShortcutLinksUIV1.tsx
```diff
@@ -0,0 +1,212 @@
...
+  onMenuClick: React.MouseEventHandler<Element>;
+  onOptionsOpen: () => void;
+  onV1Hide: () => void;
+  shortcutLinks: string[];
+  shouldUseListFeedLayout: boolean;
+  showTopSites: boolean;
+  toggleShowTopSites: () => void;
+};
+
+const ShortCutV1Placeholder: FC<{
```
rebelchris, line 35: Not pro/against but we never really use FC format anywhere.

packages/extension/src/newtab/ShortcutLinks/experiments/ShortcutLinksUIV1.tsx
```diff
@@ -0,0 +1,212 @@
...
+    <div className="group flex flex-col items-center">
+      <div className="mb-2 flex size-12 items-center justify-center rounded-full bg-surface-float text-text-secondary">
+        {children}
+      </div>
+      <span className="h-2 w-10 rounded-10 bg-surface-float" />
+    </div>
+  );
+};
+
+export const ShortcutLinksUIV1: FC<ShortcutLinksV1Props> = (props) => {
```
rebelchris, line 104: This one is now missing all the tracking events basically.
We'll need to probably find a way to add a hook that has all actions as they are shared between the two.

ilasw: I've passed all the functions from the parent isolating the UI from logic but they should be still running (like in the attached sample). I'm not fully aware on how the events works, do I need to use them in the same component?

image

packages/extension/src/newtab/ShortcutLinks/experiments/ShortcutLinksUIV1.tsx
ilasw: Okay! If preferred I can switch to `function Name(props: Props): ReactElement` is that better?

.idea/prettier.xml
ilasw: I've seen that `.idea` is not included in `.gitignore` so i kept it, but I can remove if may cause errors 👀

packages/extension/src/newtab/ShortcutLinks/experiments/ShortcutLinksUIV1.tsx
rebelchris: nope sorry my bad, missed that part and thought you rendered them individual now 😅

packages/extension/src/newtab/ShortcutLinks/experiments/ShortcutLinksUIV1.tsx
rebelchris: It's more inline, but honestly don't mind.
Maybe we can open it up on [#guild-web](https://dailydotdev.slack.com/archives/C02E2C3C13R) slack channel.

I do prefer 1 or the other just so everyone uses the same.

.idea/prettier.xml
rebelchris: Not sure, but it becomes a personal overwrite everytime then.
Maybe better we add it to gitignore :)

packages/extension/src/newtab/ShortcutLinks/experiments/ShortcutLinksUIV1.tsx
ilasw: Okay, just keeping the old one for now and then will see ;)

rebelchris (approved):

nensidosari (approved):
Looks great! Great work on your first PR!! 🎉

PR #3298, opened 2024-07-03
fix: solve profile rerenders

rebelchris:
## Changes

Technically the change by @sshanzel from `RenderForm` to `form` fixes everything.
But edge-case where photo/timezone update could cause re-mount would still loose state.

This is more failsafe addition as we already build it 😅

## Events

## Experiment

## Manual Testing

> [!CAUTION]

### Preview domain
https://fix-profile-state-renders.preview.app.daily.dev


packages/webapp/pages/account/profile.tsx
```diff
@@ -63,6 +63,7 @@ const AccountProfilePage = (): ReactElement => {
       update: true,
     }),
   );
+  const [updatedUser, setUpdatedUser] = useState(user);
```
capJavert, line 66: only thing that comes to mind is that `user` here could be `undefined` on initial render and then we get no values in `updatedUser` but it should not happen since it comes from boot? We need forms lib!

packages/shared/src/components/fields/Textarea.tsx
```diff
@@ -103,7 +104,12 @@ function Textarea({
         id={inputId}
         ref={inputRef}
         onFocus={onFocus}
-        onBlur={onBlur}
+        onBlur={(e) => {
+          if (onExternalBlur) {
+            onExternalBlur(e);
+          }
+          onBlur();
+        }}
```
capJavert, line 112: This was a fix for focus or was missing all along on `Textarea`?

capJavert (approved):

packages/webapp/pages/account/profile.tsx
rebelchris: I'm open to just deleting this whole PR as Lee's fix is enough for now.

packages/shared/src/components/fields/Textarea.tsx
rebelchris: Missing in textarea :)

nensidosari (approved):

rebelchris:
Closing due to complexity/already being solved differently.

PR #3299, opened 2024-07-03
fix: cleanup onboarding_most_visited

rebelchris:
## Changes

Cleanup onboarding_most_visited it won but only for users after set date.
Decided for this reason to abstract into a hook so the logic had one source of truth.

## Events

## Experiment

## Manual Testing

> [!CAUTION]

AS-425 #done 

### Preview domain
https://as-425-cleanup-onboarding-most-visited.preview.app.daily.dev

### DIFF
--- a/packages/shared/src/hooks/utils/useShortcuts.ts
+++ b/packages/shared/src/hooks/utils/useShortcuts.ts
@@ -1,0 +1,24 @@
+import { useMemo } from 'react';
+import { useAuthContext } from '../../contexts/AuthContext';
+
+interface UseShortcuts {
+  isShortcutsV1: boolean;
+}
+
+/**
+ * Hook to determine if the user is on the old version of shortcuts
+ * More information in this thread: https://dailydotdev.slack.com/archives/C02JAUF8HJL/p1720008775281939
+ * Keywords: shortcutsUI, onboarding_most_visited
+ */
+export const useShortcuts = (): UseShortcuts => {
+  const { user } = useAuthContext();
+
+  const isShortcutsV1 = useMemo(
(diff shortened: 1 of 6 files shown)


nensidosari (approved):

capJavert (approved):

packages/shared/src/hooks/utils/useShortcuts.ts
```diff
@@ -0,0 +1,24 @@
...
+/**
+ * Hook to determine if the user is on the old version of shortcuts
+ * More information in this thread: https://dailydotdev.slack.com/archives/C02JAUF8HJL/p1720008775281939
+ * Keywords: shortcutsUI, onboarding_most_visited
+ */
+export const useShortcuts = (): UseShortcuts => {
+  const { user } = useAuthContext();
+
+  const isShortcutsV1 = useMemo(
+    () => user?.createdAt >= '2024-04-16T00:00:00.000Z',
```
omBratteng, line 17: I know it's valid, I just really dislike that is is possible to do `>=` against a string :triggered:

rebelchris: Alternative I can wrap it into a date conversion lib function?
Just that this was quickest 😅

omBratteng:
```suggestion
    () => new Date(user?.createdAt) >= new Date('2024-04-16T00:00:00.000Z'),
```

Could also extract that `2024-04-16T00:00:00.000Z`, think I saw it used more places?

rebelchris: It's used one more space, but just "random" that two experiments finished on one day

PR #3300, opened 2024-07-03
fix: dropdown should focus on trigger when context menu closes;

ilasw:
## Changes
Improved Dropdown accessibility.

- Dropdown should focus on trigger when context menu closes (sample reference: https://pauljadam.com/demos/bootstrap-dropdown.html );
- Updated random id generation with react 18 `useId()`;

## Manual Testing

- [ ] Checked on Chrome;
- [ ] Checked on Firefox;

### Preview domain
https://fix-dropdown-accessibility.preview.app.daily.dev


packages/shared/src/components/fields/Dropdown.tsx
```diff
@@ -123,6 +126,12 @@ export function Dropdown({
 
   const fullScreen = openFullScreen ?? isMobile;
 
+  useLayoutEffect(() => {
+    if (wasVisible === 'true' && !isVisible) {
+      triggerRef?.current?.focus?.();
+    }
+  }, [isVisible]);
```
sshanzel, line 133: I think we can avoid the additional hooks but just adding the `triggerRef?.current?.focus?.();` at the `onHidden` callback of the `Menu` component.

ilasw: Hi @sshanzel, thanks for comment!  I've tried, but unfortunately `onHidden` is called on the very first render of the `Menu` component, resulting in an autofocus behavior on the last dropdown on the page :(

rebelchris (approved):
Thanks for the pro-active fix 🙏
If we can't avoid the layoutEffect (which probably we can't due to focus state) I'm happy with it.

nensidosari (approved):
🚀

packages/shared/src/components/fields/Dropdown.tsx
sshanzel: Oh, then I think there is something wrong with the callback. Anyways, we will replace our context menu completely in the future. I think spending more time to check why that happens is probably unnecessary. Should be good to go here! 🥳

PR #3304, opened 2024-07-04
fix: flaky test

sshanzel:
## Changes
- Checking the value of the input box at the end doesn't really prove anything.
- It may have some relevance, but at the start of the flow. Hence, moved it.
- The important factor is whether the mock of updating the settings was executed. Should be tested here now.

## Events

## Experiment

## Manual Testing

> [!CAUTION]

### DIFF
--- a/packages/webapp/__tests__/AccountSecurityPage.tsx
+++ b/packages/webapp/__tests__/AccountSecurityPage.tsx
@@ -121,12 +121,10 @@ const verifySession = async (email = defaultLoggedUser.email) => {
     method: 'password',
   };
   mockKratosPost({ action, params }, verifiedLoginData);
-  await act(async () => {
-    const submitLogin = await screen.findByText('Verify');
-    fireEvent.click(submitLogin);
-    await waitForNock();
-    expect(refetchBoot).toHaveBeenCalled();
-  });
+  const submitLogin = await screen.findByText('Verify');
+  fireEvent.click(submitLogin);
+  await waitForNock();
+  expect(refetchBoot).toHaveBeenCalled();
   return true;
 };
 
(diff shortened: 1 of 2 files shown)


packages/webapp/__tests__/AccountSecurityPage.tsx
```diff
@@ -262,10 +262,10 @@ it('should allow setting new password but require to verify session', async () =
   const submitResetPassword = await screen.findByText('Set password');
   fireEvent.click(submitResetPassword);
   await waitForNock();
-  mockSettingsValidation(params);
+  const mock = mockSettingsValidation(params);
   await verifySession();
-  const input = await screen.findByPlaceholderText('Password');
-  expect(input).toHaveValue('#123xAbc');
+  await waitForNock();
+  expect(mock.isDone()).toBeTruthy();
```
sshanzel, line 268: To prove that the expectation works, I tried to comment out the line `await verifySession` - and it failed. We need the verification first so the mock will be triggered.

omBratteng (approved):
lgtm, lets see it it fixes flakiness

sshanzel:
> lgtm, lets see it it fixes flakiness

Whew. I thought it did not. There are two flaky test, the other one is fixed, let me check this bookmarks one.

sshanzel:
They just keep popping up :hidethepain:

Will check the extension flaky test (though I think I've seen this once too).

sshanzel:
@omBratteng everything finally turned green 🟢

packages/extension/src/newtab/ShortcutLinks.spec.tsx
```diff
@@ -1,8 +1,7 @@
-import React from 'react';
+import React, { act } from 'react';
```
rebelchris, line 1: You removed act right?
Not sure is it used elsewhere in file?

sshanzel: Ah, yes, but that is for the other file.

PR #3305, opened 2024-07-04
fix: use profile hook looping reloads

sshanzel:
## Changes
- The `useProfile` hook is used in different places, for some reason, if we use placeholderData, it triggers loop of re-rendering. I think it could be that the data referenced object is becoming unstable (highly unlikely though).
- To fix the issue, we used `initialData` which sets the returned object than using a placeholder object.
- Also applied StaleTime.

## Events

## Experiment

## Manual Testing

> [!CAUTION]

### Preview domain
https://fix-profile.preview.app.daily.dev


rebelchris (approved):
It's still slow, but does seem to fix it?
https://fix-profile.preview.app.daily.dev/dante979

sshanzel:
> It's still slow, but does seem to fix it? https://fix-profile.preview.app.daily.dev/dante979

It loads fast to me now. Probably cache? Let's see on prod maybe. But yeah, fixes the tabs issue.

PR #3307, opened 2024-07-04
fix: flaky extension test for shortcut links

sshanzel:
## Changes
- Fix the extension flaky test. Sometimes the evaluation happens without the re-render not yet finished. Hence, flaky result.

## Events

## Experiment

## Manual Testing

> [!CAUTION]

### Preview domain
https://fix-flaky-extension.preview.app.daily.dev

### DIFF
--- a/packages/extension/src/newtab/ShortcutLinks.spec.tsx
+++ b/packages/extension/src/newtab/ShortcutLinks.spec.tsx
@@ -70,14 +70,7 @@ jest.mock('webextension-polyfill', () => {
   };
 });
 
-beforeEach(() => {
-  nock.cleanAll();
-  jest.clearAllMocks();
-});
-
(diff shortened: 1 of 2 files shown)


packages/extension/src/newtab/ShortcutLinks.spec.tsx
```diff
@@ -94,7 +94,6 @@ const defaultSettings: RemoteSettings = {
     'http://custom2.com',
     'http://custom3.com',
     'http://custom4.com',
-    'http://custom5.com',
```
sshanzel, line 97: Had to reduce the amount to make it easier to debug since the count for placeholder and this list is the same. Also why I had to reduce the count of the tests below by 1.

rebelchris (approved):

PR #3311, opened 2024-07-05
fix: showing streak when disabled

sshanzel:
## Changes
- Was not able to reproduce the actual issue but I observed all the evaluations we did when not showing the milestone.
- The only evaluation that could be the fault is the checking of actions they've made to hide the popup.
- There could be a race where the value returned is `undefined` which is considered as falsy.
- Also refactored usage of let. Not a fan of it as it is prone to being mutated somewhere else.

## Events

## Experiment

## Manual Testing

> [!CAUTION]

MI-410

### Preview domain
https://mi-410.preview.app.daily.dev


rebelchris (approved):
Worth a try. Agree let are not ideal. Think i vaguely remember it being a discussion on the initial PR

nensidosari (approved):
Lets try 🙌

PR #3315, opened 2024-07-05
feat: slack integration

capJavert:
## Changes

Slack integration
- new settings section for managing this and future integrations
- slack connect button on squad pages (visible to admins)
- slack modal for managing slack <-> source integrations
- new hooks for managing integrations and slack oauth
- new hook for joining private source from post page

For more info/testing also see [here](https://dailydotdev.slack.com/archives/C0795F3KKAP/p1723017198466439)

## Events

| New | set integration | target_id: slack, extra: { source } |
| New | remove integration | target_id: slack, extra: { source } |
| New | revoke integration access | target_id: slack |
| New | start adding workspace | target_id: slack, extra: { origin, source } |
| New | start adding integration | target_id: slack, extra: { source, has_integrations }) |
| New | manage integration | target_id: slack, extra: { source } |

## Experiment

## Manual Testing

> [!CAUTION]

AS-473
AS-474
AS-475

### Preview domain
https://as-413-slack-integration.preview.app.daily.dev


packages/shared/src/components/MainLayout.tsx
```diff
@@ -154,8 +154,14 @@ function MainLayoutComponent({
...
 
+  const ignoredUtmMediumForLogin = ['slack'];
   const utmSource = router?.query?.utm_source;
-  const shouldShowLogin = !user && isAuthReady && utmSource === 'notification';
+  const utmMedium = router?.query?.utm_medium;
+  const shouldShowLogin =
+    !user &&
+    isAuthReady &&
+    utmSource === 'notification' &&
+    !ignoredUtmMediumForLogin.includes(utmMedium as string);
```
rebelchris, line 164: Lol this is gonna be a pain 😂

packages/shared/src/components/modals/SlackIntegrationModal.tsx
```diff
@@ -0,0 +1,351 @@
...
+
+  const { data: sourceIntegration } = useSourceIntegration({
+    userIntegrationType: UserIntegrationType.Slack,
+    sourceId: source.id,
+  });
+
+  const { data: slackIntegrations, isLoading: isLoadingIntegrations } =
+    useIntegrations({
+      queryOptions: {
+        select: useCallback((data) => {
```
rebelchris, line 64: This is just a future proofing for different integrations?
Feels like a lot of overhead now, but maybe it's fine.

Just wondering if `useIntegrations` is still valid, but then have a `useSlackIntegration` which does abstraction on top is better?

packages/shared/src/components/modals/SlackIntegrationModal.tsx
```diff
@@ -0,0 +1,351 @@
...
+  const modalProps: ModalProps = {
+    kind: Modal.Kind.FlexibleCenter,
+    size: Modal.Size.XSmall,
+    isDrawerOnMobile: true,
+    ...props,
+  };
+
+  if (isLoadingIntegrations) {
+    return (
+      <Modal {...modalProps}>
```
rebelchris, line 144: Won't this flash? by showing/hiding a whole modal?

packages/shared/src/components/modals/SlackIntegrationModal.tsx
```diff
@@ -0,0 +1,351 @@
...
+        <Modal.Body className="flex items-center justify-center">
+          <Loader />
+        </Modal.Body>
+      </Modal>
+    );
+  }
+
+  if (noIntegrations) {
+    return (
```
rebelchris, line 153: I'm kind of wondering if we should not just change only the inside of the modal, not the full modal?

packages/shared/src/components/modals/SlackIntegrationModal.tsx
```diff
@@ -0,0 +1,351 @@
...
+              return (
+                <span className="flex gap-2 typo-callout">
+                  <ItemIcon />
+                  {slackIntegrations[index].name}
+                </span>
+              );
+            }}
+            icon={<SlackIcon />}
+            options={slackIntegrations?.map((item) => item.name)}
+            onChange={(value, index) => {
```
rebelchris, line 255: Quite a lot of big inline functions, can we somehow maybe move them top level or to a hook?

packages/shared/src/components/modals/SlackIntegrationModal.tsx
```diff
@@ -0,0 +1,351 @@
...
+            color={TypographyColor.Tertiary}
+          >
+            Select channels
+            {!!channels?.length && (
+              <Typography
+                className="ml-1 flex min-h-6 min-w-6 items-center justify-center rounded-full bg-surface-float p-0.5"
+                tag={TypographyTag.Span}
+                type={TypographyType.Callout}
+                color={TypographyColor.Tertiary}
+                bold={false}
```
rebelchris, line 288: Isn't this the default?

packages/shared/src/components/modals/SlackIntegrationModal.tsx
```diff
@@ -0,0 +1,351 @@
...
+              </Typography>
+            )}
+          </Typography>
+          {!!channels?.length && (
+            <Dropdown
+              placeholder="Select channel"
+              shouldIndicateSelected
+              buttonSize={ButtonSize.Medium}
+              iconOnly={false}
+              key="feed"
```
rebelchris, line 300: feed is ok, not sure it's used already?

packages/shared/src/hooks/integrations/useIntegration.ts
```diff
@@ -0,0 +1,111 @@
...
+  removeSourceIntegration: ({
+    sourceId,
+    integrationId,
+  }: {
+    sourceId: string;
+    integrationId: string;
+  }) => Promise<void>;
+};
+
+export const useIntegration = (): UseIntegration => {
```
rebelchris, line 32: Why did you split between singular and not?

packages/shared/src/lib/integrations.ts
```diff
@@ -0,0 +1,38 @@
...
+};
+
+export const getIconForIntegration = (
+  type: UserIntegrationType,
+): FunctionComponent<IconProps> => {
+  return integrationTypeToIconMap[type] || AppIcon;
+};
+
+export const deleteSourceIntegrationPromptOptions: PromptOptions = {
+  title: labels.integrations.prompt.deleteSourceIntegration.title,
```
rebelchris, line 23: Are these actually re-used or only in this file? (the labels)

rebelchris (commented):
Nice work, mainly some questions, feel free to re-request review.

omBratteng:
image
Are these buttons supposed to be different sizes?

And it feels off/ strange that the close button hovers over the image that way 😬

packages/shared/src/lib/integrations.ts
capJavert: for now no, but was easier to just push them to lib because they could easily be

packages/shared/src/hooks/integrations/useIntegration.ts
capJavert: yeah, naming wise maybe a bit confusing but basically useIntegrations is query to list integration and this hook is more of a util for mutations, I might refactor query ones to `useIntegrationsQuery` and so on.

packages/shared/src/components/modals/SlackIntegrationModal.tsx
capJavert: it is per component, but yeah, better to update, its weird

packages/shared/src/components/modals/SlackIntegrationModal.tsx
capJavert: it does not from what I see, but regarding your comment https://github.com/dailydotdev/apps/pull/3315#discussion_r1709628746 I will refactor the insides

packages/shared/src/components/modals/SlackIntegrationModal.tsx
capJavert: its valid yes because `useSlackIntegration` checks single integration for source

packages/shared/src/components/MainLayout.tsx
capJavert: yep 😆

packages/shared/src/components/modals/SlackIntegrationModal.tsx
capJavert: Refactored here 46ae06df59c3cda83080ad6b1bdfc273e8a6d3e4

packages/shared/src/hooks/integrations/useIntegration.ts
capJavert: Renamed all query based hooks to end with `Query` so they don't overlap with utility hooks 438fa52b6d4f30cb652b2ea5ba8b3ae2c1677c9f

capJavert:
@omBratteng fixed the modal close button and sizes.

omBratteng:
image

omBratteng:
After I connect, `?lzym=slackIntegration` gets added to URL. This should maybe removed from URL after I've pressed save or closed the modal?

capJavert:
image

@omBratteng hopefully fixed!

packages/shared/src/lib/integrations.ts
rebelchris: Yeah no big feelings about it but wondering if it was one spot injection or not.

packages/shared/src/components/typography/Typography.tsx
```diff
@@ -62,9 +64,16 @@ export function Typography<TagName extends AllowedTags>({
...
+  truncate = false,
   ...props
 }: TypographyProps<TagName>): ReactElement {
-  const classes = classNames(className, type, { 'font-bold': bold }, color);
+  const classes = classNames(
+    className,
+    type,
+    { 'font-bold': bold },
+    color,
+    truncate && truncateTextClassNames,
```
rebelchris, line 75: Oh nice addition!

rebelchris (approved):
It's a big PR so did a high level re-check and manual testing.
Nothing major pops up so from my perspective good to roll :)

capJavert:
@rebelchris thanks for testing! Yeah, ended up a bit bigger then the POC, but think we should be good.

packages/shared/src/components/typography/Typography.tsx
omBratteng: image

omBratteng:
image
Lots of empty space beneath channel list?

packages/shared/src/graphql/integrations.ts
```diff
@@ -0,0 +1,127 @@
...
+  channelIds: string[];
+};
+
+export type SlackChannelConnection = {
+  data: SlackChannel[];
+};
+
+export const SLACK_CHANNELS_QUERY = gql`
+  query SlackChannels($integrationId: ID!) {
+    slackChannels(integrationId: $integrationId, limit: 999) {
```
omBratteng, line 37: Is the `limit: 999` intended to be that high?

omBratteng (approved):
Looks good, only one last comment ☝️

packages/shared/src/graphql/integrations.ts
capJavert: and for https://github.com/dailydotdev/apps/pull/3315#issuecomment-2277915616 I did not really touch dropdown component so it looks as it looks I guess 🙈

capJavert: I'll take a quick look if I can fix it though

capJavert: basically when `scrollable` is set on dropdown it gets fixed height, recon is something to do with styling so leaving it as is.

packages/shared/src/hooks/integrations/slack/useSlack.ts
```diff
@@ -0,0 +1,94 @@
...
+  const { user, isLoggedIn } = useAuthContext();
+  const { logEvent } = useLogContext();
+  const { value: slackIntegration, isLoading: isFeatureLoading } =
+    useConditionalFeature({
+      feature: feature.slackIntegration,
+      shouldEvaluate: isLoggedIn,
+    });
+
+  return {
+    connect: useCallback(
```
nensidosari, line 45: Why not declare these above, think it makes what is being returned clearer

packages/shared/src/hooks/integrations/slack/useSlackChannelsQuery.ts
```diff
@@ -0,0 +1,67 @@
...
+      return result.slackChannels.data.sort((a, b) => {
+        if (a.name < b.name) {
+          return -1;
+        }
+
+        if (a.name > b.name) {
+          return 1;
+        }
+
+        return 0;
```
nensidosari, line 53: Maybe we can abstract this into a general func to sort alphabetically

packages/webapp/pages/account/integrations/index.tsx
```diff
@@ -0,0 +1,160 @@
...
+                size={ButtonSize.Small}
+                onClick={onConnectNew}
+              >
+                Connect
+              </Button>
+            </div>
+          </div>
+        )}
+        {data?.map((integration) => {
+          return (
```
nensidosari, line 108: seems like u can return directly

packages/webapp/pages/account/integrations/index.tsx
```diff
@@ -0,0 +1,160 @@
...
+            <div className="ml-auto flex items-center gap-2">
+              <Button
+                variant={ButtonVariant.Secondary}
+                size={ButtonSize.Small}
+                onClick={onConnectNew}
+              >
+                Connect
+              </Button>
+            </div>
+          </div>
```
nensidosari, line 151: Could this be a separate component in another file

packages/webapp/pages/account/integrations/index.tsx
```diff
@@ -0,0 +1,160 @@
...
```
nensidosari, line 105: Could this be a separate component in another file

nensidosari (approved):
Looks good - but quite chunky to review 😬 

None of my comments are blocking

packages/webapp/pages/account/integrations/index.tsx
capJavert: Its just a style preference, it does the same thing code wise, I like this one because when you have to do some logic inside `map` you don't have to change more lines

packages/shared/src/components/modals/SlackIntegrationModal/SlackIntegrationModal.tsx
```diff
@@ -0,0 +1,258 @@
...
+  const isStartTracked = useRef(false);
+
+  useEffect(() => {
+    if (!trackStart || isLoadingIntegrations) {
+      return;
+    }
+
+    if (isStartTracked.current) {
+      return;
+    }
```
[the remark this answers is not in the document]
capJavert: The ref here is specific to protecting useEffect from executing more then once. This is why its better to keep it in separate `if` for readability. Code wise the bundler will merge it either way so it ends up being the same thing eg. single `if`.

packages/shared/src/components/modals/SlackIntegrationModal/SlackIntegrationModal.tsx
```diff
@@ -0,0 +1,258 @@
...
+  const isManageTracked = useRef(false);
+
+  useEffect(() => {
+    if (!hasIntegrations) {
+      return;
+    }
+
+    if (isManageTracked.current) {
+      return;
+    }
```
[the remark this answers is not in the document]
capJavert: Same as https://github.com/dailydotdev/apps/pull/3315#discussion_r1713604916

packages/shared/src/components/integrations/UserSourceIntegrationList.tsx
```diff
@@ -0,0 +1,103 @@
...
+
+                  await removeSourceIntegration({
+                    sourceId: sourceIntegration.source.id,
+                    integrationId,
+                    integrationType: sourceIntegration.userIntegration.type,
+                  });
+                }}
+              >
+                <MiniCloseIcon />
+              </button>
```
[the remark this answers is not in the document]
capJavert: It's just a button though?

packages/shared/src/hooks/integrations/slack/useSlackChannelsQuery.ts
capJavert: For sure!

packages/shared/src/components/integrations/UserSourceIntegrationList.tsx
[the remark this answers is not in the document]
nensidosari: Mostly on my mind was a component inside render but just to separate the main return. Or maybe setting onClicks as functions before the return could make it a bit more clean. 
It's a minor thing so wouldn't argue too much which is best as I'm not 100% sure myself. up to u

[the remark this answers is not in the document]
nensidosari: 🙌

packages/shared/src/lib/func.ts
```diff
@@ -72,3 +72,17 @@ export const wrapStopPropagation =
     event.stopPropagation();
     callback();
   };
+
+export const sortAlphabeticallyByProperty =
```
nensidosari, line 76: I would argue that for a generic function it doesn't need to actually care if it has a property or not, it would just expect a and b as 2 strings.

capJavert: Yes but we almost always order by property and if I had not created it like that you would have to do `array.sort((a, b) => sortAlphabetically(a.name, b.name))` vs now you can do `array.sort(sortAlphabeticallyByProperty('name'))` which I think is more convenient. 

Additionally in the first one you are calling 2 functions in each iteration: closure and then `sortAlphabetically` vs in second example you only call one function each iteration that gets created by `sortAlphabeticallyByProperty`.

capJavert: We can add additional one that is flat if needed in the future when needed.

nensidosari: Makes sense for the usage, we can always create a separate func then if needed for plain strings.

PR #3318, opened 2024-07-05
feat: add animated upvotes in feed cards

ilasw:
## Changes
- Implement upvote animation as experiment ([figma reference](https://www.figma.com/proto/zLYukUtJGDplw6qBK59J3q/daily.dev---Mockup-(Q1_2024)-(Copy)?node-id=2914-10453&t=RbsA7knTYpvhLNnj-0&scaling=min-zoom&content-scaling=fixed&page-id=2753%3A8083&starting-point-node-id=2914%3A10453&show-proto-sidebar=1));
- Refactored Storybook stories for easier finding;
- Refactored Storybook mock system for make possibile stories for Experiments;

> This is an experiment in adding animations to the web app. I've implemented the upvote animation using CSS (by adding a new CSS file for ActionButtons), but there are many different ways to do this. We should consider the idea of implementing them with [Lottie](https://airbnb.io/lottie) or [Motion](https://www.framer.com/motion/) if more animations come up. Open to feedback on this topic!

## Events

No events added

## Experiment

> [!IMPORTANT]

## Manual Testing

> [!CAUTION]

### Preview domain
- Webapp: https://mi-406-animated-action-buttons-experiment.preview.app.daily.dev
- Storybook: https://storybook-8hpmoej5g-dailydotdev.vercel.app/?path=/story/components-actionbuttons--animated-upvote&globals=theme:dark


packages/shared/src/components/cards/ActionsButtons/ActionButtons.tsx
```diff
@@ -84,9 +117,12 @@ export default function ActionButtons({
         <QuaternaryButton
           id={`post-${post.id}-upvote-btn`}
           icon={
-            <UpvoteIcon secondary={post?.userState?.vote === UserVote.Up} />
+            <span className="pointer-events-none relative">
+              <UpvoteIcon secondary={post?.userState?.vote === UserVote.Up} />
+              {isAnimatedVersion && isUpvoteActive && <AnimatedUpvoteIcons />}
```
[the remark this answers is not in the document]
ilasw: Thanks for the feedback! I've added a control using `usePrevious` custom hook!

Speaking about it I've added a dependency in this hook (for perf) and changed to accept all kind of values, is that ok or there is a reason for no dependency array?

packages/shared/src/lib/featureManagement.ts
```diff
@@ -29,6 +33,7 @@ const feature = {
   showRoadmap: new Feature('show_roadmap', false),
   seoSidebar: new Feature('seo_sidebar', SeoSidebarExperiment.Control),
   onboardingChecklist: new Feature('onboarding_checklist', false),
+  upvote: new Feature('animated_upvote', UpvoteExperiment.Control),
```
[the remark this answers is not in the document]
ilasw: How can I check for features in growthbook? 👀

packages/shared/src/components/cards/ActionsButtons/ActionButtons.module.css
```diff
@@ -0,0 +1,131 @@
...
+    left: var(--upvote-start-x);
+    opacity: 0;
+    rotate: var(--upvote-start-rotate);
+    top: var(--upvote-start-y);
+    transform: translate3d(-50%, -50%, 0);
+    will-change: left, top, transform;
+
+    @media screen and (prefers-reduced-motion) {
+      animation: none;
+    }
```
nensidosari, line 27: Thanks for adding this, I think it's great to have!

Just one thing I'm concerned about, is there any way we can catch that in js? 

Experiment wise, if users are enrolled but see no animation because they have the reduced motion enabled might give us fake data, which means now at experimentation phase we might either want to not add this, or add this, but also guard via js with `useConditionalFeature` where the condition would be users that do not have this enabled.

packages/shared/src/components/cards/ActionsButtons/ActionButtons.tsx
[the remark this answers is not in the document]
nensidosari: Great thank u!

No don't think it would make a difference either way with or without the dependency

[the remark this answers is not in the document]
nensidosari: Btw don't think the animation works properly now when trying on preview. It stops without finishing the animation.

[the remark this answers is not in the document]
ilasw: Tried to fix with a useState tracking for click! Simple and direct :D

packages/shared/src/components/cards/ActionsButtons/ActionButtons.module.css
```diff
@@ -0,0 +1,127 @@
+.upvotes {
```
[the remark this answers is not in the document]
nensidosari: They do? I had no idea

packages/shared/src/lib/featureManagement.ts
```diff
@@ -28,6 +32,7 @@ const feature = {
...
```
rebelchris, line 35: I would say let's make the first key also more explanatory.
eg: `animatedUpvote`

packages/shared/src/lib/featureValues.ts
```diff
@@ -23,10 +23,16 @@ export interface FeatureTheme {
...
 
 export enum SeoSidebarExperiment {
   Control = 'control',
   V1 = 'v1',
 }
+
+export enum UpvoteExperiment {
+  Control = 'control',
+  Animated = 'v1',
+}
```
rebelchris, line 38: I don't think this needs to be a enum feature, can simply be boolean if it's on/off :)

packages/storybook/mock/GrowthBookProvider.tsx
```diff
@@ -0,0 +1,8 @@
+import { fn } from '@storybook/test';
+import * as actual from '@dailydotdev/shared/src/components/GrowthBookProvider';
+
+export * from '@dailydotdev/shared/src/components/GrowthBookProvider';
+
+export const useFeature = fn(actual.useFeature)
+  .mockName('useFeature')
+  .mockReturnValue('control');
```
rebelchris, line 8: This will mock all useFeature to `control`? it might impact some other values in that case

packages/webapp/next.config.js
```diff
@@ -10,8 +10,8 @@ const withBundleAnalyzer = require('@next/bundle-analyzer')({
 
 const securityHeaders = [
```
rebelchris, line 11: Many changes here, I think it's just formatting, but can we ensure nothing. changed?
It's highly sensitive to make changes in this file if not needed.

packages/shared/src/components/cards/ActionsButtons/ActionButtons.tsx
```diff
@@ -28,6 +37,48 @@ export interface ActionButtonsProps {
   className?: string;
 }
 
+const AnimatedUpvoteIcons = React.memo(function InnerAnimatedUpvoteIcons() {
```
rebelchris, line 40: I think we can actually move this to a custom file.
Then have the file use the https://storybook.daily.dev/?path=/docs/experiments-helpers--docs#withexperiment-hoc method to either return UpvotedButton (if part of experiment) or fallback to regular button.
That way it's cleaner in this file and all logic will live in one file.

Plus easier to manage enrollment and eventually cleanup.

rebelchris (changes requested):
Nice more of a minor comments on abstracting to a custom file (easier to maintain)

packages/shared/src/lib/featureValues.ts
nensidosari: Yeah I hadn't specified in the ticket and hadn't created the experiment yet, to not make Luca change anything since he was already done with this just went and created it like this.

packages/storybook/mock/GrowthBookProvider.tsx
ilasw: I would like to add an initial value, just for common cases, but it is editable in each component story! Should I remove it?

packages/webapp/next.config.js
ilasw: I tried to revert the changes, but my IDE still changes the formatting. There should not be any changes at all!

packages/shared/src/components/cards/ActionsButtons/ActionButtons.module.css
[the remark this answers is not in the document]
ilasw: Should I ask them for a Lottie JSON or we can keep the CSS version? 👀 
@nensidosari @rebelchris

packages/storybook/mock/GrowthBookProvider.tsx
rebelchris: Hmm if we can somehow set it for each story that be great?
Since you're one is the only one exposing it for now, we can probably go ahead like this.

packages/shared/src/lib/featureValues.ts
nensidosari: yeah sure

packages/shared/src/lib/featureManagement.ts
nensidosari:
```suggestion
  animatedUpvote: new Feature('animated_upvote', false),
```

@ilasw mind changing this to a boolean value 🙏 based on https://github.com/dailydotdev/apps/pull/3318#discussion_r1669892128 . I'll change the value in growthbook as well

ilasw: Done ✅

packages/shared/src/components/cards/ActionsButtons/ActionButtons.tsx
ilasw: Done ✅

packages/shared/src/components/cards/ActionsButtons/ActionButtons.module.css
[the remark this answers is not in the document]
nensidosari: I'm totally fine with both, your animation looks great, and since it's contained in one file it's easy to access/manage anyways even if we want to remove and replace with the new thing. 

Would though suggest u do pop the question to design so that we 100% sure they use Lottie and if we can use it as well and be aware for this and any future implementation.

packages/shared/src/components/cards/ActionsButtons/UpvoteButtonIcon.tsx
```diff
@@ -0,0 +1,77 @@
...
+    [userPrefersReducedMotions.replace('@media', '')],
+    [false],
+    false,
+    false,
+  );
+  const currentVersion = useConditionalFeature({
+    feature: feature.animatedUpvote,
+    shouldEvaluate: !haveUserPrefersReducedMotions,
+  });
+  const isAnimatedVersion = !!currentVersion?.value;
```
rebelchris, line 26: I'm not sure you need it here. I think basically the evaluation can move to withExperiment right @capJavert ?

packages/shared/src/lib/featureManagement.ts
nensidosari: Great I've changed the grwothbook value as well 🙌

rebelchris (changes requested):
Nice!
Thanks for the changes, think we can just simply the render to expect it to always return the experiment or the fallback. *(based on set criteria?)

Also not 100% sure the reduces motion is inline with experimentation expectation (I think it's a good addition, but something we clearly need to mention for evaluation)

Second note: We'll need to ask Ido for enrolment check based on PR here.
Do mention enrolment happens on render (due to animation effort) and based on reduces motion preference.

packages/shared/src/components/cards/ActionsButtons/ActionButtons.module.css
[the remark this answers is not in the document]
rebelchris: I think let's leave it for now as is.
Asked here: https://dailydotdev.slack.com/archives/C0381BQUEN9/p1720513195733419

nensidosari:
> Also not 100% sure the reduces motion is inline with experimentation expectation (I think it's a good addition, but something we clearly need to mention for evaluation)

@rebelchris we should definitely mention, but it wouldn't effect the experiment. We're already only enrolling users that don't have this enabled, so this shouldn't effect experiment results.

packages/shared/src/components/cards/ActionsButtons/UpvoteButtonIcon.tsx
ilasw: It is still mandatory because of the `reduces motion` I didn't find a way to insert this in the `withExperiment ` logic  :(

rebelchris: Ah yeah was wondering if we could put that inside the `value` evaluation there?
(Ante might know best since he made it)

ilasw: Trying to implement it 💪

ilasw: Now it should works without the `useConditionalFeature ` :)

rebelchris (approved):
Nice works for me :) 
Additionally we could have used `useConditionalFeature` inside `withExperiment` but I think this is perfect as well.

Just make sure to check enrolment with Ido (@nensidosari  can tell you more) before we merge it.

packages/shared/src/components/cards/ActionsButtons/UpvoteButtonIcon.tsx
```diff
@@ -0,0 +1,66 @@
...
+    props: AnimatedButtonIconProps,
+  ): ReactElement {
+    const haveUserPrefersReducedMotions = useMedia(
+      [userPrefersReducedMotions.replace('@media', '')],
+      [false],
+      false,
+      false,
+    );
+
+    const Experiment = withExperiment(AnimatedButtonIcon, {
```
capJavert, line 55: This will create a new component each time which will cause re-mount on each parent re-render, maybe better to `useConditonalFeature` here?

Are you using `memo` because of that?

rebelchris: He wants either the regular component or this animated one.
I suggested your `withExperiment` was probably best one right?

(With addition it needs to be conditional (so he added it inside))
Does that work?

packages/shared/src/components/cards/ActionsButtons/ActionButtons.tsx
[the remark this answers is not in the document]
nensidosari: great 👍 useRef is another option but definitely fine the way it is

packages/shared/src/components/cards/ActionsButtons/UpvoteButtonIcon.tsx
capJavert: Yeah, but `withExperiment` should not be defined inline, it should be const outside.

capJavert: I would say in this case it's better to use `useConditionalFeature` since he needs other dynamic data for user prefers motion.

rebelchris: Oh got it, I guess he did it because he needed the conditional eveluation 😅
Guess that brings trouble then?

So we'd need to maybe move back (Sorry Luca 😅)

nensidosari: ⬆️  😅

ilasw: Okay, let's revert 😢

ilasw: @rebelchris @nensidosari @capJavert updated PR!

packages/shared/src/components/withExperiment.tsx
```diff
@@ -25,12 +25,14 @@ export const withExperiment = <
...
     fallback?: ComponentType<Props>;
+    shouldEvaluate?: boolean;
   },
 ): ComponentType<Props> => {
   const WithExperiment = (props: Props): ReactElement => {
     const featureValue = useFeature(options.feature);
+    const { shouldEvaluate = true } = options;
 
-    if (featureValue !== options.value) {
+    if (!shouldEvaluate || featureValue !== options.value) {
```
nensidosari, line 35: let's revert this as well since we're going with useConditionalFeature

nensidosari (approved):
Looks great, and apologise for the mess here 😅 

Just one more thing to revert and then last thing is this:

## Experiment

If yes please remove the comment HTML comment tags and follow the instructions below

Don't forget to send a message to the [#experiments](https://dailydotdev.slack.com/archives/C02JAUF8HJL/p1715175315620999) channel, following the template in slack, and adding a link to the message here.

> [!IMPORTANT]

I can help you out with the enrolment details for the mesage

packages/shared/src/components/cards/ActionsButtons/UpvoteButtonIcon.tsx
```diff
@@ -0,0 +1,55 @@
...
+  const haveUserPrefersReducedMotions = useMedia(
+    [userPrefersReducedMotions.replace('@media', '')],
+    [false],
+    false,
+    false,
+  );
+
+  const { value: isAnimatedVersion } = useConditionalFeature({
+    feature: feature.animatedUpvote,
+    shouldEvaluate: !haveUserPrefersReducedMotions,
```
rebelchris, line 28: @nensidosari not 100% sure on what the idea was but don't we want the evaluation to include these?
```&& userClicked && isUpvoteActive```

**Review by rebelchris — APPROVED** [review]
Minor check, but will be raised in the evaluation check.
This will do (initially I thought we could still have 2, but only export one via the hook, but let's leave it for now 🙏

File: packages/shared/src/components/cards/ActionsButtons/UpvoteButtonIcon.tsx (hunk 13 above)
Comment (nensidosari) [reply packages/shared/src/components/cards/ActionsButtons/UpvoteButtonIcon.tsx:28]: well yes if we want to enrol only users who upvoted from when we roll this experiment. It's quite a specific enrolment, we can check with Ido

**Comment by nensidosari** [comment]
Last thing I just checked, this doesn't work on list card, did we want it there? not sure

**Comment by ilasw** [comment]
@nensidosari added mobile and fixed icon size! ✅

File: packages/shared/src/components/cards/ActionsButtons/UpvoteButtonIcon.tsx [hunk 14]
```diff
@@ -0,0 +1,55 @@
...
```
nensidosari, line 28:
```suggestion
    shouldEvaluate: !haveUserPrefersReducedMotions && userClicked && isUpvoteActive,
```

@ilasw let's add this as well since I confirmed it with Ido [here](https://dailydotdev.slack.com/archives/C02JAUF8HJL/p1720534108129839?thread_ts=1720532942.394519&cid=C02JAUF8HJL)

nensidosari:
@ilasw Awesome, please see my last comment 🙏 
Also please send this message to experiments channel and wait for Ido's ✅ . 

New feature for evaluation:

Name: [animated_upvote](https://app.growthbook.io/features/animated_upvote)
Available options: `true` | `false`
Enrolment:
- All users in the moment of upvoting through feed cards on desktop/tablet/mobile who don't have the "prefer reduced motion" enabled

Notes: Already enabled for team + [PR](https://github.com/dailydotdev/apps/pull/3318)|
If you want to check it our follow this [preview link](https://mi-406-animated-action-buttons-experiment.preview.app.daily.dev)https://mi-406-animated-action-buttons-experiment.preview.app.daily.dev/

CC: @Ido Shamun

packages/shared/src/components/cards/ActionsButtons/UpvoteButtonIcon.tsx
nensidosari: https://github.com/dailydotdev/apps/pull/3318#discussion_r1671676473

ilasw:
> @ilasw Awesome, please see my last comment 🙏 and also fix the test and linting failing.

I've found that `useConditionalFeature` requires an additional context in tests, added right now! ✅

PR #3326, opened 2024-07-08
fix: unable to open context menu in squads from mobile;

ilasw:
## Changes
- Update ContextMenu with status that comes from context;
- fixes [#1260](https://github.com/dailydotdev/daily/issues/1260) ;

## Manual Testing

From mobile:

```
1. Go to 'Squads' 
2. Click on 'Your Squad X > option'
3. Open menu  '3 vertical dots' on the squad landing page
4. Observe
```

- [x] iOS (Chrome and Safari)
- [ ] Android

### Preview domain
https://mi-419-fix-unable-to-leave-squad-on-mobile.preview.app.daily.dev/squads/nextjs

### DIFF
--- a/packages/shared/src/components/squads/SquadHeaderMenu.tsx
+++ b/packages/shared/src/components/squads/SquadHeaderMenu.tsx
@@ -27,6 +27,7 @@ import { useSquadInvitation } from '../../hooks/useSquadInvitation';
 import { Origin } from '../../lib/log';
 import { useAuthContext } from '../../contexts/AuthContext';
 import { ContextMenu as ContextMenuIds } from '../../hooks/constants';
+import useContextMenu from '../../hooks/useContextMenu';
 
 const ContextMenu = dynamic(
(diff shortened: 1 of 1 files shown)


nensidosari (approved):

sshanzel (commented):
Looks amazing, though I think we can move this fix to the `ContextMenu` component to tackle all cases. Basically, we will pass the same `isOpen` property to the `ContextMenuDrawer` implementation.

ilasw:
> Looks amazing, though I think we can move this fix to the `ContextMenu` component to tackle all cases. Basically, we will pass the same `isOpen` property to the `ContextMenuDrawer` implementation.

Yes, I think this should be better! But I don't know all the edge cases, there is somewhere a feature that require to override the context status?

sshanzel:
> Yes, I think this should be better! But I don't know all the edge cases, there is somewhere a feature that require to override the context status?

Good point. We only use the passed props for mobile (hence, the initial issue), but let's maybe park it for now so we don't have to spend too much time checking all the edge cases since as mentioned before, we have plans to replace this component completely.

PR #3332, opened 2024-07-09
feat: use api. domain for API requests

capJavert:
## Changes

- move from `app.daily.dev/api` to `api.daily.dev` for API requests
- refactor graphql `request` to `gqlClient.request` because only in client we can set `credentials: 'include'`  so that auth cookies are passed for cross origin requests
- this will hopefully save some costs on vercel proxy that was used before
- this also introduces CORS requests due to API now being on different origins, added caching for those https://github.com/dailydotdev/daily-api/pull/2048

## Events

## Experiment

## Manual Testing

> [!CAUTION]

### Preview domain
https://as-441-cors-handling.preview.app.daily.dev


packages/shared/src/components/CommentFeed.tsx
```diff
@@ -43,7 +42,7 @@ export default function CommentFeed<T>({
   const queryResult = useInfiniteQuery<CommentFeedData>(
     feedQueryKey,
     ({ pageParam }) =>
-      request(graphqlUrl, query, {
+      gqlClient.request(query, {
```
capJavert, line 45: Most change are just `request(url, query, variables)` to `gqlClient.request(query, variables)`

packages/eslint-config/index.js
```diff
@@ -10,7 +10,7 @@ module.exports = {
     'plugin:prettier/recommended',
     'plugin:react-hooks/recommended',
   ],
-  plugins: ['prettier', 'tailwindcss', 'react-hooks'],
+  plugins: ['prettier', 'tailwindcss', 'react-hooks', 'unused-imports'],
```
capJavert, line 13: Added this so I can autofix imports after refactor, we have it in API as well.

omBratteng: I both love and hate that... so many times I've temporarily comment out some code, then my editor autofixed lint issues and removed whatever the commented out code had imported :triggered:

packages/extension/src/companion/companionRequest.tsx
```diff
@@ -1,22 +1,23 @@
...
-      headers: args?.[3],
+      headers: args?.[2],
     });
 
     return initial ?? null;
   },
 };
 
-export const companionRequest = new Proxy(request, proxyRequest);
+export const companionRequest = new Proxy(gqlRequest, proxyRequest);
```
rebelchris, line 23: This one we'll have to double verify.
I think I tried this before and it might works, but best to double test it as the weird proxy was needed for some reason.

rebelchris (commented):
Most are ok, only concern would be around the ones where we pass different clients for companion.
Just needs a manual check 🙏

packages/extension/src/companion/companionRequest.tsx
capJavert: I tried it yes, basically first argument (url) was always graphqlUrl which now is no longer needed since it is set on the client init.

capJavert: This proxy is so that all requests go through background script.

packages/eslint-config/index.js
capJavert: I can see that, but most of the time I think it speeds up. I am open to removing it as I literally needed it to fix stuff in this PR 😄

omBratteng: I'm fine with keeping it, it's just _sometimes_ a bit annoying to have to re-import things 😅 Not that it's super difficult.
And often it'a nice to have, as you remove code, but forgot to remove the import. We just need an AI version of the plugin, so it knows when to remove it and not 😂

rebelchris (approved):

omBratteng (approved):

PR #3333, opened 2024-07-10
feat: use read it later hook

sshanzel:
## Changes
- The hook that will manage to send reminders and display the toast message along with the Undo action.

## Events

## Experiment

## Manual Testing

> [!CAUTION]

MI-434

### Preview domain
https://mi-434.preview.app.daily.dev


packages/shared/src/hooks/notifications/useBookmarkReminder.ts
```diff
@@ -0,0 +1,104 @@
...
+} from '../../graphql/bookmarks';
+import { useToastNotification } from '../useToastNotification';
+
+export enum BookmarkReminderPreference {
+  OneHour = 'In 1 hour',
+  LaterToday = 'Later today',
+  Tomorrow = 'Tomorrow',
+  TwoDays = 'In 2 days',
+  NextWeek = 'Next week',
+}
```
ilasw, line 16: I've seen a lot of enum's usage in app, but what if we use and object here?

```ts

const BOOKMARK_REMINDER_PREFERENCE = {
  ONE_HOUR: 'In 1 hour',
  LATER_TODAY: 'Later today',
  TOMORROW: 'Tomorrow',
  TWO_DAYS: 'In 2 days',
  NEXT_WEEK: 'Next week',
} as const;

type BookmarkReminderPreference = typeof bookmarkReminderPreferences;
type BookmarkReminderPreferenceKeys = keyof BookmarkReminderPreference;

```

this would make easier iterate over options in modal/dropdown and centralize options for further implementations 👀

capJavert: enums are just objects in compiled javascript so you can do that as well

packages/shared/src/hooks/notifications/useBookmarkReminder.ts
```diff
@@ -0,0 +1,104 @@
...
+  nextMonday.setDate(date.getDate() + daysUntilNextMonday);
+  return nextMonday;
+}
+
+const getRemindAt = (date: Date, preference: BookmarkReminderPreference) => {
+  const isPastLaterToday = date.getHours() >= 19;
+  if (
+    !Object.values(BookmarkReminderPreference).includes(preference) ||
+    (BookmarkReminderPreference.LaterToday === preference && isPastLaterToday)
+  ) {
```
capJavert, line 46: This check repeats few times, can we abstract?

packages/shared/src/hooks/notifications/useBookmarkReminder.ts
ilasw: Okay nevermind, never tried to Object.keys an enum 😱

packages/shared/src/hooks/notifications/useBookmarkReminder.ts
```diff
@@ -0,0 +1,104 @@
...
+    throw new Error('Invalid preference');
+  }
+
+  switch (preference) {
+    case BookmarkReminderPreference.OneHour:
+      return addHours(date, 1);
+    case BookmarkReminderPreference.LaterToday:
+      return new Date(date.setHours(19));
```
capJavert, line 54: why not use date-fns for this as well?

packages/shared/src/hooks/notifications/useBookmarkReminder.ts
```diff
@@ -0,0 +1,104 @@
...
+}
+
+interface UseBookmarkReminder {
+  onBookmarkReminder: (props: BookmarkReminderProps) => void;
+}
+
+export function getNextMonday(date: Date): Date {
+  const nextMonday = new Date(date);
+  const day = date.getDay();
+  const daysUntilNextMonday = (8 - day) % 7 || 7; // 8 % 7 = 1 (if today is Monday, get the next Monday)
```
capJavert, line 36: I think you can use `nextDay` from `date-fns`, should be easier.

packages/shared/src/hooks/notifications/useBookmarkReminder.spec.tsx
```diff
@@ -0,0 +1,179 @@
...
+const client = new QueryClient();
+let mockedNow: Date;
+
+const Wrapper = ({ children }) => (
+  <QueryClientProvider client={client}>{children}</QueryClientProvider>
+);
+
+describe('useBookmarkReminder hook', () => {
+  beforeEach(() => {
+    process.env.TZ = 'UTC'; // unfortunately this doesn't work. the offset is still there
```
capJavert, line 26: What does this break?

capJavert (commented):
Looks good, love to see the tests as well for the hook, some minor comments!

nensidosari (commented):
Few minor comments, looks great otherwise.

packages/shared/src/hooks/notifications/useBookmarkReminder.ts
sshanzel: Oh, let me check their docs about it.

packages/shared/src/hooks/notifications/useBookmarkReminder.spec.tsx
sshanzel: I was hoping to set some specific TZ to just have some validation that we send 0 offset. Setting the env variable didn't work.

packages/shared/src/hooks/notifications/useBookmarkReminder.ts
```diff
@@ -0,0 +1,104 @@
...
+  );
+
+  const onBookmarkReminder = useCallback(
+    (props: BookmarkReminderProps) => {
+      const { preference } = props;
+      const now = new Date();
+      const isPastLaterToday = now.getHours() >= 19;
+
+      if (
+        !Object.values(BookmarkReminderPreference).includes(preference) ||
```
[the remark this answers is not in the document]
sshanzel: Yeah, makes sense. Let's do that.

packages/shared/src/hooks/notifications/useBookmarkReminder.ts
sshanzel: Ah, my bad, should only be in a single place. No point in checking again here since we have checked it at the `onBookmarkReminder`.

packages/shared/src/hooks/notifications/useBookmarkReminder.ts
sshanzel: Let me check their docs about this 👀

packages/shared/src/hooks/notifications/useBookmarkReminder.ts
```diff
@@ -0,0 +1,104 @@
...
+
+export function getNextMonday(date: Date): Date {
+  const nextMonday = new Date(date);
+  const day = date.getDay();
+  const daysUntilNextMonday = (8 - day) % 7 || 7; // 8 % 7 = 1 (if today is Monday, get the next Monday)
+  nextMonday.setDate(date.getDate() + daysUntilNextMonday);
+  return nextMonday;
+}
+
+const getRemindAt = (date: Date, preference: BookmarkReminderPreference) => {
```
ilasw: I think this method should be exported, as it is very useful in components like the modal.
In the UI, we need to show times and dates for "next week," so we can centralize the functionality :)

capJavert (approved):

nensidosari (approved):

packages/shared/src/hooks/notifications/useBookmarkReminder.ts
sshanzel: Sure. Have exported it now. Feel free to use it 🥳

PR #3334, opened 2024-07-10
fix: null on report modal

nensidosari:
## Changes
- quick fix for null value in title report modal 
- issue reported here: https://dailydotdev.slack.com/archives/C01L0QXQJNB/p1720605400997199

## Events

## Experiment

## Manual Testing

> [!CAUTION]

### DIFF
--- a/packages/shared/src/components/modals/report/ReportPostModal.tsx
+++ b/packages/shared/src/components/modals/report/ReportPostModal.tsx
@@ -164,7 +164,7 @@ export function ReportPostModal({
       onReport={onReportPost}
       reasons={reportOptionsForActiveReason}
       heading="Report post"
-      title={`"${post?.title}"`}
+      title={post?.title}
       footer={
         <Checkbox ref={inputRef} name="blockSource" className="font-normal">
           Don&apos;t show posts from {post?.source?.name}


packages/shared/src/components/modals/report/ReportPostModal.tsx
```diff
@@ -164,7 +164,7 @@ export function ReportPostModal({
       onReport={onReportPost}
       reasons={reportOptionsForActiveReason}
       heading="Report post"
-      title={`"${post?.title}"`}
+      title={post?.title}
```
capJavert, line 167: Not sure if we wanted the title "quoted", maybe this preserves it?
```suggestion
      title={post?.title ? `"${post.title}"` : undefined}
```

capJavert (approved):

packages/shared/src/components/modals/report/ReportPostModal.tsx
nensidosari: Ah yes didn't think of that 🤦‍♀️ thanks for noticing

rebelchris: Also in some places we use `DEFAULT_POST_TITLE` not sure if that was intended here maybe as well?

PR #3335, opened 2024-07-10
feat: added ReminderModal component

ilasw:
> [!CAUTION]
> Do not merge, I'm using invalid property from `Post`, as added in other RIL PRs

> I didn't bring the logic inside the modal because it is always used inside components where the hook was already called. Maybe a callback `onReminderSet` is enough.

## Changes
- added `ReminderModal` component;
- added utility `useToggle` hook;

## Events

## Experiment

## Manual Testing

> [!CAUTION]

### Preview domain
Web App: https://mi-430-read-it-later-modal.preview.app.daily.dev
Storybook: https://storybook-git-mi-430-read-it-later-modal-dailydotdev.vercel.app/?path=/story/components-notifications-remindermodal--default


packages/shared/src/components/notifications/ReminderModal.tsx
```diff
@@ -0,0 +1,118 @@
...
+import { ModalSize } from '../modals/common/types';
+import { Button, ButtonVariant } from '../buttons/Button';
+import {
+  getRemindAt,
+  ReminderPreference,
+} from '../../hooks/notifications/useBookmarkReminder';
+
+export type ReminderModalProps = {
+  onReminderSet: (reminder: string) => void;
+} & Pick<ModalProps, 'isOpen' | 'onRequestClose'>;
```
sshanzel, line 16: Most of the time, we prefer interface over type intersections. More information here: https://github.com/microsoft/TypeScript/wiki/Performance#preferring-interfaces-over-intersections

In summary, extending interfaces is more effective, cached, and displays conflicts.

packages/shared/src/components/notifications/ReminderModal.tsx
```diff
@@ -0,0 +1,118 @@
...
+  const defaultOption = 'OneHour';
+  const [selectedOption, setSelectedOption] = useState<string>(defaultOption);
+
+  const handleSubmit: FormEventHandler = (e) => {
+    e.preventDefault();
+    onReminderSet(selectedOption);
+  };
+
+  return (
+    <>
```
sshanzel, line 84: I think we should not need the Fragment here.

sshanzel (approved):
Looks amazing already to me. Just make sure to verify it on mobile and tablet size as well.

packages/shared/src/components/notifications/ReminderModal.tsx
ilasw: Cool, thanks for sharing!

packages/shared/src/components/notifications/ReminderModal.tsx
ilasw: Removing 👍

packages/shared/src/components/notifications/ReminderModal.tsx
```diff
@@ -0,0 +1,129 @@
...
+};
+
+export const ReminderModal = (props: ReminderModalProps): ReactElement => {
+  const { isOpen, onRequestClose, onReminderSet } = props;
+  const defaultOption = 'OneHour';
+  const [selectedOption, setSelectedOption] = useState<string>(defaultOption);
+
+  const handleSubmit: FormEventHandler = (e) => {
+    e.preventDefault();
+    onReminderSet(selectedOption);
```
sshanzel, line 81: I think this is going to be an issue when using the hook to process the submission. The hook accepts a `preference` prop, and the acceptable values are the labels defined in the enum. So for the `MODAL_OPTIONS` we should also use the `value` as the same as what we put on the label.

Then for the `defaultOption` const, we can then use `defaultOption = ReminderPreference.OneHour` to avoid magic strings.

sshanzel (changes requested):
Just noticed one thing with regards to the data that we will pass. Also, I think we can just define the hook here rather than instantiate it every time we need to open the modal. We can just pass the `post` as a prop to this modal and we should have everything we need to make the hook work.

nensidosari (commented):
Looks really good, agree with Lee's comment above as well. 

One thing I noticed but that I think it's more related to the PR that was merged before this is about the hours shown in the modal for selection:
Screenshot 2024-07-11 at 9 38 20 AM

Didn't we want the hours to be 7:00, 9:00 - or we want the mins to be based on the actual time?

packages/shared/src/hooks/useToggle.ts
```diff
@@ -0,0 +1,18 @@
+import { useState } from 'react';
+
+type UseToggle = (
+  initialState: boolean,
+) => readonly [boolean, (forced?: boolean) => void];
+
+export const useToggle: UseToggle = (initialState) => {
```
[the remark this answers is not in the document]
ilasw: I can remove if can be confusing! Just added because I love to use it in different projects 👀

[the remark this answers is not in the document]
nensidosari: I like it just more of a message to think how we can promote and make it a thing we use

ilasw:
> Didn't we want the hours to be 7:00, 9:00 - or we want the mins to be based on the actual time? cc. @sshanzel

In design is always to 0 minutes, I wrote a function to change it in hook, ready to revert if we don't want it anymore!

nensidosari:
> In design is always to 0 minutes, I wrote a function to change it in hook, ready to revert if we don't want it anymore!


@ilasw u say u changed this so it will be 9:00 now and so on? I don't see that reflected on storybook, still shows as :55 for me based on my time

ilasw:
> @ilasw u say u changed this so it will be 9:00 now and so on? I don't see that reflected on storybook, still shows as :55 for me based on my time

Yes it should work right now, I've to force for storybook redeploy because it is marked as optional and sometime it is skipped :(

packages/shared/src/hooks/notifications/useBookmarkReminder.ts
```diff
@@ -27,24 +28,26 @@ interface BookmarkReminderProps {
...
 interface UseBookmarkReminder {
-  onBookmarkReminder: (props: BookmarkReminderProps) => void;
+  onBookmarkReminder: (props: BookmarkReminderProps) => Promise<EmptyResponse>;
 }
 
 export const getRemindAt = (
   date: Date,
   preference: ReminderPreference,
 ): Date => {
+  const atNineAM = (value: Date) => set(value, { hours: 9, minutes: 0 });
```
nensidosari, line 38: shall we modify the tests for this change?

nensidosari (approved):
Looks good on my opinion, only the last thing for tests if needed

packages/shared/src/hooks/notifications/useBookmarkReminder.ts
sshanzel, line 38: Nice. Forgot to consider the minutes. Good catch!

sshanzel (approved):
Looks amazing! Great fix on the minute-level of calculation 🚀

packages/shared/src/hooks/notifications/useBookmarkReminder.ts
sshanzel: Yep. Should be part of the tests to ensure accurate result.

ilasw: Done ✅

rebelchris (approved):
Thanks for keeping stories going, I love it!
Non blocking feedback from my side.

packages/shared/src/components/notifications/ReminderModal.tsx
```diff
@@ -0,0 +1,143 @@
...
+
+  return (
+    <Button
+      aria-checked={isActive}
+      aria-label={value}
+      className="w-full"
+      name="bookmarkReminder"
+      onClick={onClick}
+      role="radio"
+      type="button"
```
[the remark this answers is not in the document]
ilasw: It behaves like a radio button so this role should improve accessibility, I would like to use the radio group but it was too much styling :\

packages/shared/src/components/notifications/ReminderModal.tsx
```diff
@@ -0,0 +1,143 @@
...
+    });
+  };
+
+  return (
+    <Modal
+      drawerProps={{
+        title: 'Set reminder',
+        className: {
+          title: '!block text-center border-transparent !typo-title2 pb-2',
+        },
```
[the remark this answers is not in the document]
ilasw: There is another way of doing this? Classes cannot be override otherwise... We should use that `twMerge` 🤔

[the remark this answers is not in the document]
rebelchris: No i mean why are they different from our guideline modals more 🙏

packages/shared/src/components/notifications/ReminderModal.tsx
```diff
@@ -0,0 +1,143 @@
...
+      }}
+      isDrawerOnMobile={isOpen}
+      isOpen={isOpen}
+      onRequestClose={onRequestClose}
+      size={ModalSize.Small}
+    >
+      <ModalHeader title="Set reminder" />
```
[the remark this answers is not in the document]
sshanzel: I think this one is for Desktop. Does it not display the same header on mobile without the `title` prop? If so we might really need it.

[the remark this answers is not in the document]
ilasw: @sshanzel this ModalHeader will not show at mobile breakpoints, only adding title to `drawerProps` will make it works.

nensidosari:
@ilasw I see your description to not merge this PR, if still valid maybe consider pointing this PR to this feature branch [MI-432](https://github.com/dailydotdev/apps/tree/MI-432) so everything about bookmarks can be there before we merge the final PR

packages/shared/src/components/notifications/ReminderModal.tsx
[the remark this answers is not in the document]
sshanzel: I see. Then we really need it in this case.

PR #3338, opened 2024-07-11
feat: post card overlay remind me later

sshanzel:
## Changes
- Post card overlay when the user bookmarked a post to allow them to set a reminder.
- Refactored some reused components.
- Introduced a hook to manage overlays on cards.
- As discussed, the priority is the share overlay, then bookmark overlay.

Note: the bookmark icon change is for another PR.

Preview:
Uploading Screen Recording 2024-07-11 at 5.08.20 PM.mov…

## Events

## Experiment

## Manual Testing

> [!CAUTION]

MI-432

### Preview domain
https://mi-432.preview.app.daily.dev

### DIFF
--- a/packages/shared/src/components/Feed.tsx
+++ b/packages/shared/src/components/Feed.tsx
@@ -1,6 +1,7 @@
 import React, {
   ReactElement,
   ReactNode,
+  useCallback,
   useContext,
   useEffect,
   useMemo,
@@ -282,6 +283,12 @@ export default function Feed<T>({
     }
   }, [selectedPost]);
 
+  const onShareClick = useCallback(
+    (post: Post, row?: number, column?: number) =>
+      openSharePost({ post, columns: virtualizedNumCards, column, row }),
+    [openSharePost, virtualizedNumCards],
+  );
+
   if (!loadedSettings) {
     return <></>;
   }
@@ -377,9 +384,6 @@ export default function Feed<T>({
         ...feedLogExtra(feedName, ranking),
       },
     });
-
-  const onShareClick = (post: Post, row?: number, column?: number) =>
-    openSharePost({ post, columns: virtualizedNumCards, column, row });
 
   const post = (items[postMenuIndex] as PostItem)?.post;
   const commonMenuItems = {
--- a/packages/shared/src/components/cards/common/CardCoverContainer.tsx
+++ b/packages/shared/src/components/cards/common/CardCoverContainer.tsx
@@ -1,0 +1,20 @@
+import React, { ReactElement, ReactNode } from 'react';
+
+interface CardCoverContainerProps {
+  title: string;
+  children: ReactNode;
+}
+
+export function CardCoverContainer({
+  title,
+  children,
+}: CardCoverContainerProps): ReactElement {
+  return (
+    <span className="absolute inset-0 z-1 flex flex-col items-center justify-center">
+      <p className="mt-5 text-center font-bold typo-callout laptopL:mt-0">
+        {title}
+      </p>
+      {children}
+    </span>
+  );
+}
--- a/packages/shared/src/components/cards/list/CardCover.tsx
+++ b/packages/shared/src/components/cards/list/CardCover.tsx
@@ -1,62 +1,35 @@
 import React, { ReactElement } from 'react';
 import classNames from 'classnames';
-import { ImageProps, ImageType } from '../../image/Image';
-import VideoImage, { VideoImageProps } from '../../image/VideoImage';
 import ConditionalWrapper from '../../ConditionalWrapper';
-import { CardImage } from './ListCard';
-import { CommonCardCoverProps } from '../common';
-import { useCardCover } from '../../../hooks/feed/useCardCover';
+import {
+  SharedCardCover,
+  SharedCardCoverProps,
+} from '../common/SharedCardCover';
 
-interface CardCoverProps extends CommonCardCoverProps {
-  imageProps: ImageProps;
-  videoProps?: Omit<VideoImageProps, 'imageProps'>;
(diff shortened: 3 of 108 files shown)


packages/shared/src/components/Feed.tsx
```diff
@@ -282,6 +283,12 @@ export default function Feed<T>({
     }
   }, [selectedPost]);
 
+  const onShareClick = useCallback(
```
rebelchris, line 286: Why was the callback needed here, did it have issues?

packages/shared/src/components/cards/common/CardCoverContainer.tsx
```diff
@@ -0,0 +1,20 @@
+import React, { ReactElement, ReactNode } from 'react';
+
+interface CardCoverContainerProps {
+  title: string;
+  children: ReactNode;
+}
+
+export function CardCoverContainer({
```
rebelchris, line 8: Love the generic stuff 🙏

packages/shared/src/components/cards/list/CardCover.tsx
```diff
@@ -23,28 +22,18 @@ export function CardCoverList({
...
-    <CardCoverShare
-      post={post}
-      onShare={() => {
-        onInteract();
-        onShare(post);
-      }}
-      onCopy={onInteract}
-    />
-  );
+  const { overlay } = useCardCover({ post, onShare });
```
rebelchris, line 25: More of a generic question, what are the diffs between this one on list/card?
Wondering if maybe we can make all the shared stuff/actions uniform better.

packages/shared/src/hooks/feed/useCardCover.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+export const useCardCover = ({
+  post,
+  onShare,
+}: UseCardCoverProps): UseCardCover => {
+  const { shouldShowOverlay, onInteract } = usePostShareLoop(post);
+  const shouldShowReminder = useBookmarkReminderEnrollment(post);
+
+  const overlay = useMemo(() => {
+    if (shouldShowOverlay) {
```
rebelchris, line 27: This one should take priority over the other one?

packages/shared/src/hooks/notifications/useBookmarkReminder.ts
```diff
@@ -63,6 +68,20 @@ export const useBookmarkReminder = (): UseBookmarkReminder => {
         updatePostCache(client, postId, (post) => ({
           bookmark: { ...post.bookmark, remindAt },
         }));
+
+        if (feedQueryKey) {
```
rebelchris, line 72: Just something to really test thorogouly, could easily creep in issues that are hard to spot.

rebelchris (changes requested):
Nice think it's important to do a big final review on the big PR :) 
Some minor comments, please check.

packages/shared/src/hooks/notifications/useBookmarkReminder.ts
sshanzel: We should be safe here, this was taken from the process of `useBookmarkPost` - the only difference is the object we are trying to pass for update.

packages/shared/src/hooks/feed/useCardCover.tsx
sshanzel: Yes, as per [product's response](https://dailydotdev.slack.com/archives/C077EU278D7/p1719313817976609?thread_ts=1719313376.734949&cid=C077EU278D7).

packages/shared/src/components/cards/list/CardCover.tsx
sshanzel: After checking, the only difference seems to be with the `wrapper` for the list mode. Not sure if there was any reason behind adding almost the same thing. Probably to avoid any unnecessary change on card that affects list.

packages/shared/src/components/Feed.tsx
sshanzel: We will use this function as a dependency way down so we need it to be a stable callback.

packages/shared/src/components/cards/list/CardCover.tsx
rebelchris: So maybe we can abstract the inner to a common file? 🙏

sshanzel: Yeah, makes sense. We can abstract it even further. Let me check quickly.

nensidosari:
Btw, if u intended to post an image/video at the description, unfortunately that has failed halfway 😅 
@sshanzel

sshanzel:
> Btw, if u intended to post an image/video at the description, unfortunately that has failed halfway 😅
> @sshanzel

We can maybe enable this for members, so we can better check on prod. Wdyt?

nensidosari:
Screenshot 2024-07-11 at 4 32 32 PM
@sshanzel there's an issue with the images now

sshanzel:
> Screenshot 2024-07-11 at 4 32 32 PM @sshanzel there's an issue with the images now

Probably due to the refactor. Let me check quickly where it is coming from.

sshanzel:
@nensidosari should be back to normal now 🥳

packages/shared/src/components/cards/list/CardCover.tsx
sshanzel: Refactored and using the same component now 🚀

sshanzel:
@rebelchris @nensidosari should be ready for final review 🥳

rebelchris (approved):

nensidosari (approved):
Looks good on my side as well 🎉

PR #3339, opened 2024-07-11
fix: clicking shortcuts options did nothing

capJavert:
## Changes

Not sure when it broke but when testing something else I noticed the below button does nothing. I pushed it to use the same method to open menu as control version.

image

## Events

## Experiment

## Manual Testing

> [!CAUTION]


rebelchris (approved):
Oh damn!
just pushed a extension already.
Will have to do one tomorrow then.

Do wonder how this broke 😅

PR #3340, opened 2024-07-11
refactor: bookmark actions to show popup on click when reminder is set

sshanzel:
## Changes
- Introduce a `BookmarkButton` to make it easy to manage state and actions that are possible for the user to take.
- Fix some naming issues.
- Update bookmark buttons below to use the new component:
  - Post card
  - Post list
  - Post modal/page
- Fix cache when the reminder is set.

https://github.com/dailydotdev/apps/assets/13744167/f4350b81-6cca-4929-a401-efcddaaf039a

## Events

## Experiment

## Manual Testing

> [!CAUTION]

MI-437

### Preview domain
https://mi-437.preview.app.daily.dev


rebelchris (commented):
More generic, isn't this a experiment, this would now always show right?

packages/shared/src/components/buttons/QuaternaryButton.tsx
```diff
@@ -9,11 +9,11 @@ import React, {
 import classNames from 'classnames';
 import { AllowedTags, Button, ButtonElementType, ButtonProps } from './Button';
 
-type QuandaryButtonProps = {
-  id: string;
-  reverse?: boolean;
-  responsiveLabelClass?: string;
-};
+export type QuaternaryButtonProps<TagName extends AllowedTags> =
```
nensidosari, line 12: why do we  use this as a separate button again? 😅

nensidosari (approved):
Looks good

packages/shared/src/components/buttons/QuaternaryButton.tsx
sshanzel: One word. Legacy 😓

Hopefully, we can spend some time in the future and understand whether we can actually remove this.

packages/shared/src/components/buttons/BookmarkButton.tsx
```diff
@@ -0,0 +1,72 @@
...
+      return onMenuClick(e);
+    }
+
+    return buttonProps.onClick?.(e);
+  };
+
+  return (
+    <>
+      <SimpleTooltip content={post.bookmarked ? 'Remove bookmark' : 'Bookmark'}>
+        <QuaternaryButton
```
rebelchris, line 44: Are you sure this is right though?
I mean the current actionButtons render as:

```
<Button
            id={`post-${post.id}-bookmark-btn`}
            className="pointer-events-auto ml-2"
            icon={<BookmarkIcon secondary={post.bookmarked} />}
            onClick={() => onBookmarkClick(post)}
            color={ButtonColor.Bun}
            pressed={post.bookmarked}
            variant={ButtonVariant.Float}
          />
```

sshanzel: Let me check where this is.

sshanzel: Check the difference between the buttons. Should practically be the same, the point of the Quaternary button is to have a wrapper so that when you add a label, the label will not have the same background as the button. Since this is for the list feed, we don't show any label from the actions there in any format. We should be good on that end.

sshanzel: I think we should be able to merge here, and if we really want to avoid the extra wrapper, I think we can work on it if the experiment wins. Thoughts?

PR #3342, opened 2024-07-11
fix: manage to do some more strict related items

rebelchris:
## Changes

Some more work around strict mode for webapp (so far)

## Events

## Experiment

## Manual Testing

> [!CAUTION]

### Preview domain
https://chore-strict-v2.preview.app.daily.dev

### DIFF
--- a/packages/shared/src/contexts/AuthContext.tsx
+++ b/packages/shared/src/contexts/AuthContext.tsx
@@ -99,7 +99,7 @@ const logout = async (reason: string): Promise<void> => {
 };
 
 export type AuthContextProviderProps = {
-  user: LoggedUser | AnonymousUser | undefined;
+  user?: LoggedUser | AnonymousUser;
   isFetched?: boolean;
   isLegacyLogout?: boolean;
   firstLoad?: boolean;
--- a/packages/webapp/pages/[userId]/posts.tsx
+++ b/packages/webapp/pages/[userId]/posts.tsx
@@ -22,7 +22,7 @@ export const getStaticPaths = getProfileStaticPaths;
 const ProfilePostsPage = ({ user }: ProfileLayoutProps): ReactElement => {
   const { user: loggedUser } = useContext(AuthContext);
   const { shouldUseListFeedLayout } = useFeedLayout();
-  const isSameUser = loggedUser?.id === user.id;
+  const isSameUser = user && loggedUser?.id === user.id;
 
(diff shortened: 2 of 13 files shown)


packages/shared/src/contexts/AuthContext.tsx
```diff
@@ -99,7 +99,7 @@ const logout = async (reason: string): Promise<void> => {
 };
 
 export type AuthContextProviderProps = {
-  user: LoggedUser | AnonymousUser | undefined;
+  user?: LoggedUser | AnonymousUser;
```
rebelchris, line 102: I personally prefer this, but let me know if people liked the otherside better

omBratteng: Same, there's a reason why `?:` exists 😸

sshanzel (approved):
Looks amazing! We'll get there eventually 🥳

packages/shared/src/contexts/AuthContext.tsx
capJavert: +1

packages/webapp/pages/[userId]/posts.tsx
```diff
@@ -42,7 +42,7 @@ const ProfilePostsPage = ({ user }: ProfileLayoutProps): ReactElement => {
       />
     ) : (
       <ProfileEmptyScreen
-        title={`${user.name} hasn't posted yet`}
+        title={`${user?.name ?? 'User'} hasn't posted yet`}
```
Comment (capJavert) [inline packages/webapp/pages/[userId]/posts.tsx:45]: Just had similar type issue on API side 😄

packages/webapp/pages/account/profile.tsx
```diff
@@ -94,7 +94,7 @@ const AccountProfilePage = (): ReactElement => {
       }),
     {
       onSuccess: async (res) => {
-        await updateUser({ ...user, cover: res.user.cover });
+        await updateUser({ ...user, cover: res.user.cover } as LoggedUser);
```
capJavert, line 97: it does not infer?  ah probably because `user` can be `undefined`

capJavert (approved):

packages/webapp/pages/account/profile.tsx
rebelchris: It's the difference between `LoggedUser` and `AnnonymousUser` one has mandatory cover, other doesn't basically.

PR #3344, opened 2024-07-12
fix: application for public squad

ilasw:
## Changes
- created `PublicSquadSubmissionActions`;
- added missing cta for public request in checklist;
- implemented component for shared logic;
- fixed wrong condition in `isQueryEnabled`  on `Details.tsx`;

## Events

## Experiment

## Manual Testing

> [!CAUTION]

### Preview domain
https://mi-439-application-for-public-squad-doesn-t-work.preview.app.daily.dev

### DIFF
--- a/packages/shared/src/components/squads/PublicSquadSubmissionActions.tsx
+++ b/packages/shared/src/components/squads/PublicSquadSubmissionActions.tsx
@@ -15,7 +15,7 @@ interface PublicSquadSubmissionActionsProps {
   isDetailsVisible?: boolean;
 }
 
-export const PublicSquadSubmissionActions = (
+const PublicSquadSubmissionActions = (
   props: PublicSquadSubmissionActionsProps,
 ): ReactElement => {
   const { squad, isDetailsVisible = true } = props;
--- a/packages/shared/src/components/squads/PublicSquadSubmissionActions.tsx
+++ b/packages/shared/src/components/squads/PublicSquadSubmissionActions.tsx
@@ -1,0 +1,80 @@
+import React, { useCallback } from 'react';
+import { Button, ButtonSize, ButtonVariant } from '../buttons/Button';
+import { squadsPublicGuide } from '../../lib/constants';
+import { anchorDefaultRel } from '../../lib/strings';
+import { SquadStatus } from './settings';
+import { TimerIcon } from '../icons';
+import { usePublicSquadRequests } from '../../hooks';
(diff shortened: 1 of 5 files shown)


packages/shared/src/components/squads/PublicSquadSubmissionActions.tsx
```diff
@@ -0,0 +1,78 @@
...
+const PublicSquadSubmissionActions = (
+  props: PublicSquadSubmissionActionsProps,
+): ReactElement => {
+  const { squad, isDetailsVisible = true } = props;
+  const { isFetched, status } = usePublicSquadRequests({
+    sourceId: squad?.id,
+  });
+  const { openModal } = useLazyModal();
+  const postsCount = squad?.flags?.totalPosts ?? 0;
+  const isEligible = postsCount >= PUBLIC_SQUAD_REQUEST_REQUIREMENT;
```
rebelchris, line 27: It's non blocking, but maybe consideration if we can export it somewhere as a shared function? (not sure if we already have a hook or something we can re-use?

packages/shared/src/components/squads/PublicSquadSubmissionActions.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+  const onSubmit = useCallback(() => {
+    if (!squad.id) {
```
rebelchris, line 30: this means squad can be null?
Maybe we should make squad optional in the props then already?

rebelchris (approved):
Looks good nothing blocking.

packages/shared/src/components/squads/PublicSquadSubmissionActions.tsx
ilasw: I'm removing since it should not be optional, copy-paste went wrong for this ahahha
Thanks for the feedback! :(

packages/shared/src/components/squads/PublicSquadSubmissionActions.tsx
ilasw: Current hooks do not accept the post count. 
- I can move this piece of code in a shared function but maybe it will be lost in the codebase and it will be hard to be found. 
- Another option is change the arguments used by `usePublicSquadRequests` and insert there, this should be ok?

rebelchris: Also fine to leave it.
Just want to avoid if things change to eval criteria we only modify 1 space by accident.

PR #3345, opened 2024-07-13
chore: round 3 of tsc

rebelchris:
## Changes

More..

For the team:
I see we actually often "expect" everything to be there (for mutations/queries) but potentially these can be undefined which would results in broken pages/mutations.
(mostly worried about mutations eg updating a source where we can potentially send: where: { id: undefined|null }

## Events

## Experiment

## Manual Testing

> [!CAUTION]

### Preview domain
https://chore-tsc-v3.preview.app.daily.dev


sshanzel (approved):
Looks great! 🚀

PR #3347, opened 2024-07-15
feat: try before buy welcome page

capJavert:
## Changes

New page that will be directed from campaigns. 
- it is not experiment
- it is mobile only (desktop goes to onboarding)

TODO
- [x] base SEO (and robots block)
- [x] sticky header (just need to adjust it better per design)

## Events

## Experiment

## Manual Testing

> [!CAUTION]

### Preview domain
https://as-410-try-before-buy.preview.app.daily.dev/welcome (make sure you are not logged in and on mobile)


packages/webapp/pages/welcome/index.tsx
```diff
@@ -0,0 +1,97 @@
...
+    query: ANONYMOUS_FEED_QUERY,
+    pageSize: 10,
+    allowFetchMore: false,
+  };
+
+  useEffect(() => {
+    if (isLaptop) {
+      router.replace(onboardingUrl);
+    }
+  }, [isLaptop, router]);
```
capJavert, line 41: Probably also, if user is authenticated, go to homepage?

packages/webapp/pages/welcome/index.tsx
```diff
@@ -0,0 +1,112 @@
...
+
+  return (
+    <>
+      <NextSeo {...seo} />
+      <div
+        className={classNames(
+          'sticky top-0 z-max flex h-12 w-full justify-between border-b border-accent-cabbage-default px-4 py-2',
```
capJavert, line 57: Enginnering decision, its always purple I don't know of way of making it purple only when stick (scrolls from offset), except from JS but that seems overkill here. 
image

packages/webapp/pages/welcome/index.tsx
capJavert: Done in 57962fd46acc4b1aa904d081a0add185aa5dd23a

nensidosari (changes requested):
I'm guessing this new page should be behind an experiment right?

capJavert:
> I'm guessing this new page should be behind an experiment right?

Nope, this is not experiment, they will just link directly in campaigns where needed.

nensidosari:
@capJavert sorry about that, missed the description, and was basing it on specs as they mention experiment... 🤦‍♀️

nensidosari (approved):
Looks good, one last thing before you merge, this is using desktop cards, u need to add this route to the useFeedLayout hook `FeedLayoutMobileFeedPages` so it can use the proper mobile cards

packages/webapp/pages/welcome/index.tsx
```diff
@@ -0,0 +1,118 @@
...
+    feedQueryKey: generateQueryKey(OtherFeedPage.Welcome, user),
+    query: ANONYMOUS_FEED_QUERY,
+    pageSize: 10,
+    allowFetchMore: false,
+  };
+
+  useEffect(() => {
+    if (isLaptop) {
+      router.replace(onboardingUrl);
+    }
```
rebelchris, line 52: It's mobile + tablet or mobile only?

rebelchris (approved):
Overall code looks good.
Just the 2 issues raised on slack :)

packages/webapp/pages/welcome/index.tsx
capJavert: I've put it on tablet as well since sone phones might be large, page also works on desktop actually but I think this is ok.

PR #3350, opened 2024-07-15
chore: round 4 of TSC fixes

rebelchris:
## Events

## Experiment

## Manual Testing

> [!CAUTION]

### Preview domain
https://chore-tsc-v4.preview.app.daily.dev


packages/shared/src/components/RecommendedTags.tsx
```diff
@@ -5,7 +5,7 @@ import type { Tag } from '../graphql/feedSettings';
 
 interface RecommendedTagsProps {
   isLoading: boolean;
-  tags: Tag[];
+  tags?: Tag[];
```
nensidosari, line 8: now these optional things are starting to scare me 😅

packages/webapp/pages/tags/[tag].tsx
```diff
@@ -164,15 +170,16 @@ const TagPage = ({ tag, initialData }: TagPageProps): ReactElement => {
...
     ) {
       return 'blocked';
     }
     if (
-      feedSettings.includeTags?.findIndex(
-        (includedTag) => tag === includedTag,
-      ) > -1
+      feedSettings.includeTags &&
+      feedSettings.includeTags.findIndex((includedTag) => tag === includedTag) >
+        -1
```
Comment (nensidosari) [inline packages/webapp/pages/tags/[tag].tsx:182]: Not related to PR but seems like this could be abstracted in a function for this and the above if

nensidosari (approved):
Looks good, will do few testings around the new optional props, where they are used just to see if we can avoid any issues, and then hope for the best :D

sshanzel (approved):
Looks great

packages/shared/src/components/RecommendedTags.tsx
capJavert: Its actually better since when we turn on strict ts option it will scream if you try to access something that can be undefined. Only thing.. a lot of work to fix all the types.. 😆

packages/webapp/pages/feeds/[slugOrId]/edit.tsx
```diff
@@ -81,7 +81,8 @@ const EditFeedPage = (): ReactElement => {
   const { user } = useAuthContext();
   const { feedSettings } = useFeedSettings({ feedId, enabled: !!feedId });
   const [isPreviewFeedVisible, setPreviewFeedVisible] = useState(true);
-  const isPreviewFeedEnabled = feedSettings?.includeTags?.length >= 1;
+  const isPreviewFeedEnabled =
+    feedSettings?.includeTags && feedSettings.includeTags.length >= 1;
```
Comment (capJavert) [inline packages/webapp/pages/feeds/[slugOrId]/edit.tsx:85]: probably want to do `!!feedSettings?.includeTags` to keep `isPreviewFeedEnabled` boolean always

packages/webapp/pages/popup/notifications/enable.tsx
```diff
@@ -79,7 +79,7 @@ function Enable(): React.ReactElement {
   }, [isInitialized]);
 
   if (!isInitialized) {
-    return null;
+    return <></>;
```
capJavert, line 82: What error does it say here, `null` should be a valid thing to return?

capJavert (approved):
Some minor sanity checks, keep them coming 💪

packages/webapp/pages/feeds/[slugOrId]/edit.tsx
Comment (sshanzel) [reply packages/webapp/pages/feeds/[slugOrId]/edit.tsx:85]: Good catch, keeping it a boolean is a must, or else, the `shouldEvaluate` on `useConditionalFeature` would break because react-query only accepts a Boolean value and if not, it will throw an error and crash the app on prod. Don't ask me how I know 😭

packages/webapp/pages/account/invite.tsx
```diff
@@ -68,10 +68,10 @@ const AccountInvitePage = (): ReactElement => {
...
   );
   const users: UserShortProfile[] = useMemo(() => {
-    const list = [];
+    const list: UserShortProfile[] = [];
     usersResult.data?.pages.forEach((page) => {
       page?.referredUsers.edges.forEach(({ node }) => {
-        list.push(node);
+        list.push(node as UserShortProfile);
       });
     }, []);
```
ilasw, line 76: We are forcing node to be a `UserShortProfile` here but we can gentle make it works with something like this:

```ts
const users: UserShortProfile[] = useMemo(() => {
    return usersResult.data?.pages.flatMap((page) =>
      page?.referredUsers?.edges.map(({ node }) => {
        const {
          username,
          id,
          name,
          image,
          bio,
          createdAt,
          reputation,
          permalink,
        } = node;
        return {
          username,
          id,
          name,
          image,
          bio,
          createdAt,
          reputation,
          permalink,
        };
      }),
    );
  }, [usersResult]);
```

It's typesafe and we could create a transform utility from `PublicProfile` to `UserShortProfile` to make it consistent and reusable. What do you think about this approach?

PR #3353, opened 2024-07-15
refactor: bookmark reminder hook for analytics

sshanzel:
## Changes
- Thankfully we centralized all business logic for the bookmark reminder, adding the analytics was a breeze.
- Add two new events: `set bookmark reminder` and `remove bookmark reminder`.

## Events

| New | set bookmark reminder  | extra: { remind_in: `selected option` } |
| New | remove bookmark reminder  |  |

## Experiment

## Manual Testing

> [!CAUTION]

MI-438 #done

### Preview domain
https://mi-438.preview.app.daily.dev


packages/shared/src/hooks/notifications/useBookmarkReminder.ts
```diff
@@ -59,14 +62,15 @@ export const getRemindAt = (
   }
 };
 
-export const useBookmarkReminder = (): UseBookmarkReminder => {
+export const useBookmarkReminder = (post: Post): UseBookmarkReminder => {
```
capJavert, line 65: Let's do object syntax `{ post: Post }` as with other hooks.

packages/shared/src/hooks/notifications/useBookmarkReminder.ts
```diff
@@ -117,19 +121,34 @@ export const useBookmarkReminder = (): UseBookmarkReminder => {
...
+
       return onSetBookmarkReminder({
         ...props,
         remindAt: getRemindAt(now, preference),
       });
     },
-    [onSetBookmarkReminder],
+
+    // logEvent is known to be unstable
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
capJavert, line 137: Better to just put it on? It is not as bad as in `useEffect`.

packages/shared/src/hooks/notifications/useBookmarkReminder.ts
```diff
@@ -117,19 +121,34 @@ export const useBookmarkReminder = (): UseBookmarkReminder => {
...
     onRemoveReminder: useCallback(
-      (postId) => onUndoReminder({ postId, remindAt: null }),
-      [onUndoReminder],
+      (postId) => {
+        logEvent(postLogEvent(LogEvent.RemoveBookmarkReminder, post));
+
+        return onUndoReminder({ postId, remindAt: null });
+      },
+      // logEvent is known to be unstable
+      // eslint-disable-next-line react-hooks/exhaustive-deps
```
capJavert, line 150: same as above

packages/shared/src/hooks/notifications/useBookmarkReminder.spec.tsx
```diff
@@ -136,7 +137,7 @@ describe('useBookmarkReminder hook', () => {
   });
 
   it('should properly set to next Monday', async () => {
-    const { result } = renderHook(() => useBookmarkReminder(), {
+    const { result } = renderHook(() => useBookmarkReminder(post), {
```
capJavert, line 140: We usually add tests for `trackEvent` so if you had time it would be cool to add 🙏 (non blocking)

capJavert (commented):
Some minor comments, overall smooth implementation indeed 👌

packages/shared/src/hooks/notifications/useBookmarkReminder.ts
sshanzel: Right. Yeah, we can do that.

packages/shared/src/hooks/notifications/useBookmarkReminder.spec.tsx
sshanzel: Oh, right. Yeah, I think it is a good practice. Let's add that.

packages/shared/src/hooks/notifications/useBookmarkReminder.ts
sshanzel: Oh, yep. Totally forgot about that.

capJavert: Yeah, easier to add additional props if we need to 👌

sshanzel:
@capJavert added some events to see what props are sent to our analytics context. Also realized there was no test for removing the reminder so added it 😅

capJavert (approved):
I usually just use `LogContext.Provider`, but this works too, tnx for adding 👌

sshanzel:
> I usually just use `LogContext.Provider`, but this works too, tnx for adding 👌

Yeah, just so it will be easier on the next ones 🚀

packages/shared/src/hooks/notifications/useBookmarkReminder.ts
```diff
@@ -59,14 +62,21 @@ export const getRemindAt = (
   }
 };
 
-export const useBookmarkReminder = (): UseBookmarkReminder => {
+interface UseBookmarkReminderProps {
+  post: Post;
```
rebelchris, line 66: Are you sure this works well?
I just vaguely remember it being an issue from somewhere (maybe modal?) or share. 
Where we only get access to the post on click not when initializing the hook.

Eg does something like `onBookmark(passPostHere)`

rebelchris (approved):
Looks good, maybe just sanity check if this hook doesn't affect anything negatively 😂

packages/shared/src/hooks/notifications/useBookmarkReminder.ts
sshanzel: Yes, this should work, I think that can happen on the context menu in extreme cases, but since we don't send the reminder here through the context menu (we open a modal when clicked), we should be safe. Other areas we can set a reminder is through the card and modal which does not disappear when action is made.

sshanzel:
> Looks good, maybe just sanity check if this hook doesn't affect anything negatively 😂

Verified anon pages seem to render still 😂 😭

nensidosari (approved):
looks great

sshanzel:
@idoshamun for your final review about the analytics involved 👀

idoshamun (approved):
Fantastic!

PR #3354, opened 2024-07-16
refactor: usage of isloading within the same mutation

sshanzel:
## Changes
- The primary issue was the upvote not working.
- To temporarily fix, we use the state instead.
- After further investigation, there seems to be a race condition when using the `isLoading` from the same mutation. Since we are using the `isLoading` from the same mutation, instantly after calling it (before reaching the first line of the function), the state has already updated that `isLoading` is `true` in turn, it just returns early on.
- Introduced a wrapper function to handle the `isloading` beforehand.

## Events

## Experiment

## Manual Testing

> [!CAUTION]

### Preview domain
https://mi-436.preview.app.daily.dev

### DIFF
--- a/packages/shared/src/hooks/vote/useVote.ts
+++ b/packages/shared/src/hooks/vote/useVote.ts
@@ -1,4 +1,4 @@
-import { useContext, useCallback, useState } from 'react';
+import { useContext, useCallback } from 'react';
 import { useMutation } from '@tanstack/react-query';
 import AuthContext from '../../contexts/AuthContext';
 import { UserVote } from '../../graphql/posts';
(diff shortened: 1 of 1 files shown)


packages/shared/src/hooks/vote/useVote.ts
```diff
@@ -40,11 +33,21 @@ const useVote = ({ onMutate, entity, variables }: UseVoteProps): UseVote => {
...
+  const voteEntity: typeof onVoteEntity = useCallback(
+    (props) => {
+      if (isLoading) {
+        return Promise.resolve(null);
+      }
+
+      return onVoteEntity(props);
+    },
+    [isLoading, onVoteEntity],
+  );
```
nensidosari, line 49: Just wondering to avoid anything like this in other places would it make sense to have any wrapper function for useMutation that would have this wrapper by default for mutateAsync? 

Of course changing that in all the places and making sure they work is a lot and not sure worth it, but we can consider in case we ever are faced with the same issue again

nensidosari (approved):
Looks valid, was looking to test this on preview but it leads me to 404.

capJavert (approved):
Interesting edge case, I would not apply it globally since its not needed when not using variables from useMutation directly inside mutationFn but something to keep in mind when we have to do similar stuff in the future.

Nice find!

rebelchris (approved):
But in this case isn't the state resolver we put in just as good?
Just want to avoid this still having some edge-case we might have missed.

Otherwise if you feel this is better, feel free to merge.
Just make extra sure to test it out with actually seeing data go all way through.

sshanzel:
> But in this case isn't the state resolver we put in just as good?

Setting a state produces a re-render, even though the mutation would likely still send a re-render, probably better to be one less usage of state.

packages/shared/src/hooks/vote/useVote.ts
sshanzel: I am still not in the case where I think we should wrap everything with this. We don't often do this as we normally prioritize optimistic UI. The root of us doing this was a deeper issue that happens if you spam the upvote.

PR #3360, opened 2024-07-16
feat: add "notify button" experiment to sources

ilasw:
If you want to test locally have to pull from this: https://github.com/dailydotdev/daily-api/pull/2070

## Changes
- Added `BellAddIcon`;
- Added follow button in `SourceSubscribeButton`;
- Refactor "block" button on source page, only have one source of "actions" (UI change if follow, hiding Block)
- Added tracking events
- Added toast notifications
- Add `follow/unfollow` mutation;
- Add `withExperiment` logic;

Todo:
- Add confirmation box for notifications;
- Try to add to StoryBook;

## Events

## Experiment

## Manual Testing

> [!CAUTION]

### Preview domain
https://mi-422-micro-notify-button-step-source-subscription.preview.app.daily.dev


packages/shared/src/lib/element.ts
```diff
@@ -60,6 +60,6 @@ export const parentClassContains = (
 };
 
 export const isCompanionActivated = (): boolean =>
-  !!document
```
ilasw, line 63: This caused a lot of errors in dev environment

packages/shared/src/components/widgets/PostUsersHighlights.tsx
```diff
@@ -31,18 +32,14 @@ enum UserType {
...
-interface SourceAuthorProps {
-  id?: string;
-  handle?: string;
-  image: string;
-  name: string;
-  username?: string;
-  userType?: UserType;
-  permalink: string;
-  reputation?: number;
-}
```
ilasw, line 43: Needed to switch from `interface` to `type` for managing cases with dynamic props.

packages/shared/src/components/widgets/PostUsersHighlights.tsx
```diff
@@ -106,18 +103,19 @@ const Image = (props: ImageProps) => {
...
-    userType = UserType.Source,
-    reputation,
     allowSubscribe = true,
     className,
     showReputation = false,
-  } = props;
+  } = source;
+
+  const typeHandle = 'handle' in source ? source.handle : source.username;
+  const reputation = 'reputation' in source ? source.reputation : NaN;
```
ilasw, line 117: Those props do not exist in every prop type so need for initial check before initialize

packages/shared/src/components/widgets/PostUsersHighlights.tsx
```diff
@@ -194,38 +192,41 @@ export const UserHighlight = (props: UserHighlightProps): ReactElement => {
...
-        <SourceSubscribeButton
-          className="ml-2"
-          variant={ButtonVariant.Secondary}
-          source={{ id }}
-        />
-      )}
+      {!isSourceBlocked &&
+        isUserTypeSource &&
+        allowSubscribe &&
+        'handle' in source && (
```
ilasw, line 211: Need to extra check if is a Source using `handle` as key prop

packages/shared/src/hooks/useMutateFilters.ts
```diff
@@ -163,17 +165,54 @@ export default function useMutateFilters(
...
+    () => void
+  >(
+    ({ source, key = 'excludeSources' }) =>
+      gqlClient.request(REMOVE_FILTERS_FROM_FEED_MUTATION, {
+        filters: {
+          [key]: [source.id],
+        },
+      }),
+    {
+      // onMutate: onUnblockSource,
```
rebelchris, line 181: This is missed?

packages/shared/src/hooks/useTagAndSource.ts
```diff
@@ -199,18 +203,80 @@ export default function useTagAndSource({
...
 
+  const onUnblockSource = useCallback(
+    async ({ source, requireLogin }: SourceActionArguments) => {
+      if (shouldShowLogin(requireLogin)) {
+        showLogin({ trigger: origin as AuthTriggersType });
+        return { successful: false };
+      }
+      logEvent({
+        event_name: 'unblock',
+        target_type: 'source',
```
rebelchris, line 214: These should be types enums somewhere already existing something like `EventName.{something}`

packages/shared/src/lib/element.ts
rebelchris: I rather not move this since as it has nothing to do with the PR in general 🙏

rebelchris (commented):
Wow love the fact it got abstracted so nicely.
Sorry only had time for a shallow review so will trust the rest of web team to test it over.

One thing I couldn't see clearly did we manage the change that blocking if you notifying will also unfollow/unnotify? in one go for control?

packages/shared/src/hooks/useMutateFilters.ts
ilasw: This is the generic function I've written for refactoring and do not repeat the mutation over follow/notify/etc...
I'm removing the comment! :)

packages/shared/src/hooks/source/useSourceActionsNotify.ts
```diff
@@ -9,31 +9,32 @@ import { useNotificationPreferenceToggle } from '../notifications';
...
 
 export type UseSourceSubscription = {
-  isSubscribed: boolean;
+  haveNotifications: boolean;
   isReady: boolean;
-  onSubscribe: () => Promise<void>;
+  onNotify: () => Promise<void>;
 };
 
-export const useSourceSubscription = ({
```
capJavert, line 21: Is it really needed the name change, introduces a lot of changes to the code I think not really needed? Also its experiment which means if it fails we will revert this so better to keep changes minimal until we know its a win. Interested also to hear what others think.

packages/webapp/pages/sources/index.tsx
```diff
@@ -48,6 +51,7 @@ const SourceTopList = ({
           <ListItem key={item.id} index={i + 1} href={item.permalink}>
             <UserHighlight
               {...item}
+              userType={UserType.Source}
```
capJavert, line 54: This does not come from item?

packages/shared/src/lib/log.ts
```diff
@@ -149,9 +149,20 @@ export enum LogEvent {
   ScheduleStreakReminder = 'schedule streak reminder',
   // 404 page
   View404Page = '404 page',
-  // source
+  // Source Actions - start
+  BlockSource = 'block source',
+  UnblockSource = 'unblock source',
```
capJavert, line 154: Just verify with Ido if this is ok so it can be added to taxonomy sheet (also with other events related changes)

packages/shared/src/components/PostOptionsMenu.tsx
```diff
@@ -350,13 +350,17 @@ export default function PostOptionsMenu({
...
         />
       ),
       label: `${
-        sourceSubscribe.isSubscribed ? 'Unsubscribe from' : 'Subscribe to'
+        sourceSubscribe.haveNotifications
+          ? 'Receive notifications from'
+          : 'Receive notifications to'
       } ${post?.source?.name}`,
-      action: sourceSubscribe.isReady ? sourceSubscribe.onSubscribe : undefined,
+      action: sourceSubscribe.isReady ? sourceSubscribe.onNotify : undefined,
```
capJavert, line 363: Should this and related changes also be behind experiment? To preserve control copy.

packages/shared/src/components/widgets/PostUsersHighlights.tsx
```diff
@@ -31,18 +33,16 @@ enum UserType {
...
+  | ({ userType?: null } & CommentAuthor)
+  | ({
+      userType: UserType.Featured | UserType.Author | UserType.Scout;
+    } & CommentAuthor)
+  | ({
+      userType: UserType.Source;
+    } & ISource);
 
-interface UserHighlightProps extends SourceAuthorProps {
+type UserHighlightProps = SourceAuthorProps & {
```
capJavert, line 45: As Chris said, if there is no meaningful need for this change let's remove since it adds a mental overload when looking at changed code in the PR.

packages/shared/src/components/widgets/PostUsersHighlights.tsx
```diff
@@ -172,19 +173,16 @@ export const UserHighlight = (props: UserHighlightProps): ReactElement => {
...
             <ProfileLink
-              className={classNames(
-                'truncate font-bold typo-callout',
-                className?.name,
-              )}
+              className={classNames('font-bold typo-callout', className?.name)}
               href={permalink}
             >
-              {name}
+              <TruncateText>{name}</TruncateText>
```
capJavert, line 185: Please just double check manually this truncates correctly, because of lot of wrappers this component is known to break and clip in a weird ways. 🙏

packages/shared/src/hooks/useMutateFilters.ts
```diff
@@ -423,27 +515,19 @@ export default function useMutateFilters(
...
-    () => Promise<void>
-  >(
-    ({ source }) =>
-      gqlClient.request(ADD_FILTERS_TO_FEED_MUTATION, {
-        filters: {
-          excludeSources: [source.id],
+  const blockSourceRemote = useCallback(
+    async ({ source }: SourceMutationProps) => {
+      await addFiltersRemote(
+        { source, key: 'excludeSources' },
```
capJavert, line 521: Can we make this static and two mutations? The `key` pattern seems a bit fragile and also not typed.

capJavert (commented):
Overall I think implementation is good but a lot of unrelated changes/refactors add a lot of mental overhead when reviewing. While I appreciate some of those refactors I think we should remove that additional code/refactor from this PR. Additional reason for it being this is experiment so if it fails we again have to revert a lot of code.

packages/shared/src/components/widgets/PostUsersHighlights.tsx
ilasw: @capJavert I need this change because `block` action need more than the `id` and there is no way to have this data not changing this prop and how it behave :(

packages/shared/src/components/widgets/PostUsersHighlights.tsx
ilasw: It works! 
image

packages/shared/src/components/PostOptionsMenu.tsx
ilasw: Yes, I've asked for updates on this component [here](https://dailydotdev.slack.com/archives/C07ATUPBPJ9/p1721396757912089) and they have updated the UI early today. I've to update this!

packages/webapp/pages/sources/index.tsx
ilasw: Not really, removing this line make the component broke for ts check :(

packages/shared/src/hooks/source/useSourceActionsNotify.ts
ilasw: I've changed the name everywhere after [this slack thread](https://dailydotdev.slack.com/archives/C02E2C3C13R/p1721205082828849?thread_ts=1721204862.202729&cid=C02E2C3C13R)

packages/shared/src/components/widgets/PostUsersHighlights.tsx
capJavert: but the `interface` to `type` change is not needed right?

capJavert: Missed this comment from all the changes https://github.com/dailydotdev/apps/pull/3360/files#r1682862007, my mistake.

packages/webapp/pages/sources/index.tsx
capJavert: `userType` was optional before and `UserHighlight` had it default to `UserType.Source`, I am actually more in favour to keeping this 👍

packages/shared/src/hooks/useMutateFilters.ts
ilasw: @capJavert updated ✅

nensidosari (commented):
Great work Luca, like the refactoring, but can say definitely a really hard PR to review 😢 Makes it difficult to understand what are the changes only refactoring and what are the actually changes for the experiment.

Since both Chris and Ante have mentioned, would you be up for splitting the PR into 2? One for all the refactoring of current code, another one with additions for the experiment?
That would make it easier for a last review to make sure the changes are correct, easier to test as we merge both PR's one at a time, and easier to cleanup if needed as mentioned if experiment looses. 

@ilasw what do u think? would that be a lot of work?

packages/shared/src/hooks/useMutateFilters.ts
capJavert: Nice!

packages/shared/src/hooks/source/useSourceActionsNotify.ts
capJavert: That was my logic yes 👍

packages/shared/src/hooks/source/useSourceActionsBlock.ts
```diff
@@ -22,10 +22,11 @@ export function useSourceActionsBlock(
   const { displayToast } = useToastNotification();
 
   const isBlocked = useMemo(() => {
-    if (!feedSettings) {
+    if (!feedSettings?.excludeSources?.length) {
       return false;
     }
-    return !!feedSettings.excludeSources?.find(
+    console.log(feedSettings.excludeSources);
```
nensidosari, line 28: Probably not ment to be pushed :D

PR #3368, opened 2024-07-17
feat: auth modal tracking

capJavert:
## Changes

- add tracking pixel hook for signup to work on `AuthModal` (and not only `/onboarding`

## Events

## Experiment

## Manual Testing

> [!CAUTION]

### Preview domain
https://auth-modal-tracking.preview.app.daily.dev


packages/shared/src/components/auth/AuthModal.tsx
```diff
@@ -45,6 +46,12 @@ export default function AuthModal({
     closeLogin();
   };
 
+  const onSuccessfulRegistration = (newUser?: LoggedUser | AnonymousUser) => {
+    loginState?.onRegistrationSuccess?.(newUser);
+
+    logSignUp({ experienceLevel: (newUser as LoggedUser)?.experienceLevel });
```
capJavert, line 52: I added it globally to all calls but technically inside `logSignUp` method there is a check if pixel tracking is intialized so for example it won't fire from post page. I think thats ok @idoshamun @rebelchris ?

packages/shared/src/components/profile/SquadsList.tsx
```diff
@@ -60,8 +60,8 @@ function SquadItem({
       showLogin({
         trigger: AuthTriggers.JoinSquad,
         options: {
-          onLoginSuccess: joinSquad,
-          onRegistrationSuccess: joinSquad,
+          onLoginSuccess: () => joinSquad(),
```
capJavert, line 63: Function was always sending `(user) => unknown` but now TS was complaining when I updated correct type in `AuthModal`

nensidosari: Do we need the return anywhere? if not, maybe instead we can remove the return

capJavert: Wanted to keep it as is that if we need to return we don't have to change all implementations again. Its not really doing anything.

rebelchris (approved):
Looks ok to me, but let's wait for @idoshamun final say if it's ok

idoshamun (approved):

PR #3371, opened 2024-07-18
feat: add admin rep for public squad treshold

rebelchris:
## Changes

Add squad admin reputation of 100 to the public squad requirement.

| Status | Card | List | Settings | Checklist |
|--------|--------|--------|--------|--------|
| Open | Screenshot 2024-07-18 at 09 56 50 | Screenshot 2024-07-18 at 09 57 07 | Screenshot 2024-07-18 at 09 57 22 | Screenshot 2024-07-18 at 09 56 55 |
| Ready | Screenshot 2024-07-18 at 09 59 18 | Screenshot 2024-07-18 at 09 59 10  | Screenshot 2024-07-18 at 09 59 02 | Screenshot 2024-07-18 at 09 59 32 | 

## Events

## Experiment

## Manual Testing

> [!CAUTION]

AS-445 #done 

### Preview domain
https://as-445-public-squad-admin-rep.preview.app.daily.dev


packages/shared/src/components/ProfilePicture.tsx
```diff
@@ -7,6 +7,7 @@ import { Image, ImageType } from './image/Image';
 import { useRequestProtocol } from '../hooks/useRequestProtocol';
 
 export enum ProfileImageSize {
+  Size16 = 'size16',
```
nensidosari, line 10: 🙈 this didn't fit in any of the ones we have already defined?

packages/shared/src/components/ProfilePicture.tsx
```diff
@@ -32,6 +33,7 @@ export interface ProfilePictureProps
 }
 
 export const sizeClasses: Record<ProfileImageSize, string> = {
+  size16: 'size-4',
```
nensidosari, line 36: Maybe xxSmall to keep the same convention we have so far?

packages/shared/src/components/fields/ProgressBar.tsx
```diff
@@ -35,8 +36,9 @@ export function ProgressBar({
     >
       <meter
         className={classNames(
-          'appearance-none bg-accent-cabbage-default transition-[width]',
+          'appearance-none transition-[width]',
           className?.bar,
+          className?.barColor ?? 'bg-accent-cabbage-default',
```
nensidosari, line 41: we should really really consider that tailwind merge, this wouldn't be necessary

nensidosari (approved):
Seems good to me, couldn't test though in preview, link points to 404

packages/shared/src/components/ProfilePicture.tsx
rebelchris: Nope super tiny format this one.

packages/shared/src/components/fields/ProgressBar.tsx
rebelchris: This actually fixes it as a neater solution I think.
If the background is variable (which it is) it should not be defined on the top level. (or should be a prop)

TW merge is technically just another variant of !important but on top level.
Not bad, but think where possible you should still avoid it if you can abstract things away.

packages/shared/src/components/ProfilePicture.tsx
nensidosari: yeah it's clearer this way, just until then it becomes a mess to have both.
Just wondering, since this is particular to this use case, wouldn't an override - custom size be better then defining?

This is more general discussions, since I see the sizing defined which should be finite, just grows with every usecase, but not a blocker for the PR.

packages/shared/src/components/fields/ProgressBar.tsx
nensidosari: If we're accepting just a set of colors, this or that or that, then yes a prop makes more sense. 
How it's done at the moment where every possible color set as a classname, I think `className?.bar` would have been more then enough with tailwind merge or !important.

PR #3372, opened 2024-07-18
feat: update welcome page feed variables

capJavert:
## Changes

`version` & `ranking` were missing

## Events

## Experiment

## Manual Testing

> [!CAUTION]

### Preview domain
https://welcome-page-feed-variables.preview.app.daily.dev

### DIFF
--- a/packages/webapp/pages/welcome/index.tsx
+++ b/packages/webapp/pages/welcome/index.tsx
@@ -1,6 +1,9 @@
 import React, { ReactElement, useEffect, useState } from 'react';
 import Feed, { FeedProps } from '@dailydotdev/shared/src/components/Feed';
-import { ANONYMOUS_FEED_QUERY } from '@dailydotdev/shared/src/graphql/feed';
+import {
+  ANONYMOUS_FEED_QUERY,
+  RankingAlgorithm,
+} from '@dailydotdev/shared/src/graphql/feed';
 import {
   OtherFeedPage,
   StaleTime,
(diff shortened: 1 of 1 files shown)


idoshamun (approved):

packages/webapp/pages/welcome/index.tsx
```diff
@@ -63,6 +78,10 @@ const DemoPage = (): ReactElement => {
     options: {
       staleTime: StaleTime.Default,
     },
+    variables: {
+      version: isDevelopment ? 1 : feedVersion,
```
rebelchris, line 82: Actually reminds me we should really fix this 😂

packages/webapp/pages/welcome/index.tsx
```diff
@@ -124,7 +143,7 @@ const DemoPage = (): ReactElement => {
           what&apos;s out there.
         </p>
       </div>
-      <Feed className={feedProps.className} {...feedProps} />
+      {featuresReady && <Feed className={feedProps.className} {...feedProps} />}
```
rebelchris, line 146: Does this delay rendering? Because placeholders only happen inside here

rebelchris (approved):
Probably nothing blocking

packages/webapp/pages/welcome/index.tsx
capJavert: A bit, but not that much.

capJavert: Placeholder shows delayed either way since it is not in the first pain in SSR, but did not want to look at why the useFeed does that in this PR.

capJavert: `useFeed` placeholder is delayed due to `tokenRefreshed` condition on `enabled` but I don't want to remove it.

PR #3373, opened 2024-07-19
feat: content type onboarding

nensidosari:
## Changes
- new content type section for onboarding
- new custom checkbox (got from last experiment with list card view on onboarding)
- refactor: move edit tag section in new file
- refactor: move onboarding inner sections in a new directory
- refactor: moved all logic that needed to be duplicated from advanced settings useAdvancedSettings
- added more logic to create feed button as it needs to change depending on the step (extracted some parts only for content type in a helpers function for separation)
- all logic of showing content types in onboarding is behind flag

## Events

## Experiment

Did you introduce any new experiments? Yes

https://dailydotdev.slack.com/archives/C02JAUF8HJL/p1721369865457159

## Manual Testing

> [!CAUTION]

- [x] MobileL (420px)
- [x] Tablet (656px)
- [x] Laptop (1020px)

I've tested variations with and without content types in both mobile and desktop

MI-441 #done

### Preview domain
https://mi-441-content-type-onboarding.preview.app.daily.dev


packages/shared/src/components/auth/OnboardingSteps/ContentTypes/CustomCheckbox.tsx
```diff
@@ -0,0 +1,47 @@
...
+import { Checkbox } from '../../../fields/Checkbox';
+
+interface CustomCheckboxProps {
+  checked: boolean;
+  title: string;
+  description: string;
+  onCheckboxToggle: () => void;
+  name: string;
+}
+export const CustomCheckbox = ({
```
nensidosari, line 13: Styling kept almost the same as the checkbox created for layout experiment in onboarding

packages/shared/src/components/auth/OnboardingSteps/EditTag.tsx
nensidosari, line 1: Just moved this to a separate file to make onboarding.tsx a bit better

packages/shared/src/components/auth/OnboardingSteps/ReadingReminder.tsx
nensidosari, line 1: moved in a directory for all onboarding steps

packages/shared/src/components/filters/AdvancedSettings.tsx
```diff
@@ -1,58 +1,23 @@
...
-import { AdvancedSettingsGroup } from '../../graphql/feedSettings';
 import { useAdvancedSettings } from '../../hooks/feed';
-import useTagAndSource from '../../hooks/useTagAndSource';
-import { Origin } from '../../lib/log';
-import { Source } from '../../graphql/sources';
 
 const ADVANCED_SETTINGS_KEY = 'advancedSettings';
 
 function AdvancedSettingsFilter(): ReactElement {
-  const { advancedSettings, isLoading, feedSettings } = useFeedSettings();
```
nensidosari, line 13: All code here was moved inside the useFeedSettings hook and useAdvancedSettings without being changed so it could be easily reused in content type onboarding

packages/shared/src/components/filters/ContentTypesFilter.tsx
```diff
@@ -4,27 +4,24 @@ import { useAdvancedSettings } from '../../hooks/feed';
 import useFeedSettings from '../../hooks/useFeedSettings';
 
 export function ContentTypesFilter(): ReactElement {
-  const { advancedSettings, isLoading } = useFeedSettings();
+  const { videoSetting, isLoading } = useFeedSettings();
```
nensidosari, line 7: Just moved inside useFeedSettings so the logic could be reused in content type onboarding

packages/shared/src/hooks/feed/useAdvancedSettings.ts
```diff
@@ -61,5 +66,34 @@ export const useAdvancedSettings = (
     [alerts?.filter, selectedSettings, onUpdateSettings, updateAlerts, user],
   );
 
-  return { selectedSettings, onToggleSettings, onUpdateSettings };
+  const checkSourceBlocked = useCallback(
```
nensidosari, line 69: All existing logic just moved from advanced settings

packages/shared/src/hooks/useFeedSettings.ts
```diff
@@ -80,12 +84,39 @@ export default function useFeedSettings({
     [advancedSettings, feedSettings],
   );
 
+  const contentSourceList = useMemo(
```
nensidosari, line 87: All existing logic just moved from advanced settings

packages/shared/src/lib/featureManagement.ts
```diff
@@ -29,6 +29,7 @@ const feature = {
   bookmarkReminder: new Feature('bookmark_reminder', false),
   animatedUpvote: new Feature('animated_upvote', false),
   mobileExploreTab: new Feature('mobile_explore_tab', false),
+  onboardingContentType: new Feature('onboarding_content_type', false),
```
nensidosari, line 32: In growthbook https://app.growthbook.io/features/onboarding_content_type and already enabled for team

packages/webapp/pages/onboarding.tsx
```diff
@@ -159,6 +161,19 @@ export function OnboardPage(): ReactElement {
   };
 
   const onClickCreateFeed = () => {
+    if (activeScreen === OnboardingStep.EditTag) {
+      const onboardingContentType = getFeatureValue(
+        feature.onboardingContentType,
+      );
```
nensidosari, line 167: Enrolment only happens when user is in edit tag page and has clicked the continue button

packages/shared/src/components/auth/OnboardingSteps/ContentTypes/ContentTypes.tsx
```diff
@@ -0,0 +1,49 @@
...
+  const {
+    selectedSettings,
+    onToggleSettings,
+    checkSourceBlocked,
+    onToggleSource,
+  } = useAdvancedSettings();
+  const { advancedSettingsExceptCommunity } = useContentTypesOnboarding();
+
+  return (
+    <div className="flex max-w-[63.75rem] flex-col tablet:px-10">
```
rebelchris, line 19: Where does this max-w come from? feels a bit odd

packages/shared/src/components/auth/OnboardingSteps/ContentTypes/useContentTypesOnboarding.tsx
```diff
@@ -0,0 +1,36 @@
+// TODO: move in right directory if the onboardingContentType experiment is successful
```
rebelchris, line 1: What would be the right directory?

packages/shared/src/components/auth/OnboardingSteps/ContentTypes/ContentTypes.tsx
```diff
@@ -0,0 +1,49 @@
+import React, { ReactElement } from 'react';
+import useFeedSettings from '../../../../hooks/useFeedSettings';
+
+import { useAdvancedSettings } from '../../../../hooks';
+import { CustomCheckbox } from './CustomCheckbox';
+import { useContentTypesOnboarding } from './useContentTypesOnboarding';
+
+export const ContentTypes = (): ReactElement => {
+  const { contentSourceList } = useFeedSettings();
```
rebelchris, line 9: Why don't you get this all from the same hook?
(EG how `AdvancedSettings` does it now?

Just find the naming a bit odd "expectCommunity" what happens if tomorrow we add another `ContentSource`.
Feel it might be better if we expose `ContentCurationList` and `ContentSourceList` from somewhere in one go?

(Let me know what you think)

packages/shared/src/components/filters/AdvancedSettings.tsx
rebelchris: I like this export more, why don't we use this in the `ContentTypes` this makes more sense to me actually.9

packages/shared/src/components/auth/OnboardingSteps/ContentTypes/ContentTypes.tsx
```diff
@@ -0,0 +1,49 @@
...
+  const {
+    selectedSettings,
+    onToggleSettings,
+    checkSourceBlocked,
+    onToggleSource,
+  } = useAdvancedSettings();
+  const { advancedSettingsExceptCommunity } = useContentTypesOnboarding();
```
rebelchris, line 16: I'm just confused why we need this, you can just use `contentCurationList` from the above hook right?

packages/webapp/pages/onboarding.tsx
```diff
@@ -124,6 +113,8 @@ export function OnboardPage(): ReactElement {
   const targetId: string = ExperimentWinner.OnboardingV4;
   const formRef = useRef<HTMLFormElement>();
   const [activeScreen, setActiveScreen] = useState(OnboardingStep.Intro);
+  const shouldEnrollContentType = useRef(false);
```
rebelchris, line 116: The name is a bit confusing here as this state has nothing to do with enrolment.
It should be something like `shouldShowContentTypeScreen` maybe?

packages/webapp/pages/onboarding.tsx
```diff
@@ -159,6 +161,19 @@ export function OnboardPage(): ReactElement {
...
+        feature.onboardingContentType,
+      );
+
+      if (onboardingContentType) {
+        shouldEnrollContentType.current = true;
+        return onClickNext();
+      }
+    } else if (activeScreen === OnboardingStep.ContentTypes) {
+      shouldEnrollContentType.current = false;
+    }
```
rebelchris, line 175: I think you might want this inside line 132 right?
This could be some extreme edge-case where it re-renders and sets to false without ever showing it.
Think it's safer to do it when it shows to reset.

rebelchris (commented):
Nice work!
Some minor comments, but don't feel blocked by me please 🙏
(The only "concern" would be the enrolment race condition of setting the ref to false there)

packages/shared/src/components/filters/AdvancedSettings.tsx
nensidosari: This is still exported and can be used for sure, just wanted to make sure to have the same behaviour so thought of exporting from here.
But makes sense, this is a bit more clear, just for not having duplicate code, thinking maybe we can do the filterings and what is needed in any common function

packages/shared/src/components/auth/OnboardingSteps/ContentTypes/ContentTypes.tsx
nensidosari: So the box does not expand more then it should in big screens. it is a set width in figma

packages/shared/src/components/auth/OnboardingSteps/ContentTypes/useContentTypesOnboarding.tsx
nensidosari: Well this is a good place, unless we'll use this in other places in the future where hooks directory is the best place. 
Thinking this doesn't have to be a hook really, just a utils function for content type should be enough

packages/shared/src/components/auth/OnboardingSteps/ContentTypes/ContentTypes.tsx
nensidosari: Seemed like the right approach on my head, but let me give another look to the whole thing to see what makes more sense. Definitely what u're saying makes sense

packages/shared/src/components/filters/AdvancedSettings.tsx
nensidosari: Got back to this export for useFeedSettings

packages/shared/src/components/auth/OnboardingSteps/ContentTypes/useContentTypesOnboarding.tsx
nensidosari: removed the hook at all, felt unnecessary, instead created a helper function for separating functionality

packages/shared/src/components/auth/OnboardingSteps/ContentTypes/ContentTypes.tsx
nensidosari: Made a helper function to get the lists based on advanced settings

packages/shared/src/components/auth/OnboardingSteps/ContentTypes/ContentTypes.tsx
nensidosari: yes, it's removed

packages/webapp/pages/onboarding.tsx
nensidosari: renamed to enableContentTypeStep

packages/webapp/pages/onboarding.tsx
nensidosari: removed the move to false, since we don't really need it.

omBratteng:
PR title

nensidosari:
@omBratteng fixed

packages/shared/src/components/auth/OnboardingSteps/ContentTypes/CustomCheckbox.tsx
```diff
@@ -0,0 +1,47 @@
...
+  onCheckboxToggle,
+  name,
+}: CustomCheckboxProps): ReactElement => {
+  return (
+    <button
+      type="button"
+      className={classNames(
+        'h-[8.25rem] max-w-80 cursor-pointer rounded-16 border p-4 text-left',
+        checked
+          ? 'border-accent-cabbage-bolder bg-surface-float text-text-primary'
```
omBratteng, line 26: In the design, the border (and checkboxes) use a gradient instead of solid colour. Is this intended to differ?

packages/shared/src/components/auth/OnboardingSteps/ContentTypes/helpers.ts
```diff
@@ -0,0 +1,47 @@
...
+  const contentCurationList = getContentCurationList(advancedSettings);
+
+  const videoSetting = getVideoSetting(advancedSettings);
+
+  const advancedSettingsSelected = (settings: AdvancedSettings[]) =>
+    !!settings
+      .map(({ id, defaultEnabledState }) => {
+        return selectedSettings[id] ?? defaultEnabledState;
+      })
+      .find((setting) => setting === true);
```
sshanzel, line 30: I think we can simplify this and avoid the `!!` by using some.
```suggestion
    settings
      .map(({ id, defaultEnabledState }) => {
        return selectedSettings[id] ?? defaultEnabledState;
      })
      .some(Boolean);
```

packages/shared/src/components/auth/OnboardingSteps/ContentTypes/helpers.ts
```diff
@@ -0,0 +1,47 @@
...
+
+  const advancedSettingsCurationListSelected =
+    advancedSettingsSelected(contentCurationList);
+
+  const advancedSettingsVideoSelected =
+    !!videoSetting && advancedSettingsSelected([videoSetting]);
+
+  const sourceListSelected = !!contentSourceList
+    .map(({ options }) => options.source)
+    .find((source) => !checkSourceBlocked(source));
```
sshanzel, line 40: Same goes here, nothing blocking.
```suggestion
  const sourceListSelected = contentSourceList
    .map(({ options }) => options.source)
    .some((source) => !checkSourceBlocked(source));
```

packages/shared/src/components/auth/OnboardingSteps/ContentTypes/ContentTypes.tsx
rebelchris: I think we can use max-w-laptop then right? (Or something like that)

rebelchris (approved):

packages/shared/src/components/auth/OnboardingSteps/ContentTypes/ContentTypes.tsx
nensidosari: oh yeah true

packages/shared/src/components/auth/OnboardingSteps/ContentTypes/CustomCheckbox.tsx
```diff
@@ -0,0 +1,47 @@
...
+      <div className="h-full">
+        <div className="flex">
+          <p className="mr-auto font-bold typo-title3">{title}</p>
+          {checked && (
+            <Checkbox
+              className="pointer-events-none !p-0"
+              checkmarkClassName="!mr-0"
+              name={name}
+              checked
+            />
```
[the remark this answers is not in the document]
nensidosari: Good point, adding this to the cleanup ticket together with the gradient if experiment wins.

packages/shared/src/components/auth/OnboardingSteps/ContentTypes/CustomCheckbox.tsx
nensidosari: Because this component was created and accepted in a very recent implementation for layout on onboarding, decided to go this way without the gradient. Once this experiment wins will add the gradient for sure, just not to put in extra work now.

packages/shared/src/components/auth/OnboardingSteps/ContentTypes/CustomCheckbox.tsx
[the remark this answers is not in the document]
nensidosari: https://dailydotdev.atlassian.net/browse/MI-451

packages/shared/src/components/auth/OnboardingSteps/ContentTypes/CustomCheckbox.tsx
nensidosari: https://dailydotdev.atlassian.net/browse/MI-451

packages/shared/src/components/auth/OnboardingSteps/ContentTypes/CustomCheckbox.tsx
```diff
@@ -0,0 +1,47 @@
...
+      type="button"
+      className={classNames(
+        'h-[8.25rem] max-w-80 cursor-pointer rounded-16 border p-4 text-left',
+        checked
+          ? 'border-accent-cabbage-bolder bg-surface-float text-text-primary'
+          : 'text-text-tertiary',
+      )}
+      onClick={onCheckboxToggle}
+    >
+      <div className="h-full">
```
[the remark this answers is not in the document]
nensidosari: not needed

PR #3374, opened 2024-07-19
fix: toast timer cleanup

sshanzel:
## Changes
- While working on reading reminders, I noticed an issue on our toast.
- Interval wasn't getting cleaned up when setting a new one.
- There would be a point where toast won't even be readable to users.

Before:

https://github.com/user-attachments/assets/0a9ae19e-9921-4692-90a0-f83eadfee6d4

After:

https://github.com/user-attachments/assets/1c586559-a881-4d97-b7eb-5ead42b5a0a0

## Events

## Experiment

## Manual Testing

> [!CAUTION]

### Preview domain
https://fix-toast-timer.preview.app.daily.dev


capJavert (approved):
Makes sense to me 👍

packages/shared/src/hooks/useTimedAnimation.ts
```diff
@@ -47,6 +47,10 @@ export const useTimedAnimation = ({
 
   const startAnimation = useCallback(
     (duration: number) => {
+      if (interval.current) {
+        window.clearInterval(interval.current);
+      }
```
capJavert, line 52: Would you maybe use `clearInterval` function we already have for this?

rebelchris (approved):
Oh nice find!
Can you double check it doesn't break the on navigation maintain toast flow though? (don't think it will, so approving anyway)

packages/shared/src/hooks/useTimedAnimation.ts
sshanzel: Oh, right. We can use that.

sshanzel:
> Can you double check it doesn't break the on navigation maintain toast flow though? (don't think it will, so approving anyway)

IIRC, it should actually close right? Or is it only the behavior on mobile?

rebelchris:
@sshanzel I think we have this option where it should persist between navigation.
Not 100% sure on this though.

sshanzel:
> @sshanzel I think we have this option where it should persist between navigation. Not 100% sure on this though.

All cool. We can probably discuss it at a different time, at least we fixed an actual issue here.

PR #3376, opened 2024-07-19
fix: max count on card tags

sshanzel:
## Changes
- Introduced some calculations to identify how many tags we can put.
- ~~This is only applied to list mode, as per spec, but we can actually use this on cards as well. I think it is better to consolidate the calculation.~~
- Both cards and list will use the same calculation as it is based on what value you will pass to the hook.
- Feel free to test on preview.

Preview:

Screenshot 2024-07-19 at 11 08 11 PM
Screenshot 2024-07-19 at 10 57 23 PM
Screenshot 2024-07-19 at 10 57 06 PM
Screenshot 2024-07-19 at 10 57 04 PM
Screenshot 2024-07-19 at 10 56 59 PM

## Events

## Experiment

## Manual Testing

> [!CAUTION]

MI-448 #done

# Preview domain

https://mi-448.preview.app.daily.dev/

### Preview domain
https://mi-448.preview.app.daily.dev

### DIFF
--- a/packages/shared/src/components/cards/list/CardCover.tsx
+++ b/packages/shared/src/components/cards/list/CardCover.tsx
@@ -13,11 +13,16 @@ interface CardCoverProps extends ReusedCardCoverProps {
 
 export function CardCoverList({
   className,
+  imageProps,
   ...props
 }: CardCoverProps): ReactElement {
   return (
     <SharedCardCover
       {...props}
+      imageProps={{
+        ...imageProps,
+        className: classNames(imageProps.className, 'tablet:w-60'),
+      }}
       CardImageComponent={CardImage}
       renderOverlay={({ overlay, image }) => (
         <ConditionalWrapper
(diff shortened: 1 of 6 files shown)


packages/shared/src/components/cards/list/CardCover.tsx
```diff
@@ -13,11 +13,16 @@ interface CardCoverProps extends ReusedCardCoverProps {
...
+  imageProps,
   ...props
 }: CardCoverProps): ReactElement {
   return (
     <SharedCardCover
       {...props}
+      imageProps={{
+        ...imageProps,
+        className: classNames(imageProps.className, 'tablet:w-60'),
+      }}
```
sshanzel, line 25: Fixed the inconsistent sizes of images which produces content shift.

Before:

https://github.com/user-attachments/assets/028d7ea5-c7e1-4cfd-984b-e772fe02ba2d

Figma:
Screenshot 2024-07-19 at 11 08 46 PM

After:

Screenshot 2024-07-19 at 11 09 33 PM

rebelchris (approved):
Nice looks great, couldn't really find any missed ones.

PR #3377, opened 2024-07-19
fix: tldr height for shared links

rebelchris:
## Changes

We have this [edge case](https://dailydotdev.slack.com/archives/C01L0QXQJNB/p1720729579314099) when sharing posts in squads.
This is a place where we render the TLDR with the option to hide it.
For this we use some JS calculations, which don't take the `show more` rendered height into account.

(Doesn't affect existing implementations as it just renders without the need for height wrapping.

Instead of relying on JS calculations I've decided to simply fix it with CSS magic :) 
First time ever using a [youtube video](https://www.youtube.com/watch?v=B_n4YONte5A) as a resource 😅

In action: (with CSS only)
https://github.com/user-attachments/assets/d79c2903-8be0-4498-b0de-314adbe0ae03

## Events

## Experiment

## Manual Testing

> [!CAUTION]

AS-452 #done 

### Preview domain
https://as-452-tldr-height.preview.app.daily.dev

### DIFF
--- a/packages/shared/src/components/cards/PostSummary.tsx
+++ b/packages/shared/src/components/cards/PostSummary.tsx
@@ -18,6 +18,7 @@ function PostSummary(
   return (
     <SummaryContainer ref={ref} {...props}>
       <ShowMoreContent
+        className="overflow-hidden"
         content={summary}
         charactersLimit={330}
         threshold={50}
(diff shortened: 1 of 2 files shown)


capJavert (approved):
I am not much of a CSS guy but its hell of a better then JS calculations for sure :rocket:

packages/shared/src/components/cards/PostSummary.tsx
```diff
@@ -18,6 +18,7 @@ function PostSummary(
   return (
     <SummaryContainer ref={ref} {...props}>
       <ShowMoreContent
+        className="overflow-hidden"
```
capJavert, line 21: works on iOS/safari? Just because its a new IE 😆

rebelchris: Tested on ios: 👍

https://as-452-tldr-height.preview.app.daily.dev/posts/shameless-self-promotion--7qfatxo1a

capJavert (approved):

PR #3383, opened 2024-07-23
refactor: source actions

ilasw:
Part 1 from splitted https://github.com/dailydotdev/apps/pull/3360 

## Changes
- Refactor `Source Actions` (rename block/follow to block/unblock)

## Events

## Experiment

## Manual Testing

> [!CAUTION]

### Preview domain
https://refactor-source-actions.preview.app.daily.dev


packages/shared/src/components/PostOptionsMenu.tsx
```diff
@@ -350,13 +350,15 @@ export default function PostOptionsMenu({
...
         <MenuIcon
-          Icon={sourceSubscribe.isSubscribed ? BellSubscribedIcon : BellIcon}
+          Icon={
+            sourceSubscribe.haveNotifications ? BellSubscribedIcon : BellIcon
+          }
         />
       ),
-      label: `${
-        sourceSubscribe.isSubscribed ? 'Unsubscribe from' : 'Subscribe to'
+      label: `Receive notifications ${
```
nensidosari, line 358: Is this label change not part of the experiment?

nensidosari: I'm not sure if the copy makes sense in

`Receive notifications to`

packages/shared/src/components/icons/Bell/Add/filled.svg
nensidosari, line 1: is this icon addition for experiment?

packages/shared/src/components/icons/Bell/Add/index.tsx
nensidosari, line 1: is this icon addition for experiment?

packages/shared/src/components/icons/Bell/Add/outlined.svg
nensidosari, line 1: is this icon addition for experiment?

packages/shared/src/components/icons/Bell/filled.svg
nensidosari, line 1: is this icon addition for experiment?

packages/shared/src/components/widgets/PostUsersHighlights.tsx
```diff
@@ -194,24 +192,26 @@ export const UserHighlight = (props: UserHighlightProps): ReactElement => {
               />
             )}
           </div>
-          {(handle || username || id) && (
+          {(handleOrUsername || id) && (
```
nensidosari, line 195: Why not handleOrUsernameOrId 😅

packages/shared/src/lib/log.ts
```diff
@@ -149,9 +149,19 @@ export enum LogEvent {
   ScheduleStreakReminder = 'schedule streak reminder',
   // 404 page
   View404Page = '404 page',
-  // source
+  // Source Actions - start
```
nensidosari, line 152: Just referencing this comment here from previous PR https://github.com/dailydotdev/apps/pull/3360#discussion_r1686342509 from @capJavert 

> Just verify with Ido if this is ok so it can be added to taxonomy sheet (also with other events related changes)

packages/shared/src/hooks/source/useSourceActionsNotify.ts
```diff
@@ -9,60 +9,62 @@ import { useNotificationPreferenceToggle } from '../notifications';
...
 
+    const displayName = 'name' in source ? source.name : source?.id;
+
     displayToast(
-      result.isSubscribed
-        ? '✅ You are now subscribed'
-        : '⛔️ You are now unsubscribed',
+      notifications.isSubscribed
+        ? `✅ You'll get notified every time ${displayName} posts`
+        : `⛔️ You'll no longer get notified about ${displayName} posts`,
```
nensidosari, line 61: Is this copy change intended even without experiment?

nensidosari (commented):
Mostly comments to check if somethings are intended for this PR or not. Will test in preview in few mins just to make sure we're not changing anything at this moment

packages/shared/src/hooks/source/useSourceActionsNotify.ts
ilasw: I kept because it make sense and clarify whats going on, but maybe is more a part of the experiment and should revert :\

nensidosari: Yeah, we either move this behind experiment, or we check with product in this case for copy change.

packages/shared/src/lib/log.ts
ilasw: What do you think about this? @idoshamun

idoshamun: In the taxonomy we have these events already and they follow our legacy convention which is event_name and target_type. It already exists in the code base if you look for it.
We have `block` and `unblock` as event_name and `source` and `tag` as the `target_type`. Let's not change it please and keep logging as it is
image

ilasw: Currently we are firing `follow` events when user `Unblock` the source, I think this is confusing and is not implemented as the table you mentioned. 
You can check yourself at `packages/shared/src/hooks/useTagAndSource.ts` in `line 209`. Should I keep this? @idoshamun @nensidosari 

And if yes how the event names changes with the new experiment?

nensidosari: First, do u have access to view this doc https://docs.google.com/spreadsheets/d/18Lv7zXges9QfVX5VYL1a-Hyl0e1sQ3sLr0OK8YZWKXI/edit?gid=0#gid=0?
Screenshot 2024-07-23 at 2 15 01 PM

Screenshot 2024-07-23 at 2 14 28 PM

I think these are the events that refer to follow/block or what we're naming them after block/unblock. 
What Ido is mentioning is that even with the new experiment where we are changing the naming, let's not change events names.

So for this PR we don't want any changes in event names, nor in names for buttons. We can see at the experiment PR if we want to do any changes or add any new events if missing. 🙏

idoshamun: > Currently we are firing follow events when user Unblock the source, I think this is confusing and is not implemented as the table you mentioned.
You can check yourself at packages/shared/src/hooks/useTagAndSource.ts in line 209. Should I keep this? @idoshamun @nensidosari

Oh that's not good at all :triggered:

idoshamun: @ilasw let's fix it then as suggested. but please stick to the taxonomy definitions

ilasw: For this PR events are ok, updated even in the new one https://github.com/dailydotdev/apps/pull/3385

idoshamun: I'm not sure why are we adding more events to the enum. I think there's a confusion here. Please note that in the taxonomy we separate event_name and target_type and here you merge them to one. It's a little tricky so happy to guide you through if needed

nensidosari (approved):
Looks great, approving, just please fix the last detail from testing 🙏 

Follow has now the block icon, in prod it has a plus icon - let's stick to that
Screenshot 2024-07-23 at 3 26 10 PM

nensidosari (approved):

packages/shared/src/hooks/useTagAndSource.ts
```diff
@@ -199,7 +199,7 @@ export default function useTagAndSource({
     ],
   );
 
-  const onFollowSource = useCallback(
+  const onUnblockSource = useCallback(
```
sshanzel, line 202: Why did we end up changing the definitions of these functions, I think it contributed a lot to refactors we could've avoided. It is great that we do refactors, though I think for experiments, we should keep it to minimum since things might still change after the result.

packages/shared/src/hooks/useMutateFilters.ts
```diff
@@ -389,25 +389,23 @@ export default function useMutateFilters(
     [user, queryClient, feedId],
   );
 
-  const { mutateAsync: followSourceRemote } = useMutation<
+  const { mutateAsync: unblockSourceRemote } = useMutation<
```
sshanzel, line 392: In my humble opinion, I think we should not have refactored these ones, it made a lot of avoidable changes.

sshanzel (changes requested):
Love how things were simplified, but I am a bit in doubt with the large number of refactors we had to do.

packages/shared/src/hooks/useTagAndSource.ts
ilasw: the problem is that the new event is literally called "follow" and would conflict with this one which is an "unblock". Beyond the experiment it might be better to clarify what the method does by giving it a consistent name. What do you think?

packages/shared/src/components/sources/SourceActions/SourceSubscribeButton.tsx
```diff
@@ -65,21 +67,24 @@ const SourceSubscribeButton = ({
...
     : SourceSubscribeButtonRegular;
 
   return (
-    <ButtonComponent
-      className={className}
-      isFetching={isLoggedIn && !isReady}
-      onClick={onSubscribe}
-      variant={variant}
-    />
+    <>
```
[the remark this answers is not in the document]
ilasw: Removed!

packages/shared/src/hooks/useTagAndSource.ts
sshanzel: I agree, we should have these updated, but I think now is not yet the best time as this would still be running in an experiment. I can't exactly remember how things look now if the process still fits the narrative of these functions, but still, I think we should make the refactor at the end of the experiments.

packages/shared/src/lib/log.ts
```diff
@@ -149,9 +149,19 @@ export enum LogEvent {
...
+  // Source Actions - end
   SubscribeSource = 'subscribe source',
   UnsubscribeSource = 'unsubscribe source',
+  // Source Actions - end
+  // Tags - start
+  FollowTag = 'follow tag',
+  UnfollowTag = 'unfollow tag',
+  BlockTag = 'block tag',
+  UnblockTag = 'unblock tag',
+  // Tags - end
```
ilasw, line 164: remove those

nensidosari: Only if these change from the naming events had - if it's just moving from plain string to enum with the same content then you can leave.

nensidosari (approved):

sshanzel:
Feel free to move ahead, if Nensi's approval has been given :shipit:

PR #3389, opened 2024-07-24
feat: add card downvote

capJavert:
## Changes

- add downvote to desktop cards
- downvote panel also added
- adjusted spacing based on new design
- all small icons per new design
- events send the same as from options -> downvote (which we already had)

## Events

## Experiment

## Manual Testing

> [!CAUTION]

AS-446 #done

### Preview domain
https://as-446-card-downvote.preview.app.daily.dev

### DIFF
--- a/packages/shared/src/components/buttons/QuaternaryButton.tsx
+++ b/packages/shared/src/components/buttons/QuaternaryButton.tsx
@@ -13,6 +13,7 @@ export type QuaternaryButtonProps<TagName extends AllowedTags> =
   ButtonProps<TagName> & {
     reverse?: boolean;
     responsiveLabelClass?: string;
+    labelClassName?: string;
   };
 
 function QuaternaryButtonComponent<TagName extends AllowedTags>(
@@ -24,6 +25,7 @@ function QuaternaryButtonComponent<TagName extends AllowedTags>(
     reverse,
     responsiveLabelClass,
     tag = 'button',
+    labelClassName,
     ...props
   }: QuaternaryButtonProps<TagName>,
   ref?: Ref<ButtonElementType<TagName>>,
@@ -73,6 +75,7 @@ function QuaternaryButtonComponent<TagName extends AllowedTags>(
             'cursor-pointer items-center font-bold typo-callout',
             labelDisplay,
             { readOnly: props.disabled },
+            labelClassName,
           )}
         >
           {children}
--- a/packages/shared/src/components/cards/ActionsButtons/ActionButtons.tsx
+++ b/packages/shared/src/components/cards/ActionsButtons/ActionButtons.tsx
@@ -7,6 +7,7 @@ import {
   DiscussIcon as CommentIcon,
   BookmarkIcon,
   LinkIcon,
+  DownvoteIcon,
 } from '../../icons';
 import {
   Button,
@@ -19,6 +20,10 @@ import { SimpleTooltip } from '../../tooltips/SimpleTooltip';
 import { useFeedPreviewMode } from '../../../hooks';
 import { UpvoteButtonIcon } from './UpvoteButtonIcon';
 import { BookmarkButton } from '../../buttons';
+import { feature } from '../../../lib/featureManagement';
+import { withExperiment } from '../../withExperiment';
+import { IconSize } from '../../Icon';
+import { useBlockPostPanel } from '../../../hooks/post/useBlockPostPanel';
 
 export interface ActionButtonsProps {
   post: Post;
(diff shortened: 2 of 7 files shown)


packages/shared/src/components/buttons/QuaternaryButton.tsx
```diff
@@ -73,6 +75,7 @@ function QuaternaryButtonComponent<TagName extends AllowedTags>(
             'cursor-pointer items-center font-bold typo-callout',
             labelDisplay,
             { readOnly: props.disabled },
+            labelClassName,
```
capJavert, line 78: I needed this to override padding

packages/shared/src/components/cards/ActionsButtons/ActionButtons.tsx
```diff
@@ -119,4 +125,137 @@ export default function ActionButtons({
       {lastActions}
     </div>
   );
-}
+};
+
+const ActionButtonsDownvote: typeof ActionButtonsControl = ({
```
capJavert, line 130: There were too many changes to margins/paddings to it was cleaner to just create separate component and utilize `withExperiment` HOC

sshanzel (commented):
Looks great overall. Although, I think we need one for list type?

capJavert:
@sshanzel from what I can see lists already have downvote together with mobile feed layout?

sshanzel:
> @sshanzel from what I can see lists already have downvote together with mobile feed layout?

Oh, in that case, we should be good here :shipit:

sshanzel (approved):
🚀

nensidosari:
@capJavert one thing from testing so far: hight of cards changes when the downvote feedback shows. I'm guessing it should stay the same
Screenshot 2024-07-25 at 8 00 41 AM

nensidosari:
Another thing, when clicking on block it says "No topics were blocked", is this right?

nensidosari:
If I click downvote and then upvote again I would expect the downvote panel to be removed (list view)
Screenshot 2024-07-25 at 8 04 18 AM

packages/shared/src/components/buttons/QuaternaryButton.tsx
```diff
@@ -24,6 +25,7 @@ function QuaternaryButtonComponent<TagName extends AllowedTags>(
     reverse,
     responsiveLabelClass,
     tag = 'button',
+    labelClassName,
```
nensidosari, line 28: Why not opt in for classname as an object? Or too many changes that way?

nensidosari (commented):
Nothing to add codewise, looks good and happy to approve.

Just the tests, take a look and let me know 🙏 Though seems like the behaviour of the downvote pannel, block action works the same in prod just don't make sense.

packages/shared/src/components/buttons/QuaternaryButton.tsx
rebelchris: We should really get rid of this button comp anyway 🥲

packages/shared/src/components/cards/WelcomePostCard.tsx
```diff
@@ -26,6 +26,7 @@ export const WelcomePostCard = forwardRef(function SharePostCard(
     children,
     enableSourceHeader = false,
     domProps = {},
+    onDownvoteClick,
```
rebelchris, line 29: I really wish we had one source to only mod this (no action needed maybe good backlog)

rebelchris (approved):
Besides nensis feedback not much.
Was it intended to show the downvote actions (dont have access to figma atm)

packages/shared/src/components/buttons/QuaternaryButton.tsx
nensidosari: agree

capJavert: because I did not want to refactor all other usages

capJavert: @nensidosari as to downvote/block actions I think we should fix it separate since this only adds the button and reuses logic already used for lists from before?

As to height issue, I will fix it, good catch 👌

nensidosari: @capJavert Makes sense 🙌🏻 I'll create a ticket for what I found that are not relevant here

capJavert: Fixed height in a9869501031b98c4099c343c5ce59b5e199756eb, I am not sure why `max-h` was used but from testing looks good without it and tags panel just expands into card container. 
image

nensidosari (approved):

PR #3391, opened 2024-07-25
feat: open login modal when origin is notification

sshanzel:
## Changes
- When users click the notifications or email pointing towards dailydev, we include a `utm_source` parameter.
- We will use that parameter to tell whether they came from clicking notifications and such.
- Currently applied on post pages, but let me know if we want this behavior across the app.

Feel free to open the link below using incognito:
https://mi-453.preview.app.daily.dev/posts/jx5hH81YR?utm_source=notification&utm_medium=email&utm_campaign=squad_post_added

## Events

## Experiment

## Manual Testing

> [!CAUTION]

MI-453

### Preview domain
https://mi-453.preview.app.daily.dev


packages/webapp/pages/posts/[id]/index.tsx
```diff
@@ -123,6 +128,16 @@ const PostPage = ({ id, initialData }: Props): ReactElement => {
...
   });
 
+  const shouldShowLogin = !user && utmSource && isAuthReady;
+
+  useEffect(() => {
+    if (!shouldShowLogin) {
+      return;
+    }
+
+    showLogin({ trigger: AuthTriggers.FromNotification });
```
Comment (omBratteng) [inline packages/webapp/pages/posts/[id]/index.tsx:138]:
```suggestion
    showLogin({ trigger: AuthTriggers.FromNotification, isLogin: true });
```
Think this should open the modal on the login view, and not the signup view

Comment (sshanzel) [reply packages/webapp/pages/posts/[id]/index.tsx:138]: Good catch. Yes, that is right.

nensidosari:
@sshanzel how much work would it be to make it across app not only on post pages?
some feedback from Chris: We probably do want it app wide if not too much work as some notifications can end up at random pages.

I think it's a valid concern. Maybe do check with product if u want but might be better this way.

nensidosari:
Go for it then @sshanzel 😁

sshanzel:
@nensidosari updated it now 🥳

packages/shared/src/components/MainLayout.tsx
```diff
@@ -153,6 +154,20 @@ function MainLayoutComponent({
     router.push(`${onboardingUrl}?${params.toString()}`);
   }, [shouldRedirectOnboarding, router]);
 
+  const utmSource = router?.query?.utm_source;
```
nensidosari, line 157: is this a param coming from both notifications in app, push notifications and emails?

nensidosari (approved):
looks good

packages/shared/src/components/MainLayout.tsx
sshanzel: Yeah, if it is in-app, that means they are logged in. For both email and push notifications, they both have this param.

PR #3395, opened 2024-07-25
feat: refetch profile after update

capJavert:
## Changes

- refetch profile after update, noticed when testing social links
- this is done by manually reloading the page (so that query client refetches)
  - proper fix for this would be to utilize [on demand revalidation](https://nextjs.org/docs/pages/building-your-application/data-fetching/incremental-static-regeneration#on-demand-revalidation), verified this method works locally but I did not want to complicate here since revalidating could be expensive with the new Vercel pricing)
- the SSR cache will still contain old data for next 60s (due to ssr cache) but on hydration it changes to correct one

## Events

## Experiment

## Manual Testing

> [!CAUTION]

### Preview domain
https://fix-profile-stale-after-update.preview.app.daily.dev

### DIFF
--- a/packages/webapp/pages/account/profile.tsx
+++ b/packages/webapp/pages/account/profile.tsx
@@ -95,7 +95,10 @@ const AccountProfilePage = (): ReactElement => {
       linkedin: values.linkedin,
       mastodon: values.mastodon ? withHttps(values.mastodon) : undefined,
       experienceLevel: values.experienceLevel,
-      onUpdateSuccess: () => router.push(`/${values.username}`),
+      onUpdateSuccess: () =>
+        router.push(`/${values.username}`).then(() => {
+          router.reload();
+        }),
     };
     updateUserProfile(params);
   };


packages/shared/src/hooks/profile/useProfile.ts
```diff
@@ -13,8 +12,7 @@ export function useProfile(initialUser?: PublicProfile): PublicProfile {
     generateQueryKey(RequestKey.Profile, initialUser),
     () => getProfile(initialUser?.id),
     {
-      ...disabledRefetch,
```
capJavert, line 16: We don't really need to disable refetch because we now have stale time (it wont refetch until that is elapsed or we use `invalidateQueries` (like during profile update)

packages/shared/src/hooks/profile/useProfile.ts
```diff
@@ -13,8 +12,7 @@ export function useProfile(initialUser?: PublicProfile): PublicProfile {
...
-      cacheTime: StaleTime.OneHour,
+      staleTime: StaleTime.OneHour,
```
capJavert, line 15: This was probably meant to be `staleTime` since `cacheTime` does not really keep from refetching.

capJavert: Referencing PR that added it https://github.com/dailydotdev/apps/pull/3305

packages/shared/src/contexts/BootProvider.tsx
```diff
@@ -202,9 +202,10 @@ export const BootDataProvider = ({
     async (newUser: LoggedUser | AnonymousUser) => {
       const updated = updateLocalBootData(cachedBootData, { user: newUser });
       setCachedBootData(updated);
-      await queryClient.invalidateQueries(
-        generateQueryKey(RequestKey.Profile, newUser),
-      );
+      await queryClient.invalidateQueries({
+        queryKey: generateQueryKey(RequestKey.Profile, newUser),
+        refetchType: 'all',
```
capJavert, line 207: refetch inactive query as well (since profile one will be due to `useProfile` hook not being mounted until redirect to user profile page)

packages/webapp/pages/account/profile.tsx
```diff
@@ -95,7 +95,10 @@ const AccountProfilePage = (): ReactElement => {
       linkedin: values.linkedin,
       mastodon: values.mastodon ? withHttps(values.mastodon) : undefined,
       experienceLevel: values.experienceLevel,
-      onUpdateSuccess: () => router.push(`/${values.username}`),
+      onUpdateSuccess: () =>
+        router.push(`/${values.username}`).then(() => {
+          router.reload();
```
capJavert, line 100: quick fix to reload the whole page after update

nensidosari (approved):

omBratteng (approved):

PR #3396, opened 2024-07-25
fix: icon has wrong alignment in profile experience dropdown

ilasw:
## Changes
- Fixed `ExperienceLevelDropdown`'s icon margin and width

## Events

## Experiment

## Manual Testing

> [!CAUTION]

### Preview domain
https://mi-442-align-profile-input-icons.preview.app.daily.dev


packages/shared/src/components/profile/ExperienceLevelDropdown.tsx
```diff
@@ -79,7 +79,7 @@ const ExperienceLevelDropdown = ({
         }}
         onOpenChange={setOpen}
         placeholder="Experience level"
-        icon={<TerminalIcon className="mr-2" />}
+        icon={<TerminalIcon className="ml-0 mr-2 !w-5" />}
```
nensidosari, line 82: We have icon sizes, wouldn't the w-5 be IconSize.XSmall?

ilasw: Yes, but we also have `!w-6` class forced by the button so we need to override it :(

ilasw: We can use this but the icon will be slightly bigger the the previous 👀 

```tsx
icon={<TerminalIcon className="ml-0 mr-1" size={IconSize.XSmall} />}
```

image

nensidosari: But I think since the issue was alignment, then only `icon={<TerminalIcon className="ml-0 mr-1" />}` would work just fine, no?

nensidosari: Screenshot 2024-07-25 at 5 38 05 PM

ilasw: Okay, move to this approach 👍

nensidosari (approved):

PR #3397, opened 2024-07-25
fix: dev card has too much sources;

ilasw:
## Changes
- Limited source count shown in DevCard

## Events

## Experiment

## Manual Testing

> [!CAUTION]

### Preview domain
https://mi-443-devcard-sources-smooshed.preview.app.daily.dev

### DIFF
--- a/packages/shared/src/components/profile/devcard/DevCardFooter.tsx
+++ b/packages/shared/src/components/profile/devcard/DevCardFooter.tsx
@@ -4,7 +4,7 @@ import { TagLinks } from '../../TagLinks';
 import { ButtonVariant } from '../../buttons/common';
 import SourceButton from '../../cards/SourceButton';
 import Logo, { LogoPosition } from '../../Logo';
-import { DevCardTheme, DevCardType } from './common';
+import { DevCardConstraints, DevCardTheme, DevCardType } from './common';
 import { Source } from '../../../graphql/sources';
 import { checkLowercaseEquality } from '../../../lib/strings';
 import { ProfileImageSize } from '../../ProfilePicture';
@@ -21,11 +21,13 @@ export interface DevCardFooterProps {
 export function DevCardFooter({
   type,
   tags,
-  sources,
+  sources = [],
   theme,
   shouldShowLogo = true,
   elementsClickable = true,
 }: DevCardFooterProps): ReactElement {
+  const topSources = sources.slice(0, DevCardConstraints.MaxSourceCount);
+
   return (
     <>
       <TagLinks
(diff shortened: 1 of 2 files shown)


packages/shared/src/components/profile/devcard/DevCardFooter.tsx
```diff
@@ -21,11 +21,13 @@ export interface DevCardFooterProps {
 export function DevCardFooter({
   type,
   tags,
-  sources,
+  sources = [],
   theme,
   shouldShowLogo = true,
   elementsClickable = true,
 }: DevCardFooterProps): ReactElement {
+  const topSources = sources.slice(0, DevCardConstraints.MaxSourceCount);
```
rebelchris, line 29: Wouldn't it be better to limit what we query already?
(Not sure top of head where the information comes from now)

capJavert: on the query side limit is already 5

rebelchris: So how come we show more then 5? 👀😅

nensidosari: 👀 what

capJavert: so changes in this PR would do nothing since API will never return more then 5

nensidosari: But how come in the issue reported we showed 6??

capJavert: > So how come we show more then 5? 👀😅

I checked on API and it only returns 5, from the image Chris provided in report it actually looks like its just the source image that looks bad

image

nensidosari: Oh wow... nice issue 😅 @ilasw u can close this PR and the ticket apparently

capJavert: `DevCardById` query uses `getFavoriteSources` which has `LIMIT 5`.

ilasw: ...closing PR!

capJavert: I found it, this is the squad https://app.daily.dev/squads/dotnetangularfullstack

verified locally image looks like that
image

capJavert: cc: @nensidosari @ilasw @rebelchris

rebelchris: Omg facepalm 🤦‍♂️😢😂

PR #3401, opened 2024-07-26
feat: generate tldr button experiment

sshanzel:
## Changes
- A new experiment about TLDR.
- We will add an overlay when the user hasn't clicked the button yet.
- Should be applied across the app and to Companion.

Preview:

https://github.com/user-attachments/assets/8ca31696-a510-48f3-8315-63cb8c036884

## Events

## Experiment

## Manual Testing

> [!CAUTION]

MI-450

### Preview domain
https://mi-450.preview.app.daily.dev

### DIFF
--- a/packages/shared/src/components/post/PostSummaryButton.tsx
+++ b/packages/shared/src/components/post/PostSummaryButton.tsx
@@ -1,0 +1,38 @@
+import React, { ReactElement } from 'react';
+import { useSummaryActivation } from '../../hooks/experiments';
+import { Loader } from '../Loader';
+import { MagicIcon } from '../icons';
+import { ButtonIconPosition, ButtonVariant } from '../buttons/common';
+import { Button, ButtonColor } from '../buttons/Button';
+
+interface PostSummaryButtonProps {
+  summary: string;
+}
+
+export function PostSummaryButton({
+  summary,
+}: PostSummaryButtonProps): ReactElement {
+  const { shouldShowOverlay, isLoading, onClickSummary } = useSummaryActivation(
+    { summary },
+  );
+
+  if (!shouldShowOverlay) {
+    return null;
+  }
+
+  return (
+    <>
+      <div className="absolute inset-0 bg-gradient-to-t from-background-default from-60% to-overlay-primary-pepper" />
+      <Button
+        icon={isLoading ? <Loader /> : <MagicIcon secondary />}
+        iconPosition={ButtonIconPosition.Right}
+        variant={ButtonVariant.Primary}
+        color={ButtonColor.Cabbage}
+        className="absolute left-1/2 top-1/2 -translate-x-1/2 -translate-y-1/2"
+        onClick={onClickSummary}
+      >
+        Auto-summarize
+      </Button>
+    </>
+  );
+}
--- a/packages/shared/src/components/post/PostSummaryButton.tsx
+++ b/packages/shared/src/components/post/PostSummaryButton.tsx
@@ -6,7 +6,7 @@ import { ButtonIconPosition, ButtonVariant } from '../buttons/common';
 import { Button, ButtonColor } from '../buttons/Button';
 
 interface PostSummaryButtonProps {
-  summary: string;
+  summary?: string;
 }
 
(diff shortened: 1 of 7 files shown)


packages/shared/src/components/post/PostSummaryButton.tsx
```diff
@@ -0,0 +1,38 @@
+import React, { ReactElement } from 'react';
+import { useSummaryActivation } from '../../hooks/experiments';
+import { Loader } from '../Loader';
+import { MagicIcon } from '../icons';
+import { ButtonIconPosition, ButtonVariant } from '../buttons/common';
+import { Button, ButtonColor } from '../buttons/Button';
+
+interface PostSummaryButtonProps {
+  summary: string;
```
capJavert, line 9: type wise summary can be undefined, I know TS is not complaining now but in strict TS it would (same for hook).

packages/shared/src/hooks/experiments/useSummaryActivation.ts
```diff
@@ -0,0 +1,59 @@
...
+        clearTimeout(timeoutRef.current);
+      }
+    };
+  }, []);
+
+  const onClickSummary = useCallback(() => {
+    setIsLoading(true);
+    completeAction(ActionType.ShowSummary);
+
+    timeoutRef.current = window.setTimeout(() => {
```
capJavert, line 50: I don't think you need `window` here, `setTimeout` is fine as global hoisted function.

capJavert: Probably also clear previous timeout here if user clicks multiple times.

packages/shared/src/hooks/experiments/useSummaryActivation.ts
```diff
@@ -0,0 +1,59 @@
+import { useCallback, useEffect, useRef, useState } from 'react';
+import { ActionType } from '../../graphql/actions';
+import { useToastNotification } from '../useToastNotification';
+import { useAuthContext } from '../../contexts/AuthContext';
+import { useActions } from '../useActions';
+import { useConditionalFeature } from '../useConditionalFeature';
+import { feature } from '../../lib/featureManagement';
+
+interface UseSummaryActivationProps {
+  summary: string;
```
capJavert, line 10: optional as well

capJavert (commented):
Some sanity checks, overall smooth!

packages/shared/src/hooks/experiments/useSummaryActivation.ts
sshanzel: Right. Sometimes the `setTimeout` from jest is getting imported automatically and I was going crazy as to why the app is getting broken. Just to ensure it is the actual function, I added `window`.

capJavert: Yeah the part about `window` is non blocking, I get that as well sometimes 🙏

sshanzel: Added the cleanups now 🥳

packages/shared/src/components/post/PostSummaryButton.tsx
```diff
@@ -0,0 +1,39 @@
...
+    { summary },
+  );
+
+  if (!shouldShowOverlay) {
+    return null;
+  }
+
+  return (
+    <>
+      <div className="absolute inset-0 bg-gradient-to-t from-background-default from-60% to-overlay-primary-pepper" />
```
nensidosari, line 25: Just wondering, is this considering the light mode?

capJavert (approved):

nensidosari (approved):
Looks good. Just the concern on light mode

packages/shared/src/components/post/PostSummaryButton.tsx
sshanzel: Yes, since the colors are using the themed classes.

nensidosari: Ah nevermind, I thought I saw a set color

PR #3406, opened 2024-07-29
feat: added new layout experiment for search page (desktop)

ilasw:
## Changes
- Created new `SearchResultsLayout` component;
- Added conditional FeedContainer for search result page;
- Added `WidgetCard` for widgets in sidebar;
- Added tracking events for element in sidebar;

image

## Events

## Experiment

Introduced experiment `search_results_upgrade` for better search results on desktop search page.

> [!IMPORTANT]

## Manual Testing

> [!CAUTION]

### Preview domain
https://mi-462-search-result-page-upgrade.preview.app.daily.dev


packages/shared/src/hooks/search/useSearchResultsLayout.ts
```diff
@@ -0,0 +1,36 @@
...
+export const useSearchResultsLayout = (
+  props?: UseSearchResultsLayoutProps,
+): SearchResultsLayout => {
+  const { feedName } = props ?? {};
+  const isLaptop = useViewSize(ViewSize.Laptop);
+  const { value: isSearchUpgradeExperiment, isLoading } = useConditionalFeature(
+    {
+      feature: feature.searchResultsUpgrade,
+      shouldEvaluate:
+        !!isLaptop && (!feedName || feedName === SharedFeedPage.Search),
```
capJavert, line 24: Why `!feedName`?

packages/shared/src/components/search/SearchResults/SearchResultsSources.tsx
```diff
@@ -0,0 +1,45 @@
...
+interface SearchResultsSourcesProps {
+  items: Source[];
+  isLoading: boolean;
+  onSourceClick: (source: Source) => void;
+}
+
+export const SearchResultsSources = (
+  props: SearchResultsSourcesProps,
+): ReactElement => {
+  const { items, isLoading, onSourceClick } = props;
```
capJavert, line 16: I see you do this a lot, I don't mind, but why not just do inline in arrow function definition `{ items, isLoading, onSourceClick }: SearchResultsSourcesProps`?

packages/shared/src/components/search/SearchResults/SearchResultsLayout.tsx
```diff
@@ -0,0 +1,151 @@
...
+    logEvent({
+      event_name: LogEvent.Click,
+      target_type: TargetType.SearchRecommendation,
+      target_id: tag,
+      feed_item_title: tag,
+      extra: JSON.stringify({
+        origin: Origin.SearchPage,
+        provider: SearchProviderEnum.Tags,
+      }),
+    });
```
capJavert, line 83: These are not new? Please verify with Ido if he wants some other origin or tracking for this experiment (if you did not already).

packages/shared/src/components/search/SearchResults/SearchResultsLayout.tsx
```diff
@@ -0,0 +1,151 @@
...
+
+    push(`${webappUrl}tags/${tag}`);
+  };
+
+  if (!isSearchResultsUpgrade) {
+    return <>{children}</>;
```
capJavert, line 89: why not just `children`?

packages/shared/src/components/search/SearchResults/SearchResultsLayout.tsx
```diff
@@ -0,0 +1,151 @@
...
+  const { children } = props;
+
+  const { isSearchResultsUpgrade } = useSearchResultsLayout();
+  const { items = [] } = useContext(ActiveFeedContext);
+  const {
+    query: { q: query },
+    push,
+  } = useRouter();
+  const { logEvent } = useLogContext();
+  const isLoading = !items.length;
```
capJavert, line 49: Search results could have no results right, so this is not exactly correct? If you need loading specifically you can add it to `ActiveFeedContext`

packages/shared/src/components/Feed.tsx
```diff
@@ -409,18 +413,23 @@ export default function Feed<T>({
...
-        header={header}
-        inlineHeader={inlineHeader}
-        className={className}
-        showSearch={showSearch && isValidFeed}
-        shortcuts={shortcuts}
-        actionButtons={actionButtons}
-        isHorizontal={isHorizontal}
-        feedContainerRef={feedContainerRef}
-      >
+      <FeedWrapperComponent {...containerProps}>
```
capJavert, line 432: Just sanity check the control please since `Feed` is used all over 🙏

packages/shared/src/components/search/SearchResults/SearchResultsLayout.tsx
```diff
@@ -0,0 +1,151 @@
...
+    const ids = new Set<string>();
+    return postItems
+      .map((item) => item.post.source)
+      .filter((source) => {
+        if (ids.has(source.id)) {
+          return false;
+        }
+        ids.add(source.id);
+        return true;
+      });
```
capJavert, line 68: You are trying to get unique sources? Maybe `reduce` is more simple to do in one go?

capJavert:
On tablet and below there is a padding

image

capJavert:
On no results the button for AI provider is missing and there is no padding (vs what production has).

image

packages/shared/src/components/search/SearchResults/SearchResultsLayout.tsx
ilasw: Children is `ReactNode`, just tried to be consistent

packages/shared/src/components/search/SearchResults/SearchResultsLayout.tsx
ilasw: @capJavert the tags events are the same for the search suggestions as specified in the kickoff, the new ones are the `sources`. @idoshamun there are some directives for those?

packages/shared/src/components/search/SearchResults/SearchResultsSources.tsx
ilasw: I do not like so much destructuring in parameters just because how is formatted, but is not a heavy preference! Should i put there for consistency?

packages/shared/src/hooks/search/useSearchResultsLayout.ts
ilasw: because props was optional but I'm making them mandatory for consistency and for avoiding errors!

packages/shared/src/components/search/SearchResults/SearchResultsLayout.tsx
ilasw: Not really needed, I'm removing! Thanks for pointing at :)

packages/shared/src/components/search/SearchResults/SearchResultsLayout.tsx
ilasw: Updated! ✅

packages/shared/src/components/search/SearchResults/SearchResultsSources.tsx
capJavert: I was just wondering why you do it, up to you.

packages/shared/src/hooks/search/useSearchResultsLayout.ts
capJavert: Should we just make it mandatory then or load it from `ActiveFeedContext`, to make sure hook behaves the same in all places?

ilasw: Cool! I've switched to 

```ts
```

packages/shared/src/components/MainFeedLayout.tsx
```diff
@@ -325,22 +336,24 @@ export default function MainFeedLayout({
...
     setSelectedAlgo,
+    getFeatureValue,
+    isSearchResultsUpgrade,
+    search,
+    tab,
     router.pathname,
+    isAnyExplore,
+    selectedPeriod,
+    isExploreLatest,
+    isLaptop,
```
capJavert, line 356: Why is this re-ordered?

packages/shared/src/components/search/SearchResults/SearchResultsLayout.tsx
```diff
@@ -0,0 +1,160 @@
...
+  const { isListMode } = useFeedLayout();
+  const { isSearchResultsUpgrade } = useSearchResultsLayout();
+  const { items = [] } = useContext(ActiveFeedContext);
+  const {
+    query: { q: query },
+    push,
+  } = useRouter();
+  const { logEvent } = useLogContext();
+
+  const postItems = items.filter(isItemPost);
```
capJavert, line 54: also memoize filter call here?

capJavert (approved):
Looks good to me, aside from checking from where do we want to extract related sources/tags (that I mentioned on slack).

packages/shared/src/hooks/useFeedLayout.ts
```diff
@@ -75,17 +76,19 @@ const getFeedPageLayoutComponent = ({
...
 > & {
   shouldUseListFeedLayoutOnMobileTablet: boolean;
+  isSearchResultsUpgrade: boolean;
 }): UseFeedLayoutReturn['FeedPageLayoutComponent'] => {
   if (shouldUseCommentFeedLayout) {
     return CommentFeedPage;
   }
 
-  if (shouldUseListMode) {
+  if (shouldUseListMode && !isSearchResultsUpgrade) {
```
sshanzel, line 91: Does this mean even if the user is in list mode but enrolled in the experiment, they will see the grid layout?

packages/shared/src/hooks/search/useSearchResultsLayout.ts
```diff
@@ -0,0 +1,29 @@
...
+  isSearchResultsUpgrade: boolean;
+}
+
+export const useSearchResultsLayout = (): SearchResultsLayout => {
+  const { feedName } = useActiveFeedNameContext();
+  const isLaptop = useViewSize(ViewSize.Laptop);
+  const { value: isSearchUpgradeExperiment, isLoading } = useConditionalFeature(
+    {
+      feature: feature.searchResultsUpgrade,
+      shouldEvaluate: !!isLaptop && feedName === SharedFeedPage.Search,
```
sshanzel, line 17: We should put this condition in a variable and use it on the condition for `isSearchResultsUpgrade`. The thing is, once the evaluation has finished, the result stays the same even if the value of the `shouldEvaluate` has changed.

For example: if you are viewing the search page, the result of `isSearchUpgradeExperiment` then becomes true, which would make `isSearchResultsUpgrade` become `true` also (let's say the other conditions are true as well). Now, if you switch pages, even though the `shouldEvaluate` has changed, the value of `isSearchUpgradeExperiment` from the cache is still the same which is still `true`.

packages/shared/src/components/search/SearchResults/SearchResultsTags.tsx
```diff
@@ -0,0 +1,40 @@
...
+  onTagClick: (suggestion: SearchSuggestion) => void;
+}
+
+export const SearchResultsTags = (
+  props: SearchResultsTagsProps,
+): ReactElement => {
+  const { items = [], isLoading, onTagClick } = props;
+  return (
+    <WidgetCard heading="Related tags">
+      {!!items.length && (
```
sshanzel, line 19: If you pass `null` on the `items` prop, it will not fallback to `[]`.
```suggestion
      {!!items?.length && (
```

packages/shared/src/components/search/SearchResults/SearchResultsSources.tsx
```diff
@@ -0,0 +1,45 @@
...
+  return (
+    <WidgetCard heading="Related sources">
+      {!!items.length && (
```
sshanzel, line 19: Same here, if there is a possibility for the value to be `null`.
```suggestion
      {!!items?.length && (
```

packages/shared/src/components/search/SearchResults/SearchResultsLayout.tsx
```diff
@@ -0,0 +1,160 @@
+import React, {
```
sshanzel, line 1: Is this just a renamed file?

sshanzel (commented):
Looks amazing overall. Just had some minor comments but I see this working already 💪

sshanzel (changes requested):
Sorry, I see that this has been approved, but I have to block the PR until the condition is fixed.

packages/shared/src/components/search/SearchResults/SearchResultsLayout.tsx
capJavert: won't be needed since we are gonna use `useSearchProviderSuggestions`

sshanzel:
Also, we must do a lot of sanity checks:
- Grid layout:
  - Laptop
  - Tablet
  - Mobile
  - Squads page
- List layout
  - Tablet
  - Mobile
  - Laptop
  - Squads page

Can't exactly remember if there are more, but would be great if we can verify, at least, these ones.

packages/shared/src/hooks/useFeedLayout.ts
ilasw: `isSearchResultsUpgrade` is true only in the search page feed, only scree width > laptop and if you enrolled the experiment. None of the other feeds should be changed by this addition

sshanzel (approved):
Looks great! 🥳

packages/shared/src/lib/featureManagement.ts
```diff
@@ -22,6 +22,7 @@ const feature = {
   }),
   feedAdSpot: new Feature('feed_ad_spot', 0),
   searchVersion: new Feature('search_version', 2),
+  searchResultsUpgrade: new Feature('search_results_upgrade', false),
```
rebelchris, line 25: Feature added in [GB](https://app.growthbook.io/features/search_results_upgrade).
@ilasw did you check the enrolment with Ido in the experiment channel already?

packages/shared/src/components/search/SearchResults/SearchResultsLayout.tsx
```diff
@@ -0,0 +1,145 @@
...
+        <div className="flex-1 border-r border-border-subtlest-tertiary">
+          <h2 className="px-4 py-4 font-bold text-text-primary typo-body">
+            Related posts
+          </h2>
+          <div
+            role="list"
+            className={classNames(
+              'mt-2.5',
+              isListMode
+                ? 'flex flex-col [&>article]:rounded-none'
```
rebelchris, line 90: Why is this `[&>article]:rounded-none` needed?
The actual list vs card should have no special needs right?

packages/shared/src/components/search/SearchResults/SearchResultsLayout.tsx
```diff
@@ -0,0 +1,145 @@
...
+                : 'grid w-96 grid-cols-1 gap-8 px-4 laptopL:w-auto laptopL:grid-cols-2',
```
rebelchris, line 91: We actually have to take density into account here (grid settings, probably best to check if they are supported now and render neatly)

packages/shared/src/components/search/SearchResults/SearchResultsTags.tsx
```diff
@@ -0,0 +1,45 @@
...
+  }
+
+  return (
+    <WidgetCard heading="Related tags">
+      {!!items?.length && (
+        <div className="flex flex-wrap gap-3" role="list">
+          {items.map((tag) => (
+            <TagLink
+              key={tag}
+              tag={tag}
```
rebelchris, line 29: I think you are using the wrong value here.
It renders as the name, but should be the tag value to be consistent.

Screenshot 2024-08-01 at 09 57 29

packages/shared/src/components/widgets/WidgetCard.tsx
```diff
@@ -0,0 +1,26 @@
+import React, { PropsWithChildren, ReactElement } from 'react';
+import classNames from 'classnames';
+
+interface WidgetCardProps extends PropsWithChildren {
+  heading?: string;
+  className?: string;
+}
+
+export const WidgetCard = (props: WidgetCardProps): ReactElement => {
```
rebelchris, line 9: Nice, didn't we have such a component already for what renders on other pages with sidebar?

rebelchris (changes requested):
Please check the enrolment (block it for now on that)
Also minor rendering with tags as names.

packages/shared/src/components/widgets/WidgetCard.tsx
ilasw: It is weird but didn't having one, right now for example `SimilarPosts` and `BestDiscussions` have 2 different implementation (I discovered recently working on another ticket) 👀

packages/shared/src/components/search/SearchResults/SearchResultsTags.tsx
ilasw: Updated ✅

packages/shared/src/components/search/SearchResults/SearchResultsLayout.tsx
ilasw: Updated! ✅

packages/shared/src/components/search/SearchResults/SearchResultsLayout.tsx
ilasw: Because [on figma](https://www.figma.com/design/nmfWPS7x3kzLUvYMkBx2kW/daily.dev---Dev-Mode?node-id=13419-45716&t=I5YN0t9vaVkyaSw4-0) they are not rounded, should I skip on this detail?

packages/shared/src/lib/featureManagement.ts
ilasw: Not already, was waiting for some reviews on PR first 👀

ilasw: Approved ✅

packages/shared/src/components/widgets/WidgetCard.tsx
rebelchris: Ah ok, all good I like this was surprised to learn we didn't have it 😂

PR #3421, opened 2024-08-02
fix: page paddings, breadcrumb a11y, FeedNav last item 

ilasw:
## Changes
- Fixed page wrapper paddings;
- Fixed `BreadCrumb` accessibility;
- Fixed mobile `FeedNav` unreachable last item;

image

## Events

## Experiment

## Manual Testing

> [!CAUTION]

Pages to test:
- Leaderboard: https://mi-460-align-breadcrumbs-in-all-pages.preview.app.daily.dev/users
- Sources: https://mi-460-align-breadcrumbs-in-all-pages.preview.app.daily.dev/sources
- Tags: https://mi-460-align-breadcrumbs-in-all-pages.preview.app.daily.dev/tags
- Explore: https://mi-460-align-breadcrumbs-in-all-pages.preview.app.daily.dev/posts

### Preview domain
https://mi-460-align-breadcrumbs-in-all-pages.preview.app.daily.dev


packages/shared/src/components/header/BreadCrumbs.tsx
```diff
@@ -1,31 +1,44 @@
...
+            icon={<HomeIcon secondary />}
+            tag="a"
+            href={process.env.NEXT_PUBLIC_WEBAPP_URL}
+            size={ButtonSize.XSmall}
+          />
+          <span aria-hidden>/</span>
+        </li>
+        <li className="flex flex-row items-center gap-1 px-2 font-bold text-text-primary">
+          <a
+            href={pathname}
```
rebelchris, line 34: Shouldn't the current one not be a link as it will just reload?

packages/webapp/pages/sources/index.tsx
```diff
@@ -102,7 +103,7 @@ const SourcesPage = ({
   return (
     <>
       <NextSeo {...seo} />
-      <main className="py-6 tablet:px-4 laptop:px-10">
+      <PageWrapperLayout className="py-6">
```
rebelchris, line 106: Should this still have the py-6 then?

packages/webapp/pages/tags/index.tsx
```diff
@@ -123,7 +124,7 @@ const TagsPage = ({
   return (
     <>
       <NextSeo {...seo} />
-      <main className="flex flex-col gap-4 p-0 tablet:px-10 tablet:py-6">
+      <PageWrapperLayout className="flex flex-col gap-4">
```
rebelchris, line 127: Same here does it need the gap etc?

rebelchris (commented):
More questions, probably all fine 🙏

packages/shared/src/components/header/BreadCrumbs.tsx
ilasw: I've put the `pointer-events:none;` for not making it clickable, but should be present for semantic. 
But this is optional, I can remove it if this make troubles!

packages/webapp/pages/sources/index.tsx
ilasw: Due to internal layout we need the padding here for mobile because the button end really near to the header 
image

packages/webapp/pages/tags/index.tsx
ilasw: I can move the `flex flex-col gap-4` to a child element, but it is needed for the page layout 👀

packages/webapp/pages/sources/index.tsx
rebelchris: Ah i see, mainly wanted to avoid it having some impact on breadcrumbs (same as below basically)

packages/shared/src/components/header/BreadCrumbs.tsx
rebelchris: Not sure just seems pointless then right?
Why is it semantic needed if it doesn't do anything?

ilasw: Just for having all the `li` with a `a`, and keeping them consistent through screen readers. But I'm removing since is optional 👍

rebelchris (approved):

PR #3422, opened 2024-08-03
chore: cleanup content types

sshanzel:
## Changes
- Clean up content types as the experiment won.
- Introduce a new component called `CardCheckbox` inspired by the CustomCheckbox from this initiative.

Preview:

https://github.com/user-attachments/assets/1745573a-4e11-48eb-8ab9-fb27c238f84e

The borders look a little bit weird from the recording above, but doing a screenshot reflects the actual colors like below:

Screenshot 2024-08-04 at 5 55 07 AM

## Events

## Experiment

## Manual Testing

> [!CAUTION]

MI-451


packages/shared/src/components/typography/Typography.tsx
```diff
@@ -62,6 +63,7 @@ export function Typography<TagName extends AllowedTags>({
   className,
   ...props
 }: TypographyProps<TagName>): ReactElement {
-  const Tag = classed(tag, className, type, bold && 'font-bold', color);
+  const classes = classNames(className, type, bold && 'font-bold', color);
+  const Tag = classed(tag, classes);
```
sshanzel, line 67: There's an issue with the `classed` component when the first parameter `className` is null or undefined. If that is the case, the following classes are not being applied.

packages/shared/src/components/fields/CardCheckbox.tsx
```diff
@@ -0,0 +1,72 @@
+import React, { InputHTMLAttributes, ReactElement } from 'react';
```
sshanzel, line 1: Things would have been simpler without the gradient borders 😒

packages/shared/src/components/typography/Typography.tsx
rebelchris: wow good find!
CC: @capJavert just so you know, maybe this was the case 🤯

rebelchris (approved):
Looks good to me!

omBratteng (approved):

packages/shared/src/components/typography/Typography.tsx
sshanzel: Didn't want to touch the `classed` file yet, but we can maybe write a ticket for it if anyone wants 😅

PR #3424, opened 2024-08-04
fix: add provided bg color to ad in post page

idoshamun:
.

### Preview domain
https://ad-bg-color.preview.app.daily.dev


omBratteng (approved):

packages/shared/src/components/comments/AdAsComment.tsx
```diff
@@ -81,6 +81,7 @@ export const AdAsComment = ({ postId }: AdAsCommentProps): ReactElement => {
         size={ProfileImageSize.Large}
         user={{ id: providerId, username: source, image }}
         fallbackSrc={cloudinary.post.imageCoverPlaceholder}
+        style={{ backgroundColor: data?.backgroundColor }}
```
rebelchris, line 84: Its only something we want for comment ads?

idoshamun: this is the only comment we got from carbon so i guess yes 😓

PR #3430, opened 2024-08-05
fix: removal of tags panel after upvoting a downvoted post

sshanzel:
## Changes
- When toggling the upvote from a downvote state, we should close the tags panel.

## Events

## Experiment

## Manual Testing

> [!CAUTION]

MI-474

### Preview domain
https://mi-474.preview.app.daily.dev


packages/shared/src/components/cards/list/ActionButtons.tsx
```diff
@@ -53,6 +53,15 @@ export default function ActionButtons({
     await onDownvoteClick?.(post);
   };
 
+  const onToggleUpvote = () => {
+    onUpvoteClick?.(post);
+    setUserUpvoted(true);
+
+    if (post.userState?.vote === UserVote.Down && !!showTagsPanel) {
+      onClose(true);
+    }
```
capJavert, line 62: Since `onUpvoteClick` is async it would always execute after this `if` but I think it makes more sense to move this part at the top of the function so you first check the current state and then execute upvote.

capJavert (approved):
Nice catch!

packages/shared/src/components/cards/list/ActionButtons.tsx
sshanzel: Yeah, sure, we can do that.

PR #3432, opened 2024-08-06
fix: clicking on block after downvoting on a card

ilasw:
## Changes
- added tooltip that explain how to block source that prevents unexpected `No topics were blocked`

With no selection:
image

With selected tag:
image

Now this appear only when closing the panel using the "X" button:
image

## Events

## Experiment

## Manual Testing

> [!CAUTION]

### Preview domain
https://mi-473-block-after-downvote.preview.app.daily.dev


packages/webapp/__tests__/PostPage.tsx
```diff
@@ -527,48 +527,48 @@ it('should show author onboarding when the query param is set', async () => {
...
-    postsEngaged: {
-      id: 'asd',
-      numUpvotes: 15,
-      numComments: 0,
-    },
-  });
-  const el = screen.queryByTestId('statsBar');
-  expect(el).not.toBeInTheDocument();
-});
+ it('should update post on subscription message', async () => {
```
nensidosari, line 530: Why is this formatted, think we can avoid

nensidosari (approved):
Looks good, just a comment on the test formatting

packages/shared/src/components/post/block/PostTagsPanel.tsx
```diff
@@ -124,13 +129,23 @@ export function PostTagsPanel({
...
         </Button>
-        <Button
-          variant={ButtonVariant.Primary}
-          color={ButtonColor.Cabbage}
-          onClick={() => onBlock(tags, shouldBlockSource)}
+        <ConditionalWrapper
+          condition={isBlockDisabled}
+          wrapper={(children: ReactElement) => (
+            <SimpleTooltip content="Select at least one tag or source to block">
+              <span>{children}</span>
```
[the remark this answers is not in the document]
rebelchris: Just double check it has no side-effect.
ConditionalWrapper on a state had some funky behaviour in the past, but can't fully recall the case again.
Maybe @sshanzel remembers what this was exactly.

packages/shared/src/hooks/post/useBlockPostPanel.ts
```diff
@@ -142,22 +142,32 @@ export const useBlockPostPanel = (
...
-      setShowTagsPanel({ showTagsPanel: forceClose ? undefined : false }),
-    [setShowTagsPanel],
+    (forceClose?: boolean) => {
+      const isRemovingDownvote = typeof forceClose === 'boolean';
+
+      if (!isRemovingDownvote) {
+        displayToast(getBlockedMessage(0, false), {
+          onUndo: onDismissPermanently,
+          undoCopy: `Don't ask again`,
+        });
```
rebelchris, line 158: What's the difference between this and the one below now?

rebelchris (approved):

packages/shared/src/hooks/post/useBlockPostPanel.ts
ilasw: This will remove the `block panel` forever, the other below will revert the "just blocked" sources/tags

packages/shared/src/components/post/block/PostTagsPanel.tsx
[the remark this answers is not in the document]
ilasw: I've tested manually and should work fine as I can see

PR #3433, opened 2024-08-06
chore: cleanup experiment for search layout upgrade;

ilasw:
## Changes
- cleanup experiment for search layout upgrade;

## Events

## Experiment

## Manual Testing

> [!CAUTION]

### Preview domain
https://mi-478-cleanup-search-page.preview.app.daily.dev

Search page: https://mi-478-cleanup-search-page.preview.app.daily.dev/search?q=react


packages/shared/src/components/widgets/WidgetCard.tsx
```diff
@@ -1,19 +1,20 @@
...
 export const WidgetCard = (props: WidgetCardProps): ReactElement => {
-  const { heading, className, children } = props;
+  const { heading, className, children, ...attrs } = props;
   return (
     <section
       className={classNames(
         'flex flex-col rounded-16 border border-border-subtlest-tertiary',
         className,
       )}
+      {...attrs}
```
capJavert, line 17: What is this for?

packages/shared/src/hooks/search/useSearchResultsLayout.ts
```diff
@@ -1,33 +1,18 @@
...
-    },
-  );
-
-  const isSearchResultsUpgrade =
-    !isLoading && isSearchUpgradeExperiment && shouldShowNewLayout;
-
   return {
-    isSearchResultsUpgrade,
-    isLoading: !isTesting && shouldShowNewLayout && isLoading,
+    isSearchResultsUpgrade: shouldShowNewLayout,
```
capJavert, line 15: This should be always true now right if experiment won, so we don't need the hook at all potentially?

packages/shared/src/components/widgets/WidgetCard.tsx
ilasw: I'm using it to pass potential attributes that comes from `ComponentProps<'section'>`. 

I'm passing  `data-testid="related-tags"` in Related tags for eg.

packages/shared/src/hooks/search/useSearchResultsLayout.ts
ilasw: We still have two check to do in multiple places: `!isLaptop && feedName === SharedFeedPage.Search`. 
I can remove it but have to replace with this logic in 4 files 🤔

capJavert (approved):

packages/shared/src/hooks/search/useSearchResultsLayout.ts
nensidosari: Seems like in some pages where this is used, looks like only the isLaptop check is needed rather then checking what feed page we're in.

If we're keeping this though, I would name it isSearchPageLaptop.

packages/webapp/__tests__/SearchResultsPage.tsx
```diff
@@ -0,0 +1,85 @@
+import React from 'react';
```
nensidosari, line 1: Thank u

packages/shared/src/hooks/search/useSearchResultsLayout.ts
nensidosari, line 15: If we're keeping this though, I would name it isSearchPageLaptop.

nensidosari (approved):
Just one minor thing, else looks good

packages/shared/src/hooks/search/useSearchResultsLayout.ts
ilasw: Updated! ✅

nensidosari (approved):

PR #3439, opened 2024-08-07
feat: StreakRecoverPopup component;

ilasw:
## Changes
- Fixed types on `TestBootProvider`;
- Update `BootPopups` for add the new Recover Modal;
- Added types for the new query;
- Added `StreakRecoverPopup` with almost full test coverage (need to add the fail case but some work from be is needed);
- Added `useStreakRecover` with all the feature logic; 

> In order to test this component need to edit the be, waiting the the other PR to be merged. Just change the `boot.ts` to have the new field and add a mock query for the streak recovery.

## Events

## Experiment

## Manual Testing

> [!CAUTION]

### Preview domain
https://mi-484-reading-streak-popover.preview.app.daily.dev


packages/shared/__tests__/helpers/boot.tsx
```diff
@@ -91,7 +111,10 @@ export const TestBootProvider = ({
               }}
             >
               <SettingsContext.Provider
-                value={{ ...settingsContext, ...settings }}
+                value={{
+                  ...settingsContext,
+                  ...settings,
+                }}
```
nensidosari, line 117: Why this change

packages/shared/src/components/modals/streaks/StreakRecoverModal.tsx
```diff
@@ -0,0 +1,226 @@
...
+
+export interface StreakRecoverModalProps {
+  isOpen: boolean;
+  onRequestClose?: () => void;
+  user: LoggedUser;
+}
+
+const StreakRecoverCover = () => (
+  <div className="-mx-4 -mt-6" role="presentation" aria-hidden>
+    <svg viewBox="0 0 324 131" fill="none" xmlns="http://www.w3.org/2000/svg">
```
nensidosari, line 33: Is this and svg from product or we can get a cloudinary link for it? If we're keeping it as svg let's add it in a file.

packages/shared/src/components/modals/streaks/StreakRecoverModal.tsx
```diff
@@ -0,0 +1,226 @@
...
+    type={TypographyType.Title1}
+    data-testid="streak-recover-modal-heading"
+  >
+    Oh no! Your
+    <span className="text-accent-bacon-default"> {days} day streak </span> has
+    been broken!
+  </Typography>
+);
+
+const StreakRecoveryCopy = ({
```
nensidosari, line 93: Would this better be StreakRecoveryBody or sth like it

packages/shared/src/components/modals/streaks/StreakRecoverModal.tsx
```diff
@@ -0,0 +1,226 @@
...
+  );
+
+  return (
+    <Typography
+      className="text-center"
+      tag={TypographyTag.P}
+      type={TypographyType.Body}
+      data-testid="streak-recovery-copy"
+    >
+      {isFree && (
```
nensidosari, line 118: maybe you can define these texts above and then conditionally render here

isFreeText, canRecoverText, noRecoverText

packages/shared/src/components/modals/streaks/StreakRecoverModal.tsx
nensidosari: I think all these modal components can be set in their own files

packages/shared/src/components/modals/streaks/StreakRecoverModal.tsx
```diff
@@ -0,0 +1,226 @@
...
+  props: StreakRecoverModalProps,
+): ReactElement => {
+  const { isOpen, onRequestClose, user } = props;
+
+  const id = useId();
+  const { recover, hideForever, onClose, onRecover } = useStreakRecover({
+    onRequestClose,
+  });
+
+  if (!user || !recover.canDo || recover.isLoading || recover.isDisabled) {
```
nensidosari, line 173: how does the `recover.canDo` and `canUserRecover` calculated below change?

packages/shared/src/hooks/streaks/useStreakRecover.ts
```diff
@@ -0,0 +1,89 @@
...
+    hideForever: {
+      isChecked: hideForever,
+      toggle: toggleHideForever,
+    },
+    onClose,
+    onRecover,
+    recover: {
+      ...data?.recoverStreak,
+      isLoading,
+      isDisabled: checkHasCompleted(ActionType.DisableReadingStreakRecover),
```
nensidosari, line 84: Please use isActionsFetched as well from useActions(), we've seen some weird edge cases with these
Sth like:
 `!isActionsFetched || checkHasCompleted(ActionType.DisableReadingStreakRecover)`

nensidosari (commented):
Looks good, minor comments. Will wait to approve till we have backend ready

packages/shared/src/components/modals/streaks/StreakRecoverModal.tsx
ilasw: `recover.canDo` is coming from be, `canUserRecover` is much more `canUserAffordTheCost`, maybe we want to change the name of that variable 👀

packages/shared/src/components/modals/streaks/StreakRecoverModal.tsx
ilasw: I didn't moved those files because the component is still below 250 lines of code (even with the SVG). 
I divided it in component to make easy to find out the business logic, is this confusing?

packages/shared/src/components/modals/streaks/StreakRecoverModal.tsx
nensidosari: Well for images we usually use links, this is more like an icon sth, so maybe we can just save it in its own file

packages/shared/src/components/modals/streaks/StreakRecoverModal.tsx
nensidosari: recover.canDo -> comes from be yes but what does it mean in this case?

ilasw: We use it to check if the user still can do the recover at the moment of the popup showing, if is false the component just return null. 
With @sshanzel today we saw that this may be a good point for block popup from showing and update the `Alerts` table for not making too much controls on cron.

packages/shared/src/components/modals/streaks/StreakRecoverModal.tsx
nensidosari: I love the split in components, just thought maybe would be better readable in separate files. But don't have a very strong opinion on this either, some people prefer to not move between files rather have components in the same file since it's just components used here. 

So I'll leave this up to u

packages/shared/src/components/modals/streaks/StreakRecoverModal.tsx
nensidosari: But recover.canDo also checks if user can afford the cost right? we're checking here again to help any edge case where user has less reputation then when we fetched the data recover.canDo?

ilasw: No currently do not check if the user can afford the cost, only check if is eligible, just because we want to show the popup with "you can recover the streak but currently not have enough point" if the user cannot afford it. 
Maybe `recover.isEligible` is a better name?

nensidosari: hmm maybe, both give the idea that that check might already be complete that's why the confusion on my side. What makes the user eligible in this case? The fact he only broke the streak today (1day - 24h?)? anything else?

packages/shared/__tests__/helpers/boot.tsx
```diff
@@ -35,18 +36,35 @@ interface TestBootProviderProps {
   gb?: GrowthBook;
 }
 
-export const settingsContext: Partial<SettingsContextData> = {
-  spaciness: 'eco',
+export const settingsContext: SettingsContextData = {
```
rebelchris, line 39: One day we should really make a neat settings wrapper else everytime there's something new we need to add it in 100 tests

packages/shared/src/components/modals/streaks/StreakRecoverModal.tsx
rebelchris: It should be an image I think actually it's cheaper to load from cloudinary then from vercel :)

packages/shared/src/components/modals/streaks/StreakRecoverModal.tsx
```diff
@@ -0,0 +1,229 @@
...
+}: Record<'cost', number> & ButtonProps<'button'>) => (
+  <Button
+    {...props}
+    className="relative"
+    variant={ButtonVariant.Primary}
+    size={ButtonSize.Large}
+    data-testid="streak-recover-button"
+  >
+    Restore my streak
+    <span className="absolute -right-0.5 -top-0.5 flex items-center whitespace-nowrap rounded-bl-14 rounded-tr-14 bg-accent-onion-default px-1.5 py-1 text-text-primary">
```
rebelchris, line 153: Why do we need all this?

packages/shared/src/components/modals/streaks/StreakRecoverPopup.spec.tsx
```diff
@@ -0,0 +1,323 @@
...
+
+  // expect mutation to be called
+  expect(mutationCalled).toBeTruthy();
+
+  // expect success message
+  const alert = await screen.findByRole('alert');
+  expect(alert).toContainHTML('Lucky you! Your streak has been restored');
+});
+
+// it('Should show error message on recover fail', async () => {
```
rebelchris, line 271: this one can be removed?

rebelchris (changes requested):
Minor comments.
General sub comment, in the test you have a lot of comments, we can pretty much omit all of them as the tests seem very explanatory already.

nensidosari:
Except PR comments is there anything that blocks us from merging this PR before having BE functionality in?

cc. @ilasw

packages/shared/src/components/modals/streaks/StreakRecoverModal.tsx
ilasw: User can see the popup if lost streak in the past day, but can recover only if have enough points. 
We can differentiate with a `recover.isVisible`? But is that name ok from the be?

nensidosari: I wouldn't worry at this point anymore, think what we have is ok at the moment. 👍

packages/shared/src/components/modals/streaks/StreakRecoverModal.tsx
ilasw: Yes, this is the style for this badge 👀 
image

packages/shared/src/components/modals/streaks/StreakRecoverPopup.spec.tsx
ilasw: I commented because I wanted to ask product team if add the error alert, but forgot to ask in the end 😅

nensidosari (approved):
Looks good, nth to add. Just might need some testing once we have BE in place

rebelchris (changes requested):
Minor comments,
Main question, was there not also a option to do it from the streak popup? (when clicking on the day lost) (can't fully recall)

packages/shared/src/hooks/notifications/useStreakRecoverModal.ts
```diff
@@ -0,0 +1,28 @@
+import { useEffect } from 'react';
+import { useRouter } from 'next/router';
+import { useLazyModal } from '../useLazyModal';
+import { useAuthContext } from '../../contexts/AuthContext';
+import { LazyModal } from '../../components/modals/common/types';
+
+export const useStreakRecoverModal = (): void => {
```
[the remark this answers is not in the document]
ilasw: Yes, click on notification will make the router change and adds the `streak_restore` query string, so we can check on this one!

packages/shared/src/components/modals/streaks/StreakRecoverModal.tsx
```diff
@@ -0,0 +1,207 @@
...
+);
+
+const StreakRecoveryCopy = ({
+  reputation,
+  cost,
+}: Record<'reputation' | 'cost', number>) => {
+  const isFree = cost === 0;
+  const canRecover = reputation >= cost;
+  const reputationLink = (
+    <a
```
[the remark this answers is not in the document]
ilasw: I've tried to use the button at first but it applies too much padding/styles not needed, so moved to `<a>`. So i preferred to use a link with 2 classes over using !important in classes, do you think that I should revert?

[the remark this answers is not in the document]
rebelchris: Nah let's leave it then :) (I thought it probably had a reason)

packages/shared/src/hooks/notifications/useStreakRecoverModal.ts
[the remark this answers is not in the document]
rebelchris: It route changes only on the notification page?

packages/shared/src/components/modals/streaks/StreakRecoverModal.tsx
```diff
@@ -0,0 +1,207 @@
...
+  onRequestClose: () => void;
+  user: LoggedUser;
+}
+
+const StreakRecoverCover = () => (
+  <div className="overflow-hidden tablet:-mx-4" role="presentation" aria-hidden>
+    <img
+      alt="Broken reading streak"
+      className="h-auto w-full object-contain"
+      height="131"
```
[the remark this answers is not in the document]
ilasw: We don't have a design for mobile, but added width/height to prevent layout shift once the image is loading (it is lazy for now for performance). Should i remove them?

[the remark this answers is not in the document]
rebelchris: It just sounded like probably more prone to issues down the line then having it dynamic.
(Which we do now everywhere else from what I can see)

CC: @nensidosari Not really having a strong opinion, just want to avoid accidentally missing some view)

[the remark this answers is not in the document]
nensidosari: True, we never set them usually. I would suggest to remove for now, if we see layout shift is becoming an issue in this case we can consider it

rebelchris (approved):

PR #3440, opened 2024-08-07
fix: truncate preview subdomain to max 32 chars

omBratteng:
### Preview domain
https://fix-preview-max-length-of-the-su.preview.app.daily.dev


sshanzel (approved):
🚀

PR #3443, opened 2024-08-08
refactor: search layout as list only is experiment now

ilasw:
## Changes
- search layout as list only is experiment now

## Events

## Experiment

Name: `search_list_mode` 
Available options: true | false
Enrolment:
- When user lands search page from laptop

## Manual Testing

> [!CAUTION]

### Preview domain
https://mi-482-search-page-only-list.preview.app.daily.dev


packages/shared/src/hooks/useFeedLayout.ts
```diff
@@ -111,7 +114,13 @@ export const useFeedLayout = ({
...
   const { isSearchPageLaptop } = useSearchResultsLayout();
 
-  const isListMode = isSearchPageLaptop || insaneMode;
+  const { value: isSearchListModeExperiment, isLoading: isLoadingExperiment } =
+    useConditionalFeature({
+      feature: feature.searchListMode,
+      shouldEvaluate: isSearchPageLaptop,
+    });
+
+  const isListMode = isSearchListModeExperiment || insaneMode;
```
nensidosari, line 123: why not do this in the search results layout, this way we don't have to export anything like loading experiment from here?

nensidosari (approved):
Looks good. Please don't forget to send the experiment enrolment message to experiments channel for Ido to approve before merging 🙏 

One suggestion maybe since isListMode is used by every page (just to rule out any weird edge case), we can export a isSearchPageListMode only for this experiment, that depends on the experiment variable.

PR #3444, opened 2024-08-08
chore: update default og

nensidosari:
Optimise default OG image - see thread https://dailydotdev.slack.com/archives/C02AA18RD1D/p1723099768261979?thread_ts=1722759609.853999&cid=C02AA18RD1D

### Preview domain
https://chore-update-default-og.preview.app.daily.dev

### DIFF
--- a/packages/webapp/next-seo.ts
+++ b/packages/webapp/next-seo.ts
@@ -23,7 +23,7 @@ export const defaultSeo: Partial<NextSeoProps> = {
 export const defaultOpenGraph: Partial<OpenGraph> = {
   images: [
     {
-      url: 'https://daily-now-res.cloudinary.com/image/upload/v1680599225/opengraph/Open_Graph_3.png',
+      url: 'https://res.cloudinary.com/daily-now/image/upload/s--VAY5ToZt--/f_auto/v1722860388/public/daily.dev%20-%20open%20graph',
     },
   ],
 };


packages/webapp/next-seo.ts
```diff
@@ -23,7 +23,7 @@ export const defaultSeo: Partial<NextSeoProps> = {
 export const defaultOpenGraph: Partial<OpenGraph> = {
   images: [
     {
-      url: 'https://daily-now-res.cloudinary.com/image/upload/v1680599225/opengraph/Open_Graph_3.png',
+      url: 'https://res.cloudinary.com/daily-now/image/upload/s--VAY5ToZt--/f_auto/v1722860388/public/daily.dev%20-%20open%20graph',
```
capJavert, line 26: should we also add `q_auto`

capJavert (approved):

packages/webapp/next-seo.ts
idoshamun: no need in this case. also since these are signed urls you can't easily add them (fyi)

idoshamun (approved):

packages/webapp/next-seo.ts
omBratteng: I think so, it goes from 404 KiB to 154 KiB with `q_auto`

idoshamun: yes but we want to keep the quality of the image. at some point design commented about q_auto messing with the quality

nensidosari: Well it's exactly about quality. I'll leave it to what we have atm unless instructed otherwise.

capJavert: 404kb seems a bit large for og image, but yeah, if they requested it (can't see thread, it opens nothing) 🙈

PR #3450, opened 2024-08-10
fix: set hour dropdown to up when there's no push enabled

omBratteng:
## Changes

Changes the dropdown for the personalized digest time to drop up when there is no push notifications enabled.
This is a hotfix, and we should probably make it a bit more spatially aware to automagically select drop down direction

image

### Preview domain
https://hotfix-digest-dropdown.preview.app.daily.dev

### DIFF
--- a/packages/webapp/pages/account/notifications.tsx
+++ b/packages/webapp/pages/account/notifications.tsx
@@ -403,7 +403,10 @@ const AccountNotificationsPage = (): ReactElement => {
               </h3>
             </div>
             <HourDropdown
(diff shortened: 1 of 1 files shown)


rebelchris (approved):
Good for now.
Think it falls under the "we want to change context menu" category anyway.
Which includes direction aware opening etc.

omBratteng:
> Think it falls under the "we want to change context menu" category anyway.
> Which includes direction aware opening etc.

Yeah, so this is more of a quick fix so that people who doesn't have notifications enabled, can actually use the dropdown.

PR #3452, opened 2024-08-12
chore: cleanup source_notify_button experiment it won

rebelchris:
## Changes

Cleanup for source_notify_button experiment it won.

### Todo

- [x] Waiting for Luca if there was any DB action we needed to retro fix for this

## Events

## Experiment

## Manual Testing

> [!CAUTION]

AS-515 #done 

### Preview domain
https://as-515-cleanup-source-notify-but.preview.app.daily.dev


packages/shared/src/components/sources/SourceActions/SourceActionsBlock.tsx
```diff
@@ -13,10 +11,8 @@ const SourceActionsBlock = ({
   isBlocked,
   onClick,
 }: SourceActionBlockProps): ReactElement => {
-  const isNotifyExperiment = useFeature(feature.sourceNotifyButton);
-  const unblockLabel = isNotifyExperiment ? 'Unblock' : 'Follow';
-  const defaultIcon = isBlocked ? <PlusIcon /> : <BlockIcon />;
-  const icon = isNotifyExperiment ? <BlockIcon /> : defaultIcon;
+  const unblockLabel = 'Unblock';
+  const icon = <BlockIcon />;
```
ilasw, line 15: Maybe we can to put this <BlockIcon /> directly inside the `icon` prop

packages/shared/src/components/sources/SourceActions/index.tsx
```diff
@@ -1,44 +1,73 @@
 import React, { ReactElement } from 'react';
-import { SourceSubscribeButton } from './SourceSubscribeButton';
```
ilasw, line 2: Looks good but maybe we should remove even the `SourceSubscribeButton` component, now it should have no use in the codebase

ilasw (changes requested):
Not sure if the `SourceSubscribeButton` have other uses in the codebase, but remember was used only here. 
Should we have to remove this component?

ilasw (approved):
Cool! Thanks!

PR #3459, opened 2024-08-13
feat: adjust styles for squad actions

capJavert:
## Changes

Based on [thread](https://dailydotdev.slack.com/archives/C069XNYMDGV/p1723553196263239)  I adjusted styles a bit with CEO of the night. 

Before:
image

After:
image

## Events

## Experiment

## Manual Testing

> [!CAUTION]

### Preview domain
https://squad-actions-styles.preview.app.daily.dev

### DIFF
--- a/packages/shared/src/components/squads/SquadMemberShortList.tsx
+++ b/packages/shared/src/components/squads/SquadMemberShortList.tsx
@@ -41,7 +41,7 @@ function SquadMemberShortList({
       <button
         type="button"
         className={classNames(
-          'flex flex-row-reverse items-center border border-border-subtlest-secondary pl-3 hover:bg-surface-hover active:bg-theme-active',
+          'flex flex-row-reverse items-center border border-border-subtlest-secondary pl-3 pr-1 hover:bg-surface-hover active:bg-theme-active',
           className,
           roundClasses[size],
         )}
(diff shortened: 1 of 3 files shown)


packages/shared/src/components/squads/SquadMemberShortList.tsx
```diff
@@ -37,8 +41,9 @@ function SquadMemberShortList({
       <button
```
rebelchris, line 41: I guess ideally this would just become a actual Button right?
What makes it different now?

capJavert: its used in different places so I don't know, trying to just make it work for this change

capJavert: Main issue is that internal `ProfilePicture` sizes do not match sizes of Button so I would have to introduce multiple props.

rebelchris: Ah I see :( 
It's fine for now I guess, but probably something we want to fix as guidelined after this.

capJavert: For sure

nensidosari (approved):

PR #3462, opened 2024-08-14
feat: shortcuts followup experiment

capJavert:
## Changes

- experiment is now going to work for existing users
- connect existing feature flag again instead of hardcoded
- do not show the popup if user already has some top sites

## Events

## Experiment

## Manual Testing

> [!CAUTION]

<!--
If relevant, please remove the comment HTML comment tags and fill the checkboxes below

- [ ] MobileL (420px)
- [ ] Tablet (656px)
- [ ] Laptop (1020px)

- [ ] iOS (Chrome and Safari)
- [ ] Android

AS-468 #done

### Preview domain
https://as-468-shortcuts-followup.preview.app.daily.dev

### DIFF
--- a/packages/extension/src/newtab/ShortcutLinks/experiments/ShortcutLinksUIV1.tsx
+++ b/packages/extension/src/newtab/ShortcutLinks/experiments/ShortcutLinksUIV1.tsx
@@ -34,6 +34,7 @@ type ShortcutLinksV1Props = {
   shouldUseListFeedLayout: boolean;
   showTopSites: boolean;
   toggleShowTopSites: () => void;
+  hasCheckedPermission: boolean;
 };
 
 function ShortCutV1Placeholder({
@@ -121,12 +122,18 @@ export function ShortcutLinksUIV1(props: ShortcutLinksV1Props): ReactElement {
     shouldUseListFeedLayout,
     showTopSites,
     toggleShowTopSites,
+    hasCheckedPermission,
   } = props;
 
   const { checkHasCompleted, isActionsFetched, completeAction } = useActions();
   const { githubShortcut } = useThemedAsset();
 
+  if (!hasCheckedPermission) {
+    return null;
+  }
+
   if (
+    !shortcutLinks?.length &&
     showTopSites &&
     isActionsFetched &&
     !checkHasCompleted(ActionType.FirstShortcutsSession)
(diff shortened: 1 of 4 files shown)


packages/extension/src/newtab/ShortcutLinks/experiments/ShortcutLinksUIV1.tsx
```diff
@@ -121,12 +122,18 @@ export function ShortcutLinksUIV1(props: ShortcutLinksV1Props): ReactElement {
...
     toggleShowTopSites,
+    hasCheckedPermission,
   } = props;
 
   const { checkHasCompleted, isActionsFetched, completeAction } = useActions();
   const { githubShortcut } = useThemedAsset();
 
+  if (!hasCheckedPermission) {
+    return null;
+  }
```
capJavert, line 133: This is basically loading state for top sites so I just don't render anything until browser check has passed to avoid flash of popup.

nensidosari (approved):

packages/shared/src/hooks/utils/useShortcuts.ts
```diff
@@ -11,13 +15,15 @@ interface UseShortcuts {
  * Keywords: shortcutsUI, onboarding_most_visited
  */
 export const useShortcuts = (): UseShortcuts => {
-  const { user } = useAuthContext();
+  const { isLoggedIn, user } = useAuthContext();
+  const { value: shortcutsUIVersion } = useConditionalFeature({
+    feature: feature.shortcutsUI,
+    shouldEvaluate: isLoggedIn && isExtension,
```
rebelchris, line 21: but won't this "re-set" people who already where proven to have the experiment as a win?
Not sure how to fix it, but something as in everyone after set date should also ALWAYS have it right?

rebelchris (changes requested):
Just not sure on the enrollment yet.
(Also do check in #experiment channel please)

packages/shared/src/hooks/utils/useShortcuts.ts
nensidosari: That could be part of enrolment from GB

rebelchris: In that case we need to be very explicit towards experiment owners :) 
So yeah maybe let's raise the experiment on slack and see what will be advised

rebelchris (approved):
Guess we good to go, just ensure the rules are matching before merging in GB :)

PR #3463, opened 2024-08-16
chore: add search debounce

rebelchris:
## Changes

Noticed our old useDebounce hook is more useDebounceFn,
in this case I just want a debounced value so added the normal version as well.

## Events

## Experiment

## Manual Testing

> [!CAUTION]

### Preview domain
https://chore-search-debounce.preview.app.daily.dev

### DIFF
--- a/packages/shared/src/hooks/search/useSearchProviderSuggestions.ts
+++ b/packages/shared/src/hooks/search/useSearchProviderSuggestions.ts
@@ -11,6 +11,8 @@ import {
   defaultSearchSuggestionsLimit,
   minSearchQueryLength,
 } from '../../graphql/search';
+import useDebounce from '../useDebounce';
+import { defaultSearchDebounceMs } from '../../lib/func';
 
 export type UseSearchProviderSuggestionsProps = {
   limit?: number;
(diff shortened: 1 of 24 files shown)


packages/shared/src/hooks/useDebounceFn.ts
```diff
@@ -0,0 +1,56 @@
+import { useCallback, useEffect, useRef, useMemo } from 'react';
+
+export type StartFn<T, R = void> = (params?: T) => R;
+export type CancelEvent = () => void;
+
+export default function useDebounceFn<T = unknown>(
```
capJavert, line 6: This is just the old file renamed?

capJavert: Non blocking but to avoid this many changes the new hook could have been `useDebouncedValue`.

packages/shared/src/hooks/search/useSearchProviderSuggestions.ts
```diff
@@ -28,15 +30,16 @@ export const useSearchProviderSuggestions = ({
 }: UseSearchProviderSuggestionsProps): UseSearchProviderSuggestions => {
   const { user } = useAuthContext();
   const { getSuggestions } = useSearchProvider();
+  const debouncedQuery = useDebounce(query, defaultSearchDebounceMs);
```
capJavert, line 33: I would put it up at something a bit bigger, like 600ms, since 200ms is not doing anything for someone typing more leisurely or on mobile. We can even have different debounce on desktop/mobile.

packages/shared/src/hooks/useDebounceFn.ts
rebelchris: If you prefer I can do that?

capJavert: Since you already did the work I am ok with this as well 👍

packages/shared/src/hooks/search/useSearchProviderSuggestions.ts
capJavert: Yeah, 500 should be better as well. Only thing it will introduce a bit of latency for panel to show/update suggestions first but personally for me personally it is not an issue.

capJavert (approved):

packages/shared/src/hooks/search/useSearchProviderSuggestions.ts
rebelchris: It will actually hit on the first letter, then debounce the extra ones (due to query key debounce)

Thought it was a nice bonus 😂

PR #3469, opened 2024-08-20
feat: post translations

capJavert:
## Changes

- [x] add language to registration forms
- [x] set gqlClient language header from boot
- [x] update/manage/invalidate query keys to include language
- [x] add update language to feed settings

https://github.com/dailydotdev/daily-api/pull/2162

## Events

| Change | change settings  | target_type: language, target_id: {langCode} |

## Experiment

- `post_title_language`

Don't forget to send a message to the [#experiments](https://dailydotdev.slack.com/archives/C02JAUF8HJL/p1715175315620999) channel, following the template in slack, and adding a link to the message here.

> [!IMPORTANT]

## Manual Testing

> [!CAUTION]

### Preview domain
https://post-translations.preview.app.daily.dev


packages/shared/src/components/filters/ContentTypesFilter.tsx
```diff
@@ -1,40 +1,172 @@
...
+          RequestKey.Bookmarks,
+          RequestKey.ReadingHistory,
+          RequestKey.RelatedPosts,
+          RequestKey.SourceFeed,
+          RequestKey.SourceMostUpvoted,
+          RequestKey.SourceBestDiscussed,
+          RequestKey.TagFeed,
+          RequestKey.TagsMostUpvoted,
+          RequestKey.TagsBestDiscussed,
+        ];
```
capJavert, line 76: Instead of changing query keys all over the place I opted into invalidating data that contains posts. This will mean that when user again goes to some feed or post the data will simply refetch then and show new translated titles if exist.

packages/shared/src/lib/query.ts
```diff
@@ -135,6 +135,12 @@ export enum RequestKey {
   SlackChannels = 'slack_channels',
   UserIntegrations = 'user_integrations',
   UserSourceIntegrations = 'user_source_integrations',
+  SourceFeed = 'sourceFeed',
+  SourceMostUpvoted = 'sourceMostUpvoted',
+  SourceBestDiscussed = 'sourceBestDiscussed',
+  TagFeed = 'tagFeed',
+  TagsMostUpvoted = 'tagsMostUpvoted',
+  TagsBestDiscussed = 'tagsBestDiscussed',
```
capJavert, line 143: Just some missing constants

packages/shared/src/contexts/BootProvider.tsx
```diff
@@ -223,6 +224,19 @@ export const BootDataProvider = ({
...
+    const partialUser = user as Partial<LoggedUser>;
+
+    gqlClient.setHeader(
+      'content-language',
+      partialUser?.language || ContentLanguage.English,
+    );
+
+    return () => {
+      gqlClient.setHeader('content-language', ContentLanguage.English);
+    };
```
capJavert, line 237: This sets the language globally for all gql requests made with a client.

packages/shared/src/components/auth/RegistrationForm.tsx
```diff
@@ -247,6 +249,17 @@ export const RegistrationForm = ({
           }
           saveHintSpace
         />
+        <LanguageDropdown
```
rebelchris, line 252: This shouldn't be behind exp?

packages/shared/src/components/auth/SocialRegistrationForm.tsx
```diff
@@ -260,6 +263,18 @@ export const SocialRegistrationForm = ({
           hint={experienceLevelHint}
           saveHintSpace
         />
+        <LanguageDropdown
```
rebelchris, line 266: Same not behind exp?

packages/shared/src/components/filters/ContentTypesFilter.tsx
```diff
@@ -1,40 +1,172 @@
...
+  const queryClient = useQueryClient();
   const { advancedSettings, isLoading } = useFeedSettings();
   const { selectedSettings, onToggleSettings } = useAdvancedSettings();
+  const { user, updateUser } = useAuthContext();
+  const { displayToast } = useToastNotification();
+  const { logEvent } = useLogContext();
 
   const videoSetting = getVideoSetting(advancedSettings);
 
+  const { mutate: onLanguageChange } = useMutation(
```
rebelchris, line 38: since it's quite a bloat would consider it in a hook,. but not highly important

packages/shared/src/components/filters/ContentTypesFilter.tsx
rebelchris: Hmm don't we prefix them with feed or something?
That would have been smart 😂

packages/shared/src/contexts/BootProvider.tsx
```diff
@@ -223,6 +224,11 @@ export const BootDataProvider = ({
     setCachedBootData((cachedData) => updateLocalBootData(cachedData, { exp }));
   }, []);
 
+  gqlClient.setHeader(
+    'content-language',
+    (user as Partial<LoggedUser>)?.language || ContentLanguage.English,
+  );
```
capJavert, line 230: This sets the language for gql client for all requests. This is earliest I could push it so we load the language as soon as we know it from boot data.

packages/shared/src/components/filters/ContentTypesFilter.tsx
```diff
@@ -1,40 +1,172 @@
...
+              >
+                Choose your preferred language for the post titles on the feed
+              </Typography>
+            </div>
+            <LanguageDropdown
+              className={{ container: 'w-full max-w-60' }}
+              name="language"
+              defaultValue={user.language}
+              onChange={(value) => onLanguageChange(value)}
+              icon={null}
```
rebelchris, line 123:
```suggestion
```

rebelchris: Surely we can omit this?

packages/shared/src/components/filters/ContentTypesFilter.tsx
capJavert: Bad thing here is that if user changes the language in other tab it will not invalidate but for experiment I think its ok. If this wins and we want to add more translations eg. to content we have to introduce `lang` to query keys so they are invalidated automatically.

packages/shared/src/components/filters/ContentTypesFilter.tsx
```diff
@@ -1,40 +1,172 @@
...
-            selectedSettings[videoSetting.id] ??
-            videoSetting.defaultEnabledState
-          }
-          onToggle={() =>
-            onToggleSettings(videoSetting.id, videoSetting.defaultEnabledState)
-          }
-        >
-          {videoSetting.title}
-        </FilterCheckbox>
+    <div className="flex flex-col gap-4 px-6">
```
rebelchris, line 97: Overall space here feels too big, should only be 24px allround.
Screenshot 2024-08-21 at 15 52 17

packages/shared/src/components/profile/LanguageDropdown.tsx
```diff
@@ -0,0 +1,115 @@
...
+
+const LanguageDropdownDefault = ({
+  className = {},
+  onChange,
+  hint = '',
+  valid = true,
+  defaultValue,
+  name,
+  saveHintSpace = false,
+  icon = defaultIcon,
```
rebelchris, line 33: Ah I see 😅

packages/shared/src/components/profile/LanguageDropdown.tsx
```diff
@@ -0,0 +1,115 @@
...
+    >
+      <Dropdown
+        className={{
+          label: classNames(labelClassName, 'typo-body'),
+          button: classNames(
+            buttonClassName,
+            '!px-3',
+            !open &&
+              !valid &&
+              '!shadow-[inset_0.125rem_0_0_var(--status-error)]',
```
rebelchris, line 65: Lol we need a better way to do this.
But basically it just copies from exp level dropdown right?

Can we log a backlog item to consolidate the styling otherwise? *(not important/urgent)

packages/shared/src/contexts/BootProvider.tsx
rebelchris: But why in a useEffect?
Can't we solve it without just on the fact if `provider in user` or something?

Just want to avoid 1000s of rerenders on bootprovider where possible.

rebelchris (changes requested):
Nice once this wins we should add ticket for OS community to convert language dropdown into storybook :) 

Just some minor questions, think you can validate most.

packages/shared/src/contexts/BootProvider.tsx
capJavert: Its no longer in useEffect, check again. I did it inline because useEffect would be too late for extension.

packages/shared/src/components/filters/ContentTypesFilter.tsx
capJavert: By default it has the icon and here it does not. I can reverse it if you want?

packages/shared/src/components/filters/ContentTypesFilter.tsx
capJavert: I left it as is on other filter views. Top one is actually a bit smaller the other. I can ping design but if you agree I can do engineering decision and just put 24 as you say on all of them?

packages/shared/src/contexts/BootProvider.tsx
capJavert: https://github.com/dailydotdev/apps/pull/3469#discussion_r1725091836

packages/shared/src/components/profile/LanguageDropdown.tsx
capJavert: yes :hidethepain:

packages/shared/src/components/profile/LanguageDropdown.tsx
capJavert: AS-533

packages/shared/src/components/auth/RegistrationForm.tsx
rebelchris: Ah right sorry didn't go back to this comment after seeing that.

packages/shared/src/components/filters/ContentTypesFilter.tsx
rebelchris: Ah just put 24 on all for now then :)

packages/shared/src/contexts/BootProvider.tsx
rebelchris: This looks good to me!

packages/shared/src/components/filters/ContentTypesFilter.tsx
capJavert: > Hmm don't we prefix them with feed or something?
That would have been smart 😂

I though we did and then I seen https://github.com/dailydotdev/apps/pull/3469#discussion_r1725077628

packages/shared/src/components/filters/ContentTypesFilter.tsx
capJavert: Moved here 29b6d9fbf09b1fa3aa9bb4eff2a714e7f0700b81

packages/shared/src/components/filters/ContentTypesFilter.tsx
capJavert: https://github.com/dailydotdev/apps/pull/3474

rebelchris (approved):
Looks good to me :)

rebelchris (approved):

PR #3471, opened 2024-08-21
fix: give source handle priority in redirect

rebelchris:
## Changes

Give source handle priority in redirect

## Events

## Experiment

## Manual Testing

> [!CAUTION]

AS-530 #done 

### Preview domain
https://as-530-source-handle-prio.preview.app.daily.dev

### DIFF
--- a/packages/shared/src/components/search/SearchPanel/SearchPanelSourceSuggestions.tsx
+++ b/packages/shared/src/components/search/SearchPanel/SearchPanelSourceSuggestions.tsx
@@ -71,7 +71,7 @@ export const SearchPanelSourceSuggestions = ({
   });
 
   const onSuggestionClick = (suggestion: SearchSuggestion) => {
-    const source = suggestion.id || suggestion.subtitle.toLowerCase();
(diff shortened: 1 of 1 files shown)


capJavert (approved):

packages/shared/src/components/search/SearchPanel/SearchPanelSourceSuggestions.tsx
```diff
@@ -71,7 +71,7 @@ export const SearchPanelSourceSuggestions = ({
   });
 
   const onSuggestionClick = (suggestion: SearchSuggestion) => {
-    const source = suggestion.id || suggestion.subtitle.toLowerCase();
+    const source = suggestion.subtitle.toLowerCase() || suggestion.id;
```
capJavert, line 74: I think `suggesstion.id` will never evaluate now because if `suggestion.subtitle` was undefined it would just break when calling `toLowerCase`, so maybe just?
```suggestion
    const source = suggestion.subtitle?.toLowerCase() || suggestion.id;
```

PR #3484, opened 2024-08-27
chore: attempt to fix flaky test

rebelchris:
## Changes

Since our tests run concurrent I actually don't know which one cause the fail.
In general we should avoid having 2 same named tests,  so this should fix that.

At very best we get more insights in which of the 2 is flaky.

## Events

## Experiment

## Manual Testing

> [!CAUTION]

### Preview domain
https://chore-attempt-flaky-fix.preview.app.daily.dev


packages/shared/src/components/Feed.spec.tsx
```diff
@@ -410,7 +410,7 @@ it('should send add bookmark mutation', async () => {
   await waitFor(() => expect(mutationCalled).toBeTruthy());
 });
 
-it('should send remove bookmark mutation', async () => {
+it('should send remove bookmark mutation (feed)', async () => {
```
capJavert, line 413: Better way is to use describe as we do in newer files and then in test results you can see under which one it ran.

capJavert (approved):

PR #3487, opened 2024-08-28
Mi 486 horizontal scroll

nensidosari:
## Events

## Experiment

## Manual Testing

> [!CAUTION]

### DIFF
--- a/packages/shared/src/components/HorizontalScroll/HorizontalScroll.tsx
+++ b/packages/shared/src/components/HorizontalScroll/HorizontalScroll.tsx
@@ -15,7 +15,7 @@ interface HorizontalScrollProps {
   className?: string;
 }
 
-export default function HorizontalScroll<T>({
+export default function HorizontalScroll({
   title,
   children,
   onScroll,
--- a/packages/shared/src/components/HorizontalScroll/HorizontalScroll.tsx
+++ b/packages/shared/src/components/HorizontalScroll/HorizontalScroll.tsx
@@ -39,12 +39,12 @@ export default function HorizontalScroll<T>({
       const element = ref.current.firstElementChild as HTMLElement;
       const elementWidth = element.offsetWidth;
       const elementMarginRight = parseFloat(
-        getComputedStyle(element).marginRight,
+        getComputedStyle(element).marginRight || '0',
       );
       const elementMarginLeft = parseFloat(
-        getComputedStyle(element).marginLeft,
+        getComputedStyle(element).marginLeft || '0',
       );
-      const containerGap = parseFloat(getComputedStyle(ref.current).gap);
+      const containerGap = parseFloat(getComputedStyle(ref.current).gap || '0');
       const elementWidthWithMarginsAndGap =
         elementWidth + elementMarginRight + elementMarginLeft + containerGap;
 
@@ -133,6 +133,8 @@ export default function HorizontalScroll<T>({
           'no-scrollbar grid auto-cols-max grid-flow-col overflow-x-scroll scroll-smooth',
           className,
         )}
+        role="region"
+        aria-label="Horizontal Scroll"
       >
         {children}
       </div>
--- a/packages/shared/src/components/HorizontalScroll/HorizontalScroll.tsx
+++ b/packages/shared/src/components/HorizontalScroll/HorizontalScroll.tsx
@@ -8,6 +8,7 @@ import React, {
   RefObject,
   MouseEventHandler,
 } from 'react';
+import classNames from 'classnames';
 import { Button, ButtonVariant } from '../buttons/Button';
 import { ArrowIcon } from '../icons';
 import { useScrollManagement } from './useScrollManagement';
@@ -17,6 +18,7 @@ interface HorizontalScrollProps {
   children: ReactNode;
   onScroll?: (ref: RefObject<HTMLElement>) => void;
   onClickSeeAll?: MouseEventHandler;
+  className?: string;
 }
 
 export default function HorizontalScroll<T>({
@@ -24,6 +26,7 @@ export default function HorizontalScroll<T>({
   children,
   onScroll,
   onClickSeeAll,
+  className,
 }: HorizontalScrollProps): ReactElement {
   const ref = useRef<HTMLDivElement>(null);
   const [scrollableElementWidth, setScrollableElementWidth] =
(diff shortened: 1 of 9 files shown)


packages/shared/src/components/HorizontalScroll/useHorizontalScrollHeader.tsx
```diff
@@ -0,0 +1,127 @@
...
+  onClickSeeAll,
+  title,
+}: HorizontalScrollHeaderProps): HorizontalScrollHeaderReturn => {
+  const ref = useRef<HTMLDivElement>(null);
+  const [scrollableElementWidth, setScrollableElementWidth] =
+    useState<number>(0);
+  const [numCards, setNumCards] = useState<number>(0);
+
+  // Calculate the width of elements and the number of visible cards
+  const calculateVisibleCards = useCallback(() => {
```
rebelchris, line 36: This seems like a highly expensive calculation right?
do we really need to do it as a "seperate one" meaning, maybe we can just use the computedStyle of the main container? (as you already know how many children you render)

packages/shared/src/components/HorizontalScroll/useScrollManagement.tsx
```diff
@@ -0,0 +1,40 @@
...
+      const { scrollLeft, scrollWidth, clientWidth } = ref.current;
+
+      setIsAtStart(scrollLeft === 0);
+      setIsAtEnd(scrollLeft + clientWidth >= scrollWidth);
+    }
+  }, [ref, onScroll]);
+
+  const [debouncedOnScroll] = useDebounceFn(checkScrollPosition, 100);
+
+  useEffect(() => {
```
rebelchris, line 27: Probably yes, but this needs to be a effect?
(Eg we cant leverage useCallback as we technically only add event listener?)

rebelchris (approved):
Looks to work perfectly!
So code wise happy, just wondering if we can optimize the heavy calculation.

Also edge case (which we had already, but maybe can fix now?)
If there's exact amount of items (4/4) the right button is still active:
https://mi-486-horizontal-scroll.preview.app.daily.dev/sources/TechWithTim (the second row)

nensidosari:
Also edge case (which we had already, but maybe can fix now?)
If there's exact amount of items (4/4) the right button is still active:
https://mi-486-horizontal-scroll.preview.app.daily.dev/sources/TechWithTim (the second row)

I know 😬 I tried to fix today morning, it works ok on storybook, but on the feed it messes up a little bit because how the scrollWidth is calculated in first load. So need to think a better solution for it but didn't want to block the whole thing.

packages/shared/src/components/HorizontalScroll/useScrollManagement.tsx
capJavert: We can probably use useEventListener and I would use requestAnimationFrame instead (or in addition) to debounce to make it smooth and not impact scrolling.

packages/shared/src/components/HorizontalScroll/HorizontalScroll.tsx
```diff
@@ -0,0 +1,47 @@
...
+    <>
+      <Header />
+      <div
+        ref={ref}
+        className={classNames(
+          'no-scrollbar grid auto-cols-max grid-flow-col overflow-x-scroll scroll-smooth',
+          className,
+        )}
+        role="region"
+        aria-label="Horizontal Scroll"
```
ilasw, line 41: If we have always the `<Header />` component with the title for this section maybe we would like to use `aria-labelledby` pointing to the title. 
So user that enter into this section with a screen reader will hear something like: `Related posts` instead `Horizontal Scroll`. What do you think?

Reference: https://www.aditus.io/aria/aria-label/#aria-label-vs-aria-labelledby

nensidosari: Thank u!! Will change :D

packages/shared/src/components/HorizontalScroll/HorizontalScroll.tsx
```diff
@@ -0,0 +1,49 @@
...
+  const { ref, Header } = useHorizontalScrollHeader({
+    title,
+    onScroll,
+    onClickSeeAll,
+  });
+
+  const titleId = `horizontal-scroll-title-${Math.random()}`;
+
+  return (
+    <>
```
sshanzel, line 34: I think we should make this an actual wrapper of the header and content to make it more predictable.
```suggestion
    <div className="flex flex-col">
```

sshanzel (approved):
Don't have anything to add on top of what is already mentioned, everything looked amazing. Love the fact that we have proper tests.

Just a reminder to move the hooks to their proper folders before merging after addressing the concerns.

packages/shared/src/components/HorizontalScroll/HorizontalScroll.tsx
nensidosari: Yeah makes sense, thank u! will add

packages/shared/src/components/HorizontalScroll/HorizontalScroll.tsx
```diff
@@ -0,0 +1,49 @@
...
+  onClickSeeAll,
+  className,
+}: HorizontalScrollProps): ReactElement {
+  const { ref, Header } = useHorizontalScrollHeader({
+    title,
+    onScroll,
+    onClickSeeAll,
+  });
+
+  const titleId = `horizontal-scroll-title-${Math.random()}`;
```
ilasw, line 31: Thanks for updating! Just a small suggestion!

```suggestion
  const id = useId();
  const titleId = `horizontal-scroll-title-${id}`;
```

nensidosari:
As for this: Just a reminder to move the hooks to their proper folders before merging after addressing the concerns.

@sshanzel 

I'm a bit in doubt of wether that is the best case or not, and don't want to go against our setup. But feels like these hooks are not hooks that can be used on any occasion, they are tightly coupled to HorizontalScroll in general. So would it make sense to leave here?

Not sure what others think as well, @dailydotdev/web-team

capJavert:
@nensidosari I agree with hooks staying where they are, they are internal hooks for the component that we use only there, I did the same thing for SearchPanel hooks.

rebelchris (approved):

packages/shared/src/components/HorizontalScroll/useScrollManagement.tsx
nensidosari: Addressed feedback as @capJavert suggested

nensidosari: Based on Ante's comment

> hey all good, one thing I might have missed is that we don't really need requestAnimationFrame since you are not doing any layout manipulation directly, you are just reading values, so if not much of a hassle I think remove it to not complicate the code, sorry if going back and forth 

removing requestAnimationFrame additions

capJavert: Approved as well 🙌

PR #3490, opened 2024-08-28
refactor: create Squad

sshanzel:
## Changes
- This needs the API changes, so might be able to test it on local only.
- This is purely Create Squad page, don't mind the edit page, I will raise a separate PR for that pointing to this one.
- Tested locally both the dropdown and opening a public Squad ✅ 
- Dropdown component to have a disabled state.
- Refactored the `Details` component to be broken down into `Sections`. Some of the things you might see are purely copy-pasted.

Preview:

Screenshot 2024-08-28 at 8 35 01 PM
Screenshot 2024-08-28 at 8 34 56 PM

Mobile:
Screenshot 2024-08-28 at 8 40 17 PM
Screenshot 2024-08-28 at 8 43 52 PM
Screenshot 2024-08-28 at 8 40 26 PM

## Events

## Experiment

## Manual Testing

> [!CAUTION]

MI-493 #done

### Preview domain
https://mi-493.preview.app.daily.dev


packages/shared/src/components/squads/settings/SquadPrivacySection.tsx
```diff
@@ -0,0 +1,132 @@
+import { useInfiniteQuery } from '@tanstack/react-query';
```
sshanzel, line 1: Inspired by the deleted `useSquadPrivacyOptions`.

packages/shared/src/components/squads/settings/SquadPrivacySection.tsx
```diff
@@ -0,0 +1,132 @@
...
+            )}
+            {list && !isPrivate && (
+              <input
+                type="hidden"
+                id="categoryId"
+                name="categoryId"
+                value={list[selected]?.id}
+              />
+            )}
+          </div>
```
nensidosari, line 103: Let's move this into a component, even if we want to keep it in the same file

nensidosari (changes requested):
Looks good, just one small thing on my side. 

Blocking just until api changes are in

packages/shared/src/components/squads/settings/SquadPrivacySection.tsx
sshanzel: I tried to, but there's a lot of moving objects on this section, and decided to leave it here.

nensidosari: It could be probably within the component no? that way u don't need to pass anything just separating the code

sshanzel: Yes, that's what I hoped when I tried it at first, but most values here are used within this whole form.

sshanzel: I am doing some refactoring on the form as a whole, I will check again if something can still be extracted easily.

nensidosari: Yeah cool, no worries if it brings too much trouble

packages/shared/src/components/squads/settings/PermissionSection.tsx
```diff
@@ -0,0 +1,71 @@
...
+  );
+  const [memberInviteRole, setMemberInviteRole] = useState(
+    () => initialMemberInviteRole || SourceMemberRole.Member,
+  );
+
+  return (
+    <div className="mt-2 flex flex-col gap-4 rounded-16 border border-border-subtlest-tertiary">
+      <Typography
+        type={TypographyType.Body}
+        className="flex items-center border-b border-border-subtlest-tertiary p-4 font-bold"
```
rebelchris, line 38: bold is already a option on `Typography`

packages/shared/src/components/squads/settings/SquadPrivacySection.tsx
```diff
@@ -0,0 +1,132 @@
...
+  );
+  const isPrivate = privacy === PrivacyOption.Private;
+  const [selected, setSelected] = useState(-1);
+  const { data } = useInfiniteQuery<SourceCategoryData>(
+    generateQueryKey(RequestKey.Source, null, 'categories'),
+    ({ pageParam }) => gqlClient.request(SOURCE_CATEGORIES_QUERY, pageParam),
+  );
+  const list = useMemo(() => flattenInfiniteQuery(data, 'categories'), [data]);
+
+  useEffect(() => {
```
rebelchris, line 52: This combi seems like it might give some issues?
- query without staletime + computation + useEffect on result

Not 100% sure but seems we can do some optimizations here

packages/shared/src/components/squads/settings/common.tsx
```diff
@@ -5,7 +5,7 @@ import { IconSize } from '../../Icon';
 
 export const StatusDescription = classed(
   'p',
-  'typo-subhead text-text-tertiary',
+  'typo-callout text-text-secondary',
 );
```
rebelchris, line 9: Seeing as we already changing it, maybe go for Typography? 👀

packages/shared/src/components/squads/utils.tsx
```diff
@@ -1,14 +1,8 @@
...
-export const SquadTitle = classed(
-  'h1',
-  'text-center typo-large-title font-bold',
-);
-export const SquadSubTitle = classed(
-  'p',
-  'text-center typo-title3 text-text-tertiary',
-);
+export const SquadTitle = classed('h1', 'typo-large-title font-bold');
+export const SquadSubTitle = classed('p', 'typo-body text-text-tertiary');
```
rebelchris, line 5: Same comment here on Typography

packages/shared/src/lib/image.ts
```diff
@@ -58,8 +58,12 @@ export const cloudinary = {
     },
     imageFallback:
       'https://daily-now-res.cloudinary.com/image/upload/s--LrHsyt2T--/f_auto/v1692632054/squad_placeholder_sfwkmj',
-    createSquad:
-      'https://daily-now-res.cloudinary.com/image/upload/s--4nZR_i2V--/f_auto/v1686299591/Squads_eary_access_e3lo4u',
+    createSquad: {
+      beyondMobile:
```
rebelchris, line 62: What's beyond mobile? :jenkins-triggered:

Maybe just keep `mobile/desktop` naming think it's clear enough

packages/shared/src/lib/query.ts
```diff
@@ -299,3 +300,17 @@ export const updateReadingHistoryListPost = ({
     queryClient.setQueryData(queryKey, oldData);
   };
 };
+
+export function flattenInfiniteQuery<
```
rebelchris, line 304: More generic, do we need a infiniteQuery then if we expect to flatten the whole result anyway?
EG wouldn't we just one a once-off "all categories"?

packages/shared/src/components/squads/settings/SquadPrivacySection.tsx
```diff
@@ -0,0 +1,132 @@
...
+  isPublic = true,
+  categoryHint,
+  onCategoryChange,
+}: CategoryDropdownProps): ReactElement {
+  const [privacy, setPrivacy] = useState(
+    isPublic ? PrivacyOption.Public : PrivacyOption.Private,
+  );
+  const isPrivate = privacy === PrivacyOption.Private;
+  const [selected, setSelected] = useState(-1);
+  const { data } = useInfiniteQuery<SourceCategoryData>(
```
rebelchris, line 46: this can 100% use long staleTime actually, don't expect them to change often.

rebelchris (changes requested):

packages/shared/src/components/squads/settings/SquadPrivacySection.tsx
sshanzel: All we would care about here is the list which is already after the computation. This would not run until that changes. This just basically updates the dropdown if we pass initial value of the category but we have to wait for the list to be fetched.

packages/shared/src/lib/image.ts
sshanzel: Desktop is a bit inaccurate since it can also be applied to tablets.

packages/shared/src/components/squads/settings/SquadPrivacySection.tsx
sshanzel: Yeah, we can add stale time here.

packages/shared/src/lib/query.ts
sshanzel: Oh, right. I haven't actually have thought of that 😂

sshanzel: Though still a useful utility as often flatten results. However, in this case, we don't have to if we just use `useQuery`.

packages/shared/src/components/squads/settings/PermissionSection.tsx
sshanzel: Right. Will update this one.

packages/shared/src/components/squads/settings/SquadPrivacySection.tsx
sshanzel: My mistake. I overlooked it, I thought the comment was referring to the name/handle/description being extracted too. This was about the dropdown. You're right. We should be able to move this one.

packages/shared/src/components/squads/utils.tsx
sshanzel: I think it looked a little verbose. Might be nice if we can come up with a function similar to this `classed` func but outputs a Typography.

rebelchris: It would just be Typography?
I mean just omit this whole file and simply use `Typography` inside component?

rebelchris (changes requested):
Wow cleans up nicely!

packages/shared/src/components/squads/settings/SquadCategoryDropdown.tsx
```diff
@@ -0,0 +1,90 @@
...
+  initialCategory,
+  isDisabled,
+  categoryHint,
+  onCategoryChange,
+}: SquadCategoryDropdownProps): ReactElement {
+  const [selected, setSelected] = useState(-1);
+  const { data } = useQuery<SourceCategoryData>(
+    generateQueryKey(RequestKey.Source, null, 'categories'),
+    () => gqlClient.request(SOURCE_CATEGORIES_QUERY),
+    { staleTime: Infinity },
```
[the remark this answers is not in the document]
sshanzel: A day would most likely have the user refreshed from the create page, wouldnt they?

packages/webapp/pages/squads/new.tsx
```diff
@@ -51,24 +53,32 @@ const NewSquad = (): ReactElement => {
     <ManageSquadPageContainer>
       <NextSeo {...seo} titleTemplate="%s | daily.dev" noindex nofollow />
       <SquadDetails
-        form={{ public: false }}
+        form={{}}
```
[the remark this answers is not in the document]
sshanzel: This form is form data, not really form name.

packages/shared/src/components/squads/settings/SquadCategoryDropdown.tsx
```diff
@@ -0,0 +1,90 @@
...
+export function SquadCategoryDropdown({
+  initialCategory,
+  isDisabled,
+  categoryHint,
+  onCategoryChange,
+}: SquadCategoryDropdownProps): ReactElement {
+  const [selected, setSelected] = useState(-1);
+  const { data } = useQuery<SourceCategoryData>(
+    generateQueryKey(RequestKey.Source, null, 'categories'),
+    () => gqlClient.request(SOURCE_CATEGORIES_QUERY),
```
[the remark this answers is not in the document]
sshanzel: Limit is 100 per default on our API.

packages/shared/src/components/squads/settings/SquadCategoryDropdown.tsx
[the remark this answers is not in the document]
rebelchris: It's not cached per user in above example so would be vercel cached no? 👀
AFAIK

packages/shared/src/components/squads/settings/SquadCategoryDropdown.tsx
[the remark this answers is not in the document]
rebelchris: Is it enough? (not sure how many there are, but assume we want to show "all" right?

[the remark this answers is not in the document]
sshanzel: Yeah, we only have 10 I believe.

packages/shared/src/components/squads/settings/SquadCategoryDropdown.tsx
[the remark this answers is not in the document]
sshanzel: No. This is per user per session. Like a state in the app.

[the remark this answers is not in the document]
rebelchris: oh wait yeah it's client side invoked my bad :)

rebelchris (approved):
Perfect should be good from my side :)

nensidosari (approved):
Looks good on my side as well :)

PR #3496, opened 2024-08-29
refactor: create/edit squad form and its types

sshanzel:
## Changes
- Refactored both process to make them behave the same rather than one doing a pre-calculation, and then the other happening outside the mutation function call side.
- While refactoring, I realized some properties are not used anymore and, hence, removed (like commentary).
  - At first, I planned to introduce a separate client-side function to handle squad creation with commentary, but the mutation itself does not have the said property anymore.
- Unified the types as the original one was combining two things at once. The form data and initial squad data on edit.
  - To avoid that, we want to make it clear that anything initial data related, should come from `squad` property. Other than that, it is the form to decide.
  - Due to the previous setup, TypeScript was going crazy due to certain type usages of the Form.

## Events

## Experiment

## Manual Testing

> [!CAUTION]

### Preview domain
https://refactor-forms.preview.app.daily.dev


sshanzel:
Root cause on what triggered me to make this PR 😂 https://github.com/dailydotdev/apps/pull/3496/commits/78977915ff645ae6b9106fc9c08983318ecc2a89

packages/shared/src/components/squads/Details.tsx
```diff
@@ -49,25 +48,27 @@ const getFormData = async (
 
   const input = document.getElementById(squadImageId) as HTMLInputElement;
   const file = input.files[0];
-  const base64 = await blobToBase64(file);
 
-  return { ...current, file: base64 };
```
rebelchris, line 54: Why is this removed?

rebelchris (changes requested):
Just one query, where did this base64 stuff go now?

packages/shared/src/components/squads/Details.tsx
sshanzel: The type requirement of the mutation is a `File`, here we convert it to `string`, then before sending it, we then again turn it into a `File.

rebelchris: Hmm just make sure it doesn't break implementation (as locally the upload is not e2e testable)
And maybe considering this: https://dailydotdev.slack.com/archives/C01L0QXQJNB/p1724993128390989

rebelchris (approved):
Good to go, but just verify in prod it works with upload specific :)

sshanzel:
> Good to go, but just verify in prod it works with upload specific :)

Yeah, just waiting for the go signal if we can merge without updating the feed yet. After that, I will run some tests.

PR #3499, opened 2024-08-30
feat: report source modal

sshanzel:
## Changes
- Report source modal

## Events

## Experiment

## Manual Testing

> [!CAUTION]

MI-497

### DIFF
--- a/packages/shared/src/components/modals/report/ReportSourceModal.tsx
+++ b/packages/shared/src/components/modals/report/ReportSourceModal.tsx
@@ -1,0 +1,97 @@
+import React, { ReactElement, useRef } from 'react';
+import { useMutation } from '@tanstack/react-query';
+import { Checkbox } from '../../fields/Checkbox';
+import { ModalProps } from '../common/Modal';
+import { useLogContext } from '../../../contexts/LogContext';
+import { ReportModal } from './ReportModal';
+import {
+  CommonReportReason,
+  ReportEntity,
+  sendReport,
+  SourceReportReason,
+  SourceReportReasonType,
+} from '../../../report';
+import { Squad } from '../../../graphql/sources';
+import { useLeaveSquad } from '../../../hooks';
+import { useAuthContext } from '../../../contexts/AuthContext';
+import { LogEvent } from '../../../lib/log';
+
+interface SouceReportModalProps extends ModalProps {
+  squad: Squad;
+  onReported?: () => void;
+}
+
+const reportReasons: { value: string; label: string }[] = [
+  { value: CommonReportReason.Nsfw, label: 'NSFW or inappropriate content' },
+  { value: SourceReportReason.Spam, label: 'Spam or excessive self-promotion' },
+  { value: SourceReportReason.Bullying, label: 'Harassment and bullying' },
+  {
+    value: SourceReportReason.Hateful,
+    label: 'Hate speech and discrimination',
+  },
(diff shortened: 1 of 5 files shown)


packages/shared/src/components/modals/common.tsx
```diff
@@ -159,6 +159,13 @@ const SlackIntegrationModal = dynamic(
     ),
 );
 
+const ReportSourceModal = dynamic(
+  () =>
+    import(
+      /* webpackChunkName: "reportSquadModal" */ './report/ReportSourceModal'
```
rebelchris, line 165: Probably meant this?
```suggestion
      /* webpackChunkName: "reportSourceModal" */ './report/ReportSourceModal'
```

packages/shared/src/components/modals/report/ReportSourceModal.tsx
```diff
@@ -0,0 +1,92 @@
...
+  { value: ReportReason.Illegal, label: 'Illegal activities and scams' },
+  { value: ReportReason.Other, label: 'Other' },
+];
+
+interface SubmitReportProps {
+  reason: ReportReason;
+  comment?: string;
+}
+
+export function ReportPostModal({
```
rebelchris, line 38:
```suggestion
export function ReportSourceModal({
```

packages/shared/src/components/modals/report/ReportSourceModal.tsx
```diff
@@ -0,0 +1,92 @@
...
+  squad,
+  onReported,
+  onRequestClose,
+  ...props
+}: SouceReportModalProps): ReactElement {
+  const { logEvent } = useLogContext();
+  const { squads } = useAuthContext();
+  const inputRef = useRef<HTMLInputElement>();
+  const onLeaveSquad = useLeaveSquad({ squad });
```
rebelchris, line 47: Is it specific to squads or also sources?

packages/shared/src/report.ts
```diff
@@ -0,0 +1,68 @@
+import { gql } from 'graphql-request';
+import { EmptyResponse } from './graphql/emptyResponse';
+import { gqlClient } from './graphql/common';
+
+export enum ReportEntity {
+  Post = 'post',
+  Source = 'source',
+}
```
rebelchris, line 8: comment also exist right?

rebelchris (changes requested):
Nice work on abstracting it, mainly have some cleanup/questions.

packages/shared/src/components/modals/report/ReportSourceModal.tsx
sshanzel: Should probably become applicable to sources too, but should not matter here as we won't show the checkbox if the user is not a member, which is always going to be true.

packages/shared/src/report.ts
sshanzel: Right. Let's update comments.

packages/shared/src/components/modals/report/ReportSourceModal.tsx
rebelchris: I see, we can start with this then, and modify if we need it for regular sources.

rebelchris (approved):
Cool looks good, just make sure perhaps to try all 3 once we deploy to prod?
(Just notify Ido so he doesn't action on those)

nensidosari (approved):

PR #3502, opened 2024-09-02
chore: cleanup post_title_language experiment (won)

rebelchris:
## Changes

Experiment post_title_language won

## Events

## Experiment

## Manual Testing

> [!CAUTION]

AS-545 #done 

### Preview domain
https://as-545-cleanup-post-title-langua.preview.app.daily.dev


packages/shared/src/components/profile/LanguageDropdown.tsx
```diff
@@ -22,7 +20,7 @@ type Props = {
 
 const defaultIcon = <LanguageIcon className="ml-0 mr-1" />;
 
-const LanguageDropdownDefault = ({
+export const LanguageDropdown = ({
```
capJavert, line 23: We should also see down the line (after some more adoption?) to make language required in registration flow on API? (not blocker for this PR)

capJavert (approved):

PR #3503, opened 2024-09-02
chore: fix user short info everywhere

rebelchris:
## Changes

Noticed we not loading usershortinfo fragment everywhere
This fixes that, so we get all short info needs everyhwere

## Events

## Experiment

## Manual Testing

> [!CAUTION]

### Preview domain
https://chore-user-short-info.preview.app.daily.dev


capJavert (approved):

capJavert:
As a note (not a blocker), short info fragment starting to be bigger and bigger so it could be good to also split into another fragment for some additional info that is not required anywhere. Something to think about when we add more stuff.

PR #3506, opened 2024-09-03
fix: handle garmr retry exclusion

rebelchris:
## Changes

This will handle the garmr error and avoid retries when it happens.
Follow up will be dedicated error screen.

## Events

## Experiment

## Manual Testing

> [!CAUTION]

### Preview domain
https://fix-handle-garmr-retry.preview.app.daily.dev

### DIFF
--- a/packages/shared/src/hooks/useFeed.ts
+++ b/packages/shared/src/hooks/useFeed.ts
@@ -5,6 +5,7 @@ import {
   UseInfiniteQueryOptions,
   useQueryClient,
 } from '@tanstack/react-query';
+import { ClientError } from 'graphql-request';
 import {
   Ad,
   FeedData,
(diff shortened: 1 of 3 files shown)


packages/shared/src/hooks/useFeed.ts
```diff
@@ -138,6 +139,15 @@ export default function useFeed<T>(
     {
       refetchOnMount: false,
       ...options,
+      retry: (failureCount, error) => {
```
idoshamun, line 142: Is there any way to add globally?

rebelchris: Yeah also, but for now it's only feed and want to raise a specific `onError` here in the next iteration.
So not seeing a point to do it now if that makes sense?

rebelchris: Noticed we can just overwrite it then and use it seperate :)

idoshamun: but we also limit the number of retries. i'm afraid in case of an issue frontend will hammer api other endpoints

idoshamun (approved):

packages/shared/src/hooks/useFeed.ts
rebelchris: Limit to? I mean we can remove retries all together?

capJavert (approved):

PR #3508, opened 2024-09-03
feat: add option to make TabList button be a element

omBratteng:
## Changes

- Converts tabs on explore page to use `a` instead of `button`

### Preview domain
https://as-543-buttons-to-links.preview.app.daily.dev

### DIFF
--- a/packages/shared/src/components/tabs/TabList.tsx
+++ b/packages/shared/src/components/tabs/TabList.tsx
@@ -45,6 +45,7 @@ function TabList<T extends string = string>({
     offset: 0,
     indicatorOffset: 0,
   });
+  const isAnchor = Tag === 'a';
 
   const scrollIfNotInView = useCallback(() => {
     const { activeTabRect, offset } = dimensions;
@@ -130,6 +131,7 @@ function TabList<T extends string = string>({
               className.item,
               'relative p-2 py-4 text-center font-bold typo-callout',
               isActive ? '' : 'text-text-tertiary',
+              isAnchor && 'cursor-pointer',
             )}
             onClick={() => onClick(tab)}
             type="button"
--- a/packages/shared/src/components/tabs/TabList.tsx
+++ b/packages/shared/src/components/tabs/TabList.tsx
@@ -134,8 +134,12 @@ function TabList<T extends string = string>({
               isAnchor && 'cursor-pointer',
             )}
             onClick={() => onClick(tab)}
-            type="button"
-            role="menuitem"
+            {...(isAnchor
+              ? {
+                  'aria-label': tab,
+                  title: tab,
+                }
+              : { type: 'button', role: 'menuitem' })}
           >
             {renderedTab}
           </Tag>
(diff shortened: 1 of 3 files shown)


sshanzel (approved):

rebelchris (changes requested):
More going of visual testing.
Can we add `cursor-pointer` to hovering it?

rebelchris (approved):

packages/shared/src/components/tabs/TabList.tsx
```diff
@@ -113,7 +118,7 @@ function TabList<T extends string = string>({
         );
 
         return (
-          <button
+          <Tag
```
ilasw, line 121: If now is an anchor maybe we want to add useful attributes like `aria-label`, `title`, `href` (this should be required imo). What do you think?

packages/shared/src/components/tabs/TabList.tsx
```diff
@@ -126,13 +131,14 @@ function TabList<T extends string = string>({
               className.item,
               'relative p-2 py-4 text-center font-bold typo-callout',
               isActive ? '' : 'text-text-tertiary',
+              isAnchor && 'cursor-pointer',
```
ilasw, line 134: Non mandatory: if we have more conditional classes maybe the object structure can be more readable, wdyt?

```ts
classNames(
              className.item,
              'relative p-2 py-4 text-center font-bold typo-callout',
              { 'text-text-tertiary': !isActive, 'cursor-pointer': isAnchor},
)
```

packages/shared/src/components/tabs/TabList.tsx
```diff
@@ -126,13 +131,14 @@ function TabList<T extends string = string>({
...
             )}
             onClick={() => onClick(tab)}
             type="button"
```
ilasw, line 137: If anchor type button should not be there

ilasw (changes requested):

packages/shared/src/components/tabs/TabList.tsx
omBratteng: Removed

packages/shared/src/components/tabs/TabList.tsx
omBratteng: Don't think we'll have any more now

packages/shared/src/components/tabs/TabList.tsx
omBratteng: I've added label and title. Adding href needs a bit more work I see

ilasw: If it's hard to implement, also we can put `role="button"` and `href="#"` if the behaviour is fully JS-side

omBratteng: Had a chat with @rebelchris, we'll merge it without the href and then run it through the SEO tool to look. If it still complains, I'll refactor it so that it gets proper href.

ilasw (approved):

PR #3510, opened 2024-09-04
fix: hide popup after first close;

ilasw:
## Changes
- update alerts on modal close so the user will not see the popup again;

## Events

## Experiment

## Manual Testing

> [!CAUTION]

### Preview domain
https://fix-recover-modal-keep-show.preview.app.daily.dev

### DIFF
--- a/packages/shared/src/hooks/streaks/useStreakRecover.ts
+++ b/packages/shared/src/hooks/streaks/useStreakRecover.ts
@@ -13,6 +13,7 @@ import { gqlClient } from '../../graphql/common';
 import { useToastNotification } from '../useToastNotification';
 import { useLogContext } from '../../contexts/LogContext';
 import { LogEvent, TargetType } from '../../lib/log';
+import { useAlertsContext } from '../../contexts/AlertContext';
 
 interface UseStreakRecoverProps {
   onAfterClose?: () => void;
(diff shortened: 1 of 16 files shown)


haiderRizvii (approved):

packages/shared/src/hooks/streaks/useStreakRecover.ts
```diff
@@ -58,13 +60,18 @@ export const useStreakRecover = ({
...
         .then((res) => res.recoverStreak),
     onSuccess: () => {
-      displayToast('Lucky you! Your streak has been restored');
       logEvent({
         event_name: LogEvent.StreakRecover,
       });
     },
   });
 
+  const hideRemoteAlert = useCallback(async () => {
```
sshanzel, line 69: I think it doesn't have to be at the close function. At the mount of the popup, we should send a mutation. The pop-up will stay open anyway even after that. Thoughts? cc @dailydotdev/web-team

ilasw: Yes, I put on close because maybe using extension you open a new tab but don't even read what's inside. 
Maybe if you close you had the chance!

sshanzel: Yeah, makes sense.

sshanzel (approved):
🚀

PR #3514, opened 2024-09-05
chore: add error state screens

rebelchris:
## Changes

Add two new states that hopefully should never trigger but can be:
- Boot doesn't load (either 500+ error) or gcp basically is fully down (catch) -> immediately render error, no need to go through the rest of rendering
- GB set user to error (manual) this one I also stop rendering as soon as possible so we avoid queries we don't need to make.

Fully open for ideas/suggestions though.

## Events

## Experiment

## Manual Testing

> [!CAUTION]

### Preview domain
https://chore-error-screens.preview.app.daily.dev

### DIFF
--- a/packages/shared/src/contexts/BootProvider.tsx
+++ b/packages/shared/src/contexts/BootProvider.tsx
@@ -7,6 +7,7 @@ import React, {
   useState,
 } from 'react';
 import { useQuery, useQueryClient } from '@tanstack/react-query';
+import dynamic from 'next/dynamic';
 import { BootApp, BootCacheData, getBootData } from '../lib/boot';
 import { AuthContextProvider } from './AuthContext';
 import { AnonymousUser, ContentLanguage, LoggedUser } from '../lib/user';
@@ -27,6 +28,13 @@ import { useHostStatus } from '../hooks/useHostPermissionStatus';
 import { checkIsExtension } from '../lib/func';
 import { Feed, FeedList } from '../graphql/feed';
 import { gqlClient } from '../graphql/common';
+
+const ServerError = dynamic(
+  () =>
+    import(
+      /* webpackChunkName: "serverError" */ '../components/errors/ServerError'
+    ),
+);
 
 function filteredProps<T extends Record<string, unknown>>(
   obj: T,
@@ -94,6 +102,7 @@ export const BootDataProvider = ({
   getRedirectUri,
   getPage,
 }: BootDataProviderProps): ReactElement => {
+  const [hasGlobalError, setGlobalError] = useState<boolean>(false);
   const { hostGranted } = useHostStatus();
   const isExtension = checkIsExtension();
 
@@ -133,13 +142,19 @@ export const BootDataProvider = ({
   } = useQuery(
     BOOT_QUERY_KEY,
     async () => {
-      const result = await getBootData(app);
+      try {
+        const result = await getBootData(app);
 
-      preloadFeedsRef.current({ feeds: result.feeds, user: result.user });
+        preloadFeedsRef.current({ feeds: result.feeds, user: result.user });
 
-      return result;
+        return result;
+      } catch (e) {
+        setGlobalError(true);
+        return null;
+      }
     },
     {
+      retry: hasGlobalError ? false : 3,
       refetchOnWindowFocus: loggedUser,
       staleTime: STALE_TIME,
       enabled: isExtension ? !!hostGranted : true,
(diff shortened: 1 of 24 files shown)


packages/shared/src/contexts/BootProvider.tsx
```diff
@@ -133,13 +142,19 @@ export const BootDataProvider = ({
...
+      try {
+        const result = await getBootData(app);
 
-      preloadFeedsRef.current({ feeds: result.feeds, user: result.user });
+        preloadFeedsRef.current({ feeds: result.feeds, user: result.user });
 
-      return result;
+        return result;
+      } catch (e) {
+        setGlobalError(true);
```
idoshamun, line 152: This will not retry right? I think we should retry a bit before setting global error

packages/shared/src/lib/boot.ts
```diff
@@ -1,13 +1,13 @@
...
 import { AnonymousUser, ContentLanguage, LoggedUser } from './user';
-import { apiUrl } from './config';
 import { Alerts } from '../graphql/alerts';
 import { RemoteSettings } from '../graphql/settings';
 import { Post } from '../graphql/posts';
 import { Squad } from '../graphql/sources';
-import { decrypt } from '../components/crypto';
 import { MarketingCta } from '../components/marketingCta/common';
 import { Feed } from '../graphql/feed';
+import { apiUrl } from './config';
```
omBratteng, line 9: Any reason these changed positions in import?

packages/shared/src/lib/boot.ts
```diff
@@ -87,21 +87,32 @@ export async function getBootData(app: string, url?: string): Promise<Boot> {
...
+    });
 
-  const features = await decrypt(
-    result.exp.f,
-    process.env.NEXT_PUBLIC_EXPERIMENTATION_KEY,
-    'AES-CBC',
-    128,
-  );
+    if (res.status >= 500) {
+      throw new Error('Server error');
```
idoshamun, line 98: You throw an error inside a try catch and then catch it to throw again. Seems a bit weird

omBratteng (approved):
Lgtm

packages/shared/src/contexts/BootProvider.tsx
rebelchris: It won't retry based on the globalError (see retry in props)
I mean we can, but in most of these it's pretty much fire anyway right? (so user refresh would be ok)?

packages/shared/src/lib/boot.ts
rebelchris: The main issue is that new error handling state of react query 😢 
Maybe @capJavert knows how to better avoid it

packages/shared/src/contexts/BootProvider.tsx
idoshamun: yes but once in a request may fail. retries are essential to keep the experience relatively smooth

rebelchris: Alright I can remove that eval prop for this case and only leave it to 0 for actual circuit breaker fails :)

rebelchris: Actually moved it around so there's only 1 try..catch and it retries 3 times before throwing the global error now.

packages/shared/src/contexts/BootProvider.tsx
```diff
@@ -146,6 +155,12 @@ export const BootDataProvider = ({
     },
   );
 
+  useEffect(() => {
+    if (error && !hasGlobalError) {
+      setGlobalError(true);
+    }
+  }, [error, hasGlobalError]);
+
```
capJavert, line 163: Could this not be just evaluation, if boot has error return `<ServerError />`?

rebelchris: Yeah tested in this new flow we don't need state/effect :)

packages/shared/src/lib/boot.ts
capJavert: This part is looks ok since fetch won't throw on non 200 status

packages/shared/src/lib/boot.ts
```diff
@@ -87,21 +87,32 @@ export async function getBootData(app: string, url?: string): Promise<Boot> {
...
+      'AES-CBC',
+      128,
+    );
+
+    result.exp.features = JSON.parse(features);
+
+    return result;
+  } catch (e) {
+    throw new Error('Server error');
+  }
```
capJavert, line 115: This I think we don't need? It will just throw by itself?

rebelchris: Yeah not sure, think I accidently repushed this in 1 commit 🤦

capJavert (approved):

PR #3516, opened 2024-09-05
refactor: convert card overlay from button to anchor

omBratteng:
## Changes

Refactors card overlay to use `a` instead of `button`

### Preview domain
https://as-543-card-button-to-anchor.preview.app.daily.dev


packages/shared/src/components/cards/common/CardOverlay.tsx
```diff
@@ -18,7 +20,14 @@ const CardOverlay = ({
     return null;
   }
 
-  return <CardButton title={post.title} onClick={onPostCardClick} />;
+  return (
```
rebelchris, line 23: Oh interesting, I kind of thought it was a a already, looking at list card for instance 😅
Maybe we can unify the two into 1 component now?

packages/shared/src/components/cards/ArticlePostCard.tsx
```diff
@@ -72,7 +74,11 @@ export const ArticlePostCard = forwardRef(function PostCard(
       flagProps={{ pinnedAt, trending }}
       bookmarked={post.bookmarked && !showFeedback}
     >
-      <CardOverlay post={post} onPostCardClick={onPostCardClick} />
+      <CardOverlay
```
rebelchris, line 77: We actually have `combinedClicks` which can do the same, as it's both the same function right?

rebelchris (commented):
Wouldn't we just be ok by passing 1 function and simply using `combinedClicks`?

packages/shared/src/components/cards/ArticlePostCard.tsx
omBratteng:
```ts
export const combinedClicks = (
  func: React.MouseEventHandler,
): CombinedClicks => {
  return {
    onAuxClick: (event) => event.button === 1 && func(event),
    onClick: func,
  };
};
```

I tried to use that, but it open modal if you aux clicked. You just want it to open new tab

rebelchris: Hmm this just feels like we could easily miss one of the two which I'd like to avoid if that makes sense? (Eg if we need to add tracking to clicks or so)

rebelchris (commented):
Just one concern, but maybe it's ok for now?

rebelchris (approved):

packages/shared/src/components/cards/ArticlePostCard.tsx
omBratteng: Discussed this in Gather. It re-uses the same top function, and because of prop-drilling and re-creating of the functions, it's the easiest way to do it now.

PR #3517, opened 2024-09-05
refactor: convert buttons in sidebar to anchors

omBratteng:
## Changes
- convert "New squad" squad page button to anchor
- convert "New squad" sidebar button to anchor
- always show anchor element in sidebar

### Preview domain
https://as-543-more-anchors.preview.app.daily.dev

### DIFF
--- a/packages/shared/src/components/sidebar/ClickableNavItem.tsx
+++ b/packages/shared/src/components/sidebar/ClickableNavItem.tsx
@@ -1,7 +1,13 @@
-import React, { HTMLAttributes, ReactElement, ReactNode } from 'react';
+import React, {
+  HTMLAttributes,
+  ReactElement,
+  ReactNode,
+  useCallback,
+} from 'react';
 import classNames from 'classnames';
 import Link from '../utilities/Link';
 import { navBtnClass, SidebarMenuItem } from './common';
+import { combinedClicks } from '../../lib/click';
 
 interface ClickableNavItemProps
   extends HTMLAttributes<HTMLButtonElement | HTMLAnchorElement> {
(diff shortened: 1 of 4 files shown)


packages/shared/src/components/sidebar/ActivitySection.tsx
```diff
@@ -31,6 +31,7 @@ export function ActivitySection({
         ),
         title: 'Bookmarks',
         path: `${process.env.NEXT_PUBLIC_WEBAPP_URL}bookmarks`,
```
rebelchris, line 33: I think you can actually just set this to be `href` and you don't need `forcedLink`
The forcedLink is more meant for `extension || webapp` where we have pages, but for webapp you want to force a link or something like that.

packages/shared/src/components/sidebar/SquadSection.tsx
```diff
@@ -40,6 +40,8 @@ export function SquadSection(props: SectionCommonProps): ReactElement {
     icon: () => <NewSquadIcon />,
     title: 'New Squad',
     action: () => openNewSquad({ origin: Origin.Sidebar }),
+    target: newSquadUrl,
```
rebelchris, line 43: Same here basically.

rebelchris (approved):
Non blocking, but we can clean it up a little bit.

packages/shared/src/components/sidebar/ActivitySection.tsx
omBratteng: Noticed in preview that it doesn't work, because if you're not logged in, it always returns a button. But I think we probably can remove that since we want anchor for SEO?

packages/shared/src/components/sidebar/ClickableNavItem.tsx
```diff
@@ -18,25 +24,24 @@ export function ClickableNavItem({
...
+    [showLogin, item],
+  );
 
   if (!isButton && (!item.action || item.path)) {
     return (
       <Link href={item.path} passHref prefetch={false}>
         <a
-          {...(item.action && { onClick: item.action })}
           {...props}
+          {...combinedClicks(onClick)}
```
rebelchris, line 44: I'm not sure this needs combinedClicks though, you stay in the app anyway so maybe same page is fine for this?

rebelchris (approved):
Non blocking comment

sshanzel (approved):
:shipit:

packages/shared/src/components/sidebar/ClickableNavItem.tsx
omBratteng: combinedClicks doesn't open anything in new tab for you, it just sends the input function to both onClick and onAuxClick.
It would be the same as doing
```ts
onClick={onClick}
onAuxClick={(event) => event.button === 1 && onClick}
```

sshanzel:
I think we forgot the Notification bell button 😂

PR #3522, opened 2024-09-06
test: try to remove flaky tests

ilasw:
## Changes
- Basically moved from check in UI (alerts) to `displayToast` function call with the right args
- Other small changes to tests

## Events

## Experiment

## Manual Testing

> [!CAUTION]

### Preview domain
https://fix-flacky-test-recover.preview.app.daily.dev

### DIFF
--- a/packages/shared/src/components/modals/streaks/StreakRecoverModal.spec.tsx
+++ b/packages/shared/src/components/modals/streaks/StreakRecoverModal.spec.tsx
@@ -14,6 +14,7 @@ import {
 } from '../../../graphql/users';
 import { mockGraphQL } from '../../../../__tests__/helpers/graphql';
 import * as actionHook from '../../../hooks/useActions';
+import * as toastHook from '../../../hooks/useToastNotification';
 import { ActionType } from '../../../graphql/actions';
 import Toast from '../../notifications/Toast';
 
@@ -36,26 +37,9 @@ const alertsWithStreakRecovery: Alerts = {
 const checkHasCompleted = jest.fn();
 const completeAction = jest.fn();
 const updateAlerts = jest.fn();
+const displayToast = jest.fn();
 
 const queryClient = new QueryClient({});
-
-const renderComponent = (props: TestProps) => {
-  const {
-    user = { ...loggedUser, reputation: 10 },
-    alerts = alertsWithStreakRecovery,
-  } = props;
-
-  return render(
-    <TestBootProvider
-      client={queryClient}
-      auth={{ user }}
-      alerts={{ alerts, updateAlerts }}
-    >
-      <BootPopups />
-      <Toast autoDismissNotifications={false} />
-    </TestBootProvider>,
-  );
-};
 
 const mockRecoveryQuery = (
   data: UserStreakRecoverData,
@@ -106,18 +90,44 @@ const mockAlertsMutation = () => {
   });
 };
 
+const renderComponent = (props: TestProps) => {
+  const {
+    user = { ...loggedUser, reputation: 10 },
+    alerts = alertsWithStreakRecovery,
+  } = props;
+
+  return render(
+    <TestBootProvider
+      client={queryClient}
+      auth={{ user }}
+      alerts={{ alerts, updateAlerts }}
+    >
+      <BootPopups />
+      <Toast autoDismissNotifications={false} />
+    </TestBootProvider>,
+  );
+};
+
 beforeEach(async () => {
-  // need to reset the mock on useActions
-  checkHasCompleted.mockReset();
   jest.spyOn(actionHook, 'useActions').mockReturnValue({
     completeAction,
     checkHasCompleted,
     isActionsFetched: true,
     actions: [],
   });
+  jest.spyOn(toastHook, 'useToastNotification').mockReturnValue({
+    displayToast,
+    dismissToast: jest.fn(),
+    subject: undefined,
+  });
 
   // need to reset the query cache
   queryClient.clear();
+  // need to reset the mock
+  checkHasCompleted.mockReset();
+  completeAction.mockReset();
+  updateAlerts.mockReset();
+  displayToast.mockReset();
 });
 
 it('should not render if user is not logged in', async () => {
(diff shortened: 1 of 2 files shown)


packages/shared/src/components/modals/streaks/StreakRecoverModal.spec.tsx
```diff
@@ -287,13 +322,17 @@ it('Should show success message on recover', async () => {
 
   // expect mutation to be called
   expect(mutationCalled).toBeTruthy();
+  await waitForNock();
```
capJavert, line 325: I think wait for nock might better be above since nock guarantees the above callback that sets `mutationCalled` has been executed?

capJavert: or you can use `waitFor(() => expect(mutationCalled).toBeTruthy())`

capJavert (approved):
Looks good, maybe test if flaky now with repeated execution @rebelchris you shared how to do it before?

packages/shared/src/components/modals/streaks/StreakRecoverModal.spec.tsx
ilasw: I've tried right now changing this but time to time test are failing again :(

capJavert: I think different one failed right now on ci? https://app.circleci.com/pipelines/github/dailydotdev/apps/19713/workflows/3c038725-d05f-43b1-82d7-adc25c032203/jobs/51077

ilasw: Yes, it is not from the recovery but I can  also quickly check for it 👀

capJavert: Let's keep it scoped, I was just referring to you saying it failed. So the test you are fixing is failing locally then?

ilasw: Yes, if I remove `await waitForNock();` and put `waitFor(() => expect(mutationCalled).toBeTruthy())` is failing locally once in a while

capJavert: not remove it but reverse order:
```
await waitForNock();
expect(mutationCalled).toBeTruthy();
```

ilasw: Nope, not working :\ 
But anyway before this line there is another `waitForNock` so we have one before and one after!

capJavert: And its working when nock is after?

capJavert: I see its two waitForNock now, do you know why we need second one?

ilasw: I think that is because of multiple mutation have been called before. I've tried to change it like this:

```ts
let mutationCalled = false;
  mockRecoveryMutation({}, () => {
    mutationCalled = true;
  });
  await waitForNock();

  mockAlertsMutation();
  await waitForNock();

  // expect mutation to be called
  await waitFor(() => expect(mutationCalled).toBeTruthy());

```

and seems working, I think is easier to understand and maintain, wdyt?

ilasw: Nevermind, if I remove the wait for nock after expect(mutationCalled) is failing time to time, had to restore it. :\

capJavert: From my perspective you would need only one waitForNock at the end, but flakiness is hard. 

`mockRecoveryMutation`  and `mockAlertsMutation` are just handlers that get setup and wait for requests so they should go first before any actions in the tests. 

I don't see the reason why we would call them incrementally one by one and then wait for them, thats why I asked if you know why its multiple waitForNock calls?

ilasw: I've tried to remove some, but component do some api call one after another and I think that some more waiting is needed for this reason?

capJavert: so the flow of the test how I see it:

1. mock recovery query
2. renderComponent -> this consumes above mock
3. waitForNock (to make sure we wait for above consumes to go through)
4. mock mutation and alerts mutation 
5. fireEvent.click (click on recover) -> this consumes above mock
6. waitForNock (to make sure we wait for above consumes to go through) (this fails 🔴 )
7. evaluate mutation called
8. evaluate toast

In your solution you have 4. and 5. reversed. But it still fails with both so I think the root cause is something else. From what I can se You can run test multiple times by replace `it()` with following:
```js
test.each(Array(100).fill(undefined))(
  'Should show success message on recover',
  async () => {
    // test code
  });
```

it takes a bit of time but it will show if test is flaky.

nensidosari (approved):

packages/shared/src/components/modals/streaks/StreakRecoverModal.spec.tsx
rebelchris: This was my approach, but above works better most cases as concurrently will break almost all our test anyway.

https://dailydotdev.slack.com/archives/C02E2C3C13R/p1723804495112589

packages/shared/src/components/modals/streaks/StreakRecoverModal.spec.tsx
```diff
@@ -106,18 +90,44 @@ const mockAlertsMutation = () => {
...
+  jest.spyOn(toastHook, 'useToastNotification').mockReturnValue({
+    displayToast,
+    dismissToast: jest.fn(),
+    subject: undefined,
+  });
 
   // need to reset the query cache
   queryClient.clear();
+  // need to reset the mock
+  checkHasCompleted.mockReset();
```
rebelchris, line 127: I think mock reset all works as well?

packages/shared/src/components/modals/streaks/StreakRecoverModal.spec.tsx
```diff
@@ -273,27 +308,30 @@ it('Should show success message on recover', async () => {
...
-  expect(alert).toContainHTML('Lucky you! Your streak has been restored');
+  await waitForNock();
+  expect(displayToast).toHaveBeenCalledWith(
+    'Lucky you! Your streak has been restored',
+  );
 });
 
-it('Should show error message on recover fail', async () => {
+it('Should dismiss popup on close if checked option', async () => {
+  // window.scrollTo = jest.fn();
```
rebelchris, line 333: Either we should bring it back or simply delete it :)

rebelchris (approved):
Not really anything smart to say, toast === flaky :( 
I've found better luck adjusting with fakeTimers and relying on 1 tick

sshanzel (approved):
Thank you for your effort on this, working with flaky tests are not easy, especially with our toast notification 🙏

PR #3523, opened 2024-09-06
feat: new squad directory page

sshanzel:
## Changes
- Squad directory page. Functionality-wise, it works now, just waiting for the background image to be uploaded in cloudinary.
- Lots of cleanup for unused functionalities relating to the variety of the Squad feed card.
- After running through the conditions on when to display some components over the other, found out that we won't be needing most conditions now that we removed the card that is not Squad but was using the "Squad" card.
  - SquadJoinButton wrapper is not needed anymore as we now only need the JoinButton.
  - SquadImage is the same case.

## Events

## Experiment

## Manual Testing

> [!CAUTION]

MI-495

### Preview domain
https://mi-509.preview.app.daily.dev


packages/webapp/pages/squads/index.tsx
```diff
@@ -146,61 +140,9 @@ const SquadsPage = ({ initialData }: Props): ReactElement => {
                 className="mt-5 px-6"
                 inlineHeader
               >
-                {queryResult?.data?.pages?.length > 0 && (
```
sshanzel, line 149: image

packages/shared/src/components/cards/squad/SquadGrid.tsx
```diff
@@ -53,79 +43,62 @@ const borderColorToClassName: Record<SourceCardBorderColor, string> = {
...
   source,
-  title,
-  subtitle,
-  icon,
-  action,
-  description,
-  borderColor = SourceCardBorderColor.Avocado,
-  banner,
-}: SourceCardProps): ReactElement => {
+}: UnFeaturedSquadCardProps): ReactElement => {
```
sshanzel, line 46: Simplified the component to just base the content from a Squad since we don't need the extra props anymore.

rebelchris (approved):
Looks good to me :)

PR #3541, opened 2024-09-16
refactor: search by id and title

sshanzel:
## Changes
- Use the title on the URL rather than the uuid.

## Events

## Experiment

## Manual Testing

> [!CAUTION]

### Preview domain
https://mi-533.preview.app.daily.dev


ilasw (changes requested):
On preview https://mi-533.preview.app.daily.dev/squads/discover if you try to click on category tabs nothing happens 😮

sshanzel:
> On preview https://mi-533.preview.app.daily.dev/squads/discover if you try to click on category tabs nothing happens 😮

Yeah, the API is not yet merged https://github.com/dailydotdev/daily-api/pull/2258

ilasw (approved):
Nevermind if the API is getting merged for me is ok :)

nensidosari (approved):

packages/shared/src/components/squads/layout/useSquadDirectoryLayout.ts
```diff
@@ -35,7 +35,7 @@ export const useSquadDirectoryLayout = (): SquadDirectoryLayoutReturn => {
     return flatCategories.reduce(
       (result, { node }) => ({
         ...result,
-        [node.title]: `/squads/discover/${node.id}`,
+        [node.title]: `/squads/discover/${node.title.toLowerCase()}`,
```
capJavert, line 38: I see now we do this, do we maybe want to use generated column for "slug" on API (like we do for post) so we then always have reliable slug over which we can properly search? Or even just add slug to the API?

capJavert: I think it simplifies the whole searching since actually you don't search, you just want to match by slug which is better for SEO?

rebelchris: I agree with Ante, it's a generic approach as we did for posts, we should uniform it like we did with slug for posts (calculated column on API side)
You can then use findBy({id || slug}) it should make it more uniform/easier.

rebelchris (changes requested):
Agreed with Ante's comment

packages/shared/src/components/squads/layout/useSquadDirectoryLayout.ts
sshanzel: Sure, why not.

nensidosari (approved):

packages/shared/src/graphql/fragments.ts
```diff
@@ -236,3 +236,12 @@ export const USER_STREAK_FRAGMENT = gql`
     weekStart
   }
 `;
+
+export const SOURCE_CATEGORY_FRAGMENT = gql`
+  fragment SourceCategoryFragment on SourceCategory {
+    id
+    slug
+    title
+    createdAt
```
rebelchris, line 245: do you need createdAt?
Just be careful with testing this as it might break due to that conversion issue we have.

rebelchris (approved):
Non blocking, but do check it please.

packages/shared/src/graphql/fragments.ts
sshanzel: Not at the moment. But yeah probably be safe than sorry. Will remove it quickly.

capJavert (approved):

PR #3543, opened 2024-09-16
Mi 535

sshanzel:
## Changes
- Memoize the `useHorizontalScroll` variables.
- Refactored it by taking out a section that can be reused somewhere else.

## Events

## Experiment

## Manual Testing

> [!CAUTION]

MI-535

### DIFF
--- a/packages/shared/src/components/HorizontalScroll/useCalculateVisibleElements.ts
+++ b/packages/shared/src/components/HorizontalScroll/useCalculateVisibleElements.ts
@@ -1,0 +1,72 @@
+import { MutableRefObject, useCallback, useEffect, useState } from 'react';
+import { useToggle } from '../../hooks/useToggle';
+import { useEventListener } from '../../hooks';
+
+interface UseCalculateVisibleElementsProps<El extends HTMLElement> {
+  ref: MutableRefObject<El | null>;
+}
+
+interface UseCalculateVisibleElements {
+  calculateVisibleCards: () => void;
+  isOverflowing: boolean;
+  elementsCount: number;
+  scrollableElementWidth: number;
+}
+
+export const useCalculateVisibleElements = <El extends HTMLElement>({
+  ref,
+}: UseCalculateVisibleElementsProps<El>): UseCalculateVisibleElements => {
+  const [isOverflowing, setIsOverflowing] = useToggle(false);
+  const [elementsCount, setElementsCount] = useState<number>(0);
+  const [scrollableElementWidth, setScrollableElementWidth] =
+    useState<number>(0);
+
+  const calculateVisibleCards = useCallback(() => {
+    const currentRef = ref.current;
+    if (
+      !currentRef ||
+      !currentRef.firstElementChild ||
+      !(currentRef.firstElementChild instanceof HTMLElement)
+    ) {
+      return;
+    }
+
+    setIsOverflowing(currentRef.scrollWidth > currentRef.clientWidth);
+
+    if (currentRef && currentRef.firstElementChild) {
+      const element = currentRef.firstElementChild;
+      const elementWidth = element.offsetWidth;
+      const elementMarginRight = parseFloat(
+        getComputedStyle(element).marginRight || '0',
+      );
+      const elementMarginLeft = parseFloat(
+        getComputedStyle(element).marginLeft || '0',
+      );
+      const containerGap = parseFloat(getComputedStyle(currentRef).gap || '0');
+      const elementWidthWithMarginsAndGap =
+        elementWidth + elementMarginRight + elementMarginLeft + containerGap;
+
+      setScrollableElementWidth(elementWidthWithMarginsAndGap);
+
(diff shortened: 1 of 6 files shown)


packages/storybook/stories/components/HorizontalScroll.stories.tsx
```diff
@@ -88,33 +88,37 @@ export const HorizontalScrollSquadStory: Story = {
...
-            /* eslint-disable react/no-array-index-key */
-            <UnfeaturedSquadGrid
-              {...args}
-              key={`horizontal scroll item ${i}`}
-            />
-          ))}
-        </HorizontalScroll>
-      </>
+      <QueryClientProvider>
+        <AuthContext.Provider value={{}}>
```
nensidosari, line 97: Are these necessary, just asking since they weren't before.

nensidosari (commented):
Looks good!

One more thing, not sure if in the main PR we've made any changes to the header of the horizontal scroll, but it has broken the header of the horizontal scroll on tags/sources page 
Screenshot 2024-09-17 at 7 53 54 AM

vs in prod:
Screenshot 2024-09-17 at 7 56 23 AM


Let's please fix this last thing as well 🙏

packages/shared/src/components/HorizontalScroll/useCalculateVisibleElements.ts
```diff
@@ -0,0 +1,72 @@
...
+        elementWidth + elementMarginRight + elementMarginLeft + containerGap;
+
+      setScrollableElementWidth(elementWidthWithMarginsAndGap);
+
+      const mainContainerWidth = currentRef.offsetWidth;
+      setElementsCount(
+        Math.floor(mainContainerWidth / elementWidthWithMarginsAndGap),
+      );
+    }
+  }, [ref, setIsOverflowing]);
```
rebelchris, line 56: The ref won't trigger anymore on client side right, making the useEffect actually not re-trigger as well? (didn't probe any issues locally?)

packages/storybook/stories/components/HorizontalScroll.stories.tsx
```diff
@@ -12,15 +12,23 @@ import {
...
+
+const ScrollableElement = ({
+  item,
+  index,
+  ...props
+}: {
+  item: { value: string };
+  index: number;
+}): ReactElement => (
+  <div {...props} className="w-full max-w-40 border p-4">
```
rebelchris, line 27: border won't do anything without color?

rebelchris (commented):
Mostly question around the hook as it's a server side calculation thus wondering if it's reliable to use as the only dependency.

packages/storybook/stories/components/HorizontalScroll.stories.tsx
sshanzel: The storybook is not rendering without it.

packages/shared/src/components/HorizontalScroll/useCalculateVisibleElements.ts
sshanzel: Yeah, this won't trigger since it is a ref. Just the linter asking it to be part of the deps since it came from the props.

packages/storybook/stories/components/HorizontalScroll.stories.tsx
sshanzel: solid black is what being used with just border. This is just for display anyway, hence made it look like it is not really part of the design system.

packages/shared/src/components/HorizontalScroll/useCalculateVisibleElements.ts
rebelchris: Ah yes, but my concern is wider in that case.
Since that won't ever trigger does the calculation ever happen in the ref is not yet set for instance but you rely on it to give accurate adjustments?

packages/storybook/stories/components/HorizontalScroll.stories.tsx
rebelchris: Ah I see :) thanks for the explanation.

sshanzel:
> One more thing, not sure if in the main PR we've made any changes to the header of the horizontal scroll, but it has broken the header of the horizontal scroll on tags/sources page

May have been left unchecked introducing the refactor as we did not change any styling on the header itself. Will push a fix.

packages/shared/src/components/HorizontalScroll/useCalculateVisibleElements.ts
rebelchris: Yeah just saying if it's not available on server render yet, it will never retrigger it.
But if you tested it and had no issues it's probably fine.

sshanzel: Yeah, the calculation result will be useful once the user can click the left/right buttons. Server-side we should be good.

sshanzel:
@nensidosari should be fixed by now.

nensidosari (approved):

PR #3551, opened 2024-09-18
feat: sorting comments with fixing new comments

sshanzel:
## Changes
- Bring back the sorting feature.
- Properly update the possible query cache data. Kindly check this specific commit to make review easier: https://github.com/dailydotdev/apps/commit/da603b6f1ebf28669f3b0bd8e1830139fc26cb00

TODO:
- ~~Test on preview.~~

## Events

## Experiment

## Manual Testing

> [!CAUTION]

MI-538

### Preview domain
https://mi-538.preview.app.daily.dev


packages/shared/src/components/post/PostEngagements.tsx
```diff
@@ -23,6 +23,9 @@ import { useActions } from '../../hooks';
 import { ActionType } from '../../graphql/actions';
 import { AdAsComment } from '../comments/AdAsComment';
 import { PostContentReminder } from './common/PostContentReminder';
+import { Typography, TypographyType } from '../typography/Typography';
+import { Button, ButtonIconPosition, ButtonSize } from '../buttons/Button';
+import { TimeSortIcon } from '../icons/Sort/Time';
```
rebelchris, line 28: I think you forgot to add it to the icon index file.

packages/shared/src/components/modals/post/CommentModal.tsx
```diff
@@ -65,6 +48,37 @@ const getCommentFromCache = ({
...
+  postId,
+  commentId,
+}: GetCommentFromCacheProps): Comment | undefined => {
+  if (!commentId) {
+    return undefined;
+  }
+
+  const keys = getAllCommentsQuery(postId);
+  // eslint-disable-next-line no-restricted-syntax
+  for (const key of keys) {
```
rebelchris, line 68: Do you need a for loop here even? you only care about one right? 
Maybe we can just use `keys[0]`? 

Or is there some high likelyhood of it not existing?

packages/shared/src/components/post/PostComments.tsx
```diff
@@ -51,14 +54,14 @@ export function PostComments({
   const container = useRef<HTMLDivElement>();
   const { tokenRefreshed } = useContext(AuthContext);
   const { requestMethod } = useRequestProtocol();
-  const queryKey = generateQueryKey(RequestKey.PostComments, null, id);
+  const queryKey = generateCommentsQueryKey({ postId: id, sortBy });
```
rebelchris, line 57: Personally not sure if this abstraction adds much, I'd prefer just to use `generateQueryKey` here as it needs to be as close to the query to avoid issues later on. 
(that said, also not strongly against it)

packages/shared/src/hooks/usePostById.ts
```diff
@@ -66,8 +65,8 @@ export const removePostComments = (
   commentId: string,
   parentId: string,
 ): void => {
-  const key = generateQueryKey(RequestKey.PostComments, null, post.id);
-  client.setQueryData<PostCommentsData>(key, (data) => {
+  const keys = getAllCommentsQuery(post.id);
+  const removeCachnedComment = (data: PostCommentsData) => {
```
rebelchris, line 69:
```suggestion
  const removeCachedComment = (data: PostCommentsData) => {
```

packages/shared/src/hooks/usePostById.ts
```diff
@@ -95,6 +94,10 @@ export const removePostComments = (
     );
 
     return data;
+  };
+
+  keys.forEach((key) => {
+    client.setQueryData(key, removeCachnedComment);
```
rebelchris, line 100:
```suggestion
    client.setQueryData(key, removeCachedComment);
```

rebelchris (changes requested):
Minor feedback, haven't tried e2e :)

nensidosari (approved):
Tested on preview and seems to work fine.

packages/shared/src/components/post/PostEngagements.tsx
sshanzel: I added it here since it is a Sort icon too, but with different design but same name. And both exists from the guideline.

packages/shared/src/components/modals/post/CommentModal.tsx
sshanzel: Yes, we only need now, but we return early so we won't have to worry to iterating everything.

packages/shared/src/components/post/PostComments.tsx
sshanzel: Wanted to do it this way so nothing would be missed like I did. Forgot to realize that we have other areas that uses the same key.

sshanzel:
@rebelchris would love to hear what you think 👀

sshanzel:
@nensidosari @rebelchris after thinking about this more, I think the cache updates should be:
- Check if data exists.
- If yes, add the comment.
- If not, invalidate the query.

This is to avoid not fetching the actual data for the comments when you flip the sorting. For example:
- Made a comment.
- Insert the comment on each cache keys.
- But since one of the keys was not fetched initially (key1: with comments, key2: empty), we now set a value to that key.
- If the user flips the sort, the cache would think there should be valid data. When in fact it could be missing the other comments since the previous cache did not have those and we just inserted the comment.

packages/shared/src/hooks/post/useMutateComment.ts
```diff
@@ -126,6 +126,23 @@ export const useMutateComment = ({
...
+          forInvalidation.push(queryKey);
+          return null;
+        }
+
+        return updateQueryData(data);
+      });
+    });
+
+    forInvalidation.forEach(async (queryKey) => {
+      await client.invalidateQueries(queryKey);
```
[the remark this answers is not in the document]
rebelchris: This will just conflict with the read replica on this call though.
I validated it today as it was only using the optimistic update.
With invalidation we can't guarantee its available unless we remove read replica for this.

I still think dealing with it optimistic might be better?
Cc @capJavert as it impacts api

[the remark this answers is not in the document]
sshanzel: Good point. I think we can actually even omit the invalidation of queries. The case I am worried about is that:
- User opens a post with many comments.
- Made a comment.
- This pushes a comment to the cache on all keys.
- The key for the currently displayed data is correct since it has the previous comments data.
- The other key without pre-existing data, will now have some data as we will push it, but then it would be missing all of the other comments data since this is now considered as valid (not stale) data.

Let me update it one more time.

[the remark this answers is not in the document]
capJavert: I think core issue is that we are trying to invalidate every key we know could exist vs invalidating what is active on the screen. 

But to focus on problem on hand, in regards to read replica we should always focus to optimistic update current view/screen/component and everything else theoretically could be just invalidated because until user navigates to it replication should be done.

To summarize:
- optimistic update comment list user is currently on
- invalidate other lists (with top comment/post key)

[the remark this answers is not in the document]
sshanzel: Right. Then we should be good here if that is the case.

packages/shared/src/components/post/PostEngagements.tsx
rebelchris: I mean to export it from the barrel file :)

rebelchris (approved):

packages/webapp/tsconfig.json
```diff
@@ -17,13 +17,20 @@
...
     "isolatedModules": true,
     "jsx": "preserve",
-    "incremental": true
+    "incremental": true,
+    "plugins": [
+      {
+        "name": "next"
+      }
+    ],
+    "strictNullChecks": true
```
sshanzel, line 26: @rebelchris I did not change this explicitly, was there some reverted commits from the main branch about tsconfig?

sshanzel:
Tested everything in preview ✅

PR #3552, opened 2024-09-18
feat: toast for user that are no longer available to recover

ilasw:
> Previous merge was reverted due to users that reported seeing the toast even if no action was done.

## Changes
- Add toast for user that are no longer available to recover
- Disable recover popup for user that don't want it

## Events

## Experiment

## Manual Testing

> [!CAUTION]

### Preview domain
https://mi-465-alert-on-unpossible-recov.preview.app.daily.dev


packages/shared/src/hooks/streaks/useStreakRecover.ts
```diff
@@ -78,6 +74,22 @@ export const useStreakRecover = ({
     });
   }, [updateAlerts]);
 
+  const { data, isLoading } = useQuery<StreakQueryData>({
+    queryKey: generateQueryKey(RequestKey.UserStreakRecover),
+    queryFn: async () => {
+      const res = await gqlClient.request(USER_STREAK_RECOVER_QUERY);
+
+      if (!res?.streakRecover?.canRecover) {
```
rebelchris, line 82: Oh wait doesn't this also return false if in general you can't recover?
(so not if you didn't just loose it?)

I feel this should only trigger on a user action right?

ilasw: I can put a check on query string put from the notification. This exist only if user clicked. What do you think?

ilasw: Updated ✅

nensidosari (approved):

packages/shared/src/hooks/streaks/useStreakRecover.ts
```diff
@@ -52,10 +53,9 @@ export const useStreakRecover = ({
   const { updateAlerts } = useAlertsContext();
   const { user } = useAuthContext();
   const client = useQueryClient();
-  const { data, isLoading } = useQuery<StreakQueryData>({
-    queryKey: generateQueryKey(RequestKey.UserStreakRecover),
-    queryFn: async () => await gqlClient.request(USER_STREAK_RECOVER_QUERY),
-  });
+  const {
+    query: { streak_restore: streakRestore },
+  } = useRouter();
```
rebelchris, line 58: This seems a bit pointless right?
The modal never shows if that router element is not available already or if it didn't come from bootPopup?

packages/shared/src/hooks/streaks/useStreakRecover.ts
```diff
@@ -78,6 +78,22 @@ export const useStreakRecover = ({
     });
   }, [updateAlerts]);
 
+  const { data, isLoading } = useQuery<StreakQueryData>({
```
rebelchris, line 81: Maybe a more failsafe approach would be to convert this to a const, and pass it to this hook as `enabled`?
https://github.com/dailydotdev/apps/blob/b65cd5575b653380bfa9927d6e93851f4ffbb7fe/packages/shared/src/components/modals/streaks/StreakRecoverModal.tsx#L174-L182

You can then just early return the hook, or set the query to be disabled?

rebelchris (commented):
More generic feedback:
It's feels a bit like we failsafe in different ways and would have ideally seen 1 way that fixes it everywhere. (but non blocking/change needed) just take a look at the suggestions please.

packages/shared/src/hooks/streaks/useStreakRecover.ts
ilasw: If modal is opened by BootPopup we do not have this query string. 
Added just for checking if the user actively opened the modal, avoiding toast without context!

packages/shared/src/hooks/streaks/useStreakRecover.ts
ilasw: `recover` comes from query but adding `user` and `isStreaksEnabled` would be a great addition! Trying it :)

packages/shared/src/hooks/streaks/useStreakRecover.ts
```diff
@@ -83,7 +85,9 @@ export const useStreakRecover = ({
     queryFn: async () => {
       const res = await gqlClient.request(USER_STREAK_RECOVER_QUERY);
 
-      if (!res?.streakRecover?.canRecover && !!streakRestore) {
+      const userCantRecoverInNotificationCenter =
+        !res?.streakRecover?.canRecover && !!streakRestore;
```
rebelchris, line 89: Ah sorry I meant you could actually pass it to the hook as `enabled`?
then it can be whatever from top level already so you dont need secondary things inside here.

ilasw: res comes from API call so we need here. We can pass from outside `streakRestore` but the API call need to be fired even if it is _true_, we need the check just to open the toast only in the notification center @rebelchris

rebelchris: But you already have it outside where you iniate the hook? 
Eg listening for this specific router param or from boot where you have `canRecover` 

Not sure i'm being clear with what I'm seeing?

ilasw: Yes, but this handle the Viktor's edge case, where you have boot alert set to true (and modal opening in background) and you keep seeing the toast. In this case you want user to not see the toast!

ilasw: PS: I've put here because if I use displayToast inside a useEffect it makes the component rerender in a infinite loop :(

rebelchris (approved):
I guess it's just extra failsafe, so ok with this :) 
Haven't tested e2e, but assume you did.

nensidosari (approved):

PR #3554, opened 2024-09-18
feat: follow button

capJavert:
## Changes

- reuse base components from `SourceActions` components (also adjusted them to match new follow designs)
- add `useContentPreference` hook for related mutations AS-580
- add `FollowButton` component (~~currently sends mutation but page refresh is needed to view new state~~)
- add support for follow of author/scout AS-572
- adjust fragments/queries 

TODO:
- [x] handle mutations and optimistic query data update
- [x] e2e test a bit more (merging to unblock other tasks)

**!!! Opening this for review until I figure out invalidation of local cache. We can also probably merge this and use it to add the button around the app to not block other tasks.**

**!!! Update mutations handled for single post page only, we should be able to apply the concept elsewhere**

image

## Events

## Experiment

## Manual Testing

> [!CAUTION]

AS-571

### Preview domain
https://as-571-follow-button.preview.app.daily.dev

### DIFF
--- a/packages/shared/src/components/sources/SourceActions/SourceActionsFollow.tsx
+++ b/packages/shared/src/components/sources/SourceActions/SourceActionsFollow.tsx
@@ -1,4 +1,4 @@
-import React, { ReactElement, useState } from 'react';
+import React, { ReactElement } from 'react';
 import classNames from 'classnames';
 import { ButtonSize, ButtonVariant } from '../../buttons/common';
 import { Button } from '../../buttons/Button';
@@ -13,34 +13,28 @@ interface SourceActionsFollowProps {
 
 const SourceActionsFollow = (props: SourceActionsFollowProps): ReactElement => {
   const { className, isSubscribed, isFetching, onClick, variant } = props;
-  const [isHovered, setIsHovered] = useState(false);
-
-  let label = isSubscribed ? 'Following' : 'Follow';
-
-  if (isHovered && isSubscribed) {
-    label = 'Unfollow';
-  }
 
   return (
     <Button
-      aria-label={label}
+      aria-label={`Toggle follow status, currently you are ${
+        isSubscribed ? 'following' : 'not following'
+      }`}
       className={classNames(
         isSubscribed &&
-          'min-w-24 hover:bg-overlay-float-ketchup hover:text-accent-ketchup-default',
+          'group min-w-24 hover:bg-overlay-float-ketchup hover:text-accent-ketchup-default',
         className,
       )}
       disabled={isFetching}
       onClick={onClick}
       size={ButtonSize.Small}
       variant={isSubscribed ? ButtonVariant.Subtle : variant}
-      onMouseEnter={() => {
-        setIsHovered(true);
-      }}
-      onMouseOut={() => {
-        setIsHovered(false);
-      }}
     >
-      {label}
+      <span className="group-hover:hidden">
+        {isSubscribed ? 'Following' : 'Follow'}
+      </span>
+      <span className="hidden group-hover:block">
+        {isSubscribed ? 'Unfollow' : 'Follow'}
+      </span>
     </Button>
   );
 };
--- a/packages/shared/src/components/sources/SourceActions/SourceActionsFollow.tsx
+++ b/packages/shared/src/components/sources/SourceActions/SourceActionsFollow.tsx
@@ -1,4 +1,5 @@
-import React, { ReactElement } from 'react';
+import React, { ReactElement, useState } from 'react';
+import classNames from 'classnames';
 import { ButtonSize, ButtonVariant } from '../../buttons/common';
 import { Button } from '../../buttons/Button';
 
@@ -12,15 +13,32 @@ interface SourceActionsFollowProps {
 
 const SourceActionsFollow = (props: SourceActionsFollowProps): ReactElement => {
   const { className, isSubscribed, isFetching, onClick, variant } = props;
-  const label = isSubscribed ? 'Unfollow' : 'Follow';
+  const [isHovered, setIsHovered] = useState(false);
+
+  let label = isSubscribed ? 'Following' : 'Follow';
+
+  if (isHovered && isSubscribed) {
+    label = 'Unfollow';
+  }
+
   return (
     <Button
       aria-label={label}
-      className={className}
+      className={classNames(
+        isSubscribed &&
+          'min-w-24 hover:bg-overlay-float-ketchup hover:text-accent-ketchup-default',
+        className,
+      )}
       disabled={isFetching}
       onClick={onClick}
       size={ButtonSize.Small}
-      variant={isSubscribed ? ButtonVariant.Tertiary : variant}
+      variant={isSubscribed ? ButtonVariant.Subtle : variant}
+      onMouseEnter={() => {
+        setIsHovered(true);
+      }}
+      onMouseOut={() => {
+        setIsHovered(false);
+      }}
     >
       {label}
     </Button>
--- a/packages/shared/src/components/widgets/PostUsersHighlights.tsx
+++ b/packages/shared/src/components/widgets/PostUsersHighlights.tsx
@@ -226,7 +226,7 @@ export const UserHighlight = (props: UserHighlightProps): ReactElement => {
       )}
       {isUserType && (
         <FollowButton
-          userId="testuser"
+          userId={id}
           type={ContentPreferenceType.User}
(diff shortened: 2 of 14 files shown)


packages/shared/src/graphql/fragments.ts
```diff
@@ -32,6 +32,46 @@ export const USER_SHORT_INFO_FRAGMENT = gql`
...
 
+export const CONTENT_PREFERENCE_FRAMENT = gql`
+  fragment ContentPreferenceFragment on ContentPreference {
+    referenceId
+    user {
+      id
+      name
+      image
+      username
+    }
```
capJavert, line 43: I lowered amount of data we need here, I would even say that we might not even need anything, since `ContentPreference` is always on parent entity which will have its own id. Will log it so we can remove if not needed at the end of implementation at least.

omBratteng:
It's getting cramped there 🙈 
image

Who am I following? Can't even see first letter in the name 😬 
image

packages/shared/src/hooks/usePostById.ts
```diff
@@ -115,6 +126,52 @@ const usePostById = ({ id, options = {} }: UsePostByIdProps): UsePostById => {
   );
   const post = postById || (options?.initialData as PostData);
 
+  useMutationSubscription({
+    matcher: contentPreferenceMutationMatcher,
+    callback: ({ mutation, variables: mutationVariables, queryClient }) => {
```
capJavert, line 131: This is handling optimistic updates of contentPreference state on post author/scout. Ideally we can use `useMutationSubscription` in other places like feed and user listings as well to update cache in those queries.

packages/shared/src/components/sources/SourceActions/SourceActionsFollow.tsx
```diff
@@ -12,15 +13,32 @@ interface SourceActionsFollowProps {
...
-  const label = isSubscribed ? 'Unfollow' : 'Follow';
+  const [isHovered, setIsHovered] = useState(false);
+
+  let label = isSubscribed ? 'Following' : 'Follow';
+
+  if (isHovered && isSubscribed) {
+    label = 'Unfollow';
+  }
+
   return (
```
rebelchris, line 24: This one couldn't be the generic FollowButton now?
(not sure why it would be different?)

packages/shared/src/components/widgets/PostUsersHighlights.tsx
```diff
@@ -219,6 +224,14 @@ export const UserHighlight = (props: UserHighlightProps): ReactElement => {
           source={user}
         />
       )}
+      {isUserType && (
+        <FollowButton
+          userId={id}
+          type={ContentPreferenceType.User}
+          status={(user as CommentAuthor).contentPreference?.status}
+          entityName={`@${handleOrUsernameOrId}`}
```
rebelchris, line 232: Why entityName? I assumed this would be the `entity` name like `User` or `Source`?

packages/shared/src/hooks/contentPreference/useContentPreference.ts
```diff
@@ -0,0 +1,112 @@
...
+      entity,
+      entityName,
+    }: PropsParameters<UseContentPreference['follow']>) => {
+      await gqlClient.request(CONTENT_PREFERENCE_FOLLOW_MUTATION, {
+        id,
+        entity,
+        status: ContentPreferenceStatus.Follow,
+      });
+
+      displayToast(`✅ You are now following ${entityName}`);
```
rebelchris, line 37: Oh I see, it's more of a display thing :)

packages/shared/src/hooks/contentPreference/useContentPreference.ts
```diff
@@ -0,0 +1,112 @@
...
+      entity,
+      entityName,
+    }: PropsParameters<UseContentPreference['subscribe']>) => {
+      await gqlClient.request(CONTENT_PREFERENCE_FOLLOW_MUTATION, {
+        id,
+        entity,
+        status: ContentPreferenceStatus.Follow,
+      });
+
+      displayToast(`✅ You are no longer subscribed to ${entityName}`);
```
rebelchris, line 96: This needs the ✅?

packages/shared/src/hooks/usePostById.ts
rebelchris: Why do we expose preference in sub elements?
Can't we just rely on the main query and use that as a source of truth rather? so only 1 query needs to validate (you are following User[])?

rebelchris (commented):
Some minor questions.
Mainly wondering if we can't abstract the knowledge of whether you follow something to it's own logic vs creeping it into things?

packages/shared/src/components/sources/SourceActions/SourceActionsFollow.tsx
rebelchris: I guess because we only allow it for users for now :)

capJavert: yes, I used only base components and we will transition to `FollowButton` in phase 2 & 3

packages/shared/src/hooks/contentPreference/useContentPreference.ts
capJavert: yes, just wanted to slim it down and make generic when we introduce other follow entities.

packages/shared/src/hooks/contentPreference/useContentPreference.ts
capJavert: no 🙈

packages/shared/src/hooks/usePostById.ts
capJavert: It does not really work like that, most of the entities will get contentPreference from a field, which means it will live in different queries.

capJavert: For example in post modal `contentPreference` will come from feed query and each post item inside the feed. I will add that example next.

rebelchris: Hmm I see, ideally it was a sole query, but yeah get that it's not possible :)

packages/shared/src/components/sources/SourceActions/SourceActionsFollow.tsx
```diff
@@ -12,15 +13,32 @@ interface SourceActionsFollowProps {
...
 const SourceActionsFollow = (props: SourceActionsFollowProps): ReactElement => {
   const { className, isSubscribed, isFetching, onClick, variant } = props;
-  const label = isSubscribed ? 'Unfollow' : 'Follow';
+  const [isHovered, setIsHovered] = useState(false);
+
+  let label = isSubscribed ? 'Following' : 'Follow';
+
+  if (isHovered && isSubscribed) {
+    label = 'Unfollow';
+  }
```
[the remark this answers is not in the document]
capJavert: Yeah I tried handling it with CSS, I gave `after:content` a try but to no luck. 

Above is ok but kinda feels wrong to render two divs all the time?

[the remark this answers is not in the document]
capJavert: I like it since its more performant then JS for sure though.

packages/shared/src/hooks/contentPreference/useContentPreference.ts
capJavert: updated

packages/shared/src/components/sources/SourceActions/SourceActionsFollow.tsx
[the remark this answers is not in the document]
ilasw: Two div are present but only one is rendered if we use display:none, so this is a ok situation of accessibility (display none is not read nor rendered by any device) 👀 

I suggest maybe to use a neutral sentence for `aria-label`, something like this `Toggle follow status, currently you are {subscribed/x}`.

Most of the people that use aria attributes are using a screen reader that focus on element more then hover so this could be a good trade off. 

What do you think?

rebelchris (approved):
Nice

packages/shared/src/components/sources/SourceActions/SourceActionsFollow.tsx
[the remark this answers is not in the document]
capJavert: Let's do it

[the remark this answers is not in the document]
capJavert: updated!

ilasw (approved):
Looks good 💯

PR #3555, opened 2024-09-18
feat: following tab in feed preferences

omBratteng:
## Changes

- Add new tab in feed preferences
- Adds two new checkboxes for notifications

## Events

## Experiment

## Manual Testing

> [!CAUTION]

AS-579

### Preview domain
https://as-579-settings.preview.app.daily.dev

### DIFF
--- a/packages/shared/src/components/feeds/MobileFeedActions.tsx
+++ b/packages/shared/src/components/feeds/MobileFeedActions.tsx
@@ -42,6 +42,7 @@ export function MobileFeedActions(): ReactElement {
           onOpenFeedFilters={() =>
             openModal({
               type: LazyModal.FeedFilters,
+              persistOnRouteChange: true,
             })
           }
         />
--- a/packages/shared/src/components/feeds/MobileFeedActions.tsx
+++ b/packages/shared/src/components/feeds/MobileFeedActions.tsx
@@ -42,7 +42,6 @@ export function MobileFeedActions(): ReactElement {
           onOpenFeedFilters={() =>
             openModal({
               type: LazyModal.FeedFilters,
-              persistOnRouteChange: true,
             })
           }
         />
--- a/packages/shared/src/components/filters/FollowingFilter.tsx
+++ b/packages/shared/src/components/filters/FollowingFilter.tsx
@@ -4,7 +4,7 @@ import { OpenLinkIcon } from '../icons';
 import { useAuthContext } from '../../contexts/AuthContext';
 import Link from '../utilities/Link';
 
-export const FollowingFilter = (): JSX.Element => {
+export const FollowingFilter = (): ReactElement=> {
   const { user } = useAuthContext();
   return (
     <div className="flex flex-col items-center gap-4 px-6 py-10 text-center text-text-secondary">
--- a/packages/shared/src/components/filters/FollowingFilter.tsx
+++ b/packages/shared/src/components/filters/FollowingFilter.tsx
@@ -8,11 +8,10 @@ import { FlexCentered } from '../utilities';
 import { checkFetchMore } from '../containers/InfiniteScrolling';
 import { AddUserIcon } from '../icons';
 import { IconSize } from '../Icon';
-import { useLazyModal } from '../../hooks/useLazyModal';
 
 export const FollowingFilter = (): ReactElement => {
   const { user } = useAuthContext();
-  const { closeModal } = useLazyModal();
+
   const queryResult = useFollowingQuery({
     id: user.id,
     entity: ContentPreferenceType.User,
@@ -38,9 +37,6 @@ export const FollowingFilter = (): ReactElement => {
             <Link
               href={`${process.env.NEXT_PUBLIC_WEBAPP_URL}users`}
               passHref
-              onClick={() => {
-                closeModal();
-              }}
               className="text-text-link"
             >
               Leaderboards
@@ -53,11 +49,6 @@ export const FollowingFilter = (): ReactElement => {
         isFetchingNextPage,
         canFetchMore: checkFetchMore(queryResult),
         fetchNextPage,
-      }}
-      userInfoProps={{
-        onClick: () => {
-          closeModal();
-        },
       }}
     />
   );
--- a/packages/shared/src/components/filters/FollowingFilter.tsx
+++ b/packages/shared/src/components/filters/FollowingFilter.tsx
@@ -1,10 +1,10 @@
-import React from 'react';
+import React, { type ReactElement } from 'react';
 import { IconSize } from '../Icon';
 import { OpenLinkIcon } from '../icons';
 import { useAuthContext } from '../../contexts/AuthContext';
 import Link from '../utilities/Link';
 
-export const FollowingFilter = (): ReactElement=> {
+export const FollowingFilter = (): ReactElement => {
   const { user } = useAuthContext();
   return (
     <div className="flex flex-col items-center gap-4 px-6 py-10 text-center text-text-secondary">
--- a/packages/shared/src/components/filters/FollowingFilter.tsx
+++ b/packages/shared/src/components/filters/FollowingFilter.tsx
@@ -54,6 +54,11 @@ export const FollowingFilter = (): ReactElement => {
         canFetchMore: checkFetchMore(queryResult),
         fetchNextPage,
       }}
+      userInfoProps={{
+        onClick: () => {
+          closeModal();
+        },
+      }}
     />
   );
 };
--- a/packages/shared/src/components/filters/FollowingFilter.tsx
+++ b/packages/shared/src/components/filters/FollowingFilter.tsx
@@ -1,13 +1,18 @@
 import React, { type ReactElement } from 'react';
+import Link from 'next/link';
 import { useAuthContext } from '../../contexts/AuthContext';
 import UserList from '../profile/UserList';
 import { useFollowingQuery } from '../../hooks/contentPreference/useFollowingQuery';
 import { ContentPreferenceType } from '../../graphql/contentPreference';
 import { FlexCentered } from '../utilities';
 import { checkFetchMore } from '../containers/InfiniteScrolling';
+import { AddUserIcon } from '../icons';
+import { IconSize } from '../Icon';
+import { useLazyModal } from '../../hooks/useLazyModal';
 
 export const FollowingFilter = (): ReactElement => {
   const { user } = useAuthContext();
+  const { closeModal } = useLazyModal();
   const queryResult = useFollowingQuery({
     id: user.id,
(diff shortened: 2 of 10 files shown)


packages/shared/src/components/layout/common.tsx
```diff
@@ -82,8 +82,7 @@ export const SearchControlHeader = ({
     return null;
   }
 
-  const openFeedFilters = () =>
-    openModal({ type: LazyModal.FeedFilters, persistOnRouteChange: true });
```
omBratteng, line 86: A bit unsure if, and why, this modal had to persist route change

packages/shared/src/components/feeds/MobileFeedActions.tsx
```diff
@@ -42,7 +42,6 @@ export function MobileFeedActions(): ReactElement {
           onOpenFeedFilters={() =>
             openModal({
               type: LazyModal.FeedFilters,
-              persistOnRouteChange: true,
```
rebelchris, line 45: Best to check with @sshanzel I think.
Is it not something on submit you get redirected, but we wanted to persist the modal open?

packages/shared/src/components/filters/FollowingFilter.tsx
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
+import { IconSize } from '../Icon';
+import { OpenLinkIcon } from '../icons';
+import { useAuthContext } from '../../contexts/AuthContext';
+import Link from '../utilities/Link';
+
+export const FollowingFilter = (): JSX.Element => {
```
rebelchris, line 7:
```suggestion
export const FollowingFilter = (): ReactElement=> {
```

rebelchris: Why do we even have this component, i don't really get it...

packages/webapp/pages/account/notifications.tsx
```diff
@@ -378,6 +399,14 @@ const AccountNotificationsPage = (): ReactElement => {
         >
           Community updates
         </Checkbox>
+        <Checkbox
+          name="following"
+          data-testid="following-switch"
+          checked={followingEmail}
+          onToggleCallback={onToggleFollowingEmail}
+        >
+          Updates from followed sources, squads and users
```
rebelchris, line 408: We might have to check with @capJavert if we want to change the copy (for the time being to:)
`Updates from followed users`

packages/webapp/pages/account/notifications.tsx
```diff
@@ -419,7 +448,7 @@ const AccountNotificationsPage = (): ReactElement => {
           </>
         )}
       </div>
-      {isPushSupported && (
+      {true && (
```
rebelchris, line 451: Intentional?

rebelchris (changes requested):
Minor queries :)

packages/webapp/pages/account/notifications.tsx
omBratteng: Nope, just had it for local dev, will fix

packages/shared/src/components/filters/FollowingFilter.tsx
omBratteng: I separated it so that the parent component didn't need to add the auth context hook just for the link. But I'm also fine with moving it back into the parent

rebelchris: I mean more from ux perspective, what does it bring anyone this empty screen 🤣
https://dailydotdev.slack.com/archives/C07ATUPBPJ9/p1726728429332939

omBratteng: Oh yeah, was wondering about that too 😬

packages/webapp/pages/account/notifications.tsx
```diff
@@ -509,6 +538,14 @@ const AccountNotificationsPage = (): ReactElement => {
             >
               Reading reminder (Mon-Fri)
             </Checkbox>
+            <Checkbox
+              name="followingPush"
+              data-testid="following-push-switch"
+              checked={followNotifications}
+              onToggleCallback={onToggleFollowingNotifications}
+            >
+              Updates from followed sources, squads and users
```
capJavert, line 547: Update here as well

packages/webapp/pages/account/notifications.tsx
```diff
@@ -96,7 +96,12 @@ const AccountNotificationsPage = (): ReactElement => {
   const personalizedDigestType =
     personalizedDigest?.flags?.sendType || (!isLoading ? SendType.Off : null);
 
-  const { acceptedMarketing, notificationEmail } = user ?? {};
+  const {
+    acceptedMarketing,
+    notificationEmail,
+    followingEmail,
+    followNotifications,
+  } = user ?? {};
```
capJavert, line 104: Is this user object coming from boot or separate user query?

omBratteng: From boot

capJavert: ok, let's keep it like that on API side as well, we will refactor when new notifications settings arrive

packages/webapp/pages/account/notifications.tsx
omBratteng: Updated

packages/webapp/pages/account/notifications.tsx
omBratteng: Fixed

packages/webapp/pages/account/notifications.tsx
omBratteng: Updated

rebelchris (approved):

packages/shared/src/components/feeds/MobileFeedActions.tsx
omBratteng: There's no "submit" on the modal. And I tested in preview that can toggle any setting in the modal without it closing while I have `persistOnRouteChange` disabled.

omBratteng: I had to disable it so that modal would close when I navigate to the profile page 🙈
Could remove the `Link` and have it behave like normal link, that does a full navigation, but it's slower.

capJavert: We can maybe manually close the modal when Link is clicked using `onClick`?

capJavert (approved):

packages/shared/src/components/feeds/MobileFeedActions.tsx
omBratteng: I got a circular import error when I tried to import useLazyModal to get access to closeModal 🙈

omBratteng: actually, think I fixed it, just changed import to type in useLazyModal

omBratteng:
Ah fml, see design is different

omBratteng:
@rebelchris @capJavert ready for re-review, design has changed since last review.

packages/shared/src/components/filters/FollowingFilter.tsx
```diff
@@ -0,0 +1,64 @@
...
+          <AddUserIcon size={IconSize.XXXLarge} />
+          <p>
+            You haven&apos;t follow any User yet.
+            <br /> Explore our{' '}
+            <Link
+              href={`${process.env.NEXT_PUBLIC_WEBAPP_URL}users`}
+              passHref
+              onClick={() => {
+                closeModal();
+              }}
```
rebelchris, line 43: Why do we need to manually close it?
Isn't just putting a `<a>` inside enough to force page load and close it?

packages/shared/src/components/filters/FollowingFilter.tsx
```diff
@@ -0,0 +1,64 @@
...
+      scrollingProps={{
+        isFetchingNextPage,
+        canFetchMore: checkFetchMore(queryResult),
+        fetchNextPage,
+      }}
+      userInfoProps={{
+        onClick: () => {
+          closeModal();
+        },
+      }}
```
rebelchris, line 61: Same with this one, not sure why we need it?
(EG other modals (upvote list etc) don't do this right?

packages/shared/src/components/filters/FollowingFilter.tsx
```diff
@@ -0,0 +1,64 @@
...
+    });
+
+    return acc;
+  }, []);
+
+  return (
+    <UserList
+      users={users}
+      emptyPlaceholder={
+        <FlexCentered className="flex-col gap-4 px-6 py-10 text-center text-text-tertiary typo-callout">
```
rebelchris, line 33: Generic: It seems we need to remove some space (see purple parts) that will make it inline with the other tabs:
Screenshot 2024-09-20 at 07 51 31

rebelchris (changes requested):
Small feedback points.

packages/shared/src/components/filters/FollowingFilter.tsx
omBratteng: An anchor does indeed force page load, but it loads the _entire_ page again, not just next.js magic loading the new content

packages/shared/src/components/filters/FollowingFilter.tsx
omBratteng: Force close because modal is persist on navigation. But saw @sshanzel commented that it's probably legacy, so we can probably remove it and not have to use close-modal

rebelchris: Yeah let's remove it, it's weird to have to do manual stuff basically.

packages/shared/src/components/filters/FollowingFilter.tsx
omBratteng: Fixed... had to set a `!p-0` on the modal body element.

packages/shared/src/components/filters/FollowingFilter.tsx
omBratteng: Removed and removed manual close.

packages/shared/src/components/filters/FollowingFilter.tsx
omBratteng: Removed. Do note in preview it does an entire new page load because of it pointing to prod domain 😬

rebelchris (approved):
Only generic thing, but @capJavert can add seperate ticket:
-  unfollow doesn't invalidate here
- manage should not be needed here (in toast)?

Screenshot 2024-09-20 at 10 52 57

capJavert:
Invalidation is on the list, as for "Manage" from what I see it should show everywhere, honestly I would remove it altogether but will iron out with design/product

PR #3557, opened 2024-09-19
Mi 541 fix scroll actions

nensidosari:
## Changes
- disables next button when there are not enough elements in the view (fixes tag/sources page horizontal scroll as well)

design decision [here](https://dailydotdev.slack.com/archives/C069XNYMDGV/p1726732424418799?thread_ts=1726649072.226739&cid=C069XNYMDGV)

## Events

## Experiment

## Manual Testing

> [!CAUTION]

MI-541 #done

### Preview domain
https://mi-541-fix-scroll-actions.preview.app.daily.dev

### DIFF
--- a/packages/shared/src/components/HorizontalScroll/useCalculateVisibleElements.ts
+++ b/packages/shared/src/components/HorizontalScroll/useCalculateVisibleElements.ts
@@ -33,7 +33,7 @@ export const useCalculateVisibleElements = <El extends HTMLElement>({
 
     setTimeout(() => {
       setIsOverflowing(currentRef.scrollWidth > currentRef.clientWidth);
-    }, 100);
+    }, 1000);
 
     if (currentRef && currentRef.firstElementChild) {
       const element = currentRef.firstElementChild;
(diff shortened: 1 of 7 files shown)


packages/shared/src/components/HorizontalScroll/useCalculateVisibleElements.ts
```diff
@@ -31,7 +31,9 @@ export const useCalculateVisibleElements = <El extends HTMLElement>({
       return;
     }
 
-    setIsOverflowing(currentRef.scrollWidth > currentRef.clientWidth);
+    setTimeout(() => {
+      setIsOverflowing(currentRef.scrollWidth > currentRef.clientWidth);
+    }, 500);
```
nensidosari, line 36: I don't like this very much so open for any other ideas...why this was added was because of the implementation on tags/sources, the scrollWidth takes a bit of time to actually update to the real value, meaning without the timeout, it won't show the buttons at all (it does if u refresh...) :/

nensidosari: @dailydotdev/web-team

sshanzel (approved):
🚀

rebelchris (approved):

PR #3563, opened 2024-09-19
feat: post option follow

capJavert:
## Changes

- add option to menu
- mutations/invalidations of query cache work out of the box

## Events

## Experiment

## Manual Testing

> [!CAUTION]

<!--
Copy and paste the below line outside the HTML comment tags to link this PR to the ticket in Jira

AS-569


packages/shared/src/components/PostOptionsMenu.tsx
```diff
@@ -372,6 +379,34 @@ export default function PostOptionsMenu({
     }
   }
 
+  if (isLoggedIn && post?.author) {
+    const authorName = post.author.name || post.author.username;
+    const isFollowingUser = [
+      ContentPreferenceStatus.Follow,
+      ContentPreferenceStatus.Subscribed,
```
rebelchris, line 386: Nothing urgent, but maybe we need to consolidate this "fn" at some point.
Pretty sure we are going to need it elsewhere.

packages/shared/src/components/PostOptionsMenu.tsx
```diff
@@ -372,6 +379,34 @@ export default function PostOptionsMenu({
...
+          follow({
+            id: post.author.id,
+            entity: ContentPreferenceType.User,
+            entityName: authorName,
+          });
+        } else {
+          unfollow({
+            id: post.author.id,
+            entity: ContentPreferenceType.User,
+            entityName: authorName,
```
rebelchris, line 403: Also I did `@name` (probably should remove the @) but maybe we can have some rules around it internally?

rebelchris (approved):
This part looks good :) 
How come the invalidation worked here already?

packages/shared/src/components/PostOptionsMenu.tsx
capJavert: Added util 6f1d155476649456dea14214d4298990d12160d9

packages/shared/src/components/PostOptionsMenu.tsx
capJavert: yep dc3b29712bdfd86cf5408cfeb6bcada7606e0630

PR #3564, opened 2024-09-19
feat: comment option follow

capJavert:
## Changes

- add comment option menu item
- add mutation subscription handler to update post comments cache

## Events

## Experiment

## Manual Testing

> [!CAUTION]

<!--
Copy and paste the below line outside the HTML comment tags to link this PR to the ticket in Jira

AS-570

### DIFF
--- a/packages/shared/src/components/comments/CommentActionButtons.tsx
+++ b/packages/shared/src/components/comments/CommentActionButtons.tsx
@@ -1,7 +1,7 @@
 import React, { ReactElement, useContext, useEffect, useState } from 'react';
 import classNames from 'classnames';
 import { useQueryClient } from '@tanstack/react-query';
-import AuthContext from '../../contexts/AuthContext';
+import AuthContext, { useAuthContext } from '../../contexts/AuthContext';
 import {
   UpvoteIcon,
   DiscussIcon as CommentIcon,
@@ -10,6 +10,7 @@ import {
   ShareIcon,
   FlagIcon,
   DownvoteIcon,
+  AddUserIcon,
 } from '../icons';
 import { Comment } from '../../graphql/comments';
 import { Roles } from '../../lib/user';
@@ -40,6 +41,9 @@ import {
 import { RequestKey } from '../../lib/query';
 import { useRequestProtocol } from '../../hooks/useRequestProtocol';
 import { getCompanionWrapper } from '../../lib/extension';
+import { useContentPreference } from '../../hooks/contentPreference/useContentPreference';
+import { ContentPreferenceType } from '../../graphql/contentPreference';
+import { isFollowingContent } from '../../hooks/contentPreference/types';
 
 export interface CommentActionProps {
   onComment: (comment: Comment, parentId: string | null) => void;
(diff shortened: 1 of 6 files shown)


packages/shared/src/components/comments/CommentActionButtons.tsx
```diff
@@ -167,6 +173,33 @@ export default function CommentActionButtons({
     });
   }
 
+  if (isLoggedIn && comment?.author) {
```
rebelchris, line 176: Just something that popped up in my head:
1. We should explicit not render if you are the author
2. We should probably exclude 404 user :)

rebelchris: The 404 user is generic for the whole follow system actually, maybe we can make 1 generic `fn` that checks if it's you or 404 it shouldn't continue.

packages/shared/src/components/post/PostComments.tsx
```diff
@@ -67,6 +68,9 @@ export function PostComments({
         refetchOnWindowFocus: false,
       },
     );
+
+  useCommentContentPreferenceMutationSubscription({ queryKey });
```
rebelchris, line 72: This does start to feel a bit dangerous already.
Eg companion comments are now not accounted for (not that, that specific case is a big issue, but just saying with manually having to mutate other stuff it becomes quickly easy to forget a spot or not do it.

rebelchris (approved):
As mentioned before, not against it, but the mutation subscribe way is a bit dangerous to account for everywhere.

packages/shared/src/components/post/PostComments.tsx
capJavert: I have to double check companion still yes, but it is using the same component, probably issue is that it does not use QueryClient 🙈 

I would not say its dangerous but introduces a mindset shift where we think about invalidation as separate component which we need to take care of when building components and if components need it. Muck like state, styling etc. Would love to go deeper into this approach and discuss pros/cons.

packages/shared/src/components/comments/CommentActionButtons.tsx
capJavert: yes, good thinking

packages/shared/src/components/post/PostComments.tsx
capJavert: I remove follow option on comments in companion for now, will recheck it at the end. Extension works.

packages/shared/src/components/comments/CommentActionButtons.tsx
```diff
@@ -167,6 +174,39 @@ export default function CommentActionButtons({
...
   }
 
+  const shouldShowFollow =
+    !useIsSpecialUser({ userId: comment?.author?.id }) &&
+    isLoggedIn &&
+    comment?.author &&
+    !isCompanion;
+
+  if (shouldShowFollow) {
+    const authorName = comment.author.name || `@${comment.author.username}`;
```
rebelchris, line 184: Maybe let's add a ticket to make utility function for the `entityName` just because I've noticed we do different things :)

rebelchris (approved):

packages/shared/src/components/comments/CommentActionButtons.tsx
capJavert: https://dailydotdev.atlassian.net/browse/AS-594

PR #3568, opened 2024-09-20
fix: squad list modal

rebelchris:
## Changes

Added some changes so we can render items before and after the follow button in UserShortInfo (needed for squads now)

Screenshot 2024-09-20 at 13 53 40

## Events

## Experiment

## Manual Testing

> [!CAUTION]

AS-593 #done

### DIFF
--- a/packages/shared/src/components/modals/SquadMemberModal.tsx
+++ b/packages/shared/src/components/modals/SquadMemberModal.tsx
@@ -16,12 +16,13 @@ import { LinkIcon } from '../icons';
 import { useSquadInvitation } from '../../hooks/useSquadInvitation';
 import { FlexCentered } from '../utilities';
 import { useSquadActions } from '../../hooks';
-import SquadMemberItemAdditionalContent from '../squads/SquadMemberItemAdditionalContent';
+import SquadMemberItemRole from '../squads/SquadMemberItemRole';
 import { verifyPermission } from '../../graphql/squads';
 import useDebounceFn from '../../hooks/useDebounceFn';
 import { defaultSearchDebounceMs } from '../../lib/func';
 import { BlockedMembersPlaceholder } from '../squads/Members';
 import { ContextMenu } from '../../hooks/constants';
+import SquadMemberItemOptionsButton from '../squads/SquadMemberItemOptionsButton';
 
 enum SquadMemberTab {
   AllMembers = 'Squad members',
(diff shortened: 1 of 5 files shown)


packages/shared/src/components/modals/SquadMemberModal.tsx
```diff
@@ -122,8 +123,13 @@ export function SquadMemberModal({
           onScroll: hideMenu,
         }}
         userListProps={{
-          additionalContent: (user, index) => (
-            <SquadMemberItemAdditionalContent
+          additionalContent: (user, index) => [
```
rebelchris, line 126: Not 100% sure if I like this array object flow, so let me know please 🤣 
It was just the easiest.

nensidosari: Yeah not sure I like it either... 😅 

Had to dig in to see how we have composed this components. I think maybe ideally the UserShortInfo component should allow for all of these additional details to user if provided props to show it. Still there seems to be a lot of info being passed around until u end up to UserShortInfo so might need a better look at it all if we want to refactor. 

Saying that, if we want to move fast with it, think we're good. Let's just note it for further improvements.

packages/shared/src/components/profile/UserList.tsx
```diff
@@ -50,8 +53,9 @@ function UserList({
               tag="a"
               href={user.permalink}
               user={user}
+              endChildren={additionalContent?.(user, i)[1]}
             >
-              {additionalContent?.(user, i)}
+              {additionalContent?.(user, i)[0]}
```
nensidosari, line 58: why 0 here?

packages/shared/src/components/profile/UserShortInfo.tsx
```diff
@@ -32,6 +32,7 @@ export interface UserShortInfoProps<
   scrollingContainer?: HTMLElement;
   appendTooltipTo?: HTMLElement;
   children?: ReactNode;
+  endChildren?: ReactNode;
```
nensidosari, line 35: Now it became more scary 😅

nensidosari (approved):
Except the issue or not best solution u've mentioned, looks good!!

packages/shared/src/components/modals/SquadMemberModal.tsx
rebelchris: 100% could probably do with a refactor.
In general it's not needing anything, only this squad one adds "for now" but expect it to have exceptions later on again.

packages/shared/src/components/profile/UserList.tsx
rebelchris: Index zero on the array of components 😅
That's why I mentioned it's not super readable/clean

nensidosari: Yeah no, that's why I wasn't expecting this 😅 was expecting all of the return array to be mapped here somewhere

rebelchris:
@nensidosari Decided to fully do it the "nice" way :) 
@capJavert invalidation works :)

capJavert (approved):
To me this looks good 👌 Tnx for the extra effort

PR #3573, opened 2024-09-21
fix: button not clicking in mobile, members in my squads

nensidosari:
## Changes
- not show members in my squad https://dailydotdev.slack.com/archives/C069XNYMDGV/p1726872246644799
- button working on mobile https://dailydotdev.slack.com/archives/C069XNYMDGV/p1726897403685179 & https://dailydotdev.slack.com/archives/C069XNYMDGV/p1726898361240879

### Preview domain
https://fix-public-squad-issues.preview.app.daily.dev

### DIFF
--- a/packages/shared/src/components/cards/squad/SquadsDirectoryFeed.tsx
+++ b/packages/shared/src/components/cards/squad/SquadsDirectoryFeed.tsx
@@ -1,6 +1,4 @@
 import React, { ReactElement, ReactNode } from 'react';
-import Link from 'next/link';
-import { useRouter } from 'next/router';
 import { Squad } from '../../../graphql/sources';
 import {
   SourcesQueryProps,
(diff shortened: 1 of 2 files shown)


packages/shared/src/components/cards/squad/SquadsDirectoryFeed.tsx
```diff
@@ -1,4 +1,6 @@
 import React, { ReactElement, ReactNode } from 'react';
+import Link from 'next/link';
```
rebelchris, line 2: This is the wrong one right we need the custom one since next 14

packages/shared/src/components/cards/squad/SquadsDirectoryFeed.tsx
```diff
@@ -57,7 +60,16 @@ export function SquadsDirectoryFeed({
         {children}
         <header className="mb-2 flex flex-row items-center justify-between">
           {title}
-          <Button variant={ButtonVariant.Tertiary}>See all</Button>
+          <Link href={linkToSeeAll} passHref>
+            <Button
+              variant={ButtonVariant.Tertiary}
+              onClick={() => router.push(linkToSeeAll)}
```
rebelchris, line 66: Why does it do onclick as well as href? One should be enough

rebelchris (changes requested):

packages/shared/src/components/cards/squad/SquadsDirectoryFeed.tsx
nensidosari: 🤦‍♀️

rebelchris (approved):

PR #3575, opened 2024-09-23
feat: profile follow button

rebelchris:
## Changes

Since we currently only fetch client side on same user or no initial data we had to remove this.
As server isn't aware of the logged user.

There is a stale time so should be fine :) 
Also added the mutation invalidation handler.

## Events

## Experiment

## Manual Testing

> [!CAUTION]

AS-597 #done 

### Preview domain
https://as-597-profile-follow-button.preview.app.daily.dev

### DIFF
--- a/packages/shared/src/components/contentPreference/FollowButton.tsx
+++ b/packages/shared/src/components/contentPreference/FollowButton.tsx
@@ -67,7 +67,7 @@ export const FollowButton = ({
 
   const isLoading = isLoadingFollow || isLoadingNotify;
 
-  if (user.id === userId) {
+  if (!user || user.id === userId) {
     return null;
   }
 
--- a/packages/shared/src/components/profile/Header.tsx
+++ b/packages/shared/src/components/profile/Header.tsx
@@ -81,15 +81,13 @@ export function Header({
         icon={<ShareIcon />}
         onClick={() => onShareOrCopyLink()}
       />
-      {!isSameUser && (
-        <FollowButton
-          userId={user.id}
-          type={ContentPreferenceType.User}
-          status={(user as LoggedUser).contentPreference?.status}
-          entityName={`@${user.username}`}
-          className="ml-2"
-        />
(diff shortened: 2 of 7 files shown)


packages/shared/src/components/contentPreference/FollowButton.tsx
```diff
@@ -67,7 +67,7 @@ export const FollowButton = ({
 
   const isLoading = isLoadingFollow || isLoadingNotify;
 
-  if (user.id === userId) {
+  if (!user || user.id === userId) {
```
rebelchris, line 70: Edge-case :)

packages/shared/src/components/profile/Header.tsx
```diff
@@ -34,7 +36,7 @@ export function Header({
   });
   const isMobile = useViewSize(ViewSize.MobileL);
   const [isMenuOpen, setIsMenuOpen] = useState(false);
-
+  console.log(user);
```
omBratteng:
```suggestion
```

packages/shared/src/components/profile/Header.tsx
```diff
@@ -80,6 +81,15 @@ export function Header({
         icon={<ShareIcon />}
         onClick={() => onShareOrCopyLink()}
       />
+      {!isSameUser && (
```
capJavert, line 84: you don't need this since its inside the button?

packages/shared/src/hooks/profile/useProfile.ts
```diff
@@ -1,24 +1,33 @@
...
     {
       ...disabledRefetch,
       cacheTime: StaleTime.OneHour,
       initialData: initialUser,
-      enabled: !!initialUser && isSameUser,
     },
   );
 
-  return user;
+  return useMemo(
```
capJavert, line 26: no need for useMemo

capJavert (approved):

PR #3580, opened 2024-09-23
fix: devcard positioning after hover

sshanzel:
## Changes
- The primary issue of the devcard getting cut was due to when we trigger the tooltip element and fetch the data.
- We first render the element of the tooltip, which calculates whether it would fit the screen, and then we fetch data which then when the element gets filled up, the tooltip thinks it should fit.
- I introduced a wrapper to keep backwards support, especially with the GitHub integration and download feature.

Tooltip is now able again to dynamically position itself as can be seen from the video below.

https://github.com/user-attachments/assets/2ba60ba2-733d-45de-8e8f-d96d3e77ebbb

## Events

## Experiment

## Manual Testing

> [!CAUTION]

### Preview domain
https://mi-549-screen-cutoff.preview.app.daily.dev


packages/shared/src/components/profile/ProfileTooltip.tsx
```diff
@@ -19,49 +20,54 @@ export interface ProfileTooltipProps extends ProfileTooltipContentProps {
...
-  user,
+  userId,
   link,
   scrollingContainer,
   tooltip = {},
-}: Omit<ProfileTooltipProps, 'user'> & {
-  user?: Partial<Author>;
-}): ReactElement {
+}: Omit<ProfileTooltipProps, 'user'>): ReactElement {
+  const [id, setId] = useState<string>();
```
sshanzel, line 36: We set it in a state so that we won't have to immediately request the devcard data, until only when the user triggers the tooltip.

sshanzel:
@dailydotdev/web-team should be ready for review 🥳

nensidosari (approved):
Looks great

PR #3581, opened 2024-09-24
fix: stop expanding the post author so tooltip wont trigger

sshanzel:
## Changes
- The primary concern here is the post content, when the author is present, the container expands so when hovering the right side of the container, tooltip is triggered which it should not.

## Events

## Experiment

## Manual Testing

> [!CAUTION]

### DIFF
--- a/packages/shared/src/components/post/SquadPostAuthor.tsx
+++ b/packages/shared/src/components/post/SquadPostAuthor.tsx
@@ -68,7 +68,7 @@ function SquadPostAuthor({
         <a
           href={author.permalink}
           className={classNames(
-            'ml-4 flex flex-1 flex-col overflow-hidden',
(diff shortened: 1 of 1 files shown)


packages/shared/src/components/post/SquadPostAuthor.tsx
```diff
@@ -68,7 +68,7 @@ function SquadPostAuthor({
         <a
           href={author.permalink}
           className={classNames(
-            'ml-4 flex flex-1 flex-col overflow-hidden',
```
sshanzel, line 71: There are only two places where we use this specific component. And on the other one, we pass the `flex-1` already, and the broken one does not need this.

rebelchris (approved):

omBratteng (approved):

PR #3584, opened 2024-09-25
chore: better a11y for onboarding;

ilasw:
## Changes
- Change on form inputs to make sure that can accept accessibility props;
- Updated Onboarding removing for screen readers presentational elements and adding label for a good navigation;
- Autofocus the first access point for inputs. 

> I added autofocus attribute into form fields, but we should test how it behaves for users. Scared that the browser will scroll down bringing the input inside the viewport.

## Events

## Experiment

## Manual Testing

> [!CAUTION]

### Preview domain
https://fix-onboarding-accessibility.preview.app.daily.dev


packages/shared/src/components/ProfilePicture.tsx
```diff
@@ -101,13 +101,15 @@ function ProfilePictureComponent(
     return null;
   }
 
+  const imageAlt = `${user.username || user.name || user.id}'s profile`;
```
ilasw, line 104: Alt for those kind of images was "Undedined's profile picture"

packages/shared/src/components/auth/AuthHeader.tsx
```diff
@@ -21,9 +21,14 @@ function AuthHeader({
...
   className,
   onBack,
+  ...attrs
 }: AuthHeaderProps): ReactElement {
   if (simplified) {
-    return <h2 className="text-center font-bold typo-title2">{title}</h2>;
+    return (
+      <h2 {...attrs} className="text-center font-bold typo-title2">
+        {title}
+      </h2>
```
ilasw, line 30: Need to pass id for `aria-labelledby`

packages/shared/src/components/auth/AuthSignBack.tsx
```diff
@@ -73,7 +73,7 @@ export const AuthSignBack = ({
         <ConditionalWrapper
           condition={provider !== 'password'}
           wrapper={(component) => (
-            <div className="relative">
+            <div className="relative" aria-hidden>
```
ilasw, line 76: Onboarding's profile image is not a required information on screen reader, maybe is better to skip just to the login button since image cannot be seen in those cases.

packages/shared/src/components/auth/LoginForm.tsx
```diff
@@ -52,6 +52,7 @@ function LoginForm({
 
   return (
     <AuthForm
+      aria-label="Login using email and password"
```
ilasw, line 55: Made this to give a name to the form, this should be mandatory but, as a general optimization,  we can also make it point with `aria-labelledby` to the form title if we are 100% sure to have one

packages/shared/src/components/auth/OrDivider.tsx
```diff
@@ -9,10 +9,12 @@ interface OrDividerProps {
 function OrDivider({ className, label = 'or' }: OrDividerProps): ReactElement {
   return (
     <div
+      aria-hidden
       className={classNames(
         'flex items-center justify-center text-text-quaternary typo-callout',
         className,
       )}
+      role="separator"
```
ilasw, line 17: I don't think that "or" need to be read, but open to discussion

packages/shared/src/components/auth/SignBackButton.tsx
```diff
@@ -21,6 +21,9 @@ export function SignBackButton({
       className="btn-signback btn-primary"
       type="button"
       onClick={onClick}
+      aria-label={`Continue as ${
+        signBack.name || signBack.email
+      } from ${provider}`}
```
ilasw, line 26: trying to give more context to the user

packages/shared/src/components/footer/FooterLinks.tsx
```diff
@@ -9,33 +9,35 @@ export type FooterLinksProps = {
...
-        <a href={`${webappUrl}sources`}>Sources</a>
-      </li>
-      <li>
-        <a href={`${webappUrl}squads`}>Squads</a>
-      </li>
-      <li>
-        <a href={`${webappUrl}users`}>Leaderboard</a>
-      </li>
-    </ul>
+    <footer>
```
ilasw, line 12: Those links are always in the end of the page, footer is a better tag element imo

packages/shared/src/components/Feed.spec.tsx
```diff
@@ -846,7 +846,7 @@ describe('Feed', () => {
     const title = await screen.findByTestId('post-modal-title');
     expect(title).toHaveTextContent(firstPost.node.title);
 
-    await screen.findByRole('navigation');
+    await screen.findAllByRole('navigation');
```
ilasw, line 849: wasn't working because is finding more than one element that has this role

packages/shared/src/components/Icon.tsx
```diff
@@ -27,11 +27,9 @@ export const iconSizeToClassName: Record<IconSize, string> = {
 
 type ItemType = React.ComponentType<{ className }>;
 
-export interface IconProps {
+export interface IconProps extends ComponentProps<'svg'> {
   secondary?: boolean;
   size?: IconSize;
```
ilasw, line 32: Need to make the icons able to receive `aria-` attributes

packages/shared/src/components/fields/Checkbox.tsx
```diff
@@ -30,13 +31,14 @@ export const Checkbox = forwardRef(function Checkbox(
...
     onToggleCallback,
-    id,
+    id = '',
     disabled,
     ...props
   }: CheckboxProps,
   ref: LegacyRef<HTMLInputElement>,
 ): ReactElement {
   const [actualChecked, setActualChecked] = useState(checked);
+  const checkId = useId();
```
ilasw, line 41: Make sure to have a valid id on each use, sometime right now we do not have id and as a consequence

packages/shared/src/components/auth/AuthDefault.tsx
```diff
@@ -179,6 +198,7 @@ const AuthDefault = ({
               button: shouldLogin ? 'Sign up' : 'Log in',
             }}
             onClick={() => {
+              focusFirstSocialLink();
```
ilasw, line 201: By clicking this element the component will refresh and we want to move the focus on top just to make the navigation easier

packages/shared/src/components/auth/EmailSignupForm.tsx
```diff
@@ -18,26 +18,44 @@ function EmailSignupForm({
...
   const [email, setEmail] = useState(null);
+  const inputId = useId();
+  const isSubmitDisabled = !email || !isReady;
 
   return (
-    <AuthForm className="gap-2" onSubmit={onSubmit}>
+    <AuthForm
+      aria-label="Signup using email"
+      className="gap-2"
+      onSubmit={onSubmit}
```
ilasw, line 28: Every form should be announced correctly by having the `aria-label`

packages/shared/src/components/fields/SearchField.tsx
```diff
@@ -45,8 +47,17 @@ export interface SearchFieldProps
...
+  isPrimary,
+  ...attrs
+}: IconProps & {
+  isPrimary: boolean;
+}) =>
+  isPrimary ? (
+    <CloseIcon {...attrs} />
+  ) : (
+    <ArrowIcon {...attrs} className="rotate-90" />
+  );
```
ilasw, line 60: Need to pass the `aria-` attributes to the icons

packages/shared/src/components/Icon.tsx
rebelchris: ComponentProps is prefered anyway :)

packages/shared/src/components/auth/EmailSignupForm.tsx
```diff
@@ -18,26 +18,44 @@ function EmailSignupForm({
...
+            aria-label="Submit and sign me up, accepting terms and conditions"
             size={ButtonSize.Small}
             variant={ButtonVariant.Primary}
-            icon={<ArrowIcon className="rotate-90" />}
+            icon={
+              <ArrowIcon
+                aria-hidden
+                className="rotate-90"
+                role="presentation"
+              />
```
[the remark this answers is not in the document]
rebelchris: not urgent, but we should probably add some config to `Icon` component maybe to make it easier.

rebelchris (approved):
Minor feedback, we just need to be careful about the focus and manually test on some devices.
As well as the impact of the layout focus addition.

packages/shared/src/components/Logo.tsx
```diff
@@ -72,19 +75,20 @@ export default function Logo({
...
     <LinkWithTooltip
-      href={process.env.NEXT_PUBLIC_WEBAPP_URL}
+      href={homeHref}
       passHref
       prefetch={false}
       tooltip={{ placement: 'right', content: 'Home' }}
     >
-      {/* eslint-disable-next-line jsx-a11y/click-events-have-key-events,jsx-a11y/no-static-element-interactions */}
       <a
+        aria-disabled={linkDisabled}
```
sshanzel, line 84: Should we also add `disabled={linkDisabled}`? Just so the behavior would work as how it is explained by the aria.

packages/shared/src/components/auth/AuthDefault.tsx
```diff
@@ -80,6 +94,10 @@ const AuthDefault = ({
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [shouldLogin]);
 
+  useLayoutEffect(() => {
+    focusFirstSocialLink();
+  }, []);
```
sshanzel, line 99: @rebelchris I remember one session we had with tkDodo about use effects. Is it maybe possible to have this within the `ref` callback of the `div` element itself? I think so it should.

packages/shared/src/components/fields/Checkbox.tsx
```diff
@@ -57,12 +59,13 @@ export const Checkbox = forwardRef(function Checkbox(
         { checked: actualChecked, disabled },
       )}
       style={{ transition: 'color 0.1s linear' }}
-      htmlFor={id}
+      htmlFor={id.concat(checkId)}
```
sshanzel, line 62: We should maybe put this in a variable, to avoid any issue down the line.

sshanzel (approved):
Just had one comment, but looks amazing overall. Great work btw 🙏

packages/shared/src/components/Logo.tsx
ilasw: I've tried but linter suggested to remove because we should not use it in links

packages/shared/src/components/fields/Checkbox.tsx
ilasw: Updated ✅

nensidosari (approved):
Great work!

packages/shared/src/components/Logo.tsx
```diff
@@ -59,6 +60,8 @@ interface LogoProps {
   linkDisabled?: boolean;
 }
 
+const homeHref = process.env.NEXT_PUBLIC_WEBAPP_URL;
```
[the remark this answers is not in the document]
ilasw: Updated! ✅

packages/shared/src/components/auth/AuthDefault.tsx
rebelchris: Yeah should be possible indeed

PR #3591, opened 2024-09-26
fix: change to only moderator post

nensidosari:
As per request https://dailydotdev.slack.com/archives/C02AA18RD1D/p1727269244960479

### Preview domain
https://fix-oly-mod-post.preview.app.daily.dev


omBratteng (approved):

idoshamun (approved):
Thanks! Can you also make sure we apply this by default when we create squad automatically via the "create new squad" drop-down in the new post button

nensidosari:
Merged as at the moment we have no defaults for the new post, so this won't change a thing. Will raise a pr for that in a min

PR #3597, opened 2024-09-28
fix: Display storybook components correctly

AmarTrebinjac:
## Changes

Initially this was supposed to be a task to implement ActionButtons into Storybook, but this task had already been completed.

However, several components were rendering the server error page due to recent changes instead of the actual component. This PR solves those issues.

## Events

## Experiment

## Manual Testing

> [!CAUTION]

MI-555

### Preview domain
https://storybook-git-mi-555-dailydotdev.vercel.app/?path=/story/components-actionbuttons--default


packages/storybook/mock/next-router.ts
```diff
@@ -1,7 +1,8 @@
 import { fn } from '@storybook/test';
 
-export const useRouter = fn().mockReturnValue({
+export const useRouter = () => ({
```
sshanzel, line 3: Would there be any difference in using this over the previous one? Asking just so we can have clarity and which one to go for on the next ones.

packages/storybook/stories/extension/_providers.tsx
```diff
@@ -1,11 +1,13 @@
...
-import { BootApp } from '@dailydotdev/shared/src/lib/boot';
+import {
+  QueryClient,
+  QueryClientProvider as TanStackQueryClientProvider,
+} from '@tanstack/react-query';
 import { Browser } from 'webextension-polyfill';
 import type { FC, PropsWithChildren } from 'react';
 
-const app = BootApp.Extension;
+const app = 'extension';
```
sshanzel, line 10: Why did we use literal string over the enum?

sshanzel (approved):
Looks amazing. Just had some questions but nothing blocking 🚀

packages/storybook/stories/extension/_providers.tsx
rebelchris: Yeah think this should just be the enum :)

rebelchris (approved):

packages/storybook/stories/extension/_providers.tsx
ilasw: We tried to use the enum but a `Detected cycle while resolving name 'BootApp' in '/mock/boot.ts` error popped up. 

I think that fixing this issue could need more than half a day, just because we can not import the files with an absolute path and a mock for both @dailydotdev/shared/src/lib/boot' and '../lib/boot' is needed in `.storybook/main.ts` :(

We can further investigate if needed!

packages/storybook/mock/next-router.ts
AmarTrebinjac: When using "mockReturnValue" the router was erroring with a "Cannot destructure property 'query' of 'useRouter(...)' as it is undefined.". We didn't really figure out why it didn't work, but @ilasw suggested we just remove the mock because the router does not need to be tested at this point, so we can just add a skeleton so that it doesn't cause errors.

ilasw (approved):

PR #3598, opened 2024-09-28
fix: HTML Email validator on account page fix

AmarTrebinjac:
## Changes

The submit buttons in the form were calling "onClick" instead of "onSubmit", which prevented the HTML validation from triggering. However, when having two submit buttons in one form we have to differentiate between which one is clicked. Now there's a shared submit handler that checks the name of the submit button and calls the appropriate submit function.

## Events

## Experiment

## Manual Testing

> [!CAUTION]

MI-501

### Preview domain
https://mi-501.preview.app.daily.dev

### DIFF
--- a/packages/webapp/__tests__/AccountSecurityPage.tsx
+++ b/packages/webapp/__tests__/AccountSecurityPage.tsx
@@ -158,8 +158,17 @@ it('should allow changing of email', async () => {
   };
   mockWhoAmIFlow(email);
   mockSettingsValidation(params);
-  const submitChanges = await screen.findByText('Send code');
-  fireEvent.click(submitChanges);
+
+  const firstSubmitButton = await screen.findByText('Send code');
+  const mockFirstSubmit = new Event('submit', {
+    bubbles: true,
+    cancelable: true,
+  });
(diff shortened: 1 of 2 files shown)


packages/webapp/__tests__/AccountSecurityPage.tsx
```diff
@@ -158,8 +158,17 @@ it('should allow changing of email', async () => {
...
-  fireEvent.click(submitChanges);
+
+  const firstSubmitButton = await screen.findByText('Send code');
+  const mockFirstSubmit = new Event('submit', {
+    bubbles: true,
+    cancelable: true,
+  });
+  Object.defineProperty(mockFirstSubmit, 'submitter', {
+    value: firstSubmitButton,
+  });
```
AmarTrebinjac, line 169: React Testing Library does not replicate a "real" form submission event, meaning some properties get omitted, such as submitter. One way to solve that is to add the add the event manually.

sshanzel: Oh, wow. I did not know about this one. Great info.

sshanzel (approved):
TIL :shipit:

PR #3608, opened 2024-10-01
fix: change squad activeFeed name to reflect new path;

ilasw:
## Changes
On mobile Squad directory page was not marked as active on navigation, quickly checked and was a 2 line fix

## Events

## Experiment

## Manual Testing

> [!CAUTION]

### Preview domain
https://fix-squad-mobile-tab-item-active.preview.app.daily.dev

### DIFF
--- a/packages/shared/src/lib/query.ts
+++ b/packages/shared/src/lib/query.ts
@@ -26,8 +26,8 @@ import {
 export enum OtherFeedPage {
   Tag = 'tag',
   Tags = 'tags',
-  Squad = 'squads',
-  SquadPage = 'squads[handle]',
+  Squad = 'squadsdiscover',
+  SquadPage = 'squadsdiscover[id]',
   Source = 'source',
   Sources = 'sources',
   Leaderboard = 'users',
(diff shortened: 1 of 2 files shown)


capJavert (approved):

nensidosari (approved):

packages/shared/src/lib/query.ts
```diff
@@ -26,8 +26,8 @@ import {
 export enum OtherFeedPage {
   Tag = 'tag',
   Tags = 'tags',
-  Squad = 'squads',
-  SquadPage = 'squads[handle]',
+  Squad = 'squadsdiscover',
```
rebelchris, line 29: I think we might do some eval somewhere, but not sure it's on this.
To check if page === some value it can render the name differently.

Just noting it so we keep a close eye on mobile reports after deploying :)

rebelchris (approved):
Non blocking, just need to monitor closely.

packages/shared/src/lib/query.ts
capJavert: I think the tests failed because of it, we have some string evals/types.

rebelchris: @ilasw Yeah it also breaks on mobile now.
The squad is not shown as active at the moment.

Maybe best to revert this and re-look at it.

PR #3609, opened 2024-10-01
refactor: avoid unnecessary calls and properties in squads directory page

sshanzel:
## Changes
- Adjusted the fragments so that only the necessary (or at least not heavy) properties are fetched for the cards.
- Utilized in view to check whether the element is in view to decide only then we will fetch the values.

Preview:

https://github.com/user-attachments/assets/88da62bc-0183-40fa-8061-66e1166eeb35

## Events

## Experiment

## Manual Testing

> [!CAUTION]

MI-561

### Preview domain
https://mi-561.preview.app.daily.dev

### DIFF
--- a/packages/shared/src/components/cards/squad/SquadsDirectoryFeed.tsx
+++ b/packages/shared/src/components/cards/squad/SquadsDirectoryFeed.tsx
@@ -59,7 +59,7 @@ export function SquadsDirectoryFeed({
 
   if (isMobile) {
     return (
-      <div className="relative flex flex-col gap-3 pb-6">
+      <div ref={ref} className="relative flex flex-col gap-3 pb-6">
         {children}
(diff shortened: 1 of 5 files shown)


packages/shared/src/components/cards/squad/SquadsDirectoryFeed.tsx
```diff
@@ -41,14 +42,18 @@ export function SquadsDirectoryFeed({
   className,
   children,
 }: SquadHorizontalListProps): ReactElement {
-  const { result } = useSources<Squad>({ query });
-  const { isInitialLoading } = result;
```
sshanzel, line 45: This prop is actually going to be removed at next major upgrade.

nensidosari (approved):

rebelchris (approved):
Just monitor it on prod, but code wise looks good to me.

PR #3611, opened 2024-10-02
fix: missing date range on extension and mobile

sshanzel:
## Changes
- Fix missing date range on mobile and extension.
- Removed unnecessary dropdown about algo (since explore tabs, we don't need that anymore).

## Events

## Experiment

## Manual Testing

> [!CAUTION]

MI-508 #done

### Preview domain
https://mi-508.preview.app.daily.dev


packages/shared/src/components/header/FeedExploreHeader.tsx
```diff
@@ -117,7 +122,7 @@ export function FeedExploreHeader({
         )}
         {showDropdown && (
           <span className="ml-auto">
-            {withDateRange.includes(path as OtherFeedPage) && (
+            {shouldShowDropdown && (
```
nensidosari, line 125: would it make sense to use ExploreHeader here as well?

packages/shared/src/components/feeds/FeedNav.tsx
```diff
@@ -152,27 +137,6 @@ function FeedNav(): ReactElement {
             <Tab key={label} label={label} url={url} />
           ))}
         </TabContainer>
-
```
nensidosari, line 155: which one is this we're removing, I just can't seem to recall?

sshanzel: I should revert this. Give me one moment. Reference: https://dailydotdev.slack.com/archives/C02AA18RD1D/p1727856011408519

packages/shared/src/components/header/FeedExploreHeader.tsx
sshanzel: Won't work here since the other is used for mobile display and the layout is different.

nensidosari: Would prefer then a more specific naming, like MobileExploreHeading for the new one if it makes sense.

sshanzel: Yeah, makes sense. Let's update it.

sshanzel: Updated the component now.

packages/shared/src/components/header/ExploreHeader.tsx
nensidosari, line 1: @sshanzel let's change the file name as well 🙏 😅

nensidosari (approved):

PR #3613, opened 2024-10-02
feat: micro experiment search placeholder

AmarTrebinjac:
## Changes
Introduced a feature for experimenting with search text, particularly placeholder in this case.

## Events

## Experiment

## Manual Testing

> [!CAUTION]

MI-564

### Preview domain
https://mi-564.preview.app.daily.dev

### DIFF
--- a/packages/shared/src/components/search/SearchPanel/SearchPanelInput.tsx
+++ b/packages/shared/src/components/search/SearchPanel/SearchPanelInput.tsx
@@ -75,7 +75,7 @@ export const SearchPanelInput = ({
     useInputField(value, valueChanged);
   const { isLoggedIn, showLogin } = useAuthContext();
   const isLaptop = useViewSize(ViewSize.Laptop);
-  const searchText = useFeature(feature.searchText);
+  const placeholder = useFeature(feature.searchPlaceholder);
 
   const onInputClick = () => {
     if (!isLoggedIn) {
@@ -193,7 +193,7 @@ export const SearchPanelInput = ({
             placeholder={
               searchPanel.isActive
                 ? 'Search posts or ask a question...'
-                : searchText.placeholder
+                : placeholder
             }
             ref={inputRef}
             onFocus={(event) => {
--- a/packages/shared/src/components/search/SearchPanel/SearchPanelInput.tsx
+++ b/packages/shared/src/components/search/SearchPanel/SearchPanelInput.tsx
@@ -191,7 +191,9 @@ export const SearchPanelInput = ({
             {...inputProps}
             data-search-panel-item="true"
             placeholder={
-              searchPanel.isActive ? searchText.placeholder : 'Search'
+              searchPanel.isActive
+                ? 'Search posts or ask a question...'
+                : searchText.placeholder
             }
             ref={inputRef}
             onFocus={(event) => {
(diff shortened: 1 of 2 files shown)


packages/shared/src/components/search/SearchPanel/SearchPanelInput.tsx
```diff
@@ -68,14 +70,12 @@ export const SearchPanelInput = ({
     onFocus: externalOnFocus,
     onBlur: externalOnBlur,
     onClick: externalOnClick,
-    placeholder = searchPanel.isActive
-      ? 'Search posts or ask a question...'
-      : 'Search',
```
nensidosari, line 73: Let's make sure here they want to experiment with both of these versions 🙏

AmarTrebinjac: > Let's make sure here they want to experiment with both of these versions 🙏

You mean you want the experiment to only be displayed when the search panel is "active"? 😄

nensidosari: Well, yeah that's what we need to check with product in this case, highlight to them that in the meantime we have 2 placeholders for when the panel is active and when panel is not active. 
Just because with the changes we're doing now in this PR, we're defaulting to only one placeholder for both.

nensidosari: As for where to check this with product, u can either do on product-chatter channel, or in the experiment approval thread you added in experiments channel by tagging @product

packages/shared/src/lib/featureManagement.ts
```diff
@@ -30,6 +30,9 @@ const feature = {
   onboardingChecklist: new Feature('onboarding_checklist', true),
   searchUsers: new Feature('search_users', false),
   showCodeSnippets: new Feature('show_code_snippets', false),
+  searchText: new Feature('search_text', {
+    placeholder: 'Search',
+  }),
```
sshanzel, line 35: Normally, we resort to using just a string when it comes to copy changes so this can be just a string literal.
```suggestion
  searchText: new Feature('search_text', 'Search'),
```

sshanzel (commented):
Looks amazing! Just about how we normally define the values. We use object for complex experiments such as onboarding where we are trying to change multiple things at once.

packages/shared/src/lib/featureManagement.ts
nensidosari: @sshanzel I created that just in case product wants to play with different parts of search text. But we can convert if we want to I don't mind.

nensidosari: Changed to string only since it doesn't seem product want to experiment differently 

https://app.growthbook.io/features/search_placeholder

flag is search_placeholder and is a string defaulting to Search

cc. @AmarTrebinjac

packages/shared/src/components/search/SearchPanel/SearchPanelInput.tsx
```diff
@@ -44,7 +46,7 @@ export type SearchPanelInputClassName = {
 export type SearchPanelInputProps = {
   className?: SearchPanelInputClassName;
   valueChanged?: (value: string) => void;
-  inputProps?: InputHTMLAttributes<HTMLInputElement>;
+  inputProps?: Omit<InputHTMLAttributes<HTMLInputElement>, 'placeholder'>;
```
nensidosari, line 49: why this change?

nensidosari: Hmm, seems ok to remove actually since it's not being used

nensidosari (approved):
Looks good. Please before merging, update the experiment message with the new values of feature flag

packages/shared/src/components/search/SearchPanel/SearchPanelInput.tsx
AmarTrebinjac: @nensidosari Like you said, it's not being used, so omitting it as to not have it show up as an available property, giving the impression that it does something  😅

PR #3616, opened 2024-10-02
fix: gitpod cors issue

sshanzel:
## Changes
- Fix GitPod CORS issue by proxying the api.
- Tested in GitPod. Works as intended.

## Events

## Experiment

## Manual Testing

> [!CAUTION]

### Preview domain
https://fix-gitpod.preview.app.daily.dev

### DIFF
--- a/packages/webapp/next.config.js
+++ b/packages/webapp/next.config.js
@@ -83,26 +83,38 @@ module.exports = {
       env: {
         CURRENT_VERSION: version,
       },
-      rewrites: () => [
-        {
-          source: '/api/sitemaps/:path*',
-          destination: `${process.env.NEXT_PUBLIC_API_URL}/sitemaps/:path*`,
-        },
-        {
-          source: '/search',
-          destination: '/search/:provider',
-          has: [
-            {
-              type: 'query',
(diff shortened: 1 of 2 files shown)


packages/webapp/next.config.js
```diff
@@ -83,26 +83,38 @@ module.exports = {
...
+            ]
+          },
+          {
+            source: '/search',
+            destination: '/search/posts'
+          }
+        ];
+
+        // to support GitPod environment and avoid CORS issues, we need to proxy the API requests
+        if (process.env.NEXT_PUBLIC_DOMAIN === 'localhost') {
```
rebelchris, line 109: This env only exists in gitpod? Need to be 100% sure it doesn't accidentally got set on prod infra 🙈

rebelchris (approved):

nensidosari (approved):

packages/webapp/next.config.js
sshanzel: This is also used in prod but with a different value. So we should be safe 🥳 

Screenshot 2024-10-03 at 2 46 27 PM

PR #3617, opened 2024-10-03
feat: shuffle onboarding tags experiment

omBratteng:
## Changes

Micro experiment to shuffle tags during onboarding.

Needs API to be merged first
https://github.com/dailydotdev/daily-api/pull/2308

## Events

## Experiment

## Manual Testing

> [!CAUTION]

AS-619

### Preview domain
https://as-619-exp-random-tags.preview.app.daily.dev


packages/shared/src/components/onboarding/FilterOnboardingV4.tsx
```diff
@@ -63,6 +65,7 @@ export function FilterOnboardingV4({
     feedId,
     shouldFilterLocally,
   });
+  const shouldShuffleTags = useFeature(feature.onboardingShuffleTags);
```
rebelchris, line 68: This file is also used for custom feeds, we need to ensure those users don't get enrolled/evaluated or see this.

rebelchris: Maybe we can evaluate it on `EditTag` component level?

rebelchris (changes requested):
Small thing :)

packages/shared/src/components/onboarding/FilterOnboardingV4.tsx
omBratteng: Yeah, could probably move it up in the tree to onboarding specific. Or use router to check if path is onboarding

nensidosari (approved):
Looks great with these changes!! Discussing some more possible refactoring we can do with Chris in DMs.

omBratteng:
Can't approve changes since it's my PR. But lgtm @rebelchris

packages/shared/src/components/tags/TagSelection.tsx
```diff
@@ -102,6 +104,15 @@ export function FilterOnboardingV4({
...
     },
   );
+
+  const onboardingTags = useMemo(() => {
+    if (!shouldShuffleTags) {
+      return onboardingTagsRaw;
+    }
+
+    return onboardingTagsRaw?.sort(() => Math.random() - 0.5);
+  }, [shouldShuffleTags, onboardingTagsRaw]);
```
capJavert, line 114: One thing that comes to mind if this is SSR-ed it could point to server client mismatch because it randomises once on the server and once on the client. But I don't think its rendered on server side?

PR #3618, opened 2024-10-03
fix: url for extension

nensidosari:
## Changes

Url not working in extension

### Preview domain
https://fix-url-extension.preview.app.daily.dev

### DIFF
--- a/packages/shared/src/components/sidebar/Sidebar.tsx
+++ b/packages/shared/src/components/sidebar/Sidebar.tsx
@@ -48,7 +48,7 @@ import { IconSize } from '../Icon';
 import { CreatePostButton } from '../post/write';
 import useActiveNav from '../../hooks/useActiveNav';
 import { useFeatureTheme } from '../../hooks/utils/useFeatureTheme';
-import { squadCategoriesPaths, webappUrl } from '../../lib/constants';
+import { webappUrl } from '../../lib/constants';
 import { AlertColor, AlertDot } from '../AlertDot';
 
 const SidebarOnboardingChecklistCard = dynamic(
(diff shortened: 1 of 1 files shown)


packages/shared/src/components/sidebar/Sidebar.tsx
```diff
@@ -88,8 +88,8 @@ export default function Sidebar({
   });
   const hasSquads = squads?.length > 0;
   const squadsUrl = hasSquads
-    ? squadCategoriesPaths['My Squads']
-    : squadCategoriesPaths.discover;
+    ? `${webappUrl}squads/discover/my`
+    : `${webappUrl}squads/discover`;
```
rebelchris, line 92: It's probably safer to use the defined constant in case we ever change them.

```suggestion
    ? `${webappUrl}${squadCategoriesPaths['My Squads']}`
    : `${webappUrl}${squadCategoriesPaths.discover}`;
```

rebelchris (changes requested):
Minor request for safety :) 
I see we do use this `squadCategoriesPath` elsewhere not sure it might be an issue somewhere else? (probably not due to no render)

packages/shared/src/components/sidebar/Sidebar.tsx
nensidosari: but the constant has an extra / in the beginning, which isn't a big deal but might cause anything unexpected

sshanzel: Also, there is an extra `/` character if we use both together. Although this is for extension and the links here are only used on tablet. Shouldn't be too much of an issue I believe.

rebelchris: We can maybe use some string replacement rather?
I just want to avoid having magic strings that are actually const elsewhere 🙈

sshanzel: Yeah, I agree. Always better to use defined consts.

rebelchris (approved):
That works :)

PR #3619, opened 2024-10-03
chore: cleanup unused flow

rebelchris:
## Changes

We had some legacy flow for onboarding via post page (this got changed to uniform onboarding page)
However some elements still referred to this flow although it should't exist.

The only edge-case was the `registerLocalFilters` function, which was called on `onboarding` page if user didn't select tags. It's technically impossible, but saw in clickhouse there's about 3 a week that have this. (Not sure how)

Removing it doesn't break anything though as their flow still works.

Related discussion:
https://dailydotdev.slack.com/archives/C01L0QXQJNB/p1727951097888609

## Events

## Experiment

## Manual Testing

> [!CAUTION]

AS-631 #done 

### Preview domain
https://as-631-cleanup-unused-code.preview.app.daily.dev


omBratteng (approved):

packages/shared/src/components/modals/ArticlePostModal.tsx
```diff
@@ -24,11 +23,11 @@ export default function ArticlePostModal({
   onRemovePost,
   ...props
 }: ArticlePostModalProps): ReactElement {
-  const { showArticleOnboarding } = useContext(OnboardingContext);
   const { position, onLoad } = usePostNavigationPosition({
     isDisplayed: props.isOpen,
-    offset: showArticleOnboarding ? ONBOARDING_OFFSET : 0,
+    offset: 0,
   });
+  console.log('position', position);
```
rebelchris, line 30:
```suggestion
```

packages/storybook/stories/components/HorizontalScroll.stories.tsx
```diff
@@ -2,7 +2,6 @@ import type { Meta, StoryObj } from '@storybook/react';
 import HorizontalScroll from '@dailydotdev/shared/src/components/HorizontalScroll/HorizontalScroll';
 import React, {
   ReactElement,
-  Ref,
```
rebelchris, line 5: unused

packages/storybook/stories/extension/ShortcutLinks.stories.tsx
```diff
@@ -1,11 +1,5 @@
 import type { Meta, StoryObj } from '@storybook/react';
-import { HttpResponse, http } from 'msw';
```
rebelchris, line 2: All unused

packages/shared/src/components/Feed.spec.tsx
```diff
@@ -232,23 +231,13 @@ const renderComponent = (
...
-            <Feed
-              feedQueryKey={['feed']}
-              feedName={feedName}
-              query={ANONYMOUS_FEED_QUERY}
-              variables={variables}
-            />
-          </OnboardingContext.Provider>
+          <Toast autoDismissNotifications={false} />
+          <Feed
+            feedQueryKey={['feed']}
```
capJavert, line 236: We have OtherFeed or some constant for this?

capJavert (approved):

packages/shared/src/components/Feed.spec.tsx
rebelchris: The key you mean?
Probably didn't really change anything here 😅

capJavert: non blocking just checking if we have constant in the meantime 👍

rebelchris: I actually checked, but we don't specific that one somewhere at the moment as enum.

PR #3623, opened 2024-10-03
feat: new tab overlay to keep it

sshanzel:
## Changes
- New tab behavior to display some overlay.
- The thing is, we won't be able to detect whether the user clicks "Keep it" or not, so the overlay will have to stay until the user interacts with it.

Preview:

https://github.com/user-attachments/assets/c6186d67-34b6-40c6-87e9-ce68aab8960e

## Events

## Experiment

## Manual Testing

> [!CAUTION]

### Preview domain
https://mi-587.preview.app.daily.dev


capJavert (approved):
Looks ok to me 👍

nensidosari (changes requested):
Looks all good, just one thing missing from the requirements. 

> The overlay will be there for x amount of seconds (5, for instance)

Shall we add this as well?

sshanzel:
> Looks all good, just one thing missing from the requirements.
> 
> > The overlay will be there for x amount of seconds (5, for instance)
> 
> Shall we add this as well?

Right, let's add this.

> Nice love the install cache solution here. Just to check we don't want specific events for this?

Let me check with Ido 👀

nensidosari (approved):

packages/extension/src/newtab/App.tsx
```diff
@@ -134,12 +141,29 @@ function InternalApp(): ReactElement {
     }
   }, [contentScriptGranted]);
 
+  useEffect(() => {
+    if (!user || shouldShowLogin) {
```
idoshamun, line 145: Why do you have this condition?

I'm not sure it's needed, we need to show the overlay on the first tab regardless if the user is logged in or not

packages/extension/src/newtab/KeepItOverlay.tsx
```diff
@@ -0,0 +1,95 @@
...
+    await onDelete();
+    onClose();
+  }, [onDelete, onClose, logEvent]);
+
+  // when this overlay is opened, there is nothing the user can do, clicking anywhere should close it.
+  useEventListener(globalThis?.document, 'click', onClick);
+
+  useEffect(() => {
+    onDelete();
+  }, [onDelete]);
```
idoshamun, line 47: What's this?

idoshamun (commented):
* Shouldn't this be an experiment?
* Can we show the overlay faster? I feel there's a significant delay between the Chrome prompt and our overlay. I'd like to minimize this gap if possible

packages/extension/src/newtab/App.tsx
sshanzel: Right. Our flow is install -> registration -> new tab. But yeah, if the user skips installation, we should still show this. Let me remove it.

packages/extension/src/newtab/KeepItOverlay.tsx
sshanzel: This is to delete the cache once this opens. On the second reload, they have kept the extension already, without this, there could be a possibility of showing the overlay again.

sshanzel:
> Shouldn't this be an experiment?

Have added the right flag now.

> Can we show the overlay faster? I feel there's a significant delay between the Chrome prompt and our overlay. I'd like to minimize this gap if possible

There is just one call now that needs to happen for us to show this, which is the feature flag fetch. We'll do some tests on prod to ensure it shows up as intended. Locally, it works well.

packages/extension/src/newtab/App.tsx
```diff
@@ -134,12 +146,29 @@ function InternalApp(): ReactElement {
...
   }, [contentScriptGranted]);
 
+  useEffect(() => {
+    if (shouldShowLogin) {
+      return;
+    }
+
+    getCache(INSTALLATION_STORAGE_KEY).then((value) => {
+      if (value) {
+        logEvent({ event_name: LogEvent.ShowNewTabPermission });
```
idoshamun, line 156: Let's add the event to the component itself since this can be true while the user isn't part of the experiment

packages/extension/src/newtab/App.tsx
idoshamun: Can you show a video where the user doesn't signup and we still show the overlay please?

idoshamun (approved):

PR #3625, opened 2024-10-04
feat: search focus experiment

rebelchris:
## Changes

Changed to the following states: (Already checked with Tomer)
We also added focus blue ring to default and disable the autofocus on mobile.

Screenshot 2024-10-08 at 09 23 06

## Events

## Experiment

## Manual Testing

> [!CAUTION]

AS-633 #done 

### Preview domain
https://as-633-search-focus-experiment.preview.app.daily.dev

### DIFF
--- a/packages/shared/src/components/fields/SearchField.tsx
+++ b/packages/shared/src/components/fields/SearchField.tsx
@@ -18,6 +18,11 @@ import {
 import { getFieldFontColor } from './BaseFieldContainer';
 import { IconProps } from '../Icon';
 
+export enum SearchStyleVersion {
+  Default = 'default',
+  AlwaysFocus = 'always_focus',
+}
+
 export interface SearchFieldProps
   extends Pick<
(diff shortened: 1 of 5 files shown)


packages/shared/src/components/fields/SearchField.tsx
```diff
@@ -136,10 +147,10 @@ export const SearchField = forwardRef(function SearchField(
             aria-hidden
             className="icon mr-2 text-2xl"
             role="presentation"
-            secondary={focused}
+            secondary={focused || isFocusVersion}
             style={{
               color:
-                focused || hasInput
+                focused || hasInput || isFocusVersion
```
capJavert, line 153: I agree, there should be other ways to make search more prominent without forcing focus state.

nensidosari (approved):
Code wise looks ok if we decide to move forward with this

nensidosari (approved):
Looks good

PR #3627, opened 2024-10-04
feat: Post approval switch

AmarTrebinjac:
## Changes
Implemented a switch for letting admins toggle if they want to approve every single user post or not.

## Events

## Experiment

## Manual Testing

> [!CAUTION]

MI-571

### Preview domain
https://mi-571.preview.app.daily.dev


packages/shared/src/components/squads/settings/SquadModerationSettingsSection.tsx
```diff
@@ -0,0 +1,117 @@
...
+              disabled={memberPostingRole === SourceMemberRole.Moderator}
+              checked={moderationRequired}
+              onToggle={() => setModerationRequired(!moderationRequired)}
+            >
+              <span>{moderationRequired ? 'On' : 'Off'}</span>
+            </Switch>
+          </SimpleTooltip>
+        </SquadSettingsSection>
+        <SquadSettingsSection
+          title="Invite others"
```
AmarTrebinjac, line 104: I took the liberty to update this title as well, as it used to say "Invitation others", which is not grammatically correct

packages/shared/src/components/squads/settings/SquadModerationSettingsSection.tsx
```diff
@@ -0,0 +1,117 @@
...
+import { SquadSettingsSection } from './SquadSettingsSection';
+import { Radio } from '../../fields/Radio';
+import { SourceMemberRole } from '../../../graphql/sources';
+import { Switch } from '../../fields/Switch';
+import { WidgetCard } from '../../widgets/WidgetCard';
+import { SimpleTooltip } from '../../tooltips';
+import { useViewSize, ViewSize } from '../../../hooks/useViewSize';
+import { MiniCloseIcon } from '../../icons';
+
+interface PermissionSectionProps {
```
ilasw, line 11: We should maintain the name consistent, maybe rename into `SquadModerationSettingsSectionProps`?

packages/shared/src/components/squads/settings/SquadModerationSettingsSection.tsx
```diff
@@ -0,0 +1,117 @@
...
+    value: SourceMemberRole.Moderator,
+  },
+];
+
+export function SquadModerationSettingsSection({
+  initialMemberInviteRole,
+  initialMemberPostingRole,
+  initialModerationRequired,
+}: PermissionSectionProps): ReactElement {
+  const [tooltipDismissed, setTooltipDismissed] = useState(false);
```
ilasw, line 33: if we want to manage a boolean we can use `useToggle` for it ;)

packages/shared/src/components/squads/settings/SquadModerationSettingsSection.tsx
```diff
@@ -0,0 +1,117 @@
...
+        >
+          <SimpleTooltip
+            placement={isMobile ? 'bottom' : 'right'}
+            interactive
+            content={
+              <button
+                className="flex cursor-pointer"
+                type="button"
+                onClick={() => setTooltipDismissed(true)}
+              >
```
rebelchris, line 79: I don't think we need this right?
Also isn't this the latest version?

rebelchris (changes requested):
Don't think we need the manual button on the tooltip :)

packages/shared/src/components/squads/settings/SquadModerationSettingsSection.tsx
```diff
@@ -0,0 +1,117 @@
...
+}: PermissionSectionProps): ReactElement {
+  const [tooltipDismissed, setTooltipDismissed] = useState(false);
+  const isMobile = useViewSize(ViewSize.MobileL);
+  const [memberPostingRole, setMemberPostingRole] = useState(
+    () => initialMemberPostingRole || SourceMemberRole.Moderator,
+  );
+  const [memberInviteRole, setMemberInviteRole] = useState(
+    () => initialMemberInviteRole || SourceMemberRole.Member,
+  );
+  const [moderationRequired, setModerationRequired] = useState(
```
ilasw, line 41:
```ts
const [moderationRequired, toggleModeration] = useToggle(initialModerationRequired);

```

ilasw (changes requested):
Overall seems good! Just small suggestions

packages/shared/src/components/squads/settings/SquadModerationSettingsSection.tsx
AmarTrebinjac: You are entirely correct! 

Changing

packages/shared/src/components/squads/settings/SquadModerationSettingsSection.tsx
```diff
@@ -0,0 +1,117 @@
...
+                <span className="mr-1">
+                  Cannot be enabled in moderator only mode
+                </span>
+                <MiniCloseIcon />
+              </button>
+            }
+            visible={
+              !tooltipDismissed &&
+              memberPostingRole === SourceMemberRole.Moderator
+            }
```
nensidosari, line 89: This should only be visible on switch hover if the condition is right, based on the specs docs.

nensidosari (changes requested):

packages/shared/src/components/squads/settings/SquadModerationSettingsSection.tsx
AmarTrebinjac: Agreed, I missed this one. Thanks!

packages/shared/src/components/squads/settings/SquadModerationSettingsSection.tsx
```diff
@@ -0,0 +1,107 @@
...
+            visible={
+              isHoveringSwitch &&
+              memberPostingRole === SourceMemberRole.Moderator
+            }
+          >
+            <span
+              className="max-w-fit cursor-pointer"
+              onMouseEnter={() => setIsHoveringSwitch(true)}
+              onMouseLeave={() => setIsHoveringSwitch(false)}
+            >
```
AmarTrebinjac, line 79: Wrapped the switch in a span because disabled inputs do not fire events.

packages/shared/src/graphql/squads.ts
```diff
@@ -461,6 +470,10 @@ const formToInput = (form: SquadForm): SharedSquadInput => ({
   memberPostingRole: form.memberPostingRole,
   memberInviteRole: form.memberInviteRole,
   categoryId: form.categoryId,
+  moderationRequired:
+    form.memberPostingRole === SourceMemberRole.Moderator
+      ? false
+      : form.moderationRequired,
```
AmarTrebinjac, line 476: According to design we want the switch to still show "on" when it's disabled, so we have to check manually if the posting role is set to "moderators"

packages/shared/src/components/fields/Switch.tsx
```diff
@@ -41,6 +41,7 @@ function SwitchComponent(
         className,
         'group relative flex items-center',
         disabled ? 'pointer-events-none cursor-not-allowed' : 'cursor-pointer',
+        disabled && checked && 'opacity-16',
```
AmarTrebinjac, line 44: Currently when the switch is disabled and checked, there isn't any darkening effect, so we add it here.

packages/shared/src/components/squads/settings/SquadModerationSettingsSection.tsx
```diff
@@ -0,0 +1,107 @@
...
+          />
+        </SquadSettingsSection>
+        <SquadSettingsSection
+          title="Require post approval"
+          description="Turn this on to have admins or moderators approve every new post before it's published."
+        >
+          <SimpleTooltip
+            placement="top-start"
+            content={
+              <span className="p- max-w-[188px] p-2 text-center">
```
rebelchris, line 65: I don't think you need this whole span anyway? Just render as is.
The tooltip should handle the rendering for plain text ones.

packages/shared/src/components/squads/settings/SquadModerationSettingsSection.tsx
rebelchris: I don't really see why this is needed?
The tooltip should just show automatically on hovering any type of element?

packages/shared/src/graphql/squads.ts
rebelchris: That doesn't seem right? 😅
Might just be a design mistake no?

packages/shared/src/components/squads/settings/SquadModerationSettingsSection.tsx
```diff
@@ -0,0 +1,107 @@
...
+  initialModerationRequired?: boolean;
+}
+
+const memberRoleOptions = [
+  {
+    label: 'All members (recommended)',
+    value: SourceMemberRole.Member,
+  },
+  {
+    label: 'Only Moderators',
```
rebelchris, line 22: We never do this, expect for product terms (Squad for instance)
```suggestion
    label: 'Only moderators',
```

rebelchris (changes requested):
Some minor feedback, but almost there.

packages/shared/src/components/squads/settings/SquadModerationSettingsSection.tsx
AmarTrebinjac: In figma it says "the moderation settings sections include text changes, and new "require post approval" option.", There's multiple texts that don't match with live, so I assumed it referred to all text changes. "Moderators" is capitalized in the design

packages/shared/src/components/squads/settings/SquadModerationSettingsSection.tsx
AmarTrebinjac: Yes, if there is no condition set in the "visible" property, it will automatically show on hover on any type of element.

However, as mentioned, disabled inputs do not trigger events, so the "onMouseEnter" event is not being triggered by its child.

We also additionally only want to show the hover if moderation role is set to "moderator" _and_ the switch is disabled

nensidosari: Not sure but in these cases maybe we want to suggest to design that it would be best if for these things we had any ℹ️ icon in the section that if user hovers on it, it shows the tooltip with the information. 
Rather then hack these things where it might not make too much sense 🤔

packages/shared/src/components/squads/settings/SquadModerationSettingsSection.tsx
AmarTrebinjac: > I don't think you need this whole span anyway? Just render as is. The tooltip should handle the rendering for plain text ones.

You're right. I did it this way to make it look more like the design:
Screenshot 2024-10-04 at 14 48 31

the design:
Screenshot 2024-10-04 at 14 52 45



Without the span it looks like this:
Screenshot 2024-10-04 at 14 48 19

Also on mobile the text is a little bit too long, so it creates an overflow and horizontal scroll.

packages/shared/src/components/squads/settings/SquadModerationSettingsSection.tsx
AmarTrebinjac: > Not sure but in these cases maybe we want to suggest to design that it would be best if for these things we had any ℹ️ icon in the section that if user hovers on it, it shows the tooltip with the information. Rather then hack these things where it might not make too much sense 🤔

I agree. I did wonder if this would've been better off as a "hint" below the switch, but an info icon works well, too

nensidosari:
@AmarTrebinjac one more thing, when I click on the preview link, no feed loads, is it because backend is not in place yet?

AmarTrebinjac:
> @AmarTrebinjac one more thing, when I click on the preview link, no feed loads, is it because backend is not in place yet?

@nensidosari Correct! This is why me and 

> @AmarTrebinjac one more thing, when I click on the preview link, no feed loads, is it because backend is not in place yet?

@nensidosari Correct.  It's erroring because graphql is querying a field that does not exist yet. 😅

This is why me and @ilasw decided to merge to a intermediate branch before going into main

packages/shared/src/components/squads/settings/SquadModerationSettingsSection.tsx
rebelchris: Hmm yeah probably something we should fix with our new context menu (if we get one)
To make it more inline maybe use this approach instead:
https://github.com/dailydotdev/apps/blob/ee7c37145463d0988b081608d3217e05e2fd2349/packages/shared/src/components/streak/popup/DayStreak.tsx#L79

Also maybe you can use the `show` variable to determine if it needs to render (as shown above in the example)

AmarTrebinjac: > Hmm yeah probably something we should fix with our new context menu (if we get one) To make it more inline maybe use this approach instead:
> 
> https://github.com/dailydotdev/apps/blob/ee7c37145463d0988b081608d3217e05e2fd2349/packages/shared/src/components/streak/popup/DayStreak.tsx#L79
> 
> Also maybe you can use the `show` variable to determine if it needs to render (as shown above in the example)

@rebelchris Great tips!

I changed it to use the container instead of the span.

When it comes to the hover trigger: With your solution it allows us to remove the state, but the span needs to remain for the events

Great feedback!

packages/shared/src/components/squads/settings/SquadModerationSettingsSection.tsx
```diff
@@ -0,0 +1,106 @@
...
+        </SquadSettingsSection>
+        <SquadSettingsSection
+          title="Require post approval"
+          description="Turn this on to have admins or moderators approve every new post before it's published."
+        >
+          <SimpleTooltip
+            placement="top-start"
+            content="
+                Only admins and moderators can post; their posts are
+                auto-published.
```
rebelchris, line 73: The formatting is a bit weird here?

rebelchris (approved):
Nice, just formatting, but looks great :)

nensidosari (approved):
Amazing!

packages/shared/src/components/squads/settings/SquadModerationSettingsSection.tsx
```diff
@@ -0,0 +1,103 @@
...
+];
+
+export function SquadModerationSettingsSection({
+  initialMemberInviteRole,
+  initialMemberPostingRole,
+  initialModerationRequired,
+}: SquadModerationSettingsSectionProps): ReactElement {
+  const [isHoveringSwitch, setIsHoveringSwitch] = useState(false);
+  const [memberPostingRole, setMemberPostingRole] = useState(
+    () => initialMemberPostingRole || SourceMemberRole.Moderator,
```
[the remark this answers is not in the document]
AmarTrebinjac: > IMO the initial state should not be a function, just the value is fine!

This is only showing up as a change since I renamed the component, as it was like this from before, but I can do a little refactoring while I'm at it :)

packages/shared/src/components/squads/settings/SquadModerationSettingsSection.tsx
```diff
@@ -0,0 +1,103 @@
...
+    value: SourceMemberRole.Moderator,
+  },
+];
+
+export function SquadModerationSettingsSection({
+  initialMemberInviteRole,
+  initialMemberPostingRole,
+  initialModerationRequired,
+}: SquadModerationSettingsSectionProps): ReactElement {
+  const [isHoveringSwitch, setIsHoveringSwitch] = useState(false);
```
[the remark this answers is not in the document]
AmarTrebinjac: We do not!

Removed

PR #3630, opened 2024-10-05
feat: Image drag and drop for ImageInput

Abhi-Bohora:
## Changes
Linked Issue: (https://github.com/dailydotdev/daily/issues/1369)

As said in the issue `We have an ImageInput for uploading images (as the name implies), but it lacks drag-and-drop functionality`. so this means the user cannot simply drag and drop their images so this PR aims to fix that

Any suggestions, Improvements 🙏

https://github.com/user-attachments/assets/583804c7-6c3b-447c-bcef-6589940b996e

## Events

## Experiment

## Manual Testing

> [!CAUTION]


packages/shared/src/components/fields/ImageInput.tsx
```diff
@@ -102,6 +103,32 @@ function ImageInput({
     onChange?.(base64, file);
   };
 
+  const onDragOver = (event: DragEvent<HTMLButtonElement>) => {
+    event.preventDefault();
+  };
+
+  const onDrop = async (event: DragEvent<HTMLButtonElement>) => {
```
nensidosari, line 110: Seems like both onFileChange and onDrop do almost the same things after they have the file. Would be nice to unify these functions rather then repeat code.

nensidosari:
Hi @Abhi-Bohora thank you for your contribution. Have left just one small comment regarding code, other then that things seem good :D

sshanzel (approved):
I don't have anything to add on top of Nensi's comment so I'm approving. Really looks amazing with the screen recording 😍

packages/shared/src/components/fields/ImageInput.tsx
sshanzel: Maybe introduce a function named `validateFileInput` or something to contain the logic.

PR #3638, opened 2024-10-08
fix: remove google auth from in app browsers

nensidosari:
## Changes
- remove google auth from in app browsers

MI-594 #done

### Preview domain
https://mi-594-remove-google-inapp.preview.app.daily.dev


omBratteng (approved):
Interesting issue

PR #3640, opened 2024-10-08
fix: detect more inapp web views

nensidosari:
## Changes
- detect more inapp web views

MI-594 #done

### Preview domain
https://mi-594-remove-google-inapp.preview.app.daily.dev


rebelchris (approved):
Maybe we can copy some of this logic though?
think the safari ones are a big one as well:
https://github.com/f2etw/detect-inapp/blob/master/src/inapp.js

rebelchris (approved):
looks good to me :) 
Perhaps just try running on actual iPhone (Chrome/Safari) that those still show it :)

nensidosari:
This "hasLimitedFeatures" actually has false positive when in private mode - will see now if I can avoid that, else I'll remove

idoshamun (approved):

nensidosari:
Tested in web desktop/mobile for firefox/edge/chrome/safari
Will test more when deployed when opening in app

PR #3641, opened 2024-10-09
feat: squad post approval

ilasw:
## Changes
- Added post approval switch in squad settings
- 

## Events

## Experiment

## Manual Testing

> [!CAUTION]


packages/webapp/pages/squads/[handle]/moderate.tsx
```diff
@@ -0,0 +1,101 @@
...
+  handle: string;
+}
+
+export async function getServerSideProps({
+  params,
+  res,
+}: GetServerSidePropsContext<SquadPageParams>): Promise<
+  GetServerSidePropsResult<ModerateSquadPageProps>
+> {
+  const setCacheHeader = () => {
```
Comment (rebelchris) [inline packages/webapp/pages/squads/[handle]/moderate.tsx:75]: This might cause an issue if you change the setting of the squad?
Not sure but check here:
https://dailydotdev.slack.com/archives/C07GFU7DX8X/p1731565434782739

Comment (sshanzel) [reply packages/webapp/pages/squads/[handle]/moderate.tsx:75]: Right. We should maybe leave this one out since it won't really matter to have this cached, or at least lower the time.

Comment (ilasw) [reply packages/webapp/pages/squads/[handle]/moderate.tsx:75]: I think that, most of the time, users will have already fetched the squad client-side by the time they land on the /moderate page, and the squad query is usually already freshly cached.

In my opinion, we could try making this page client-side only and remove this fetch since we mainly want fresh data. What do you think?

packages/shared/src/components/post/SharePostContent.tsx
```diff
@@ -39,82 +41,98 @@ const SharePostContentSkeleton = () => (
...
   };
 
-  if (!post.sharedPost) {
+  if (!sharedPost) {
     return <SharePostContentSkeleton />;
   }
 
   const shouldUseInternalLink =
-    isSharedPostSquadPost(post) || isInternalReadType(post.sharedPost);
+    isSharedPostSquadPost({ sharedPost }) || isInternalReadType(sharedPost);
```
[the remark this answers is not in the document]
ilasw: This was changed because SharePostContent is now used both on real posts and pending ones (in preview modal). 
Everything is marked as changes, but seems that the component is the same, just split between the common part and the full-feature one. 

image

packages/shared/src/components/squads/moderation/SquadModerationItemContextMenu.tsx
```diff
@@ -0,0 +1,58 @@
...
+      <OptionsButton
+        onClick={onMenuClick}
+        className="z-1 !my-0"
+        tooltipPlacement="right"
+      />
+      <ContextMenu
+        options={[
+          // {
+          //   label: 'Edit post',
+          //   action: () => console.log('Edit'), // todo: implement after SourcePostModeration edit page is created
```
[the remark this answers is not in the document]
ilasw: Yes, we need to redirect to edit page once is implemented ( https://github.com/dailydotdev/apps/pull/3796 )

packages/shared/src/hooks/usePostContent.ts
```diff
@@ -26,14 +29,35 @@ export interface UsePostContentProps {
   post: Post;
 }
 
+interface UseReadArticle {
+  origin: Origin;
+  post: Post | SharedPost;
+}
+
+export const useReadArticle = ({
```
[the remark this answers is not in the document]
sshanzel: This is just the callback on what happens, the event listener is handle in the component level. This is so we don't have to copy-paste everything and might miss out the analytics part.

packages/shared/src/components/post/common/SharedPostLink.tsx
```diff
@@ -1,33 +1,35 @@
...
+  mainSource: Post['source'];
   children: ReactNode;
   className?: string;
   onGoToLinkProps?: CombinedClicks;
 }
 
 export const SharedPostLink = ({
-  post,
+  sharedPost,
+  mainSource,
```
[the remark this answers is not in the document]
ilasw: We don't have any variable named `source` here and we can distinguish between the main and the shared because we have to use `sharedPost.source` for the second one. 
I'm trying to change this just to stay consistent through the codebase ✔️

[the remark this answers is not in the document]
rebelchris: source and sharedPostSource would be better then wouldn't you agree?

packages/shared/src/hooks/squads/usePostToSquad.ts
```diff
@@ -67,25 +89,83 @@ export const usePostToSquad = ({
   const { completeAction } = useActions();
   const [preview, setPreview] = useState(initialPreview);
   const { requestMethod } = useRequestProtocol();
+
+  const {
+    mutateAsync: onCreatePost,
+    isPending: isLoadingFreeform,
```
rebelchris, line 95: Ideally we shouldn't recast to loading to avoid confusion between devs.
But non blocking

packages/shared/src/hooks/squads/usePostToSquad.ts
```diff
@@ -142,24 +222,42 @@ export const usePostToSquad = ({
   });
 
   const isPosting =
```
rebelchris, line 224: Jooooh this file is insanity guys 😅
I'm gonna approve it, but it's quite messy

rebelchris (approved):
From my in person tests it looks good.
Just find some files very overwhelming to review and can't really go in details in this PR anymore.

Maybe something to consider moving forward, rather split stuff up in smaller hooks that can re-use each other?
@capJavert did something with useIntegration for example. (maybe have a look there)

packages/shared/src/hooks/squads/usePostToSquad.ts
AmarTrebinjac: Agree, I think it would be better for each posting type to have its own hook

PR #3644, opened 2024-10-09
fix: form content lost when switching tabs

Abhi-Bohora:
## Changes
Linked Issue: (https://github.com/dailydotdev/daily/issues/1261)

The issue here is that when we switch the tabs between `New Post` and `Share a link` the form content in the `Share a link` gets lost if we navigate to `New Post` and again navigate back to `Share a link`.  By default, shouldMountInactive is set to `false`. This means that when a tab is inactive, its content is unmounted from the DOM. So I have come up with this fix. But not sure if this is a good approach or not.

Before

https://github.com/user-attachments/assets/448d80ce-1313-418a-b8cf-a2610f2e6114

After

https://github.com/user-attachments/assets/a42ca0c8-9eb1-4c83-b7f0-7afe4a9d3958

## Events

## Experiment

## Manual Testing

> [!CAUTION]

### DIFF
--- a/packages/webapp/pages/squads/create.tsx
+++ b/packages/webapp/pages/squads/create.tsx
@@ -191,7 +191,7 @@ function CreatePost(): ReactElement {
         <TabContainer<WriteFormTab>
           onActiveChange={(active) => setDisplay(active)}
           controlledActive={display}
-          shouldMountInactive={false}
+          shouldMountInactive={true}
           className={{ header: 'px-1' }}
           showHeader={isTablet}
         >


packages/webapp/pages/squads/create.tsx
```diff
@@ -191,7 +191,7 @@ function CreatePost(): ReactElement {
         <TabContainer<WriteFormTab>
           onActiveChange={(active) => setDisplay(active)}
           controlledActive={display}
-          shouldMountInactive={false}
+          shouldMountInactive={true}
```
sshanzel, line 194: > By default, shouldMountInactive is set to false. This means that when a tab is inactive, its content is unmounted from the DOM. So I have come up with this fix. But not sure if this is a good approach or not.

Normally we do this just to be more optimized than having the elements rendered in the background, but in cases where we need the state to be consistent, we turn this to `true`. So this means, the approach was not wrong.

sshanzel:
Also, feel free to comment on the main issue so we can assign it to you and avoid other amazing contributors to work on it at the same time 🙌

packages/webapp/pages/squads/create.tsx
Abhi-Bohora: Yes, I thought so if there were many tabs then it would be slow to keep this always true. But in the case of just two tabs, this shouldn't be slow that was my thinking

Abhi-Bohora:
> Really good spot. This can work. My only concern at first was, why the `draft` process is not working as intended since we store the data on what you have typed, it should be available on the cached draft.

We have to dig deeper I think for that but for now, I was just looking at this from the point of state inconsistencies between two tabs or forms(going back and forth between them)

Abhi-Bohora:
> Also, feel free to comment on the main issue so we can assign it to you and avoid other amazing contributors to work on it at the same time 🙌

Sure thanks

packages/webapp/pages/squads/create.tsx
sshanzel: Yeah, this should be good enough for this fix 🚀

sshanzel (approved):
Looks good to me! 🤩 

cc @rebelchris @nensidosari

rebelchris (approved):

PR #3648, opened 2024-10-09
fix: Use slug instead of id when clicking category

AmarTrebinjac:
## Changes

The category in squad header was using the squad id instead of the slug, which would result in an empty discover page.

## Events

## Experiment

## Manual Testing

> [!CAUTION]

MI-592

### Preview domain
https://mi-592.preview.app.daily.dev


sshanzel (approved):
🚀

rebelchris (approved):

PR #3649, opened 2024-10-10
fix: Move create new squad to bottom of dropdown

AmarTrebinjac:
## Changes
 "create new squad" option in the squad dropdown of "new post" is now the last option

## Events

## Experiment

## Manual Testing

> [!CAUTION]

MI-591

### Preview domain
https://mi-591.preview.app.daily.dev


packages/webapp/pages/squads/create.tsx
```diff
@@ -120,7 +120,10 @@ function CreatePost(): ReactElement {
   const { onCreateSquad, isLoading } = useSquadCreate({
     onSuccess: (newSquad) => {
       const form = formToJson<CreatePostProps>(formRef.current);
-
+      /* If the squad gets created, the squad list gets updated and then the selected index becomes the last squad in the list.
+      Therefore on success, we set it to the first item in the list, which is the newly created squad.
+      */
+      setSelected(0);
```
AmarTrebinjac, line 126: This is mostly just to make the UI not look janky, as the time between the squad being created and navigation happening might take a second, which causes this to happen:


https://github.com/user-attachments/assets/f1df7f38-1436-498d-b44b-0792088086a8

packages/webapp/pages/squads/create.tsx
```diff
@@ -121,6 +121,7 @@ function CreatePost(): ReactElement {
     onSuccess: (newSquad) => {
       const form = formToJson<CreatePostProps>(formRef.current);
 
+      setSelected(selected + 1);
```
AmarTrebinjac, line 124: Moving the "create new squad" option to the bottom creates some jank in the UI if the navigation is slow after creating the post:

https://github.com/user-attachments/assets/a6895a54-bcde-4c54-b99d-e9e3b8e10530

It happens because the list of squads gets updated before the navigation happens.

Updating the selected index was the quickest and easiest fix. It works because the update of the list of squads and setSelected are batched together.

AmarTrebinjac:
Moving the "create new squad" option to the bottom created some jank in the UI. The squad list gets updated, therefore the index changes to a different squad for a second before navigating: 

https://github.com/user-attachments/assets/5cf42df0-e787-44ea-8aba-e10b20d1dbdc

Will try to fix tomorrow

packages/webapp/pages/squads/create.tsx
AmarTrebinjac, line 124: A minor hack for aesthetic purposes.

Moving the "create new squad" option to the bottom created some funky UI behavior. The navigation often takes a second before it happens after the squad is created. This results in the wrong squad being shown as selected for a brief moment as the list grows, but the selected index remains the same:


https://github.com/user-attachments/assets/ce595de2-49b6-41cb-b93c-27acc4f89bdd



There's certainly a root problem that should/could be addressed here, but I think this is an OK quick fix for now.

nensidosari: 😬

nensidosari: I would actually avoid the quick fix and this fix in this case then have something not really stable for solving the issue.  🤔 
Also what's also annoying me that I hadn't noticed before is the flickering in the dropdown when u select a value

AmarTrebinjac: @nensidosari The potentially wonky incident I can see is if the squad gets created, but post doesn't. In that case if the user clicks post again, he will create a second squad. But yeah, if you want to avoid the hack, or let this issue rest all together for now it's understandable.

The crux of the issue here is that the dropdown uses indexes and not unique identifiers, but refactoring that seemed like way out of the scope of this task. Also I've seen talk about potentially adopting radix in the future, which would make fixing the dropdown redundant.


> Also what's also annoying me that I hadn't noticed before is the flickering in the dropdown when u select a value

Yeah, seems to be caused by multiple rerenders. Quite unsightly 😅

sshanzel: > Also what's also annoying me that I hadn't noticed before is the flickering in the dropdown when u select a value

That's an on-going issue with our context menu 😅

nensidosari: @AmarTrebinjac the problem we're fixing here, does it mean we fetch squads again in the dropdown now that a new squad is added that causes the weird behaviour?

sshanzel (approved):
Don't have anything blocking here. Looks good to me 🚀

rebelchris (commented):
Isn't it more an issue of `filtered` not being stale?

sshanzel:
@AmarTrebinjac what's the state of this PR? Is it ready for merging?

packages/webapp/pages/squads/create.tsx
sshanzel: @AmarTrebinjac I think we can just remove this line. After creation, I think it won't matter whether the label from "Create new Squad" would become "User's private Squad". The intention was to create a new Squad and the label itself displays the accurate value.

sshanzel (approved):
🚀

PR #3651, opened 2024-10-10
fix: follow button showing by default in user lists

capJavert:
## Changes

Default value for `UserList` was `true` which then shows follow button on places we don't need currently. 

## Events

## Experiment

## Manual Testing

> [!CAUTION]

### Preview domain
https://fix-invite-friends-follow.preview.app.daily.dev

### DIFF
--- a/packages/shared/src/components/profile/UserList.tsx
+++ b/packages/shared/src/components/profile/UserList.tsx
@@ -53,7 +53,6 @@ function UserList({
               href={user.permalink}
               user={user}
               afterContent={afterContent?.(user, i)}
-              showFollow
             >
               {additionalContent?.(user, i)}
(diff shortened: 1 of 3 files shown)


packages/shared/src/components/profile/UserList.tsx
```diff
@@ -53,7 +53,6 @@ function UserList({
               href={user.permalink}
               user={user}
               afterContent={afterContent?.(user, i)}
-              showFollow
```
capJavert, line 56: Now the default is false and I also enabled it on cases where we support following list user list modal and following filter settings.

rebelchris (approved):

sshanzel (approved):

PR #3657, opened 2024-10-10
feat: set default values to false for listicles and memes on senior devs

nensidosari:
## Changes
- check if user has selected any values from 4+ years on experience and update the settings to exclude the listicles and memes
- added experiment flag
- removed legacy ref for content types

## Events

## Experiment

Did you introduce any new experiments? Yes

https://dailydotdev.slack.com/archives/C02JAUF8HJL/p1728569507880959

## Manual Testing

> [!CAUTION]

MI-595 #done

### Preview domain
https://mi-595-senior-dev-custom-content.preview.app.daily.dev


packages/webapp/pages/onboarding.tsx
```diff
@@ -110,7 +123,15 @@ export function OnboardPage(): ReactElement {
   const targetId: string = ExperimentWinner.OnboardingV4;
   const formRef = useRef<HTMLFormElement>();
   const [activeScreen, setActiveScreen] = useState(OnboardingStep.Intro);
-  const enableContentTypeStep = useRef(false);
+  const { updateAdvancedSettings } = useMutateFilters(user);
+  const shouldUpdateAdvancedSettings = useRef(true);
+  const isSeniorUser =
+    EXPERIENCE_TO_SENIORITY[user?.experienceLevel] === 'senior' ||
+    user?.experienceLevel === 'MORE_THAN_4_YEARS';
```
rebelchris, line 130: Why doesn't `EXPERIENCE_TO_SENIORITY` do both cases?

packages/webapp/pages/onboarding.tsx
```diff
@@ -122,18 +143,11 @@ export function OnboardPage(): ReactElement {
       return setActiveScreen(OnboardingStep.EditTag);
     }
 
-    if (
```
rebelchris, line 125: This was before it was rolled out to 100%?
(Can't recall)

packages/webapp/pages/onboarding.tsx
```diff
@@ -179,13 +190,38 @@ export function OnboardPage(): ReactElement {
...
 
+  const isFeedSettingsDefined = useMemo(() => !!feedSettings, [feedSettings]);
+
+  const updateSettingsBasedOnExperience = useCallback(() => {
+    if (isSeniorUser && isFeedSettingsDefined) {
+      shouldUpdateAdvancedSettings.current = false;
+      updateAdvancedSettings({
+        advancedSettings: [
+          { id: 5, enabled: false },
+          { id: 10, enabled: false },
```
rebelchris, line 201: Magic number here, maybe we can use a different mapping rather?

packages/webapp/pages/onboarding.tsx
```diff
@@ -179,13 +190,38 @@ export function OnboardPage(): ReactElement {
...
       experienceLevel: userRefetched?.experienceLevel,
     });
     setActiveScreen(OnboardingStep.EditTag);
   };
 
+  useEffect(() => {
+    if (!shouldUpdateAdvancedSettings.current || !seniorContentOnboarding) {
+      return;
+    }
+    updateSettingsBasedOnExperience();
```
rebelchris, line 218: Do we even need this update here?
Isn't it enough to simply render with some values and on clicking next we persist whatever the state is?

It should then avoid this useEffect and the whole function basically.

rebelchris (commented):
Probably ok, just some sanity checks :)

packages/webapp/pages/onboarding.tsx
nensidosari: Cuz this is not senior 'MORE_THAN_4_YEARS' in that map, and product wanted also this category for experiment. 

We already user EXPERIENCE_TO_SENIORITY in some logs and didn't want to mess with that. So I think after the experiment results we can see what we want to do overall.

packages/webapp/pages/onboarding.tsx
nensidosari: Heh yeah, it just that we rarely seem to change these numbers, and since we depend on them to actually do the updates probably ok to do this. 
I can get these from the titles "Listicles" and "Memes" if u think is best, just thought either way doesn't make much difference since in this case, if we were to change the titles just a little bit then it would still mess up.

packages/webapp/pages/onboarding.tsx
rebelchris: Ah ok no worries, thanks for the cleanup in that case :) 
Makes things a lot easier to follow.

packages/webapp/pages/onboarding.tsx
rebelchris: Nah maybe just define them as `const LISTICLE_ADVANCED_SETTINGS_ID = 10` (just a bit clearer for others).

packages/webapp/pages/onboarding.tsx
nensidosari: What do u mean render with some values? 

There are few points to consider:
1. we've seen scenarios that user can get out of the select tags and whatever onboarding steps that come next meaning if we were to do this somewhere later on we might miss disabling listicles and memes for some devs (of course it wouldn't much impact experiment as we would probably do enrolment only when we can actually disable, but this way we can gain more users to enroll)
2. we can do this when user clicks next on tag selection - probably easy win in this point as we can avoid the use effect, but we'll miss what I mentioned above 
3. we can render the content types with some default values not considering feed settings as default and then update all in the end, but we'll miss what I mentioned above + not a fan of this as the way we are doing things at the moment makes it quite consistent with the feed settings in feed, so wouldn't want to change much.

packages/webapp/pages/onboarding.tsx
nensidosari: Yup makes sense, in a moment

packages/webapp/pages/onboarding.tsx
rebelchris: You can still have the enrollment check, but instead of "pre-pushing" it, we can just change the rendering?
Not sure without looking at that code.

The fact if people abort is a very edge-casey thing we should solve that differently regardless I think.

nensidosari: > but instead of "pre-pushing" it, we can just change the rendering?

I'm still not sure what u mean by this, u mean the rendering on content types, or?

nensidosari: Discussed and decided to move on with removing useEffect and updating the values onClickNext on edit tag page

packages/webapp/pages/onboarding.tsx
```diff
@@ -110,7 +123,30 @@ export function OnboardPage(): ReactElement {
...
   const formRef = useRef<HTMLFormElement>();
   const [activeScreen, setActiveScreen] = useState(OnboardingStep.Intro);
-  const enableContentTypeStep = useRef(false);
+  const { updateAdvancedSettings } = useMutateFilters(user);
+  const isSeniorUser =
+    EXPERIENCE_TO_SENIORITY[user?.experienceLevel] === 'senior' ||
+    user?.experienceLevel === 'MORE_THAN_4_YEARS';
+  const { value: seniorContentOnboarding } = useConditionalFeature({
+    feature: feature.seniorContentOnboarding,
+    shouldEvaluate: isSeniorUser,
```
rebelchris, line 132: We should probably add a state to say only enrol when screen is about to hit.
So default false state, but set to true when onNext passes here:
```
if (seniorContentOnboarding) {
        updateSettingsBasedOnExperience();
      }
```

rebelchris (changes requested):
One final addition and it will be failsafe :)

packages/webapp/pages/onboarding.tsx
nensidosari: changed to using getFeatureValue inside the onClick rather then useConditionalFeature

packages/shared/src/components/onboarding/common.ts
```diff
@@ -19,11 +18,6 @@ export const OnboardingGradientClasses =
   'font-bold text-transparent bg-clip-text bg-gradient-to-r from-accent-bacon-default to-accent-cabbage-default';
 export const OnboardingTitleGradient = classed('h1', OnboardingGradientClasses);
 
-export interface OnboardingStepProps {
```
nensidosari, line 22: unused

rebelchris (approved):
Neat!

PR #3661, opened 2024-10-11
refactor: invalid username error message to clarify the accepter format

PredictiveManish:
### https://github.com/dailydotdev/docs/issues/400 
This issue is solved in this PR.

### This PR changes the error message seen by user during registering and having their username.
- When people select/enter their username- It was showing username is invalid but not showing what is the requirements.
- This issue is raised in the docs repo of daily.dev.

image

### Summary-
This issue will help users to find out name format so that they can easily register and make their account freely.


packages/shared/src/graphql/common.ts
```diff
@@ -119,7 +119,7 @@ export const DEFAULT_ERROR = 'An error occurred, please try again';
 
 export const errorMessage = {
   profile: {
-    invalidUsername: 'Invalid characters found in username!',
+    invalidUsername: 'Username can only contain Alphanumeric (A-z, a-z, 1234..)',
```
sshanzel, line 122: There is actually a specific regex condition we use for checking usernames, and it is not limited to this.

PredictiveManish: Ok, so it is automated or can be changed?

PredictiveManish: image
This one but it is checking the condition and the issue was just showcasing the error message isn't these both are different? Please clear my doubt.

sshanzel: The one we use for usernames is the `handleRegex`

PredictiveManish: Thanks Updating and solving issue!

sshanzel: The message is still inaccurate like mentioned.

PredictiveManish: Sorry, due to some other work I was busy solving it in 1 hour.

PredictiveManish: When i am going to testuser profile page without doing any changes I am facing this below mentioned issue
image
and in code I am facing this issue:
Screenshot 2024-10-15 122504

sshanzel: All cool, no worries. Thank you for trying! 🙏

PR #3663, opened 2024-10-13
fix: Your Website and Mastodon empty update not working

Abhi-Bohora:
## Changes
Linked Issue: (https://github.com/dailydotdev/daily/issues/1396)

In the`https://app.daily.dev/account/profile` in  `Profile Social Links` if the user tries to update the fields `Your Website` and `Mastodon` with empty values to remove their links then the old values would remain even after updating the value with empty values. So this PR aims to fix the issue so that users can remove profile links by setting empty values or update with valid values. other fields were working normally.

## Events

## Experiment

## Manual Testing

> [!CAUTION]


sshanzel (commented):
I think this goes the same to the other options such as twitter, gh, etc. - can you check if it actually is?

Abhi-Bohora:
@sshanzel other field work as expected GitHub, twitter, etc these two fields only have a problem maybe due to a ternary condition set to undefined?

sshanzel (approved):
In that case, then yeah, this should fix the case 💪

sshanzel (commented):
Sorry, one more thing I realized. We can maybe add a `type="url"` to those two inputs to ensure you either fill it or leave it blank.

Abhi-Bohora:
@sshanzel, I just added `type="url"` to those input fields. sorry I didn't understand  if you want to keep this  `portfolio: values.portfolio ? withHttps(values.portfolio) : null` as `undefined`  same as before instead of `null`, if we keep it as `undefined` it didn't work it was same behaviour as before. so I kept it as `null`.

and one more thing I did git rebase because I cannot push changes it was giving me error maybe because there were many commits made already  let me know If this cause any problem. I don't know much about this. Thanks

rebelchris (approved):
Looks worth a try :) 
Perhaps we can try end 2 end on prod once merged?

sshanzel (approved):
Looks amazing! 🚀

sshanzel:
@Abhi-Bohora tested in prod, worked as intended! 🥳

sshanzel:
@Abhi-Bohora can you maybe link a social media profile of yours where I can send you some message?

